PR #2536, opened 2016-12-19
okhttp: catch Throwable to avoid breaking preconditions.

zhangkun83:
TransportSet mandates that transportShutdown() has been called prior to
transportTerminated().  OkHttp transport tries to meet this requirement
by calling startGoAway() which calls transportShutdown() in both the
normal path and the error path.  However, the error path only caught
Exception.  If an Error is thrown from the try-block, calling
transportTerminated() in the finally block will cause a check failure in
TransportSet, shadowing the original Error, which is undesirable for
debugging.  Catching Throwable here is more helpful.

### DIFF
--- a/okhttp/src/test/java/io/grpc/okhttp/OkHttpClientTransportTest.java
+++ b/okhttp/src/test/java/io/grpc/okhttp/OkHttpClientTransportTest.java
@@ -244,6 +244,7 @@ public class OkHttpClientTransportTest {
     frameReader.throwIoExceptionForNextFrame();
     listener1.waitUntilStreamClosed();
     listener2.waitUntilStreamClosed();
+
     assertEquals(0, activeStreamCount());
     assertEquals(Status.UNAVAILABLE.getCode(), listener1.status.getCode());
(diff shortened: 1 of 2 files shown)


carl-mastrangelo:
LGTM, but could you also add a test?

zhangkun83:
@carl-mastrangelo, test added. PTAL.

okhttp/src/test/java/io/grpc/okhttp/OkHttpClientTransportTest.java
```diff
@@ -253,6 +254,30 @@ public void nextFrameThrowIoException() throws Exception {
...
+    initTransport();
+    MockStreamListener listener = new MockStreamListener();
+    OkHttpClientStream stream = clientTransport.newStream(method, new Metadata());
+    stream.start(listener);
+    stream.request(1);
+    assertEquals(1, activeStreamCount());
+    assertContainStream(3);
+    frameReader.throwErrorForNextFrame();
+    listener.waitUntilStreamClosed();
+    assertEquals(0, activeStreamCount());
```
zhangkun83: Done.

Issue #1993, opened 2016-06-29
Implement status_code_and_message interop test in Java


a11r:
There is a spec, and implementations in some languages, but none in Java yet.

ericgribkoff:
This was implemented with #2538.

PR #2538, opened 2016-12-20
interop-testing: Add status_code_and_message interop test

ericgribkoff:
This implements the [status_code_and_message](https://github.com/grpc/grpc/blob/f29de2619476622e61c894559f8885ca3cc8971c/doc/interop-test-descriptions.md#status_code_and_message)  interop test (#1993, https://github.com/grpc/grpc/issues/8481)

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -927,16 +927,16 @@ public abstract class AbstractInteropTest {
 
   @Test(timeout = 10000)
   public void statusCodeAndMessage() throws Exception {
-    final int errorCode = 2;
-    final String errorMessage = "test status message";
-    final EchoStatus responseStatus = EchoStatus.newBuilder()
+    int errorCode = 2;
+    String errorMessage = "test status message";
+    EchoStatus responseStatus = EchoStatus.newBuilder()
         .setCode(errorCode)
         .setMessage(errorMessage)
         .build();
-    final SimpleRequest simpleRequest = SimpleRequest.newBuilder()
+    SimpleRequest simpleRequest = SimpleRequest.newBuilder()
         .setResponseStatus(responseStatus)
         .build();
-    final StreamingOutputCallRequest streamingRequest = StreamingOutputCallRequest.newBuilder()
+    StreamingOutputCallRequest streamingRequest = StreamingOutputCallRequest.newBuilder()
         .setResponseStatus(responseStatus)
         .build();
 
@@ -954,7 +954,8 @@ public abstract class AbstractInteropTest {
 
     // Test FullDuplexCall
     @SuppressWarnings("unchecked")
-    StreamObserver<StreamingOutputCallResponse> responseObserver = mock(StreamObserver.class);
+    StreamObserver<StreamingOutputCallResponse> responseObserver =
+        (StreamObserver<StreamingOutputCallResponse>) mock(StreamObserver.class);
     StreamObserver<StreamingOutputCallRequest> requestObserver
         = asyncStub.fullDuplexCall(responseObserver);
     requestObserver.onNext(streamingRequest);
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -78,6 +78,7 @@ import io.grpc.stub.MetadataUtils;
 import io.grpc.stub.StreamObserver;
 import io.grpc.testing.StreamRecorder;
 import io.grpc.testing.TestUtils;
+import io.grpc.testing.integration.Messages.EchoStatus;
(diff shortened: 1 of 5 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -843,6 +844,50 @@ public void gracefulShutdown() throws Exception {
...
+    if (metricsExpected()) {
+      assertClientMetrics("grpc.testing.TestService/UnaryCall", Status.Code.UNKNOWN);
+    }
+
+    // Test FullDuplexCall
+    @SuppressWarnings("unchecked")
+    StreamObserver<StreamingOutputCallResponse> responseObserver = mock(StreamObserver.class);
+    StreamObserver<StreamingOutputCallRequest> requestObserver
+        = asyncStub.fullDuplexCall(responseObserver);
+    requestObserver.onNext(streamingRequest);
```
ericgribkoff: The test spec requires the client to half-close here. Working on adding this now.

ericgribkoff: Done.

interop-testing/src/main/java/io/grpc/testing/integration/TestServiceImpl.java
```diff
@@ -186,13 +195,22 @@ public void onError(Throwable cause) {
     return new StreamObserver<StreamingOutputCallRequest>() {
       @Override
```
dapengzhang0: I would rather add the boolean field `canceled` in this anonymous `StreamObserver` class than in the `ResponseDispatcher` class. (add `ResponseDispatcher#clear()` if necessary)

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -924,6 +925,51 @@ public void customMetadata() throws Exception {
     }
   }
 
+  @Test(timeout = 10000)
+  public void statusCodeAndMessage() throws Exception {
+    final int errorCode = 2;
```
ericgribkoff: Done

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -924,6 +925,51 @@ public void customMetadata() throws Exception {
...
+      assertEquals(Status.UNKNOWN.getCode(), e.getStatus().getCode());
+      assertEquals(errorMessage, e.getStatus().getDescription());
+    }
+    if (metricsExpected()) {
+      assertClientMetrics("grpc.testing.TestService/UnaryCall", Status.Code.UNKNOWN);
+    }
+
+    // Test FullDuplexCall
+    @SuppressWarnings("unchecked")
+    StreamObserver<StreamingOutputCallResponse> responseObserver = mock(StreamObserver.class);
```
ericgribkoff: Done

interop-testing/src/main/java/io/grpc/testing/integration/TestServiceImpl.java
ericgribkoff: Why is the boolean in the `StreamObserver` preferred? I can do this, but it still requires some kind of check in `ResponseDispatcher#dispatchChunk()`, see below...

interop-testing/src/main/java/io/grpc/testing/integration/TestServiceImpl.java
```diff
@@ -296,10 +315,28 @@ public ResponseDispatcher completeInput() {
...
+    public synchronized boolean isCancelled() {
+      return cancelled;
+    }
+
+    /**
      * Dispatches the current response chunk to the client. This is only called by the executor. At
      * any time, a given dispatch task should only be registered with the executor once.
      */
     private synchronized void dispatchChunk() {
+      if (cancelled) {
```
ericgribkoff: If `canceled` is no longer part of the `ResponseDispatcher` state, all I can do here is check if the queue is empty, e.g.:

```
    private synchronized void dispatchChunk() {
      // If the service already called cancel(), the queue may be empty.
      if (chunks.isEmpty()) {
        return;
      }
      ...
```

It seems like this has the potential to mask other problems with the code by potentially masking cases where the queue is incorrectly empty at this stage. This doesn't seem to be possible with the current code, but still hesitant about adding this kind of check.

interop-testing/src/main/java/io/grpc/testing/integration/TestServiceImpl.java
```diff
@@ -296,10 +315,28 @@ public ResponseDispatcher completeInput() {
     }
 
     /**
+     * Allows the service to cancel the remaining responses.
+     */
+    public synchronized void cancel() {
+      if (cancelled) {
```
dapengzhang0: nit: `Precondition.checkState(!cancelled, "Dispatcher already cancelled");`

ericgribkoff:
retest this please

PR #2539, opened 2016-12-20
compiler: reduce synchronzed invocation

dapengzhang0:
not necessary to synchronze every time calling
getServiceDescriptor(), if the descriptor has been created already;

go with the double-checked locking idom
http://www.oracle.com/technetwork/articles/javase/bloch-effective-08-qa-140880.html
https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java

### DIFF
--- a/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
+++ b/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
@@ -269,25 +269,15 @@ public class BenchmarkServiceGrpc {
     }
   }
 
-  private static io.grpc.ServiceDescriptor serviceDescriptor;
-
-  public static io.grpc.ServiceDescriptor getServiceDescriptor() {
-    if (serviceDescriptor != null) {
-      return serviceDescriptor;
-    }
-    return newServiceDescriptor();
-  }
-
-  private static synchronized io.grpc.ServiceDescriptor newServiceDescriptor() {
-    if (serviceDescriptor != null) {
-      return serviceDescriptor;
-    }
-    io.grpc.ServiceDescriptor serviceDescriptorCopy = new io.grpc.ServiceDescriptor(
+  private static final class LazyServiceDescriptorHolder {
+    static final io.grpc.ServiceDescriptor SERVICE_DESCRIPTOR = new io.grpc.ServiceDescriptor(
         SERVICE_NAME,
         new BenchmarkServiceDescriptorSupplier(),
         METHOD_UNARY_CALL,
         METHOD_STREAMING_CALL);
-    serviceDescriptor = serviceDescriptorCopy;
-    return serviceDescriptor;
+  }
+
+  public static io.grpc.ServiceDescriptor getServiceDescriptor() {
+    return LazyServiceDescriptorHolder.SERVICE_DESCRIPTOR;
   }
 }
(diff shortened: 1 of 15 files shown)


benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
```diff
@@ -271,14 +271,23 @@ public void invoke(Req request, io.grpc.stub.StreamObserver<Resp> responseObserv
...
   private static io.grpc.ServiceDescriptor serviceDescriptor;
 
-  public static synchronized io.grpc.ServiceDescriptor getServiceDescriptor() {
-    if (serviceDescriptor == null) {
-      serviceDescriptor = new io.grpc.ServiceDescriptor(SERVICE_NAME,
-          new BenchmarkServiceDescriptorSupplier(),
-          METHOD_UNARY_CALL,
-          METHOD_STREAMING_CALL);
+  public static io.grpc.ServiceDescriptor getServiceDescriptor() {
+    if (serviceDescriptor != null) {
```
carl-mastrangelo: this is not thread safe.  you would need to make serviceDescriptor volatile, and double checked lock it.

carl-mastrangelo: serviceDescriptor is being concurrently read and modified outside of a lock.  It is not locked here, but locked below.  concurrent calls might get two copies.  That is not thread safe.

dapengzhang0: concurrent calls can not get two copies because `newServiceDescriptor()` is synchronized, the serviceDescriptor field can not be overwritten, only the same instance can be returned.

dapengzhang0:
Ran the qps benchmark. QPS for the first second increased. The result could be better if duration parameter could be a fraction. Updated the PR description.

carl-mastrangelo:
With only one channel, there is going to be very little contention.  Those benchmark results are in the noise.

dapengzhang0:
updated with 10 channel's result.

carl-mastrangelo:
There is still extreme variation in the results.  They are not statistically significant.

ejona86:
I think we want to continue with the double-checked locking. "Soon" we will want to do the same for the `MethodDescriptors` in order to lazily-create them. But using a different class for each `MethodDescriptor` is far too expensive. So we'd either want to use `volatile` or if we need to, avoid synchronization using an object with all-`final` fields. But today I'd expect there is no gain, because the benchmarks and most code should only call the method once and only on the server-side.

The idiom used here works well when used sparingly; it isn't as nice when generating lots of code, because of its class overhead.

dapengzhang0:
now go with the double-checked locking idom, with volatile
http://www.oracle.com/technetwork/articles/javase/bloch-effective-08-qa-140880.html
https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java

compiler/src/java_plugin/cpp/java_generator.cpp
```diff
@@ -905,44 +905,40 @@ static void PrintGetServiceDescriptorMethod(const ServiceDescriptor* service,
...
+      "public static $ServiceDescriptor$ getServiceDescriptor() {\n");
+  p->Indent();
+  p->Print(
+      *vars,
+      "$ServiceDescriptor$ result = serviceDescriptor;\n");
+  p->Print("if (result == null) {\n");
+  p->Indent();
+  p->Print(
+      *vars,
+      "synchronized($service_class_name$.class) {\n");
```
ejona86: Add a space before the `(`

carl-mastrangelo:
I agree with @ejona86 that this will likely have no performance improvement.  I would prefer that the code be left simpler unless there are clear benefits from making it more complicated.

PR #2544, opened 2016-12-21
all: add max message size to client calls

carl-mastrangelo:


core/src/main/java/io/grpc/internal/MessageFramer.java
```diff
@@ -155,6 +166,12 @@ private int writeUncompressed(InputStream message, int messageLength) throws IOE
     }
     BufferChainOutputStream bufferChain = new BufferChainOutputStream();
     int written = writeToOutputStream(message, bufferChain);
+    if (maxOutboundMessageSize >= 0 && written > maxOutboundMessageSize) {
+      throw Status.INTERNAL
+          .withDescription(
+              String.format("message too large %d > %d", written > maxOutboundMessageSize))
```
zhangkun83: The second `>` should be `,`

core/src/main/java/io/grpc/internal/MessageFramer.java
```diff
@@ -169,6 +186,12 @@ private int writeCompressed(InputStream message, int messageLength) throws IOExc
     } finally {
       compressingStream.close();
     }
+    if (maxOutboundMessageSize >= 0 && written > maxOutboundMessageSize) {
+      throw Status.INTERNAL
+          .withDescription(
+              String.format("message too large %d > %d", written > maxOutboundMessageSize))
```
zhangkun83: ditto

core/src/main/java/io/grpc/internal/MessageDeframer.java
```diff
@@ -338,10 +342,10 @@ private void processHeader() {
 
     // Update the required length to include the length of the frame.
     requiredLength = nextFrame.readInt();
-    if (requiredLength < 0 || requiredLength > maxMessageSize) {
+    if (requiredLength < 0 || requiredLength > maxInboundMessageSize) {
       throw Status.INTERNAL.withDescription(String.format("Frame size %d exceeds maximum: %d. "
               + "If this is normal, increase the maxMessageSize in the channel/server builder",
```
zhangkun83: This can be misleading because CallOptions and service config will be able to override the message size set in the builder. Maybe just delete the second sentence?

core/src/main/java/io/grpc/CallOptions.java
```diff
@@ -363,6 +371,70 @@ public boolean isWaitForReady() {
   }
 
   /**
+   * Sets the maximum allowed message size acceptable from the remote peer.  If unset, this will
+   * default to the value set on the {@link ManagedChannel}.
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/FIXME")
```
carl-mastrangelo: Done

core/src/main/java/io/grpc/CallOptions.java
```diff
@@ -363,6 +371,70 @@ public boolean isWaitForReady() {
...
+    newOptions.maxOutboundMessageSize = maxSize;
+    return newOptions;
+  }
+
+  /**
+   * Clears the maximum allowed message size acceptable from the remote peer.  The value set on the
+   * {@link ManagedChannel} may still be enforced.
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/FIXME")
+  public CallOptions withoutMaxInboundMessageSize() {
```
carl-mastrangelo: Using a null to clear the value forces boxing.  This currently does box, but that isn't part of the API; we could change it later to not box.

This is also slightly different than other APIs, because if the call option is not set, the channel default comes into play.  I don't think any of the other call options behave this way.

Thoughts?

core/src/main/java/io/grpc/internal/AbstractStream.java
```diff
@@ -138,6 +138,14 @@ public void endOfStream() {
         statsTraceCtx);
   }
 
+  protected void setMaxInboundMessageSizeProtected(int maxSize) {
```
carl-mastrangelo: Done.

core/src/main/java/io/grpc/internal/MessageFramer.java
carl-mastrangelo: yikes!

carl-mastrangelo:
@zhangkun83  PTAL

core/src/main/java/io/grpc/CallOptions.java
zhangkun83: > Using a null to clear the value forces boxing. This currently does box, but that isn't part of the API; we could change it later to not box.

Why not make boxing part of the API?

> This is also slightly different than other APIs, because if the call option is not set, the channel default comes into play. I don't think any of the other call options behave this way.

`withAuthority()` and `withExecutor()` work this way too. Although `setExecutor()` doesn't have `@Nullable`, it doesn't `checkNotNull` either, so it seems reasonable to add `@Nullable`.

core/src/main/java/io/grpc/internal/MessageDeframer.java
zhangkun83: What about this?

carl-mastrangelo: Done.

core/src/main/java/io/grpc/CallOptions.java
carl-mastrangelo: Boxing means allocation.  Primitives are better API.

carl-mastrangelo:
Added some interop tests PTAL

core/src/main/java/io/grpc/CallOptions.java
zhangkun83: API consistency is also important. I don't want to see two styles of reverting to channel default on the same interface.

On the other hand, I don't think boxing vs primitive in this particular case makes any difference in terms of memory allocation. Suppose we find out the following code allocates so many Integers that we want to get rid of them:
```java
stub.withMaxInboundMessageSize(maxSize).hello(req);
```
It's also allocating that many `CallOptions` at the same time, and we would also have the reason to optimize it out since it's larger than Integer.  The optimization would likely be saving the configured stub instead of configuring it per-call. After we stop allocating that many `CallOptions`, the boxing if Integers will become a non-issue.

carl-mastrangelo: There are not two ways.  There is the primitive way of setting the max, and the clearing without method.  There is no `Integer` way of clearing.  This is consistent with `withWaitForReady` and `withoutWaitForReady`.  You imply that that should be `withWaitForReady(Boolean)`.

It's also Effective Java Item 49 says Prefer primitives to Boxed.

Creating lots of stubs would have a memory problem, but unboxing the integer on every call seems unnecessary.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -785,6 +787,75 @@ public void deadlineInPast() throws Exception {
     }
   }
 
+  @Test(timeout = 10000)
+  public void maxInboundSize_exact() {
+    // warm up the channel and JVM
+    blockingStub.emptyCall(Empty.getDefaultInstance());
+    // 5 is the size of the encoded response proto
+    TestServiceGrpc.TestServiceBlockingStub stub = TestServiceGrpc.newBlockingStub(channel)
+        .withMaxInboundMessageSize(5);
```
zhangkun83: This is fragile, and if the response proto changes, it would be annoying to figure out the new size.

Some test methods construct golden responses. You could do the same and get the serialized size programatically. The actual response can also be compared with the golden response to make sure the response is of the expected size.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -785,6 +787,75 @@ public void deadlineInPast() throws Exception {
...
+    stub.streamingOutputCall(request).next();
+  }
+
+  @Test(timeout = 10000)
+  public void maxInboundSize_tooBig() {
+    // warm up the channel and JVM
+    blockingStub.emptyCall(Empty.getDefaultInstance());
+    // 6 is the size of the encoded response proto
+    TestServiceGrpc.TestServiceBlockingStub stub = TestServiceGrpc.newBlockingStub(channel)
+        .withMaxInboundMessageSize(6 - 1);
```
zhangkun83: ditto

stub/src/main/java/io/grpc/stub/AbstractStub.java
```diff
@@ -189,4 +189,20 @@ public final S withCallCredentials(CallCredentials credentials) {
   public final S withWaitForReady() {
     return build(channel, callOptions.withWaitForReady());
   }
+
+  /**
+   * Returns a new stub that limits the maximum acceptable message size from a remote peer.
```
zhangkun83: Please also mention defaulting to channel setting if not set.

stub/src/main/java/io/grpc/stub/AbstractStub.java
```diff
@@ -189,4 +189,20 @@ public final S withCallCredentials(CallCredentials credentials) {
...
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/2563")
+  public final S withMaxInboundMessageSize(int maxSize) {
+    return build(channel, callOptions.withMaxInboundMessageSize(maxSize));
+  }
+
+  /**
+   * Returns a new stub that limits the maximum acceptable message size to send a remote peer.
```
zhangkun83: ditto.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
carl-mastrangelo: Done.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
carl-mastrangelo: Done.

stub/src/main/java/io/grpc/stub/AbstractStub.java
carl-mastrangelo: Done.

stub/src/main/java/io/grpc/stub/AbstractStub.java
carl-mastrangelo: There actually isn't an equivalent of this on the Channel.   This is new.

zhangkun83: Then `CallOptions.withMaxOutboundMessageSize()` shouldn't refer to ManagedChannel either.

carl-mastrangelo:
PTAL

PR #2547, opened 2016-12-21
Add support for http forward proxy with CONNECT (backport)

ejona86:
Note that this is for the v1.0.x branch.

This is a squash and modification of master commits that also includes:
netty,okhttp: Fix CONNECT and its error handling

This commit has been modified to reduce its size to substantially reduce
risk of it breaking Netty error handling. But that also means proxy
error handling just provides a useless "there was an error" sort of
message.

There is no Java API to enable the proxy support. Instead, you must set
the GRPC_PROXY_EXP environment variable which should be set to a
host:port string. The environment variable is temporary; it will not
exist in future releases. It exists to provide support without needing
explicit code to enable the future, while at the same time not risking
enabling it for existing users.

CC @lukaszx0

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
+++ b/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
@@ -322,6 +322,16 @@ public class NettyChannelBuilder extends AbstractManagedChannelImplBuilder<Netty
       }
       ProtocolNegotiator negotiator = protocolNegotiator != null ? protocolNegotiator :
           createProtocolNegotiator(authority, negotiationType, sslContext);
+      String proxy = System.getenv("GRPC_PROXY_EXP");
+      if (proxy != null) {
+        String[] parts = proxy.split(":", 2);
+        int port = 80;
+        if (parts.length > 1) {
+          port = Integer.parseInt(parts[1]);
+        }
+        InetSocketAddress proxyAddress = new InetSocketAddress(parts[0], port);
+        negotiator = ProtocolNegotiators.httpProxy(proxyAddress, null, null, negotiator);
+      }
       return newClientTransport(serverAddress, authority, userAgent, negotiator);
     }
 
--- a/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
+++ b/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
@@ -44,6 +44,7 @@ import io.netty.channel.ChannelDuplexHandler;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerAdapter;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandler;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelPipeline;
 import io.netty.channel.ChannelPromise;
@@ -54,6 +55,9 @@ import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpVersion;
 import io.netty.handler.codec.http2.Http2ClientUpgradeCodec;
 import io.netty.handler.codec.http2.Http2ConnectionHandler;
+import io.netty.handler.proxy.HttpProxyHandler;
+import io.netty.handler.proxy.ProxyConnectionEvent;
+import io.netty.handler.proxy.ProxyHandler;
 import io.netty.handler.ssl.OpenSsl;
 import io.netty.handler.ssl.OpenSslEngine;
 import io.netty.handler.ssl.SslContext;
@@ -63,6 +67,7 @@ import io.netty.util.AsciiString;
 import io.netty.util.Attribute;
 import io.netty.util.ReferenceCountUtil;
 
+import java.net.SocketAddress;
 import java.net.URI;
 import java.util.ArrayDeque;
 import java.util.Arrays;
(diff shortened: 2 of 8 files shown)


zhangkun83:
> This commit has been modified to reduce its size to substantially reduce
risk of it breaking Netty error handling. But that also means proxy
error handling just provides a useless "there was an error" sort of
message.

At least the user can tell proxy errors from backend connection errors, right?

ejona86:
> At least the user can tell proxy errors from backend connection errors, right?

Backend errors sent via grpc-status are unaffected. If an error occurs during negotiation then they receive ["Connection closed with unknown cause"](https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/NettyClientTransport.java#L235-L236), because the closeFuture completes before the proxy negotiator reports its error.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -365,10 +437,22 @@ public void channelRegistered(ChannelHandlerContext ctx) throws Exception {
        * lifetime and we only want to configure it once.
        */
       if (handlers != null) {
-        ctx.pipeline().addFirst(handlers);
+        for (ChannelHandler handler : handlers) {
```
zhangkun83: Why is this change?

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -365,10 +437,22 @@ public void channelRegistered(ChannelHandlerContext ctx) throws Exception {
...
+          ctx.pipeline().addBefore(ctx.name(), null, handler);
+        }
+        ChannelHandler handler0 = handlers[0];
```
zhangkun83: And why this?

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
ejona86: Previously there was only one negotiator. Now there can be multiple. addFirst assumed it was the only thing in the channel pipeline. Now we make sure to put the other handlers immediately before this handler, which causes nesting to work correctly.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
ejona86: The channelRegistered event is delivered to the front. By adding these handlers before the current one, they won't receive the event. But the nested negotiator needs the channelRegistered event. Thus I trigger the event to the handlers that missed it. That causes this handler to receive the event twice, so I squelch one of the two.

It would be possible to teach the negotiation handlers to observe other events, but it would be rare that that code would be run, so it seems more bug-prone.

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -322,6 +322,16 @@ public ConnectionClientTransport newClientTransport(
       }
       ProtocolNegotiator negotiator = protocolNegotiator != null ? protocolNegotiator :
           createProtocolNegotiator(authority, negotiationType, sslContext);
+      String proxy = System.getenv("GRPC_PROXY_EXP");
```
lukaszx0: Sorry, but it isn't immediately obvious to me what does the `_EXP` suffix stand for?

ejona86: Experimental. I agree it isn't clear, but it's also going to be trashed. There's another PR that's currently out that uses the exp abbreviation for ALPN.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -365,10 +437,22 @@ public void channelRegistered(ChannelHandlerContext ctx) throws Exception {
...
+        ChannelHandlerContext handler0Ctx = ctx.pipeline().context(handlers[0]);
         handlers = null;
+        if (handler0Ctx != null) { // The handler may have removed itself immediately
```
zhangkun83: Does this comment apply to the case where `handler0Ctx == null`, where we don't need to call `channelRegistered()` on it?

To confirm my understanding, take `BufferUntilProxyTunnelledHandler(HttpProxyHandler, BufferUntilTlsNegotiatedHandler` for example, when  `BufferUntilProxyTunnelledHandler.channelRegistered()` is running, the pipe line will become
```
[HttpProxyHandler, BufferUntilTlsNegotiatedHandler, BufferUntilProxyTunnelledHandler]
```
Thus `handler0` will be `HttpProxyHandler`. Calling `handler0.fireChannelRegistered()` will call `fireChannelRegistered()` on every handler in the pipeline, including `BufferUntilTlsNegotiatedHandler` which would turn the pipeline into:
```
[HttpProxyHandler, SslHandler, BufferUntilTlsNegotiatedHandler, BufferUntilProxyTunnelledHandler]
```

Am I right?

netty/src/test/java/io/grpc/netty/ProtocolNegotiatorsTest.java
```diff
@@ -252,4 +276,93 @@ public void tls_invalidHost() throws SSLException {
...
+    ChannelHandler handler = nego.newHandler(grpcHandler);
+    Channel channel = new Bootstrap().group(elg).channel(LocalChannel.class).handler(handler)
+        .register().sync().channel();
+    pipeline = channel.pipeline();
+    // Wait for initialization to complete
+    channel.eventLoop().submit(NOOP_RUNNABLE).sync();
+    // The grpcHandler must be in the pipeline, but we don't actually want it during our test
+    // because it will consume all events since it is a mock. We only use it because it is required
+    // to construct the Handler.
+    pipeline.remove(grpcHandler);
```
zhangkun83: In production code, `grpcHandler` will be the one that does the HTTP2 framing? Without `grpcHandler`, `golden` will just be sent to the socket as-is?

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
lukaszx0: Ah I see, thanks - I wasn't familiar with this convention.

netty/src/test/java/io/grpc/netty/ProtocolNegotiatorsTest.java
ejona86: Yes. grpcHandler is our main handler. After negotiation, it may be the only handler (SslHandler may be another). It does all of the HTTP/2. Normally unrecognized messages passthrough the pipeline untouched. (Mocks don't participate correctly d consume everything)

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
ejona86: > Does this comment apply to the case where handler0Ctx == null, where we don't need to call channelRegistered() on it?

Yes. `handler0Ctx == null` means the handler is not in the pipeline. Since it isn't in the pipeline, it shouldn't receive further events. I will note that it is possible that we may need to call `channelRegistered()` on a `handler1Ctx` instead, and that isn't handled in the code. The only time that `handler0Ctx == null` today is `plaintext()`, which completes immediately and does nothing.

`handler0Ctx.fireChannelRegistered()` (the context) does _not_ call `channelRegistered()` on `HttpProxyHandler` in your example. It calls the _next_ in the pipeline, since the context is intended for handlers to call to fire events, and it wouldn't want the event to notify itself.  Note that I also can't use the _previous_ context to call the handler, because there may not be a previous. This has been annoying multiple times.

So instead of using the context, most of the time we'll call the handler directly because it is a `ChannelInboundHandler` and so has the `channelRegistered()` method. That will trigger the event to go through the rest of the pipeline (including this handler, and since `handlers == null`, it'll continue the chain via `super.channelRegistered(ctx)`, which just calls `ctx.fireChannelRegistered()`). In the case that the handler doesn't have `channelRegistered()`, then using the context is fine because `handler0` doesn't care about the event anyway.

To be clear, I'm not happy with the complexity and knowledge required here, but the alternatives seemed to be worse (e.g., buggier and higher coupling). On the other hand, this is based on just Netty knowledge and the only bug I encountered while testing is when `handler0Ctx == null`.

ejona86: Oh, and your view of the channel pipeline before and after the `channelRegistered()`s is correct.

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
ejona86: Calling it a _convention_ is probably giving it too much credit. I would have explained what it meant, but I instead used the words to describe it being deleted immediately (and thus, it shouldn't matter).

PR #2564, opened 2017-01-04
benchmarks: Do not set done to true when HistogramFuture#get

kkaneda:
IIUC, `done` should be set to true when we reach the code.


ejona86:
Jenkins, okay to test

ejona86:
@carl-mastrangelo, mind taking a look, too?

carl-mastrangelo:
hmm, I thought I commented on this, but I guess not.  

I asked why it was there to begin with, but looking at the original code I don't see.  It looks like a typo.

LGTM

PR #2571, opened 2017-01-06
context: expose Ticker interface and testing method

### LINKED ISSUE #2531: Make Deadline test-friendly
In its current form, io.grpc.Deadline is not suitable for testing.  It is not possible to create a deadline with an absolute offset, making it impossible to reliably create two relative deadlines.  The only public constructor is `after()` which is actually a timeout!

carl-mastrangelo:
The motivation is to make it easier for users to test their
deadlines.  Currently, only packages in io.grpc can set up tests
to use a fake ticker, while all other packages have to rely on the
System ticker.  This makes tests that compare deadlines flaky.

This exposes an interface in the context package for testing.
The previous abstract class was made to match the Guava version,
so for outside users to be able to continue using Guava, it
cannot be an abstract class.

The `baseInstant` has been left private, since I think it is still
easy to test deadlines without having access to it.

Fixes #2531

### DIFF
--- a/context/src/main/java/io/grpc/Deadline.java
+++ b/context/src/main/java/io/grpc/Deadline.java
@@ -55,20 +55,20 @@ public final class Deadline implements Comparable<Deadline> {
     return new Deadline(SystemTicker.INSTANCE, units.toNanos(duration), true);
   }
 
-  public static Deadline after(long duration, TimeUnit units, DeadlineTicker ticker) {
+  public static Deadline after(long duration, TimeUnit units, Ticker ticker) {
     checkNotNull(units, "units");
     return new Deadline(ticker, units.toNanos(duration), true);
   }
 
-  private final DeadlineTicker ticker;
+  private final Ticker ticker;
   private final long deadlineNanos;
   private volatile boolean expired;
 
-  private Deadline(DeadlineTicker ticker, long offset, boolean baseInstantAlreadyExpired) {
+  private Deadline(Ticker ticker, long offset, boolean baseInstantAlreadyExpired) {
     this(ticker, ticker.read(), offset, baseInstantAlreadyExpired);
   }
 
-  private Deadline(DeadlineTicker ticker, long baseInstant, long offset,
+  private Deadline(Ticker ticker, long baseInstant, long offset,
       boolean baseInstantAlreadyExpired) {
     this.ticker = ticker;
     // Clamp to range [MIN_OFFSET, MAX_OFFSET]
@@ -170,8 +170,7 @@ public final class Deadline implements Comparable<Deadline> {
     }
     Deadline that = (Deadline) obj;
     return this.deadlineNanos == that.deadlineNanos
-        && this.ticker == that.ticker
-        && this.expired == that.expired;
+        && ticker.equals(that.ticker);
   }
 
   /**
@@ -179,9 +178,7 @@ public final class Deadline implements Comparable<Deadline> {
    */
   @Override
   public int hashCode() {
-    return ((int)(deadlineNanos ^ (deadlineNanos >>> 32)))
-        ^ (expired ? 1231 : 1237)
-        ^ ticker.hashCode();
+    return ((int)(deadlineNanos ^ (deadlineNanos >>> 32))) ^ ticker.hashCode();
   }
 
   /**
@@ -189,12 +186,12 @@ public final class Deadline implements Comparable<Deadline> {
    *
    * <p>This is public for testing purposes.
    */
-  public interface DeadlineTicker {
+  public interface Ticker {
     /** Returns the number of nanoseconds since this source's epoch. */
     long read();
   }
 
-  private enum SystemTicker implements DeadlineTicker {
+  private enum SystemTicker implements Ticker {
     INSTANCE;
 
     @Override
--- a/context/src/main/java/io/grpc/Deadline.java
+++ b/context/src/main/java/io/grpc/Deadline.java
@@ -52,7 +52,7 @@ public final class Deadline implements Comparable<Deadline> {
    */
   public static Deadline after(long duration, TimeUnit units) {
     checkNotNull(units, "units");
-    return new Deadline(SystemTicker.INTANCE, units.toNanos(duration), true);
+    return new Deadline(SystemTicker.INSTANCE, units.toNanos(duration), true);
   }
 
   public static Deadline after(long duration, TimeUnit units, DeadlineTicker ticker) {
@@ -174,7 +174,7 @@ public final class Deadline implements Comparable<Deadline> {
   }
 
   private enum SystemTicker implements DeadlineTicker {
-    INTANCE;
+    INSTANCE;
 
     @Override
     public long read() {
--- a/context/src/main/java/io/grpc/Deadline.java
+++ b/context/src/main/java/io/grpc/Deadline.java
@@ -163,6 +163,27 @@ public final class Deadline implements Comparable<Deadline> {
     return 0;
   }
 
+  @Override
+  public boolean equals(Object obj) {
+    if (!(obj instanceof Deadline)) {
+      return false;
+    }
+    Deadline that = (Deadline) obj;
+    return this.deadlineNanos == that.deadlineNanos
+        && this.ticker == that.ticker
+        && this.expired == that.expired;
+  }
+
+  /**
+   * It is not expected that this will be used, but is added for completeness.
+   */
+  @Override
+  public int hashCode() {
+    return ((int)(deadlineNanos ^ (deadlineNanos >>> 32)))
+        ^ (expired ? 1231 : 1237)
+        ^ ticker.hashCode();
+  }
+
   /**
    * Time source representing nanoseconds since fixed but arbitrary point in time.
    *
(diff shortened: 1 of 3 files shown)


carl-mastrangelo:
Also, incase not obvious, the name `DeadlineTicker` was picked as to not conflict with the Guava `Ticker` class, which I expect to be the preferred implementation.

ejona86:
This alone won't help a single have less-flaky tests, since we convert it to timeouts both locally and "remotely."

Also, the implementation isn't really ready to handle exposing ticker, since comparing Deadlines with different tickers is pretty dangerous.

carl-mastrangelo:
Right now, there is no way to create two deadlines that target the same time in the future.   For example, if given a timeout (from a proto), you can't make an expected deadline to compare the one set in the call options.

I made the name "forTest" to make it clear that it is only suitable in tests, and custom tickers aren't something for everyday use.  My own personal use case is a ClientInterceptor test, not a remote endpoint.

context/src/main/java/io/grpc/Deadline.java
```diff
@@ -164,13 +164,19 @@ public int compareTo(Deadline that) {
...
   }
 
-  /** Time source representing nanoseconds since fixed but arbitrary point in time. */
-  abstract static class Ticker {
+  /**
+   * Time source representing nanoseconds since fixed but arbitrary point in time.
+   *
+   * <p>This is public for testing purposes.
+   */
+  public interface DeadlineTicker {
```
ejona86: Why not just call this Ticker and then people use Deadline.Ticker as the name? Let's not stutter unless we need to.

context/src/main/java/io/grpc/Deadline.java
```diff
@@ -164,13 +164,19 @@ public int compareTo(Deadline that) {
...
     /** Returns the number of nanoseconds since this source's epoch. */
-    public abstract long read();
+    long read();
   }
 
-  private static class SystemTicker extends Ticker {
```
ejona86: I don't see any value avoiding creating the small singleton for testing. But meh.

context/src/main/java/io/grpc/Deadline.java
```diff
@@ -164,13 +164,19 @@ public int compareTo(Deadline that) {
...
+  private enum SystemTicker implements DeadlineTicker {
+    INTANCE;
```
ejona86: s/INTANCE/INSTANCE/

context/src/main/java/io/grpc/Deadline.java
```diff
@@ -52,24 +51,25 @@
...
   public static Deadline after(long duration, TimeUnit units) {
-    return after(duration, units, SYSTEM_TICKER);
+    checkNotNull(units, "units");
+    return new Deadline(SystemTicker.INTANCE, units.toNanos(duration), true);
   }
 
-  // For testing
-  static Deadline after(long duration, TimeUnit units, Ticker ticker) {
+  //For testing
+  public static Deadline forTest(long duration, TimeUnit units, DeadlineTicker ticker) {
```
ejona86: Isn't this constructor hard to use in code without an easily accessible instance of SystemTicker? Although it sounds like you've created your own custom system ticker that is both a DeadlineTicker and cgc.base.Ticker.

ejona86:
> Right now, there is no way to create two deadlines that target the same time in the future.

True. But that's also not strictly required. If you took the Deadlines and the beginning and end of your test you could have a very strong bound for the acceptable output range.

> I made the name "forTest" to make it clear that it is only suitable in tests

But it's still a public API and tests are trying to measure correctness. I'm not comfortable having such a gaping hole in the implementation if others will be using it for their tests. I'm not against this in principal, but against a half-hearted implementation.

context/src/main/java/io/grpc/Deadline.java
zhangkun83: It's technically impossible to stop people from using it in production for whatever reason. It is mimicking the guava Ticker which is not test-only, so I don't see a reason this has to be test-only.

zhangkun83:
> But it's still a public API and tests are trying to measure correctness. I'm not comfortable having such a gaping hole in the implementation if others will be using it for their tests. I'm not against this in principal, but against a half-hearted implementation.

How about throwing from `compareTo()` if Tickers are not the same? And I don't see any benefit from making it `for-test`. I'd prefer it being a normal public API.

carl-mastrangelo:
Making the compare method fail if the tickers are different seems reasonable to me.  Eric said that the prod ticker will have to be the same as the test ticker, so a test only method probably doesn't make sense.

PTAL

context/src/main/java/io/grpc/Deadline.java
carl-mastrangelo: Because I don't want users to stutter when using the Guava ticker

`class MyTicker extends Ticker implements io.grpc.Ticker {}`

Naming it DeadlineTicker limits scope, and prevent people from autocompleting the wrong one.

context/src/main/java/io/grpc/Deadline.java
carl-mastrangelo: actually the singleton is for prod, for users who don't care.

context/src/main/java/io/grpc/Deadline.java
carl-mastrangelo: doh!

context/src/main/java/io/grpc/Deadline.java
ejona86: ?? I don't see what stuttering goes away with the `DeadlineTicker` name when using Guava. And that isn't how it would look. It would look like:

```java
class MyTicker extends Ticker implements Deadline.Ticker {}
```

Vs your rename:
```java
import io.grpc.Deadline.DeadlineTicker;

class MyTicker extends Ticker implements DeadlineTicker {}
```

I don't understand how it limits scope, but I do see the autocomplete argument.

context/src/main/java/io/grpc/Deadline.java
ejona86: We already had a singleton: `SYSTEM_TICKER`. This is a different way of doing a singleton that has the only benefit of avoiding creating the singleton if it goes unused. Since test code is the only place it would be unused, it seemed like you made the change to benefit test code.

Either that, or you just felt like changing the code for the heck of it.

context/src/main/java/io/grpc/Deadline.java
```diff
@@ -164,13 +163,40 @@ public int compareTo(Deadline that) {
     return 0;
   }
 
-  /** Time source representing nanoseconds since fixed but arbitrary point in time. */
-  abstract static class Ticker {
+  @Override
+  public boolean equals(Object obj) {
```
ejona86: Why is this now being implemented and compareTo wasn't changed?

context/src/main/java/io/grpc/Deadline.java
```diff
@@ -164,13 +163,40 @@ public int compareTo(Deadline that) {
...
+    if (!(obj instanceof Deadline)) {
+      return false;
+    }
+    Deadline that = (Deadline) obj;
+    return this.deadlineNanos == that.deadlineNanos
+        && this.ticker == that.ticker
+        && this.expired == that.expired;
```
ejona86: Don't compare expired. One instance may not have noticed it is actually expired yet.

context/src/main/java/io/grpc/Deadline.java
```diff
@@ -164,13 +163,40 @@ public int compareTo(Deadline that) {
...
+  }
+
+  /**
+   * It is not expected that this will be used, but is added for completeness.
+   */
+  @Override
+  public int hashCode() {
+    return ((int)(deadlineNanos ^ (deadlineNanos >>> 32)))
+        ^ (expired ? 1231 : 1237)
```
ejona86: Don't use expired in hashCode. That would make the class appear to mutate.

carl-mastrangelo: Done

context/src/main/java/io/grpc/Deadline.java
carl-mastrangelo: done

context/src/main/java/io/grpc/Deadline.java
zhangkun83: If `compareTo()` is going to throw when tickers are different, so should `isBefore()`

context/src/main/java/io/grpc/Deadline.java
```diff
@@ -164,13 +163,37 @@ public int compareTo(Deadline that) {
...
+   */
+  @Override
+  public int hashCode() {
+    return ((int)(deadlineNanos ^ (deadlineNanos >>> 32))) ^ ticker.hashCode();
+  }
+
+  /**
+   * Time source representing nanoseconds since fixed but arbitrary point in time.
+   *
+   * <p>This is public for testing purposes.
```
zhangkun83: Maybe just don't say this at all?

carl-mastrangelo:
I believe this has been eclipsed by @zpencer s changes.  Originally @zhangkun83 and I talked about this and decided on some sort of Deadline Factory, but I didn't get a chance to make this a reality.

zpencer:
I didn't realize there was previously an effort to expose unit testable deadlines. I think we should move grpc-core's `FakeClock` class into the context package, and use it as the basis of the Deadline factory. That class has a `ScheduledExecutorService` that only executes runnables when `forwardTime` is called, which gives us a complete solution for Context cancellations. I have a diff to try and do this but am running into some gradle issues.

PR #2573, opened 2017-01-06
testing: Remove accidental cast to char in file IO

ejona86:
Also add forgotten close() for the InputStream.

### DIFF
--- a/testing/src/main/java/io/grpc/testing/TestUtils.java
+++ b/testing/src/main/java/io/grpc/testing/TestUtils.java
@@ -31,8 +31,6 @@
 
 package io.grpc.testing;
 
-import static com.google.common.base.Charsets.UTF_8;
-
 import io.grpc.ExperimentalApi;
 import io.grpc.ForwardingServerCall.SimpleForwardingServerCall;
 import io.grpc.Metadata;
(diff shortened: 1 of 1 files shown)


testing/src/main/java/io/grpc/testing/TestUtils.java
```diff
@@ -255,19 +253,20 @@ public static InetSocketAddress testServerAddress(int port) {
...
+    InputStream in = new BufferedInputStream(TestUtils.class.getResourceAsStream("/certs/" + name));
     File tmpFile = File.createTempFile(name, "");
     tmpFile.deleteOnExit();
 
-    BufferedWriter writer =
-        new BufferedWriter(new OutputStreamWriter(new FileOutputStream(tmpFile), UTF_8));
+    OutputStream os = new BufferedOutputStream(new FileOutputStream(tmpFile));
     try {
       int b;
       while ((b = in.read()) != -1) {
```
dapengzhang0: why just read/write a single byte at a time? seems not taking advantage of **Buffered**Input/OutputStream

ejona86: If we read chunks at a time, we wouldn't actually need the **Buffered**Input/OutputStream. The Buffered streams keep a buffer internally and do reads and writes of chunks at a time.

We read and write a single byte at a time because it is dead-simple and we don't care about performance here.

ejona86:
Jenkins, retest this please

PR #2583, opened 2017-01-10
core: abstract channel builder to accept LoadBalancer2

zhangkun83:
If a `LoadBalancer2` is passed in, the builder will create `ManagedChannelImpl2` instead of `ManagedChannelImpl`. This allows us to test the LBv2 classes on a large scale.

### DIFF
--- a/core/src/main/java/io/grpc/internal/SharedResourcePool.java
+++ b/core/src/main/java/io/grpc/internal/SharedResourcePool.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2016, Google Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/grpc/internal/SharedResourcePool.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  @Override
+  public T getObject() {
+    return SharedResourceHolder.get(resource);
+  }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public T returnObject(Object object) {
```
ejona86: Why not `T object`? Hmm... I guess that's pre-existing.

zhangkun83: Right, this is pre-existing. In order to fix this, I will change all `ObjectPool<? extend Executor>` to `ObjectPool<Executor>`, `GrpcUtil.SHARED_CHANNEL_EXECUTOR` from `SharedResourceHolder.Resource<ExecutorService>` to `SharedResourceHolder.Resource<Executor>`. Does it sound better to you?

ejona86: Oh, I see. Yeah, we've going to have issues both ways. This is fine.

PR #2587, opened 2017-01-10
interop-testing: script for running http2 client

ericgribkoff:
Needed for running interop tests, https://github.com/grpc/grpc/pull/9298


ericgribkoff:
@makdharma fyi

ejona86:
So we discussed it and looked at the script, and we don't think there is much point to the script. I believe the `cd` is vestigial in which case it only executes the binary with an exec. So we're preferring you just calling the binary directly instead.

PR #2588, opened 2017-01-11
core,testing: make MethodDescriptor final and add Test helper

carl-mastrangelo:
MethodDescriptor was accidentally non final, allowing people to mock it.  Prevent that and provide a proper workaround.


core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -48,7 +48,7 @@
  * <p>Can be constructed manually but will often be generated by stub code generators.
  */
 @Immutable
-public class MethodDescriptor<ReqT, RespT> {
+public final class MethodDescriptor<ReqT, RespT> {
 
   private final MethodType type;
```
carl-mastrangelo: Yes, the extra finals are not needed. I secretly suspect that the final added to the class might need some transition period internally, so I don't want to lose the finals  in that case.  I would be happy to remove them once this is merged internally.

testing/src/main/java/io/grpc/testing/TestMethodDescriptors.java
```diff
@@ -0,0 +1,77 @@
...
+ * a descriptor, but don't really care how it works.
+ */
+public final class TestMethodDescriptors {
+  private TestMethodDescriptors() {}
+
+  public static MethodDescriptor<Void, Void> noopMethod() {
+    return noopMarshaller("service_foo", "method_bar");
+  }
+
+  private static MethodDescriptor<Void, Void> noopMarshaller(
```
zhangkun83: This method should be named `noopMethod()`.

carl-mastrangelo: Done.

testing/src/main/java/io/grpc/testing/TestMethodDescriptors.java
```diff
@@ -0,0 +1,77 @@
...
+import java.io.InputStream;
+
+/**
+ * A collection of method descriptor constructors useful for tests.  These are useful if you need
+ * a descriptor, but don't really care how it works.
+ */
+public final class TestMethodDescriptors {
+  private TestMethodDescriptors() {}
+
+  public static MethodDescriptor<Void, Void> noopMethod() {
```
zhangkun83: It's necessary to have a javadoc says that it creates a new instance each time, in case a test author wants to have multiple noop methods and verify against specific ones.

carl-mastrangelo: done

PR #2594, opened 2017-01-11
core: promote experimental methods in MethodDescriptor

carl-mastrangelo:
* `create` is being used from generated code, so really it isn't experimental.  To remove it now, would break stubs generated at 1.0.x.

* `generateFullMethodName` is non experimental for the same reason

* `extractFullServiceName` is more nuanced.  It is referenced by packages `util` and `auth`, which may be at earlier versions of gRPC than core is at.  If someone was using util:1.0.0 with core:1.n.m, they would expect it to work.  Importantly, the methods that call `extractFullServiceName` are themselves not experimental, which would seems to unintentionally promote them.  

If we did want to get rid of these, at best we could mark them deprecated and hope that no one calls them again.

### DIFF
--- a/core/src/main/java/io/grpc/MethodDescriptor.java
+++ b/core/src/main/java/io/grpc/MethodDescriptor.java
@@ -197,7 +197,6 @@ public class MethodDescriptor<ReqT, RespT> {
    * @param requestMarshaller the marshaller used to encode and decode requests
    * @param responseMarshaller the marshaller used to encode and decode responses
    */
-  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1774")
   public static <RequestT, ResponseT> MethodDescriptor<RequestT, ResponseT> create(
       MethodType type, String fullMethodName,
       Marshaller<RequestT> requestMarshaller,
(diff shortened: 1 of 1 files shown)


ejona86:
For `extractFullServiceName`, `util` reference isn't a big deal, because `util` is part of `core`. Auth is a bit more interesting. We haven't stated that you are permitted to mix-and-match grpc artifacts, but we do try to [version-pin](https://github.com/grpc/grpc-java/blob/v1.0.x/build.gradle#L297) when there are usages of `internal` by other artifacts. We could just say that it was a bug that `util` didn't version-pin (which I'd completely believe; I wouldn't be surprised if it wasn't the only such artifact with such a bug).

But I'm fine with it being non-experimental anyway.

@zhangkun83, you feel comfortable making `extractFullServiceName` non-experimental?

ejona86:
Hmm... We should probably throw some checkNotNulls() in those methods?

zhangkun83:
Yes, making `extractFullServiceName` non-experimental SGTM.

carl-mastrangelo:
Added some checkNotNulls

noel-yap:
Should `extractFullServiceName` error if the full method name is `/method` or `service/`?

noel-yap:
Should `extractFullServiceName` instead throw an exception if the full method name doesn't contain a `/`?

PR #2598, opened 2017-01-11
core: fix bug with Metadata merging

carl-mastrangelo:


core/src/main/java/io/grpc/Metadata.java
```diff
@@ -443,7 +443,7 @@ public void merge(Metadata other) {
     }
     int remaining = cap() - len();
     if (isEmpty() || remaining < other.len()) {
```
dapengzhang0: `isEmpty()` is not needed

carl-mastrangelo: It is needed since the underlying array could be null.  Merging two empty MDs would throw.

PR #2603, opened 2017-01-12
all: bump to netty 4.1.7

carl-mastrangelo:
too large headers now returns 431.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -161,9 +161,9 @@ subprojects {
                 protobuf_plugin: 'com.google.protobuf:protobuf-gradle-plugin:0.8.0',
                 protobuf_util: "com.google.protobuf:protobuf-java-util:${protobufVersion}",
 
-                netty: 'io.netty:netty-codec-http2:[4.1.6.Final]',
-                netty_epoll: 'io.netty:netty-transport-native-epoll:4.1.6.Final' + epoll_suffix,
-                netty_tcnative: 'io.netty:netty-tcnative-boringssl-static:1.1.33.Fork23',
+                netty: 'io.netty:netty-codec-http2:[4.1.7.Final]',
+                netty_epoll: 'io.netty:netty-transport-native-epoll:4.1.7.Final' + epoll_suffix,
+                netty_tcnative: 'io.netty:netty-tcnative-boringssl-static:1.1.33.Fork25',
 
                 // Test dependencies.
                 junit: 'junit:junit:4.11',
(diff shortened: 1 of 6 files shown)


carl-mastrangelo:
jenkins test this please.
jenkins retest this please.

build.gradle
```diff
@@ -161,9 +161,9 @@ subprojects {
                 protobuf_plugin: 'com.google.protobuf:protobuf-gradle-plugin:0.8.0',
                 protobuf_util: "com.google.protobuf:protobuf-java-util:${protobufVersion}",
 
-                netty: 'io.netty:netty-codec-http2:[4.1.6.Final]',
-                netty_epoll: 'io.netty:netty-transport-native-epoll:4.1.6.Final' + epoll_suffix,
-                netty_tcnative: 'io.netty:netty-tcnative-boringssl-static:1.1.33.Fork23',
+                netty: 'io.netty:netty-codec-http2:[4.1.7.Final]',
+                netty_epoll: 'io.netty:netty-transport-native-epoll:4.1.7.Final' + epoll_suffix,
+                netty_tcnative: 'io.netty:netty-tcnative-boringssl-static:1.1.33.Fork25',
```
ejona86: Also update SECURITY.md.

core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -182,6 +182,7 @@ public static Status httpStatusToGrpcStatus(int httpStatusCode) {
     }
     switch (httpStatusCode) {
       case HttpURLConnection.HTTP_BAD_REQUEST:  // 400
+      case 431: // Request Header Fields Too Large
```
ejona86: A change like this needs to be done in https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md first. Why is this change necessary?

carl-mastrangelo: Header processing fails in NettyClientTransport test otherwise I think.  The change is not made there because this preserves existing error behavior (INTERNAL instead of UNKNOWN).

build.gradle
carl-mastrangelo: Done.

core/src/main/java/io/grpc/internal/GrpcUtil.java
ejona86: At least add a TODO or similar comment calling out that it needs to be fixed.

PR #2605, opened 2017-01-12
	core: ServerImpl returns shared resources at termination

### LINKED ISSUE #2034: ServerImpl may use executor after it is released
Since 777e92853, ServerImpl [may reference `executor`](https://github.com/grpc/grpc-java/blob/v0.15.0/core/src/main/java/io/grpc/internal/ServerImpl.java#L331) and [may schedule Runnables on `executor`](https://github.com/grpc/grpc-java/blob/v0.15.0/core/src/main/java/io/grpc/internal/ServerImpl.java#L478) after `executor` is released. The executor should be released on termination, not shutdown.

zhangkun83:
Previously it does it at shutdown, which was wrong because executor may
still be used before the server is terminated.

Resolves #2034

Uses ObjectPool to make this change testable.  Cleans up test and makes
it mostly single-threaded, except for two deadlock tests that have to be
multi-threaded.

### DIFF
--- a/core/src/main/java/io/grpc/Server.java
+++ b/core/src/main/java/io/grpc/Server.java
@@ -49,7 +49,8 @@ public abstract class Server {
    *
    * @return {@code this} object
    * @throws IllegalStateException if already started
-   * @throws IOException if unable to bind
+   * @throws IOException if unable to bind.  {@link #shutdown} or {@link #shutdownNow} still needs
+   *         to be called to release any resource that may have been acquired by the server
    */
   public abstract Server start() throws IOException;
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/grpc/Server.java
```diff
@@ -49,7 +49,8 @@
    *
    * @return {@code this} object
    * @throws IllegalStateException if already started
-   * @throws IOException if unable to bind
+   * @throws IOException if unable to bind.  {@link #shutdown} or {@link #shutdownNow} still needs
```
zhangkun83: Done. PTAL.

Issue #2604, opened 2017-01-12
GRPC is logging NOT_FOUND errors as SEVERE and spamming the logs


r-tock:
gRPC = 1.0.3

We are using NOT FOUND to indicate that a resource is missing on the server. This is an ok response. However, this is getting logged as a SEVERE error on the client side and we can't find a way to control how to turn off logging for this particular error or reduce its severity to INFO.

Can someone help fix this/explain why NOT_FOUND is severe? 

cc: @jacob-tock

r-tock:
```
com.tocktix.ssm.ServerSideStateManagerImplTest > handleTwoSequentialNotFoundErrors STANDARD_ERROR
    Jan 12, 2017 8:57:36 PM io.grpc.internal.ManagedChannelImpl <init>
    INFO: [ManagedChannelImpl@4f4af133] Created with target directaddress:///io.grpc.inprocess.InProcessSocketAddress@48eb32b9
    Jan 12, 2017 8:57:36 PM io.grpc.internal.SerializeReentrantCallsDirectExecutor execute
    SEVERE: Exception while executing runnable io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$2@2620b663
    io.grpc.StatusRuntimeException: NOT_FOUND
    	at com.tocktix.ssm.ServerSideStateManagerImplTest$DelegatingSsmService.query(ServerSideStateManagerImplTest.java:359)
    	at com.tocktix.messages.SsmServiceGrpc$MethodHandlers.invoke(SsmServiceGrpc.java:340)
    	at io.grpc.stub.ServerCalls$1$1.onHalfClose(ServerCalls.java:150)
    	at io.grpc.internal.ServerCallImpl$ServerStreamListenerImpl.halfClosed(ServerCallImpl.java:261)
    	at io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$2.runInContext(ServerImpl.java:503)
    	at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:52)
    	at io.grpc.internal.SerializeReentrantCallsDirectExecutor.execute(SerializeReentrantCallsDirectExecutor.java:65)
    	at io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener.halfClosed(ServerImpl.java:499)
    	at io.grpc.inprocess.InProcessTransport$InProcessStream$InProcessClientStream.halfClose(InProcessTransport.java:530)
[...]

PR #2606, opened 2017-01-13
core: more logs in ManagedChannelImpl.

### LINKED ISSUE #2299: More logs in channel
We have insufficient logging in the channel layer. In particular:
- Name resolution success and failures. Although failures are plumbed to application when it makes a call.
- ManagedChannelImpl goes into and out of idleness

Whether they should be INFO or FINE are not clear, but at least they should start as FINE.

zhangkun83:
They have already been added to ManagedChannelImpl2.

Resolves #2299

### DIFF
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -234,6 +234,7 @@ public final class ManagedChannelImpl extends ManagedChannel implements WithLogI
               IDLE_GRACE_PERIOD_MILLIS, TimeUnit.MILLISECONDS);
           return;
         }
+        log.log(Level.FINE, "[{0}] Entering idle mode", getLogId());
         // Enter idle mode
         savedBalancer = graceLoadBalancer;
         graceLoadBalancer = null;
@@ -288,6 +289,7 @@ public final class ManagedChannelImpl extends ManagedChannel implements WithLogI
       if (loadBalancer != null) {
         return loadBalancer;
       }
+      log.log(Level.FINE, "[{0}] Exiting idle mode", getLogId());
       balancer = loadBalancerFactory.newLoadBalancer(nameResolver.getServiceAuthority(), tm);
       this.loadBalancer = balancer;
       resolver = this.nameResolver;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -764,10 +769,13 @@ public void onUpdate(List<ResolvedServerInfoGroup> servers, Attributes config) {
         onError(Status.UNAVAILABLE.withDescription("NameResolver returned an empty list"));
         return;
       }
+      log.log(Level.FINE, "[{0}] resolved address: {1}, config={2}",
+          new Object[] {getLogId(), servers, config});
 
       try {
         balancer.handleResolvedAddresses(servers, config);
       } catch (Throwable e) {
+        log.log(Level.WARNING, "[" + getLogId() + "] Caught exception from LoadBalancer", e);
```
carl-mastrangelo: Why is this here?  shouldnt the balancer handle this below?

zhangkun83: So that we don't need to repeat the log in each LoadBalancer. Also we don't control 3rd-party LoadBalancers and cannot guarantee they will handle it or log it. This is for our own benefit, in that we always have a way to know what's happening.

carl-mastrangelo: Won't a lot of other alarm bells go off?  This is expected to happen for network outages in common case right?

zhangkun83: No, it's only logged if there is a bug in `LoadBalancer`, because it's not expected to throw.

Here we catch and pass the exception back to `LoadBalancer`, in hope that it will be propagated to the application by LoadBalancer. However, LoadBalancer may not propagate or log as we expect, or there may not even be RPC yet to be propagated to, so we need to log here as the last resort.

We don't catch in `onError()` below, because if `handleNameResolutionError()` throws, there isn't a really good way to propagate this error to application. The exception will be caught by transport thread (or `ChannelExecutor` in v2 impl) and logged there.

carl-mastrangelo: If this is unexpected, could you put that in the message?  This [issue](https://github.com/grpc/grpc-java/issues/2604) from yesterday came from someone not realizing they should have been catching the exception, and reading the log message.

zhangkun83: Done. PTAL.

PR #2611, opened 2017-01-17
RoundRobinLoadBalancer minor cleanups

lukaszx0:
A bunch of code clean up in `RoundRobinLoadBalancer2`

/cc @kkaneda  @zhangkun83

### DIFF
--- a/core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory2.java
+++ b/core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory2.java
@@ -65,6 +65,7 @@ import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 
 import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
 
 /**
  * A {@link LoadBalancer} that provides round-robin load balancing mechanism over the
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory2.java
```diff
@@ -241,28 +241,27 @@ private Status getAggregatedError() {
     final Status status;
```
kkaneda: add `private` for consistency?

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory2.java
```diff
@@ -241,28 +241,27 @@ private Status getAggregatedError() {
...
     private final List<Subchannel> list;
     private int index;
```
kkaneda: maybe mention that `index` is guarded by this?

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory2.java
lukaszx0: Actually it's used in tests. I've added pkg private getter and made it private

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory2.java
lukaszx0: Done

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory2.java
```diff
@@ -238,31 +238,31 @@ private Status getAggregatedError() {
...
+      if (status != null) {
+        return PickResult.withError(status);
+      }
+
+      return PickResult.withNoResult();
     }
 
     private Subchannel nextSubchannel() {
-      if (empty) {
+      if (list.isEmpty()) {
```
zhangkun83: How about saving `size` as a member variable? This way you can eliminate `isEmpty()` call here and the `list.size()` call on line 271 and 272. In fact line 272 could be `index = 0`.

lukaszx0: Good call, fixed. Reshuffled and changed first commit, which is now last (might be worth to merge those fixes as a separate commits, that's why I'm keeping them separate - your call)

zhangkun83: They are all very small changes. I am going to squash them.

PR #2526, opened 2016-12-16
core,netty: add getAttributes to ClientStream and ClientCall

dapengzhang0:
add `getAttributes()` to `ClientStream` and `ClientCall` to be able to share clientTransport
information such as socket TOS with higher lever API's, once the RPC picks up an active transport that is ready to use.

Issue #1779, opened 2016-05-03
Tracking Issue for ServerCall Attributes being Experimental.


carl-mastrangelo:
Specific usages:

daicoden:
Following the headers example: https://github.com/grpc/grpc-java/tree/master/examples/src/main/java/io/grpc/examples/header

What is the difference between Metadata and Attributes?  I was going to add OU to metadata via a client interceptor, but saw this comment on attributes.

```
   * Returns properties of a single call. This is a generic container which can contain any kind of
   * information describing call like for example remote address, TLS information (OU etc.)
```

carl-mastrangelo:
Metadata represents headers that are sent on the wire.  They have stronger restrictions about their values, and have higher parsing costs.

Attributes are more like an escape hatch to pass values between parts of gRPC.  They do not get transmitted as part of RPCs.

dapengzhang0:
we should also rename the API as "getAttributes" to be consistent with other getters and setters API in the same class.

larry-safran:
Has already been renamed to getAttributes

larry-safran:
Blocked on class `Attributes` being Experimental #1764

PR #2615, opened 2017-01-18
Rename attributes() to getAttributes() to make it consistent

lukaszx0:
Make it consistent with everything else. Addresses https://github.com/grpc/grpc-java/issues/1779 and https://github.com/grpc/grpc-java/pull/2526#discussion_r96547749

/cc @ejona86 @dapengzhang0

### DIFF
--- a/core/src/main/java/io/grpc/ServerCall.java
+++ b/core/src/main/java/io/grpc/ServerCall.java
@@ -58,7 +58,7 @@ import javax.net.ssl.SSLSession;
 public abstract class ServerCall<ReqT, RespT> {
   /**
    * {@link Attributes.Key} for the remote address of server call attributes
-   * {@link ServerCall#attributes()}
+   * {@link ServerCall#getAttributes()}
    *
    * @deprecated use the equivalent {@link io.grpc.Grpc#TRANSPORT_ATTR_REMOTE_ADDR} instead
    */
@@ -69,7 +69,7 @@ public abstract class ServerCall<ReqT, RespT> {
 
   /**
    * {@link Attributes.Key} for the SSL session of server call attributes
-   * {@link ServerCall#attributes()}
+   * {@link ServerCall#getAttributes()}
    *
    * @deprecated use the equivalent {@link io.grpc.Grpc#TRANSPORT_ATTR_SSL_SESSION} instead
    */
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/grpc/internal/ServerCallImpl.java
```diff
@@ -192,7 +192,7 @@ ServerStreamListener newServerStreamListener(ServerCall.Listener<ReqT> listener)
   }
 
   @Override
-  public Attributes attributes() {
+  public Attributes getAttributes() {
     return stream.attributes();
```
lukaszx0: @ejona86 Heh, what about this one?

lukaszx0: There's also setter in this interface - https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/ServerStream.java#L83

At least it's in `internal` package, so I guess we should rename it to `getAttributes` too?

core/src/main/java/io/grpc/ServerCall.java
```diff
@@ -237,7 +237,7 @@ public void setCompression(String compressor) {
    * @return Attributes container
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1779")
-  public Attributes attributes() {
+  public Attributes getAttributes() {
```
ejona86: Go ahead and have a `@Deprecated` and `@ExperimentalApi` `attributes()` method that calls the new method. That is very easy to do and makes it much easier to migrate.

core/src/main/java/io/grpc/internal/ServerCallImpl.java
ejona86: Meh. I'm not concerned either way.

core/src/main/java/io/grpc/ServerCall.java
lukaszx0: Done.

core/src/main/java/io/grpc/internal/ServerCallImpl.java
lukaszx0: I'll leave it that way then

lukaszx0: I've just discovered that `(Client|Server)Stream` are inconsistent in the same way, in which case I think it's worth adjusting: https://github.com/grpc/grpc-java/pull/2616

PR #2616, opened 2017-01-18
core: rename attributes() -> getAttributes()

lukaszx0:
Make it consistent with its counterpart, `ClientStream#getAttributes()`


lukaszx0:
@ejona86 resolved conflict, should build green now

lukaszx0:
:shipit: ?

PR #2617, opened 2017-01-18
core: rename getAttr() and use standard getAttributes()

lukaszx0:
Makes it consistent across the board - it's `getAttributes` all the way down.

/cc @ejona86


ejona86:
I saw this test failure:
```
io.grpc.okhttp.OkHttpTransportTest > clientCancel FAILED
    java.lang.AssertionError
```

That strikes me as very strange. I'm very disappointed it didn't give more backtrace information. Probably a flake, but it'd be nice to know what assertion failed.

lukaszx0:
:shipit: ?

PR #2618, opened 2017-01-18
core: DelayedStream cancels provided stream if not using it.

### LINKED ISSUE #1537: DelayedStream.startStream() should cancel the provided stream if not using it
If `cancel()` or `startStream()` was previously called then startStream()` throws away the stream argument. It should `cancel()` the stream to make sure resources are freed, although I think only DelayedClientTransport benefits.

This should not be done until after #1536. I don't know if OkHttp suffers a similar problem. If so, we may want to revisit whether it makes sense to allow streams to be cancelled before `start()`.

zhangkun83:
Resolves #1537

Also disallow cancel() before start().
DelayedClientTransport.shutdownNow() races with stream start(), thus it
shouldn't call cancel() directly.  It would delay the cancellation until
the stream is started.

### DIFF
--- a/core/src/main/java/io/grpc/internal/DelayedClientTransport.java
+++ b/core/src/main/java/io/grpc/internal/DelayedClientTransport.java
@@ -218,7 +218,7 @@ class DelayedClientTransport implements ManagedClientTransport {
     }
     if (savedPendingStreams != null) {
       for (PendingStream stream : savedPendingStreams) {
-        stream.cancel(status);
+        stream.cancelInternal(status);
       }
       listener.transportTerminated();
     }
@@ -477,6 +477,12 @@ class DelayedClientTransport implements ManagedClientTransport {
     private final Context context;
     private final StatsTraceContext statsTraceCtx;
 
+    private final Object pendingStreamLock = new Object();
+    @GuardedBy("pendingStreamLock")
+    private boolean started;
+    @GuardedBy("pendingStreamLock")
+    private Status pendingCancelReason;
+
     private PendingStream(MethodDescriptor<?, ?> method, Metadata headers,
         CallOptions callOptions, StatsTraceContext statsTraceCtx) {
       this.method = method;
@@ -497,8 +503,32 @@ class DelayedClientTransport implements ManagedClientTransport {
       setStream(realStream);
     }
 
+    // This may be called concurrently with other methods on the stream
+    private void cancelInternal(Status reason) {
+      synchronized (pendingStreamLock) {
+        if (!started) {
+          pendingCancelReason = reason;
+          return;
+        }
+      }
+      cancel(reason);
+    }
+
     @Override
-    public void cancel(Status reason) {
+    public final void start(ClientStreamListener listener) {
+      Status savedPendingCancelReason;
+      synchronized (pendingStreamLock) {
+        started = true;
+        savedPendingCancelReason = pendingCancelReason;
+      }
+      super.start(listener);
+      if (savedPendingCancelReason != null) {
+        cancel(savedPendingCancelReason);
+      }
+    }
+
+    @Override
+    public final void cancel(Status reason) {
       super.cancel(reason);
       synchronized (lock) {
         if (pendingStreams != null) {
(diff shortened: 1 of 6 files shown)


carl-mastrangelo:
@ejona86 can you look at this, it is blocking 1.1

core/src/main/java/io/grpc/internal/DelayedStream.java
```diff
@@ -104,19 +119,40 @@ public void run() {
...
-    drainPendingCalls();
+    if (savedRealStream == null) {
+      drainPendingCalls();
+    } else {
+      // If realStream was not null, then either setStream() or cancel() must had been called,
+      // we will cancel and discard the given stream.
+      // ClientStream.cancel() must be called after start()
+      stream.start(NOOP_STREAM_LISTENER);
+      if (savedError != null) {
+        stream.cancel(savedError);
```
ejona86: Let's not cancel with `savedError`. We're cancelling the stream because of the incorrect API usage, not because of a previous error.

At that point `error` itself can be deleted.

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
```diff
@@ -497,8 +503,32 @@ private void createRealStream(ClientTransport transport) {
...
+    }
+
+    @Override
+    public final void start(ClientStreamListener listener) {
+      Status savedPendingCancelReason;
+      synchronized (pendingStreamLock) {
+        started = true;
+        savedPendingCancelReason = pendingCancelReason;
+      }
+      super.start(listener);
```
zhangkun83: Why does this matter?

core/src/main/java/io/grpc/internal/DelayedStream.java
zhangkun83: The misuse case is calling `setStream()` twice, which is the `else` branch.
This branch is for calling `setStream()` after `cancel()`, which is legit, and I think the real stream should be cancelled with `savedError` which was used to cancel the delayed stream.

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
ejona86: If we're immediately cancelling, it seems pretty weak to start if it is going to do I/O. Doing unnecessary I/O in presence of cancellation is generally a recipe of cascading failure. There are two things important here though: 1) the cancellation involved here is expected to be rare and only impact a few streams since it is only due to a race between the stream creation and the delayed client transport shutdown and 2) even when it happens, no I/O is done.

core/src/main/java/io/grpc/internal/DelayedStream.java
ejona86: > I think the real stream should be cancelled with savedError which was used to cancel the delayed stream.

That probably doesn't matter at all. A simple Status.CANCELLED.withDescription("cancelled before setStream") or "the stream was only started so it could be cancelled" would seem fine to me. The application won't end up seeing the status. When debugging, I would probably prefer to see the hard-coded status instead of `savedError` as well, since the reason for the cancel is the limited API that required the start() in the first place.

> This branch is for calling setStream() after cancel(), which is legit

Hmm... If it is legit, we have to make sure it doesn't happen frequently and could only impact a few streams. And it seems that doesn't hold for `MetadataApplierImpl` and may not hold for `DelayedClientTransport{,2}`. That would be the fix could actually cause stability problems whereas the current code isn't causing any. Hmm...

zhangkun83: For my own record, the issue is `setStream()` may be scheduled for an unbounded period of time into the future. For example, `CallCredentials` may take a long time to fetch the credentials before triggering `setStream()` (in `MetadataApplierImpl`, prior to which the RPC may be cancelled due to deadline-exceeded. In those cases, real stream are created before `setStream()` is called, only to be started (as required by cancel()) and then cancelled. `start()` typically involve I/O, e.g., sending headers, which are wasted. If the delay before `setStream()` is long enough, this will have a high chance to happen.

zhangkun83:
As discussed in #1537, putting this on-hold.

Issue #2613, opened 2017-01-18
Put server address in ClientCall#attributes()


lukaszx0:
It would be very helpful to put the target address into `ClientCall#attributes()`. One of the use cases is the ability to access it in interceptors.

Without first class support for retries, we're using [RetryingInterceptor](https://github.com/grpc/grpc-java/pull/1570/files). We want to be able to write LB logic which in the case of failed attempt, would pick different backed on next call. There's currently no way to achieve that and ability to get information about which address is targeted in given call would solve it (the `RetryingInterceptor` would then append target address to `BACKENDS_TRIED` key in affinity attributes, on every attempt).

/cc @dapengzhang0 @zhangkun83  @kkaneda

lukaszx0:
I'm happy to work on that once #2526 will be merged.

lukaszx0:
I've started digging into this today and while working on it I realized that #2526 might've ship with everything we needed to convey target address to interceptors.

The [`ClientCallImpl#getAttributes`](https://github.com/grpc/grpc-java/blob/d0d2c1e7961aa28e7d0518fc07e560803309c1fa/core/src/main/java/io/grpc/internal/ClientCallImpl.java#L415-L421) in their current form delegate to `stream` which should contain `Grpc.TRANSPORT_ATTR_REMOTE_ADDR` set in [`NettyServerHandler#handlerAdded`](https://github.com/grpc/grpc-java/blob/d0d2c1e7961aa28e7d0518fc07e560803309c1fa/netty/src/main/java/io/grpc/netty/NettyServerHandler.java#L171-L173).

I think we have everything we need. @zhangkun83 does that sound about right?

zhangkun83:
@lukaszx0 `NettyServerHandler#handlerAdded` is on server side. What you need is the client-side counterpart of it.

lukaszx0:
Omg, I've confused myself again with server code. Sorry, ignore my previous comment 🤦‍♂️

PR #2619, opened 2017-01-18
core: pass down addr to which client is connected

lukaszx0:
This addresses #2613

### DIFF
--- a/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
+++ b/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
@@ -95,6 +95,10 @@ public final class ProtocolNegotiators {
           public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
             // Just replace this handler with the gRPC handler.
             ctx.pipeline().replace(this, null, handler);
+            // TODO(lukaszx0) not sure if this it the best/right place to do this
+            handler.handleProtocolNegotiationCompleted(Attributes.newBuilder()
+                .set(Grpc.TRANSPORT_ATTR_REMOTE_ADDR, ctx.channel().remoteAddress())
+                .build());
           }
 
           @Override
@@ -157,6 +161,7 @@ public final class ProtocolNegotiators {
                 Attributes.newBuilder()
                     .set(Grpc.TRANSPORT_ATTR_SSL_SESSION,
                         sslHandler(ctx.pipeline()).engine().getSession())
+                    .set(Grpc.TRANSPORT_ATTR_REMOTE_ADDR, ctx.channel().remoteAddress())
                     .build());
             // Replace this handler with the GRPC handler.
             ctx.pipeline().replace(this, null, grpcHandler);
@@ -511,6 +516,7 @@ public final class ProtocolNegotiators {
             grpcHandler.handleProtocolNegotiationCompleted(
                 Attributes.newBuilder()
                     .set(Grpc.TRANSPORT_ATTR_SSL_SESSION, handler.engine().getSession())
+                    .set(Grpc.TRANSPORT_ATTR_REMOTE_ADDR, ctx.channel().remoteAddress())
                     .build());
             writeBufferedAndRemove(ctx);
           } else {
--- a/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
+++ b/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
@@ -93,12 +93,12 @@ public final class ProtocolNegotiators {
         class PlaintextHandler extends ChannelHandlerAdapter implements Handler {
           @Override
           public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
-            // Just replace this handler with the gRPC handler.
-            ctx.pipeline().replace(this, null, handler);
-            // TODO(lukaszx0) not sure if this it the best/right place to do this
+            // Set sttributes before replace to be sure we pass it before accepting any requests.
             handler.handleProtocolNegotiationCompleted(Attributes.newBuilder()
                 .set(Grpc.TRANSPORT_ATTR_REMOTE_ADDR, ctx.channel().remoteAddress())
                 .build());
+            // Just replace this handler with the gRPC handler.
+            ctx.pipeline().replace(this, null, handler);
           }
 
           @Override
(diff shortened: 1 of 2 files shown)


lukaszx0:
This is first stab at conveying target address down to `ClientCall`. It's not fully functional yet, @zhangkun83 let me know if this is the way how would like to be implemented?

cc @kkaneda

core/src/main/java/io/grpc/Attributes.java
```diff
@@ -152,6 +152,21 @@ public int hashCode() {
     return data.hashCode();
   }
 
+  /**
+   * Merges two attributes objects into one.
+   */
+  public static Attributes merge(Attributes a, Attributes b) {
```
kkaneda: test for this method?

core/src/main/java/io/grpc/Attributes.java
```diff
@@ -152,6 +152,21 @@ public int hashCode() {
...
+    if (a == Attributes.EMPTY) {
```
kkaneda: just curious, how bad if we just always do `return Attributes.newBuilder(a).setAll(b).build();`? I guess we can also do something like 

```
if (a == Attributes.EMPTY && b == Attributes.EMPTY) {
  return Attributes.EMPTY;
}
return Attributes.newBuilder(a).setAll(b).build();
```

netty/src/main/java/io/grpc/netty/NettyClientTransport.java
```diff
@@ -82,6 +83,7 @@
   private final int flowControlWindow;
   private final int maxMessageSize;
   private final int maxHeaderListSize;
+  private Attributes attributes;
```
kkaneda: can this be `final`?

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -217,6 +219,9 @@ public void runInContext() {
       updateTimeoutHeaders(effectiveDeadline, callOptions.getDeadline(),
           context.getDeadline(), headers);
       ClientTransport transport = clientTransportProvider.get(callOptions, headers);
+      if (transport instanceof ConnectionClientTransport) {
```
kkaneda: hmm, should we add `getAttrs` to `ClientTransport` to avoid cast here? I'm not sure that's a right change, but we would then kill `ConnectionClientTransport`?

netty/src/main/java/io/grpc/netty/NettyClientTransport.java
```diff
@@ -110,6 +112,8 @@
     this.maxHeaderListSize = maxHeaderListSize;
     this.authority = new AsciiString(authority);
     this.userAgent = new AsciiString(GrpcUtil.getGrpcUserAgent("netty", userAgent));
+    // TODO(lukaszx0): do we want to reuse, or create new key for server/client?
+    this.attributes = Attributes.newBuilder().set(Grpc.TRANSPORT_ATTR_REMOTE_ADDR, address).build();
```
kkaneda: Do we need to make simpler change to other classes inheriting from `ConnectionClientTransport` (e.g., `InProcessTransport`)?

ejona86:
You should call handleProtocolNegotiationCompleted from the protocol negotiator, similar to what is done in ServerTlsHandler today. It does require teaching each negotiator to add the IP address, but we do know different negotiators will be able to add different information.

lukaszx0:
@ejona86 ah, yeah thanks - that makes sense.

Do we need to update all of the `BufferUntil*Handler`classes to call `handleProtocolNegotiationCompleted` ?

ejona86:
> Do we need to update all of the BufferUntil*Handlerclasses to call handleProtocolNegotiationCompleted ?

Eventually. I would have liked it if that was done when the client-side attributes were added, but it isn't hurting anything yet.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -95,6 +95,10 @@ public Handler newHandler(final GrpcHttp2ConnectionHandler handler) {
           public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
             // Just replace this handler with the gRPC handler.
             ctx.pipeline().replace(this, null, handler);
+            // TODO(lukaszx0) not sure if this it the best/right place to do this
```
ejona86: Do it right before the replace(), since after this point `handler` may be receiving requests.

lukaszx0: Fixed.

kkaneda:
Question: Do we still need the following code in`NettyServerHandler`?

```
  @Override
    serverWriteQueue = new WriteQueue(ctx.channel());
    attributes = transportListener.transportReady(Attributes.newBuilder(protocolNegotationAttrs)
        .set(Grpc.TRANSPORT_ATTR_REMOTE_ADDR, ctx.channel().remoteAddress())
        .build());
    super.handlerAdded(ctx);
  }
```

ejona86:
> Question: Do we still need the following code inNettyServerHandler?

Saving `serverWriteQueue`, yes. Calling `transportReady()`, yes. However, calling `transportReady()` from `handleProtocolNegotiationCompleted()` would probably be better. We'd just need to verify that all server negotiators call `handleProtocolNegotiationCompleted()`.

But that's fine to do later (and I'd honestly prefer doing that right after we cut the v1.1.x branch).

kkaneda:
I see. Thanks for the explanation!

lukaszx0:
@ejona86 are we missing anything from the changes, as they are now? (except the things that you've pointed out that'd be better to take care of later)

zhangkun83:
Same question for this code:
```java
  @Override
    serverWriteQueue = new WriteQueue(ctx.channel());
    attributes = transportListener.transportReady(Attributes.newBuilder(protocolNegotationAttrs)
        .set(Grpc.TRANSPORT_ATTR_REMOTE_ADDR, ctx.channel().remoteAddress())
        .build());
    super.handlerAdded(ctx);
  }
```
Isn't setting `TRANSPORT_ATTR_REMOTE_ADDR` here redundant with what is added to the negotiators?

dapengzhang0:
> Isn't setting TRANSPORT_ATTR_REMOTE_ADDR here redundant with what is added to the negotiators?

@zhangkun83 that's right. And the `attributes` field could be set in `NettyServerHandler#handleProtocolNegotiationCompleted` instead, doing that, the method `handleProtocolNegotiationCompleted` must not be forgotten to invoke in every ProtocolNegotiator implementation.

lukaszx0:
Nice, thanks @dapengzhang0 !

PR #2620, opened 2017-01-18
all: ErrorProne fixes and avoid @Beta in Guava

ejona86:


core/src/main/java/io/grpc/internal/ManagedChannelImpl2.java
```diff
@@ -631,12 +628,12 @@ void onStateChange(InternalSubchannel is, ConnectivityStateInfo newState) {
               }
 
               @Override
-              public void onInUse(InternalSubchannel is) {
+              protected void onInUse(InternalSubchannel is) {
```
zhangkun83: They should actually be package-private.

ejona86: Done.

Issue #2157, opened 2016-08-11
Flaky CompressionTest


zhangkun83:
Found in https://travis-ci.org/grpc/grpc-java/jobs/151357545

```
io.grpc.testing.integration.CompressionTest > compression[23] FAILED
    io.grpc.StatusRuntimeException: INTERNAL: Exception deframing message
        at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:230)
        at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:211)
        at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:144)
        at io.grpc.testing.integration.TestServiceGrpc$TestServiceBlockingStub.unaryCall(TestServiceGrpc.java:369)
        at io.grpc.testing.integration.CompressionTest.compression(CompressionTest.java:217)
        Caused by:
        io.grpc.StatusRuntimeException: INTERNAL: Can't decode compressed frame as compression not configured.
```

ejona86:
Also seen at:
https://travis-ci.org/grpc/grpc-java/jobs/149904961
https://travis-ci.org/grpc/grpc-java/jobs/156283171 (since rebuilt)

kleinsch:
Just saw this in our client code that's using compression.

```
Caused by: io.grpc.StatusRuntimeException: INTERNAL: Can't decode compressed frame as compression not configured.
```

ejona86:
Seen at https://travis-ci.org/grpc/grpc-java/jobs/193463553
```
io.grpc.testing.integration.CompressionTest > compression[55] FAILED
    io.grpc.StatusRuntimeException: INTERNAL: io.grpc.netty.NettyClientTransport$2: Can't decode compressed frame as compression not configured.
        at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:230)
        at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:211)
        at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:144)
        at io.grpc.testing.integration.TestServiceGrpc$TestServiceBlockingStub.unaryCall(TestServiceGrpc.java:408)
        at io.grpc.testing.integration.CompressionTest.compression(CompressionTest.java:217)
```

ejona86:
Seen at https://travis-ci.org/grpc/grpc-java/jobs/194406178 (since rebuilt)
```
[...]

PR #2622, opened 2017-01-18
core: add debug information in MessageDeframer.

zhangkun83:
So that MessageDeframer would include the class name of the enclosing
stream when emitting errors.  This should give us more information about #2157


zhangkun83:
I was meant to refer to #2157 in the commit message, but it was wrapped at the beginning of a line and treated as comment.

PR #2625, opened 2017-01-19
core: change method descriptor to be builder based

carl-mastrangelo:
This change is not as big as it seems, the core of the change is in MethodDescriptor.

Still to do is to move over generated code, and remove the "with methods"

### DIFF
--- a/core/src/main/java/io/grpc/MethodDescriptor.java
+++ b/core/src/main/java/io/grpc/MethodDescriptor.java
@@ -372,8 +372,19 @@ public final class MethodDescriptor<ReqT, RespT> {
    * Creates a new builder for a {@link MethodDescriptor}.
    */
   @CheckReturnValue
-  public static Builder<?, ?> newBuilder() {
-    return new Builder<Object, Object>();
+  public static <ReqT, RespT> Builder<ReqT, RespT> newBuilder() {
+    return newBuilder(null, null);
+  }
+
+  /**
+   * Creates a new builder for a {@link MethodDescriptor}.
+   */
+  @CheckReturnValue
+  public static <ReqT, RespT> Builder<ReqT, RespT> newBuilder(
+      Marshaller<ReqT> requestMarshaller, Marshaller<RespT> responseMarshaller) {
+    return new Builder<ReqT, RespT>()
+        .setRequestMarshaller(requestMarshaller)
+        .setResponseMarshaller(responseMarshaller);
   }
 
   /**
@@ -381,7 +392,16 @@ public final class MethodDescriptor<ReqT, RespT> {
    */
   @CheckReturnValue
   public Builder<ReqT, RespT> toBuilder() {
-    return newBuilder()
+    return toBuilder(requestMarshaller, responseMarshaller);
+  }
+
+  /**
+   * Turns this descriptor into a builder, replacing the request and response marshallers.
+   */
+  @CheckReturnValue
+  public <NewReqT, NewRespT> Builder<NewReqT, NewRespT> toBuilder(
+      Marshaller<NewReqT> requestMarshaller, Marshaller<NewRespT> responseMarshaller) {
+    return MethodDescriptor.<NewReqT, NewRespT>newBuilder()
         .setRequestMarshaller(requestMarshaller)
         .setResponseMarshaller(responseMarshaller)
         .setType(type)
@@ -408,10 +428,9 @@ public final class MethodDescriptor<ReqT, RespT> {
      * Sets the request marshaller.
      * @param requestMarshaller the marshaller to use.
      */
-    @SuppressWarnings("unchecked")
-    public <T> Builder<T, RespT> setRequestMarshaller(Marshaller<T> requestMarshaller) {
-      this.requestMarshaller = (Marshaller<ReqT>) requestMarshaller;
-      return (Builder<T, RespT>) this;
+    public Builder<ReqT, RespT> setRequestMarshaller(Marshaller<ReqT> requestMarshaller) {
+      this.requestMarshaller = requestMarshaller;
+      return this;
     }
 
     /**
--- a/core/src/main/java/io/grpc/MethodDescriptor.java
+++ b/core/src/main/java/io/grpc/MethodDescriptor.java
@@ -38,6 +38,7 @@ import com.google.common.base.Preconditions;
 import java.io.InputStream;
 import java.util.concurrent.atomic.AtomicReferenceArray;
 
+import javax.annotation.CheckReturnValue;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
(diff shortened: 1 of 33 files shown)


core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -365,4 +367,111 @@ public static String extractFullServiceName(String fullMethodName) {
...
+  @CheckReturnValue
+  public static Builder<?, ?> newBuilder() {
+    return new Builder<Object, Object>();
+  }
+
+  /**
+   * Turns this descriptor into a builder.
+   */
+  @CheckReturnValue
+  public Builder<ReqT, RespT> toBuilder() {
```
zhangkun83: `Attributes.newBuilder(Attributes)` is another style. Not saying that's better, but we should try to maintain some style consistency across the code path. Although we already have inconsistencies here and there, at least we can avoid introducing yet another new style to make it more fragmented.

p.s. After I wrote the above paragraph, I did a search and I realized Protobuf uses `toBuilder()`, so it's not completely divergent. I post this comment anyway to raise the awareness of the problem, but I am fine with it as-is.

zhangkun83:
LGTM.
I feel much better for this than the current mixture of `create()` and `with()`. Unfortunately it's too late to remove `create()`. Should `with()`es be marked as deprecated?

core/src/main/java/io/grpc/MethodDescriptor.java
carl-mastrangelo: I believe Proto is more widely used, and was the inspiration for this builder.

carl-mastrangelo:
I plan on removing with as soon as this is in.

carl-mastrangelo:
@ejona86 & @zhangkun83 PTAL

core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -365,4 +367,130 @@ public static String extractFullServiceName(String fullMethodName) {
...
+      this.requestMarshaller = requestMarshaller;
+      return this;
+    }
+
+    /**
+     * Sets the response marshaller.
+     * @param responseMarshaller the marshaller to use.
+     */
+    @SuppressWarnings("unchecked")
+    public <T> Builder<ReqT, T> setResponseMarshaller(Marshaller<T> responseMarshaller) {
```
ejona86: This one looks like it was forgotten to be updated.

core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -365,4 +367,130 @@ public static String extractFullServiceName(String fullMethodName) {
...
+  }
+
+  /**
+   * A builder for a {@link MethodDescriptor}.
+   */
+  public static final class Builder<ReqT, RespT> {
+
+    private Marshaller<ReqT> requestMarshaller;
+    private Marshaller<RespT> responseMarshaller;
+    private MethodType type;
```
ejona86: Do we want to require type, or initialize it to UNKNOWN?

core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -365,4 +367,130 @@ public static String extractFullServiceName(String fullMethodName) {
...
+        .setType(type)
+        .setFullMethodName(fullMethodName)
+        .setIdempotent(idempotent)
+        .setSafe(safe);
+  }
+
+  /**
+   * A builder for a {@link MethodDescriptor}.
+   */
+  public static final class Builder<ReqT, RespT> {
```
ejona86: Let's go with ExperimentalApi? (here and above)

core/src/main/java/io/grpc/MethodDescriptor.java
carl-mastrangelo: It was, but i reverted it while playing around with the signatures.  Fixed.

core/src/main/java/io/grpc/MethodDescriptor.java
carl-mastrangelo: This was supposed to be used by the generated code in the next commit, which would implicitly make it Non experimental.  I can leave the stub code alone bu that would mean they would still need to use MD.create.  Also, I would probably have to back out all the changes from the non-code packages.

core/src/main/java/io/grpc/MethodDescriptor.java
carl-mastrangelo: meh, Id rather it be set explicitly by the user.  That can be relaxed in the future, but never added back.

core/src/main/java/io/grpc/MethodDescriptor.java
ejona86: `MethodDescriptor.create()` isn't going away, so I don't see what the rush is to avoid using it in generated code. And I do have concern with committing a new API and releasing it as stable a couple days later.

> Also, I would probably have to back out all the changes from the non-code packages.

What do you mean? non-code?

carl-mastrangelo: derp, non-core.  This is widely used, so if we remove it in 1.2, 1.1 might not be compatible with 1.2.  Maybe thats okay.

carl-mastrangelo: Also, I know create isn't going away, but I was hoping to scare people away from it by marking it deprecated.  Old generated code may use it but and human authors should be using the builder.  

For people who take inspiration from the generated code, I want them to use the builder pattern too.

ejona86: Non-core isn't an issue. Ideally we'd just want to make sure that such projects are in [the version-pinning list](https://github.com/grpc/grpc-java/blob/d5eb248/build.gradle#L299).

zhangkun83:
@carl-mastrangelo LGTM

core/src/main/java/io/grpc/MethodDescriptor.java
carl-mastrangelo: I'll add it, but have little confidence of it sticking.

core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -365,4 +367,131 @@ public static String extractFullServiceName(String fullMethodName) {
     }
     return fullMethodName.substring(0, index);
   }
+
+  /**
+   * Creates a new builder for a {@link MethodDescriptor}.
+   */
+  @CheckReturnValue
+  public static <ReqT, RespT> Builder<ReqT, RespT> newBuilder() {
```
ejona86: experimental. Ditto 3x below.

carl-mastrangelo: Done

PR #2630, opened 2017-01-20
docs: clarify requirements of detach().

zhangkun83:


context/src/main/java/io/grpc/Context.java
```diff
@@ -379,8 +379,13 @@ public Context attach() {
 
   /**
    * Detach the current context and attach the provided replacement which should be the context of
-   * the outer scope, thus exit the current scope.  If this context is not {@link #current()} a
-   * SEVERE message will be logged but the context to attach will still be bound.
+   * the outer scope, thus exit the current scope.
+   *
+   * <p>This context should be {@link #current()}.  If it's not the case, a SEVERE message will be
```
ejona86: I don't think this is the way we want to describe it. The worst thing to is `Context.current().detach()`. Instead, you should say this context should be the attached context (the exact reverse of the previous attach). And then include other details about if the attached context isn't current, then probably some other code is broken and not detaching correctly, but it will be logged here.

zhangkun83: Rephrased. PTAL.

context/src/main/java/io/grpc/Context.java
```diff
@@ -379,8 +379,18 @@ public Context attach() {
...
+   * <p>This context should be the same context that was previously {@link #attach attached}.  The
+   * provided replacement should be what was returned by the matching {@link #attach attach()} call.
+   * If an {@code attach()} and a {@code detach()} meet above requirements, they match.
+   *
+   * <p>It is expected that all {@code attach()}es and {@code detach()}es between this {@code
+   * detach()} and the matching {@code attach()} also perfectly match.  If this method finds that
+   * this context is not {@link #current current}, either you or some code in-between are not
+   * detaching correctly, and a SEVERE message will be logged but the context to attach will still
+   * be bound.  <strong>Never</strong> use {@code Context.current().detach()}, as this will
+   * compromise this error-detecing mechanism.
```
ejona86: s/detecing/detecting/

PR #2634, opened 2017-01-20
core: make StatsContextFactory setters protected

zhangkun83:
These are only used in internal tests.  In production,
StatsContextFactory is loaded by the "Instrumentation" library and must
be one per process, thus we won't allow setting it on a per-channel or
per-server basis.

### DIFF
--- a/core/src/test/java/io/grpc/internal/TestUtils.java
+++ b/core/src/test/java/io/grpc/internal/TestUtils.java
@@ -36,6 +36,8 @@ import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import com.google.instrumentation.stats.StatsContextFactory;
+
 import io.grpc.CallOptions;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
@@ -51,7 +53,7 @@ import javax.annotation.Nullable;
 /**
  * Common utility methods for tests.
  */
-final class TestUtils {
+public final class TestUtils {
 
   static class MockClientTransportInfo {
     /**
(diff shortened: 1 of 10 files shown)


core/src/test/java/io/grpc/internal/TestUtils.java
```diff
@@ -51,7 +53,7 @@
 /**
  * Common utility methods for tests.
  */
-final class TestUtils {
+public final class TestUtils {
```
carl-mastrangelo: needs a private constructor now

interop-testing/build.gradle
```diff
@@ -16,6 +16,10 @@ dependencies {
             libraries.mockito,
             libraries.netty_tcnative,
             libraries.oauth_client
+
+    // Tests depend on base class defined by core module.
+    compile project(':grpc-core').sourceSets.test.output
```
carl-mastrangelo: is this correct?  how come the project(':grpc-core') dep above is not enough?

zhangkun83: Normal deps only include `main` sourceSet, not `test`.

core/src/test/java/io/grpc/internal/TestUtils.java
zhangkun83: Done.

interop-testing/src/test/java/io/grpc/testing/integration/TransportCompressionTest.java
```diff
@@ -190,8 +189,9 @@ public void onHeaders(Metadata headers) {
             };
           }
         })
-        .usePlaintext(true)
-        .build();
+        .usePlaintext(true);
+    io.grpc.internal.TestUtils.setStatsContextFactory(builder,getClientStatsFactory());
```
carl-mastrangelo: add space after comma

zhangkun83: Done.

Issue #2357, opened 2016-10-19
ServiceDescriptor lacks validation


ejona86:
ServerServiceDefinition has validation, but it seems that should actually be in ServiceDescriptor (moved or copied). The method names should match the service name, and there shouldn't be two methods with the same name.

PR #2637, opened 2017-01-21
core: make ServiceDescriptor check input values

carl-mastrangelo:
Fixes: #2357 

Slightly different than the validation from `ServerServiceDefinition` since it has a more complicated use case.  We can unify these after 1.1

### DIFF
--- a/core/src/main/java/io/grpc/ServiceDescriptor.java
+++ b/core/src/main/java/io/grpc/ServiceDescriptor.java
@@ -105,7 +105,7 @@ public final class ServiceDescriptor {
       String methodServiceName =
           MethodDescriptor.extractFullServiceName(method.getFullMethodName());
       checkArgument(serviceName.equals(methodServiceName),
-          "service names %s !=%s", methodServiceName, serviceName);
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/grpc/ServiceDescriptor.java
```diff
@@ -92,4 +98,16 @@ public String getName() {
...
     return marshallerDescriptor;
   }
+
+  static void validateMethodNames(String serviceName, Collection<MethodDescriptor<?, ?>> methods) {
+    Set<String> allNames = new HashSet<String>(methods.size());
+    for (MethodDescriptor<?, ?> method : methods) {
+      String methodServiceName =
+          MethodDescriptor.extractFullServiceName(method.getFullMethodName());
+      checkArgument(serviceName.equals(methodServiceName),
+          "service names %s !=%s", methodServiceName, serviceName);
```
ejona86: Space before second %s

carl-mastrangelo: Done

PR #2639, opened 2017-01-23
Replace Mockito with StreamObserver for several interop tests

kpayson64:
Mockito is restricted on AppEngine.

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -116,6 +116,8 @@ import org.mockito.verification.VerificationMode;
 
 /**
  * Abstract base class for all GRPC transport tests.
+ *
+ * <p> New tests should avoid using Mockito to support running on AppEngine.</p>
  */
 public abstract class AbstractInteropTest {
   /** Must be at least {@link #unaryPayloadLength()}, plus some to account for encoding overhead. */
@@ -354,29 +356,41 @@ public abstract class AbstractInteropTest {
                 .setBody(ByteString.copyFrom(new byte[58979])))
             .build());
 
-    @SuppressWarnings("unchecked")
-    StreamObserver<StreamingOutputCallResponse> responseObserver = mock(StreamObserver.class);
+    final ArrayBlockingQueue<Object> queue = new ArrayBlockingQueue<Object>(5);
     StreamObserver<StreamingOutputCallRequest> requestObserver
-        = asyncStub.fullDuplexCall(responseObserver);
+        = asyncStub.fullDuplexCall(new StreamObserver<StreamingOutputCallResponse>() {
+          @Override
+          public void onNext(StreamingOutputCallResponse response) {
+            queue.add(response);
+          }
+
+          @Override
+          public void onError(Throwable t) {
+            queue.add(t);
+          }
+
+          @Override
+          public void onCompleted() {
+            queue.add("Completed");
+          }
+        });
     for (int i = 0; i < requests.size(); i++) {
+      assertNull(queue.peek());
       requestObserver.onNext(requests.get(i));
-      verify(responseObserver, timeout(operationTimeoutMillis())).onNext(goldenResponses.get(i));
-      verifyNoMoreInteractions(responseObserver);
+      assertEquals(goldenResponses.get(i),
+                   queue.poll(operationTimeoutMillis(), TimeUnit.MILLISECONDS));
     }
     requestObserver.onCompleted();
-    verify(responseObserver, timeout(operationTimeoutMillis())).onCompleted();
-    verifyNoMoreInteractions(responseObserver);
+    assertEquals("Completed", queue.poll(operationTimeoutMillis(), TimeUnit.MILLISECONDS));
   }
 
   @Test(timeout = 10000)
   public void emptyStream() throws Exception {
-    @SuppressWarnings("unchecked")
-    StreamObserver<StreamingOutputCallResponse> responseObserver = mock(StreamObserver.class);
+    StreamRecorder<StreamingOutputCallResponse> responseObserver = StreamRecorder.create();
     StreamObserver<StreamingOutputCallRequest> requestObserver
         = asyncStub.fullDuplexCall(responseObserver);
     requestObserver.onCompleted();
-    verify(responseObserver, timeout(operationTimeoutMillis())).onCompleted();
-    verifyNoMoreInteractions(responseObserver);
+    responseObserver.awaitCompletion(operationTimeoutMillis(), TimeUnit.MILLISECONDS);
   }
 
   @Test(timeout = 10000)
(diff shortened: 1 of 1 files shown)


carl-mastrangelo:
Reassiging to ejona, who said he would be removing StreamRecorder

ejona86:
What was the driving force for this change?

I'm okay with us using StreamRecorder; I'm mainly against it being an exposed API for others to use, because it can have sharp corners.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -359,28 +359,31 @@ public void pingPong() throws Exception {
             .build());
 
     @SuppressWarnings("unchecked")
-    StreamObserver<StreamingOutputCallResponse> responseObserver = mock(StreamObserver.class);
+    StreamRecorder<StreamingOutputCallResponse> responseObserver = StreamRecorder.create();
     StreamObserver<StreamingOutputCallRequest> requestObserver
         = asyncStub.fullDuplexCall(responseObserver);
     for (int i = 0; i < requests.size(); i++) {
       requestObserver.onNext(requests.get(i));
-      verify(responseObserver, timeout(operationTimeoutMillis())).onNext(goldenResponses.get(i));
```
ejona86: This is a ping-pong test. You have to assert we received another response.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -359,28 +359,31 @@ public void pingPong() throws Exception {
...
+    assertEquals(goldenResponses.size(), responseObserver.getValues().size());
+    for (int ix = 0; ix < responseObserver.getValues().size(); ++ix) {
+      StreamingOutputCallResponse response = responseObserver.getValues().get(ix);
+      assertEquals(goldenResponses.get(ix), response);
+    }
   }
 
   @Test(timeout = 10000)
   public void emptyStream() throws Exception {
     @SuppressWarnings("unchecked")
```
ejona86: Remove the `@SuppressWarnings`. Ditto elsewhere (where you removed the `mock()`).

kpayson64: Fixedbb

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
kpayson64: In the section below I compare the responses from the StreamRecorder and the goldenResponses.

ejona86: But that's not all of what this was trying to check. It was making sure that the channel is full-duplex. We should be able to receive the response before sending the next request.

HTTP is commonly implemented as half-duplex, where the client's request is fully received before the server starts processing. Being able to do ping-pong within a single stream is a very, very important case that can break in-practice.

ejona86:
FYI: while we do rebase commits when submitting to master, we try to avoid rebasing during code review, because it is very difficult to see the modifications that we made. I'd have to basically re-review this PR now.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
kpayson64: I've added a wait mechanism to the StreamObserver to make sure we get a message.

testing/src/main/java/io/grpc/testing/StreamRecorder.java
```diff
@@ -127,4 +142,18 @@ public boolean awaitCompletion(int timeout, TimeUnit unit) throws Exception {
...
+
+  /**
+   * Waits for a value to be received from the stream since the last
+   * call to awaitNext() or construction.  Useful for testing streaming
+   * call patterns.
+   */
+  public void awaitNext() {
+    nextValueLock.lock();
+    if (!seenNextValue) {
+      nextValueCondition.awaitUninterruptibly();
```
ejona86: Don't wait uninterruptibly. Just let the `InterruptionException` propagate.

testing/src/main/java/io/grpc/testing/StreamRecorder.java
```diff
@@ -127,4 +142,18 @@ public boolean awaitCompletion(int timeout, TimeUnit unit) throws Exception {
   public ListenableFuture<T> firstValue() {
     return firstValue;
   }
+
+  /**
+   * Waits for a value to be received from the stream since the last
+   * call to awaitNext() or construction.  Useful for testing streaming
+   * call patterns.
+   */
+  public void awaitNext() {
```
ejona86: Tests hanging when they fail can be horrible to debug. Pass in a timeout (in the form `long, TimeUnit`). You then either need to return a boolean of whether next occurred in time (and assertTrue() at each usage) or throw a `TimeoutException`.

testing/src/main/java/io/grpc/testing/StreamRecorder.java
```diff
@@ -127,4 +142,18 @@ public boolean awaitCompletion(int timeout, TimeUnit unit) throws Exception {
...
+    nextValueLock.lock();
+    if (!seenNextValue) {
```
ejona86: Always use a while loop for conditions. Conditions can have spurious wakeups.

Note that makes it a bit annoying for calling `await(long, TimeUnit)` because the time keeps changing. You can do `nanos = timeUnit.toNanos(time)` and then use `nanos = nextValueCondition.awaitNanos(nanos)`.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
ejona86: I added comments about the implementation. However, it may be easier to use `BlockingQueue<Object>` (like `LinkedBlockingQueue`). You can create your own `StreamObserver` that adds any `StreamingOutputCallResponse` or `Throwable` to the queue. That would also naturally produce a reasonable error message if an entry isn't what you expect. See `serverStreamingShouldBeFlowControlled()` for an example.

ejona86:
FYI: we avoid merge commits in grpc-java. We just rebase instead.

ejona86:
retest this please

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
kpayson64: I like this idea better.  Changed.

kpayson64:
I've addressed the comments.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -354,29 +356,41 @@ public void pingPong() throws Exception {
...
+          public void onCompleted() {
+            queue.add("Completed");
+          }
+        });
     for (int i = 0; i < requests.size(); i++) {
+      assertNull(queue.peek());
       requestObserver.onNext(requests.get(i));
-      verify(responseObserver, timeout(operationTimeoutMillis())).onNext(goldenResponses.get(i));
-      verifyNoMoreInteractions(responseObserver);
+      assertEquals(queue.poll(operationTimeoutMillis(), TimeUnit.MILLISECONDS),
```
ejona86: Reverse the arguments to this `assertEquals()`. It should be `assertEquals(expected, actual)`.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -354,29 +356,41 @@ public void pingPong() throws Exception {
...
   public void emptyStream() throws Exception {
-    @SuppressWarnings("unchecked")
-    StreamObserver<StreamingOutputCallResponse> responseObserver = mock(StreamObserver.class);
+    StreamRecorder<StreamingOutputCallResponse> responseObserver = StreamRecorder.create();
     StreamObserver<StreamingOutputCallRequest> requestObserver
         = asyncStub.fullDuplexCall(responseObserver);
     requestObserver.onCompleted();
-    verify(responseObserver, timeout(operationTimeoutMillis())).onCompleted();
-    verifyNoMoreInteractions(responseObserver);
+    responseObserver.awaitCompletion();
```
ejona86: Please use the `awaitCompletion(int, TimeUnit)` version of this method. Ditto elsewhere.

PR #2642, opened 2017-01-23
all: avoid DNS with GRPC_PROXY_EXP

ejona86:
In some environments DNS is not available and is performed by the
CONNECT proxy. Nothing "special" should need to be done for these
environments, but the previous support took shortcuts which knowingly
would not support such environments.

This change should fix both OkHttp and Netty. Netty's
Bootstrap.connect() resolved the name immediately whereas using
ChannelPipeline.connect() waits until the address reaches the end of the
pipeline. Netty uses NetUtil.toSocketAddressString() to get the name of
the address, which uses InetSocketAddress.getHostString() when
available.

OkHttp is still using InetSocketAddress.getHostName() which may issue
reverse DNS lookups. However, if the reverse DNS lookup fails, it should
convert the IP to a textual string like getHostString(). So as long as
the reverse DNS maps to the same machine as the IP, there should only be
performance concerns, not correctness issues. Since the DnsNameResolver
is creating unresolved addresses, the reverse DNS lookups shouldn't
occur in the common case.

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
@@ -290,16 +290,6 @@ public class TestServiceClient {
     @Override
     protected ManagedChannel createChannel() {
       if (!useOkHttp) {
-        InetAddress address;
-        try {
-          address = InetAddress.getByName(serverHost);
-          if (serverHostOverride != null) {
-            // Force the hostname to match the cert the server uses.
-            address = InetAddress.getByAddress(serverHostOverride, address.getAddress());
-          }
-        } catch (UnknownHostException ex) {
-          throw new RuntimeException(ex);
-        }
         SslContext sslContext = null;
         if (useTestCa) {
           try {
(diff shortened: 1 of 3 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
```diff
@@ -290,16 +290,6 @@ private void runTest(TestCases testCase) throws Exception {
...
-        InetAddress address;
-        try {
-          address = InetAddress.getByName(serverHost);
-          if (serverHostOverride != null) {
-            // Force the hostname to match the cert the server uses.
-            address = InetAddress.getByAddress(serverHostOverride, address.getAddress());
-          }
-        } catch (UnknownHostException ex) {
-          throw new RuntimeException(ex);
-        }
```
zhangkun83: Why is this removed?

ejona86: `forAddress(InetSocketAddress)` doesn't work with proxies. This is very old code, before we had `forTarget(String)` and `forAdddress(String, port)`. Back in the day when testing TLS you had to resolve the IP yourself and then make the InetAddress manually so that it had the right name. We removed the need for that when we added `overrideAuthority()` before we went beta. Updating to the "current way" is clearer and works with CONNECT.

zhangkun83:
Please fix the CI errors.

interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
ejona86: Errr... at least, I'm not trying to make `forAddress(InetSocketAddress)` work with proxies. Since it avoids the NameResolver.

ejona86:
Oh, FYI @lukaszx0

lukaszx0:
thanks @ejona86 👍

Issue #2444, opened 2016-11-25
Non-determinism when using InProcessTransport with direct executors


grandseiken:
### What version of gRPC are you using?

1.0.1

### What JVM are you using (`java -version`)?

1.8.0_101

### What did you do?

I am attempting to write unit tests for RPC systems without mocking stubs, which would be lovely if it worked. Recent discussions (e.g. [here](https://github.com/grpc/grpc-java/issues/1469), [1.0.1 release notes](https://github.com/grpc/grpc-java/releases/tag/v1.0.1), [this example](https://github.com/grpc/grpc-java/commit/7306df42662b8eb96f3524e3940b3e65d37deb73)) suggest this is the new done thing and is supposed to work. To this end I am running an `InProcessServer` and calling it over an `InProcessChannel` with both ends using a `DirectExecutor`.

### What did you expect to see?

The 1.0.1 release notes imply that this is now deterministic (after the removal of the `grpc-ready` thread pool, which was causing issues for me before), so I expected to see determinism in my unit test, and for the server handler to be called on the same thread that made the stub method call. The RPC method is a client-side streaming RPC, but I don't know if that is relevant.

### What did you see instead?

The unit test is not deterministic, since the service handler seems to end up being called on a `grpc-timer` thread provided by the `ScheduledExecutorService` created by `GrpcUtil.TIMER_SERVICE`. The task that causes the call is  the `NameResolverStartTask` in `ManagedChannelImpl`, and there appears to be no way to provide an alternative timer service.

[See ManagedChannelImpl.java line 307.](https://github.com/grpc/grpc-java/blob/496a621da7c57a252286f041e9b41e59b722ae9b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java#L307)

For reference, the stack trace that I see on the handler side is as follows:

```
        ...
	at foo.bar.FooServiceImpl$1.onNext(FooServiceImpl.java:33)
	at io.grpc.stub.ServerCalls$2$1.onMessage(ServerCalls.java:206)
	at io.grpc.internal.ServerCallImpl$ServerStreamListenerImpl.messageRead(ServerCallImpl.java:237)
	at io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1.runInContext(ServerImpl.java:485)
	at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:52)
[...]

PR #2644, opened 2017-01-23
services: Remove no-op call in reflection service test

ericgribkoff:
Now that the fix for #2444 is in, there's no need for the initial noop call.


ericgribkoff:
retest this please

PR #2652, opened 2017-01-25
core: stop "testing" from depending on "core"'s test.

### LINKED ISSUE #2651: AbstractInteropTest should not depend on TestUtils
PR #2634 accidentally added a dep from src/tests to another src/test directory, which is uncommon.  Shared test library code needs to be under a /testing/ directory in the src/java which can be reused by multiple test suites.

zhangkun83:
Because "core"'s test source already depends on "testing", e.g.,
`core/src/test/java/io/grpc/internal/ServerCallImplTest.java` uses
`testing/src/main/java/io/grpc/internal/testing/StatsTestUtils.java`,
which forms a circular dependency.

This change moves the StatsContext setter accessors from "core"'s test
source to "testing".

Resolves #2651

### DIFF
--- a/core/src/test/java/io/grpc/internal/TestUtils.java
+++ b/core/src/test/java/io/grpc/internal/TestUtils.java
@@ -36,8 +36,6 @@ import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
-import com.google.instrumentation.stats.StatsContextFactory;
-
 import io.grpc.CallOptions;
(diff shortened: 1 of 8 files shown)


core/src/test/java/io/grpc/internal/TestUtils.java
```diff
@@ -53,7 +51,7 @@
 /**
  * Common utility methods for tests.
  */
-public final class TestUtils {
+final class TestUtils {
```
carl-mastrangelo: good idea!

Issue #2653, opened 2017-01-25
Create GrpclbLoadBalancerFactory2


zhangkun83:
If #2650 is resolved by keeping LBv2 in 1.1, we should also add `GrpclbLoadBalancerFactory2` for 1.1, otherwise `GrpclbLoadBalancer2` will not be accessible to users.

PR #2658, opened 2017-01-26
core: open up LBv2 APIs for early adopters

### LINKED ISSUE #2650: Decide whether LBv2 should be kept in 1.1
If we do, we could provide a better (non-atomic) migration path for people (implementors) implementing `LoadBalancer`.

During 1.1, we keep the old API intact, and include the v2 API as `LoadBalancer2`. Implementors re-write their LBs with the v2 API, and migrate their customers to their v2 LB. The two versions of LB impls can coexist.

In 1.2, we replace `LoadBalancer` with `LoadBalancer2`, and keep `LoadBalancer2` as an alias to `LoadBalancer` by extending it. The old LB impls stop working. Before switching to 1.2, the implementors must have switched all their customers to the v2 API. During 1.2, they can migrate their customers to the proper `LoadBalancer` name.

In 1.3, we delete `LoadBalancer2`.

The caveat is that it increases the size of Android packages in 1.1. It shouldn't be a significant amount compared with the existing code size. We should at least calculate the exact size and note it in release notes.

If we are going to do this, we should also change the `@Internal` annotations to `@ExeprimentalApi` on the v2 APIs.

zhangkun83:
Also add GrpclbLoadBalancerFactory2 so that GrpclbLoadBalancer2 can be used.

Resolves #2650

### DIFF
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -159,9 +159,28 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
    *
    * <p>If this method is not called, the builder will use {@link PickFirstBalancerFactory}
    * for the channel.
+   *
+   * <p>This overrides what was set by {@link #loadBalancerFactory(LoadBalancer2.Factory)}.
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1771")
   public abstract T loadBalancerFactory(LoadBalancer.Factory loadBalancerFactory);
+
+  /**
+   * Provides a custom {@link LoadBalancer2.Factory} for the channel.
+   *
+   * <p>If this method is not called, the builder will use {@link PickFirstBalancerFactory}
+   * for the channel.
+   *
+   * <p>This overrides what was set by {@link #loadBalancerFactory(LoadBalancer.Factory)}.
+   *
+   * <p><strong>IMPORTANT: </strong>Calling this will make the channel to run the LBv2 code path,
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -159,11 +159,30 @@
...
+   * for the channel.
+   *
+   * <p>This overrides what was set by {@link #loadBalancerFactory(LoadBalancer.Factory)}.
+   *
+   * <p><strong>IMPORTANT: </strong>Calling this will make the channel to run the LBv2 code path,
+   * which is new and not as well-exercised as the current one.  You should use it only if you want
+   * to try out LBv2, or you are migrating your LoadBalancer implementation to LBv2.  See
+   * <a href="https://github.com/grpc/grpc-java/issues/2656" target="_blank">#2656</a> for more
+   * information.
+   */
```
carl-mastrangelo: This is also deprecated, the method signature will go away next version

zhangkun83: No, it won't. Next version will still have two signatures, while `LoadBalancer2` will just be an alias of `LoadBalancer`.

At 1.1, we expect users of stock LoadBalancers to keep using the first signature, and users of custom LoadBalancers to migrate to the second signature. Therefore, neither of them are deprecated.

At 1.2, the one with `LoadBalancer2` will be deprecated.

zhangkun83:
Also resolves #2653

carl-mastrangelo:
This probably should not have been rushed in.   This is a breaking API change.  Adding an abstract method to ManagedChannelBuilder breaks all implementations.

PR #2663, opened 2017-01-26
netty: call transportReady in handleProtocolNegotiationCompleted

dapengzhang0:
implements the comments mentioned in #2619

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
@@ -41,8 +41,8 @@ import static io.netty.handler.codec.http2.DefaultHttp2LocalFlowController.DEFAU
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+
 import io.grpc.Attributes;
-import io.grpc.Grpc;
 import io.grpc.Metadata;
 import io.grpc.Status;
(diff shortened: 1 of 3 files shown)


ejona86:
CC @lukaszx0, @kkaneda

kkaneda:
Thanks! LGTM.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -41,8 +41,8 @@
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+
```
ejona86: Remove the newlines; that's not the new style. Ditto elsewhere.

netty/src/main/java/io/grpc/netty/ProtocolNegotiator.java
```diff
@@ -53,7 +53,8 @@
 
   /**
    * Creates a new handler to control the protocol negotiation. Once the negotiation
-   * has completed successfully, the provided handler is installed.
+   * has completed successfully, the provided handler is installed. Must call {@code
+   * grpcHandler.onHandleProtocolNegotiationCompleted()} at certain point.
```
ejona86: If the negotiation fails, it isn't necessary to call the completed callback.

Issue #2636, opened 2017-01-21
ServiceDescriptor constructor is error prone


carl-mastrangelo:
The `ServiceDescriptor(String, Collection<MethodDescriptor<?, ?>)` constructor is error prone.  Getting the generics slightly wrong results in `ServiceDescriptor(String, Object, MethodDescriptor<?, ? ...)` being invoked instead.  This can happen if the second arg is `List<MethodDescriptor<Void, Void>>` for example.

ejona86:
Those constructors are not ExperimentalApi, so we have to address it in some way now.

carl-mastrangelo:
Deprecated sounds appropriate.(an idle threat)

ejona86:
The Object constructor was added since 1.0, so we can mark it as experimental now, except for any issues caused by codegen. And I do think codegen may be using one of them.

carl-mastrangelo:
If code gen is using it, it cant be experimental.  It has either got to be removed, or deprecated.  Else it would break compat.

PR #2667, opened 2017-01-27
core: make ServiceDescriptor use the Builder pattern

carl-mastrangelo:
Fixes: #2636

Also fixes a minor bug where methods could have been null inside the method list.

### DIFF
--- a/core/src/main/java/io/grpc/ServiceDescriptor.java
+++ b/core/src/main/java/io/grpc/ServiceDescriptor.java
@@ -55,7 +55,7 @@ public final class ServiceDescriptor {
   private final Object schemaDescriptor;
 
   /**
-   * Constructs a new Service Descriptor.  Users are encouraged to used {@link #newBuilder()}
+   * Constructs a new Service Descriptor.  Users are encouraged to use {@link #newBuilder()}
    * instead.
    *
    * @param name The name of the service
@@ -67,7 +67,7 @@ public final class ServiceDescriptor {
   }
 
   /**
-   * Constructs a new Service Descriptor.  Users are encouraged to used {@link #newBuilder()}
+   * Constructs a new Service Descriptor.  Users are encouraged to use {@link #newBuilder()}
    * instead.
    *
    * @param name The name of the service
@@ -134,7 +134,8 @@ public final class ServiceDescriptor {
   }
 
   /**
-   * @return Creates a new builder for a {@link ServiceDescriptor}.
+   * Creates a new builder for a {@link ServiceDescriptor}.
+   *
    * @since 1.1.0
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/2666")
--- a/core/src/main/java/io/grpc/ServiceDescriptor.java
+++ b/core/src/main/java/io/grpc/ServiceDescriptor.java
@@ -79,7 +79,7 @@ public final class ServiceDescriptor {
   }
 
   private ServiceDescriptor(Builder b) {
-    this.name = checkNotNull(b.name, "name");
+    this.name = b.name;
     validateMethodNames(name, b.methods);
     this.methods = Collections.unmodifiableList(new ArrayList<MethodDescriptor<?, ?>>(b.methods));
     this.schemaDescriptor = b.schemaDescriptor;
--- a/core/src/main/java/io/grpc/ServiceDescriptor.java
+++ b/core/src/main/java/io/grpc/ServiceDescriptor.java
@@ -163,6 +163,7 @@ public final class ServiceDescriptor {
      * @return this builder.
      * @since 1.1.0
      */
+    @ExperimentalApi("https://github.com/grpc/grpc-java/issues/2666")
     public Builder setName(String name) {
       this.name = checkNotNull(name, "name");
       return this;
@@ -176,7 +177,7 @@ public final class ServiceDescriptor {
      * @since 1.1.0
      */
     public Builder addMethod(MethodDescriptor<?, ?> method) {
-      methods.add(method);
+      methods.add(checkNotNull(method, "method"));
       return this;
     }
 
--- a/core/src/main/java/io/grpc/ServiceDescriptor.java
+++ b/core/src/main/java/io/grpc/ServiceDescriptor.java
@@ -32,49 +32,64 @@
 package io.grpc;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.common.base.Preconditions;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 import javax.annotation.Nullable;
 
 /**
  * Descriptor for a service.
+ *
+ * @since 1.0.0
  */
 public final class ServiceDescriptor {
 
   private final String name;
   private final Collection<MethodDescriptor<?, ?>> methods;
-  private final Object marshallerDescriptor;
+  private final Object schemaDescriptor;
 
+  /**
+   * Constructs a new Service Descriptor.  Users are encouraged to used {@link #newBuilder()}
+   * instead.
+   *
+   * @param name The name of the service
+   * @param methods The methods that are part of the service
+   * @since 1.0.0
+   */
   public ServiceDescriptor(String name, MethodDescriptor<?, ?>... methods) {
-    this(name, null, Arrays.asList(methods));
+    this(name, Arrays.asList(methods));
   }
 
+  /**
+   * Constructs a new Service Descriptor.  Users are encouraged to used {@link #newBuilder()}
+   * instead.
+   *
+   * @param name The name of the service
+   * @param methods The methods that are part of the service
+   * @since 1.0.0
+   */
   public ServiceDescriptor(String name, Collection<MethodDescriptor<?, ?>> methods) {
-    this(name, null, methods);
+    this(newBuilder().setName(name).addAllMethods(checkNotNull(methods, "methods")));
   }
 
-  public ServiceDescriptor(String name, Object marshallerDescriptor,
-                           MethodDescriptor<?, ?>... methods) {
-    this(name, marshallerDescriptor, Arrays.asList(methods));
+  private ServiceDescriptor(Builder b) {
+    this.name = checkNotNull(b.name, "name");
+    validateMethodNames(name, b.methods);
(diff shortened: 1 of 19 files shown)


ejona86:
Jenkins, retest this please

core/src/main/java/io/grpc/ServiceDescriptor.java
```diff
@@ -32,74 +32,98 @@
...
  */
 public final class ServiceDescriptor {
 
   private final String name;
   private final Collection<MethodDescriptor<?, ?>> methods;
-  private final Object marshallerDescriptor;
+  private final Object schemaDescriptor;
 
+  /**
+   * Constructs a new Service Descriptor.  Users are encouraged to used {@link #newBuilder()}
```
ericgribkoff: s/used/use

core/src/main/java/io/grpc/ServiceDescriptor.java
```diff
@@ -32,74 +32,98 @@
...
+   * @param methods The methods that are part of the service
+   * @since 1.0.0
+   */
   public ServiceDescriptor(String name, MethodDescriptor<?, ?>... methods) {
-    this(name, null, Arrays.asList(methods));
+    this(name, Arrays.asList(methods));
   }
 
+  /**
+   * Constructs a new Service Descriptor.  Users are encouraged to used {@link #newBuilder()}
```
ericgribkoff: s/used/use

core/src/main/java/io/grpc/ServiceDescriptor.java
```diff
@@ -108,4 +132,88 @@ static void validateMethodNames(String serviceName, Collection<MethodDescriptor<
           "duplicate name %s", method.getFullMethodName());
     }
   }
+
+  /**
+   * @return Creates a new builder for a {@link ServiceDescriptor}.
```
zhangkun83: Remove `@return`?

ejona86:
retest this please

core/src/main/java/io/grpc/ServiceDescriptor.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/ServiceDescriptor.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/ServiceDescriptor.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/ServiceDescriptor.java
```diff
@@ -32,74 +32,98 @@
...
-  /** Creates a new ServiceDescriptor. */
-  public ServiceDescriptor(String name, Object marshallerDescriptor,
-                           Collection<MethodDescriptor<?, ?>> methods) {
-    this.name = Preconditions.checkNotNull(name, "name");
-    Preconditions.checkNotNull(methods, "methods");
-    validateMethodNames(name, methods);
-    this.marshallerDescriptor = marshallerDescriptor;
-    this.methods = Collections.unmodifiableList(new ArrayList<MethodDescriptor<?, ?>>(methods));
+  private ServiceDescriptor(Builder b) {
+    this.name = checkNotNull(b.name, "name");
```
ejona86: I think this needs to be a IllegalStateException instead of IllegalArgumentException. So checkState.

core/src/main/java/io/grpc/ServiceDescriptor.java
```diff
@@ -108,4 +132,89 @@ static void validateMethodNames(String serviceName, Collection<MethodDescriptor<
...
     }
   }
+
+  /**
+   * Creates a new builder for a {@link ServiceDescriptor}.
+   *
+   * @since 1.1.0
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/2666")
+  public static Builder newBuilder() {
```
ejona86: Bikeshedding: I'd like name to be specified here, because then you can call build() at any time without needing to worry about runtime exceptions.

core/src/main/java/io/grpc/ServiceDescriptor.java
```diff
@@ -108,4 +132,89 @@ static void validateMethodNames(String serviceName, Collection<MethodDescriptor<
...
+    return new Builder();
+  }
+
+  /**
+   * A builder for a {@link ServiceDescriptor}.
+   *
+   * @since 1.1.0
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/2666")
```
ejona86: It can't be experimental, if being used by the generated code.

core/src/main/java/io/grpc/ServiceDescriptor.java
```diff
@@ -108,4 +132,89 @@ static void validateMethodNames(String serviceName, Collection<MethodDescriptor<
...
+     * @return this builder.
+     * @since 1.1.0
+     */
+    public Builder setName(String name) {
+      this.name = name;
+      return this;
+    }
+
+    /**
+     * Adds a methods to this builder.  This should be non-{@code null}.
```
ejona86: s/methods/method/

ejona86: Maybe "Adds a method to this service"?

core/src/main/java/io/grpc/ServiceDescriptor.java
carl-mastrangelo: I have come from writing a lot of proto code, where `build()` is when invariants are enforced.  I feel there are a lot of people who also have this expectations, just based on the number of them who write code at Google.  

@zhangkun83  thoughts?

zhangkun83: They both have pros and cons. I slightly lean towards not having name here, because it's more consistent. `build()` is usually very close to `newBuilder()`, so I think it should be easy to make sure `setName()` is called. `build()` should have a javadoc says that `setName()` must be called.

ejona86: I don't think proto is a good comparison, because it could not do the pattern I'm suggesting. And I think with proto people expect to build() to fail the first couple of times they run their program. That is needless in this case. If there are a lot of invariants I'd expect to do something different.

Looking through Guava, I don't see a _single_ Builder that can be in an incomplete state. Their API is designed where you simply can't build it wrong. The main non-trivial example is [GraphBuilder](http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/graph/GraphBuilder.html)/NetworkBuilder/ValueGraphBuilder with directed()/undirected(). There are some older+weaker examples like [ThreadFactoryBuilder](http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/util/concurrent/ThreadFactoryBuilder.html) which falls back to defaults and [MinMaxPriorityQueue.Builder](http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/MinMaxPriorityQueue.Builder.html) which you provide an Iterable to the `create()` (which is basically equivalent to `build()`).

carl-mastrangelo: Guava Builders is not nearly as used as Proto Builders, so that isn't a sufficient reason to mirror the proto behavior.  (Imagine if proto builders forced required fields be in the `newBuilder()` constructor.) 

Your proposal is not *clearly* better than the existing one.  It may be, but it isn't certainly better.  Considering that there is significant precedent for not requiring it, I would prefer to leave it as is.

core/src/main/java/io/grpc/ServiceDescriptor.java
carl-mastrangelo: Done.

carl-mastrangelo: Done.

core/src/main/java/io/grpc/ServiceDescriptor.java
carl-mastrangelo: Added a javadoc comment @zhangkun83

core/src/main/java/io/grpc/ServiceDescriptor.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/ServiceDescriptor.java
ejona86: Comparing to protobuf doesn't help the discussion. It doesn't matter what protobuf does because it could not have done what I'm describing. If I was making the protobuf builder I would have done exactly like they would have, because there is no other option.

More examples requiring:
 * [ProcessBuilder](http://docs.oracle.com/javase/8/docs/api/java/lang/ProcessBuilder.html)
 * [DateTimeFormatterBuilder](http://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatterBuilder.html)'s `toFormatter(Locale)`

I (surprisingly, honestly) have not found an example in a similar circumstance that follows your recommendation.

ejona86: I'll also note that ManagedChannelBuilder, ServerBuilder, and ServerServiceDefinition require required fields when making the builder.

carl-mastrangelo:
Done.  PTAL

core/src/main/java/io/grpc/ServiceDescriptor.java
```diff
@@ -108,4 +132,90 @@ static void validateMethodNames(String serviceName, Collection<MethodDescriptor<
...
+    private Object schemaDescriptor;
+
+    /**
+     * Sets the name.  This should be non-{@code null}.
+     *
+     * @param name The name of the service.
+     * @return this builder.
+     * @since 1.1.0
+     */
+    public Builder setName(String name) {
```
carl-mastrangelo: Done.

core/src/main/java/io/grpc/ServiceDescriptor.java
```diff
@@ -108,4 +132,90 @@ static void validateMethodNames(String serviceName, Collection<MethodDescriptor<
...
+
+    /**
+     * Adds a method to this service.  This should be non-{@code null}.
+     *
+     * @param method the method to add to the descriptor.
+     * @return this builder.
+     * @since 1.1.0
+     */
+    public Builder addMethod(MethodDescriptor<?, ?> method) {
+      methods.add(method);
```
carl-mastrangelo: Done.

carl-mastrangelo:
retest this please

PR #2669, opened 2017-01-27
testing: change names of noopMarshaller to void marshaller

carl-mastrangelo:
This is needed because in interceptor tests, often the types cannot
be changed.  The void methods stay for users who are writing tests
where they actually don't care about types.  The noop methods
require types to be specified.  This is for users who don't care
about the implementation.  These represent different levels of
commitment.

This eases the transistion of code Mocking MethodDescriptor, which
breaks in this release.


ejona86:
It's not clear: are you planning to delete noopMarshaller? "change names" would seem to imply so, but your commit description seems to imply not so.

carl-mastrangelo:
no, I think both methods are useful.

PR #2671, opened 2017-01-30
netty: work around netty bug when RPCs may be reset early

carl-mastrangelo:
For https://github.com/netty/netty/issues/6296


carl-mastrangelo:
Also, I manually confirmed this test works with 100 runs.

Scottmitch:
I commented on https://github.com/netty/netty/issues/6296 ... I'm not yet convinced this is a Netty bug and need more info.

PR #2672, opened 2017-01-30
netty: fix race in server channel startup and shutdown

carl-mastrangelo:
This is an expansion of https://github.com/grpc/grpc-java/commit/2292cbfca091a4c6bb315a291929c656a81b90b6

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServer.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServer.java
@@ -123,13 +123,6 @@ class NettyServer implements InternalServer {
     b.childHandler(new ChannelInitializer<Channel>() {
       @Override
       public void initChannel(Channel ch) throws Exception {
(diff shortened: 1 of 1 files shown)


netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -140,6 +133,15 @@ public void operationComplete(ChannelFuture future) {
             ch.close();
             return;
           }
+          // `channel` shutdown can race with `ch` initialization, so this is only safe to increment
+          // inside the lock.
+          eventLoopReferenceCounter.retain();
+          ch.closeFuture().addListener(new ChannelFutureListener() {
```
ejona86: This addListener _could_ be placed after the synchronized block, but I honestly don't care.

carl-mastrangelo: I moved the whole chain.

PR #2673, opened 2017-01-31
core: record server_elapsed_time on client

zhangkun83:
It is defined as the time between the client sends out the headers, and
the RPC finishes.

### DIFF
--- a/core/src/main/java/io/grpc/internal/StatsTraceContext.java
+++ b/core/src/main/java/io/grpc/internal/StatsTraceContext.java
@@ -251,7 +251,7 @@ public final class StatsTraceContext {
     }
     long roundtripNanos = stopwatch.elapsed(TimeUnit.NANOSECONDS);
     MeasurementMap.Builder builder = MeasurementMap.builder()
-        .put(latencyMetric, roundtripNanos / NANOS_PER_MILLI)
+        .put(latencyMetric, roundtripNanos / NANOS_PER_MILLI)  // in double
         .put(wireBytesSentMetric, wireBytesSent)
         .put(wireBytesReceivedMetric, wireBytesReceived)
         .put(uncompressedBytesSentMetric, uncompressedBytesSent)
@@ -260,7 +260,7 @@ public final class StatsTraceContext {
       if (clientPendingNanos >= 0) {
         builder.put(
             RpcConstants.RPC_CLIENT_SERVER_ELAPSED_TIME,
-            (roundtripNanos - clientPendingNanos) / NANOS_PER_MILLI);
+            (roundtripNanos - clientPendingNanos) / NANOS_PER_MILLI);  // in double
       }
     }
     statsCtx.with(RpcConstants.RPC_STATUS, TagValue.create(status.getCode().toString()))
--- a/core/src/main/java/io/grpc/internal/StatsTraceContext.java
+++ b/core/src/main/java/io/grpc/internal/StatsTraceContext.java
@@ -32,6 +32,7 @@
 package io.grpc.internal;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Stopwatch;
 import com.google.common.base.Supplier;
 import com.google.instrumentation.stats.MeasurementDescriptor;
@@ -72,6 +73,7 @@ public final class StatsTraceContext {
   private final Stopwatch stopwatch;
   private final Side side;
   private final Metadata.Key<StatsContext> statsHeader;
+  private volatile long clientPendingMillis = -1;
   private volatile long wireBytesSent;
   private volatile long wireBytesReceived;
   private volatile long uncompressedBytesSent;
@@ -205,6 +207,18 @@ public final class StatsTraceContext {
   }
 
   /**
+   * Mark the time when the headers, which are the first bytes of the RPC, are sent from the client.
+   * This is specific to transport implementation, thus should be called from transports.  Calling
+   * it the second time or more is a no-op.
+   */
+  public void clientHeadersSent() {
+    Preconditions.checkState(side == Side.CLIENT, "Must be called on client-side");
+    if (clientPendingMillis < 0) {
+      clientPendingMillis = stopwatch.elapsed(TimeUnit.MILLISECONDS);
+    }
+  }
+
+  /**
    * Record a finished all and mark the current time as the end time.
    *
    * <p>Can be called from any thread without synchronization.  Calling it the second time or more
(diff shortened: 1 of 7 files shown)


zhangkun83:
retest this please

core/src/main/java/io/grpc/internal/StatsTraceContext.java
```diff
@@ -72,6 +73,7 @@
   private final Stopwatch stopwatch;
   private final Side side;
   private final Metadata.Key<StatsContext> statsHeader;
+  private volatile long clientPendingMillis = -1;
```
carl-mastrangelo: millis are rather coarse, no?  I think it would be better if we used nanos for all time information throughout the library.

core/src/main/java/io/grpc/internal/StatsTraceContext.java
```diff
@@ -233,14 +247,20 @@ void callEnded(Status status) {
...
-    statsCtx
-        .with(RpcConstants.RPC_STATUS, TagValue.create(status.getCode().toString()))
-        .record(MeasurementMap.builder()
-            .put(latencyMetric, stopwatch.elapsed(TimeUnit.MILLISECONDS))
-            .put(wireBytesSentMetric, wireBytesSent)
-            .put(wireBytesReceivedMetric, wireBytesReceived)
-            .put(uncompressedBytesSentMetric, uncompressedBytesSent)
-            .put(uncompressedBytesReceivedMetric, uncompressedBytesReceived)
-            .build());
+    long roundtrip = stopwatch.elapsed(TimeUnit.MILLISECONDS);
```
carl-mastrangelo: include the units in the variable name?

core/src/test/java/io/grpc/internal/StatsTraceContextTest.java
```diff
@@ -0,0 +1,229 @@
...
+ * Test for {@link StatsTraceContext}.
+ */
+@RunWith(JUnit4.class)
+public class StatsTraceContextTest {
+  private FakeClock fakeClock;
+  private FakeStatsContextFactory statsCtxFactory;
+
+  @Before
+  public void setUp() {
+    fakeClock = new FakeClock();
```
carl-mastrangelo: Just initialize these above?

core/src/test/java/io/grpc/internal/StatsTraceContextTest.java
```diff
@@ -0,0 +1,229 @@
...
+
+  @After
+  public void allRecordsVerified() {
+    assertNull(statsCtxFactory.pollRecord());
+  }
+
+  @Test
+  public void clientBasic() {
+    StatsTraceContext ctx = StatsTraceContext.newClientContextForTesting(
+        "Service1/method1", statsCtxFactory, statsCtxFactory.getDefault(),
```
carl-mastrangelo: let's use the recommended full method name, with the leading `/`

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -98,6 +99,7 @@
       String userAgent,
       StatsTraceContext statsTraceCtx) {
     super(new OkHttpWritableBufferAllocator(), maxMessageSize, statsTraceCtx);
+    this.statsTraceCtx = statsTraceCtx;
```
carl-mastrangelo: checkNotNull?

core/src/test/java/io/grpc/internal/StatsTraceContextTest.java
zhangkun83: Any reference for that? The generated code has something like:
```java
generateFullMethodName(
              "grpc.testing.TestService", "UnaryCall"),
```
And `MethodDescriptor.generateFullMethodName()` doesn't add the leading `/`

core/src/main/java/io/grpc/internal/StatsTraceContext.java
zhangkun83: Switched to nanos.

core/src/main/java/io/grpc/internal/StatsTraceContext.java
zhangkun83: Done.

core/src/test/java/io/grpc/internal/StatsTraceContextTest.java
zhangkun83: Done.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
zhangkun83: Done.

zhangkun83:
Thanks. PTAL.

core/src/test/java/io/grpc/internal/StatsTraceContextTest.java
carl-mastrangelo: Hmm, you are right.  The spec was updated that the path should have the leading slash.  Some places add it, some does not. This is going to be confusing.

https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

zhangkun83: The spec defines path having a leading slash, but never spells out that the path and the fullly qualified method name are equivalent.

It seems throughout Java code base the method name is assumed to be sans-leading-slash. For example, `MethodDescriptor.extractFullServiceName()` does so and is used on the server-side too.

core/src/test/java/io/grpc/internal/StatsTraceContextTest.java
```diff
@@ -0,0 +1,222 @@
...
+        record.getMetricAsLongOrFail(RpcConstants.RPC_CLIENT_UNCOMPRESSED_REQUEST_BYTES));
+    assertEquals(0, record.getMetricAsLongOrFail(RpcConstants.RPC_CLIENT_RESPONSE_BYTES));
+    assertEquals(0,
+        record.getMetricAsLongOrFail(RpcConstants.RPC_CLIENT_UNCOMPRESSED_RESPONSE_BYTES));
+    assertEquals(3000, record.getMetricAsLongOrFail(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY));
+    assertNull(record.getMetric(RpcConstants.RPC_CLIENT_SERVER_ELAPSED_TIME));
+  }
+
+  @Test
+  public void tagPropagation() {
```
adriancole: maybe add javadoc about why this is interesting (in relation to stats). Ex I presume this could be a log correlation or trace identifier

core/src/main/java/io/grpc/internal/StatsTraceContext.java
```diff
@@ -233,14 +249,21 @@ void callEnded(Status status) {
...
-        .record(MeasurementMap.builder()
-            .put(latencyMetric, stopwatch.elapsed(TimeUnit.MILLISECONDS))
-            .put(wireBytesSentMetric, wireBytesSent)
-            .put(wireBytesReceivedMetric, wireBytesReceived)
-            .put(uncompressedBytesSentMetric, uncompressedBytesSent)
-            .put(uncompressedBytesReceivedMetric, uncompressedBytesReceived)
-            .build());
+    long roundtripNanos = stopwatch.elapsed(TimeUnit.NANOSECONDS);
+    MeasurementMap.Builder builder = MeasurementMap.builder()
+        .put(latencyMetric, roundtripNanos / NANOS_PER_MILLI)
```
carl-mastrangelo: TimeUnit.NANOSECONDS.toMillis

core/src/main/java/io/grpc/internal/StatsTraceContext.java
```diff
@@ -233,14 +249,21 @@ void callEnded(Status status) {
...
+        .put(wireBytesSentMetric, wireBytesSent)
+        .put(wireBytesReceivedMetric, wireBytesReceived)
+        .put(uncompressedBytesSentMetric, uncompressedBytesSent)
+        .put(uncompressedBytesReceivedMetric, uncompressedBytesReceived);
+    if (side == Side.CLIENT) {
+      long serverTime;
```
carl-mastrangelo: unused?

core/src/main/java/io/grpc/internal/StatsTraceContext.java
```diff
@@ -233,14 +249,21 @@ void callEnded(Status status) {
...
+      if (clientPendingNanos >= 0) {
+        builder.put(RpcConstants.RPC_CLIENT_SERVER_ELAPSED_TIME,
```
carl-mastrangelo: optional: rectangle rule says each arg should get its own line

core/src/test/java/io/grpc/internal/StatsTraceContextTest.java
carl-mastrangelo: Want to call `generateFullMethodName` and remove any ambiguity?

zhangkun83: Done.

core/src/test/java/io/grpc/internal/StatsTraceContextTest.java
zhangkun83: Added comments. The stats tags can be used for associating RPC costs with custom dimensions, e.g., userId. Trace identifier is similar, but will be done separately as part of the tracing work.

core/src/main/java/io/grpc/internal/StatsTraceContext.java
zhangkun83: No, the measurements are doubles. Although the unit is millisecond, it can still retain nano-level precision.

core/src/main/java/io/grpc/internal/StatsTraceContext.java
zhangkun83: Removed.

core/src/main/java/io/grpc/internal/StatsTraceContext.java
zhangkun83: Done.

zhangkun83:
@carl-mastrangelo PTAL

core/src/main/java/io/grpc/internal/StatsTraceContext.java
carl-mastrangelo: That is quite surprising, maybe worth a comment?  It isn't obvious from just reading the code.

zhangkun83: Done.

PR #2678, opened 2017-01-31
core: clean up code deprecated in 1.1

lukaszx0:
Since we're on 1.2 train, we can clean this up now :)

### DIFF
--- a/testing/src/main/java/io/grpc/testing/DeadlineSubject.java
+++ b/testing/src/main/java/io/grpc/testing/DeadlineSubject.java
@@ -110,6 +110,8 @@ public final class DeadlineSubject extends ComparableSubject<DeadlineSubject, De
      * @deprecated {@link Object#equals(Object)} is not supported on TolerantDeadlineComparison
      *     If you meant to compare deadlines, use {@link #of(Deadline)} instead.
      */
+    // Deprecation used to signal visual warning in IDE for the unaware users.
+    // This method is created as a precaution and won't be removed as part of deprecation policy.
     @Deprecated
     @Override
     public boolean equals(@Nullable Object o) {
@@ -123,6 +125,8 @@ public final class DeadlineSubject extends ComparableSubject<DeadlineSubject, De
      * @throws UnsupportedOperationException always
      * @deprecated {@link Object#hashCode()} is not supported on TolerantDeadlineComparison
      */
+    // Deprecation used to signal visual warning in IDE for the unaware users.
+    // This method is created as a precaution and won't be removed as part of deprecation policy.
     @Deprecated
     @Override
     public int hashCode() {
--- a/core/src/main/java/io/grpc/ServerCall.java
+++ b/core/src/main/java/io/grpc/ServerCall.java
@@ -32,8 +32,6 @@
 package io.grpc;
 
 import com.google.errorprone.annotations.DoNotMock;
-import java.net.SocketAddress;
-import javax.net.ssl.SSLSession;
 
 /**
  * Encapsulates a single call received from a remote client. Calls may not simply be unary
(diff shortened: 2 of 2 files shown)


lukaszx0:
What's the deprecation policy then? (ie. when we're free to clean up stuff?) Also, `ClientAuthInterceptor` seems like the one which is very easily "backportable" - just copy it into your repo internally, even leaving original package name.

ejona86:
As far as our deprecation process, we are free to delete it in 1.1 because it was deprecated in 1.0. But since we've not migrated users over to the new thing yet in our codebase we've not made sure the process is smooth. So it's easy to just leave it there for now and remove it once we're sure the next thing and process is ready.

lukaszx0:
That's fair, and I don't have the intention to push for it, but wouldn't it be better to get rid of it now, and motivate users to make a switch before 1.2 will be release? (which will take some time). That'd work unless you guys aren't building against HEAD, obviously, in which case I'll revert deletion of `ClientAuthInterceptor` :)

lukaszx0:
@ejona86 @carl-mastrangelo I've restored the `ClientAuthInterceptor`.

testing/src/main/java/io/grpc/testing/DeadlineSubject.java
```diff
@@ -102,32 +101,6 @@ private TolerantDeadlineComparison() {}
...
     public abstract void of(Deadline expectedDeadline);
-
-    /**
-     * Do not call this method.
-     *
-     * @throws UnsupportedOperationException always
-     * @deprecated {@link Object#equals(Object)} is not supported on TolerantDeadlineComparison
-     *     If you meant to compare deadlines, use {@link #of(Deadline)} instead.
-     */
-    @Deprecated
```
carl-mastrangelo: This is actually here to scare people away from it.  I think the deprecated is present to make IDEs catch on more quickly that they shouldn't be using it.  The exception is to enforce it.  Truth makes it easy to accidentally call:


assertAbout(DeadlineSubject.deadline()).that(myDeadline).equals(otherDeadline);

vs.

assertAbout(DeadlineSubject.deadline()).that(myDeadline).IsEqual(otherDeadline);

The first asserts that the DeadlineSubject itself is equal  to the deadline, where as the latter compares the subject to the object.

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -197,17 +197,6 @@ public NettyChannelBuilder flowControlWindow(int flowControlWindow) {
   }
 
   /**
-   * Sets the max message size.
-   *
-   * @deprecated Use {@link #maxInboundMessageSize} instead
-   */
-  @Deprecated
-  public NettyChannelBuilder maxMessageSize(int maxMessageSize) {
```
carl-mastrangelo: We still use this method internally, and while it can be removed per the compat guidelines, we wouldn't be able to sync gRPC internally.  We find a fair number of issues when doing the sync, so removing this now would cause more issues than it fixes.  

I am undertaking the fixing of users, but it takes time.

testing/src/main/java/io/grpc/testing/DeadlineSubject.java
lukaszx0: Ah, makes sense. I've added comment clarifying that.

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
lukaszx0: Reverted.

lukaszx0:
@carl-mastrangelo is it okay now, in this form?

PR #2690, opened 2017-02-03
services: add monitoring proto

ericgribkoff:
This is the monitoring service proto for exporting stats.

The only true new file is `monitoring.proto`. It depends on `census.proto`, and for the moment I've just copied `census.proto` from https://github.com/google/instrumentation-proto/blob/master/stats/census.proto. This should probably be imported directly from the instrumentation-proto repo rather than duplicated here.

@makdharma @a11r FYI


services/src/main/proto/monitoring.proto
```diff
@@ -0,0 +1,130 @@
...
+  string name = 1;  // name of a group of monitoring data
+}
+
+// The wrapper for custom monitoring data.
+message CustomMonitoringData {
+  // can be any application specific monitoring data
+  google.protobuf.Any contents = 1;
+}
+
+message Empty {}
```
mavrukin: this is not necessary, just use the proto3.empty type

services/src/main/proto/monitoring.proto
```diff
@@ -0,0 +1,130 @@
...
+package grpc.instrumentation.v1alpha;
+
+option java_multiple_files = true;
+option java_package = "io.grpc.instrumentation.v1alpha";
+option java_outer_classname = "MonitoringProto";
+
+service Monitoring {
+
+  // Return canonical RPC stats
+  rpc GetCanonicalRpcStats(Empty) returns (CanonicalRpcStats) {
```
mavrukin: please take a look at my comment below, i.e., you want to use google.protobuf.Empty and not roll your own

services/src/main/proto/census.proto
```diff
@@ -0,0 +1,318 @@
+// Copyright 2016, Google Inc.
```
mavrukin: What's the relationship of this census.proto w/ the monitoring.proto?  They don't seem to actually be related, at least not on the surface, does it make sense to put them in as the same commit or split them up?

services/src/generated/main/grpc/io/grpc/instrumentation/v1alpha/MonitoringGrpc.java
```diff
@@ -0,0 +1,500 @@
+package io.grpc.instrumentation.v1alpha;
```
mavrukin: We really don't want to be committing generated source, please take it out.  You can do it most easily by just adding this: https://github.com/mavrukin/grpc-java/blob/master/.gitignore#L31 to your .gitignore file and that will take care of it

services/src/main/proto/monitoring.proto
ericgribkoff: Done for now. The interop tests do not use google.protobuf.Empty and instead define their own Empty message type, so there may be a reason to prefer this here as well?

services/src/main/proto/monitoring.proto
ericgribkoff: Done.

services/src/main/proto/census.proto
ericgribkoff: This proto defines the types returned by monitoring.proto. Since it's just cloned from the instrumentation-java repository, this can be replaced with a dependency (e.g., a git submodule) later.

services/src/generated/main/grpc/io/grpc/instrumentation/v1alpha/MonitoringGrpc.java
ericgribkoff: grpc-java includes generated code. See https://github.com/grpc/grpc-java/tree/master/services/src/generated/main/java/io/grpc/health/v1

services/src/main/proto/census.proto
```diff
@@ -0,0 +1,318 @@
...
+  //
+  // The size of `bucket_count` must be no greater than N as defined in
+  // `bucket_boundaries`.
+  //
+  // Any suffix of trailing zero bucket_count fields may be omitted.
+  repeated int64 bucket_counts = 5;
+
+  // Tags associated with this DistributionAggregation. These will be filled
+  // in based on the View specification.
+  repeated Tag tags = 6;
```
adriancole: sorry if this is described somewhere else, but why not `map<string, string>`? are we intentionally supporting multiple values per key?

ericgribkoff: I am not sure, but I believe there is some support (potentially legacy) in the [`instrumentation`](https://github.com/google/instrumentation-java) library for multiple values per key. I don't think this will arise in gRPC, but we are relying on instrumentation as the source of these stats. This proto is copied from their repository, https://github.com/google/instrumentation-proto/blob/master/stats/census.proto.

ejona86:
Is there a PR that depends on this? It seems strange to add it without anything using it.

> The interop test protos instead define their own Empty message - is there any reason to avoid using google.protobuf.Empty here?

That is historical. The interop tests date back to before the well-known protos were handled well. Multiple build systems had trouble "include"ing the `.proto` and others had trouble finding the generated code. That's long been solved, but nobody's cleaned it up.

services/src/main/proto/census.proto
```diff
@@ -0,0 +1,318 @@
...
+// limitations under the License.
+
+//TODO(ericgribkoff) Depend on this directly from the instrumentation-proto
+//repository.
+
+syntax = "proto3";
+
+package google.instrumentation;
+
+option java_package = "com.google.instrumentation.stats.proto";
```
ejona86: It's strange java_multiple_files isn't true.

a-veitch: We really shouldn't make a copy of this, without some TODO reflecting that it *is* a copy, and the right thing to do is depend on the source version (https://github.com/google/instrumentation-proto/blob/master/stats/census.proto). Is there some reason you can't make that a submodule?

ericgribkoff:
The monitoring proto is now going into the `grpc-proto` repository (https://github.com/grpc/grpc-proto/pull/2). Reviewing this PR can now wait until I add the monitoring service later this week.

ericgribkoff:
This PR is superseded by https://github.com/grpc/grpc-java/pull/2736

Issue #2691, opened 2017-02-03
ManagedClientTransport.start should be better defined


carl-mastrangelo:
Calling  `ManagedClientTransport.start` returns a Runnable, but it isn't clear what to do with it, or when it should be called.  While the code is annotated `@CheckReturnValue`, the annotation doesn't get inherited across class boundaries, leading to places where the value is ignored.  `NettyClientTransportTest` contains several invocations of `start`, but ignores the return value due to `NettyClientTransport` currently returning null.  

The difficulty of describing when and how the Runnable should be used feels indicative of shaky API.  It would be nice to modify InternalSubchannel to not need to acquire locks to alleviate the need to run code outside of locks.  This is not urgent.

@zhangkun83

PR #2692, opened 2017-02-03
netty: add return value errorprone annotations

carl-mastrangelo:
Related to #2691.  This requires callers of the netty code to check their return values.  Exceptions are for the Channel builder  and the Server builder, since it's usually okay to not use the result of a setter.  

I eventually would like to add this to more parts of our code, but I'm starting with a small part first.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
+++ b/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
@@ -33,6 +33,7 @@ package io.grpc.netty;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import io.grpc.ExperimentalApi;
 import io.netty.handler.codec.http2.Http2SecurityUtil;
 import io.netty.handler.ssl.ApplicationProtocolConfig;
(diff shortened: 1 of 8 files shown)


carl-mastrangelo:
cc: @ejona86

netty/src/test/java/io/grpc/netty/NettyHandlerTestBase.java
```diff
@@ -221,6 +222,7 @@ protected final Http2Connection connection() {
     return handler().connection();
   }
 
+  @CanIgnoreReturnValue
```
zhangkun83: Why is this needed?

netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
```diff
@@ -133,6 +134,7 @@ public static SslContextBuilder forServer(
    * Set ciphers and APN appropriate for gRPC. Precisely what is set is permitted to change, so if
    * an application requires particular settings it should override the options set here.
    */
+  @CanIgnoreReturnValue
```
zhangkun83: Add `@CheckReturnValue` to the class?

netty/src/test/java/io/grpc/netty/NettyHandlerTestBase.java
carl-mastrangelo: Same reason as in `WriteQueue`, not all callers consume the future.  Netty's API usually returns a future that is passed in.  It looks like this code is trying to be like that.

netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
carl-mastrangelo: Because it is included in the package, it applies to all classes.  See package-info.java

PR #2694, opened 2017-02-03
interop-testing: fix flakes in Http2Client

ericgribkoff:
This fixes the Java flakes noted https://github.com/grpc/grpc/issues/9474.

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/Http2Client.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/Http2Client.java
@@ -153,8 +153,8 @@ public final class Http2Client {
 
   private void setUp() {
     channel = createChannel();
-    blockingStub = TestServiceGrpc.newBlockingStub(channel);
-    asyncStub = TestServiceGrpc.newStub(channel);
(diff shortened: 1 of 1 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/Http2Client.java
```diff
@@ -272,6 +272,7 @@ private void rstDuringData() throws Exception {
 
     private void goAway() throws Exception {
       assertResponseEquals(blockingStub.unaryCall(simpleRequest), goldenResponse);
+      TimeUnit.SECONDS.sleep(1);
```
dapengzhang0: I'm not clear about how the test is working. Where is verification that a GOAWAY was sent/seen?

ericgribkoff: At the moment, there is no verification on the client-side. The server asserts that the second call is seen on another stream: https://github.com/grpc/grpc/issues/9300 and https://github.com/grpc/grpc/blob/master/doc/negative-http2-interop-test-descriptions.md

PR #2695, opened 2017-02-05
core: expose #ping in Subchannel

lukaszx0:
This would be very useful. Would enable users to write load balancing logic based on the latency of the connections.

/cc @zhangkun83

### DIFF
--- a/core/src/main/java/io/grpc/LoadBalancer2.java
+++ b/core/src/main/java/io/grpc/LoadBalancer2.java
@@ -32,7 +32,9 @@
 package io.grpc;
 
 import com.google.common.base.Preconditions;
+import io.grpc.internal.ClientTransport;
 import java.util.List;
+import java.util.concurrent.Executor;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 import javax.annotation.concurrent.NotThreadSafe;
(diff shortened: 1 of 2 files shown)


ejona86:
Long-term, I question if the transports should just keep track of their latency. They know the latency initially via SETTINGS ack and then later via [flow control pinging](https://github.com/grpc/grpc-java/blob/89bc2cd/netty/src/main/java/io/grpc/netty/AbstractNettyHandler.java#L136).

But we may open this up in the short-term, since people were already using it in the previous LB setup.

lukaszx0:
Maybe it's not worth adding extending LB's API if it'll be soon replaced/removed? I've created an [issue to track this](https://github.com/grpc/grpc-java/issues/2717)

lukaszx0:
Talked with @ejona86 offline. We'd want a better solution than this and there's no urgency in having this merge for short term one.

ejona86:
We talked about having an API that is basically `notifyWhenLatencyOutsideRange(low, high)`. The important part being that most latency changes can be ignored. Didn't discuss whether the callback should be passed in or details like that. It would need to be passed the current latency. If the range is a mix of inclusive and exclusive, then `low==high` would guarantee a notification, which seems convenient.

PR #2696, opened 2017-02-06
core: replace EquivalentAddressGroup with ResolvedServerInfo

lukaszx0:
Now that we have a new LoadBalancer2 API in place, I'd like to propose removing `EquivalentAddressGroup` and replacing it with `ResolvedServerInfo`.

I think that having two, very similar objects adds unnecessary confusion and more importantly a lot of boilerplate that users need to deal with when custom LBs. With the new API and how things are designed (managing subchannels/connections _manually_ in LB) we don't need `EquivalentAddressGroup` anymore and can just add another dimension to the resolved server group of objects by allowing `ResolvedServerInfo` to store more than one address.

In my opinion, this would greatly simplify and unify `NameResolver`/`LoadBalancer` logic and code.

@zhangkun83 @ejona86 @carl-mastrangelo What do you think?

### DIFF
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -32,7 +32,11 @@
 package io.grpc.internal;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Iterators;
+import com.google.common.collect.Lists;
+
 import io.grpc.Attributes;
 import io.grpc.NameResolver;
 import io.grpc.ResolvedServerInfo;
@@ -43,7 +47,10 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.URI;
 import java.net.UnknownHostException;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
@@ -142,8 +149,7 @@ class DnsNameResolver extends NameResolver {
         try {
           if (System.getenv("GRPC_PROXY_EXP") != null) {
             ResolvedServerInfoGroup servers = ResolvedServerInfoGroup.builder()
-                .add(new ResolvedServerInfo(
-                    InetSocketAddress.createUnresolved(host, port), Attributes.EMPTY))
+                .add(new ResolvedServerInfo(InetSocketAddress.createUnresolved(host, port)))
                 .build();
             savedListener.onUpdate(Collections.singletonList(servers), Attributes.EMPTY);
             return;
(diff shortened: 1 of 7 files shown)


lukaszx0:
I just pushed a proof of concept to illustrate my proposal. It doesn't work at all and I didn't bother to change var names etc. mostly just changed types everywhere and did few minor code adjustments as I wanted to discuss it first (if we'll go with it, we probably want to consider renaming `ResolvedServerInfo` to maybe something more generic, too).

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -165,13 +171,18 @@ public void run() {
...
           for (int i = 0; i < inetAddrs.length; i++) {
-            InetAddress inetAddr = inetAddrs[i];
-            servers.add(
-                new ResolvedServerInfo(new InetSocketAddress(inetAddr, port), Attributes.EMPTY));
+            inetSocketAddrs.add(new InetSocketAddress(inetAddrs[i], port));
           }
-          savedListener.onUpdate(Collections.singletonList(servers.build()), Attributes.EMPTY);
+
+          ResolvedServerInfoGroup servers = ResolvedServerInfoGroup.builder()
+              .add(new ResolvedServerInfo(inetSocketAddrs))
```
lukaszx0: In the case of DNS name resolver, this actually makes way more sense and is semantically correct now. DNS can return multiple addresses for a given hostname, but it is logically still the same destination.

Before, we were creating multiple `ResolvedServerInfo`s which is very confusing and misleading when using DNS name resolver and custom LB - it sort of requires the implementor of LB to have a knowledge about how DNS NR in particular returns addresses and depend LB behaviour on this (making it less portable). Now we create only one object to represent one target host but backed by multiple alternative addresses.

zhangkun83:
The equality of `EquivalentAddressGroup`, which only counts the addresses but not the attributes, was needed by the v1 `ManagedChannelImpl` for de-duplicating `TransportSets`. Even though `ManagedChannelImpl2` doesn't forbid duplicating `InternalSubchannel` for the same `EquivalentAddressGroup`, the equality is still needed by most `LoadBalancer2` implementors, e.g., `Map<EquivalentAddressGroup, Subchannel> subchannels` in the round-robin LB, to determine which `InternalSubchannel`s to close and which to keep, when the NR result is updated. 

I agree that there are too many address wrappers and it has become very confusing, sometimes even to myself. I prefer to keep the `EquivalentAddressGroup` name, because it indicates how these addresses are used by the channel. So instead of deleting `EquivalentAddressGroup`, I suggest deleting `ResolvedServerInfo`, and adding attributes to `EquivalentAddressGroup`. To address the equality requirement I mentioned earlier, we could add a `withoutAttributes()` method which can be used in cases where attributes should not be counted for equality.

`ResolvedServerInfoGroup` was intended to be the attribute-ful counter-part of `EquivalentAddressGroup`. Since you are proposing to replace `EquivalentAddressGroup` with `ResolvedServerInfo`, I would expect `ResolvedServerInfoGroup` to go away instead of building yet another level on top of `ResolvedServerInfo`.

lukaszx0:
I should've start with the issue, not PR. My bad. Let's continue in #2716

PR #2697, opened 2017-02-06
core: un-abstract the LBv2 setter in ManagedChannelBuilder.

zhangkun83:
The LBv2 setter was added in #2658 as an abstract method, which breaks
anyone who implements ManagedChannelBuilder.  Here we add a default
implementation that throws.


ejona86:
Backported in #2702

PR #2235, opened 2016-09-06
netty: use custom http2 headers for decoding.

buchgr:
The DefaultHttp2Headers class is a general-purpose Http2Headers implementation
and provides much more functionality than we need in gRPC. In gRPC, when reading
headers off the wire, we only inspect a handful of them, before converting to
Metadata.

This commit introduces a Http2Headers implementation that aims for insertion
efficiency, a low memory footprint and fast conversion to Metadata.

  - Header names and values are stored in plain byte[].
  - Insertion is O(1), while lookup is now O(n).
  - Binary header values are base64 decoded as they are inserted.
  - The byte[][] returned by namesAndValues() can directly be used to construct
    a new Metadata object.
  - For HTTP/2 request headers, the pseudo headers are no longer carried over to
    Metadata.

A microbenchmark aiming to replicate the usage of Http2Headers in NettyClientHandler
and NettyServerHandler shows decent throughput gains when compared to DefaultHttp2Headers.

```
Benchmark                                             Mode  Cnt     Score    Error  Units
InboundHeadersBenchmark.defaultHeaders_clientHandler  avgt   10   262.158 ±  3.624  ns/op
InboundHeadersBenchmark.defaultHeaders_serverHandler  avgt   10  1016.699 ± 14.174  ns/op
InboundHeadersBenchmark.grpcHeaders_clientHandler     avgt   10   203.998 ±  2.329  ns/op
InboundHeadersBenchmark.grpcHeaders_serverHandler     avgt   10   549.832 ±  5.796  ns/op
```
Additionally, the memory footprint is reduced by more than 50%!
```
gRPC Request Headers: 864 bytes
Netty Request Headers: 1728 bytes
gRPC Response Headers: 216 bytes
Netty Response Headers: 528 bytes
```
Furthermore, this change does most of the gRPC groundwork necessary to be able
to cache higher ordered objects in HPACK's dynamic table, as discussed in [1].

[1] https://github.com/grpc/grpc-java/issues/2217

PR #2704, opened 2017-02-06
Implement missing authority API for HandlerRegistry

rmichela:
`HandlerRegistry.lookupMethod(String methodName, String authority)` is never never invoked with a non-null authority string. Per the discussion on the mailing list, I implemented this method.

https://groups.google.com/forum/#!topic/grpc-io/1RPka9CKAvE

### DIFF
--- a/core/src/main/java/io/grpc/internal/ServerImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerImpl.java
@@ -81,6 +81,8 @@ import javax.annotation.concurrent.GuardedBy;
  * server stops servicing new requests and waits for all connections to terminate.
  */
 public final class ServerImpl extends io.grpc.Server implements WithLogId {
+  public static final Attributes.Key<String> ATTR_AUTHORITY = Attributes.Key.of("io.grpc.authority");
+
   private static final ServerStreamListener NOOP_LISTENER = new NoopListener();
 
   private final LogId logId = LogId.allocate(getClass().getName());
@@ -418,7 +420,9 @@ public final class ServerImpl extends io.grpc.Server implements WithLogId {
             try {
               ServerMethodDefinition<?, ?> method = registry.lookupMethod(methodName);
               if (method == null) {
-                method = fallbackRegistry.lookupMethod(methodName);
+                String authority = stream.getAttributes()
+                                         .get(ATTR_AUTHORITY);
+                method = fallbackRegistry.lookupMethod(methodName, authority);
               }
               if (method == null) {
                 Status status = Status.UNIMPLEMENTED.withDescription(
--- a/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
@@ -45,9 +45,7 @@ import io.grpc.Attributes;
 import io.grpc.Grpc;
 import io.grpc.Metadata;
 import io.grpc.Status;
-import io.grpc.internal.GrpcUtil;
-import io.grpc.internal.ServerTransportListener;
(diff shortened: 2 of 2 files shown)


rmichela:
I'm not done with this (style, tests, etc) but am looking for feedback on the concept. Am I going in the right direction here?

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -45,9 +45,7 @@
 import io.grpc.Grpc;
 import io.grpc.Metadata;
 import io.grpc.Status;
-import io.grpc.internal.GrpcUtil;
-import io.grpc.internal.ServerTransportListener;
-import io.grpc.internal.StatsTraceContext;
+import io.grpc.internal.*;
```
carl-mastrangelo: We don't do `*` style imports.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -418,7 +420,9 @@ public void runInContext() {
             try {
               ServerMethodDefinition<?, ?> method = registry.lookupMethod(methodName);
               if (method == null) {
-                method = fallbackRegistry.lookupMethod(methodName);
+                String authority = stream.getAttributes()
```
carl-mastrangelo: `Attributes.newBuilder(Attributes)`

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -195,8 +193,13 @@ private void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
           checkNotNull(transportListener.methodDetermined(method, metadata), "statsTraceCtx");
       NettyServerStream.TransportState state = new NettyServerStream.TransportState(
           this, http2Stream, maxMessageSize, statsTraceCtx);
-      NettyServerStream stream = new NettyServerStream(ctx.channel(), state, attributes,
+      Attributes attributesWithAuthority = Attributes
+              .newBuilder()
```
carl-mastrangelo: This file is not particularly good about following the rule, but line continuation should be +4.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -195,8 +193,13 @@ private void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
...
+              .setAll(attributes).set(ServerImpl.ATTR_AUTHORITY, headers.authority().toString())
```
carl-mastrangelo: This is going to add extra allocation and AsciiString conversion in the serving path, where there was none before.  Almost no users need this data.  This would slow down all servers for the sake of proxies.

rmichela: Is there a cleaner way to get `authority` up to the `callbackRegistry` that doesn't involve passing `authority` along the call path?

core/src/main/java/io/grpc/internal/ServerImpl.java
rmichela: I don't understand this comment. Can you elaborate?

carl-mastrangelo: Oops, that should be in NettyServerHandler.java line 197

rmichela:
`HandlerRegistry` is still experimental. In light of #2235 do you see it's API moving away from exposing `authority`?

rmichela:
Per discussion with @ejona86 on IRC, I've re-attempted this PR using a different approach. 

New PR in #2709

PR #2705, opened 2017-02-07
all: swap to newer animalsniffer plugin

ejona86:
The new plugin uses a newer version of animalsniffer, allows overriding
the animalsniffer version used, and has up-to-date handling. The
up-to-date handling cuts fully incremental parallel build times in half,
from 5.5s to 2.7s.

The previous plugin was supposed to be verifying tests. However, either
it wasn't verifying them or its verification was broken.

### DIFF
--- a/netty/build.gradle
+++ b/netty/build.gradle
@@ -7,6 +7,7 @@ dependencies {
     // Tests depend on base class defined by core module.
     testCompile project(':grpc-core').sourceSets.test.output,
                 project(':grpc-testing')
+    signature "org.codehaus.mojo.signature:java17:+@signature"
 }
 
 test {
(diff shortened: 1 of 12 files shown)


netty/build.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
     // Tests depend on base class defined by core module.
     testCompile project(':grpc-core').sourceSets.test.output,
                 project(':grpc-testing')
+    signature "org.codehaus.mojo.signature:java17:+@signature"
```
carl-mastrangelo: 16?

ejona86: The Netty transport's TLS code doesn't actually work with JDK 6. There is a way to mark exclusions, but I didn't at this point.

PR #2709, opened 2017-02-08
Implement missing authority API for HandlerRegistry

rmichela:
`HandlerRegistry.lookupMethod(String methodName, String authority)` is never never invoked with a non-null authority string. Per the discussion on the mailing list, I implemented this method.

https://groups.google.com/forum/#!topic/grpc-io/1RPka9CKAvE

### DIFF
--- a/core/src/main/java/io/grpc/inprocess/InProcessTransport.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessTransport.java
@@ -71,6 +71,7 @@ class InProcessTransport implements ServerTransport, ConnectionClientTransport {
 
   private final LogId logId = LogId.allocate(getClass().getName());
   private final String name;
+  private final String authority;
   private ServerTransportListener serverTransportListener;
   private Attributes serverStreamAttributes;
   private ManagedClientTransport.Listener clientTransportListener;
@@ -82,8 +83,6 @@ class InProcessTransport implements ServerTransport, ConnectionClientTransport {
   private Status shutdownStatus;
   @GuardedBy("this")
   private Set<InProcessStream> streams = new HashSet<InProcessStream>();
-  @GuardedBy("this")
-  private String authority;
 
   public InProcessTransport(String name, String authority) {
     this.name = name;
@@ -141,7 +140,7 @@ class InProcessTransport implements ServerTransport, ConnectionClientTransport {
     }
     StatsTraceContext serverStatsTraceContext = serverTransportListener.methodDetermined(
         method.getFullMethodName(), headers);
-    return new InProcessStream(method, headers, serverStatsTraceContext).clientStream;
+    return new InProcessStream(method, headers, serverStatsTraceContext, authority).clientStream;
   }
 
   @Override
@@ -239,13 +238,15 @@ class InProcessTransport implements ServerTransport, ConnectionClientTransport {
     private final StatsTraceContext serverStatsTraceContext;
     private final Metadata headers;
     private final MethodDescriptor<?, ?> method;
+    private String authority;
 
     private InProcessStream(MethodDescriptor<?, ?> method, Metadata headers,
-        StatsTraceContext serverStatsTraceContext) {
+        StatsTraceContext serverStatsTraceContext, String authority) {
       this.method = checkNotNull(method, "method");
       this.headers = checkNotNull(headers, "headers");
       this.serverStatsTraceContext =
           checkNotNull(serverStatsTraceContext, "serverStatsTraceContext");
+      this.authority = authority;
     }
 
     // Can be called multiple times due to races on both client and server closing at same time.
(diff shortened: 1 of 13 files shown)


rmichela:
@ejona86 I've implemented this PR per your recommendations on IRC.

core/src/main/java/io/grpc/internal/AbstractServerStream.java
```diff
@@ -168,6 +168,11 @@ public final boolean isReady() {
   }
 
   @Override
+  public String getAuthority() {
+    return "";
```
ejona86: Given that `HandlerRegistry` is the only thing that is using the value from `getAuthority()` currently, and it is `@Nullable`, I would still feel better by defaulting to `null`. Here and in `NettyServerHandler`.

It's something we can change in the future though. We'll probably deal with it when we expose `getAuthority()` on `ServerCall`. At that point we may make it non-`null` where empty string is explicitly called out in the API.

rmichela:
I signed it!

core/src/main/java/io/grpc/inprocess/InProcessTransport.java
```diff
@@ -83,8 +84,9 @@
   @GuardedBy("this")
   private Set<InProcessStream> streams = new HashSet<InProcessStream>();
 
-  public InProcessTransport(String name) {
+  public InProcessTransport(String name, String authority) {
```
carl-mastrangelo: This would be API breaking.

core/src/main/java/io/grpc/inprocess/InProcessTransport.java
```diff
@@ -550,8 +559,7 @@ public void setMessageCompression(boolean enable) {}
 
       @Override
       public void setAuthority(String string) {
-        // TODO(ejona): Do something with this? Could be useful for testing, but can we "validate"
-        // it?
+        InProcessStream.this.authority = string;
```
carl-mastrangelo: This would not be thread safe

rmichela: `InProcessStream.this.authority` is now `volatile`. Setting and reading references are atomic in Java. Strings are immutable. This should be thread safe now, right?

rmichela:
@ejona86 @carl-mastrangelo Test cases added and feedback addressed.

rmichela:
I can't reproduce the `serverNotListening` test failure. I think it's a flapping test.

ejona86:
Jenkins, okay to test.

core/src/main/java/io/grpc/inprocess/InProcessTransport.java
ejona86: Actually, it was fine as it was before (without volatile). setAuthority "May only be called before start."

core/src/main/java/io/grpc/inprocess/InProcessTransport.java
ejona86: No, it isn't. InProcessTransport itself is package-private.

ejona86:
okay to test

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -205,6 +209,18 @@ private void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
...
   }
 
+  private String getOrUpdateAuthority(AsciiString authority) {
+    if (authority == null) {
+      return null;
+    } else if (!authority.equals(lastKnownAuthority)) {
+      lastKnownAuthority = authority;
+    }
+
+    // AsciiString.toString() is internally cached, so subsequent calls will not
```
carl-mastrangelo: This is not efficient enough.  If the client changes the authority every request and the server doesn't read from it, it will unnecessarily convert it.  Why not do the conversion when getAuthority is called?

ejona86: getAuthority() is always called.

This is plenty efficient enough. It doesn't hinder the common case at all.

Most clients, by far, won't change the authority every request. And even if they do change constantly, then any optimization here would probably be in the noise.

I'd rather get to a point we can call good enough for now and any future optimization for edge cases gets included with the HPACK tricks we're planning.

ejona86: I talked with Carl. He showed me that getAuthority() isn't always called, since it is only called when the immutable lookup fails. But we also agreed that it will 99% of the time always be the same authority. And we also believe that the HPACK table will even make it the same AsciiString instance. So Carl is now okay with this. He would still like to have our nice HPACK stuff, but that's orthogonal and future work.

ejona86:
Thank you for implementing this! I'm sorry that I didn't do it initially, but I'm glad you were able to.

PR #2712, opened 2017-02-08
core: DelayedStream getAttributes to throw IllegalStateExeption if not passThrough

dapengzhang0:


core/src/main/java/io/grpc/internal/DelayedStream.java
```diff
@@ -217,10 +217,8 @@ public void run() {
 
   @Override
   public Attributes getAttributes() {
-    if (passThrough) {
-      return realStream.getAttributes();
-    }
-    return Attributes.EMPTY;
+    checkState(passThrough, "passThrough");
```
ejona86: That string isn't helpful for users. How about "Called getAttributes before attributes are ready." Or similar.

core/src/test/java/io/grpc/internal/DelayedStreamTest.java
```diff
@@ -208,10 +205,14 @@ public void setStream_getAttributes() {
...
     stream.start(listener);
-    assertEquals(Attributes.EMPTY, stream.getAttributes());
 
-    stream.setStream(realStream);
-    assertEquals(attributes, stream.getAttributes());
+    try {
+      stream.getAttributes(); // expect to throw IllegalStateException, otherwise fail()
+      fail();
+    } catch (IllegalStateException e) {
+      stream.setStream(realStream);
```
ejona86: Move these two lines out of the catch.

PR #2718, opened 2017-02-10
netty: include SETTINGS_MAX_HEADER_LIST_SIZE in SETTINGS

### LINKED ISSUE #2350: Include SETTINGS_MAX_HEADER_LIST_SIZE in initial SETTINGS.
We currently don't send `SETTINGS_MAX_HEADER_LIST_SIZE` as part of the initial `SETTINGS` frame, although we enforce it on the decoder. This is valid as of the H2 spec

> For any given request, a lower limit than what is advertised MAY be enforced. The initial value of this setting is unlimited.

However, we would argue that we should still include it in the initial `SETTINGS` frame, cause why not? If both endpoints are gRPC Java, it would also give us better error messages, as then the local encoder would enforce the remote decode limit [1]. However, we can't currently do that cause it seems like a bug in Netty that the encoder throws a connection error and not a stream error [1], bringing down the whole connection, while only the stream should fail. I ll open a issue/PR in Netty.

[1] https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/internal/hpack/Encoder.java#L126

lukaszx0:
Closes #2350

cc @buchgr @ejona86

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
@@ -123,17 +123,19 @@ class NettyClientHandler extends AbstractNettyHandler {
     Http2Connection connection = new DefaultHttp2Connection(false);
 
     return newHandler(connection, frameReader, frameWriter, lifecycleManager, keepAliveManager,
-        flowControlWindow, ticker);
+        flowControlWindow, maxHeaderListSize, ticker);
   }
 
   @VisibleForTesting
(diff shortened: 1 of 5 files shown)


ejona86:
Tests are currently failing. It seems they are related.

lukaszx0:
@ejona86 ah, my bad. Thanks! Fixed.

carl-mastrangelo:
Tests passed, merging.  Also LGTM

PR #2723, opened 2017-02-13
all: Enable ErrorProne during compilation

ejona86:
ErrorProne provides static analysis for common issues, including
misused variables GuardedBy locks.

This increases build time by 60% for parallel builds and 30% for
non-parallel, so I've provided a way to disable the check. It is on by
default though and will be run in our CI environments.

So I've tried to do an okay job fixing up existing issues. But I wouldn't
be surprised if someone would prefer they were handled differently.

### DIFF
--- a/benchmarks/src/main/java/io/grpc/benchmarks/SocketAddressValidator.java
+++ b/benchmarks/src/main/java/io/grpc/benchmarks/SocketAddressValidator.java
@@ -31,12 +31,14 @@
 
 package io.grpc.benchmarks;
 
+import com.google.errorprone.annotations.Immutable;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 
 /**
  * Verifies whether or not the given {@link SocketAddress} is valid.
  */
+@Immutable
 public interface SocketAddressValidator {
   /**
    * Verifier for {@link InetSocketAddress}es.
--- a/build.gradle
+++ b/build.gradle
@@ -143,6 +143,7 @@ subprojects {
 
           [compileJava, compileTestJava].each() {
             // Protobuf-generated code produces some warnings.
+            // https://github.com/google/protobuf/issues/2718
             it.options.compilerArgs += ["-Xlint:-cast", "-Xep:MissingOverride:OFF",
                 "-Xep:ReferenceEquality:OFF", "-Xep:FunctionalInterfaceClash:OFF"]
           }
--- a/build.gradle
+++ b/build.gradle
@@ -2,10 +2,14 @@ buildscript {
   repositories {
     mavenCentral()
     mavenLocal()
+    maven {
+      url "https://plugins.gradle.org/m2/"
+    }
   }
   dependencies {
     classpath 'com.google.gradle:osdetector-gradle-plugin:1.4.0'
-    classpath "ru.vyarus:gradle-animalsniffer-plugin:1.2.0"
+    classpath 'ru.vyarus:gradle-animalsniffer-plugin:1.2.0'
+    classpath 'net.ltgt.gradle:gradle-errorprone-plugin:0.0.9'
   }
 }
 
@@ -20,6 +24,9 @@ subprojects {
     apply plugin: "com.google.osdetector"
     // The plugin only has an effect if a signature is specified
     apply plugin: "ru.vyarus.animalsniffer"
+    if (!rootProject.hasProperty('errorProne') || rootProject.errorProne.toBoolean()) {
+      apply plugin: "net.ltgt.errorprone"
+    }
 
     group = "io.grpc"
     version = "1.2.0-SNAPSHOT" // CURRENT_GRPC_VERSION
@@ -33,7 +40,7 @@ subprojects {
     }
 
     [compileJava, compileTestJava].each() {
-        it.options.compilerArgs += ["-Xlint:all", "-Xlint:-options"]
+        it.options.compilerArgs += ["-Xlint:all", "-Xlint:-options", "-Xlint:-path"]
         it.options.encoding = "UTF-8"
         if (rootProject.hasProperty('failOnWarnings') && rootProject.failOnWarnings.toBoolean()) {
             it.options.compilerArgs += ["-Werror"]
@@ -136,7 +143,8 @@ subprojects {
 
           [compileJava, compileTestJava].each() {
             // Protobuf-generated code produces some warnings.
-            it.options.compilerArgs += ["-Xlint:-cast"]
+            it.options.compilerArgs += ["-Xlint:-cast", "-Xep:MissingOverride:OFF",
+                "-Xep:ReferenceEquality:OFF", "-Xep:FunctionalInterfaceClash:OFF"]
           }
         }
 
@@ -198,6 +206,10 @@ subprojects {
 
         // Configuration for modules that use Netty tcnative (for OpenSSL).
         tcnative libraries.netty_tcnative
+
+        // The ErrorProne plugin defaults to the latest, which would break our
+        // build if error prone releases a new version with a new check
+        errorprone 'com.google.errorprone:error_prone_core:2.0.15'
     }
 
     signing {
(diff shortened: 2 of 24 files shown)


build.gradle
```diff
@@ -136,7 +143,8 @@ subprojects {
 
           [compileJava, compileTestJava].each() {
             // Protobuf-generated code produces some warnings.
-            it.options.compilerArgs += ["-Xlint:-cast"]
+            it.options.compilerArgs += ["-Xlint:-cast", "-Xep:MissingOverride:OFF",
+                "-Xep:ReferenceEquality:OFF", "-Xep:FunctionalInterfaceClash:OFF"]
```
carl-mastrangelo: Might be worth filing protobuf bugs

core/src/test/java/io/grpc/ConnectivityStateInfoTest.java
```diff
@@ -88,6 +88,7 @@ public void equality() {
     assertNotEquals(info4, info6);
 
     assertFalse(info1.equals(null));
-    assertFalse(info1.equals(this));
+    // Extra cast to avoid ErrorProne EqualsIncompatibleType failure
+    assertFalse(((Object) info1).equals(this));
```
carl-mastrangelo: assertNotEquals?

interop-testing/src/main/java/io/grpc/testing/integration/TestServiceImpl.java
```diff
@@ -447,6 +447,7 @@ private StreamingOutputCallResponse toResponse() {
   /**
    * Creates a buffer with data read from a file.
    */
+  @SuppressWarnings("Finally") // Not concerned about suppression; expected to be exceedingly rare
   private ByteString createBufferFromFile(String fileClassPath) {
```
carl-mastrangelo: Are we allowed java7 in test code?  That might let us use try-with safely and avoid the warning suppression

netty/src/test/java/io/grpc/netty/NettyStreamTestBase.java
```diff
@@ -199,7 +200,7 @@ public void shouldNotBeReadyForDataAfterWritingLargeMessage() throws IOException
   }
 
   protected byte[] smallMessage() {
-    return MESSAGE.getBytes();
+    return MESSAGE.getBytes(US_ASCII);
```
carl-mastrangelo: US_ASCII is deprecated I think

core/src/test/java/io/grpc/ConnectivityStateInfoTest.java
ejona86: Do you know which `equals()` method `assertNotEquals()` uses? It looks unspecified to me, so calling `equals()` manually would be the only safe way to do the test.

interop-testing/src/main/java/io/grpc/testing/integration/TestServiceImpl.java
ejona86: Currently, no, we can't do java7 in test code. This is actually something that became harder with the new animalsniffer plugin, but the old animalsniffer plugin was worrying me because it missed some things.

netty/src/test/java/io/grpc/netty/NettyStreamTestBase.java
ejona86: Versus what? The only other thing I see is StandardCharsets, which we can't use. I don't see anything lead me to believe that US_ASCII is any more deprecated than UTF_8.

build.gradle
dapengzhang0: filed one checkstyle bugs before at google/protobuf#2239

ejona86: I'm not certain protobuf should fix the warnings. Depending on your perspective it could be blamed on protobuf for not being perfect in errorprone's eyes, protobuf-gradle-plugin for mixing generated and non-generated code in the JavaCompile task, gradle for making it hard for protobuf-gradle-plugin to use a separate compilation task, or errorprone for not supporting filtering. However, if I view the problem from each project's perspective, it doesn't seem like their problem and it would be unlikely they would change.

ejona86: I've filed https://github.com/google/protobuf/issues/2718 I'll update the build.gradle to include a link.

interop-testing/src/test/java/io/grpc/testing/integration/ProxyTest.java
```diff
@@ -202,7 +202,7 @@ public void run() {
   }
 
   // server with echo and streaming modes
-  private class Server implements Runnable {
+  private static class Server implements Runnable {
```
ejona86: I don't think that relates at all to using errorprone or adding static here.

I actually prefer it the way it is. Yes, I could add final, but since it is a private class already I don't see that as doing much. Concerning the visibility, I tend to prefer visibility be "normal" inside the class and only control the class's visibility to choose whether to expose it to others. This makes them a bit easier to read and also easier to move. So I sort of prefer the way this is already done.

testing/src/main/java/io/grpc/testing/DeadlineSubject.java
```diff
@@ -92,7 +92,7 @@ public void of(Deadline expected) {
    * A partially specified proposition about an approximate relationship to a {@code deadline}
    * subject using a tolerance.
    */
-  public abstract class TolerantDeadlineComparison {
+  public abstract static class TolerantDeadlineComparison {
```
ejona86: I'm not trying to resolve FindBugs notices in this PR. And given that the equals and hashCode throw an exception unconditionally, it seems that it isn't too much of a concern.

The only reason I added static here is that if I don't the build will fail because of the errorprone checks.

benchmarks/src/main/java/io/grpc/benchmarks/SocketAddressValidator.java
```diff
@@ -31,12 +31,14 @@
...
 
+import com.google.errorprone.annotations.Immutable;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 
 /**
  * Verifies whether or not the given {@link SocketAddress} is valid.
  */
+@Immutable
 public interface SocketAddressValidator {
```
ejona86: The current code is fine and valid. Yes, we may want it to be a different way, but just adding the `@Immutable` is the least invasive and equivalent to swapping to an enum. I see rewriting code unnecessarily as counter-productive, because nobody will agree exactly how it should be implemented and so things can be rewritten again and again without bound.

ejona86:
@dapengzhang0, how do you feel about my replies to your comments?

dapengzhang0:
@ejona86 mostly still LGTM. I'm wondering what @carl-mastrangelo would say about `SocketAddressValidator`.

ejona86:
@carl-mastrangelo, how do you feel about the `SocketAddressValidator` change?

core/src/test/java/io/grpc/ConnectivityStateInfoTest.java
carl-mastrangelo: Not sure.  I think it is expected.equals(actual), but Truth seems to do it 'tother way 'round.  I'm not even sure what this line is trying to prove.  It seems checking with null would do the same thing, and avoid the cast.

Up to you.

netty/src/test/java/io/grpc/netty/NettyStreamTestBase.java
carl-mastrangelo: Like most other refs, a 

private static final US_ASCII = Charset.forName("US-ASCII);


is probably good enough.  Like `@Beta`, we probably shouldn't depend on `@Deprecated` stuff.

benchmarks/src/main/java/io/grpc/benchmarks/SocketAddressValidator.java
carl-mastrangelo: Pro enum: 
* Effective Java 19 says so, and it's usually right.
* It's probably slightly faster, cleaner, more readable.

Pro leave-this-as-is:
* It's in benchmarks, and HK?

I am ambivalent.  Since neither is objectively better than the other, I can go either way.  Right of way goes to the author so Eric will need to decide what's best long term for gRPC.  If pressed to a decision, I count myself in the clean-things-up-as-you-go camp and would pick an enum as well.

Issue #2189, opened 2016-08-20
metadata is lost when server sends StatusRuntimeException


skyao:
I would like to send a StatusRuntimeException to the client with some information saved in metadata. like this:

``` java
Metadata trailers = new Metadata();
 trailers.put(EXTENDED_STATUS_KEY, "10001");
StatusRuntimeException e = new StatusRuntimeException(Status.ABORTED, trailers);
 throw e;
```

But when I caught the exception in the client, I found that the metadata in the exception didn't contain this information. I checked the code and found that in class Server.JumpToApplicationThreadServerStreamListener:

``` java
    public void halfClosed() {
      callExecutor.execute(new ContextRunnable(context) {
        @Override
        public void runInContext() {
          try {
            getListener().halfClosed();
          } catch (RuntimeException e) {
            internalClose(Status.fromThrowable(e), new Metadata());
            throw e;
          } catch (Throwable t) {
            internalClose(Status.fromThrowable(t), new Metadata());
            throw new RuntimeException(t);
          }
        }
      });
    }
```

In this method, `Status.fromThrowable(e)` will convert this exception into Status instance, but it won't reuse the metadata saved in the exception. Instead, it create a new empty Metadata.

So in this way, all the data saved in the metadata of StatusRuntimeException is lost.

Yet in class Status I found there is a new method `Metadata trailersFromThrowable(Throwable t)` and it is used in class ServerCalls:

``` java
public void onError(Throwable t) {
      Metadata metadata = Status.trailersFromThrowable(t);
      if (metadata == null) {
        metadata = new Metadata();
      }
      call.close(Status.fromThrowable(t), metadata);
    }
```

I think this is the correct way to handle and pass metadata. 

Please help to confirm it. If so, I would like to pull a request to fix it. 

Thank you !

skyao:
Hi, I pulled a request to fix this issue. 

The code change is very small and straight.

ejona86:
Throwing an exception is an error and discouraged. The normal way to fail is to call `StreamObserver.onError()`. As you saw in `ServerCalls`, `onError()` behaves correctly.

I'm a bit concerned about `ServerInterceptor`s when there is thrown metadata, because interceptors generally won't realize they need to modify such metadata. And I don't really want to cause interceptors to need to modify thrown metadata. That seems likely to cause its own sort of breakage. :-/

[...]

PR #2730, opened 2017-02-16
Propagate trailers in runtime status exception

daicoden:
This propagates trailers to the client when server code throws a StatusRuntimeException.

Previously trailers were only propagated by invoking `responseListener.onError()`.  
This previous behavior was surprising behavior because the Status is passed back to the client when throwing exceptions, but the trailers were lost.

This will allow application code to communicate trailers without handing the responseListener through the stack.


ejona86:
See #2189

daicoden:
well shucks, thanks.

daicoden:
Would it be worth me inverting the tests so it was documented it was intentional?  It looked like a bug when investigating, and a test which set trailers in an exception, and verified they were cleared would have made it clear it was intentional.

PR #2736, opened 2017-02-17
services: add monitoring service

ericgribkoff:
This adds `monitoring.proto` and an implementation of its `GetCanonicalRpcStats` method in `MonitoringService.java`. 

There are tests for all of the new serialization methods from the various `instrumentation` types (`View`, `ViewDescriptor`, etc.) to protos.

There is no integration test for the service itself: the actual stats come from `StatsManagerImpl`, which hasn't been released yet. `MonitoringService#getInstance()` checks (via a call to the `instrumentation` API) for a `StatsManager` implementation, and throws an exception if its not found.

How this will be used, once `StatsManagerImpl` is included in an `instrumentation` release, is that user code can add a dependency on the instrumentation implementation and `MonitoringService` will pick it up at runtime. I've tested this locally with an stub version of `StatsManagerImpl`. I believe the mechanism is fairly straightforward, but let me know if it should be changed and/or a test added via a mock or some other means.


ericgribkoff:
PTAL

services/src/main/java/io/grpc/services/MonitoringService.java
```diff
@@ -0,0 +1,105 @@
...
+import io.grpc.stub.StreamObserver;
+
+/**
+ * An implementation of the gRPC monitoring service.
+ *
+ * <p>An implementation of {@link StatsManager} must be available at runtime (determined via {@link
+ * Stats#getStatsManager}) or instantiating this service will fail.
+ */
+public final class MonitoringService extends MonitoringGrpc.MonitoringImplBase {
+  private static final StatsManager STATS_MANAGER = Stats.getStatsManager();
```
zhangkun83: To make it more testable, I would make StatsManager an instance field, pass it to the constructor, call `Stats.getStatsManager()` in `getInstance()` which creates the singleton `instance`, make the constructor package-private, and inject a mock `StatsManager` in the unit test.

Then in `MonitoringServiceTest`, you can verify that `StatsManager.getView()` is called with the expected view descriptors, and that the service is filling out every field as expected (by stubbing `getView()` to return some fake results).

services/src/test/java/io/grpc/services/MonitoringUtilTest.java
```diff
@@ -0,0 +1,374 @@
...
+        intervalAggregationProto,
+        MonitoringUtil.serializeIntervalAggregation(intervalAggregation));
+  }
+
+  @Test
+  public void serializeInterval() throws Exception {
+    assertEquals(intervalProto, MonitoringUtil.serializeInterval(interval));
+  }
+
+  private int measurementUnitPower = -3;
```
zhangkun83: All fields that are not intended to be changed should be final.

services/src/main/java/io/grpc/services/MonitoringUtil.java
```diff
@@ -0,0 +1,272 @@
...
+import com.google.instrumentation.stats.ViewDescriptor.DistributionViewDescriptor;
+import com.google.instrumentation.stats.ViewDescriptor.IntervalViewDescriptor;
+import com.google.instrumentation.stats.proto.CensusProto;
+import io.grpc.instrumentation.v1alpha.CanonicalRpcStats;
+
+/** Utility methods to support {@link MonitoringService}. */
+final class MonitoringUtil {
+
+  /** Serialize a {@link View} and associated descriptors to a {@link CanonicalRpcStats.View}. */
+  public static CanonicalRpcStats.View buildCanonicalRpcStatsView(View view) {
```
ericgribkoff: Done

services/src/main/java/io/grpc/services/MonitoringService.java
ericgribkoff: Done

services/src/test/java/io/grpc/services/MonitoringUtilTest.java
ericgribkoff: Done

ericgribkoff:
I added the `@ExperimentalApi` annotation. Will create tracking issue when this is approved.

ericgribkoff:
retest this please

Issue #2737, opened 2017-02-17
Pass "args" object to LoadBalancer2.SubchannelPicker.pickSubchannel()


ejona86:
Right now the method looks like:
```java
public abstract PickResult pickSubchannel(Attributes affinity, Metadata headers);
```

Neither affinity nor headers feels inherent to the method. A bit "random." That is a symptom that we may need to add additional arguments in the future. For example, `MethodDescriptor`. To avoid overload troubles, we should probably just make it an object today. Say, something like:

```java
public abstract class PickSubchannelArgs {
  public Attributes getAffinity() { return Attributes.EMPTY; }
  public abstract Metadata getMetadata();

  // And an example of how we would add things in the future
  public CallOptions getCallOptions() {
    return CallOptions.DEFAULT;
  }
  public MethodDescriptor getMethodDescriptor() {
    return null;
    // or
    // throw new UnsupportedOperationException();
  }
}
```

zhangkun83:
A transition plan may look like

1. In the next release, old `SubchannelPicker` is still working, but will switch to implement the new signature: 
```java
abstract class SubchannelPicker {
  @Deprecated
  public PickResult pickSubchannel(Attributes affinity, Metadata headers) {
    return pickSubchannel(new PickSubchannelArgsImpl(affinity, headers));
  }
  public PickResult pickSubchannel(PickSubchannelArgs args) {
    return pickSubchannel(args.getAffinity(), args.getHeaders());
  }
}
```
2. In the subsequent release, delete the old signature

If we do this, it has to be done as a `1.1.x` patch release, because `1.2` is going to have LBv2 API promoted, which is a API-breaking change anyway.

/cc @lukaszx0 @jhspaybar

PR #2740, opened 2017-02-19
core: introduce PickSubchannelArgs

lukaszx0:
This will allow adding new arguments passed to pickSubchannel without significant API changes in core LB API in the future.

Addresses #2737

cc @ejona86 @zhangkun83

### DIFF
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -31,7 +31,6 @@
 
 package io.grpc;
 
-import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import java.util.List;
 import javax.annotation.Nullable;
@@ -163,7 +162,7 @@ public abstract class LoadBalancer {
    * The main balancing logic.  It <strong>must be thread-safe</strong>. Typically it should only
    * synchronize on its own state, and avoid synchronizing with the LoadBalancer's state.
    *
-   * Note: Implementations should override exactly one {@code pickSubchannel}.
+   * <p>Note: Implementations should override exactly one {@code pickSubchannel}.
    */
   @ThreadSafe
   public abstract static class SubchannelPicker {
@@ -173,7 +172,7 @@ public abstract class LoadBalancer {
      * @param affinity the affinity attributes provided via {@link CallOptions#withAffinity}
      * @param headers the headers container of the RPC. It can be mutated within this method.
      * @deprecated this signature is going to be removed in the next minor release. Implementations
-     * should instead override the {@link #pickSubchannel(PickSubchannelArgs)}.
+     *     should instead override the {@link #pickSubchannel(PickSubchannelArgs)}.
      */
     @Deprecated
     public PickResult pickSubchannel(Attributes affinity, Metadata headers) {
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -31,6 +31,7 @@
 
 package io.grpc;
 
+import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import java.util.List;
 import javax.annotation.Nullable;
@@ -169,8 +170,75 @@ public abstract class LoadBalancer {
      *
      * @param affinity the affinity attributes provided via {@link CallOptions#withAffinity}
      * @param headers the headers container of the RPC. It can be mutated within this method.
+     * @deprecated use {@link #pickSubchannel(PickSubchannelArgs)}
      */
-    public abstract PickResult pickSubchannel(Attributes affinity, Metadata headers);
+    @Deprecated
+    public PickResult pickSubchannel(Attributes affinity, Metadata headers) {
+      throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Make a balancing decision for a new RPC.
+     *
+     * @param args the pick arguments
+     */
+    // TODO(lukaszx0) make it abstract once deprecated overload will be removed.
+    public PickResult pickSubchannel(PickSubchannelArgs args) {
+      return pickSubchannel(args.getCallOptions().getAffinity(), args.getHeaders());
+    }
+  }
+
+  /**
+   * Provides arguments for a {@link SubchannelPicker#pickSubchannel(PickSubchannelArgs)}.
+   */
+  public abstract static class PickSubchannelArgs {
+
+    /**
+     * Call options.
+     */
+    public abstract CallOptions getCallOptions();
+
+    /**
+     * Call metadata.
+     */
+    public abstract Metadata getHeaders();
+
+    /**
+     * Call method.
+     */
+    public abstract MethodDescriptor<?, ?> getMethodDescriptor();
+
+    /**
+     * Returns true if the given object is also a {@link PickSubchannelArgs} with an equal
+     * getter values.
+     */
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+      PickSubchannelArgs that = (PickSubchannelArgs) o;
+      return Objects.equal(getCallOptions(), that.getCallOptions())
+          && Objects.equal(getHeaders(), that.getHeaders())
+          && Objects.equal(getMethodDescriptor(), that.getMethodDescriptor());
+    }
+
+    /**
+     * Returns a hash code for the subchannel args object.
+     */
+    @Override
+    public int hashCode() {
+      return Objects.hashCode(getCallOptions(), getHeaders(), getMethodDescriptor());
+    }
+
+    @Override
+    public String toString() {
+      return "[callOptions=" + getCallOptions() + " headers=" + getHeaders() + " method="
+          + getMethodDescriptor() + "]";
+    }
   }
 
   /**
@@ -340,8 +408,8 @@ public abstract class LoadBalancer {
      * Set a new picker to the channel.
      *
      * <p>When a new picker is provided via {@code updatePicker()}, the channel will apply the
-     * picker on all buffered RPCs, by calling {@link SubchannelPicker#pickSubchannel
-     * SubchannelPicker.pickSubchannel()}.
+     * picker on all buffered RPCs, by calling {@link SubchannelPicker#pickSubchannel(
+     * PickSubchannelArgs)}.
      *
      * <p>The channel will hold the picker and use it for all RPCs, until {@code updatePicker()} is
      * called again and a new picker replaces the old one.  If {@code updatePicker()} has never been
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -31,6 +31,7 @@
 
 package io.grpc;
 
+import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import java.util.List;
 import javax.annotation.Nullable;
@@ -215,7 +216,7 @@ public abstract class LoadBalancer {
      * getter values.
      */
     @Override
-    public boolean equals(Object o) {
+    public final boolean equals(Object o) {
       if (this == o) {
         return true;
       }
@@ -232,7 +233,7 @@ public abstract class LoadBalancer {
      * Returns a hash code for the subchannel args object.
      */
     @Override
-    public int hashCode() {
+    public final int hashCode() {
       return Objects.hashCode(getCallOptions(), getHeaders(), getMethodDescriptor());
     }
   }
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -210,32 +210,6 @@ public abstract class LoadBalancer {
      * Call method.
      */
     public abstract MethodDescriptor<?, ?> getMethodDescriptor();
-
-    /**
-     * Returns true if the given object is also a {@link PickSubchannelArgs} with an equal
-     * getter values.
-     */
-    @Override
-    public final boolean equals(Object o) {
-      if (this == o) {
-        return true;
-      }
-      if (o == null || getClass() != o.getClass()) {
-        return false;
-      }
-      PickSubchannelArgs that = (PickSubchannelArgs) o;
-      return Objects.equal(getCallOptions(), that.getCallOptions())
-          && Objects.equal(getHeaders(), that.getHeaders())
-          && Objects.equal(getMethodDescriptor(), that.getMethodDescriptor());
-    }
-
-    /**
-     * Returns a hash code for the subchannel args object.
-     */
-    @Override
-    public final int hashCode() {
-      return Objects.hashCode(getCallOptions(), getHeaders(), getMethodDescriptor());
-    }
   }
 
   /**
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -233,12 +233,6 @@ public abstract class LoadBalancer {
     public int hashCode() {
       return Objects.hashCode(getCallOptions(), getHeaders(), getMethodDescriptor());
     }
-
-    @Override
-    public String toString() {
-      return "[callOptions=" + getCallOptions() + " headers=" + getHeaders() + " method="
-          + getMethodDescriptor() + "]";
-    }
   }
 
   /**
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -31,7 +31,6 @@
 
 package io.grpc;
 
-import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import java.util.List;
 import javax.annotation.Nullable;
@@ -162,6 +161,8 @@ public abstract class LoadBalancer {
   /**
    * The main balancing logic.  It <strong>must be thread-safe</strong>. Typically it should only
    * synchronize on its own state, and avoid synchronizing with the LoadBalancer's state.
+   *
+   * Note: Implementations should override exactly one {@code pickSubchannel}.
    */
   @ThreadSafe
   public abstract static class SubchannelPicker {
@@ -170,7 +171,8 @@ public abstract class LoadBalancer {
      *
      * @param affinity the affinity attributes provided via {@link CallOptions#withAffinity}
      * @param headers the headers container of the RPC. It can be mutated within this method.
-     * @deprecated use {@link #pickSubchannel(PickSubchannelArgs)}
+     * @deprecated this signature is going to be removed in the next minor release. Implementations
+     * should instead override the {@link #pickSubchannel(PickSubchannelArgs)}.
      */
     @Deprecated
     public PickResult pickSubchannel(Attributes affinity, Metadata headers) {
--- a/core/src/main/java/io/grpc/internal/DelayedClientTransport.java
+++ b/core/src/main/java/io/grpc/internal/DelayedClientTransport.java
@@ -34,7 +34,9 @@ package io.grpc.internal;
 import com.google.common.annotations.VisibleForTesting;
 import io.grpc.CallOptions;
 import io.grpc.Context;
+import io.grpc.LoadBalancer;
 import io.grpc.LoadBalancer.PickResult;
+import io.grpc.LoadBalancer.PickSubchannelArgs;
 import io.grpc.LoadBalancer.SubchannelPicker;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
@@ -143,11 +145,12 @@ final class DelayedClientTransport implements ManagedClientTransport {
       CallOptions callOptions, StatsTraceContext statsTraceCtx) {
     try {
       SubchannelPicker picker = null;
+      PickSubchannelArgs args = new PickSubchannelArgsImpl(method, headers, callOptions);
       long pickerVersion = -1;
       synchronized (lock) {
         if (!shutdown) {
           if (lastPicker == null) {
-            return createPendingStream(method, headers, callOptions, statsTraceCtx);
+            return createPendingStream(args, statsTraceCtx);
           }
           picker = lastPicker;
           pickerVersion = lastPickerVersion;
@@ -155,12 +158,12 @@ final class DelayedClientTransport implements ManagedClientTransport {
       }
       if (picker != null) {
         while (true) {
-          PickResult pickResult = picker.pickSubchannel(
-              new PickSubchannelArgsImpl(method, headers, callOptions));
-          ClientTransport transport = GrpcUtil.getTransportFromPickResult(
-              pickResult, callOptions.isWaitForReady());
+          PickResult pickResult = picker.pickSubchannel(args);
+          ClientTransport transport = GrpcUtil.getTransportFromPickResult(pickResult,
+              callOptions.isWaitForReady());
           if (transport != null) {
-            return transport.newStream(method, headers, callOptions, statsTraceCtx);
+            return transport.newStream(args.getMethodDescriptor(), args.getHeaders(),
+                args.getCallOptions(), statsTraceCtx);
           }
           // This picker's conclusion is "buffer".  If there hasn't been a newer picker set
           // (possible race with reprocess()), we will buffer it.  Otherwise, will try with the new
@@ -170,7 +173,7 @@ final class DelayedClientTransport implements ManagedClientTransport {
               break;
             }
             if (pickerVersion == lastPickerVersion) {
-              return createPendingStream(method, headers, callOptions, statsTraceCtx);
+              return createPendingStream(args, statsTraceCtx);
             }
             picker = lastPicker;
             pickerVersion = lastPickerVersion;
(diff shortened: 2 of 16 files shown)


core/src/main/java/io/grpc/LoadBalancer2.java
```diff
@@ -317,6 +322,77 @@ public String toString() {
   }
 
   /**
+   * Provides arguments for a {@link SubchannelPicker#pickSubchannel(PickSubchannelArgs)}.
+   */
+  public abstract static class PickSubchannelArgs {
```
zhangkun83: Since LB impls will provide this object, we should make it convenient to create, probably a final class with a builder.

lukaszx0:
@zhangkun83 regarding `methodDescriptor`, I'd argue that it's very useful to have it and there's already a lot of use cases for it, first that immediately comes to mind - ability to have code recording metrics inside picker, based on method's name. I think we should provide it, it'll be very helpful for implementors, myself included. Don't have very strong opinion, can remove it if you're not convinced.

Addressed other comments, PTAL.

core/src/main/java/io/grpc/LoadBalancer2.java
```diff
@@ -170,13 +171,17 @@ public abstract void handleSubchannelState(
...
+
     /**
      * Make a balancing decision for a new RPC.
      *
-     * @param affinity the affinity attributes provided via {@link CallOptions#withAffinity}
-     * @param headers the headers container of the RPC. It can be mutated within this method.
+     * @param args the pick arguments
      */
-    public abstract PickResult pickSubchannel(Attributes affinity, Metadata headers);
+    public abstract PickResult pickSubchannel(PickSubchannelArgs args);
```
lukaszx0: AH yeah, good point. Thanks!

core/src/main/java/io/grpc/LoadBalancer2.java
lukaszx0: Good call.

core/src/main/java/io/grpc/LoadBalancer2.java
```diff
@@ -317,6 +322,77 @@ public String toString() {
...
+     * Call options.
+     */
+    public CallOptions getCallOptions() {
+      return CallOptions.DEFAULT;
+    }
+
+    /**
+     * Call metadata.
+     */
+    public abstract Metadata getMetadata();
```
lukaszx0: 👍

core/src/main/java/io/grpc/LoadBalancer2.java
zhangkun83: Oh sorry, I must have gone out of my mind. The args are *consumed* by LB impls, so an abstract class as described in #2737 is most appropriate.

lukaszx0: I'm confused. The final immutable class still makes sense to me in this context, no? Do you anticipate passing different types of `PickSubchannelArgs` to `pickSubchannel` internally? If so when and why?

lukaszx0: I guess the only change needed would be to either add interface in `LoadBalancer2` or promote `PickSubchannelArgs` from internal to `io.grpc`. Probably latter, since this will let implementors re-use it in tests.

zhangkun83: `PickSubchannelArgs` is public API, thus it should be in `io.grpc` (preferably as inner of `LoadBalancer2`) instead of `io.grpc.internal`.

Builder pattern exposes the construction of `PickSubchannelArgs` as public API, which is not needed by LB impls because they are only consumers of `PickSubchannelArgs`. The builder API has to define required fields (headers) and enforce it, but there isn't an ideal way to do it (there was a [previous discussion](https://github.com/grpc/grpc-java/pull/2667#discussion_r98262481) on a similar issue). It's better to avoid exposing such a builder unless it's necessary.

lukaszx0: How would you unit test your custom impl of `Picker` without access to builder? Exposing only abstract class would require you to implement it which effectively will be just copy of what's used internally because this class is just a simple container.

Anyway, unclear at this point if you think we should revert it back to `abstract` or make it `public final` (as it is now) ?

zhangkun83: I am thinking about reverting back to `abstract`. The channel would have its own private implementation which will not have a builder, but just a constructor. Unit tests would extend the abstract class to make fake values. Unfortunately Mockito would be discouraged because any unstubbed methods (e.g., new fields) will return null.

lukaszx0: That puts a bit of a burden on implementors when writing tests. Not a strong opinion but I slightly favour builder though I understand your point of view and I'm fine reverting that back. Let me know what do you think works better for you.

zhangkun83: I agree with you it's a bit inconvenient for test writers. Because it appears to me both approaches have about the same amount of problems, I would opt for a simpler API, which is the abstract class.

zhangkun83: And then the class would be small enough to fit in `LoadBalancer2`.

lukaszx0: Reverted back to abstract. PTAL.

lukaszx0:
> You may want to wait until I merge #2743 before making further changes.

It's OK, I've addressed feedback and will rebase after your migration lands. Thanks!

core/src/main/java/io/grpc/internal/PickSubchannelArgsImpl.java
```diff
@@ -0,0 +1,102 @@
...
+        && Objects.equal(method, that.method);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(callOptions, headers, method);
+  }
+
+  @Override
+  public String toString() {
```
lukaszx0: Done.

core/src/main/java/io/grpc/internal/PickSubchannelArgsImpl.java
```diff
@@ -0,0 +1,102 @@
...
+    PickSubchannelArgsImpl that = (PickSubchannelArgsImpl) o;
+    return Objects.equal(callOptions, that.callOptions)
+        && Objects.equal(headers, that.headers)
+        && Objects.equal(method, that.method);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(callOptions, headers, method);
+  }
```
lukaszx0: Good call, though it convinces me a bit that maybe using builder in this case would've been more convenient

core/src/main/java/io/grpc/LoadBalancer2.java
```diff
@@ -317,6 +325,77 @@ public String toString() {
...
 
   /**
+   * Provides arguments for a {@link SubchannelPicker#pickSubchannel(PickSubchannelArgs)}.
+   */
+  public abstract static class PickSubchannelArgs {
+
+    /**
+     * Call affinity.
+     */
+    public Attributes getAffinity() {
```
lukaszx0: Done.

core/src/main/java/io/grpc/LoadBalancer2.java
```diff
@@ -170,13 +171,20 @@ public abstract void handleSubchannelState(
...
+
     /**
      * Make a balancing decision for a new RPC.
      *
-     * @param affinity the affinity attributes provided via {@link CallOptions#withAffinity}
-     * @param headers the headers container of the RPC. It can be mutated within this method.
+     * @param args the pick arguments
      */
-    public abstract PickResult pickSubchannel(Attributes affinity, Metadata headers);
+    // TODO(lukasz) make it public once deprecated overload will be removed.
```
lukaszx0: Doh, I meant _abstract_ 🤦‍♂️

core/src/main/java/io/grpc/LoadBalancer2.java
```diff
@@ -170,13 +171,20 @@ public abstract void handleSubchannelState(
    */
   @ThreadSafe
   public abstract static class SubchannelPicker {
+    @Deprecated
+    public PickResult pickSubchannel(Attributes affinity, Metadata headers) {
+      return pickSubchannel(new DeprecatedPickSubchannelArgs(affinity, headers));
```
lukaszx0: Ah, good point. Removed.

core/src/main/java/io/grpc/LoadBalancer2.java
```diff
@@ -170,13 +171,20 @@ public abstract void handleSubchannelState(
    */
   @ThreadSafe
   public abstract static class SubchannelPicker {
+    @Deprecated
```
lukaszx0: Done.

core/src/main/java/io/grpc/internal/PickSubchannelArgsImpl.java
```diff
@@ -0,0 +1,70 @@
...
+package io.grpc.internal;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import io.grpc.CallOptions;
+import io.grpc.LoadBalancer2.PickSubchannelArgs;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+
+// TODO(lukaszx0) make this class pkg private after removing deprecated API in PickSubchannelArgs
```
zhangkun83: It is already so.

core/src/test/java/io/grpc/PickFirstLoadBalancer2Test.java
```diff
@@ -71,6 +72,7 @@
 
   private static final Attributes.Key<String> FOO = Attributes.Key.of("foo");
   private Attributes affinity = Attributes.newBuilder().set(FOO, "bar").build();
```
zhangkun83: Remove this?

core/src/test/java/io/grpc/PickFirstLoadBalancer2Test.java
```diff
@@ -71,6 +72,7 @@
...
+  private static final PickSubchannelArgs EMPTY_ARGS = mock(PickSubchannelArgs.class);
```
zhangkun83: Only if you verify at the end of the test that `EMPTY_ARGS` never had an interaction, it's fine to mock the args without any stubbing.

core/src/test/java/io/grpc/util/RoundRobinLoadBalancer2Test.java
```diff
@@ -87,6 +87,7 @@
   private Map<EquivalentAddressGroup, Subchannel> subchannels = Maps.newLinkedHashMap();
   private static final Attributes.Key<String> MAJOR_KEY = Attributes.Key.of("major-key");
   private Attributes affinity = Attributes.newBuilder().set(MAJOR_KEY, "I got the keys").build();
+  private static final PickSubchannelArgs EMPTY_ARGS = mock(PickSubchannelArgs.class);
```
zhangkun83: dittos with `PickFirstBalancerFactory2`

grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancer2Test.java
```diff
@@ -157,7 +157,7 @@ public String parse(InputStream stream) {
   private ArgumentCaptor<SubchannelPicker> pickerCaptor;
   private final SerializingExecutor channelExecutor =
       new SerializingExecutor(MoreExecutors.directExecutor());
-  private final Metadata headers = new Metadata();
+  private static final PickSubchannelArgs EMPTY_ARGS = mock(PickSubchannelArgs.class);
```
zhangkun83: dittos with PickFirstBalancerFactory2

core/src/main/java/io/grpc/internal/PickSubchannelArgsImpl.java
lukaszx0: Ooops, leftover. Thanks!

core/src/test/java/io/grpc/PickFirstLoadBalancer2Test.java
lukaszx0: It is still used for `handleResolvedAddresses` calls.

core/src/test/java/io/grpc/PickFirstLoadBalancer2Test.java
lukaszx0: Added `tearDown` with `verifyNoMoreInteractions`

core/src/test/java/io/grpc/util/RoundRobinLoadBalancer2Test.java
lukaszx0: Added `tearDown` with `verifyNoMoreInteractions`

grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancer2Test.java
lukaszx0: Added `tearDown` with `verifyNoMoreInteractions`

lukaszx0:
I didn't realize that LBv2 transition will create such bad conflicts so I pretty much just did it over again - it was just easier to do so some copy-pasting from the old version.

@zhangkun83 PTAL and review again as a whole. Thanks!

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -318,7 +318,7 @@ void exitIdleMode() {
 
   // Must be run from channelExecutor
   private void cancelIdleTimer() {
-    if (idleModeTimerFuture != null) {
+    if (idleModeTimerFuture != null && idleModeTimer != null) {
```
zhangkun83: Please don't change this line.
 1. This is not related to the subject of this PR.
 2. `idleModeTimer` should be synchronized with `idleModeTimerFuture` in terms of nullness. If one is null while the other is not, it must be a bug and we want to let it throw instead of hiding it.

core/src/main/java/io/grpc/internal/PickSubchannelArgsImpl.java
```diff
@@ -0,0 +1,69 @@
...
+package io.grpc.internal;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import io.grpc.CallOptions;
+import io.grpc.LoadBalancer.PickSubchannelArgs;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+
+class PickSubchannelArgsImpl extends PickSubchannelArgs {
```
zhangkun83: final

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -169,8 +170,75 @@ public abstract void handleSubchannelState(
...
+     * Call method.
+     */
+    public abstract MethodDescriptor<?, ?> getMethodDescriptor();
+
+    /**
+     * Returns true if the given object is also a {@link PickSubchannelArgs} with an equal
+     * getter values.
+     */
+    @Override
+    public boolean equals(Object o) {
```
zhangkun83: final

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -169,8 +170,75 @@ public abstract void handleSubchannelState(
...
+      return Objects.equal(getCallOptions(), that.getCallOptions())
+          && Objects.equal(getHeaders(), that.getHeaders())
+          && Objects.equal(getMethodDescriptor(), that.getMethodDescriptor());
+    }
+
+    /**
+     * Returns a hash code for the subchannel args object.
+     */
+    @Override
+    public int hashCode() {
```
zhangkun83: final

core/src/test/java/io/grpc/PickFirstLoadBalancerTest.java
```diff
@@ -71,6 +73,7 @@
 
   private static final Attributes.Key<String> FOO = Attributes.Key.of("foo");
   private Attributes affinity = Attributes.newBuilder().set(FOO, "bar").build();
+  private static final PickSubchannelArgs EMPTY_ARGS = mock(PickSubchannelArgs.class);
```
zhangkun83: We don't want to set an example for users and let them think they can mock `PickSubchannelArgs` whenever we want. We'd better put a comment here says because this LoadBalancer doesn't use any of the arg fields, as verified in `tearDown()`, it's safe to use a mock without any stubbing. `unusedArgs` seems to be a more accurate name. I would also make it non-static, so that if any tests violates this precondition, we will know exactly which they are.

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
```diff
@@ -87,6 +88,7 @@
   private Map<EquivalentAddressGroup, Subchannel> subchannels = Maps.newLinkedHashMap();
   private static final Attributes.Key<String> MAJOR_KEY = Attributes.Key.of("major-key");
   private Attributes affinity = Attributes.newBuilder().set(MAJOR_KEY, "I got the keys").build();
+  private static final PickSubchannelArgs EMPTY_ARGS = mock(PickSubchannelArgs.class);
```
zhangkun83: ditto as in PickFirstLoadBalancerTest

grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancerTest.java
```diff
@@ -157,7 +157,7 @@ public String parse(InputStream stream) {
   private ArgumentCaptor<SubchannelPicker> pickerCaptor;
   private final SerializingExecutor channelExecutor =
       new SerializingExecutor(MoreExecutors.directExecutor());
-  private final Metadata headers = new Metadata();
+  private static final PickSubchannelArgs EMPTY_ARGS = mock(PickSubchannelArgs.class);
```
zhangkun83: ditto as in PickFirstLoadBalancerTest

core/src/main/java/io/grpc/LoadBalancer.java
ejona86: Actually, let's not have equals/hashCode. If you're comparing them, Your Doing It Wrong™. That also prevents having the problem later where it is hard to figure out what to consider "equal".

core/src/main/java/io/grpc/internal/PickSubchannelArgsImpl.java
ejona86: Ewww... Um, the throwing problem sort of makes me think we shouldn't implement `toString()` in the abstract class either. Granted, we could call a method within try-catch in order to handle an exception, but it is still pretty icky.

The only reason this is an abstract class instead of interface is to allow adding new methods in the future. So in that way, I don't mind requiring toString to be implemented in each extending class.

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
```diff
@@ -155,7 +155,8 @@ public final ClientStream newStream(MethodDescriptor<?, ?> method, Metadata head
       }
       if (picker != null) {
         while (true) {
-          PickResult pickResult = picker.pickSubchannel(callOptions.getAffinity(), headers);
+          PickResult pickResult = picker.pickSubchannel(
+              new PickSubchannelArgsImpl(method, headers, callOptions));
```
lukaszx0: Tried to do a quick refactor, some tests were failing. Is it ok to address this in a follow up PR?

core/src/main/java/io/grpc/LoadBalancer.java
lukaszx0: Done

core/src/main/java/io/grpc/LoadBalancer.java
lukaszx0: Done

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -169,8 +170,75 @@ public abstract void handleSubchannelState(
      *
      * @param affinity the affinity attributes provided via {@link CallOptions#withAffinity}
      * @param headers the headers container of the RPC. It can be mutated within this method.
+     * @deprecated use {@link #pickSubchannel(PickSubchannelArgs)}
```
lukaszx0: Good call. Added javadocs.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
lukaszx0: Damn! I didn't intend to include this, I was experimenting with some tooling and this slipped somehow. Sorry, reverted.

core/src/main/java/io/grpc/internal/PickSubchannelArgsImpl.java
lukaszx0: Done.

core/src/test/java/io/grpc/PickFirstLoadBalancerTest.java
lukaszx0: Good points. Renamed it to `mockArgs`, which stick nicely with other fields and the convention we have.

lukaszx0:
Since comments on diff collapsed, moving the unaddressed one from @ejona86 up for discussion here:

> Actually, let's not have equals/hashCode. If you're comparing them, Your Doing It Wrong™. That also prevents having the problem later where it is hard to figure out what to consider "equal".

I agree on the principle, but turns our our tests are relying on this (as will probably the ones implemented by users).

@ejona86 @zhangkun83 thoughts?

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
zhangkun83: Hmm, I didn't expect this to be more than a trivial change. What tests broke?

ejona86:
> but turns our our tests are relying on this (as will probably the ones implemented by users)

Then the concrete implementation of PickSubchannelArgs used in the tests could implement the methods. The concrete implementation can easily avoid exceptions and API oddity.

lukaszx0:
> Then the concrete implementation of PickSubchannelArgs used in the tests could implement the methods. The concrete implementation can easily avoid exceptions and API oddity.

Then it's fine to have them in `PickSubchannelArgsImpl` too, right?

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
lukaszx0: It's possible that it is and I've made some stupid mistake, the `DelayedClientTransportTest` and one other tests were failing. I'll look into this again later today.

lukaszx0:
@ejona86 @zhangkun83 PTAL

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
```diff
@@ -347,26 +348,25 @@ public LogId getLogId() {
...
-      this.method = method;
-      this.headers = headers;
-      this.callOptions = callOptions;
-      this.context = Context.current();
+    private PendingStream(PickSubchannelArgs args, StatsTraceContext statsTraceCtx) {
+      this.args = args;
       this.statsTraceCtx = statsTraceCtx;
     }
 
+    public PickSubchannelArgs getArgs() {
```
zhangkun83: This method is not necessary, because `args` is only used from the same file.

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
```diff
@@ -290,24 +291,24 @@ final void reprocess(SubchannelPicker picker) {
...
-            @Override
-            public void run() {
-              stream.createRealStream(transport);
-            }
-          });
+          @Override
+          public void run() {
+            stream.createRealStream(transport);
+          }
+        });
```
zhangkun83: Please don't do any no-op reformatting.

core/src/test/java/io/grpc/internal/DelayedClientTransportTest.java
```diff
@@ -479,89 +501,92 @@ public void reprocess_newStreamRacesWithReprocess() throws Exception {
...
-          // Will call pickSubchannel and wait on barrier
-          delayedTransport.newStream(method, headers, callOptions, statsTraceCtx);
-        }
-      };
+      @Override
+      public void run() {
+        // Will call pickSubchannel and wait on barrier
+        delayedTransport.newStream(method, headers, callOptions, statsTraceCtx);
+      }
+    };
```
zhangkun83: ditto -- please refrain from any no-op reformatting.

core/src/test/java/io/grpc/internal/DelayedClientTransportTest.java
```diff
@@ -479,89 +501,92 @@ public void reprocess_newStreamRacesWithReprocess() throws Exception {
...
-          // Will call pickSubchannel and wait on barrier
-          delayedTransport.newStream(method, headers2, callOptions, statsTraceCtx);
-        }
-      };
+      @Override
+      public void run() {
+        // Will call pickSubchannel and wait on barrier
+        delayedTransport.newStream(method, headers2, callOptions, statsTraceCtx);
+      }
+    };
```
zhangkun83: ditto

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
lukaszx0: Right, fixed.

lukaszx0:
@zhangkun83 reverted formatting that you pointed out and found it in few more places - sorry IJ did that to me ;) Should be good now, PTAL.

ejona86:
I did some spot-checking, and this looks fine to me.

> Then it's fine to have them in `PickSubchannelArgsImpl` too, right?

Ideally, I'd say no, since applications could depend on the behavior. I think @zhangkun83 and I agree on that. However, some of the tests look very painful to fix if equals was dropped. So while disappointed, I'm okay with having equals/hashcode on the primary Impl because doing otherwise would be too onerous.

lukaszx0:
@ejona86 thanks, sounds good! @zhangkun83 are we good to go with this?

zhangkun83:
Please resolve the conflict.

lukaszx0:
Oops, didn't notice there was a conflict. Fixed!

Issue #2656, opened 2017-01-25
Switch to LBv2


zhangkun83:
LBv2 ([design doc](https://docs.google.com/document/d/1exE0_QNwbHJesD-Opl2D6qX1wujADkUm5OpFykks3C8)) consists of `LoadBalancer2` and its supporting channel implementation `ManagedChannelImpl2`. Here we outline the transition path from the current `LoadBalancer` and `ManagedChannelImpl` (a.k.a LBv1) to LBv2.

The transition is transparent to users who **only** use the default or stock LoadBalancers that are shipped with gRPC. The following transition plan should only concern advanced users who implement or use custom `LoadBalanacer`(s).

Also note all changes happen on master first. The changes scheduled for gRPC 1.2 will happen on master soon after 1.1 is released.

### During gRPC 1.1
LBv1 and LBv2 coexist. LBv1 is the default one. Both can be set to the channel builder, while LBv2 is used only if a `LoadBalancer2` implementation is set to the channel builder.

During this period, implementors re-write their LBs with the v2 API, and migrate their customers to their v2 LBs. 

For example, `FooBalancerFactory` is the current LB, while `Foo2BalancerFactory` is the new one that uses LBv2.
```java
class FooBalancerFactory extends LoadBalancer.Factory {
}

class Foo2BalancerFactory extends LoadBalancer2.Factory {
}
```
While the v1 version continues working:
```java
NettyChannelBuilder.forTarget(...).loadBalancerFactory(new FooBalancerFactory()).build();
```
Customers should be migrated to use the v2 version:
```java
NettyChannelBuilder.forTarget(...).loadBalancerFactory(new Foo2BalancerFactory()).build();
```

### During gRPC 1.2
We replace `LoadBalancer` with `LoadBalancer2`, and keep `LoadBalancer2` as an alias to `LoadBalancer` by extending it. LBv1 interfaces and codepath are deleted. All v1-based LBs stop working with gRPC. 

Prior to switching to 1.2, the implementors must have switched all their customers to the v2-based LoadBalancer2. During 1.2, they should rebase their v2-based implementations to the proper `LoadBalancer` interface.

For example, this continues working because `LoadBalancer2` extends `LoadBalancer`:
```java
class Foo2BalancerFactory extends LoadBalancer2.Factory {
}
```
.., but it should be updated to this to prevent breakage with gRPC 1.3:
```java
class Foo2BalancerFactory extends LoadBalancer.Factory {
}
```

### gRPC 1.3
[...]

PR #2743, opened 2017-02-22
core: delete LBv1 and promote LBv2

zhangkun83:
As planned out in #2656, this change will appear in gRPC 1.2

Please review the two commits individually.


zhangkun83:
Contrary to what I said in today's meeting, I am not going to rename the old classes to `*1` and wait one more week after it's imported before deleting them, because I would like to have them deleted before we cut the 1.2 branch.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -160,21 +160,6 @@
    * <p>If this method is not called, the builder will use {@link PickFirstBalancerFactory}
    * for the channel.
    *
-   * <p>This overrides what was set by {@link #loadBalancerFactory(LoadBalancer2.Factory)}.
-   *
-   * <p>Calling this will revert the channel back to the original (LBv1) implementation.
-   */
-  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1771")
-  public abstract T loadBalancerFactory(LoadBalancer.Factory loadBalancerFactory);
```
carl-mastrangelo: this has to remain abstract, since changedetector tests internally will complain about it.

core/src/main/java/io/grpc/PickFirstBalancerFactory.java
```diff
@@ -46,138 +46,122 @@
...
-    @GuardedBy("lock")
-    private InterimTransport<T> interimTransport;
-    @GuardedBy("lock")
-    private Status nameResolutionError;
-    @GuardedBy("lock")
-    private boolean closed;
-
-    private final TransportManager<T> tm;
+  @VisibleForTesting
+  static class PickFirstBalancer extends LoadBalancer {
```
carl-mastrangelo: final

core/src/main/java/io/grpc/PickFirstBalancerFactory.java
```diff
@@ -46,138 +46,122 @@
...
+    private final Helper helper;
+    private Subchannel subchannel;
 
-    private PickFirstBalancer(TransportManager<T> tm) {
-      this.tm = tm;
+    public PickFirstBalancer(Helper helper) {
+      this.helper = helper;
```
carl-mastrangelo: checkNotNull for public methods.  Also, reduce visibility to match class visibility

core/src/main/java/io/grpc/PickFirstBalancerFactory.java
```diff
@@ -46,138 +46,122 @@
...
-        addressesCopy = addresses;
-        if (addressesCopy == null) {
-          if (nameResolutionError != null) {
-            return tm.createFailingTransport(nameResolutionError);
-          }
-          if (interimTransport == null) {
-            interimTransport = tm.createInterimTransport();
-          }
-          return interimTransport.transport();
+    public void handleResolvedAddresses(List<ResolvedServerInfoGroup> servers,
```
carl-mastrangelo: Line breaks happen at highest semantic level.  Params should be on the same line.

core/src/main/java/io/grpc/PickFirstBalancerFactory.java
```diff
@@ -46,138 +46,122 @@
...
+          break;
+        case READY:
+        case IDLE:
+          pickResult = PickResult.withSubchannel(subchannel);
+          break;
+        case TRANSIENT_FAILURE:
+          pickResult = PickResult.withError(stateInfo.getStatus());
+          break;
+        default:
+          throw new IllegalStateException();
```
carl-mastrangelo: This should have been illegal argument exception, not illegal state.

core/src/main/java/io/grpc/PickFirstBalancerFactory.java
```diff
@@ -46,138 +46,122 @@
...
     }
+
+  }
+
+  /**
+   * No-op picker which doesn't add any custom picking logic. It just passes already known result
+   * received in constructor.
+   */
+  @VisibleForTesting
+  static class Picker extends LoadBalancer.SubchannelPicker {
```
carl-mastrangelo: final

core/src/main/java/io/grpc/PickFirstBalancerFactory.java
```diff
@@ -46,138 +46,122 @@
...
+    private final LoadBalancer.PickResult result;
+
+    Picker(LoadBalancer.PickResult result) {
```
carl-mastrangelo: checkNotNull for visible constructor

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -197,39 +193,11 @@ public final T nameResolverFactory(NameResolver.Factory resolverFactory) {
   }
 
   @Override
-  public final T loadBalancerFactory(LoadBalancer2.Factory loadBalancerFactory) {
+  public final T loadBalancerFactory(LoadBalancer.Factory loadBalancerFactory) {
     Preconditions.checkState(directServerAddress == null,
         "directServerAddress is set (%s), which forbids the use of LoadBalancerFactory",
```
carl-mastrangelo: This should be `LoadBalancer.Factory`

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -307,26 +275,22 @@ public ManagedChannel build() {
...
+        getNameResolverParams(),
+        firstNonNull(loadBalancerFactory, PickFirstBalancerFactory.getInstance()),
+        transportFactory,
+        firstNonNull(decompressorRegistry, DecompressorRegistry.getDefaultInstance()),
+        firstNonNull(compressorRegistry, CompressorRegistry.getDefaultInstance()),
+        SharedResourcePool.forResource(GrpcUtil.TIMER_SERVICE),
+        getExecutorPool(executor),
+        SharedResourcePool.forResource(GrpcUtil.SHARED_CHANNEL_EXECUTOR),
+        GrpcUtil.STOPWATCH_SUPPLIER, idleTimeoutMillis,
+        userAgent, interceptors, firstNonNull(statsFactory,
```
carl-mastrangelo: each of these should be on their own line, like the ones above.

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
```diff
@@ -32,156 +32,185 @@
...
-    }
-    return new FailingClientStream(Status.UNAVAILABLE.withDescription("transport shutdown"));
   }
 
   @Override
   public final ClientStream newStream(MethodDescriptor<?, ?> method, Metadata headers) {
     return newStream(method, headers, CallOptions.DEFAULT, StatsTraceContext.NOOP);
   }
 
+  // Caller must call channelExecutor.drain() outside of lock because this method may schedule
```
carl-mastrangelo: This should be in javadoc format, because some tools don't automatically associate such comments with the method.

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
```diff
@@ -32,156 +32,185 @@
...
+  // tasks on channelExecutor
+  @GuardedBy("lock")
+  private PendingStream createPendingStream(MethodDescriptor<?, ?> method, Metadata headers,
+      CallOptions callOptions, StatsTraceContext statsTraceCtx) {
+    PendingStream pendingStream = new PendingStream(method, headers, callOptions,
```
carl-mastrangelo: line wrap is wrong, it should break at the equals.

ejona86:
This looks fine to me, although it would probably be good to add the shim (`public abstract class LoadBalancer2 extends LoadBalancer {}`) as well.

zhangkun83:
The shim is there.

ejona86:
Ah, I see it now. It was lost in the sea of deletes.

lukaszx0:
Awesome 🤘

PR #2744, opened 2017-02-22
context: declare that context should not be mocked

carl-mastrangelo:
Normally I would make this final, but CancellableContext extends it, and it's public API now.  We can't really add `@DoNotMock` since it is a dependency, even as compileOnly, since the retention is runtime.  It seems a menacing comment is the best we can do.


ejona86:
Please make the comment in the JavaDoc instead.

carl-mastrangelo:
@ejona86 PTAL

PR #2748, opened 2017-02-23
compiler: update path to protoc plugin in README.md

ericgribkoff:
Now matches the actual value used for `javaPluginPath` in
https://github.com/grpc/grpc-java/blob/master/build.gradle#L65.


carl-mastrangelo:
LGTM.  Please update the commit message to be in the "package: reason why grpc is better" format.

ericgribkoff:
Commit message changed. One flaky test failing and reported

Issue #2688, opened 2017-02-02
Would it be possible to shade guava?


elandau:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?
1.0.3 and 1.1.1

### What JVM are you using (`java -version`)?

java version "1.8.0_92"
Java(TM) SE Runtime Environment (build 1.8.0_92-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.92-b14, mixed mode)

### What did you do?

GRPC 1.1+ uses Guava 20.0.   We have a very large codebase that uses both internal and external libraries that depend on older version of guava.  These libraries are not compatible with Guava 20.0.  As a result we are unable to upgrade to GRPC 1.1.+.   From what I can tell the only Guava class exposed by GRPC is ListenableFutures, via ClientCalls in grpc-stub.  Would it be possible to shade all other uses of Guava in GRPC.  

There would still be an issue with the GRPC code generator that does expose Guava's ListenableFuture.  For that use case would it be possible to make generating those stubs optional and perhaps even add an option to generate a CompletableFuture stub?

### What did you expect to see?

Newer versions of GRPC to not break other libraries depending on older version of Guava.

GEverding:
Case and point grpc 1.1.x Isn't compatible with datastax cassandra driver 3.x.

elandau:
@GEverding Looks like datastax is making progress on this.  https://github.com/datastax/java-driver/pull/784

ejona86:
We've discussed possibly shading but excluding ListenableFuture (and maybe SettableFuture), so they still come from Guava.

But in the short term we're more likely to go back to using Guava 19 like we were in v1.0.x which should restore support with older versions of Guava.

ejona86:
And yes, we've heard specific issue with Cassandra. We've also seen https://datastax-oss.atlassian.net/browse/JAVA-1328

zhangkun83:
> But in the short term we're more likely to go back to using Guava 19 like we were in v1.0.x which should restore support with older versions of Guava.

If we are also going to keep Guava 19 for further releases, we need to ask Census/instrumentation to downgrade to 19 too. Their previous releases have been using 19, but they have [bumped to 20](https://github.com/google/instrumentation-java/commit/582e13cd791bbb3a137cc11fe8e1de1887300136) on their master. @sebright

elandau:
[...]

PR #2751, opened 2017-02-24
all: Downgrade to Guava 19

ejona86:
Guava 20 introduced some overloading optimizations for Preconditions
that require using Guava 20+ at runtime. Unfortunately, Guava 20 removes
some things that is causing incompatibilities with other libraries, like
Cassandra. While the incompatibility did trigger some of those libraries
to improve compatibility for newer Guavas, we'd like to give the
community more time to work through it. See #2688

At this commit, we appear to be compatible with Guava 18+. It's not
clear if we want to actually "support" 18, but it did compile. Guava 17
doesn't have at least MoreObjects, directExecutor, and firstNotNull.
Guava 21 compiles without warnings, so it should be compatible with
Guava 22 when it is released.

One test method will fail with the upcoming Guava 22, but this won't
impact applications. I made MoreThrowables to avoid using any
known-deprecated Guava methods in our JARs, to reduce pain for those
stuck with old versions of gRPC in the future (July 2018).

In the stand-alone Android apps I removed unnecessary explicit deps
instead of syncing the version used.

### DIFF
--- a/core/src/main/java/io/grpc/internal/MoreThrowables.java
+++ b/core/src/main/java/io/grpc/internal/MoreThrowables.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016, Google Inc. All rights reserved.
+ * Copyright 2017, Google Inc. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/grpc/internal/MoreThrowables.java
```diff
@@ -0,0 +1,56 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.grpc.internal;
+
+import com.google.common.base.Preconditions;
+
+/** Utility functions when interacting with {@link Throwables}. */
```
carl-mastrangelo: TODO: remove this when upgrading.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1254,7 +1254,7 @@ protected void assertRemoteAddr(String expectedRemoteAddress) {
 
     HostAndPort remoteAddress = HostAndPort.fromString(serverCallCapture.get().getAttributes()
             .get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR).toString());
-    assertEquals(expectedRemoteAddress, remoteAddress.getHost());
+    assertEquals(expectedRemoteAddress, remoteAddress.getHostText());
```
carl-mastrangelo: This is deprecated, and will make it harder to go back up a version.   Do we even need to use it?

jhspaybar:
This is great and will let us get on the 1.1.x branch while waiting for the cassandra and other fixes.  Thank you.

ejona86:
@carl-mastrangelo, PTAL

carl-mastrangelo:
@jhspaybar FYI this will probably not make it into 1.1.x.  We plan on cutting 1.2 this week, and releasing it mid march.

ejona86:
retest this please (for Jenkins)

ejona86:
:'( I failed to squash first.

carl-mastrangelo:
:shame:

lukaszx0:
pro tip (something that I've been doing recently): never commit something with "Address review comments" message when addressing review comments hehe, even if it's a single char change. That way, in the case of something like this, the history will look good even if PR was merged in unsquashed.

Anyways, :shame: 😉

PR #2557, opened 2016-12-28
grpclb: re-implement GrpclbLoadBalancer in v2 API.

zhangkun83:
Besides API changes, this implementation is also up-to-date with the
latest design:

1. Delegate to round-robin and pick-first policies if requested by
the naming system.

2. OOB channels to LoadBalancer always use the LB authority provided by
the naming system.

3. Never send application RPCs to balancer addresses, even if the
address returns UNIMPLEMENTED error.

PR #2754, opened 2017-02-25
grpclb: fix a bug in handling server address updates.

zhangkun83:
It didn't check if the address was already there in the current
subchannel set before creating a new subchannel for it, causing a leak
of subchannels.

### DIFF
--- a/grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancerTest.java
+++ b/grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancerTest.java
@@ -601,15 +601,20 @@ public class GrpclbLoadBalancerTest {
     assertSame(error1, picker6.result.getStatus());
 
     // Update backends, with a drop entry
-    List<InetSocketAddress> backends2 = Arrays.asList(
-        new InetSocketAddress("127.0.0.1", 2030), null);
+    List<InetSocketAddress> backends2 =
+        Arrays.asList(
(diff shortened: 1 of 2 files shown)


dapengzhang0:
I remember I had a comment on this before in #2557 and you replied

> Duplicate EAGs from the balancer are intentionally preserved. They are used to distribute load unevenly across backends, a.k.a. weighted round-robin.

Was that the same concern as this PR?

dapengzhang0:
never mind, I understood that.

grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancerTest.java
```diff
@@ -601,15 +601,20 @@ public void grpclbWorking() {
     assertSame(error1, picker6.result.getStatus());
 
     // Update backends, with a drop entry
-    List<InetSocketAddress> backends2 = Arrays.asList(
-        new InetSocketAddress("127.0.0.1", 2030), null);
+    List<InetSocketAddress> backends2 =
```
dapengzhang0: This test code may be too long. Maybe it's testing multiple things that could be separated?

zhangkun83: I agree the state of this test method is not ideal. Although it's testing multiple behaviors, every test depends the state left by the the previous tests as prerequisites, therefore it's non-trivial to split them apart, and I don't have the time to do it.

dapengzhang0:
ref  #2772

PR #2755, opened 2017-02-26
 core: unify EquivalentAddressGroup and its immitators.

### LINKED ISSUE #2716: Unify usage of EAG and ResolvedServer
@zhangkun83 [wrote](https://github.com/grpc/grpc-java/pull/2696#issuecomment-277759343):

> The equality of EquivalentAddressGroup, which only counts the addresses but not the attributes, was needed by the v1 ManagedChannelImpl for de-duplicating TransportSets. Even though ManagedChannelImpl2 doesn't forbid duplicating InternalSubchannel for the same EquivalentAddressGroup, the equality is still needed by most LoadBalancer2 implementors, e.g., Map<EquivalentAddressGroup, Subchannel> subchannels in the round-robin LB, to determine which InternalSubchannels to close and which to keep, when the NR result is updated.
> 
> I agree that there are too many address wrappers and it has become very confusing, sometimes even to myself. I prefer to keep the EquivalentAddressGroup name, because it indicates how these addresses are used by the channel. So instead of deleting EquivalentAddressGroup, I suggest deleting ResolvedServerInfo, and adding attributes to EquivalentAddressGroup. To address the equality requirement I mentioned earlier, we could add a withoutAttributes() method which can be used in cases where attributes should not be counted for equality.
> 
> ResolvedServerInfoGroup was intended to be the attribute-ful counter-part of EquivalentAddressGroup. Since you are proposing to replace EquivalentAddressGroup with ResolvedServerInfo, I would expect ResolvedServerInfoGroup to go away instead of building yet another level on top of ResolvedServerInfo.

zhangkun83:
Resolves #2716

- Add attributes to EquivalentAddressGroup
- Replace ResolvedServerInfoGroup with EquivalentAddressGroup
- Delete ResolvedServerInfo, because attributes for individual addresses
  within an address group is not found to be useful.

The result is a simpler API and reduction of boilderplates.

As a related change, redefine the semantics of DnsNameResolver and
RoundRobinLoadBalancer:

- Before: DnsNameResolver returns all addresses in one address group.
  RoundRobinLoadBalancer ignores the grouping of addresses and
  round-robin on every single addresses.  It doesn't work well with the
  one-server-multiple-address setup, e.g., both IPv4 and IPv6 addresses
  are returned for a single serve, even if they are put in the same
  address group by the NameResolver.

- After: DnsNameResolver returns every address in its own
  EAG. RoundRobinLoadBalancer takes an EAG as a whole, and only
  round-robin on the list of EAGs. The new behavior is a better
  interpretation of the EAGs, and really allows the case where one
  server has more than one addresses (e.g., IPv4 and IPv6).

Caveats:

Besides trivial API breakages on NameResolver and LoadBalancer, this change will affect users that use custom LoadBalancer with the stock DnsNameResolver, and those who use custom NameResolver with the stock RoundRobinLoadBalancer. They will need to update their LoadBalancer and NameResolver implementations.

However, users who use both the stock DnsNameResolver and RoundRobinLoadBalancer or PickFirstBalancer will see no behavioral change. Because they will still round-robin on individual addresses from DNS, or do pick-first on all addresses from DNS (PickFirstBalancer flattens all addresses).


ejona86:
Talked offline. This change seems fine. I have some concern that our API is more powerful than C's, but not enough to hold it back. Conceptually I think it makes sense when you compare our original API `ResolvedServerInfo` to the new API `List<EquivalentAddressGroup>`. We just had a bit of a detour at `List<ResolvedServerInfo>`.

core/src/main/java/io/grpc/EquivalentAddressGroup.java
```diff
@@ -48,27 +48,45 @@
...
+  private final Attributes attrs;
 
   /**
    * {@link SocketAddress} docs say that the addresses are immutable, so we cache the hashCode.
    */
   private final int hashCode;
 
   /**
-   * List constructor.
+   * List constructor without Attributes.
```
zhangkun83: Done.

core/src/main/java/io/grpc/EquivalentAddressGroup.java
```diff
@@ -48,27 +48,45 @@
...
    */
   public EquivalentAddressGroup(List<SocketAddress> addrs) {
+    this(addrs, Attributes.EMPTY);
+  }
+
+  /**
+   * List constructor with Attributes.
```
zhangkun83: Done.

core/src/main/java/io/grpc/EquivalentAddressGroup.java
```diff
@@ -48,27 +48,45 @@
...
+   */
+  public EquivalentAddressGroup(List<SocketAddress> addrs, Attributes attrs) {
     Preconditions.checkArgument(!addrs.isEmpty(), "addrs is empty");
+    Preconditions.checkNotNull(attrs, "attrs");
     this.addrs = Collections.unmodifiableList(new ArrayList<SocketAddress>(addrs));
+    this.attrs = attrs;
```
zhangkun83: Done.

core/src/main/java/io/grpc/EquivalentAddressGroup.java
```diff
@@ -104,6 +129,9 @@ public boolean equals(Object other) {
         return false;
       }
     }
+    if (!attrs.equals(that.attrs)) {
+      return false;
```
zhangkun83: Done.

zhangkun83:
@lukaszx0 PTAL

zhangkun83:
Actually it turns out not very hard to keep it backward-compatible and give implementations a smoother transition path. I have filed #2848 to track the deletion of the deprecated classes.

lukaszx0:
LGTM, Thanks!

PR #2759, opened 2017-02-26
doc: fix @link to cancel in java doc

lukaszx0:
Minor fix to make linking work in IDEs.


ejona86:
The travis failures appear to be flakes. I've filed notes about the flakes in appropriate issues and restarted the jobs. But I feel fine with merging it as-is.

PR #2766, opened 2017-02-28
core: cache Accept-Encoding headers

carl-mastrangelo:
This avoids rebuilding the raw bytes for each RPC.  The decode path
is not yet optimized to avoid pulling to much into a single commit.
Decompressors may still use invalid names, but this is equivalent to
previous behavior, as cleaing happens later in the caching.

Also, internal accessors on DecompressorRegistry are now hidden.


carl-mastrangelo:
Actually, hold off on reviewing, my git has failed me, and I need to resurrect the actual commit.

carl-mastrangelo:
@ejona86 ready fro review

core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -82,8 +87,20 @@
   /**
    * {@link io.grpc.Metadata.Key} for the accepted message encodings header.
    */
-  public static final Metadata.Key<String> MESSAGE_ACCEPT_ENCODING_KEY =
-          Metadata.Key.of(GrpcUtil.MESSAGE_ACCEPT_ENCODING, Metadata.ASCII_STRING_MARSHALLER);
+  public static final Metadata.Key<byte[]> MESSAGE_ACCEPT_ENCODING_KEY =
+      InternalMetadata.keyOf(GrpcUtil.MESSAGE_ACCEPT_ENCODING, new AcceptEncodingMarshaller());
+
+  private static final class AcceptEncodingMarshaller implements TrustedAsciiMarshaller<byte[]> {
```
ejona86: Should we be concerned about the character restrictions checking that is avoided?

carl-mastrangelo: Surprisingly (in a likely bad way), I think this is okay.  Existing Ascii marshallers just do the conversion inside the AsciiKey, which seems about as bad as doing it inside of DecompressorRegistry.  The TransportFrameUtil double checks that all the characters are still in range post conversion, so I think it's okay.

Decompressor name is checked for commas when added to the registry, so I *think* it is still safe.  

The decoding path is safe for the same reasons.

carl-mastrangelo:
retest this please, Jenkins

carl-mastrangelo:
retest this please

carl-mastrangelo:
retest this please

carl-mastrangelo:
retest this please

carl-mastrangelo:
retest this please

Issue #2689, opened 2017-02-02
Reflection service can return inconsistent results


ejona86:
The service currently calls `serverReflectionIndex.updateMutableIndexIfNecessary()` after receiving each request in the stream. This is not only needless processing, but actually against the point. The reflection API is streaming so that each stream can be independently consistent. This is normally important when a load balancer may direct each request to a different backend. However, in this case it could also handle the mutable nature of the services.

Although `serverReflectionIndex` is `volatile`, it is not assigned in a thread-unsafe way; it could be non-`volatile`. However, I would suggest making `ServerReflectionIndex` immutable and create a new/derivative each time the mutable index is detected to change (probably using locking instead of volatile). I would then suggest checking for updates to the mutable index at the beginning of each RPC (`serverReflectionInfo()`) and then copying the current `ServerReflectionIndex` reference to the `ProtoReflectionStreamObserver`.

ericgribkoff:
I spoke with @dklempner about this. As you're suggesting here, the right thing to do seems to be to take a snapshot of the protobufs at the start of each stream. From a quick chat with @y-zeng, this should also be sufficient to avoid blowing up the grpc_cli tool. There is potentially some wiggle room to return an up-to-date list of the services exported, as long as the view of the protos remains consistent.

It may also be useful, at some future point, to consider adding a mechanism in the API for the service to indicate to clients that the data coming over the stream has become stale - but that's beyond the scope here.

ericgribkoff:
Fixed by https://github.com/grpc/grpc-java/pull/2767

PR #2657, opened 2017-01-25
services: make reflection test able to re-root proto

carl-mastrangelo:

PR #2767, opened 2017-02-28
services: don't update reflection index mid-stream

ericgribkoff:
This addresses https://github.com/grpc/grpc-java/issues/2689. `ServerReflectionIndex` is now immutable, and its reference is replaced by a new instance when changes are detected.

I also fixed some poor formatting choices present in the earlier code, primarily the tests - I don't think the changes are too invasive, given that much of the code has genuinely been changed here, and the formatting already popped up as an issue on https://github.com/grpc/grpc-java/pull/2657.

### DIFF
--- a/services/src/main/java/io/grpc/protobuf/service/ProtoReflectionService.java
+++ b/services/src/main/java/io/grpc/protobuf/service/ProtoReflectionService.java
@@ -66,7 +66,6 @@ import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
 import javax.annotation.Nullable;
-import javax.annotation.concurrent.GuardedBy;
 
 /**
  * Provides a reflection service for Protobuf services (including the reflection service itself).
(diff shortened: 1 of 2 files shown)


services/src/main/java/io/grpc/protobuf/service/ProtoReflectionService.java
```diff
@@ -79,49 +78,98 @@
...
-    serverReflectionIndex = new ServerReflectionIndex(checkNotNull(server, "server"));
+    this.server = checkNotNull(server);
+  }
+
+  /**
+   * Checks for updates to the server's mutable services and updates the index if any changes are
+   * detected. A change is any addition or removal in the set of file descriptors attached to the
+   * mutable services or a change in the service names.
+   */
+  private void updateIndexIfNecessary() {
```
ejona86: This is not thread-safe. `serverReflectionInfo()` may be called from multiple threads. Make it synchronized and put `@GuardedBy` on `serverReflectionIndex`?

You'd also need to return `serverReflectionIndex` from this method, to avoid grabbing the lock again in `serverReflectionInfo()`. Alternatively you could use a volatile, but meh.

ericgribkoff: Lock and `@GuardedBy` added.

ejona86: For posterity: this is actually thread-safe, due to `final` usage, but it isn't obviously so and it does take some effort to show it is safe. So we're still going to change it to something more obviously correct.

ejona86:
Travis is broken right now. Feel free to ignore it. It should be fine in a few more hours.

Issue #2292, opened 2016-09-21
Implement channel-state API in ManagedChannelImpl


zhangkun83:
The API is added by #2286, but is implemented (will throw exception) in ManagedChannelImpl.

lukaszx0:
@zhangkun83 I've talked with @ejona86 about this recently. What design do you have in mind? 

Simply adding `getState()` to picker which would return aggregate of picker's subchannels? I'm assuming we also want to keep the "old" API for `ManagedChannel` ?

```java
class ManagedChannel {
  State getState(boolean tryToConnect);
  void notifyWhenStateChanged(State lastObserved, Runnable callback);
}
```

zhangkun83:
@ejona86 has talked with me about that. The idea is that the channel will call `SubchannelPicker.getState()` only once, in `updatePicker()`, so that a new state would require a new picker.

It will work in the sense of implementing the channel state API, but I am yet to figure out how it should interact with `pickSubchannel()`.

Conceptually, `PickResult` maps to the channel state:
```
withSubchannel -> READY
withError -> TRANSIENT_FAILURE
withNoResult -> IDLE, CONNECTING
```

Should we allow `getState()` to disagree with the pick results, e.g., `getState()` returns `TRANSIENT_FAILURE` and in the mean time `pickSubchannel()` returns working subchannels?

> I'm assuming we also want to keep the "old" API for ManagedChannel ?

I am not sure what you are referring to.

lukaszx0:
> I am not sure what you are referring to.

I mean that API for `ManagedChannel2`, the snippet that I've pasted.

> It will work in the sense of implementing the channel state API, but I am yet to figure out how it should interact with pickSubchannel().

That's a tricky one. I'm not sure the `ManagedChannel` state should be tied to the result returned by `pickSubchannel` or maybe I'm misunderstanding the definition and semantics of channel's `state`.

In my opinion `ManagedChannel` state should reflect the state of underlying subchannels, in aggregate. Interaction with `pickSubchannel` would mean that channel's state could be affected / changed by `affinity` of a an RPC call (or `Metadata`).

> The idea is that the channel will call SubchannelPicker.getState() only once, in updatePicker(), so that a new state would require a new picker.

I like the idea and with what I wrote above, I think it should be as simple as adding `abstract ConnectivityState getState()` to the `SubchannelPicker` and let users arbitrarily set it based on the current state of subchannels managed by `LoadBalancer`.

[...]

PR #2768, opened 2017-02-28
core: implement state API for ManagedChannel 

lukaszx0:
Implementation of channel-state API for ManagedChannel (#2292).

### DIFF
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -63,11 +63,9 @@ import io.grpc.internal.ClientCallImpl.ClientTransportProvider;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
@@ -185,7 +183,9 @@ public final class ManagedChannelImpl extends ManagedChannel implements WithLogI
   private volatile boolean terminated;
   private final CountDownLatch terminatedLatch = new CountDownLatch(1);
 
+  @GuardedBy("this")
   private ConnectivityState state;
+  @GuardedBy("this")
   private final Queue<Runnable> stateChangeCallbacks = new LinkedList<Runnable>();
 
   // Called from channelExecutor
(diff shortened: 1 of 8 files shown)


lukaszx0:
@zhangkun83 this is very rough proof of concept. It doesn't fully compile is missing tests, docs etc. but illustrates general idea. Let me know if this is a good direction and what do you think. Thanks!

ejona86:
CC @dapengzhang0

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -527,7 +536,43 @@ public boolean isTerminated() {
...
 
-  /*
+  @Override
+  public ConnectivityState getState(boolean requestConnection) {
+    // TODO: handle requestConnection. At this point, not really sure how.
+    return state;
+  }
+
+  @Override
+  public void notifyWhenStateChanged(ConnectivityState source, Runnable callback) {
```
ejona86: I think you're mis-understanding the semantics of `source`. There is no need to remember `source` after an initial check.

`source` is the currently-believed state, to handle racing. So if `source != state`, then `callback` is run immediately. After that point though, it will be the same as the current state, so you only need to call all `callback`s when the current state changes.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -527,7 +536,43 @@ public boolean isTerminated() {
     return terminated;
   }
 
-  /*
+  @Override
+  public ConnectivityState getState(boolean requestConnection) {
+    // TODO: handle requestConnection. At this point, not really sure how.
```
ejona86: At the very least, you would exit idle mode. You could request an InternalChannel from the LB and then `obtainActiveTransport()`. But it's not clear how this boolean should work when using round robin. I guess that's because it's not clear what the `state` should be when using round robin.

That is to say, we need to work on the spec to define it.

lukaszx0: I've moved discussion to [the issue](https://github.com/grpc/grpc-java/issues/2292#issuecomment-284480422), this comment will get buried after code changes.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
lukaszx0: Ahh yes. I misunderstood it indeed. Makes sense but now my question is how many times the `Runnable` will be ran?

I assume that every time it gets triggered, the callback should be removed and if users want to continue having it registered, they should re-register it inside `Runnable`. Is that correct?

zhangkun83: Yes. The Runnable is one-shot.

lukaszx0: Addressed. PTAL and let me know if the semantics are correct and you had chance to review in general - if so, I can proceed and add tests, docs etc.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -527,7 +535,37 @@ public boolean isTerminated() {
...
+    return state;
+  }
+
+  @Override
+  public void notifyWhenStateChanged(ConnectivityState source, Runnable callback) {
+    if (source != state) {
```
zhangkun83: `notifyWhenStateChanged()` needs to be thread-safe. In fact, all methods on `ManagedChannel` are thread-safe, and that interface misses a `@ThreadSafe`.

The race between `notifyWhenStateChanged()` and `maybeUpdateState()` also needs to be correctly handled.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -202,16 +203,15 @@ private void updatePicker(@Nullable Status error) {
...
-    private Status getAggregatedError() {
-      Status status = null;
+    private ConnectivityStateInfo getAggregatedState() {
+      ConnectivityStateInfo stateInfo = null;
       for (Subchannel subchannel : getSubchannels()) {
-        ConnectivityStateInfo stateInfo = getSubchannelStateInfoRef(subchannel).get();
+        stateInfo = getSubchannelStateInfoRef(subchannel).get();
         if (stateInfo.getState() != TRANSIENT_FAILURE) {
-          return null;
+          return stateInfo;
```
zhangkun83: If there is at least one `READY`, the aggregated state should be `READY`.
Otherwise, if there is at least one `CONNECTING`, the aggregated state should be `CONNECTING`.
Otherwise, if there is at least one `IDLE`, the aggregated state should be `IDLE`.

lukaszx0:
@ejona86 @zhangkun83 pushed updates (still lacking through tests, want to get logic right first)

lukaszx0:
closing in favor of #3019

PR #2771, opened 2017-03-01
services: move proto services to io.grpc.protobuf.services

ericgribkoff:
Moves `io.grpc.protobuf.service` to `io.grpc.protobuf.services`, for consistency with `io.grpc.services` naming.

This leaves behind a `@Deprecated io.grpc.protobuf.service.ProtoReflectionService.getInstance()` method that points to the new `io.grpc.protobuf.services.ProtoReflectionService.newInstance()`.

Despite how it shows in Github's UI, `services/ProtoReflectionService.java` is a copy of the pre-existing `service/ProtoReflectionService.java` file, minus the `getInstance()` method.

I realize now that I should have held off adding `newInstance()` in https://github.com/grpc/grpc-java/pull/2767, as technically `io.grpc.protobuf.service.ProtoReflectionService.newInstance()` is now part of the API, but I'm assuming it's ok to drop this method altogether from the deprecated `io.grpc.protobuf.service.ProtoReflectionService` placeholder, since it's only existed in master a short while and not been part of a published release? Otherwise I should add `@Deprecated io.grpc.protobuf.service.ProtoReflectionService.newInstance()` here.


ejona86:
Dropping newInstance() from the old API is fine.

ejona86:
Checkstyle is failing.

services/src/test/java/io/grpc/protobuf/services/ProtoReflectionServiceTest.java
```diff
@@ -114,6 +114,12 @@ public void tearDown() {
   }
 
   @Test
+  @SuppressWarnings("deprecation")
```
ejona86: Hint (no need to change): you can mark the method as `@Deprecated` instead of suppressing the warning.

PR #2777, opened 2017-03-02
Fix shared subchannel state in RoundRobin LB

jamadeo:
handleResolvedAddresses constructs subchannels for each address in
an EquivalentAddressGroup, sharing a reference to an immutable
Atrributes. However, as noted, this Attributes instance contains
a mutable AtomicReference, eventually causing subchannel state
changes to be improperly reflected in *all* subchannels of an
EquivalentAddressGroup.

This change gives each subchannel its own Attributes instance, mitigating
this issue.

### DIFF
--- a/core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
+++ b/core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
@@ -107,7 +107,7 @@ public class RoundRobinLoadBalancerTest {
       SocketAddress addr = new FakeSocketAddress("server" + i);
       EquivalentAddressGroup eag = new EquivalentAddressGroup(addr);
       servers.put(ResolvedServerInfoGroup.builder().add(new ResolvedServerInfo(addr)).build(), eag);
-      subchannels.put(eag, createMockSubchannel());
+      subchannels.put(eag, mock(Subchannel.class));
     }
 
     when(mockHelper.createSubchannel(any(EquivalentAddressGroup.class), any(Attributes.class)))
@@ -116,7 +116,7 @@ public class RoundRobinLoadBalancerTest {
           public Subchannel answer(InvocationOnMock invocation) throws Throwable {
             Object[] args = invocation.getArguments();
             Subchannel subchannel = subchannels.get(args[0]);
-            mockSubchannelAttributes(subchannel, (Attributes)args[1]);
+            when(subchannel.getAttributes()).thenReturn((Attributes)args[1]);
             return subchannel;
           }
         });
@@ -128,8 +128,8 @@ public class RoundRobinLoadBalancerTest {
   @Test
   public void pickAfterResolved() throws Exception {
     final Subchannel readySubchannel = subchannels.values().iterator().next();
-    mockSubchannelToConnect(readySubchannel);
     loadBalancer.handleResolvedAddresses(Lists.newArrayList(servers.keySet()), affinity);
+    loadBalancer.handleSubchannelState(readySubchannel, ConnectivityStateInfo.forNonError(READY));
 
     verify(mockHelper, times(3)).createSubchannel(eagCaptor.capture(),
         any(Attributes.class));
@@ -140,7 +140,7 @@ public class RoundRobinLoadBalancerTest {
       verify(subchannel, never()).shutdown();
     }
 
-    verify(mockHelper, times(1)).updatePicker(pickerCaptor.capture());
+    verify(mockHelper, times(2)).updatePicker(pickerCaptor.capture());
 
     assertThat(pickerCaptor.getValue().getList()).containsExactly(readySubchannel);
 
@@ -155,7 +155,9 @@ public class RoundRobinLoadBalancerTest {
 
     for (Subchannel subchannel : Lists.newArrayList(removedSubchannel, oldSubchannel,
         newSubchannel)) {
-      mockSubchannelState(subchannel, ConnectivityStateInfo.forNonError(READY));
+      when(subchannel.getAttributes()).thenReturn(Attributes.newBuilder().set(STATE_INFO,
+          new AtomicReference<ConnectivityStateInfo>(
+              ConnectivityStateInfo.forNonError(READY))).build());
     }
 
     FakeSocketAddress removedAddr = new FakeSocketAddress("removed");
@@ -324,13 +326,13 @@ public class RoundRobinLoadBalancerTest {
   @Test
   public void nameResolutionErrorWithActiveChannels() throws Exception {
     final Subchannel readySubchannel = subchannels.values().iterator().next();
-    mockSubchannelToConnect(readySubchannel);
     loadBalancer.handleResolvedAddresses(Lists.newArrayList(servers.keySet()), affinity);
+    loadBalancer.handleSubchannelState(readySubchannel, ConnectivityStateInfo.forNonError(READY));
     loadBalancer.handleNameResolutionError(Status.NOT_FOUND.withDescription("nameResolutionError"));
 
     verify(mockHelper, times(3)).createSubchannel(any(EquivalentAddressGroup.class),
         any(Attributes.class));
-    verify(mockHelper, times(2)).updatePicker(pickerCaptor.capture());
+    verify(mockHelper, times(3)).updatePicker(pickerCaptor.capture());
 
     LoadBalancer.PickResult pickResult = pickerCaptor.getValue().pickSubchannel(Attributes.EMPTY,
         new Metadata());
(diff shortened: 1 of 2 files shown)


lukaszx0:
Ugh, nice catch. @jamadeo could you add a tests case for this?

jamadeo:
Thanks for taking a look @lukaszx0, just added a test case.

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
```diff
@@ -352,11 +352,40 @@ public void nameResolutionErrorWithActiveChannels() throws Exception {
...
+    final List<Subchannel> createdSubchannels = Lists.newArrayList();
+    when(mockHelper.createSubchannel(any(EquivalentAddressGroup.class), any(Attributes.class)))
+        .then(new Answer<Subchannel>() {
+          @Override
+          public Subchannel answer(InvocationOnMock invocation) throws Throwable {
+            Subchannel sc = createMockSubchannel((Attributes)invocation.getArguments()[1]);
+            createdSubchannels.add(sc);
+            return sc;
+          }
+        });
```
zhangkun83: Could you merge this change into the stubbing of `createSubchannel()` in `setUp()`?

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
```diff
@@ -352,11 +352,40 @@ public void nameResolutionErrorWithActiveChannels() throws Exception {
...
+
+    Subchannel sc1 = createdSubchannels.get(0);
+    Subchannel sc2 = createdSubchannels.get(1);
+    assertEquals(IDLE, sc1.getAttributes().get(STATE_INFO).get().getState());
+    assertEquals(IDLE, sc2.getAttributes().get(STATE_INFO).get().getState());
+
+    loadBalancer.handleSubchannelState(sc1, ConnectivityStateInfo.forNonError(READY));
+    assertEquals(READY, sc1.getAttributes().get(STATE_INFO).get().getState());
+    assertEquals(IDLE, sc2.getAttributes().get(STATE_INFO).get().getState());
+  }
```
zhangkun83: Please add a 3rd subchannel transitioned to the `TRANSIENT_FAILURE` state, and verify that it produces pickers as expected:
 - Initially, all three subchannels
 - After 2nd subchannel becomes `IDLE`: a new picker with all three subchannels
 - After 3rd subchannel bcomes `TRANSIEANT_FAILURE`: a new picker with the first two subchannels.

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
jamadeo: Done.

It's a bit different from the original in order to keep compatibility with the other tests that reference the subchannel mocks before they would be created, but I think the goal is accomplished.

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
jamadeo: Done.

Note that when a subchannel `IDLE` subchannels won't be included in the picker, but will be requested to connect. I've tried to assert this behavior in the new changes.

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
```diff
@@ -360,6 +377,25 @@ private Subchannel createMockSubchannel() {
     return subchannel;
   }
 
+  private void mockSubchannelAttributes(Subchannel subchannel, Attributes attributes) {
+    when(subchannel.getAttributes()).thenReturn(attributes);
+  }
+
+  private void mockSubchannelState(Subchannel subchannel, ConnectivityStateInfo state) {
+    mockSubchannelAttributes(subchannel, Attributes.newBuilder().set(STATE_INFO,
```
zhangkun83: The Attributes of a Subchannel is created more than once -- once in `createMockSubchannel()` and possibly multiple times in here. This is not right, because there should be only one Attributes instance for each Subchannel.

Actually, it should be the LB that set the STATE_INFO attribute. Setting it from outside is not expected by the LB. So for every call site of `mockSubchannelState()`, I would call `handleSubchannelState()` and leave the LB to update STATE_INFO. Granted it is a preexisting issue, but it's good to address it now.

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
```diff
@@ -360,6 +377,25 @@ private Subchannel createMockSubchannel() {
     return subchannel;
   }
 
+  private void mockSubchannelAttributes(Subchannel subchannel, Attributes attributes) {
+    when(subchannel.getAttributes()).thenReturn(attributes);
```
zhangkun83: This is the proper way to mock the Subchannel. Please delete `createMockSubchannel()` and replace its call sites with a plain `mock(Subchannel.class)` (without stubbing `getAttributes()`).

Because this is a one-liner with only one caller (after deleting `mockSubchannelState` as I suggest below), it's better to merge it into the call site.

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
```diff
@@ -360,6 +377,25 @@ private Subchannel createMockSubchannel() {
...
+  }
+
+  private void mockSubchannelState(Subchannel subchannel, ConnectivityStateInfo state) {
+    mockSubchannelAttributes(subchannel, Attributes.newBuilder().set(STATE_INFO,
+        new AtomicReference<ConnectivityStateInfo>(state)).build());
+  }
+
+  private void mockSubchannelToConnect(final Subchannel subchannel) {
```
zhangkun83: As I mentioned in the comment above, it should be only LB who will update STATE_INFO. In this particular case, `requestConnection()` should trigger a `handleSubchannelState()` instead of updating the STATE_INFO directly.

Also, this stubbing logic is too far away from the test logic, which is not easy to follow. I would prefer explicitly verifying `requestConnection()` and call `handleSubchannelState()` from the test method. It seems the current call sites don't even care if `requestConnection()` is called, so you could skip the verification and just call `handleSubchannelState()` to trigger state change.

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
```diff
@@ -352,6 +343,32 @@ public void nameResolutionErrorWithActiveChannels() throws Exception {
...
+    Subchannel sc2 = subchannelIterator.next();
+    Subchannel sc3 = subchannelIterator.next();
+    mockSubchannelToConnect(sc1);
+    mockSubchannelToConnect(sc2);
+    mockSubchannelToConnect(sc3);
+
+    loadBalancer.handleResolvedAddresses(Lists.newArrayList(servers.keySet()), Attributes.EMPTY);
+    loadBalancer
+        .handleSubchannelState(sc2, ConnectivityStateInfo.forNonError(IDLE));
+    mockSubchannelState(sc2, ConnectivityStateInfo.forNonError(READY));
```
jamadeo: Done

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
jamadeo: Done

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
jamadeo: Done

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
jamadeo: Done

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
```diff
@@ -105,15 +107,17 @@ public void setUp() {
...
     }
 
     when(mockHelper.createSubchannel(any(EquivalentAddressGroup.class), any(Attributes.class)))
         .then(new Answer<Subchannel>() {
           @Override
           public Subchannel answer(InvocationOnMock invocation) throws Throwable {
             Object[] args = invocation.getArguments();
-            return subchannels.get(args[0]);
+            Subchannel subchannel = subchannels.get(args[0]);
+            when(subchannel.getAttributes()).thenReturn((Attributes)args[1]);
```
zhangkun83: nit: add a space before `args[1]`.

jamadeo: Done

lukaszx0:
Nice. Thanks @jamadeo for finding and fixing this! 👍

zhangkun83:
retest this please

PR #2778, opened 2017-03-03
core/netty/okhttp: KeepAliveManager with Pinger

dapengzhang0:
* Preparing to support server side keepalive, modified KeepAliveManager to use a Pinger interface, which can send ping or shutdown transport regardless of server or client.

* On server side, keepalive is happening even the transport is idle, so the IDLE and IDLE_AND_PING_SENT states do not make sense for server. Therefore removed `IDLE`, `IDLE_AND_PING_SENT`, and `DISCONNECTED` states, and used boolean variables `isIdle`, `isShutdown`, and `noPingWhenIdle` to do more fine-grained control.

* Not to use Ping `onSuccess()` callback to cancle shutdownFuture any more, instead, regard `onDataReceived()` as ping Ack.

### DIFF
--- a/core/src/main/java/io/grpc/internal/KeepAliveManager.java
+++ b/core/src/main/java/io/grpc/internal/KeepAliveManager.java
@@ -31,10 +31,13 @@
 
 package io.grpc.internal;
 
+import static com.google.common.base.Preconditions.checkState;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.MoreExecutors;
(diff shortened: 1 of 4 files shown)


dapengzhang0:
I will split the PR into smaller parts @ejona86

ejona86:
I'm quite scared to change the state machine at this point. Changing the state machine makes it sort of equivalent to a re-write. I think we could handle servers sending keepalive on idle by just avoiding the idle state when on the server. I imagine that could be done with 3-4 lines of code, by just disabling the transition to IDLE.

PR #2780, opened 2017-03-03
core: update contributing guidelines w/ pr/commit conventions

lukaszx0:
We probably want to have more there in the future, but it's a good start.

cc @carl-mastrangelo

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,9 +41,11 @@ for reviewers:
 
   `package-name: summary of change`
 
+  where the summary finishes the sentence: `This commit improves gRPC to ____________.`
+
   for example:
 
-  `core: fix infinite loop`
+  `core,netty,interop-testing: add capacitive duractance to turbo encabulators`
 
 * Every time you receive a feedback on your pull request, push changes that
   address it as a separate one or multiple commits with a descriptive commit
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -32,6 +32,26 @@ channel on Freenode](http://webchat.freenode.net/?channels=grpc), or send an
...
+We have few conventions for keeping history clean and making code reviews easier
+for reviewers:
+
+* First line of commit messages should be in format of
+
+  `package-name: summary of change`
+
+  for example:
+
+  `core: fix infinite loop`
```
carl-mastrangelo: Thanks for doing this.  If you wouldn't mind, could you change the template on line 42 to be

```
package-name: summary of change
Where the summary finishes the sentence: "This commit improves gRPC to ____________.".
```

This format was roughly copied from Go and the Linux kernel.  Preferably, the reason starts with a lower case letter, though it seems this has not been consistently followed.  I personally don't care which, as long as it's one of them.  (it makes it way easier to scan when building the release notes!)

Also, if you wouldn't mind, could you additionally add a slightly more complicated example.  Something like:

```
core,netty,interop-testing:  add capacitive duractance to turbo encabulators
````

lukaszx0: I wanted to make it more descriptive but honestly wasn't sure what are the current conventions and what are the ones that we want so I didn't want to jump the gun cause technically I'm not even an maintainer ;)

Updated. PTAL.

PR #2786, opened 2017-03-03
core: fix keepalivemanager bug on handling IDLE_AND_PING_SENT (backport 1.0.x)

dapengzhang0:
backport of #2773 
There was a bug:
IDLE_AND_PING_SENT -> a new stream starts -> ping acked
would lead the keepalivemanager into IDLE state forever.


ejona86:
You're backporting to 1.0. Should it be to 1.1 instead? Or will there be another PR for 1.1?

dapengzhang0:
There is another PR for 1.1. Shouldn't we backport to 1.0 as well?

ejona86:
The v1.1 hadn't been created yet when I saw this. I'm not sure if/when we'll do another 1.0.x release. It's fine either way.

Issue #2738, opened 2017-02-17
Use CallOptions.withOptions instead of withAffinity


ejona86:
`withAffinity` is a full `Attributes` to be type-safe, but it will generally only have one entry and _sometimes_ two. In more recent times we've added `withOptions`. This seems it could just use it instead. This would mean the `LoadBalancer` would be passed `CallOptions` instead of `Attributes affinity`, but that doesn't seem too bad.

Probably want to do #2737 at the same time as fixing this.

zhangkun83:
The transition plan:

1. In the next release, mark `CallOptions.withAffinity()` and `getAffinity()` as deprecated
2. In the subsequent release, delete them.

zhangkun83:
@lukaszx0 @jhspaybar

infinityat0:
So, are you guys suggesting that the loadbalancer would get callOptions instead of affinity?

With that change, I'm expecting SubChannelPicker method to change from 
`public abstract PickResult pickSubchannel(Attributes affinity, Metadata headers);`

to 

`public abstract PickResult pickSubchannel(CallOptions options, Metadata headers);`

lukaszx0:
@infinityat0 this is what has been already done in #2740 linked above.

infinityat0:
@lukaszx0 Perfect. Which release are we to expect this change?

lukaszx0:
@infinityat0 It'll be included in 1.2 (if you'll go to aforementioned PR, you can see that it's part of 1.2 milestone)

infinityat0:
@lukaszx0 Oops! didn't notice that. Thank you!

PR #2791, opened 2017-03-04
core: use withOptions instead of withAffinity

lukaszx0:
Addresses #2738

cc @ejona86 @zhangkun83

### DIFF
--- a/core/src/main/java/io/grpc/CallOptions.java
+++ b/core/src/main/java/io/grpc/CallOptions.java
@@ -156,6 +156,7 @@ public final class CallOptions {
    * @deprecated use {@link #withOption(Key, Object)}.
    */
   @Deprecated
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1766")
   public CallOptions withAffinity(Attributes affinity) {
     CallOptions newOptions = new CallOptions(this);
     newOptions.affinity = Preconditions.checkNotNull(affinity, "affinity");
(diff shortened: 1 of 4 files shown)


ejona86:
The build fails due to a deprecation warning.

core/src/main/java/io/grpc/CallOptions.java
```diff
@@ -152,8 +152,10 @@ public Deadline getDeadline() {
 
   /**
    * Returns a new {@code CallOptions} with attributes for affinity-based routing.
+   *
+   * @deprecated use {@link #withOption(Key, Object)}.
    */
-  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1766")
+  @Deprecated
```
ejona86: Go ahead and leave the `@ExperimentalApi` annotation, to show that we can actually delete it.

lukaszx0: Done.

lukaszx0:
> The build fails due to a deprecation warning.

Fixed. PTAL.

lukaszx0:
@ejona86 :shipit: ?

lukaszx0:
No rush, but it'd be great to merge it and include in `1.2.x` release since "first half" of it has `PickSubchannelArgs` already landed.

ejona86:
This doesn't seem exceptional enough to include in 1.2.x, especially since the 1.2.x binary creation would happen today-ish, other than #2801. Now that we're getting on a frequent release schedule, it doesn't seem to hurt too much to just delay.

lukaszx0:
Ok, sounds good.

PR #2793, opened 2017-03-06
services: Remove unused variables

ejona86:


ejona86:
I fixed the checkstyle failure. It was caused by my change.

PR #2795, opened 2017-03-06
okhttp: Add restricted AppEngine SSL setup

kpayson64:
Per request of the VTK team, we want to bake in the needed SSL workaround for AppEngine standard.

### DIFF
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
@@ -53,6 +53,8 @@ import io.grpc.okhttp.internal.Platform;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.security.GeneralSecurityException;
+import java.security.KeyStore;
+import java.security.SecureRandom;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -60,6 +62,7 @@ import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManagerFactory;
 
 /** Convenience class for building channels with the OkHttp transport. */
 @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1785")
@@ -252,7 +255,25 @@ public class OkHttpChannelBuilder extends
       case TLS:
         try {
           if (sslSocketFactory == null) {
-            SSLContext sslContext = SSLContext.getInstance("Default", Platform.get().getProvider());
+            SSLContext sslContext;
+            if (GrpcUtil.IS_RESTRICTED_APPENGINE) {
+              // The following auth code circumvents the following AccessControlException:
+              // access denied ("java.util.PropertyPermission" "javax.net.ssl.keyStore" "read")
+              // Conscrypt will attempt to load the default KeyStore if a trust manager is not
+              // provided, which is forbidden on AppEngine
+              sslContext = SSLContext.getInstance("TLS", Platform.get().getProvider());
+              TrustManagerFactory trustManagerFactory =
+                  TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+              trustManagerFactory.init((KeyStore) null);
+              sslContext.init(
+                  null,
+                  trustManagerFactory.getTrustManagers(),
+                  // Use an algorithm that doesn't need /dev/urandom
+                  SecureRandom.getInstance("SHA1PRNG", Platform.get().getProvider()));
+
+            } else {
+              sslContext = SSLContext.getInstance("Default", Platform.get().getProvider());
+            }
             sslSocketFactory = sslContext.getSocketFactory();
           }
           return sslSocketFactory;


kpayson64:
If I do some hacky stuff to make IS_RESTRICTED_APPENGINE=true for a unit test, it still seems the JDK used for unit tests doesn't support the SHA1PRNG. 

```
io.grpc.okhttp.OkHttpChannelBuilderTest > restrictedAppEngineSocketFactory FAILED
    java.lang.RuntimeException: TLS Provider failure
        at io.grpc.okhttp.OkHttpChannelBuilder.createSocketFactory(OkHttpChannelBuilder.java:284)
        at io.grpc.okhttp.OkHttpChannelBuilderTest.restrictedAppEngineSocketFactory(OkHttpChannelBuilderTest.java:140)

        Caused by:
        java.security.NoSuchAlgorithmException: no such algorithm: SHA1PRNG for provider SunJSSE
            at sun.security.jca.GetInstance.getService(GetInstance.java:101)
            at sun.security.jca.GetInstance.getInstance(GetInstance.java:218)
            at java.security.SecureRandom.getInstance(SecureRandom.java:383)
            at io.grpc.okhttp.OkHttpChannelBuilder.createSocketFactory(OkHttpChannelBuilder.java:276)
            ... 1 more
```

This code path gets exercised in an AppEngine end to end test.

kpayson64:
@carl-mastrangelo 
After thinking this over, I don't think adding a unit tests for this with mocked providers adds much value.

If anything, it adds a false sense of security.  I wouldn't feel comfortable making any changes to this logic without testing it in AppEngine.  This code path gets exercised as part of an AppEngine end to end test, and I think that is the better method of testing.

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -252,12 +257,33 @@ SSLSocketFactory createSocketFactory() {
       case TLS:
         try {
           if (sslSocketFactory == null) {
-            SSLContext sslContext = SSLContext.getInstance("Default", Platform.get().getProvider());
+            SSLContext sslContext = null;
```
carl-mastrangelo: The assignment of null is not necessary.

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -252,12 +257,33 @@ SSLSocketFactory createSocketFactory() {
...
+            if (GrpcUtil.IS_RESTRICTED_APPENGINE) {
+              // The following auth code circumvents the following AccessControlException:
```
carl-mastrangelo: This comment is good, but it doesn't really explain why or how.  Where is that exception thrown from?  Why does this block of code change that?

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -252,12 +257,33 @@ SSLSocketFactory createSocketFactory() {
...
+              // access denied ("java.util.PropertyPermission" "javax.net.ssl.keyStore" "read")
+              sslContext = SSLContext.getInstance("TLS", Platform.get().getProvider());
+              KeyStore keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
+              keyStore.load(null, null);
```
carl-mastrangelo: what is null?

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -252,12 +257,33 @@ SSLSocketFactory createSocketFactory() {
...
+              KeyManagerFactory keyManagerFactory =
+                  KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
+              keyManagerFactory.init(keyStore, null);
```
carl-mastrangelo: here too, what does null mean?

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -252,12 +257,33 @@ SSLSocketFactory createSocketFactory() {
...
+              TrustManagerFactory trustManagerFactory =
+                  TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+              trustManagerFactory.init((KeyStore) null);
```
carl-mastrangelo: Why aren't you using the keyStore you made above?

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -252,12 +257,33 @@ SSLSocketFactory createSocketFactory() {
...
+              sslContext.init(
+                  null,
+                  trustManagerFactory.getTrustManagers(),
+                  SecureRandom.getInstance("SHA1PRNG", Platform.get().getProvider()));
```
carl-mastrangelo: What is special about SHA1PRNG?

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -252,12 +257,33 @@ SSLSocketFactory createSocketFactory() {
...
+            } else {
+              sslContext = SSLContext.getInstance("Default", Platform.get().getProvider());
+            }
             sslSocketFactory = sslContext.getSocketFactory();
           }
           return sslSocketFactory;
         } catch (GeneralSecurityException gse) {
           throw new RuntimeException("TLS Provider failure", gse);
+        } catch (IOException e) {
```
carl-mastrangelo: What throws this exception?

carl-mastrangelo:
@kpayson64 We can't really accept untested code.  You may not feel comfortable making changes to this code, but you may not be the person making changes, or know the context.  Adding unit tests proves that code is not accidentally broken, but also documents how it *should* work.  

Look at it from gRPC Java team's point of view.  This code has no tests so we don't know if changes we make to it will break.   It isn't clear why it is doing things differently.  It reduces our test coverage.  It will have a really long time between breakages and finding it, since the commits will take 1 -2 weeks to be brought in.  Lastly, we have no way to actually run these.  It is unreasonable to expect a custom test environment to be brought up to try these out.  Not all contributors may know how to do it, or even be able.  

This code is not maintainable.

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
kpayson64: Removed

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
kpayson64: I've added some details to the comment, but the root of the stack trace looks like this:

```
Caused by: java.security.AccessControlException: access denied ("java.util.PropertyPermission" "javax.net.ssl.keyStore" "read")
	at java.security.AccessControlContext.checkPermission(AccessControlContext.java:484)
	at java.security.AccessController.checkPermission(AccessController.java:698)
	at java.lang.SecurityManager.checkPermission(SecurityManager.java:549)
	at java.lang.SecurityManager.checkPropertyAccess(SecurityManager.java:1298)
	at java.lang.System.getProperty(System.java:708)
	at org.conscrypt.DefaultSSLContextImpl.getKeyManagers(DefaultSSLContextImpl.java:66)
	at org.conscrypt.OpenSSLContextImpl.<init>(OpenSSLContextImpl.java:81)
	at org.conscrypt.DefaultSSLContextImpl.<init>(DefaultSSLContextImpl.java:57)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
	at java.security.Provider$Service.newInstance(Provider.java:1253)
	... 56 more
```

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
kpayson64: As it turns out, this KeyStore logic was not needed.

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
kpayson64: This logic was extraneous and removed.

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
kpayson64: The key store was not being used.

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
kpayson64: Added a comment.

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
kpayson64: The key store init was throwing it, but it has now been removed.

louiscryan:
It seems sufficient to test that this code does not interfere with non-Appengine code paths. It's the responsibility of AppEngine folks to make sure that this works in their environment. The same would be true for other special environment detection that some other PaaS might want to do.

The alternative is to find some way to inject behavior, possibly by AppEngine doing their own channel provider but that has proven painful in the past with other platforms

PR #2798, opened 2017-03-07
Cut 1.2.0

carl-mastrangelo:
2 commits, as always


zhangkun83:
Since 1.2 is still weeks away, we should keep the current versions, and only bump it when we tag 1.2.0.

carl-mastrangelo:
Normally I agree, but I think that is for master, which is what people see when they clone the repo or visit the front page at github.  I was planning on bumping the recommended version numbers on master once 1.2.0 is on maven central.

carl-mastrangelo:
@zhangkun83 PTAL

ejona86:
I thought we make the README changes in a separate commit, so it can be "forward ported" to master.

PR #1811, opened 2016-05-09
Adding metadata to Status exceptions.

nmittler:
Fixes #681

Issue #681, opened 2015-07-29
Attaching Metadata to error Status responses...


dconnelly:
I'm not sure what the "best practice" is for returning application-level error information (would be very useful to understand Google's own internal practice for guidance), but certainly one approach is to attach trailing headers with application error information to error responses. Setting additional headers via thread-local using a server interceptor seems easy enough, but for client-side it seems particularly awkward capturing metadata on a per-call basis using an interceptor. I see examples in MetadataUtils for capturing "last set headers" on a stub or channel but this is only useful for testing and cannot capture on a per-call basis. Other option would be to use the async stub and a thread-local to pass the metadata information along to StreamObserver.onError() but this loses the convenience of using the blocking and future stubs. I'm not sure if the same approach works for a blocking or future stub since the listener may be invoked in a different thread. 

Will the context api once fully integrated make this easy, or would it make sense to have a way to "attach" metadata to a Status object (perhaps this would be desirable anyway for convenience)?

ejona86:
> I'm not sure what the "best practice" is for returning application-level error information (would be very useful to understand Google's own internal practice for guidance), but certainly one approach is to attach trailing headers with application error information to error responses.

Yes, that is what I would suggest, and I think that is what we would encourage. However, it may not be very-well supported in the various languages, simply because the use-case may not have been thought about. I simply don't know if it is. I do know that we discovered that it was hard to do in node.js recently and so the design has been changed to make it easier.

> Setting additional headers via thread-local using a server interceptor seems easy enough, but for client-side it seems particularly awkward capturing metadata on a per-call basis using an interceptor.

You can install a different interceptor per-call on client-side. In master, the reconfiguring of stubs is also easier since we've removed the builder pattern; you can do `stub.withInterceptors(someInterceptor)` to attach an interceptor to a particular call. Reconfiguring the stub per-call is not something to be avoided. Our design assumed that reconfiguring would be cheap enough and easy enough to permit frequent reconfiguration.

> Will the context api once fully integrated make this easy

I think the Context API would still have any problems outlined with the blocking stub. Future stub _could_ actually be fine, since the thread-local can be propagated through the executor (via `wrapDynamic()` method in #659).

> would it make sense to have a way to "attach" metadata to a Status object (perhaps this would be desirable anyway for convenience)?

I'm not against the idea, but there was push-back from @louiscryan when the idea was proposed internally. He felt that interceptors would function well in Java, and in general it was just a "nice to have."

[...]

PR #2799, opened 2017-03-07
protobuf: utility methods for com.google.rpc.Status

ericgribkoff:
This adds utility methods for attaching a `google/rpc/status.proto` protobuf object to a gRPC `Status(Runtime)Exception` on the server-side and retrieving it on the client-side. This reuses the `Status(Runtime)Exception` metadata mechanism introduced in https://github.com/grpc/grpc-java/pull/1811. 

There was an earlier discussion in https://github.com/grpc/grpc-java/issues/681 about the appropriate mechanism to attach metadata to status responses:

The design document for mapping from gRPC `Status` to `google/rpc/status.proto` suggests that the `google/rpc/status.proto` data should be regarded as a first-class member of the gRPC `Status` API.  Since `grpc-core` cannot depend on protobuf, an alternative approach to that chosen for this PR is to introduce a `io.grpc.GoogleRpcStatusWrapper` class with member variables `int code`, `String message`, `Object[] details` that maps 1-1 with `google/rpc/status.proto `. This would allow attaching `GoogleRpcStatusWrapper` as a member variable in gRPC `Status`. 

However, we would still have to jump through a number of hoops in `grpc-core` to serialize `io.grpc.GoogleRpcStatusWrapper` to protobuf in the trailing metadata and deserialize the protobuf to `io.grpc.GoogleRpcStatusWrapper` on the client side. It seemed cleaner and much lighter weight to leverage the existing metadata trailers mechanism for `Status(Runtime)Exception`s.

@a11r @yang-g This PR does not completely meet the design requirement that the proto serialized in the metadata as `grpc-status-details-bin` *must* match the returned gRPC status code. The utility functions of the PR enforce this requirement, but there is nothing stopping the user from sidestepping these utility functions altogether and arbitrarily setting the `grpc-status-details-bin` metadata key to whatever content they want. However, since `grpc-core` will never be able to validate the content of the serialized protobuf, I couldn't see a way to enforce this requirement on the wire, regardless of the choice of the utility functions here or introducing a wrapper class for `google/rpc/status.proto`.


ericgribkoff:
@carl-mastrangelo any estimate for when this will get reviewed?

lukaszx0:
@ericgribkoff Maybe unrelated, but I'm a bit confused why we seem to be adding _propriatary_ data structure to the project which the name seems to imply.

If it isn't meant to be _propriatary_, why call it `GoogleStatus*` ? Also seems to me like the `google/rpc/status.proto`, if intended to be generic. would be well suited for protobuf's well known types under rpc (or grpc) type.

ericgribkoff:
@lukaszx0 It's not intended to be proprietary. There's no need for it to have Google in the class name here, this was just a perhaps poor choice on my part to try to disambiguate it from the existing gRPC status (and other potential status types).

I've renamed it to `io.grpc.protobuf.StatusProto`, which more clearly captures it's intent - we want to have a standardized mechanism to return additional information with gRPC server responses, not necessarily just for Google APIs. The `google/rpc/status.proto` extends the existing gRPC status code and message plus a `repeated Any` details field, which can encapsulate the rich response type of any proto-based API.

lukaszx0:
Sounds good, I suspected that it was just weird naming, thanks for the update - `StatusProto`is much better imo!

ejona86:
We shouldn't expose rpc.Status's generated code. We need to depend on artifacts in Maven Central. I'm not sure such artifacts exist today.

ericgribkoff:
@ejona86 There is a maven central artifact for this. I'll update the PR. I'm assuming the same (should) apply to [`census.proto`](https://github.com/grpc/grpc-java/blob/master/services/src/main/proto/google/instrumentation/census.proto) as well?

ericgribkoff:
@carl-mastrangelo ping?

Re: ejona86's comment yesterday, I'm working out some issues with -Werror and depending on the maven artifact for rpc.Status, but that shouldn't require postponing review.

ericgribkoff:
@ejona86 I removed the status.proto and generated code, and added `com.google.api.grpc:googleapis-common-protos:0.0.3` as `libraries.google_api_protos` in `build.gradle`. This requires excluding its transitive dependencies on `io.grpc`, which is done in `protobuf/build.gradle`. I wasn't sure how to accomplish this directly in `build.gradle` to avoid the separation.

I spoke with the maintainers of the googleapis jars, and they should be able to release one without the `grpc` dependency, so handling the `exclude` elegantly will hopefully not be necessary.

protobuf/src/test/java/io/grpc/protobuf/StatusProtoTest.java
```diff
@@ -0,0 +1,161 @@
...
+  @Test
+  public void statusRuntimeException() throws Exception {
+    StatusRuntimeException sre = StatusProto.toStatusRuntimeException(STATUS_PROTO);
+    verifyStatusRuntimeException(sre);
+  }
+
+  @Test
+  public void statusRuntimeExceptionWithMetadata() throws Exception {
+    StatusRuntimeException sre = StatusProto.toStatusRuntimeException(STATUS_PROTO, metadata);
+    verifyStatusRuntimeException(sre);
```
carl-mastrangelo: newline between test and verify sections

protobuf/src/test/java/io/grpc/protobuf/StatusProtoTest.java
```diff
@@ -0,0 +1,161 @@
...
+    verifyMetadata(sre.getTrailers());
+  }
+
+  @Test
+  public void statusRuntimeExceptionWithNullMetadata() throws Exception {
+    try {
+      StatusProto.toStatusRuntimeException(STATUS_PROTO, null);
+      fail("NullPointerException expected");
+    } catch (NullPointerException expectedException) {
```
carl-mastrangelo: You can use this pattern if you want instead of ExpectedException, but you need to check the message of the exception to make sure its the one you expect.  Its too easy to catch the wrong NPE.

protobuf/src/main/java/io/grpc/protobuf/StatusProto.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  private static Metadata toMetadata(com.google.rpc.Status statusProto) {
+    Metadata metadata = new Metadata();
+    metadata.put(STATUS_DETAILS_KEY, statusProto);
+    return metadata;
+  }
+
+  private static Metadata toMetadata(com.google.rpc.Status statusProto, Metadata metadata) {
+    checkNotNull(metadata);
```
ericgribkoff: Done

protobuf/src/main/java/io/grpc/protobuf/StatusProto.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+    return null;
+  }
+
+  @Nullable
+  private static com.google.rpc.Status toStatusProto(Status status, Metadata trailers) {
+    if (trailers != null) {
+      com.google.rpc.Status statusProto = trailers.get(STATUS_DETAILS_KEY);
+      checkArgument(
+          status.getCode().value() == statusProto.getCode(),
```
ericgribkoff: It is a wire format requirement. I think that justifies failing if seen.

protobuf/src/test/java/io/grpc/protobuf/StatusProtoTest.java
ericgribkoff: Done

protobuf/src/test/java/io/grpc/protobuf/StatusProtoTest.java
ericgribkoff: Done

ericgribkoff:
@ejona86 I spoke again with the google API folks. It turns out `googleapis-common-protos` has been deprecated in favor of `grpc-google-common-protos`, and this actually does include grpc-generated code. They are not planning to release an artifact containing `rpc.Status` that does not depend on `grpc`. I've updated the PR to the correct dependency with the necessary exclude statements.

@

protobuf/src/test/java/io/grpc/protobuf/StatusProtoTest.java
```diff
@@ -57,60 +57,80 @@ public void setup() {
...
+      assertEquals("metadata must not be null", npe.getMessage());
+    }
+  }
+
+  @Test
+  public void toStatusRuntimeException_shouldThrowIfStatusCodeInvalid() throws Exception {
+    try {
+      StatusProto.toStatusRuntimeException(INVALID_STATUS_PROTO);
+      fail("IllegalArgumentException expected");
+    } catch (IllegalArgumentException expectedException) {
```
carl-mastrangelo: check message for code being invalid?

ericgribkoff: Done, will merge on green.

PR #2805, opened 2017-03-08
core: annotate keys with Immutable

dapengzhang0:


core/src/main/java/io/grpc/Attributes.java
```diff
@@ -85,6 +85,7 @@ public static Builder newBuilder() {
     return new Builder();
   }
 
+  @Immutable
   public static final class Key<T> {
```
lukaszx0: nit: Since it's a public class, maybe worth adding short javadoc?

core/src/main/java/io/grpc/Metadata.java
```diff
@@ -490,6 +491,7 @@ private boolean bytesEqual(byte[] left, byte[] right) {
   }
 
   /** Marshaller for metadata values that are serialized into raw binary. */
+  @Immutable
```
carl-mastrangelo: I can imagine some case where a marshaller might want to be mutable.  For example, the protobuf marshaller may want to cache a copy of the encoding settings.  

What's the reason for wanting to make these things immutable?

dapengzhang0: The metadata is supposed to be transmitted between server and client, it might be unsafe or errorprone if the marshaller is able to change settings over time. It make sense in some scenario, the marshaller might need be stateful, but it may be rare and it's better to be defensive.

carl-mastrangelo: It is good to be defensive, but in this case we really cannot enforce it.  I liken `@Immutable` to a promise, rather than a suggestion.  We can add it to classes we own, but can't enforce it for others.

dapengzhang0:
question: should `io.grpc.Context.Key` be a `final` class? should it be `@Immutable`?

carl-mastrangelo:
It should be final, though probably not immutable.  The default value is not deeply immutable.

PR #2806, opened 2017-03-08
test: bump junit to 4.12

dapengzhang0:


carl-mastrangelo:
Can't be done, as it won't match the internal version.

dapengzhang0:
What is the internal version? I though it is even later than 4.12

carl-mastrangelo:
We are stuck at 4.11 for the indefinite future.

dapengzhang0:
I see. JUNIT_VERSION = "v4_11" is currently used. thanks @carl-mastrangelo

PR #2809, opened 2017-03-09
Do not merge: test flakiness on travis

dapengzhang0:


dapengzhang0:
Ran the test for 30000 times, all passed, not able to reproduce the flake on travis linux env with this approach.

PR #2812, opened 2017-03-09
grpclb: set LB tokens to headers.

zhangkun83:
The balancer service attaches a token string for each `Server` entry it
sends to the client.  The client has to set the token to the `"lb-token"`
header when assigning that Server entry to an RPC.

For convenience of testing, also implemented `hashCode()` and `equals()` for
`PickResult`.

cc @ejona86 for the changes in `PickResult`.

### DIFF
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -31,6 +31,7 @@
 
 package io.grpc;
 
+import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import java.util.List;
 import javax.annotation.Nullable;
@@ -346,6 +347,20 @@ public abstract class LoadBalancer {
     @Override
     public String toString() {
       return "[subchannel=" + subchannel + " status=" + status + "]";
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hashCode(subchannel, status);
+    }
+
+    @Override
+    public boolean equals(Object other) {
+      if (!(other instanceof PickResult)) {
+        return false;
+      }
+      PickResult that = (PickResult) other;
+      return Objects.equal(subchannel, that.subchannel) && Objects.equal(status, that.status);
     }
   }
 
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -204,7 +204,7 @@ public abstract class LoadBalancer {
     public abstract CallOptions getCallOptions();
 
     /**
-     * Call metadata.
+     * Headers of the call. {@code pickSubchannel()} may mutate it before before returning.
      */
     public abstract Metadata getHeaders();
 
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -31,6 +31,7 @@
 
 package io.grpc;
 
+import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -476,24 +484,74 @@ public PickResult pickSubchannel(PickSubchannelArgs args) {
...
     public PickResult pickSubchannel(PickSubchannelArgs args) {
       synchronized (list) {
-        PickResult result = list.get(index);
+        RoundRobinEntry result = list.get(index);
         index++;
         if (index == list.size()) {
           index = 0;
         }
-        return result;
+        result.updateHeaders(args.getHeaders());
```
dapengzhang0: this is counter-intuitive: `public PickResult pickSubchannel(PickSubchannelArgs args)` will produce a PickResult which does not depend on the args at all, but will mutate the args.

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -347,6 +348,20 @@ public Status getStatus() {
     public String toString() {
       return "[subchannel=" + subchannel + " status=" + status + "]";
```
zhangkun83: Done.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -476,24 +484,74 @@ public PickResult pickSubchannel(PickSubchannelArgs args) {
   }
 
   @VisibleForTesting
+  static final class RoundRobinEntry {
+    final PickResult result;
```
zhangkun83: `result` is actually used in test. Made `token` private.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -476,24 +484,74 @@ public PickResult pickSubchannel(PickSubchannelArgs args) {
...
+    void updateHeaders(Metadata headers) {
+      if (token != null) {
+        headers.discardAll(TOKEN_KEY);
+        headers.put(TOKEN_KEY, token);
+      }
+    }
+
+    @Override
+    public String toString() {
+      return "[result=" + result + ", token=" + token + "]";
```

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -476,24 +484,74 @@ public PickResult pickSubchannel(PickSubchannelArgs args) {
...
+      }
+      RoundRobinEntry that = (RoundRobinEntry) other;
+      return Objects.equal(result, that.result) && Objects.equal(token, that.token);
+    }
+  }
+
+  @VisibleForTesting
   static final class RoundRobinPicker extends SubchannelPicker {
-    final List<PickResult> list;
+    final List<RoundRobinEntry> list;
```
zhangkun83: `list` is actually used in test. Made `index` private.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
zhangkun83: This is intended. The `args` doesn't only pass information to the picker, but also pass mutable objects like headers for the picker to mutate. You don't see such pattern very often because most objects in gRPC API are immutable, while `Metadata` is an exception. Whenever it comes to changing `Metadata`, it is passed in as argument instead of being returned. Another example is `CallCredentials.MetadataApplier.apply(Metadata headers)`.

dapengzhang0:
@zhangkun83  replied in the first thread.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
zhangkun83: Added comments to `LoadBalancer.PickSubchannelArgs.getHeaders()`.

Issue #2775, opened 2017-03-02
SSL error, pls help


linking12:
### What version of gRPC are you using?
Grpc version 1.0.1

### What JVM are you using (`java -version`)?
JVM: Oracle Jdk jdk-8u121

### What did you do?
```
 private SslContext buildServerSslContext() {
        try {
            InputStream certs = SslUtil.loadInputStreamCert("server.pem");
            InputStream keys = SslUtil.loadInputStreamCert("server_pkcs8.key");
            return GrpcSslContexts.configure(SslContextBuilder.forServer(certs, keys)).build();
        } catch (SSLException e) {
            throw new RpcFrameworkException(e);
        }
    }
```
```
 private SslContext buildClientSslContext() {
        try {
            InputStream certs = SslUtil.loadInputStreamCert("server.pem");
            return GrpcSslContexts.configure(SslContextBuilder.forClient()//
                                                              .trustManager(certs))//
                                  .build();
        } catch (SSLException e) {
            throw new RpcFrameworkException(e);
        }
    }
```
In client,use jetty alpn
```
-javaagent:/Users/shimingliu/.m2/repository/org/mortbay/jetty/alpn/jetty-alpn-agent/2.0.6/jetty-alpn-agent-2.0.6.jar
```

In server, use netty boringssl
```
             <dependency>
			<groupId>io.netty</groupId>
			<artifactId>netty-tcnative-boringssl-static</artifactId>
			<version>${netty-tcnative.version}</version>
			<classifier>${os.detected.classifier}</classifier>
		</dependency>
```
### What did you expect to see?
There is no error

### What did you see instead?
```
[jetty-alpn-agent] Replacing: sun/security/ssl/ClientHandshaker
[jetty-alpn-agent] Replacing: sun/security/ssl/HelloExtensions
[jetty-alpn-agent] Replacing: sun/security/ssl/ClientHandshaker$2
io.grpc.StatusRuntimeException: UNKNOWN
	at io.grpc.Status.asRuntimeException(Status.java:536)
	at com.quancheng.saluki.core.grpc.client.async.AbstractRetryingRpcListener.onClose(AbstractRetryingRpcListener.java:86)
	at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:56)
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.close(ClientCallImpl.java:481)
[...]

PR #2816, opened 2017-03-13
all: Bump Jetty ALPN agent version, for new JREs

ejona86:
Jetty ALPN was broken with jdk1.8.0_121


dapengzhang0:
is it tested with the new JRE? An issue was reported #2775

ejona86:
Yes, this fixed jdk1.8.0_121 for me.

PR #2818, opened 2017-03-14
Backport SETTINGS ACK sequencing fix

ejona86:
Reviewing each commit separately is easiest. I'll leave the two commits as separate when I merge to master.

The netty/third_party stuff is a clone of what was done in okhttp/third_party.


ejona86:
I have manually tested this fixes interop with C#. Since Netty added a test and we're going to revert to using their code once we upgrade, I didn't feel like it was worth-while to add a test in grpc.

PR #2821, opened 2017-03-14
Move okhttp to AbstractStream2

zsurocking:
One thing worth noting is that I overrided onStreamDeallocated to get notification when transportReportStatus is called and stream is considered finished. Then I can remove the stream from transport map and also reset the stream if the framer is not yet closed.

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractClientStream2.java
+++ b/core/src/main/java/io/grpc/internal/AbstractClientStream2.java
@@ -190,7 +190,6 @@ public abstract class AbstractClientStream2 extends AbstractStream2
 
     private Runnable deliveryStalledTask;
 
-    private boolean headersReceived;
     /**
      * Whether the stream is closed from the transport's perspective. This can differ from {@link
      * #listenerClosed} because there may still be messages buffered to deliver to the application.
@@ -233,7 +232,6 @@ public abstract class AbstractClientStream2 extends AbstractStream2
      */
     protected void inboundHeadersReceived(Metadata headers) {
       Preconditions.checkState(!statusReported, "Received headers on closed stream");
-      headersReceived = true;
       listener().headersRead(headers);
     }
 
@@ -248,12 +246,6 @@ public abstract class AbstractClientStream2 extends AbstractStream2
       try {
         if (statusReported) {
           log.log(Level.INFO, "Received data on closed stream");
-          return;
-        }
-        if (!headersReceived) {
-          transportReportStatus(
-              Status.INTERNAL.withDescription("headers not received before payload"),
-              false, new Metadata());
           return;
         }
 
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
@@ -38,9 +38,9 @@ import io.grpc.Attributes;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
 import io.grpc.Status;
-import io.grpc.internal.ClientStreamListener;
+import io.grpc.internal.AbstractClientStream2;
 import io.grpc.internal.GrpcUtil;
-import io.grpc.internal.Http2ClientStream;
+import io.grpc.internal.Http2ClientStreamTransportState;
 import io.grpc.internal.StatsTraceContext;
 import io.grpc.internal.WritableBuffer;
 import io.grpc.okhttp.internal.framed.ErrorCode;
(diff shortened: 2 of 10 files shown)


okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -54,38 +55,27 @@
...
-  @GuardedBy("lock")
-  private int window = Utils.DEFAULT_WINDOW_SIZE;
-  @GuardedBy("lock")
-  private int processedWindow = Utils.DEFAULT_WINDOW_SIZE;
+  public static final int ABSENT_ID = -1;
+
   private final MethodDescriptor<?, ?> method;
   /** {@code null} iff start has been called. */
   private Metadata headers;
   private final AsyncFrameWriter frameWriter;
```
ejona86: Move this to transport state? While "thread-safe", any usages outside of the lock are questionable. And today there aren't any usages outside of the lock.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -54,38 +55,27 @@
...
   private final OutboundFlowController outboundFlow;
```
ejona86: Move to transport state? It doesn't seem thread-safe, so it should only be called with the lock held.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -54,38 +55,27 @@
...
   private final OkHttpClientTransport transport;
```
ejona86: This looks like it would be good to move to transport state as well. Having a method in transport state would avoid mutating requestHeaders from TransportState as well.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -148,112 +139,163 @@ public void start(ClientStreamListener listener) {
...
+        buffer = ((OkHttpWritableBuffer) frame).buffer();
+        int size = (int) buffer.size();
+        if (size > 0) {
+          onSendingBytes(size);
         }
       }
-      if (flush) {
-        outboundFlow.flush();
+
+      synchronized (state.lock) {
```
ejona86: Move the logic in this block to transport state?

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -148,112 +139,163 @@ public void start(ClientStreamListener listener) {
...
-        pendingData.add(new PendingData(buffer, endOfStream, flush));
+    @GuardedBy("lock")
+    @Override
+    protected void onStreamAllocated() {
+      super.onStreamAllocated();
+    }
+
+    @GuardedBy("lock")
+    @Override
+    protected void onStreamDeallocated() {
```
ejona86: This is called by `transportReportStatus`. `transportReportStatus` is intended for the transport to notify that a stream has gone away. It worries me to have it implicitly close the stream, because that is hard to see and is different from netty. Is `cancel` the only code that would need to be fixed if this code was removed?

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
zsurocking: Done

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
zsurocking: Done

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
zsurocking: Done

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
zsurocking: Done

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
zsurocking: The problem is that transportReportStatus gets called in in two places:
1) AbstractClientStream2 line 252(headers not received before payload)
2) Http2ClientStreamTransportState line 156(trailer is not received)

Those two calls are not exactly from transport. The abstract layer finds out violation, reports the status and closes the stream.
And transport is not aware of these and don't know the stream gets closed. How does netty handle this?

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -387,11 +385,11 @@ public void run() {
           }
           clientFrameHandler = new ClientFrameHandler(testFrameReader);
           executor.execute(clientFrameHandler);
+          frameWriter.becomeConnected(testFrameWriter, socket);
```
zsurocking: It should be called before startPendingStreams() below which will trigger sending.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
ejona86: The call from `AbstractClientStream2` looks like it should be a call to `http2ProcessingFailed(Status, Metadata)` instead.

The call from `onStreamDeallocated` is only triggered when `endOfStream == true`. So the stream removal will already be handled by `OkHttpClientStream`.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
ejona86: Why is triggering sending a problem? Everything should be queued behind this currently-running Runnable.

zsurocking: Reverted. I thought this was the cause to one test failure.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
zsurocking: AbstractClientStream2 cannot call http2ProcessingFailed which is http2 specific and only in Http2ClientStreamTransportState.
I can move the logic of making sure headers received before payload to Http2ClientStreamTransportState, and then we can call http2ProcessingFailed. What do you think.

For the second one, yes, I can check endOfStream == true and call transportReportStatus in OkhttpClientStream.

ejona86: > I can move the logic of making sure headers received before payload to Http2ClientStreamTransportState

Yeah. That sounds good.

zsurocking: Done. PTAL

core/src/main/java/io/grpc/internal/AbstractClientStream2.java
```diff
@@ -248,12 +248,6 @@ protected void inboundDataReceived(ReadableBuffer frame) {
           log.log(Level.INFO, "Received data on closed stream");
           return;
         }
-        if (!headersReceived) {
```
ejona86: `headersReceived` is now unused. Either delete it, or turn this into an `IllegalStateException` (so that we get a clear error that a class is mis-implemented).

zsurocking: Deleted

PR #2822, opened 2017-03-14
core: keepaliveManager not to use Ping.onSuccess and ragard onDataReceive as ping Ack

dapengzhang0:
Preparing to support server side keepalive.
For the convience on server side, not to use Ping `onSuccess()` callback to cancle shutdownFuture any more, instead, regard `onDataReceived()` as ping Ack and cancel shutdownFuture in it.

This is a part of the larger PR #2778

### DIFF
--- a/core/src/main/java/io/grpc/internal/KeepAliveManager.java
+++ b/core/src/main/java/io/grpc/internal/KeepAliveManager.java
@@ -176,7 +176,7 @@ public class KeepAliveManager {
       // schedule a new ping
       state = State.PING_SCHEDULED;
       pingFuture =
-          scheduler.schedule(sendPing, nextKeepaliveTime - ticker.read(), TimeUnit.NANOSECONDS);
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/KeepAliveManager.java
```diff
@@ -162,6 +162,21 @@ public synchronized void onDataReceived() {
...
+      }
+      if (state == State.IDLE_AND_PING_SENT) {
+        // not to schedule new pings until onTransportActive
+        state = State.IDLE;
+        return;
+      }
+      // schedule a new ping
+      state = State.PING_SCHEDULED;
+      pingFuture =
+          scheduler.schedule(sendPing, nextKeepaliveTime - ticker.read(), TimeUnit.NANOSECONDS);
```
ejona86: Do keepAliveDelayInNanos, like was done before?

PR #2825, opened 2017-03-15
core: Support keepalive even when transport is idle

ejona86:
Nothing is using this yet, but it will be used on both client and
server.

### DIFF
--- a/core/src/main/java/io/grpc/internal/KeepAliveManager.java
+++ b/core/src/main/java/io/grpc/internal/KeepAliveManager.java
@@ -49,6 +49,7 @@ public class KeepAliveManager {
   private final ScheduledExecutorService scheduler;
   private final ManagedClientTransport transport;
   private final Ticker ticker;
+  private final boolean keepAliveDuringTransportIdle;
   private State state = State.IDLE;
   private long nextKeepaliveTime;
   private ScheduledFuture<?> shutdownFuture;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/grpc/internal/KeepAliveManager.java
```diff
@@ -131,27 +133,34 @@ public void run() {
...
     this.scheduler = Preconditions.checkNotNull(scheduler, "scheduler");
     this.ticker = Preconditions.checkNotNull(ticker, "ticker");
     this.keepAliveDelayInNanos = keepAliveDelayInNanos;
     this.keepAliveTimeoutInNanos = keepAliveTimeoutInNanos;
+    this.keepAliveDuringTransportIdle = keepAliveDuringTransportIdle;
     nextKeepaliveTime = ticker.read() + keepAliveDelayInNanos;
   }
 
+  /** Start keepalive monitoring. */
+  public synchronized void start() {
```
dapengzhang0: I don't think it is necessary. On server side, `channelActive` is always called at the beginning, we can put an `onTransportActive` there.

ejona86: `onTransportActive` is right now meaning there are streams. So one active != the other active. Renaming to `onTransportStarted`

PR #2826, opened 2017-03-15
core,netty,okhttp: KeepAliveManager with Pinger

dapengzhang0:
Modified KeepAliveManager to use a Pinger interface, which can send ping or shutdown transport for both server and client.

### DIFF
--- a/core/src/main/java/io/grpc/internal/KeepAliveManager.java
+++ b/core/src/main/java/io/grpc/internal/KeepAliveManager.java
@@ -31,8 +31,9 @@
 
 package io.grpc.internal;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.MoreExecutors;
 import io.grpc.Status;
 import java.util.concurrent.ScheduledExecutorService;
@@ -47,8 +48,8 @@ public class KeepAliveManager {
   private static final long MIN_KEEPALIVE_DELAY_NANOS = TimeUnit.MINUTES.toNanos(1);
 
   private final ScheduledExecutorService scheduler;
-  private final ManagedClientTransport transport;
   private final Ticker ticker;
+  private final KeepAlivePinger keepAlivePinger;
   private final boolean keepAliveDuringTransportIdle;
   private State state = State.IDLE;
   private long nextKeepaliveTime;
@@ -67,8 +68,7 @@ public class KeepAliveManager {
         }
       }
       if (shouldShutdown) {
-        transport.shutdownNow(Status.UNAVAILABLE.withDescription(
-            "Keepalive failed. The connection is likely gone"));
+        keepAlivePinger.onPingTimeout();
       }
     }
   };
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/grpc/internal/KeepAliveManager.java
```diff
@@ -234,18 +233,57 @@ public synchronized void onTransportShutdown() {
...
+    /**
+     * Callback when Ping Ack was not received in KEEPALIVE_TIMEOUT. Should shutdown the transport.
+     */
+    void onPingTimeout();
+
+    /**
+     * Specifies whether the pinger should keep sending keep-alive pings when the transport has no
+     * active RPCs.
+     */
+    boolean stillPingWhileIdle();
```
ejona86: Why is this a method instead of staying as a field? This must always return the same value, otherwise the code may break.

I don't understand what is special about this one configuration option such that it would be added to the Pinger.

core/src/test/java/io/grpc/internal/KeepAliveManagerTest.java
```diff
@@ -71,8 +73,8 @@ public long read() {
   @Before
   public void setUp() {
     MockitoAnnotations.initMocks(this);
-    keepAliveManager = new KeepAliveManager(transport, scheduler, ticker, 1000, 2000, false);
-    keepAliveManager.onTransportStarted();
+    keepAlivePinger = new ClientKeepAlivePinger(transport);
```
ejona86: Could you use a mock for the pinger for the majority of the tests? Then you can have a few tests to test the ClientKeepAlivePinger by itself, and those could mock the transport.

core/src/test/java/io/grpc/internal/KeepAliveManagerTest.java
```diff
@@ -199,13 +204,8 @@ public void keepAlivePingFailsAfterItTimesOut() {
 
     // We do not receive the ping response. Shutdown runnable runs.
     ticker.time = 3000;
-    shutdown.run();
+    keepAlivePinger.onPingTimeout();
```
ejona86: Whoa. This needs to be `shutdown.run()`, otherwise it isn't really testing much.

core/src/main/java/io/grpc/internal/KeepAliveManager.java
dapengzhang0: stayed as a field

core/src/test/java/io/grpc/internal/KeepAliveManagerTest.java
dapengzhang0: done

core/src/test/java/io/grpc/internal/KeepAliveManagerTest.java
dapengzhang0: done

dapengzhang0:
PTAL

core/src/test/java/io/grpc/internal/KeepAliveManagerTest.java
```diff
@@ -142,6 +141,8 @@ public void keepAlivePingDelayedByIncomingData() {
 
   @Test
   public void keepAlivePingFails() {
+    keepAliveManager = new KeepAliveManager(
+        new ClientKeepAlivePinger(transport), scheduler, ticker, 1000, 2000, false);
```
ejona86: Continue using pinger in this and similar tests. Only use `ClientKeepAlivePinger` in tests testing `ClientKeepAlivePinger`.

Basically you'd have about two new tests to test `ClientKeepAlivePinger` and nothing else would use it.

dapengzhang0: done

PR #2831, opened 2017-03-16
netty: fix KeepAliveManager not apply to ConnectionHandler

### LINKED ISSUE #2828: KeepAliveManager is not apply to ConnectionHandler in netty transport
GRPC 1.2.0

1.8.0_112

### What did you do?
I am try to make `KeepAlive` work with netty transport just now(grpc-java 1.2.0), and after some function test, i think the `KeepAlive` feature still not work functionally.

Debugging the code(io.grpc.netty.NettyClientTransport#start):
image

So, we known the `KeepAliveManager` only make effective in `io.grpc.netty.NettyClientHandler` instance. But the code showing the creation of `io.grpc.netty.NettyClientHandler` always receive a `null` keepAliveManager.

image

Probably because the `keepAliveManager` field is assigning in the end of `io.grpc.netty.NettyClientTransport#start()` call.

Associated commit: https://github.com/grpc/grpc-java/pull/2729: the commit resolve the problem of NPE(https://github.com/grpc/grpc-java/issues/2726)

cc @lukaszx0 @ejona86

dapengzhang0:
resolves #2828

It is awkward that the construction of Channel depends on NettyClientHandler, that of NettyClientHandler depends on KeepAliveManager, and that of KeepAliveManager depends on Channel.

- moved construction of keepAliveManager right after channel is constructed.
- pass keepAliveManager into NettyClientHandler by a setter instead of by the constructor of NettyClientHandler

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
@@ -31,11 +31,12 @@
 
 package io.grpc.netty;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
 import static io.netty.handler.codec.http2.DefaultHttp2LocalFlowController.DEFAULT_WINDOW_UPDATE_RATIO;
 import static io.netty.util.CharsetUtil.UTF_8;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
 import com.google.common.base.Stopwatch;
 import com.google.common.base.Ticker;
 import io.grpc.Attributes;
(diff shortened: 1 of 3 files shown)


dapengzhang0:
As discussed with @ejona, there are some alternative solutions:
- pass the attributes `transport/pinger`, `enableKeepAlive`, `keepAliveDelayNanos`, and `keepAliveTimeoutNanos` to the constructor of `NettyClientHandler`, and instantiate `keepAliveManager` at `handlerAdded()`
- not to pass `executorService` to the constructor of `KeepAliveManager`, instead, pass that to `onTransportStarted()`; and call `onTransportStarted(executor)` at `handlerAdded()`.
- there might be more alternatives but need dig deeper under the surface of netty.

PR #2833, opened 2017-03-17
core: use nanos more consistently

carl-mastrangelo:
Rather than use a mix of millis and nanos, let's just use nanos throughout.  We can convert to millis on API boundaries (like with Census).  Many Java APIs already speak Nanos, so we should too.


core/src/test/java/io/grpc/internal/FakeClock.java
```diff
@@ -266,8 +266,8 @@ public int forwardTime(long value, TimeUnit unit) {
    *
```
zhangkun83: Update the unit in the comment.

carl-mastrangelo: done.

PR #2834, opened 2017-03-17
netty: move startWriteQueue right after channel is constructed

dapengzhang0:
As 65e4d9f has split the channel instantiation and `connect()`, we can `startWriteQueue()` even earlier.


zhangkun83:
Does this solve any problem?

dapengzhang0:
It is commented 

> // Start the write queue as soon as the channel is constructed

otherwise the comment is not valid.

The benefit is that it may avoid a benign race that `channel.connect()` and then `transportReady()` can finish too soon, so that `startWriteQueue()` is not yet executed. The change guarantees writeQueue is available before `connect()`.

PR #2835, opened 2017-03-17
all: update to latest version of errorprone

carl-mastrangelo:


protobuf-lite/build.gradle
```diff
@@ -28,6 +28,11 @@ compileTestJava {
         "-Xep:MissingOverride:OFF", "-Xep:ReferenceEquality:OFF"]
 }
 
+[compileJava, compileTestJava].each() {
+  // Protolite isn't marked @Generated, but it should still be treated as such.
+  it.options.compilerArgs += ["-Xep:FallThrough:OFF"]
```
ejona86: Move this to the compileTestJava above? It seems you are only wanting it for the generated code, and we only generate code for the test.

carl-mastrangelo: Done.

Issue #2829, opened 2017-03-16
Define Context semantics on client-side


zhangkun83:
`Context`'s semantics is quite clear on the server-side. For each call:
 1. `ServerImpl` creates the base `Context`.
 2. `ServerInterceptor`s can create new `Context`s, which can be based on the base `Context`. The last created `Context` is what the server application sees.
 3. Server application access the `Context` from thread-local. And if it initiates outgoing calls, by default they inherit the `Context`.

For the client-side, currently the `Context` semantics is:
 1. The current `Context` when `Channel.newCall()` is called is the base `Context`.
 2. `ClientInterceptor`s create new `Context`s, which can be based on the base `Context`. The last created `Context` is what `ClientCallImpl` uses
 3. `ClientCall.Listener` is called under the `ClientCallImpl`'s `Context`.

If a new call is made from the `ClientCall.Listener`, it will inherit the previous call's `Context`. If this chain of calls is unbounded, and every call goes through interceptor(s) that create `Context` based off the current `Context`, this will create an unbounded inheritance chain of `Contexts`. Because `Context` always keeps a reference to the parent `Context`, this will lead to memory leak.

`Context` is designed to work like a stack. The expectation is that a `Context` (except the background one) should eventually be popped out of the "stack". Infinite pushing without popping, like the case from previous paragraph, should not happen. IMO, this means `Context` should only be passed in one direction. On the client-side, the direction should be **application -> interceptors -> gRPC core**. Since `ClientCall.Listener` is the other direction, it should not be called under the top-most `Context`, but rather the `Context` that application sees, which is the pre-interceptors one.

zhangkun83:
Another argument for calling the Call listeners under the original Context, is that this makes async call Context semantics consistent with blocking calls, because blocking calls handle the response in the original Context, not the interceptor-mutated Context.

zhangkun83:
Turns out running Call listeners under the original Context is not feasible, because interceptors are installed by wrapping the channel, thus `ManagecChannelImpl` cannot access the original Context under which the outmost `newCall()` is called. In @ejona86's opinion, the concept of Context doesn't go along with callbacks, and we don't have a good solution. The best we can do right now, is to document that ClientInterceptors should not override the current Context.

PR #2838, opened 2017-03-17
doc: suggest against overriding Context in ClientInterceptor

zhangkun83:
Reference: #2829

### DIFF
--- a/core/src/main/java/io/grpc/ClientInterceptor.java
+++ b/core/src/main/java/io/grpc/ClientInterceptor.java
@@ -53,6 +53,10 @@ public interface ClientInterceptor {
    * around the result of {@code next.newCall()}, whereas a simpler implementation may just modify
    * the header metadata prior to returning the result of {@code next.newCall()}.
    *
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/grpc/ClientInterceptor.java
```diff
@@ -53,6 +53,10 @@
    * around the result of {@code next.newCall()}, whereas a simpler implementation may just modify
    * the header metadata prior to returning the result of {@code next.newCall()}.
    *
+   * <p>{@code next.newCall()} <strong>must not</strong> be called under a different {@link Context}
+   * other than the current {@code Context}. The outcome of such usage is undefined and may cause
```
ejona86: It seems well-defined to me; it just seems like it is well-defined to cause problems.

zhangkun83: It's just defined by the implementation, but not what we really want. Calling it undefined means we don't support whatever the current implementation is doing and we are free to change it in the future.

PR #2842, opened 2017-03-20
services: update monitoring.proto

ericgribkoff:
We filled in some of the missing type definitions in `monitoring.proto`. One of these, `StatsResponse`, is the same as the nested type `CanonicalRpcStats.View`. This PR drops the now-redundant `CanonicalRpcStats.View` and replaces its use with the new `StatsResponse`. 

This is a breaking change to existing use of the monitoring service, but we have it marked as experimental/v1alpha so I believe this is fine. If not, I will leave in `CanonicalRpcStats.View` with a todo to remove it in a later version.

### DIFF
--- a/services/src/main/proto/grpc/instrumentation/v1alpha/monitoring.proto
+++ b/services/src/main/proto/grpc/instrumentation/v1alpha/monitoring.proto
@@ -50,8 +50,6 @@ service Monitoring {
 
   // Query the server for specific stats
   rpc GetStats(StatsRequest) returns (StatsResponse) {
-    // TODO(aveitch, ericgribkoff): Pease define the stats response message
-    // StatsRequest would specifically identify the stats to be returned.
   }
 
   // Request the server to stream back snapshots of the requested stats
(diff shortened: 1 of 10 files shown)


services/src/main/proto/grpc/instrumentation/v1alpha/monitoring.proto
```diff
@@ -76,39 +74,67 @@ service Monitoring {
...
+  // response. If view_names is also provided, then Views matching the
+  // intersection of the two are returned.
+  // TODO(aveitch): Consider making this a list of regexes or prefix matches in
+  // the future.
+  repeated string measurement_names = 2;
+
+  // By default, the MeasurementDescriptors and ViewDescriptors corresponding to
+  // the Views that are returned in a StatsResponse will be included in the
+  // first such response. Set this to true to have them not sent.
+  bool dont_include_descriptors_in_first_response = 3;
```
ericgribkoff: This is coming from https://github.com/grpc/grpc-proto/pull/3 as the source of truth, and is just copied here.

PR #2853, opened 2017-03-23
core,benchmarks: use Atomics for StatsTraceContext

carl-mastrangelo:
This removes a needless warning, and isn't much slower.  Also this
includes a benchmark for StatsTraceContext to measure the overhead
for creation.  It adds about 40ns per RPC.  Optimization will come
after structural changes are made to break the dependency on
Census.

### DIFF
--- a/core/src/main/java/io/grpc/internal/StatsTraceContext.java
+++ b/core/src/main/java/io/grpc/internal/StatsTraceContext.java
@@ -49,6 +49,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * The stats and tracing information for a call.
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/StatsTraceContext.java
```diff
@@ -255,18 +257,19 @@ void callEnded(Status status) {
...
+        .put(wireBytesReceivedMetric, wireBytesReceived.get())
+        .put(uncompressedBytesSentMetric, uncompressedBytesSent.get())
+        .put(uncompressedBytesReceivedMetric, uncompressedBytesReceived.get());
     if (!status.isOk()) {
       builder.put(errorCountMetric, 1.0);
     }
     if (side == Side.CLIENT) {
-      if (clientPendingNanos >= 0) {
+      long localClientPendingNanos;
+      if ((localClientPendingNanos = clientPendingNanos.get()) != UNSET_CLIENT_PENDING_NANOS) {
```
zhangkun83: Would it be more natural to move the assignment up a line?

carl-mastrangelo: Good point, done.

ejona86:
@carl-mastrangelo, given that we have a StatsTraceContext for each RPC, do we want to use `AtomicLongFieldUpdater` instead of `AtomicLong`?

carl-mastrangelo:
Do you mean for saving allocations?

ejona86:
To remove allocations and make not change performance compared to the original at all.

PR #2854, opened 2017-03-24
netty: Server side keep alive

dapengzhang0:
use KeepAliveManager in NettyServerHandler

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
@@ -565,8 +565,20 @@ class NettyServerHandler extends AbstractNettyHandler {
 
     @Override
     public void onPingTimeout() {
-      ctx.channel().writeAndFlush(new ForcefulCloseCommand(
-          Status.UNAVAILABLE.withDescription("Keepalive failed. The connection is likely gone")));
+      try {
+        forcefulClose(
+            ctx,
+            new ForcefulCloseCommand(Status.UNAVAILABLE
+                .withDescription("Keepalive failed. The connection is likely gone")),
+            ctx.newPromise());
+      } catch (Exception ex) {
+        try {
+          exceptionCaught(ctx, ex);
+        } catch (Exception ex2) {
+          logger.log(Level.WARNING, "Exception while propagating exception", ex2);
+          logger.log(Level.WARNING, "Original failure", ex);
+        }
+      }
     }
   }
 }
--- a/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
@@ -300,6 +300,16 @@ class NettyServerHandler extends AbstractNettyHandler {
     }
   }
 
+  @VisibleForTesting
+  KeepAliveManager getKeepAliveManagerForTest() {
+    return keepAliveManager;
+  }
+
+  @VisibleForTesting
+  void setKeepAliveManagerForTest(KeepAliveManager keepAliveManager) {
+    this.keepAliveManager = keepAliveManager;
+  }
+
   /**
    * Handler for the Channel shutting down.
    */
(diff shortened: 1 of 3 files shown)


netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -492,5 +528,46 @@ public void onPingAckRead(ChannelHandlerContext ctx, ByteBuf data) throws Http2E
...
+      ByteBuf buf = unreleasableBuffer(directBuffer(8).writeLong(0xDEADL));
+      encoder().writePing(ctx, false /* isAck */, buf, ctx.newPromise()).addListener(
+          new ChannelFutureListener() {
+            @Override
+            public void operationComplete(ChannelFuture future) throws Exception {
+              if (!future.isSuccess() && !future.isCancelled()) {
+                shutdownTransportNow();
+              }
+            }
+          });
```
dapengzhang0: will remove future listener

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -492,5 +538,46 @@ public void onPingAckRead(ChannelHandlerContext ctx, ByteBuf data) throws Http2E
...
+  private final class KeepAlivePinger implements KeepAliveManager.KeepAlivePinger {
+    final ChannelHandlerContext ctx;
+
+    KeepAlivePinger(ChannelHandlerContext ctx) {
+      this.ctx = ctx;
+    }
+
+    @Override
+    public void ping() {
+      ByteBuf buf = unreleasableBuffer(directBuffer(8).writeLong(0xDEADL));
```
dapengzhang0: static final

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
dapengzhang0: done

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
dapengzhang0: done

dapengzhang0:
PTAL

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -492,5 +540,33 @@ public void onPingAckRead(ChannelHandlerContext ctx, ByteBuf data) throws Http2E
...
+
+    @Override
+    public void ping() {
+      encoder().writePing(ctx, false /* isAck */, KEEPALIVE_PING_BUF, ctx.newPromise());
+      ctx.flush();
+    }
+
+    @Override
+    public void onPingTimeout() {
+      ctx.channel().writeAndFlush(new ForcefulCloseCommand(
```
ejona86: This is strange, since we're already in the handler. Just call forcefulClose directly?

netty/src/main/java/io/grpc/netty/NettyServerTransport.java
```diff
@@ -58,6 +59,9 @@
   private final int flowControlWindow;
   private final int maxMessageSize;
   private final int maxHeaderListSize;
+  // TODO(zdapeng): allow custom keep alive config values by NettyServerBuilder
+  private final long keepAliveTimeInNanos = TimeUnit.NANOSECONDS.convert(2L, TimeUnit.HOURS);
```
ejona86: Can you default to infinite for this PR, so the feature is disabled? Once configuration is available, then we can turn it on (so if there are problems, there is a way to disable it).

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
dapengzhang0: how to handle the checked Exception of forcefulClose?

ejona86: I ended up doing something like this:

```java
try {
  forcefulClose(ctx, new ForcefulCloseCommand(status), ctx.newPromise());
} catch (Http2Exception ex) {
  throw ex;
} catch (RuntimeException ex) {
  throw ex;
} catch (Exception ex) {
  try {
    exceptionCaught(ctx, ex);
  } catch (Exception ex2) {
    logger.log(Level.WARNING, "Exception while propagating exception", ex2);
    logger.log(Level.WARNING, "Original failure", ex);
  }
}
```

dapengzhang0:
jenkins, test this please

PR #2855, opened 2017-03-24
core: add @since annotations to MethodDescriptor

carl-mastrangelo:


core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -162,7 +171,11 @@ public final boolean serverSendsOneMessage() {
     public T parse(InputStream stream);
   }
 
-  /** A marshaller that supports retrieving it's type parameter {@code T} at runtime. */
+  /**
+   * A marshaller that supports retrieving it's type parameter {@code T} at runtime.
+   *
+   * @since 1.1.0
```
dapengzhang0: 1.1.1? As 1.1.0 is only a git tag not a maven artifact.

dapengzhang0:
what is your plan for all the files?

carl-mastrangelo:
Add them as needed.  This was left over from the 1.1.0 release.  With a rapid release cycle, knowing when methods came in is important.

Issue #2562, opened 2017-01-03
Make transport selection and stream start atomic


zhangkun83:
Right now they are done in two steps:
1. A transport that is in READY state is selected
2. `newStream()` is called on the selected transport.

If transport is shutdown (by LoadBalancer or channel idle mode) between the two steps, Step 2 will fail spuriously. Currently we work around this by adding a delay between stopping selecting a subchannel (which owns the transport) and shutting it down. As long as the delay is longer than the time between Step 1 and Step 2, the race won't happen.

This is not ideal because it relies on timing to work correctly, and will still fail in extreme cases where the time between the two steps are longer than the pre-set delay.

It would be a better solution to differentiate the racy shutdown and the intended shutdown (Channel is shutdown for good). In response to racy shutdown, transport selection will be retried. The `clientTransportProvider` in `ManagedChannelImpl` is in the best position to do this, because it knows whether the Channel has shutdown. `clientTransportProvider` would have to call `newStream()` and start the stream, and return the started stream to `ClientCallImpl` instead of a transport.

ejona86:
We should investigate whether we can use this to fix the race seen in #2857

biran0079:
What happens to on-going rpc when channel shuts down? 

pick READY transport
  channel goes idle, scheduled to shut down
newStream, rpc starts
  channel shuts down

In scenario above, if rpc did not finish before channel shuts down, would it fail?

zhangkun83:
@biran0079 I think you meant "transport shuts down" for the last item. If a transport shuts down with active RPCs, these RPCs will continue normally.

biran0079:
@zhangkun83 I see. Thanks for explaining!

zhangkun83:
The transport could also be shutdown by server-sending a GOAWAY. Unlike the case described in my first post, this case cannot be mitigated by adding a delay, thus is more problematic.

zhangkun83:
It seems transparent retry is the answer for both the local shutdown and the GOAWAY cases. Go and C++'s balancer APIs have the same issue and they are also relying on transparent retry.

PR #2857, opened 2017-03-27
core: Do not call startDeadlineTimer when is deadlineCancellationExecutor is null

kkaneda:
We got a `NullPointerException` from `ClientCallImpl#startDeadlineTimer` when a new Call is created after a Netty channel is terminated. Here is a stacktrace:

```
INTERNAL: java.lang.NullPointerException
at io.grpc.internal.ClientCallImpl.startDeadlineTimer(ClientCallImpl.java:320)
at io.grpc.internal.ClientCallImpl.start(ClientCallImpl.java:253)
```

The following code snippet reproduces the bug:

```
ManagedChannel channel = NettyChannelBuilder.forAddress(host, port)
    .usePlaintext(true)
    .build();
channel.shutdown();

Thread.sleep(1000);

GreeterGrpc.GreeterBlockingStub stub =
GreeterGrpc.newBlockingStub(channel)
    .withDeadlineAfter(10, TimeUnit.SECONDS);
stub.sayHello(HelloRequest.newBuilder().setName("world").build());
```

The issue was that `ClientCallImpl` is created from `RealChannel#newCall` *after* `ManagedChannelImpl#maybeTerminateChannel` is called and `scheduledExecutor` is set to null. In such a scenario, `deadlineCancellationExecutor` is set to null.

I think there are several ways to fix this, but one way would be to just avoid calling `startDeadlineTimer()` when `deadlineCancellationExecutor` is null. `DelayedClientTransport` will create a `FailingClientStream` with `Status.UNAVAILABLE` and we will get

```
Exception in thread "main" io.grpc.StatusRuntimeException:
UNAVAILABLE: Channel has shutdown (reported by delayed transport)
```

### DIFF
--- a/core/src/main/java/io/grpc/internal/ClientCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ClientCallImpl.java
@@ -249,7 +249,9 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT>
     context.addListener(this, directExecutor());
     if (effectiveDeadline != null
         // If the context has the effective deadline, we don't need to schedule an extra task.
-        && context.getDeadline() != effectiveDeadline) {
+        && context.getDeadline() != effectiveDeadline
+        // If the channel has been terminated, we don't need to schedule an extra task.
+        && deadlineCancellationExecutor != null) {
       deadlineCancellationFuture = startDeadlineTimer(effectiveDeadline);
     }
     if (cancelListenersShouldBeRemoved) {


ejona86:
Hmm... this is interesting, mainly because we've not seen it earlier. I wonder if we regressed somewhere. I don't quite remember what our solution to this problem was.

Although the fix helps, it seems insufficient since the code could end up using a non-`null` `scheduledExecutor` after termination, which can also fail. I think maybe we could use #2562 to address the more fundamental issue.

ejona86:
Jenkins, okay to test

kkaneda:
Thanks for the review, @ejona86 ! Agreed. This change is really a short-term workaround, and   #2562 looks like the right direction. (I don't think I can make such a change given limited experience/knowledge on gRPC, though..)

cc: @lukaszx0

ejona86:
@kkaneda, yeah, don't worry about implementing something fuller. #2562 requires being comfortable with many different parts of the stack.

lukaszx0:
Nice. Thanks @kkaneda and @ejona86 !

PR #2192, opened 2016-08-23
core: Add lockless SingleProducerSerializingExecutor

ejona86:
This is very low priority, and I'm not quite sure what benefits we need
for the change to be "worth it."

Although it is nice it is faster for many runnables, we are primarily
concerned with avoiding the lock to reduce tail latency. It uses more
memory with its linked list than SerializingExecutor (which uses array
queue), but in the future we can allow callers to provide a
LinkedRunnable (to SerializeReentrantCallsDirectExecutor as well) and
not incur an allocation each execute().

For maybe about 10% performance penality and additional complexity, the
algorithm should be able to be extended to multi-producer (via CAS on
'next'). But we don't have a strong need for multi-producer, so we favor
avoiding even more complexity.

Fixes #2122

```
Benchmark                                                 (type)    Mode Cnt      Score     Error  Units
SerializingExecutorBenchmark.manyRunnables               CLASSIC  sample 458258  65360.652 ± 129.949  ns/op
SerializingExecutorBenchmark.manyRunnables       SINGLE_PRODUCER  sample 480660  30988.375 ±  80.705  ns/op
SerializingExecutorBenchmark.oneRunnableLatency          CLASSIC  sample 629954  11756.902 ± 443.787  ns/op
SerializingExecutorBenchmark.oneRunnableLatency  SINGLE_PRODUCER  sample 629660  11413.764 ±  60.949  ns/op
```

PR #2858, opened 2017-03-28
core: make SerializingExecutor lockless

carl-mastrangelo:
This is an alternative implementation of #2192

### DIFF
--- a/core/src/main/java/io/grpc/internal/SerializingExecutor.java
+++ b/core/src/main/java/io/grpc/internal/SerializingExecutor.java
@@ -31,13 +31,16 @@
 
 package io.grpc.internal;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.google.common.base.Preconditions;
-import java.util.ArrayDeque;
 import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import javax.annotation.concurrent.GuardedBy;
+import javax.annotation.Nullable;
 
 /**
  * Executor ensuring that all {@link Runnable} tasks submitted are executed in order
@@ -45,7 +48,7 @@ import javax.annotation.concurrent.GuardedBy;
  * running at the same time.
  */
 // TODO(madongfly): figure out a way to not expose it or move it to transport package.
-public final class SerializingExecutor implements Executor {
+public final class SerializingExecutor implements Executor, Runnable {
   private static final Logger log =
       Logger.getLogger(SerializingExecutor.class.getName());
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/grpc/internal/SerializingExecutor.java
```diff
@@ -82,90 +69,48 @@ public SerializingExecutor(Executor executor) {
...
    * Runs the given runnable strictly after all Runnables that were submitted
    * before it, and using the {@code executor} passed to the constructor.     .
    */
   @Override
   public void execute(Runnable r) {
-    Preconditions.checkNotNull(r, "'r' must not be null.");
-    boolean scheduleTaskRunner = false;
-    synchronized (internalLock) {
-      waitQueue.add(r);
+    runQueue.offer(checkNotNull(r, "'r' must not be null."));
```
ejona86: For sanity's sake, use `add()`?

core/src/main/java/io/grpc/internal/SerializingExecutor.java
```diff
@@ -82,90 +69,48 @@ public SerializingExecutor(Executor executor) {
...
   }
 
-  /**
-   * Task that actually runs the Runnables.  It takes the Runnables off of the
-   * queue one by one and runs them.  After it is done with all Runnables and
-   * there are no more to run, puts the SerializingExecutor in the state where
-   * isThreadScheduled = false and returns.  This allows the current worker
-   * thread to return to the original pool.
-   */
-  private class TaskRunner implements Runnable {
```
ejona86: Why get rid of the inner TaskRunner class?

core/src/main/java/io/grpc/internal/SerializingExecutor.java
```diff
@@ -82,90 +69,48 @@ public SerializingExecutor(Executor executor) {
...
-            isThreadScheduled = false;
-          }
-        }
+        r.run();
+      } catch (RuntimeException e) {
+        // Log it and keep going.
+        log.log(Level.SEVERE, "Exception while executing runnable " + r, e);
       }
     }
+    running.set(false);
```
ejona86: This isn't protected against exceptions other than RuntimeException. It seems we would want it to deal with others.

core/src/main/java/io/grpc/internal/SerializingExecutor.java
carl-mastrangelo: I can, though you might consider an alternative:
```
boolean success = runQueue.offer(r);
assert success;
schedule();
```

I haven't done testing but this will lightly inflate the method making it harder to inline.  `add()` is just a wrapper around `offer()`.  The reason I picked offer is because I am using the [analogous](https://docs.oracle.com/javase/7/docs/api/java/util/Queue.html) function `poll()` below.  


Thoughts?

core/src/main/java/io/grpc/internal/SerializingExecutor.java
carl-mastrangelo: I was trying to drop easy allocations.  Making it the same object avoids allocating a TaskRunner.  CLQ costs an allocation per add so I was trying to pay as much of that cost back in this class.

core/src/main/java/io/grpc/internal/SerializingExecutor.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/internal/SerializingExecutor.java
ejona86: `assert success;` works fine too. While `offer()` is analogous, there is no implicit pairing between them; use whichever version you need. For instance, when using `BlockingQueue` it is quite normal to mix blocking and non-blocking operations, even though they are in a [similar table](http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/BlockingQueue.html).

ejona86:
@carl-mastrangelo, would you mind stealing the tests from #2192? I would expect most of them to work here without much difficulty.

carl-mastrangelo:
@ejona86 done

PR #2860, opened 2017-03-29
core: Fix User-Agent Javadoc in ManagedChannelBuilder

ejona86:
The behavior of how the application's User-Agent is used changed in
2247ad2. But the Javadoc was not updated.

### DIFF
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -118,8 +118,8 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   /**
    * Provides a custom {@code User-Agent} for the application.
    *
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -118,8 +118,8 @@
   /**
    * Provides a custom {@code User-Agent} for the application.
    *
-   * <p>It's an optional parameter. If provided, the given agent will be prepended by the
-   * grpc {@code User-Agent}.
+   * <p>It's an optional parameter. The library will provide a user agent independent of this
+   * option. If provided, the given agent will appear before the library's user agent information.
```
dapengzhang0: s/appear before/prepend ? Otherwise its unclear where it will appear. Http2Headers?

ejona86: Done

PR #2863, opened 2017-03-30
core: StreamTracer

zhangkun83:
Background
==========

LoadBalancer needs to track RPC measurements and status for
load-reporting.  We need to introduce a "Tracer" API for that.

Since such API is very close to the current
Census(instrumentation)-based stats reporting mechanism in terms of what
are recorded, we will migrate the Census-based stats reporting under the
new Tracer API.

Alternatives
============

We considered plumbing the LB-related information from the LoadBalancer
to the core, and recording those information along with the currently
recorded stats to Census. The LB-related information, such as LB_ID,
reason for dropping reqeusts etc, would be added to the Census
StatsContext as tags.

Since tags are held by StatsContext before eventually being recorded by
providing the measurements, and StatsContext is immutable, this would
require a way for LoadBalancer to override the StatsContext, which means
LoadBalancer API would has direct reference to the Census StatsContext.
This is undesirable because Census API is not stable yet.

Part of the LB-related information is whether the client has received
the initial headers from the server.  While such information can be
grabbed by implementing a ClientInterceptor, it must be recorded along
with other information such as LB_ID to be useful, and LB_ID is only
available in GrpclbLoadBalancer.

Bottom line, trying to use solely the Census StatsContext API to record
LB load information would require extra data plumbing channel between
ClientInterceptor, LoadBalancer and the gRPC core, as well as exposing
Census API on the gRPC API.  Even with those extensive changes, we are
yet to find a working solution. Therefore, we abandoned this idea and
propose this PR.

Summary of changes
==================

API summary
-----------
Introduce "StreamTracer" API, a callback interface for receiving stats
and tracing related updates concerning **a single stream**.
"ClientStreamTracer" and "ServerStreamTracer" add side-specific
events. A stream can have zero or more tracers and report to all of
them.

On the client-side, CallOptions now takes a list of
ClientStreamTracer.Factory. Opon creating a ClientStream, each of the
factory creates a ClientStreamTracer for the stream. This allows
ClientInterceptors to install its own tracer factories by overriding the
CallOptions.

Since StreamTracer only tracks the span of a stream, tracking of a
ClientCall needs to be done in a ClientInterceptor.  By installing its
own StreamTracer when a ClientCall is created, ClientInterceptor can
associate the updates for a Call with the updates for the Streams
created for that Call.  This is how we keep the existing Census
reporting mechanism in CensusStreamTracerModule.

On the server-side, ServerStreamTracer.Factory is added through the
ServerBuilder, and is used to create ServerStreamTracers for every
ServerStream.

The Tracer API supports propagation of stats/tracing information through
Context and metadata.  Both client-side and server-side tracer factories
have access to the headers object.  Client-side tracer relies on
interceptor to read the Context, while server-side tracer has
filterContext() method that can override the Context.

Implementation details
----------------------

Only real streams report stats.  Pseudo streams such as delayed stream,
failing stream don't report.  InProcess transport streams currently
don't report stats.

"StatsTraceContext" which used to receive updates from core and report
directly to Census (StatsContext), now delegates to the StreamTracers of
a stream.  On the client-side, the scope of a StatsTraceContext reduces
from ClientCall to a ClientStream to match the scope of StreamTracer.

The Census-specific logic that was in StatsTraceContext is moved into
CensusStreamTracerModule, which produces factories for StreamTracers
that report to Census.

Reporting with StatsTraceContext is moved out of the Channel/Server/Call layer
into Transport/Stream layer, to match the scope change of
StatsTraceContext. StatsTraceContext is now created by the transport, thus
ServerTransportListener.methodDetermined() is removed because it was solely
for facilitating ServerImpl creating StatsTraceContext and passing it back to
the stream.

Bug fixed
----------------

The end of a server-side call was reported in ServerCallImpl's
ServerStreamListenerImpl.closed(), which was wrong.  Because closed()
receiving OK doesn't necessarily mean the RPC ended with OK.  Instead it
means the server has successfully sent the final status, which may be
non-OK, to the client.

Now the end report is done in both ServerStream.close(any Status) and
before calling ServerStreamListener.closed(non-OK).  Whichever happens
first is the reported status.

TODOs
=====

A follow-up change to the LoadBalancer API will add a
ClientStreamTracer.Factory to the PickResult to complete the API needed
by load-reporting.


zhangkun83:
Reviewers: sorry for this large PR. It is an intrusive change and I had to make sure it works for all cases, thus I couldn't do it more incrementally. Here is the recommended order to review:

1. The public APIs
 - `StreamTracer`, `ClientStreamTracer`, `ServerStreamTracer`.
 - `CallOptions`
 - `ServerBuilder`
2. The core implementations
 - `AbstractManagedChannelImplBuilder`, `AbstractServerImplBuilder`
 - `StatsTraceContext`, `CensusStreamTracerModule`
 - `ManagedChannelImpl`, `ClientCallImpl`, `ServerImpl`, `ServerCallImpl`
 - `AbstractClientStream`, `AbstractClientStream2`, `AbstractServerStream`.
3. The transport implementations
 - `Netty*`, `OkHttp*`, `InProcess*`
4. The tests
5. The remaining files are mechanical changes.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -171,17 +210,41 @@ static int getPort() {
   protected TestServiceGrpc.TestServiceBlockingStub blockingStub;
   protected TestServiceGrpc.TestServiceStub asyncStub;
 
+  private final LinkedBlockingQueue<ClientStreamTracer> clientStreamTracers =
+      new LinkedBlockingQueue<ClientStreamTracer>();
+
+  private final ClientStreamTracer.Factory clientStreamTracerFactory =
```
ejona86: I've not really been happy with how the interop tests have been impacted by stats testing. I feel like its abusive of the tests and making them drift from their purpose (high level, "does this feature work" from the user's point of view). And all this boilerplate is making it worse. I'm okay with this going in, but we need to come up with a longer-term plan.

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
```diff
@@ -471,20 +535,31 @@ public void ping_afterTermination() throws Exception {
 
   @Test
   public void newStream_duringShutdown() throws Exception {
```
ejona86: I'm not excited about how many places stats are checked in these tests as well. I don't know why we would need so many asserts for stats in newStream_duringShutdown, for example.

zhangkun83:
https://travis-ci.org/grpc/grpc-java/jobs/219514837
```
:grpc-interop-testing:test
io.grpc.testing.integration.Http2NettyLocalChannelTest > cancelAfterFirstResponse FAILED
    Wanted but not invoked:
    serverStreamTracer.streamClosed(
        <Capturing argument>
    );
    -> at io.grpc.testing.integration.AbstractInteropTest.assertServerMetrics(AbstractInteropTest.java:1532)
    However, there were other interactions with this mock:
    -> at io.grpc.internal.StatsTraceContext.serverFilterContext(StatsTraceContext.java:120)
    -> at io.grpc.internal.StatsTraceContext.inboundMessage(StatsTraceContext.java:152)
    -> at io.grpc.internal.StatsTraceContext.inboundWireSize(StatsTraceContext.java:188)
    -> at io.grpc.internal.StatsTraceContext.inboundUncompressedSize(StatsTraceContext.java:179)
    -> at io.grpc.internal.StatsTraceContext.outboundMessage(StatsTraceContext.java:143)
    -> at io.grpc.internal.StatsTraceContext.outboundWireSize(StatsTraceContext.java:170)
    -> at io.grpc.internal.StatsTraceContext.outboundUncompressedSize(StatsTraceContext.java:161)

281 tests completed, 1 failed
:grpc-interop-testing:test FAILED

```

I don't understand this. In this test the server has two tracer factories, the mocked tracer and the Census tracer. The verification that Census has got the final status had passed prior to the failing point, which means the Census tracer has `streamClosed()` called. It's very unlikely that the mocked tracer doesn't have `streamClosed()` called. I couldn't reproduce this locally.

zhangkun83:
I do find the `poll()` with timeout is not doing what I wanted. The timeout should have been on the `verify().streamClosed()`. Let's see how the latest commit does.

PR #2864, opened 2017-03-30
netty: expose server side keepalive API

dapengzhang0:


netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
```diff
@@ -227,17 +236,57 @@ public NettyServerBuilder maxHeaderListSize(int maxHeaderListSize) {
     return this;
   }
 
+  /**
+   * Sets a custom KEEPALIVE_TIME. An unreasonably small value will be bumped to a reasonable value,
```
ejona86: Don't use KEEPALIVE_TIME casing, since that is meaningless to the user. You can just say "keepalive time". But you should also define what that means. Same with timeout.

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
```diff
@@ -227,17 +236,57 @@ public NettyServerBuilder maxHeaderListSize(int maxHeaderListSize) {
...
+    if (keepAliveTimeInNanos < MIN_KEEPALIVE_TIME_NANO) {
+      // Bump KEEPALIVE_TIME.
+      keepAliveTimeInNanos = MIN_KEEPALIVE_TIME_NANO;
+    }
+    return this;
+  }
+
+  /**
+   * Sets a custom KEEPALIVE_TIMEOUT. An unreasonably small value will be bumped to a reasonable
+   * value. KEEPALIVE_TIMEOUT must be much smaller than KEEPALIVE_TIME.
```
ejona86: Why must `KEEPALIVE_TIMEOUT must be much smaller than KEEPALIVE_TIME` and what does "much smaller" mean? I'd probably just get rid of this restriction.

netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -92,6 +95,8 @@
     this.flowControlWindow = flowControlWindow;
     this.maxMessageSize = maxMessageSize;
     this.maxHeaderListSize = maxHeaderListSize;
+    this.keepAliveTimeInNanos = keepAliveTimeInNanos;
+    this.keepAliveTimeoutInNanos = keepAliveTimeoutInNanos;
```
carl-mastrangelo: can these be negative?

core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -179,6 +179,20 @@
...
+   * @since 1.3.0
+   */
+  public static final long DEFAULT_SERVER_KEEPALIVE_TIME_NANOS = TimeUnit.HOURS.toNanos(2L);
+
+  /**
+   * The default timeout in nanos for a server keepalive ping request.
+   *
+   * @since 1.3.0
+   */
+  public static final long DEFAULT_SERVER_KEEPALIVE_TIMEOUT_NANOS = TimeUnit.SECONDS.toNanos(20L);
```
carl-mastrangelo: Shouldn't the "unset" variable be here too?  

public static final long SERVER_KEEPALIVE_TIME_NANOS_DISABLED = Long.MAX_VALUE;

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
```diff
@@ -59,6 +62,10 @@
 public final class NettyServerBuilder extends AbstractServerImplBuilder<NettyServerBuilder> {
   public static final int DEFAULT_FLOW_CONTROL_WINDOW = 1048576; // 1MiB
 
+  private static final long MIN_KEEPALIVE_TIME_NANO = TimeUnit.MILLISECONDS.toNanos(1L);
+  private static final long MIN_KEEPALIVE_TIMEOUT_NANO = TimeUnit.MICROSECONDS.toNanos(499L);
+  private static final long AS_LARGE_AS_INFINITE = TimeUnit.DAYS.toNanos(1000L);
```
carl-mastrangelo: This is a different definition than NettyServerHandler users.  can you use the same constant in both places?

core/src/main/java/io/grpc/internal/GrpcUtil.java
dapengzhang0: done

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
dapengzhang0: done

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
```diff
@@ -237,36 +237,37 @@ public NettyServerBuilder maxHeaderListSize(int maxHeaderListSize) {
   }
 
   /**
-   * Sets a custom KEEPALIVE_TIME. An unreasonably small value will be bumped to a reasonable value,
-   * and {@code Long.MAX_VALUE} nano seconds or an unreasonably large value will disable keep alive.
+   * Sets a custom keepalive time, the delay time for sending next keepalive ping.An unreasonably
```
ejona86: s/ping.An/ping. An/

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
```diff
@@ -227,6 +237,43 @@ public NettyServerBuilder maxHeaderListSize(int maxHeaderListSize) {
     return this;
   }
 
+  /**
+   * Sets a custom keepalive time, the delay time for sending next keepalive ping.An unreasonably
+   * small value will be bumped to a reasonable value, and {@code Long.MAX_VALUE} nano seconds or an
```
ejona86: Instead of saying "reasonable value", maybe:
> An unreasonably small value my be increased.

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
```diff
@@ -227,6 +237,43 @@ public NettyServerBuilder maxHeaderListSize(int maxHeaderListSize) {
     return this;
   }
 
+  /**
+   * Sets a custom keepalive time, the delay time for sending next keepalive ping.An unreasonably
```
ejona86: s/ping.An/ping. An/

PR #2869, opened 2017-03-31
netty: Add server keepalive enforcement

ejona86:
Everything is currently permitted, but I've tested with other
configurations and all tests pass. I'll set the restrictive default at
the same time as adding a configuration API.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/KeepAliveEnforcer.java
+++ b/netty/src/main/java/io/grpc/netty/KeepAliveEnforcer.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2017, Google Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ *    * Neither the name of Google Inc. nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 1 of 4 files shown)


netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -526,6 +553,34 @@ public void onRstStreamRead(ChannelHandlerContext ctx, int streamId, long errorC
...
     @Override
+    public void onPingRead(ChannelHandlerContext ctx, ByteBuf data) throws Http2Exception {
+      if (keepAliveManager != null) {
+        keepAliveManager.onDataReceived();
+      }
+      if (!keepAliveEnforcer.pingAcceptable()) {
+        ByteBuf debugData = ByteBufUtil.writeAscii(ctx.alloc(), "too_many_pings");
+        goAway(ctx, connection().remote().lastStreamCreated(), Http2Error.ENHANCE_YOUR_CALM.code(),
+            debugData, ctx.newPromise());
+        flush(ctx);
```
ejona86: No. ctx.flush() goes to the _next_ handler in the pipeline and so wouldn't actually flush anything from HTTP/2. This problem is mostly limited to HTTP/2. Welcome to http2+Netty.

netty/src/main/java/io/grpc/netty/KeepAliveEnforcer.java
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  @VisibleForTesting
+  interface Ticker {
+    long nanoTime();
+  }
+
+  @VisibleForTesting
+  static class SystemTicker implements Ticker {
+    public static SystemTicker INSTANCE = new SystemTicker();
```
ejona86: Done

netty/src/main/java/io/grpc/netty/KeepAliveEnforcer.java
```diff
@@ -0,0 +1,126 @@
...
+    if (!valid) {
+      pingStrikes++;
+      return !(pingStrikes > MAX_PING_STRIKES);
+    } else {
+      lastValidPingTime = now;
+      return true;
+    }
+  }
+
+  public void resetCounters() {
```
ejona86: Done

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
ejona86: So looking at it some more, it could be `ctx.flush()`. But I think I'd favor using `flush(ctx)` everywhere instead of `ctx.flush()`. Unfortunately, `flush(ctx)` sucks because it `throws Exception`, and so could be considerably more painful in situations.

ejona86: As discussed offline, this is part of the read path, so `channelReadComplete()` will do the flush for us. I've removed the flush.

ejona86:
I've simplified the error handling of the forcefulClose.

PR #2870, opened 2017-03-31
core: Clarify misleading idleTimeout documentation

ejona86:


ejona86:
Hmm... so this doc update isn't quite right. The problem is it was talking about "idle mode", not the "IDLE state". The IDLE state doesn't shut down the NR and LB, whereas idle mode does. That's far too subtle for our users.

ejona86:
We're going to discuss more in depth the subtlety separately.

PR #2871, opened 2017-03-31
netty: Fix client keepalive initialization (again)

### LINKED ISSUE #2828: KeepAliveManager is not apply to ConnectionHandler in netty transport

ejona86:
d116cc9 fixed the NPE, but the initialization of the manager happened
_after_ newHandler() was called, so a null manager was passed to the
handler.

Fixes #2828 

------

I haven't tested this manually yet. I wanted to send it out to get opinions on whether this is the fix we want. It seems considerably less invasive than some other options we were discussing.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyClientTransport.java
+++ b/netty/src/main/java/io/grpc/netty/NettyClientTransport.java
@@ -173,7 +173,8 @@ class NettyClientTransport implements ConnectionClientTransport {
           false);
     }
 
-    handler = newHandler(lifecycleManager, keepAliveManager);
+    handler = NettyClientHandler.newHandler(lifecycleManager, keepAliveManager, flowControlWindow,
+        maxHeaderListSize, Ticker.systemTicker());
     HandlerSettings.setAutoWindow(handler);
 
     negotiationHandler = negotiator.newHandler(handler);
@@ -310,13 +311,5 @@ class NettyClientTransport implements ConnectionClientTransport {
       return shutdownStatus;
     }
     return Utils.statusFromThrowable(t);
-  }
-
-  private NettyClientHandler newHandler(ClientTransportLifecycleManager lifecycleManager,
-      KeepAliveManager keepAliveManager) {
-    // Only use fields initialized before start(); have fields initialized in start() be passed in.
-    // Otherwise it is hard to ensure the fields are initialized in time.
-    return NettyClientHandler.newHandler(lifecycleManager, keepAliveManager, flowControlWindow,
-        maxHeaderListSize, Ticker.systemTicker());
   }
 }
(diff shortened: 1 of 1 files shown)


netty/src/main/java/io/grpc/netty/NettyClientTransport.java
```diff
@@ -165,14 +166,20 @@ public ClientStream newStream(MethodDescriptor<?, ?> method, Metadata headers) {
   public Runnable start(Listener transportListener) {
     lifecycleManager = new ClientTransportLifecycleManager(
```
dapengzhang0: did you forget to replace the `lifecycleManager ` field with a local variable? Ditto for the `keepaliveManager` field if it is not exposed to test.

ejona86: `lifecycleManager` is used in `statusFromFailedFuture`, so it must remain a field. I left `keepaliveManager` because it is used by the test... although I agree that is questionable given the test didn't catch this bug.

dapengzhang0: oops, I see `lifecycleManager` is also used in `statusFromFailedFuture`. But why did you add `lifecycleManager` as an extra argument of `newHandler`?

ejona86: I thought my comment explained that:
> Only use fields initialized before start()

Since `lifecycleManager` is initialized in `start()`, the flow in `start()` is more obvious when being passed as an argument. Otherwise you could move `newHandler()` within `start()` and not notice that a field is no longer initialized.

dapengzhang0: I see. I think move it inside `start()` is cleaner since it's only one line of code and used only once.

dapengzhang0: I see. I think move it inside `start()` is cleaner since it's only one line of code and used only once.

ejona86: Good idea. Done.

ejona86:
I've run manual tests and I see keepalive pings!

PR #2872, opened 2017-04-01
core/grpclb: Ignore equivalent pickers

zhangkun83:
Add `isEquivalentTo()` to `SubchannelPicker` to allow the Channel to ignore **equivalent** picker updates. Two benefits:
 1. Save the cycles spent on unnecessary "reprocessing" of pending streams
 2. Allows the Picker state (e.g., current round-robin index) to be kept if nothing has changed, resulting in less churn. In particular, because GRPCLB balancer may send identical address list to the client just to refresh the TTL, without this change the round-robin index is reset frequently enough to cause over-the-threshold imbalance in our integration test, resulting test failures.

Alternatively, LoadBalancer implementations could check whether the error status or round-robin list etc of the new Picker is the same as the current one and skip the call to `updatePicker()` if it's the case. However, a LoadBalancer may have more than one places where the Picker is constructed and the equivalency-check logic would be all over the place.

### DIFF
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -209,6 +209,16 @@ public abstract class LoadBalancer {
      * @param args the pick arguments
      */
     public abstract PickResult pickSubchannel(PickSubchannelArgs args);
+
+    /**
+     * Returns {@code true} if the two pickers should be considered equivalent.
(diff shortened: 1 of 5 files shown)


zhangkun83:
Never mind... It may not be so bad to achieve the same inside the LoadBalancer, and the API may bring more confusion than usefulness.

PR #2874, opened 2017-04-03
netty: refactor NettyChannelBuilder keepalive API

dapengzhang0:
To be in line with `NettyServerBuilder` APIs
- Deleted `enableKeepAlive(boolean enable)` and
`enableKeepAlive(boolean enable, long keepAliveDelay, TimeUnit delayUnit, long keepAliveTimeout, 
TimeUnit timeoutUnit)`
which never worked in v1.2

- Added `keepAliveTime(long keepAliveTime, TimeUnit timeUnit)` and
`keepAliveTimeout(long keepAliveTimeout, TimeUnit timeUnit)`

### DIFF
--- a/core/src/main/java/io/grpc/internal/GrpcUtil.java
+++ b/core/src/main/java/io/grpc/internal/GrpcUtil.java
@@ -180,29 +180,21 @@ public final class GrpcUtil {
 
   /**
    * The magic keepalive time value that disables client keepalive.
-   *
-   * @since 1.3.0
    */
   public static final long KEEPALIVE_TIME_NANOS_DISABLED = Long.MAX_VALUE;
 
   /**
    * The default delay in nanos for server keepalive.
-   *
-   * @since 1.3.0
    */
   public static final long DEFAULT_SERVER_KEEPALIVE_TIME_NANOS = TimeUnit.HOURS.toNanos(2L);
 
   /**
    * The default timeout in nanos for a server keepalive ping request.
-   *
-   * @since 1.3.0
    */
   public static final long DEFAULT_SERVER_KEEPALIVE_TIMEOUT_NANOS = TimeUnit.SECONDS.toNanos(20L);
 
   /**
    * The magic keepalive time value that disables keepalive.
-   *
-   * @since 1.3.0
    */
   public static final long SERVER_KEEPALIVE_TIME_NANOS_DISABLED = Long.MAX_VALUE;
 
--- a/core/src/main/java/io/grpc/internal/GrpcUtil.java
+++ b/core/src/main/java/io/grpc/internal/GrpcUtil.java
@@ -176,7 +176,14 @@ public final class GrpcUtil {
   /**
    * The default timeout in nanos for a keepalive ping request.
    */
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -176,7 +176,14 @@
   /**
    * The default timeout in nanos for a keepalive ping request.
    */
-  public static final long DEFAULT_KEEPALIVE_TIMEOUT_NANOS = TimeUnit.MINUTES.toNanos(2);
+  public static final long DEFAULT_KEEPALIVE_TIMEOUT_NANOS = TimeUnit.SECONDS.toNanos(20L);
```
lukaszx0: Just double checking - was this intended to change to `20`?

core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -176,7 +176,14 @@
...
+
+  /**
+   * The magic keepalive time value that disables client keepalive.
```
lukaszx0: Magic?

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -241,28 +243,35 @@ public NettyChannelBuilder usePlaintext(boolean skipNegotiation) {
...
-
   /**
-   * Enable keepalive with default delay and timeout.
+   * Sets a custom keepalive time, the delay time for sending next keepalive ping. An unreasonably
+   * small value might be increased, and {@code Long.MAX_VALUE} nano seconds or an unreasonably
+   * large value will disable keepalive.
+   *
+   * @since 1.3.0
    */
-  public final NettyChannelBuilder enableKeepAlive(boolean enable) {
```
ejona86: Please leave these for a cycle, to ease upgrading (even though we know that keepalive was broken). Just mark them as `@Deprecated`. They can be reimplemented by calling the new API if that makes it easier.

core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -176,7 +176,14 @@
...
+   *
+   * @since 1.3.0
```
ejona86: No need for `@since`, since this is an internal API.

core/src/main/java/io/grpc/internal/GrpcUtil.java
dapengzhang0: The default should be 20 seconds. The old value 2 minutes was used for OkHttp, should we keep that value for OkHttp @ejona86 @zsurocking ?

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
dapengzhang0: done

core/src/main/java/io/grpc/internal/GrpcUtil.java
dapengzhang0: done

dapengzhang0:
jenkins, retest this please

PR #2877, opened 2017-04-05
netty: Add config for server keepalive enforcement

ejona86:
Now that there is a config, the new defaults are now being enabled.
Previously there were no default limits. Now keepalives may not be more
frequent than every 5 minutes and only when there are outstanding RPCs.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
@@ -81,6 +81,8 @@ public final class NettyServerBuilder extends AbstractServerImplBuilder<NettySer
   private int maxHeaderListSize = GrpcUtil.DEFAULT_MAX_HEADER_LIST_SIZE;
   private long keepAliveTimeInNanos =  DEFAULT_SERVER_KEEPALIVE_TIME_NANOS;
   private long keepAliveTimeoutInNanos = DEFAULT_SERVER_KEEPALIVE_TIMEOUT_NANOS;
+  private boolean permitKeepAliveWithoutCalls;
+  private long permitKeepAliveTimeInNanos = TimeUnit.MINUTES.toNanos(5);
 
   /**
    * Creates a server builder that will bind to the given port.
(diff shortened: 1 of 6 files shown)


netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
```diff
@@ -81,6 +81,8 @@
   private int maxHeaderListSize = GrpcUtil.DEFAULT_MAX_HEADER_LIST_SIZE;
   private long keepAliveTimeInNanos =  DEFAULT_SERVER_KEEPALIVE_TIME_NANOS;
   private long keepAliveTimeoutInNanos = DEFAULT_SERVER_KEEPALIVE_TIMEOUT_NANOS;
+  private boolean permitKeepAliveWithoutCalls;
+  private long permitKeepAliveTimeInNanos = TimeUnit.MINUTES.toNanos(5);
```
dapengzhang0: make a new const `DEFAULT_PERMIT_KEEPALIVE_TIME` in `GrpcUtil.java`?

ejona86: Meh. I think it makes the code less readable by adding a layer of indirection, and it is only used in this one place. It's not like on client-side where we have multiple implementations and we want to keep them in sync. I'd rather the constant be added when it is helpful/necessary instead of now.

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
```diff
@@ -274,6 +276,51 @@ public NettyServerBuilder keepAliveTimeout(long keepAliveTimeout, TimeUnit timeU
...
+   *
+   * <p>Even though a default is defined that allows some keep-alives, clients must not use
+   * keep-alive without approval from the service owner. Otherwise, they may experience failures in
+   * the future if the service becomes more restrictive. When unthrottled, keep-alives can cause a
+   * significant amount of traffic and CPU usage, so clients and servers should be conservative in
+   * what they use and accept.
+   *
+   * @see #denyKeepAliveWithoutCalls()
+   * @see #permitKeepAliveWithoutCalls()
+   */
```
dapengzhang0: add `@since`

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
dapengzhang0: sounds good to me.

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
ejona86: Done.

dapengzhang0:
Jenkins failed, not related to the PR / introduced a flake?

ejona86:
I think it is a pre-existing flake.

ejona86:
retest this please

PR #2880, opened 2017-04-07
netty: implement server max connection age

dapengzhang0:


dapengzhang0:
rebased.
Jenkins, retest this please

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
```diff
@@ -279,6 +285,42 @@ public NettyServerBuilder keepAliveTimeout(long keepAliveTimeout, TimeUnit timeU
...
+    }
+    if (maxConnectionAgeInNanos < MIN_MAX_CONNECTION_AGE_NANO) {
+      maxConnectionAgeInNanos = MIN_MAX_CONNECTION_AGE_NANO;
+    }
+    return this;
+  }
+
+  /**
+   * Sets a custom grace time for the graceful connection termination when the connection age
+   * exceeds the max connection age. {@code Long.MAX_VALUE} nano seconds or an unreasonably large
```
ejona86: Describe what happens when this value is exceeded (i.e., streams are killed).

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -225,8 +244,41 @@ Throwable connectionError() {
...
 
   @Override
-  public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
+  public void handlerAdded(final ChannelHandlerContext ctx) throws Exception {
     serverWriteQueue = new WriteQueue(ctx.channel());
+
+    // init max connection age monitor
+    if (maxConnectionAgeInNanos != MAX_CONNECTION_AGE_NANOS_DISABLED) {
+      maxConnectionAgeMonitor = ctx.executor().schedule(
+          new Runnable() {
```
ejona86: Wrap in a LogExceptionRunnable

netty/src/test/java/io/grpc/netty/NettyServerHandlerTest.java
```diff
@@ -529,6 +533,56 @@ public void keepAliveEnforcer_noticesInactive() throws Exception {
...
+  public void maxConnectionAge() throws Exception {
+    maxConnectionAgeInNanos = TimeUnit.MILLISECONDS.toNanos(10L);
+    setUp();
+
+    assertTrue(channel().isOpen());
+
+    Thread.sleep(5L);
+    channel().runPendingTasks();
+
+    // GO_AWAY not sent yet
```
ejona86: This is going to be flaky. If you didn't do any sleep and did `runPendingTasks()`, that would be better, but 10ms still seems too small to be semi-reliable.

netty/src/test/java/io/grpc/netty/NettyServerHandlerTest.java
```diff
@@ -529,6 +533,56 @@ public void keepAliveEnforcer_noticesInactive() throws Exception {
...
+    createStream();
+
+    Thread.sleep(15L);
+    channel().runPendingTasks();
+
+    verifyWrite().writeGoAway(
+        eq(ctx()), eq(Integer.MAX_VALUE), eq(Http2Error.NO_ERROR.code()), any(ByteBuf.class),
+        any(ChannelPromise.class));
+    // channel not closed yet
+    assertTrue(channel().isOpen());
```
ejona86: This is going to be racy as well.

netty/src/test/java/io/grpc/netty/NettyServerHandlerTest.java
dapengzhang0: done

netty/src/test/java/io/grpc/netty/NettyServerHandlerTest.java
dapengzhang0: done

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
```diff
@@ -304,9 +304,10 @@ public NettyServerBuilder maxConnectionAge(long maxConnectionAge, TimeUnit timeU
   }
 
   /**
-   * Sets a custom grace time for the graceful connection termination when the connection age
-   * exceeds the max connection age. {@code Long.MAX_VALUE} nano seconds or an unreasonably large
-   * value are considered infinite.
+   * Sets a custom grace time for the graceful connection termination after the connection age
+   * exceeds the max connection age. If the connection is still open until the grace time being
```
ejona86: The grammar looks a bit wrong there. "Connection still open" also doesn't make it obvious that is caused by streams being open. We also try to use 'calls' in this part of the API. How about:

> RPCs have the grace time to complete. RPCs that do not complete in time will be cancelled, allowing the connection to terminate.

netty/src/test/java/io/grpc/netty/NettyServerHandlerTest.java
```diff
@@ -563,11 +562,11 @@ public void maxConnectionAge() throws Exception {
   @Test
   public void maxConnectionAgeGrace() throws Exception {
     maxConnectionAgeInNanos = TimeUnit.MILLISECONDS.toNanos(10L);
-    maxConnectionAgeGraceInNanos = TimeUnit.MILLISECONDS.toNanos(10L);
+    maxConnectionAgeGraceInNanos = TimeUnit.MILLISECONDS.toNanos(100L);
     setUp();
     createStream();
 
-    Thread.sleep(15L);
+    Thread.sleep(10L);
```
ejona86: Can we just get rid of this sleep? What does it do?

dapengzhang0: It tests after the max age, during the grace period, the channel is still open. Without the sleep how can I get to the grace period?

ejona86: Oh, I see. That makes sense.

netty/src/test/java/io/grpc/netty/NettyServerHandlerTest.java
ejona86: Penn pointed out that scheduling anything new only happens in `runPendingTasks()`, so this would actually be reliable.

PR #2882, opened 2017-04-08
core: allow SubchannelPicker to return a StreamTracer factory.

zhangkun83:
This allows LoadBalancers to trace the activities, including the final
status of the stream that is created as a result of the pick.

@ejona86 Although earlier we talked about switching `PickResult` to builder pattern, I found this particular change works fine with the current pattern. Let me know if you prefer doing the switch now.

### DIFF
--- a/core/src/main/java/io/grpc/internal/GrpcUtil.java
+++ b/core/src/main/java/io/grpc/internal/GrpcUtil.java
@@ -595,28 +595,27 @@ public final class GrpcUtil {
     }
     if (transport != null) {
       final ClientStreamTracer.Factory streamTracerFactory = result.getStreamTracerFactory();
-      if (streamTracerFactory != null) {
-        return new ClientTransport() {
-          @Override
-          public ClientStream newStream(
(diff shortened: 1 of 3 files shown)


lukaszx0:
LGTM

core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -582,15 +586,37 @@ public Long parseAsciiString(String serialized) {
...
     Subchannel subchannel = result.getSubchannel();
     if (subchannel != null) {
       transport = ((SubchannelImpl) subchannel).obtainActiveTransport();
     } else {
       transport = null;
     }
     if (transport != null) {
-      return transport;
+      final ClientStreamTracer.Factory streamTracerFactory = result.getStreamTracerFactory();
+      if (streamTracerFactory != null) {
```
lukaszx0: Actually small nit: maybe for better readability, make this an early return? (`if (streamTracerFactory == null) { return transport; }`)

zhangkun83: Done.

Issue #2641, opened 2017-01-23
Tracking Issue for MethodDescriptor.Builder being experimental


carl-mastrangelo:

carl-mastrangelo:
If there isn't any objection I would like to mark this as non experimental.

lukaszx0:
what does the experimental mean? I guess new, additive APIs will still be allowed and breaking API change will need to go through deprecation process?

ejona86:
@lukaszx0 This is referring to [ExperimentalApi](http://www.grpc.io/grpc-java/javadoc/io/grpc/ExperimentalApi.html), which means the API doesn't have our standard API stability guarantees.

setSafe and setIdempotent need to be marked experimental. Just use #1775 like is used for the `is*` methods.

PR #2886, opened 2017-04-10
compiler: move over to method descriptor builder

lukaszx0:
Ports compiler code to use builder code

Related to https://github.com/grpc/grpc-java/pull/2625

cc @carl-mastrangelo


ejona86:
Generated code can't call anything that is `@ExperimentalApi`. This has to wait until after #2641. Although it also seemed like that is just a formality.

lukaszx0:
rebased, it's green now

PR #2887, opened 2017-04-10
core: deprecate MethodDescriptor.create

lukaszx0:
Depends on #2886, related to #2625.

cc @carl-mastrangelo

### DIFF
--- a/core/src/main/java/io/grpc/MethodDescriptor.java
+++ b/core/src/main/java/io/grpc/MethodDescriptor.java
@@ -200,7 +200,9 @@ public final class MethodDescriptor<ReqT, RespT> {
    * @param requestMarshaller the marshaller used to encode and decode requests
    * @param responseMarshaller the marshaller used to encode and decode responses
    * @since 1.0.0
+   * @deprecated use {@link #newBuilder()}.
    */
+  @Deprecated
   public static <RequestT, ResponseT> MethodDescriptor<RequestT, ResponseT> create(
       MethodType type, String fullMethodName,
       Marshaller<RequestT> requestMarshaller,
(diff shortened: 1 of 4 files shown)


ejona86:
Is there much of a pushing factor for deprecating `create()`? It handles the common case easily and we can trivially emulate it with the builder if it became the least bit of burden. It also seems to have a cleaner creation pattern in the common case (not needing to mess with the generics; that can be avoided in the builder, but then you have indentation problems).

lukaszx0:
Not a strong opinion, we can leave it.

One question though - if we leave, do we want to grow it's arguments list when adding new fields to `MethodDescriptor` ? I've done this work mostly as a preparation step before pushing PR which would add support for service method options which will add new stuff to `MethodDescriptor`.

ejona86:
> do we want to grow it's arguments list when adding new fields to `MethodDescriptor`

No. `create()` doesn't scale. We added the builder instead of adding more arguments. And we'd want to swap to the builder in the generated code. I'm mainly saying that forcing existing users off the method probably doesn't add much value and does produce a bit uglier code. Since the method is stable, we have to maintain it anyway, so I'd figure just leave it.

I guess the counter-argument is that it would be a method we'd likely delete in a grpc 2.0. And we'll probably look for `@Deprecated` methods to delete at that point. Meh.

carl-mastrangelo:
I am in favor of deprecating `create`, if only to make it easier for users to find the Builder way.  That said, it is public API and it cannot be removed.   The tests need to still keep testing it.

carl-mastrangelo:
Since #2900 is in, I would be willing to approve this if the test were changed to just ignore deprecation, rather than use the new builder.

lukaszx0:
@carl-mastrangelo sounds good - pushed updates.

lukaszx0:
@carl-mastrangelo ping?

ejona86:
I still think that deprecating the current method probably isn't helping anyone. But meh.

PR #2892, opened 2017-04-10
netty: add jitter to max connection age

dapengzhang0:


netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
```diff
@@ -362,6 +364,13 @@ protected NettyServer buildTransportServer(
       negotiator = sslContext != null ? ProtocolNegotiators.serverTls(sslContext) :
               ProtocolNegotiators.serverPlaintext();
     }
+
+    if (maxConnectionAgeInNanos != MAX_CONNECTION_AGE_NANOS_DISABLED) {
+      // apply a random jitter of +/-10% to max connection age
+      maxConnectionAgeInNanos =
+          (long) ((.9D + new Random().nextDouble() * .2D) * maxConnectionAgeInNanos);
```
ejona86: Use `Math.random()`, since this isn't very contentious?

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
```diff
@@ -362,6 +364,13 @@ protected NettyServer buildTransportServer(
       negotiator = sslContext != null ? ProtocolNegotiators.serverTls(sslContext) :
               ProtocolNegotiators.serverPlaintext();
     }
+
+    if (maxConnectionAgeInNanos != MAX_CONNECTION_AGE_NANOS_DISABLED) {
+      // apply a random jitter of +/-10% to max connection age
```
ejona86: The intention was that this would be per-connection. That wasn't specified well. I've updated the spec doc.

ejona86:
I'd expect some sleeps in the tests need to be updated to account for the new slack time.

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
dapengzhang0: Done. I understand the semantic even it wasn't specified. Just didn't realize that the way that transports are created from the server side is different from client side.

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
dapengzhang0: done

PR #2895, opened 2017-04-10
okhttp: Add OptionalMethod support for private classes

kpayson64:
When OptionalMethod attempts to invoke a method on a private class that inherits from a superclass that has a public method, it fails.

This happens after changes to the library visibility in conscrypt.

### DIFF
--- a/okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/OptionalMethod.java
+++ b/okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/OptionalMethod.java
@@ -169,6 +169,12 @@ public class OptionalMethod<T> {
   private static Method getPublicMethod(Class<?> clazz, String methodName, Class[] parameterTypes) {
     Method method = null;
     try {
+      if (clazz == null) {
+        return null;
+      }
+      if ((clazz.getModifiers() & Modifier.PUBLIC) == 0) {
+        return getPublicMethod(clazz.getSuperclass(), methodName, parameterTypes);
+      }
       method = clazz.getMethod(methodName, parameterTypes);
       if ((method.getModifiers() & Modifier.PUBLIC) == 0) {
         method = null;
--- a/okhttp/src/test/java/io/grpc/okhttp/OptionalMethodTest.java
+++ b/okhttp/src/test/java/io/grpc/okhttp/OptionalMethodTest.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2017, Google Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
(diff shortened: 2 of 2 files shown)


okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/OptionalMethod.java
```diff
@@ -169,6 +169,12 @@ private Method getMethod(Class<?> clazz) {
   private static Method getPublicMethod(Class<?> clazz, String methodName, Class[] parameterTypes) {
     Method method = null;
     try {
+      if(clazz == null) {
```
carl-mastrangelo: why would clazz be null here?

Also, java style says add a space between `if` and `(`

kpayson64: ```getSuperclass()``` returns ```null``` for Object or other base types that don't have parents, so this serves as the base case.

I've fixed style issue.

okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/OptionalMethod.java
```diff
@@ -169,6 +169,12 @@ private Method getMethod(Class<?> clazz) {
   private static Method getPublicMethod(Class<?> clazz, String methodName, Class[] parameterTypes) {
     Method method = null;
     try {
+      if (clazz == null) {
+        return null;
+      }
+      if ((clazz.getModifiers() & Modifier.PUBLIC) == 0) {
+        return getPublicMethod(clazz.getSuperclass(), methodName, parameterTypes);
+      }
```
carl-mastrangelo: Sorry, could you explain what is going on more?  A private class with public methods should still be able to have the public methods called.  For example this works:

```
public interface Foo {
   void foo();
}
private class Bar implements Foo {
  @Override
  public void foo() {}
}

Foo f = new Bar();
f.foo();
```

Even though the class is itself private, public methods are still callable.  Can you link to the actual methods in conscrypt that were called?

kpayson64: Off the top of my head I don't remember the the exact function that caused the error.

However, if remove the changes to ```OptionalMethod.java```, I get the following result.

```
io.grpc.okhttp.OptionalMethodTest > invokeOptional FAILED
    java.lang.AssertionError: expected:<testArg> but was:<null>
        at org.junit.Assert.failNotEquals(Assert.java:743)
        at org.junit.Assert.assertEquals(Assert.java:118)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at io.grpc.okhttp.OptionalMethodTest.invokeOptional(OptionalMethodTest.java:96)

io.grpc.okhttp.OptionalMethodTest > isSupported FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:86)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at org.junit.Assert.assertFalse(Assert.java:64)
        at org.junit.Assert.assertFalse(Assert.java:74)
        at io.grpc.okhttp.OptionalMethodTest.isSupported(OptionalMethodTest.java:85)
```

The issue as I understand it only pops up when attempting to do reflection:
http://stackoverflow.com/questions/6161790/illegalaccessexception-when-using-call-dynamic-methods

PR #2904, opened 2017-04-12
Service Method Options (Protobuf)

lukaszx0:
Add support for service method options to `MethodDescriptor` and compiler.

Proto:

```protobuf
extend google.protobuf.MethodOptions {
  bool idempotent = 50001;
}

rpc EmptyCall(grpc.testing.Empty) returns (grpc.testing.Empty) {
  option (idempotent) = true;
};
```

Java runtime:

```java
((DescriptorProtos.MethodOptions) TestServiceGrpc.METHOD_EMPTY_CALL.getMethodOptions())
.getExtension(io.grpc.testing.integration.Messages.idempotent) => true
```

### DIFF
--- a/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
+++ b/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
@@ -26,7 +26,12 @@ public final class BenchmarkServiceGrpc {
 
   public static final String SERVICE_NAME = "grpc.testing.BenchmarkService";
 
+  private static final io.grpc.protobuf.ProtoDescriptorSupplier SERVICE_PROTO_DESCRIPTOR =
+    new BenchmarkServiceDescriptorSupplier();
+
   // Static method descriptors that strictly reflect the proto.
+  private static final int METHODIDX_UNARY_CALL = 0;
+
   @io.grpc.ExperimentalApi("https://github.com/grpc/grpc-java/issues/1901")
   public static final io.grpc.MethodDescriptor<io.grpc.benchmarks.proto.Messages.SimpleRequest,
       io.grpc.benchmarks.proto.Messages.SimpleResponse> METHOD_UNARY_CALL =
@@ -38,7 +43,12 @@ public final class BenchmarkServiceGrpc {
               io.grpc.benchmarks.proto.Messages.SimpleRequest.getDefaultInstance()))
           .setResponseMarshaller(io.grpc.protobuf.ProtoUtils.marshaller(
               io.grpc.benchmarks.proto.Messages.SimpleResponse.getDefaultInstance()))
+          .setMethodOptions(SERVICE_PROTO_DESCRIPTOR
+              .getMethodDescriptor(METHODIDX_UNARY_CALL).getOptions())
           .build();
+
+  private static final int METHODIDX_STREAMING_CALL = 1;
+
   @io.grpc.ExperimentalApi("https://github.com/grpc/grpc-java/issues/1901")
   public static final io.grpc.MethodDescriptor<io.grpc.benchmarks.proto.Messages.SimpleRequest,
       io.grpc.benchmarks.proto.Messages.SimpleResponse> METHOD_STREAMING_CALL =
@@ -50,6 +60,8 @@ public final class BenchmarkServiceGrpc {
               io.grpc.benchmarks.proto.Messages.SimpleRequest.getDefaultInstance()))
           .setResponseMarshaller(io.grpc.protobuf.ProtoUtils.marshaller(
               io.grpc.benchmarks.proto.Messages.SimpleResponse.getDefaultInstance()))
+          .setMethodOptions(SERVICE_PROTO_DESCRIPTOR
+              .getMethodDescriptor(METHODIDX_STREAMING_CALL).getOptions())
           .build();
 
   /**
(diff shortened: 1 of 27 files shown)


lukaszx0:
@carl-mastrangelo @ejona86 this is built on top of #2886. There are a few thing that we might want to polish but I'd like to know your opinion about this initial implementation first. I'm open to comments and suggestions. What do you think?

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
```diff
@@ -268,11 +280,21 @@ public void invoke(Req request, io.grpc.stub.StreamObserver<Resp> responseObserv
     }
   }
 
-  private static final class BenchmarkServiceDescriptorSupplier implements io.grpc.protobuf.ProtoFileDescriptorSupplier {
+  private static final class BenchmarkServiceDescriptorSupplier implements io.grpc.protobuf.ProtoDescriptorSupplier {
     @java.lang.Override
     public com.google.protobuf.Descriptors.FileDescriptor getFileDescriptor() {
       return io.grpc.benchmarks.proto.Services.getDescriptor();
```
ejona86: This call is expensive, because it loads a lot of descriptor information that is not necessary much of the time. The ProtoFileDescriptorSupplier was specifically designed to avoid loading that information when unnecessary.

Instead, I'd suggest copying what we did for the ServiceDescriptor: just place the BenchmarkServiceDescriptorSupplier instance on the object, and let the caller call `getFileDescriptor().findServiceByName(blah).findMethodByName(blah)` manually.

We could provide utility methods in `ProtoUtils` to do the searching though, since given a `MethodDescriptor` we know what values to pass to the `find*` methods.

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
```diff
@@ -268,11 +280,21 @@ public void invoke(Req request, io.grpc.stub.StreamObserver<Resp> responseObserv
...
     }
+
+    @java.lang.Override
+    public com.google.protobuf.Descriptors.ServiceDescriptor getServiceDescriptor() {
+      return getFileDescriptor().getServices().get(0);
```
ejona86: Hard-coding `0` doesn't work; there may be multiple services in the same file.

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
```diff
@@ -38,7 +43,12 @@ private BenchmarkServiceGrpc() {}
               io.grpc.benchmarks.proto.Messages.SimpleRequest.getDefaultInstance()))
           .setResponseMarshaller(io.grpc.protobuf.ProtoUtils.marshaller(
               io.grpc.benchmarks.proto.Messages.SimpleResponse.getDefaultInstance()))
+          .setMethodOptions(SERVICE_PROTO_DESCRIPTOR
+              .getMethodDescriptor(METHODIDX_UNARY_CALL).getOptions())
```
ejona86: We can't actually know the correct index for the method at compile time, because the protobuf codegen may be generated separately from the grpc codegen.

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
lukaszx0: Hmm okay. The main reason I've done this is because in order to get service/method descriptors you need to pass index. I guess adding new static `SERVICE_IDX` like we have one for methods + set of helpers in `ProtoUtils` would work. I'll change this.

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
lukaszx0: The index is hardcoded but it's not static and same for all. It uses the index of the service in the file, see https://github.com/grpc/grpc-java/pull/2904/files#diff-4e543a2f8a0243d4e0fed21737cdbf18R926

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
lukaszx0: I'm not sure I understand why? Even if the protobuf is generated separately, the indexes should remain the same - if they're different it'll mean the files are different and you have much bigger problems because it'll mean that the files are out-of-sync. I think I might be missing something, please explain.

ejona86: > the indexes should remain the same - if they're different it'll mean the files are different

Exactly. The files may be different. This is "okay" as long as the changes are backward-compatible. Adding a new method (in any order) is considered backward-compatible. But inserting a method like that would change the indexes.

Even if we were using the indexes directly, I'd want some sort of verification to fail when the index no longer matches. Otherwise it would be really painful to track down bugs caused by this.

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
ejona86: > The main reason I've done this is because in order to get service/method descriptors you need to pass index

How so? Using the `find*` methods doesn't. I thought my suggested code above would work fine without any indexes.

Even if the `find*` methods weren't available, we could write them ourselves in a utility by simply looping over the services/methods until we find the right one.

lukaszx0:
Closing in favor of #3086

PR #2910, opened 2017-04-13
netty,okhttp: handle ENHANCE_YOUR_CALM and too_many_pings

ejona86:
Any ENHANCE_YOUR_CALM is good reason to log, but too_many_pings also
triggers increase of the keepalive time for later connections.

### DIFF
--- a/core/src/main/java/io/grpc/internal/AtomicBackoff.java
+++ b/core/src/main/java/io/grpc/internal/AtomicBackoff.java
@@ -63,7 +63,7 @@ public final class AtomicBackoff {
   public final class State {
     private final long savedValue;
 
-    State(long value) {
+    private State(long value) {
       this.savedValue = value;
     }
 
--- a/core/src/main/java/io/grpc/internal/AtomicBackoff.java
+++ b/core/src/main/java/io/grpc/internal/AtomicBackoff.java
@@ -32,7 +32,7 @@
 package io.grpc.internal;
 
 import com.google.common.base.Preconditions;
-import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.concurrent.ThreadSafe;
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/grpc/internal/AtomicBackoff.java
```diff
@@ -0,0 +1,91 @@
...
+ * A {@code long} atomically updated due to errors caused by the value being too small.
+ */
+@ThreadSafe
+public final class AtomicBackoff {
+  private static final Logger log = Logger.getLogger(AtomicBackoff.class.getName());
+  private static final AtomicLongFieldUpdater<AtomicBackoff> valueUpdater =
+      AtomicLongFieldUpdater.newUpdater(AtomicBackoff.class, "value");
+
+  private final String name;
+  private volatile long value;
```
dapengzhang0: I would prefer a `private final AtomicLong` field to introducing the `static final AtomicLongFieldUpdater` here.

ejona86: Done.

core/src/main/java/io/grpc/internal/AtomicBackoff.java
```diff
@@ -0,0 +1,89 @@
...
+  /** Returns the current state. The state instance's value does not change of time. */
+  public State getState() {
+    return new State(value.get());
+  }
+
+  @ThreadSafe
+  public final class State {
+    private final long savedValue;
+
+    State(long value) {
```
dapengzhang0: private

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -1014,9 +1019,17 @@ public void ackSettings() {
 
     @Override
     public void goAway(int lastGoodStreamId, ErrorCode errorCode, ByteString debugData) {
+      if (errorCode == ErrorCode.ENHANCE_YOUR_CALM) {
+        String data = debugData.utf8();
+        log.log(Level.WARNING, String.format(
+            "%s: Received GOAWAY with ENHANCE_YOUR_CALM. Debug data: %s", this, data));
```
ejona86: This is the difference between `MessageFormat` and `String.format()`. logcat in Android generally doesn't handle the replacements, so we've worked around it by manually making sure the replacement happens before passing it to the logging infrastructure.

I'm pretty sure we could have used MessageFormat, but I didn't do that and I'm just being consistent with the rest of the code.

core/src/main/java/io/grpc/internal/AtomicBackoff.java
ejona86: Done. private can cause synthetic bridge methods to be created, but in this case I guess they won't (because the class is non-static). We're sort of ambivalent about bridge methods.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
dapengzhang0: yes, all I was asking is about `this`. Still LGTM

PR #2912, opened 2017-04-14
core: lookup TXT records when doing name resolution

carl-mastrangelo:


core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -41,25 +46,39 @@
...
+import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.GuardedBy;
+import javax.naming.NamingEnumeration;
```
ejona86: It doesn't seem like Android has javax.naming. Or at least the older ones don't. Have you tested on Android?

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -226,4 +237,163 @@ public final synchronized void shutdown() {
...
+
+  /**
+   * A resolver that uses the JNDI interface.  This class is capable of looking up both addresses
+   * and text records, but does not provide ordering guarantees.  When the {@link JdkResolver} is
+   * not available, this can be used as a fallback.  Some results are better than no results.
+   */
+  @VisibleForTesting
+  static final class JndiResolver extends DelegateResolver {
+
+    private static final String[] rrTypes = new String[]{"A", "AAAA", "TXT"};
```
ejona86: Can we just request TXT? You mentioning that ANY was used makes me nervous, and it would be rare to use the A and AAAA here. Let's just KISS and only query for what we need.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -226,4 +237,163 @@ public final synchronized void shutdown() {
...
+    ResolutionResults resolve(String host) throws Exception {
+      return new ResolutionResults(
+          Arrays.asList(InetAddress.getAllByName(host)),
+          Collections.<String>emptyList());
+    }
+  }
+
+  /**
+   * A resolver that uses the JNDI interface.  This class is capable of looking up both addresses
+   * and text records, but does not provide ordering guarantees.  When the {@link JdkResolver} is
```
carl-mastrangelo: It might not be prefered.  I don't think it can use nsAuthority, while the jndi one can.  Someone might prefer the latter behavior.  

Removed anyways.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
carl-mastrangelo: it was when I looked last :/

https://android.googlesource.com/platform/libcore/+/42cdd0bf7251074df8bf6dfa7f5d62a01c0a1ef6/ojluni/src/main/java/com/sun/jndi/dns/DnsContextFactory.java

ejona86: I saw the DnsContextFactory, which would imply that it does, but I also see things like this though: https://issues.apache.org/jira/browse/LOG4J2-703

So that may be newer versions have it. But I also didn't see javax.naming at https://developer.android.com/reference/packages.html

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -226,4 +234,153 @@ public final synchronized void shutdown() {
...
+
+  /**
+   * A composite DNS resolver that uses both the JDK and JNDI resolvers as delegate.  It is
+   * expected that two DNS queries will be executed, with the second one being from JNDI.
+   */
+  @VisibleForTesting
+  static final class CompositeResolver extends DelegateResolver {
+
+    private final DelegateResolver jdkResovler;
+    @Nullable private final DelegateResolver jndiResovler;
```
ejona86: Make jndiResolver required? It seems you only use this if it is available.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
carl-mastrangelo: Looks like it is absent.  Progaurd complains about refs to missing javax.naming classes, but not the com.sun ones.  I think if I check for both, it will properly disable itself.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
carl-mastrangelo: Done.

I feel like you are pushing into premature simplification.  This code was very easy to modify to accept results from other, future resolvers, but now it is going to be harder.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -149,10 +163,10 @@ public void run() {
             savedListener.onAddresses(Collections.singletonList(server), Attributes.EMPTY);
             return;
           }
-
+          ResolutionResults resolvedInetAddrs;
           try {
-            inetAddrs = getAllByName(host);
-          } catch (UnknownHostException e) {
+            resolvedInetAddrs = delegateResolver.resolve(host);
```
ejona86: So the TXT doesn't go anywhere?

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -226,4 +233,146 @@ public final synchronized void shutdown() {
...
+   * and text records, but does not provide ordering guarantees.  When the {@link JdkResolver} is
+   * not available, this can be used as a fallback.  Some results are better than no results.
+   */
+  @VisibleForTesting
+  static final class JndiResolver extends DelegateResolver {
+
+    private static final String[] rrTypes = new String[]{"TXT"};
+
+    @Override
+    ResolutionResults resolve(String host) throws NamingException {
```
ejona86: Do we want to catch any exceptions from this, so that a failure to get TXT doesn't hose everything?

We may want to discuss with others what should happen if we get A/AAAA but have I/O errors durring TXT.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
carl-mastrangelo: Currently No.  There are no consumers yet.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
carl-mastrangelo: I added it, but it should be fairly loud.  Not honoring the service config if present is something a user would want to fix.  It shouldn't reach this point if the JNDI resolver isn't present.

carl-mastrangelo:
PTAL

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -226,4 +233,150 @@ public final synchronized void shutdown() {
...
+    }
+  }
+
+  /**
+   * The default name resolver provided with the JDK.  This is unable to lookup TXT records, but
+   * provides address ordering sorted according to RFC 3484.  This is true on OpenJDK, because it
+   * in turn calls into libc which sorts addresses in order of reachability.
+   */
+  @VisibleForTesting
+  static final class JdkResolver extends DelegateResolver {
```
zhangkun83: This could be just a singleton object.

ejona86:
@carl-mastrangelo, what about my "but until the TXT is used for something, I'm not sure we want to do the I/O" comment?

core/src/main/java/io/grpc/internal/DnsNameResolver.java
carl-mastrangelo: Ack

carl-mastrangelo:
@ejona86  good point.  I added a boolean to disable it.  It won't see any any actual usage outside of a test.

PR #2914, opened 2017-04-14
core: record stats for GET requests

### LINKED ISSUE #2896: AbstractClientStream2.GetFramer doesn't use statsTraceCtx
GetFramer doesn't use the normal MessageFramer, so it should probably call statsTraceCtx and report the request message size and similar.

zhangkun83:
Also define the expected call sites of StatsTraceContext.

Resolves #2896

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractClientStream2.java
+++ b/core/src/main/java/io/grpc/internal/AbstractClientStream2.java
@@ -354,21 +354,21 @@ public abstract class AbstractClientStream2 extends AbstractStream2
       Preconditions.checkState(payload == null, "writePayload should not be called multiple times");
       try {
         payload = IoUtils.toByteArray(message);
-        statsTraceCtx.outboundMessage();
-        statsTraceCtx.outboundUncompressedSize(payload.length);
-        // NB(zhangkun83): this is not accurate, because the underlying transport will probably
-        // encode it using e.g., base64.  However, we are not supposed to know such detail here.
-        //
-        // We don't want to move this line to where the encoding happens either, because we'd better
-        // contain the message stats reporting in Framer as suggested in StatsTraceContext.
-        // Scattering the reporting sites increases the risk of mis-counting or double-counting.
-        //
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/grpc/internal/AbstractClientStream2.java
```diff
@@ -354,6 +354,18 @@ public void writePayload(InputStream message) {
...
+        statsTraceCtx.outboundMessage();
+        statsTraceCtx.outboundUncompressedSize(payload.length);
+        // NB(zhangkun83): this is not accurate, because the underlying transport will probably
+        // encode it using e.g., base64.  However, we are not supposed to know such detail here.
+        //
+        // We don't want to move this line to where the encoding happens either, because we'd better
+        // contain the message stats reporting in Framer as suggested in StatsTraceContext.
+        // Scattering the reporting sites increases the risk of mis-counting or double-counting.
+        //
+        // Because the payload is usually very small, and people shouldn't care about the size
```
ejona86: s/and//

core/src/main/java/io/grpc/internal/AbstractClientStream2.java
```diff
@@ -354,6 +354,18 @@ public void writePayload(InputStream message) {
       Preconditions.checkState(payload == null, "writePayload should not be called multiple times");
       try {
         payload = IoUtils.toByteArray(message);
+        statsTraceCtx.outboundMessage();
```
ejona86: Move after the try-catch?

core/src/main/java/io/grpc/internal/StatsTraceContext.java
```diff
@@ -104,6 +104,8 @@ public static StatsTraceContext newServerContext(
 
   /**
    * See {@link ClientStreamTracer#headersSent}.  For client-side only.
+   *
+   * <p>Transport-specific thus should be called by transport implementations.
```
ejona86: Add a comma before thus?

zhangkun83:
Thanks! All suggestions applied.

PR #2916, opened 2017-04-14
core: clamp keepAliveTime/Timeout in a way easier for test

dapengzhang0:


core/src/main/java/io/grpc/internal/KeepAliveManager.java
```diff
@@ -86,7 +86,10 @@ public void run() {
               TimeUnit.NANOSECONDS);
         } else if (state == State.PING_DELAYED) {
           // We have received some data. Reschedule the ping with the new time.
-          pingFuture = scheduler.schedule(sendPing, nextKeepaliveTime - ticker.read(),
+          pingFuture = scheduler.schedule(
+              sendPing,
+              // normalized as some Netty executor service does not accept a negative delay
```
ejona86: Can you file a bug? It is [part of the interface](http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ScheduledExecutorService.html):
> Zero and negative delays (but not periods) are also allowed in schedule methods

netty/src/test/java/io/grpc/netty/NettyClientTransportTest.java
```diff
@@ -465,8 +466,10 @@ private NettyClientTransport newTransport(ProtocolNegotiator negotiator, int max
     }
     NettyClientTransport transport = new NettyClientTransport(
         address, NioSocketChannel.class, new HashMap<ChannelOption<?>, Object>(), group, negotiator,
-        DEFAULT_WINDOW_SIZE, maxMsgSize, maxHeaderListSize, keepAliveTimeNano, 1L, false, authority,
-        userAgent);
+        DEFAULT_WINDOW_SIZE, maxMsgSize, maxHeaderListSize,
+        KeepAliveManager.clampKeepAliveTimeInNanos(keepAliveTimeNano),
```
ejona86: Why change the test? It seems like it should be allowed to avoid the clamping.

dapengzhang0: if `enableKeepAlive` is true, the original nano values are too small. `keepAliveTimeout` being 1 nano definitely fails the test.

core/src/main/java/io/grpc/internal/KeepAliveManager.java
dapengzhang0: filed netty/netty#6627

netty/src/test/java/io/grpc/netty/NettyClientTransportTest.java
ejona86: I don't follow. I only see values `1000` and INF used. If `1000` is too small, make it larger.

PR #2917, opened 2017-04-14
core: delete the alias LoadBalancer2.

### LINKED ISSUE #2656: Switch to LBv2

zhangkun83:
Resolves #2656

This will be backported to the 1.3.x branch.


ejona86:
Backported in #2919

PR #2918, opened 2017-04-15
core: document that NameResolver is not thread-safe.

### LINKED ISSUE #2649: NameResolver can be NotThreadSafe
With `ManagedChannelImpl2`, `NameResolver` is always called from Channel Executor, except for `getAuthority()`. After `ManagedChannelImpl2` is promoted, `NameResolver` can get rid of `@ThreadSafe`.

zhangkun83:
Resolves #2649.  This makes resolvers easier to implement.

### DIFF
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -49,7 +49,6 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
-import javax.annotation.concurrent.GuardedBy;
 
 /**
  * A DNS-based {@link NameResolver}.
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -130,18 +123,16 @@ public final synchronized void refresh() {
...
-          if (resolutionTask != null) {
-            resolutionTask.cancel(false);
-            resolutionTask = null;
-          }
-          if (shutdown) {
-            return;
-          }
-          savedListener = listener;
-          resolving = true;
+        // If this task is started by refresh(), there might already be a scheduled task.
```
carl-mastrangelo: What happens if resolutionRunnable runs on a thread other than the original one it ran on.  I don't know if ChannelExecutor is guaranteed to be serial.

zhangkun83: Hmm, you are right. ChannelExecutor needs to be exposed to NameResolver so that tasks ran in an executor can be serialized with it. I will put this PR on hold until I have time to improve it.

zhangkun83:
Superseded by #5364

PR #2920, opened 2017-04-17
core: Add support for the AppEngine development sandbox

frew:
The App Engine development appserver imposes similar ClassLoader restrictions to the production environment. This patch allows for development of grpc using code in the development appserver. First grpc patch - please let me know if I missed any of the contributing guidelines!

### DIFF
--- a/core/src/main/java/io/grpc/internal/GrpcUtil.java
+++ b/core/src/main/java/io/grpc/internal/GrpcUtil.java
@@ -70,11 +70,15 @@ public final class GrpcUtil {
 
   public static final Charset US_ASCII = Charset.forName("US-ASCII");
 
+  private static boolean isAppengineEnvironment() {
+    String appengineEnvironment = System.getProperty("com.google.appengine.runtime.environment");
+    return "Production".equals(appengineEnvironment) || "Development".equals(appengineEnvironment);
+  }
+
   // Certain production AppEngine runtimes have constraints on threading and socket handling
   // that need to be accommodated.
   public static final boolean IS_RESTRICTED_APPENGINE =
-      "Production".equals(System.getProperty("com.google.appengine.runtime.environment"))
-          && "1.7".equals(System.getProperty("java.specification.version"));
+      isAppengineEnvironment() && "1.7".equals(System.getProperty("java.specification.version"));
 
   /**
    * {@link io.grpc.Metadata.Key} for the timeout header.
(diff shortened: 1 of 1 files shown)


frew:
I signed the individual CLA. Please let me know if it's not showing up!

core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -70,11 +70,15 @@
 
   public static final Charset US_ASCII = Charset.forName("US-ASCII");
 
+  private static boolean isAppengineEnvironment() {
+    String appengineEnvironment = System.getProperty("com.google.appengine.runtime.environment");
```
ejona86: com.google.appengine.runtime.environment is either Production or Development. It has no other values.

The previous check was added explicitly. @louiscryan, would it hurt to remove the check?

frew:
@ejona86 I agree that those are the two values that are present in the App Engine documentation. Please let me know if you would prefer me to change it to just check that it's set to something - I think the case can be made for either more or less specificity.

ejona86:
@frew, yeah, go ahead and just check for the presence of the appengine property. If the dev environment has the same restriction, then it seems we don't need to go about special casing any of the environments. Other than noticing we're on appengine itself.

frew:
@ejona86 Updated - PTAL. I kept the Java 7 check there since I don't know the history of it and it should be true for all App Engine environments at present since the product doesn't have Java 8 support yet.

core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -70,15 +70,11 @@
...
-    String appengineEnvironment = System.getProperty("com.google.appengine.runtime.environment");
-    return "Production".equals(appengineEnvironment) || "Development".equals(appengineEnvironment);
-  }
-
-  // Certain production AppEngine runtimes have constraints on threading and socket handling
+  // AppEngine runtimes have constraints on threading and socket handling
   // that need to be accommodated.
   public static final boolean IS_RESTRICTED_APPENGINE =
-      isAppengineEnvironment() && "1.7".equals(System.getProperty("java.specification.version"));
+      System.getProperty("com.google.appengine.runtime.environment") != null
```
louiscryan: FWIW - I did talk to the AppEngine folks about Java 8 support and the expectation is that the runtime classloader is more lenient.

@lesv  who keeps an eye on this stuff

ejona86:
Jenkins, okay to test

ejona86:
retest this please

ejona86:
:-/ Jenkins isn't hearing my cries for it to test this. Travis-CI passed and I'm okay with skipping the Windows test for this PR. Merging.

ejona86:
Thank you, @frew!

frew:
Awesome. Thanks for the prompt and thorough review, @ejona86, @carl-mastrangelo, and @louiscryan!

PR #2921, opened 2017-04-17
core: add inboundHeaders() to ClientStreamTracer.

zhangkun83:


core/src/main/java/io/grpc/ClientStreamTracer.java
```diff
@@ -46,6 +46,12 @@ public void headersSent() {
   }
```
ejona86: Should headersSent be outboundHeaders? I was going to suggest changing inboundHeaders until I realized it matched the methods in StreamTracer better. That can be done in a later PR.

zhangkun83: Renamed in the PR, because since the tracer API has not been released, it would be easier for me to back-port the rename to the 1.3.x branch right now than doing it after the release.

ejona86:
Backport was done in #2922

PR #2923, opened 2017-04-17
netty,okhttp: handle ENHANCE_YOUR_CALM and too_many_pings (backport to v1.3)

ejona86:
Any ENHANCE_YOUR_CALM is good reason to log, but too_many_pings also
triggers increase of the keepalive time for later connections.

CC @carl-mastrangelo

### DIFF
--- a/core/src/main/java/io/grpc/internal/AtomicBackoff.java
+++ b/core/src/main/java/io/grpc/internal/AtomicBackoff.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2017, Google Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *    * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ *    * Neither the name of Google Inc. nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/grpc/internal/AtomicBackoff.java
```diff
@@ -0,0 +1,89 @@
...
+     * Causes future invocations of {@link AtomicBackoff#getState} to have a value at least double
+     * this state's value. Subsequent calls to this method will not increase the value further.
+     */
+    public void backoff() {
+      // Use max to handle overflow
+      long newValue = Math.max(savedValue * 2, savedValue);
+      boolean swapped = value.compareAndSet(savedValue, newValue);
+      // Even if swapped is false, the current value should be at least as large as newValue
+      assert value.get() >= newValue;
+      if (swapped) {
```
carl-mastrangelo: protect behind an `isLoggable` ?

core/src/main/java/io/grpc/internal/AtomicBackoff.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Causes future invocations of {@link AtomicBackoff#getState} to have a value at least double
+     * this state's value. Subsequent calls to this method will not increase the value further.
+     */
+    public void backoff() {
+      // Use max to handle overflow
+      long newValue = Math.max(savedValue * 2, savedValue);
+      boolean swapped = value.compareAndSet(savedValue, newValue);
```
carl-mastrangelo: It might be worth your time to make this state mechanism pluggable with `ExponentialBackoffPolicy`.  You could use your threadsafe version for pings, and leave the unsafe version for Channel.

ejona86: `ExponentialBackoffPolicy` is not threadsafe and is not really intended for handling configuration mutations. This API makes sure that it knows which `value` caused the `backoff()`, so it doesn't double-account for broken `value`s. This also makes no sense to apply jitter, because there isn't any "retrying" going on. They seem to have completely different constraints and behavior.

core/src/main/java/io/grpc/internal/AtomicBackoff.java
ejona86: Meh. It's generally assumed WARNING is enabled, which means we're okay with the performance this may cause. Optimizing the error path for a rare configuration whose performance is already better than the default doesn't get me excited.

ejona86:
retest this please

PR #2927, opened 2017-04-18
Move client deframing to the call executor

ericgribkoff:
This PR moves client-side deframing out of the transport thread and onto the call executor. This lays the groundwork for full-stream compression by allowing the full-stream decompression, once implemented, to occur in an application-provided thread rather than slowing down the transport.

To accomplish this, the operation of the `MessageDeframer` class is split into two parts. `MessageDeframer.Sink` is invoked from the transport thread, and queues incoming raw data frames in preparation for deframing them as gRPC messages in a separate thread. `MessageDeframer.Source` is invoked in the client call executor thread, and handles the actual deframing and our existing per-message decompression.

The `Sink` and `Source` of the `MessageDeframer` are safe to run concurrently in separate threads. The `CompositeReadableBuffer` holding raw incoming data frames is now backed by a `ConcurrentLinkedList`. Since we have exactly one thread (`MessageDeframer.Sink` in the transport thread) adding to the buffer and one thread (`MessageDeframer.Source` in the serializing call executor) reading from the buffer, `ConcurrentLinkedList` is a bit more than necessary for thread-safety, but initial benchmarks indicate that the primary runtime cost is the context switching between the transport and application threads rather than the lockless synchronization in Java's `ConcurrentLinkedList`. If it becomes the bottleneck, we can move to a data structure optimized for this one-producer, one-consumer scenario.

This PR combines the two existing deframer-to-transport notifications, `deliveryStalled()` and `endOfStream()`, into a single method: `deframerClosed()`. Previously, `deliveryStalled()` was ignored when deframing for a server, and only had an impact on the client side when `transportReportStatus()` had already been called. On the other hand, `endOfStream()` was only ever called by the deframer on the server side, as the `endOfStream` argument to `MessageDeframer#deframe(ReadableBuffer, boolean endOfStream)` was only set to true in `AbstractServerStream2`. Combining these two client- and server-specific notifications into one method, `deframerClosed()` is only invoked after the transport thread has notified the deframer that it should close. This can be in response to a normal event, such as receiving trailing metadata: in this case, the deframer will continue to send queued messages to the application and close once done. Or it can be due to an immediate close request, such as in response to a cancellation from the application code: this will cause the deframer to close and discard any remaining messages.

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractClientStream2.java
+++ b/core/src/main/java/io/grpc/internal/AbstractClientStream2.java
@@ -56,7 +56,7 @@ public abstract class AbstractClientStream2 extends AbstractStream2
    * collisions/confusion. Only called from application thread.
    */
   protected interface Sink {
-    /** 
+    /**
      * Sends the request headers to the remote end point.
      *
      * @param metadata the metadata to be sent
@@ -181,14 +181,17 @@ public abstract class AbstractClientStream2 extends AbstractStream2
     return super.isReady() && !cancelled;
   }
 
-  /** This should only called from the transport thread. */
+  /**
+   * {@link MessageDeframer.Source.Listener} methods will be called from the deframing thread.
+   * Other methods should only be called from the transport thread.
+   */
   protected abstract static class TransportState extends AbstractStream2.TransportState {
     /** Whether listener.closed() has been called. */
     private final StatsTraceContext statsTraceCtx;
     private boolean listenerClosed;
     private ClientStreamListener listener;
 
-    private Runnable deliveryStalledTask;
+    private Runnable deframerClosedTask;
 
     /**
      * Whether the stream is closed from the transport's perspective. This can differ from {@link
@@ -205,19 +208,6 @@ public abstract class AbstractClientStream2 extends AbstractStream2
     public final void setListener(ClientStreamListener listener) {
       Preconditions.checkState(this.listener == null, "Already called setListener");
       this.listener = Preconditions.checkNotNull(listener, "listener");
-    }
-
-    @Override
-    public final void deliveryStalled() {
-      if (deliveryStalledTask != null) {
-        deliveryStalledTask.run();
-        deliveryStalledTask = null;
-      }
-    }
-
-    @Override
-    public final void endOfStream() {
-      deliveryStalled();
     }
 
     @Override
@@ -251,7 +241,7 @@ public abstract class AbstractClientStream2 extends AbstractStream2
         }
 
         needToCloseFrame = false;
-        deframe(frame, false);
+        deframe(frame);
       } finally {
         if (needToCloseFrame) {
           frame.close();
@@ -280,37 +270,50 @@ public abstract class AbstractClientStream2 extends AbstractStream2
      * Report stream closure with status to the application layer if not already reported. This
      * method must be called from the transport thread.
      *
+     * <p>No further data may be sent to the deframer after this method is called. However, if
+     * {@link MessageDeframer.Source} is running in another thread, the client will receive any
+     * already queued messages before the deframer closes, unless {@code stopDelivery} is true.
+     * Even with {@code stopDelivery=true}, there is an inherent race condition so the client may
+     * still receive a pending message before the close takes effect.
+     *
      * @param status the new status to set
-     * @param stopDelivery if {@code true}, interrupts any further delivery of inbound messages that
-     *        may already be queued up in the deframer. If {@code false}, the listener will be
-     *        notified immediately after all currently completed messages in the deframer have been
-     *        delivered to the application.
+     * @param stopDelivery if {@code true}, interrupt {@link MessageDeframer.Source} even if it has
+     *     additional queued messages
      * @param trailers new instance of {@code Trailers}, either empty or those returned by the
-     *        server
+     *     server
      */
     public final void transportReportStatus(final Status status, boolean stopDelivery,
         final Metadata trailers) {
       Preconditions.checkNotNull(status, "status");
       Preconditions.checkNotNull(trailers, "trailers");
-      // If stopDelivery, we continue in case previous invocation is waiting for stall
       if (statusReported && !stopDelivery) {
         return;
       }
       statusReported = true;
       onStreamDeallocated();
 
-      // If not stopping delivery, then we must wait until the deframer is stalled (i.e., it has no
-      // complete messages to deliver).
-      if (stopDelivery || isDeframerStalled()) {
-        deliveryStalledTask = null;
-        closeListener(status, trailers);
-      } else {
-        deliveryStalledTask = new Runnable() {
-          @Override
-          public void run() {
-            closeListener(status, trailers);
-          }
-        };
+      deframerClosedTask =
+          new Runnable() {
+            @Override
+            public void run() {
+              closeListener(status, trailers);
+            }
+          };
+      if (!isDeframerScheduledToClose()) {
+        scheduleDeframerClose(stopDelivery);
+      }
+    }
+
+    /**
+     * This is the logic for listening for
+     * {@link MessageDeframer.Source.Listener#deframerClosed(boolean)}, which is invoked from the
+     * deframing thread. Subclasses must invoke this message from the transport thread.
+     */
+    protected void deframerClosedNotThreadSafe() {
+      Preconditions.checkState(deframerClosedTask != null, "deframerClosedTask is null");
+      if (deframerClosedTask != null) {
+        deframerClosedTask.run();
+        deframerClosedTask = null;
       }
     }
 
@@ -320,9 +323,9 @@ public abstract class AbstractClientStream2 extends AbstractStream2
      * @throws IllegalStateException if the call has not yet been started.
      */
     private void closeListener(Status status, Metadata trailers) {
+      Preconditions.checkState(isDeframerScheduledToClose(), "deframe not scheduled to close");
       if (!listenerClosed) {
         listenerClosed = true;
-        closeDeframer();
         statsTraceCtx.streamClosed(status);
         listener().closed(status, trailers);
       }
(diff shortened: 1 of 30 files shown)


ericgribkoff:
Jenkins: retest this please

ericgribkoff:
Jenkins: test this please

ericgribkoff:
@ejona86 I updated the PR via a force-push, following basically the same approach as the earlier, in-progress code but simplified and better organized. This is ready for review now.

ericgribkoff:
I just pushed an update that corrects some of the performance regression seen after moving client deframing to the app thread (the fix for the remainder of the regression is discussed below). 
 
Instead of waiting to notify the transport of bytes read until after the client thread actually consumes them (analogous to calls to `#deliver` from the old `MessageDeframer` implementation), `MessageDeframer.Sink#deframe` (running in the transport thread) now notifies the transport layer that the bytes have been read immediately. This avoids a source of slowdown associated with the earlier commits in the PR: the transport wouldn't be aware that bytes had been consumed until after a small but significant delay before the client-thread deframing occurred, followed by another delay to schedule that notification from the client thread back onto the transport thread. The change in the latest commit eagerly notifies the transport that the bytes have been read as soon as they are consumed: this is a behavior change when the client application has not yet `request()`ed the message (tests updated to reflect that), so it will take some discussion about whether this is semantically correct behavior.
 
The second, and more significant, source of slowdown with deframing in the client thread is due to the large number of notifications from `MessageDeframer.Sink#deframe` to `#scheduleDeframerSource`, which runs in the client thread. Since the transport thread queuing the data in `MessageDeframer.Sink#deframe` doesn't know if the http2 frames it's queuing up contain a full gRPC message or not, it has to schedule the `MessageDeframer.Source#next` to run in the client thread every time additional data comes in. For larger messages, this can occur on the order of 10K-100K times per gRPC message. Each time, the client thread has to execute a new runnable task that essentially just copies a few bytes from one buffer (`unprocessed`) into another (`nextFrame`).
 
For unary messages, we can "correct" for this effect by only scheduling the client thread once the transport thread signals the deframer to close, as this indicates the entire message has been received and it's time to "wake up" the client thread to finish deframing. In benchmarks with unary RPCs, this completely eliminates the performance gap between the existing transport-thread deframing and this PR's client-thread deframing.
 
I'm not sure how, or if, we can accomplish the same effect for non-unary messages. For the benchmarks I ran, I simulated this apriori knowledge by hard-coding the fixed message size into the deframer and only notifying the client when a full message was available. This improved performance to as good or better than deframing in the transport thread. To do this for real, we could do what C core is doing for lazy deframing, where the transport thread actually reads the gRPC message header, and then we'd know how many bytes to queue up in the deframer's `unprocessed` buffer before scheduling work on the client thread. @ejona86 and I discussed this briefly earlier this week and wanted to avoid this approach; reading the header in the transport thread won't work anyways when we have full-stream compression (the ultimate point of this work). There might be a better solution (heuristics for when to schedule work on the client, or just continuing to deframe in the transport thread for non-compressed, streaming RPCs).
 
For reference, all of the relative performances between transport-thread-deframing and client-thread-deframing mentioned above were derived from running the qps_client/qps_server in the benchmarks/ directory on my desktop, with the client's deframing occurring in the call executor and the server continuing to deframe in the transport thread. The original PR saw worse performance than the master branch across the board: scenarios with message payloads from 0 bytes to 10MB, 40 to 100+ outstanding RPCs across 4-10 channels, and streaming or unary RPCs. The fixes described above eliminate all of these regressions, getting equal or better performance with deframing in the transport thread, although in the streaming case it requires a bit of hackery to know in advance the size of the incoming message.
 
My next step is measuring performance with compressed data, both the existing single message compression and compressed streams (the point of this whole thing).
 
@ejona86 Do you still expect to have time to review this PR this week?

ejona86:
@ericgribkoff, no I'm not expecting to have time this week... err.. today. (Although... maybe you'll be lucky).

> `MessageDeframer.Sink#deframe` (running in the transport thread) now notifies the transport layer that the bytes have been read immediately

Uhh... Doesn't that disable flow control?

> The second, and more significant, source of slowdown with deframing in the client thread is due to the large number of notifications from `MessageDeframer.Sink#deframe` to `#scheduleDeframerSource`, which runs in the client thread.

This isn't surprising, but it will be a bit hard to fix. I'd suggest avoiding calling execute() with a Runnable if one is already scheduled and using our own single-producer, single-consumer queue, but that's effectively the same as what SerializingExecutor is doing today (since #2858). So we're already doing that trick.

We could probably reduce some of the costs with Netty by waiting to notify the application thread until onReadComplete. That will batch more frames together. But that solution won't help OkHttp and Cronet. With effort, Cronet could maybe get a similar optimization since it should be doing nonblocking I/O already. But it's probably a bit of a large request from Cronet. Doing an experiment with Netty to see how much we gain is probably worthwhile though.


I'm not really concerned at all about the performance of compressed messages at this point. Per-message compression should not have any real performance impact. My main performance concern was avoiding hurting the non-compressed case, which is higher-performance today.

ericgribkoff:
> > MessageDeframer.Sink#deframe (running in the transport thread) now notifies the transport layer that the bytes have been read immediately
>
> Uhh... Doesn't that disable flow control?

Yes, good point :) My goal was to isolate the cause of the slowdown on the benchmarks, where we get a single message at a time that's already been requested by the client, so this won't generalize. Probably could have just described the change rather than pushing an actual commit... but for the benchmarks, I think this functions okay as a hacky stand-in for a `Http2LocalFlowController` that will let us "return" the bytes of the gRPC message as soon as the header is consumed (ie, before the message bytes have necessarily arrived over the wire, because we know the application code has already requested the message).

ericgribkoff:
I played around a bit more with when bytes are returned to the transport and making changes here is actually not necessary to achieve performance parity. The overwhelmingly dominant cost is the repeated notifications to the client thread when a complete gRPC message is not yet available.

> We could probably reduce some of the costs with Netty by waiting to notify the application thread until onReadComplete. That will batch more frames together. But that solution won't help OkHttp and Cronet. With effort, Cronet could maybe get a similar optimization since it should be doing nonblocking I/O already. But it's probably a bit of a large request from Cronet. Doing an experiment with Netty to see how much we gain is probably worthwhile though.

I quickly searched for `onReadComplete` but didn't find that in the Netty API. Am I looking in the wrong place or did you mean a different function?

Regardless, since the only performance concern right now is deframing non-compressed messages: I'm not sure it's worth investigating an approach that will only benefit Netty and possibly Cronet and will still incur some increased overhead, when we can instead just follow C++'s approach and deframe the gRPC header in the transport thread. This lets us make a somewhat informed decision about whether to continue to deframe in the transport thread or offload to the client thread, as well as giving us performance parity with the existing transport-thread only deframing. What are your thoughts about this?

ejona86:
> I quickly searched for onReadComplete but didn't find that in the Netty API.

It's name is actually [`channelReadComplete()`](http://netty.io/4.1/api/io/netty/channel/ChannelInboundHandler.html#channelReadComplete-io.netty.channel.ChannelHandlerContext-).

> I'm not sure it's worth investigating an approach that will only benefit Netty and possibly Cronet and will still incur some increased overhead, when we can instead just follow C++'s approach and deframe the gRPC header in the transport thread.

Because deframing the header in the transport thread is neither here nor there. That is what we are doing today. Why merge this PR then? If there's a plan for stage one and stage two, that could work, but without an idea of where we are going how do we know if this change is an improvement?

The only way that is okay is if we only deframe on the network thread is when there is no compression. If you want to do it one way with compression and another without, that can work, but will be interesting from a synchronization and code health standpoint.

> I'm not sure it's worth investigating an approach that will only benefit Netty and possibly Cronet and will still incur some increased overhead

I've not heard other alternatives that may work.

But it's also not as bad as it may seem, because OkHttp has a different threading model that is closer to C's and so may not see much performance impact already. Cronet's threading model would suffer, but it may not be an issue at the throughputs expected when using Cronet. With Cronet we probably want to care more about CPU usage and the network links aren't 40+ Gbps, which means it may be acceptable to decompress on the network thread.

> This lets us make a somewhat informed decision about whether to continue to deframe in the transport thread or offload to the client thread, as well as giving us performance parity with the existing transport-thread only deframing.

I don't really understand this. How could we decompress messages on the network thread without strongly impacting performance? C does not have a "network thread." If I trust the internet, gzip decompression is on the order of 100 MB/s. Making decompression serial on the network thread seems like a bad idea.

ericgribkoff:
>> I'm not sure it's worth investigating an approach that will only benefit Netty and possibly Cronet and will still incur some increased overhead, when we can instead just follow C++'s approach and deframe the gRPC header in the transport thread.
>
>Because deframing the header in the transport thread is neither here nor there. That is what we are doing today. Why merge this PR then? If there's a plan for stage one and stage two, that could work, but without an idea of where we are going how do we know if this change is an improvement?
 
I don't really understand this. We deframe the header in the transport thread now, but not the message itself. 

My understanding of the "why" for this PR is to set things up for full-stream decompression. This requires disentangling the current single-threaded interaction between AbstractStream2 and MessageDeframer so that decompression can be moved into the client thread. Since we want to allow this without hurting performance in the existing benchmarks, I was suggesting we continue to read the header in the transport thread when the stream is not compressed.
 
>The only way that is okay is if we only deframe on the network thread is when there is no compression. If you want to do it one way with compression and another without, that can work, but will be interesting from a synchronization and code health standpoint.
 
Why can't the header always be read in the transport thread (again, in the existing case where we do not have full-stream compression) and the message body be read in the application thread?
 
>> This lets us make a somewhat informed decision about whether to continue to deframe in the transport thread or offload to the client thread, as well as giving us performance parity with the existing transport-thread only deframing.
>
> I don't really understand this. How could we decompress messages on the network thread without strongly impacting performance? C does not have a "network thread." If I trust the internet, gzip decompression is on the order of 100 MB/s. Making decompression serial on the network thread seems like a bad idea.
 
I guess I wasn't clear. I was only talking about reading the non-compressed header in the network thread,  as we do now; I was not advocating that we perform the decompression in the network thread.

ericgribkoff:
Closed in favor of https://github.com/grpc/grpc-java/pull/3145

PR #2928, opened 2017-04-19
netty: more coverage on server keepalive

dapengzhang0:
cover the missed coverage of `ping` and `onPingTimeout` in `NettyServerHandler`

### DIFF
--- a/netty/src/test/java/io/grpc/netty/NettyServerHandlerTest.java
+++ b/netty/src/test/java/io/grpc/netty/NettyServerHandlerTest.java
@@ -98,11 +98,8 @@ import java.io.InputStream;
 import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
-import org.junit.Before;
 import org.junit.Ignore;
-import org.junit.Rule;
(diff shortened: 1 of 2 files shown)


netty/src/test/java/io/grpc/netty/NettyServerHandlerTest.java
```diff
@@ -147,16 +160,28 @@ public void transportTerminated() {
     }
   }
 
+  /**
+   * Will not run {@code setUp()} automatically if the test name contains "customSetUp".
+   */
   @Before
-  public void setUp() throws Exception {
+  public void maybeRunSetup() throws Exception {
+    if (!testName.getMethodName().contains("customSetUp")) {
```
ejona86: I don't feel great about this, because its magical and can easily make things harder to debug ("why does that test method work and not this one?"). Since it seems so many tests already call `setUp()`, let's remove the `@Before` and have each test call `setUp()` explicitly.

dapengzhang0: done

PR #2933, opened 2017-04-20
Add cacheable_unary interop test

ericgribkoff:
This PR has two commits:

1. Remove the `grpc-payload-bin=` prefix from outgoing GET requests. This brings the implementation inline with C++ and our test server behind GFE.
2. Add the cacheable_unary interop test, which verifies that GET requests sent to a server behind a caching proxy are correctly sent and responses cached.

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -35,6 +35,7 @@ import static com.google.common.truth.Truth.assertThat;
 import static io.grpc.testing.integration.Messages.PayloadType.COMPRESSABLE;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
@@ -65,6 +66,7 @@ import io.grpc.CallOptions;
 import io.grpc.Channel;
 import io.grpc.ClientCall;
 import io.grpc.ClientInterceptor;
+import io.grpc.ClientInterceptors;
 import io.grpc.ClientStreamTracer;
 import io.grpc.Context;
 import io.grpc.Grpc;
@@ -85,6 +87,7 @@ import io.grpc.internal.GrpcUtil;
 import io.grpc.internal.testing.StatsTestUtils.FakeStatsContextFactory;
 import io.grpc.internal.testing.StatsTestUtils.MetricsRecord;
 import io.grpc.protobuf.ProtoUtils;
+import io.grpc.stub.ClientCalls;
 import io.grpc.stub.MetadataUtils;
 import io.grpc.stub.StreamObserver;
 import io.grpc.testing.StreamRecorder;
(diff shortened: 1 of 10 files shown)


netty/src/main/java/io/grpc/netty/NettyClientStream.java
```diff
@@ -128,8 +128,8 @@ public void writeHeaders(Metadata headers, byte[] requestPayload) {
       AsciiString httpMethod;
       if (get) {
         // Forge the query string
-        defaultPath = new AsciiString(defaultPath + "?" + GRPC_PAYLOAD_BIN_KEY + "="
-            + BaseEncoding.base64().encode(requestPayload));
+        defaultPath =
+            new AsciiString(defaultPath + "?" + BaseEncoding.base64().encode(requestPayload));
```
ejona86: Add a TODO for adding back the key.

interop-testing/src/test/java/io/grpc/testing/integration/TestCasesTest.java
```diff
@@ -54,11 +54,27 @@ public void unknownStringThrowsException() {
...
+      "client_streaming",
+      "server_streaming",
+      "ping_pong",
+      "empty_stream",
+      "compute_engine_creds",
+      "service_account_creds",
+      "jwt_token_creds",
+      "oauth2_auth_token",
+      "per_rpc_creds",
+      "cacheable_unary",
```
ejona86: This is not in the right order. It appears it should be under empty_unary to match the interop spec. It needs to be fixed elsewhere as well.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1330,6 +1333,67 @@ public void perRpcCreds(String jsonKey, InputStream credentialsStream, String oa
     oauth2AuthToken(jsonKey, credentialsStream, oauthScope);
   }
 
+  /** Sends a cacheable unary rpc using GET. Requires that the server is behind a caching proxy. */
+  public void cacheableUnary() {
+    // Build method descriptor manually to set idempotent and safe to true.
+    io.grpc.MethodDescriptor<
```
ejona86: Why all this stuff? Why isn't CacheableUnaryCall in the proto? If it were, then I'd expect `TestServiceGrpc.METHOD_CACHEABLE_UNARY_CALL.toBuilder().setSafe(true).build()` here.

If we're going to have all this goo here, it seems at least we should remove all the unnecessary fully-qualified names (like io.grpc before MethodDescriptor).

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1330,6 +1333,67 @@ public void perRpcCreds(String jsonKey, InputStream credentialsStream, String oa
...
+                .build();
+    // Set fake user IP since some proxies (GFE) won't cache requests from localhost.
+    Metadata.Key<String> userIpKey = Metadata.Key.of("x-user-ip", Metadata.ASCII_STRING_MARSHALLER);
+    Metadata metadata = new Metadata();
+    metadata.put(userIpKey, "1.2.3.4");
+    Channel channelWithUserIpKey =
+        ClientInterceptors.intercept(channel, MetadataUtils.newAttachHeadersInterceptor(metadata));
+    SimpleRequest requests1And2 =
+        SimpleRequest.newBuilder()
+            .setResponseSize(1)
```
ejona86: Why is this requesting response details if it is just going to be thrown away? It doesn't say this should be sent in the interop spec.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1330,6 +1333,67 @@ public void perRpcCreds(String jsonKey, InputStream credentialsStream, String oa
...
+                    .setBody(ByteString.copyFromUtf8(String.valueOf(System.nanoTime()))))
+            .build();
+
+    SimpleResponse response1 =
+        ClientCalls.blockingUnaryCall(
+            channelWithUserIpKey, safeCacheableUnaryCallMethod, CallOptions.DEFAULT, requests1And2);
+    SimpleResponse response2 =
+        ClientCalls.blockingUnaryCall(
+            channelWithUserIpKey, safeCacheableUnaryCallMethod, CallOptions.DEFAULT, requests1And2);
+    SimpleResponse response3 =
```
ejona86: This request is not in the interop test specification.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
ericgribkoff: My mistake. I didn't know toBuilder() existed. Fixed.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
ericgribkoff: The interop spec is going to be changed to allow validating that the server received the payload. Removed for now.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
ericgribkoff: The interop test specification needs updating, which I was holding off on to do along with adding the verificiation of the payload as mentioned above. See https://github.com/grpc/grpc/blob/d93a8bead39ab112b808a248adb4576e4c9ce766/test/cpp/interop/interop_client.cc#L921 for the desired behavior.

interop-testing/src/test/java/io/grpc/testing/integration/TestCasesTest.java
ericgribkoff: I didn't realize these were ordered according to the interop spec. Fixed.

netty/src/main/java/io/grpc/netty/NettyClientStream.java
ericgribkoff: Done

testing-proto/src/main/proto/io/grpc/testing/integration/test.proto
```diff
@@ -47,6 +47,11 @@ service TestService {
   // One request followed by one response.
   rpc UnaryCall(SimpleRequest) returns (SimpleResponse);
 
+  // One request followed by one response. Response has cache control
```
ejona86: Normally it is good to briefly sum up the important parts from the spec for the service implementation. Namely, that the request is ignored and the response is always different, except for caching based on the request.

ericgribkoff: This behavior is set to change soon, and the interop test protos have already diverged enough between here and grpc/grpc. Will update this comment once I add these files into grpc-proto and have a standardized location for the 
"ground truth".

PR #2938, opened 2017-04-21
core: integrate instrumentation-java (Census) tracing

zhangkun83:
Main implementation is in `CensusTracingModule`.

Also a few fix-ups in the stats implementation `CensusStatsModule`:
- Change header key name from `grpc-census-bin` to `grpc-tags-bin`
- Server does not fail on header parse errors. Uses the default instead.

DO NOT SUBMIT: currently using snapshot version of Census. Must switch to a release version before submitting.

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
@@ -176,13 +176,15 @@ public abstract class AbstractServerImplBuilder<T extends AbstractServerImplBuil
     StatsContextFactory statsFactory =
         this.statsFactory != null ? this.statsFactory : Stats.getStatsContextFactory();
     if (statsFactory != null) {
-      CensusStreamTracerModule census =
-          new CensusStreamTracerModule(
-              statsFactory, Tracing.getTracer(), Tracing.getBinaryPropagationHandler(),
-              GrpcUtil.STOPWATCH_SUPPLIER);
-      tracerFactories.add(census.getServerTracerFactory());
+      CensusStatsModule censusStats =
+          new CensusStatsModule(statsFactory, GrpcUtil.STOPWATCH_SUPPLIER);
+      tracerFactories.add(censusStats.getServerTracerFactory());
     }
+    CensusTracingModule censusTracing =
+        new CensusTracingModule(Tracing.getTracer(), Tracing.getBinaryPropagationHandler());
+    tracerFactories.add(censusTracing.getServerTracerFactory());
     tracerFactories.addAll(streamTracerFactories);
+
     io.grpc.internal.InternalServer transportServer =
         buildTransportServer(Collections.unmodifiableList(tracerFactories));
     ServerImpl server = new ServerImpl(getExecutorPool(),
--- a/core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
@@ -173,16 +173,20 @@ public abstract class AbstractServerImplBuilder<T extends AbstractServerImplBuil
   public ServerImpl build() {
     ArrayList<ServerStreamTracer.Factory> tracerFactories =
         new ArrayList<ServerStreamTracer.Factory>();
-    StatsContextFactory statsFactory =
-        this.statsFactory != null ? this.statsFactory : Stats.getStatsContextFactory();
-    if (statsFactory != null) {
-      CensusStatsModule censusStats =
-          new CensusStatsModule(statsFactory, GrpcUtil.STOPWATCH_SUPPLIER);
-      tracerFactories.add(censusStats.getServerTracerFactory());
+    if (GrpcUtil.enableCensusStats) {
+      StatsContextFactory statsFactory =
+          this.statsFactory != null ? this.statsFactory : Stats.getStatsContextFactory();
+      if (statsFactory != null) {
+        CensusStatsModule censusStats =
+            new CensusStatsModule(statsFactory, GrpcUtil.STOPWATCH_SUPPLIER);
+        tracerFactories.add(censusStats.getServerTracerFactory());
+      }
     }
-    CensusTracingModule censusTracing =
-        new CensusTracingModule(Tracing.getTracer(), Tracing.getBinaryPropagationHandler());
-    tracerFactories.add(censusTracing.getServerTracerFactory());
+    if (GrpcUtil.enableCensusTracing) {
+      CensusTracingModule censusTracing =
+          new CensusTracingModule(Tracing.getTracer(), Tracing.getBinaryPropagationHandler());
+      tracerFactories.add(censusTracing.getServerTracerFactory());
+    }
     tracerFactories.addAll(streamTracerFactories);
 
     io.grpc.internal.InternalServer transportServer =
--- a/core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
@@ -38,6 +38,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.instrumentation.stats.Stats;
 import com.google.instrumentation.stats.StatsContextFactory;
+import com.google.instrumentation.trace.Tracing;
 import io.grpc.BindableService;
 import io.grpc.CompressorRegistry;
 import io.grpc.Context;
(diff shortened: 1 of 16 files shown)


zhangkun83:
Please ignore the individual commits and review this PR as a whole.

adriancole:
Couple quick comments:

Does "grpc-tags-bin" imply propagated tags are shared between metrics and
tracing? Fine if so, but it is notable

Is it intentional that static references to grpc context are implied? Ex we
get the current parent using a static constant. This may be fine, but also
could use explanation (even if it is that everything gRPC uses static /
classloader configuration)

zhangkun83:
> Does "grpc-tags-bin" imply propagated tags are shared between metrics and tracing? Fine if so, but it is notable

No. Tracing has its own header: `grpc-trace-bin`.

> Is it intentional that static references to grpc context are implied? Ex we get the current parent using a static constant. This may be fine, but also could use explanation (even if it is that everything gRPC uses static / classloader configuration)

You mean `CONTEXT_SPAN_KEY`? We intend to standardize the keys for accessing the tags and span in the gRPC Context as constants from the instrumentation API. Do you have a reason to consider not doing that way?

adriancole:
No. Tracing has its own header: grpc-trace-bin.

Ok just "tags" seems generic so wondering if shared elsewhere. Thx

You mean CONTEXT_SPAN_KEY? We intend to standardize the keys for accessing
the tags and span in the gRPC Context as constants from the instrumentation
API. Do you have a reason to consider not doing that way

I get making static constants for the keys esp as iirc equals is by
identity.

I think i was more curious about use of implicit context ex
CONTEXT_SPAN_KEY.get() vs explicit CONTEXT_SPAN_KEY.get(context). Since in
some cases we are shuffling references to context anyway maybe we can
comment when we don't (ex note why when we choose to use implicit context
vs explicit)

zhangkun83:
> I think i was more curious about use of implicit context ex
CONTEXT_SPAN_KEY.get() vs explicit CONTEXT_SPAN_KEY.get(context). Since in
some cases we are shuffling references to context anyway maybe we can
comment when we don't (ex note why when we choose to use implicit context
vs explicit)

No problem. I have added comments.

core/src/main/java/io/grpc/internal/CensusTracingModule.java
```diff
@@ -0,0 +1,237 @@
...
+     *
+     * <p>Can be called from any thread without synchronization.  Calling it the second time or more
+     * is a no-op.
+     */
+    @Override
+    public void streamClosed(Status status) {
+      if (!streamClosed.compareAndSet(false, true)) {
+        return;
+      }
+      span.end();
```
bogdandrutu: You may want to pass the status via EndSpanOptions to the Span.

core/src/main/java/io/grpc/internal/CensusTracingModule.java
```diff
@@ -0,0 +1,237 @@
...
+  final class ClientCallTracer extends ClientStreamTracer.Factory {
+
+    private final String fullMethodName;
+    private final AtomicBoolean callEnded = new AtomicBoolean(false);
+    private final Span span;
+
+    ClientCallTracer(@Nullable Span parentSpan, String fullMethodName) {
+      this.span =
+          censusTracer.spanBuilder(parentSpan, makeSpanName("Sent", fullMethodName)).startSpan();
+      this.fullMethodName = checkNotNull(fullMethodName, "fullMethodName");
```
bogdandrutu: You may want to check for not null before using it in the makeSpanName

core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
```diff
@@ -175,11 +176,15 @@ public ServerImpl build() {
...
     if (statsFactory != null) {
-      CensusStreamTracerModule census =
-          new CensusStreamTracerModule(statsFactory, GrpcUtil.STOPWATCH_SUPPLIER);
-      tracerFactories.add(census.getServerTracerFactory());
+      CensusStatsModule censusStats =
+          new CensusStatsModule(statsFactory, GrpcUtil.STOPWATCH_SUPPLIER);
+      tracerFactories.add(censusStats.getServerTracerFactory());
     }
+    CensusTracingModule censusTracing =
+        new CensusTracingModule(Tracing.getTracer(), Tracing.getBinaryPropagationHandler());
```
ejona86: Hmm... so we'll send headers all the time? Do we want to flag-protect this? We can use a static field and just enable it when we want to test.

core/src/main/java/io/grpc/internal/CensusTracingModule.java
```diff
@@ -0,0 +1,237 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Provides factories for {@link StreamTracer} that records traces to Census.
+ *
+ * <p>On the client-side, a factory is created for each call, because ClientCall starts earlier than
+ * the ClientStream, and in some cases may even not create a ClientStream at all.  Therefore, it's
+ * the factory that reports the summary to Census.
+ *
+ * <p>On the server-side, a tracer is created for each call, because ServerStream starts earlier
```
ejona86: s/call/stream/ ?

core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
zhangkun83: Per offline discussion, in order to prevent premature wire-format changes from breaking existing users, we should keep both stats and tracing flag-protected until their wire format stabilized. I have added two static booleans to `GrpcUtil`.

zhangkun83:
@ejona86 These warnings are real
```
Download https://oss.sonatype.org/content/repositories/snapshots/com/google/instrumentation/instrumentation-api/0.4.1-SNAPSHOT/instrumentation-api-0.4.1-20170421.154215-2.pom
Download https://repo1.maven.org/maven2/com/google/auto/value/auto-value/1.4/auto-value-1.4.pom
Download https://oss.sonatype.org/content/repositories/snapshots/com/google/instrumentation/instrumentation-api/0.4.1-SNAPSHOT/instrumentation-api-0.4.1-20170421.154215-2.jar
Download https://repo1.maven.org/maven2/com/google/auto/value/auto-value/1.4/auto-value-1.4.jar
warning: No processor claimed any of these annotations: com.google.errorprone.annotations.ForOverride,io.grpc.ExperimentalApi,javax.annotation.concurrent.ThreadSafe,javax.annotation.concurrent.GuardedBy,javax.annotation.CheckReturnValue,io.grpc.Internal,com.google.common.annotations.VisibleForTesting,javax.annotation.concurrent.Immutable,javax.annotation.Nullable,com.google.errorprone.annotations.DoNotMock,javax.annotation.concurrent.NotThreadSafe
error: warnings found and -Werror specified
```
I tried only upgrading `instrumentation-api` to the latest snapshot in a warning-free branch, and the warnings started to show up. So this is definitely related to the changes in instrumentation. Could it be due to use of AutoValue?

zhangkun83:
Verified the issue was introduced between instrumentation 0.3.0 and 0.4.0

zhangkun83:
I have confirmed that google/instrumentation-java/pull/132 is the first commit on instrumentation `master` with which gRPC build start to show these warnings. AutoValue is super suspicious :(
```
2017-03-28 17:46 sebright           I─┐ [refs/bisect/bad] Merge pull request #132 from sebright/auto-value
2017-03-28 15:59 Bogdan Drutu       o │ [refs/bisect/good-4400097d1b3d243a8fb62cab838b3c980d5a673d] Add SpanData and TimedEvent classes. (#174)
2017-03-28 15:51 Yang Song          o │ [refs/bisect/good-eb45b7b8b8edf5f7e0cebe0d11c9fe261e3182b1] Add measurement and view descriptors for client/server started/finished~
2017-03-28 15:21 Kristen Kozak      │ o Add a note about AutoValue to CONTRIBUTING.md.
2017-03-16 16:33 Kristen Kozak      │ o Reduce visibility of MeasurementDescriptor and MeasurementUnit constructors.
2017-03-21 17:51 Kristen Kozak      │ o [HEAD] Add AutoValue to Gradle build.
2017-03-10 19:27 Kristen Kozak      │ o Tests MeasurementDescriptor and MeasurementUnit equals() and hashCode().
2017-03-10 18:49 Kristen Kozak      │ o Starts using AutoValue for classes in stats-core.
2017-03-28 10:57 Bogdan Drutu       I─┘ [refs/bisect/good-c6a59fbdd201894719979fec218977ec7bf8e8c0] Add build rule for examples to generate executable binaries. Fix javado~
```

zhangkun83:
It can be fixed by changing `instrumentation-api`'s dependency on `auto-value` from `compile` to `compileOnly` (google/instrumentation-java/pull/254)

adriancole:
Congratulations!!

PR #2941, opened 2017-04-24
Implement GRPCLB client-side load reporting

zhangkun83:
Also updated `load_balancing.proto` to the latest version.


zhangkun83:
@slash-lib, please focus on `GrpclbClientLoadRecorder` and `GrpclbLoadBalancerTest.loadReporting()` for their compliance with the spec.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -321,22 +374,80 @@ public void run() {
...
+      stopLoadReporting();
+      if (loadReportIntervalMillis > 0) {
+        loadReportTask = timerService.schedule(
+            loadReportRunnable, loadReportIntervalMillis, TimeUnit.MILLISECONDS);
+      }
+    }
+
+    private void stopLoadReporting() {
+      if (loadReportTask != null) {
+        loadReportTask.cancel(false);
```
dapengzhang0: There is a benign race here: if the timerService just started executing the task of `loadReportTask`,  namely line 346, but line 347 is not executed yet, it can not be canceled by `loadReportTask.cancel(false)`, then `sendLoadReport()` will be queued by `runSerialized`. In this case `cleanUp()` does not work.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -321,22 +374,80 @@ public void run() {
...
+      try {
+        lbRequestWriter.onNext(LoadBalanceRequest.newBuilder().setClientStats(stats).build());
+        scheduleNextLoadReport();
+      } catch (Exception e) {
+        close(e);
+      }
+    }
+
+    private void scheduleNextLoadReport() {
+      stopLoadReporting();
```
dapengzhang0: It does not make sense to call `stopLoadReporting()` here, because `scheduleNextLoadReport()` is invoked either for the first time or by the task runnable (line 349). 
`loadReportTask` is already `null` here in both cases.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
zhangkun83: This is actually not benign. The task may run and schedule yet another task anyway. I added a check for `closed` at the beginning of the task, also added a test case for this race.

grpclb/src/main/java/io/grpc/grpclb/TimeProvider.java
```diff
@@ -0,0 +1,39 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.grpc.grpclb;
+
+/**
+ * Allow time manipulation in tests.
```
dapengzhang0: interface is public by default, it is not package-private. Moving it to an inner class of package-private class like `GrpclbLoadBalancer` can make it package-private.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -78,6 +83,20 @@
   private static final Logger logger = Logger.getLogger(GrpclbLoadBalancer.class.getName());
 
   @VisibleForTesting
+  static final Map<DropType, Status> DROP_STATUSES;
+
+  static {
+    HashMap<DropType, Status> map = new HashMap<DropType, Status>();
```
dapengzhang0: * you can use `EnumMap` at the beginning.
`EnumMap<DropType, PickResult> map = new EnumMap<DropType, PickResult>(DropType.class);`
* you can use `PickResult` instead of `Status` (also rename `DROP_STATUSES` to `DROP_RESULTS`), then you don't have to create a new `PickResult` instance later each time. This can produce less garbage.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -321,22 +377,83 @@ public void run() {
...
+              "[{0}] : Skipping abundant initial response: {1}",
+              new Object[] {logId, initialResponse});
+        } else {
+          initialResponseReceived = true;
+          loadReportIntervalMillis =
+              Durations.toMillis(initialResponse.getClientStatsReportInterval());
+          scheduleNextLoadReport();
+          return;
+        }
+      }
```
dapengzhang0: should the `return;` after the bracket?

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancerFactory.java
```diff
@@ -46,6 +48,12 @@
 public class GrpclbLoadBalancerFactory extends LoadBalancer.Factory {
 
   private static final GrpclbLoadBalancerFactory instance = new GrpclbLoadBalancerFactory();
+  private static final TimeProvider realTimeProvider = new TimeProvider() {
```
zhangkun83: Done.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java

PR #2943, opened 2017-04-25
grpclb: remove non-determinism workaround

ejona86:
Issue #2444 is resolved.


ejona86:
@zhangkun83, friendly ping.

PR #2946, opened 2017-04-25
core: synchronize access to Census flags

carl-mastrangelo:


core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -61,6 +61,7 @@
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
```
zhangkun83: Not needed

carl-mastrangelo: done

Issue #2935, opened 2017-04-21
cause=io.netty.handler.codec.http2.Http2Exception: Header size exceeded max allowed size (10240)


cr22rc:
Please answer these questions before submitting your issue.

1.2.0

java.runtime.name = OpenJDK Runtime Environment
12:22:18     java.runtime.version = 1.8.0_121-8u121-b13-0ubuntu1.16.04.2-b13
12:22:18     java.specification.name = Java Platform API Specification
12:22:18     java.specification.vendor = Oracle Corporation
12:22:18     java.specification.version = 1.8
12:22:18     java.vendor = Oracle Corporation
12:22:18     java.vendor.url = http://java.oracle.com/
12:22:18     java.vendor.url.bug = http://bugreport.sun.com/bugreport/
12:22:18     java.version = 1.8.0_121
12:22:18     java.vm.info = mixed mode
12:22:18     java.vm.name = OpenJDK 64-Bit Server VM
12:22:18     java.vm.specification.name = Java Virtual Machine Specification
12:22:18     java.vm.specification.vendor = Oracle Corporation
12:22:18     java.vm.specification.version = 1.8
12:22:18     java.vm.vendor = Oracle Corporation
12:22:18     java.vm.version = 25.121-b13
12:22:18     line.separator = \

grpc async call  There really isn't any special options set on the channelbuilder
except usePlaintext it does have that set to true but a the moment not sure why.

The scenario this is running on works for all systems I'v but on these vms. 
Ubuntu vm 16.04. 

Is there some way to increase that size.
I can't imagine that the packet size would be any different between these system .. it's running the same test.  Is there any other reason this exception might happen?

cr22rc:
00:04:13 2017-04-21 00:04:13 DEBUG Peer:144 - sendProposalAsync name: peer1, url: grpc://localhost:7056

00:04:13 2017-04-21 00:04:13 DEBUG NettyClientHandler:147 - 
00:04:13 ----------------OUTBOUND--------------------
00:04:13 [id: 0xa601ff3c, L:/127.0.0.1:57166 - R:localhost/127.0.0.1:7056] SETTINGS: ack=false, settings={ENABLE_PUSH=0, MAX_CONCURRENT_STREAMS=0, INITIAL_WINDOW_SIZE=1048576, MAX_HEADER_LIST_SIZE=8192}
00:04:13 ------------------------------------
00:04:13 2017-04-21 00:04:13 DEBUG NettyClientHandler:147 - 
[...]

PR #2948, opened 2017-04-26
netty: Clarify max header list size configuration

ejona86:
CC @cr22rc

Addresses documentation issue from #2935


ejona86:

PR #2954, opened 2017-04-27
stub: add debug info for sending msg after stream closed

### LINKED ISSUE #2953: Request streamObserver can still call onNext a million times after stream is closed without being aware of that
For client streaming call, once the RPC is established, even server is shutdown, the client can still (mistakenly) keep calling `requestStreamObserver.onNext()` for a million times, and gRPC lib just silently ignores them.

dapengzhang0:
Resolves #2953
Add debug info to notice application developers when they were making the mistake described in #2953.


ejona86:
No, I don't think we should have a log message there. This happens in normal usage and is "okay." Yes, it is unfortunate that there is no limit to how long they can continue trying to send more messages, but I don't see a good way to fix that.

dapengzhang0:
Understood. Can we impose a limit, say 10 minutes, on how long user can not send more messages if the stream has been closed for?

PR #2955, opened 2017-04-27
stub: rm redundant call field in StreamObserverToCallListenerAdapter

dapengzhang0:
All that the `call` field was used for was to call `request` and nothing else. The `request` can be done by the `adapter` field, so the `call` field is redundant.


dapengzhang0:
Jenkins, test this please

dapengzhang0:
Jenkins, test this please

PR #2956, opened 2017-04-27
Channel.authority() should return the value of overrideAuthority

### LINKED ISSUE #2682: overrideAuthority does not impact authority for JWT in jwt_token_creds
https://github.com/grpc/grpc-java/commit/65e4d9f47abdad2a0a7a67fc5f1002bcf6f6fab1#diff-850c920bd69ac031d7e98cab9459dec8 caused grpc/grpc#9497

zpencer:
AbstractManagedChannelImplBuilder accepts an overrideAuthority parameter, but this value is not hooked up to the name resolver object. Ultimately, Channel.authority consults with the NameResolver, so the overrideAuthority should be hooked into the NameResolverFactory, while all other functionality should be preserved.

Also, add unit tests for all the variants of OkHttpChannelBuilder and NettyChannelBuilder constructors, namely to test the slightly different NettyChannelBuilder(SocketAddress) code path.

Fixes #2682


ejona86:
FYI: most of the time we'll rebase instead of merging, and then squash before pushing to the master branch.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -441,6 +445,59 @@ public String getDefaultScheme() {
   }
 
   /**
+   * A wrapper class that overrides the authority of a NameResolver, while preserving all other
+   * functionality.
+   */
+  @VisibleForTesting
+  protected static class OverrideAuthorityNameResolverFactory extends NameResolver.Factory {
```
ejona86: package-private instead of protected?

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -308,6 +308,10 @@ public ManagedChannel build() {
       // getResource(), then this shouldn't be a problem unless called on the UI thread.
       nameResolverFactory = NameResolverProvider.asFactory();
     }
```
ejona86: Delete AuthorityOverridingTransportFactory?

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -441,6 +445,59 @@ public String getDefaultScheme() {
...
+   * functionality.
+   */
+  @VisibleForTesting
+  protected static class OverrideAuthorityNameResolverFactory extends NameResolver.Factory {
+    final NameResolver.Factory delegate;
+    final String authorityOverride;
+
+    /**
+     * Constructor for the {@link NameResolver.Factory}
+     * @param delegate The actual underlying factory that will produce the a {@link NameResolver}
```
ejona86: Leave an empty line before the section with `@param` and similar.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -441,6 +445,59 @@ public String getDefaultScheme() {
   }
 
   /**
+   * A wrapper class that overrides the authority of a NameResolver, while preserving all other
+   * functionality.
+   */
+  @VisibleForTesting
+  protected static class OverrideAuthorityNameResolverFactory extends NameResolver.Factory {
+    final NameResolver.Factory delegate;
```
ejona86: Private? Ditto for `authorityOverride`.

core/src/test/java/io/grpc/internal/AbstractManagedChannelImplBuilderTest.java
```diff
@@ -32,11 +32,19 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import io.grpc.Attributes;
+import io.grpc.NameResolver;
 
 import java.net.InetSocketAddress;
 import java.net.URI;
 import java.util.concurrent.TimeUnit;
+
```
ejona86: Creating extra sections here is the old style. Now there are only two sections: static imports and normal imports.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
zpencer: Didn't realize that this was all that AuthorityOverridingTransportFactory was trying to accomplish. Done.

Issue #2865, opened 2017-03-30
Rare race condition in Client


carl-mastrangelo:
While more prominent when using compression, this race occurs without it as well.  The typical race looks something like:

1.  Client starts and RPC
2.  The transport to the server is not yet available, so a DelayedClientTransport is used.
3.  The server handles the RPC and sends back headers and a compressed message.
4.  The client sees there are headers, and begins executing the queued stream callbacks, **on the channel executor threads** instead of the transport thread
5.  The client sees the Data frame, and tries to decompress it on the network thread.  *This fails since the headers from 4 have not yet been processed.
6.  The stream has already failed, but the queued callback for onHeaders() is finally executed on the app thread.  

This is the root cause of #2157.  As mentioned, this isn't just for compression.  ClientInterceptors will see headers *after* data has been received.   The solution (temporary?) seems to be to move OkHttp to used AbstractClientStream2 in #2821, and then move decompression from ClientCallImpl to the stream.  That will fix the decompression, but not interceptors.

ejona86:
`ClientInterceptors will see headers after data has been received.` That is not true; the Call listeners fire in the correct order. Even the stream listeners fire in the correct order. The problem is that the StreamListener callbacks aren't done on the transport thread, so the deframer's processing of onDataReceived is out-of-band compared to the ClientCallImpl's processing of the headers.

ericgribkoff:
Self-assigning due to ongoing work on moving client-side deframing to client thread.

ejona86:
#2957 fixed the race on server-side. Client-side is still needing to be fixed.

PR #2957, opened 2017-04-28
core: set server stream decompressor in transport thread

ericgribkoff:
Currently, inbound headers are processed in the transport thread, but the resulting new stream's decompressor is not set until the `ServerCallImpl` constructor is invoked from a `wrappedExecutor` thread. The `ServerCallImpl` constructor will parse the headers and, if necessary, call `MessageDeframer#setDecompressor` to set the decompressor. This is not thread-safe, as incoming data frames may simultaneously be consumed by the deframer in the transport thread.

This PR moves the call to `setDecompressor` into the transport thread. This is a similar issue to https://github.com/grpc/grpc-java/issues/2865 on the client side. My in-progress move of client deframing to the client thread made the compression test (https://github.com/grpc/grpc-java/issues/2157) extra flaky, but due to this race condition on the server side.

If https://github.com/grpc/grpc-java/issues/2865 is the only cause for https://github.com/grpc/grpc-java/issues/2157, then moving client deframing to the app thread should also clear up the flaky compression test.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ServerCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerCallImpl.java
@@ -44,7 +44,6 @@ import io.grpc.Codec;
 import io.grpc.Compressor;
 import io.grpc.CompressorRegistry;
 import io.grpc.Context;
-import io.grpc.Decompressor;
 import io.grpc.DecompressorRegistry;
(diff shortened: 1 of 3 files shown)


carl-mastrangelo:
LGTM.

ericgribkoff:
Jenkins: retest this please

PR #2963, opened 2017-05-01
Server side cancellations should promptly inform server

### LINKED ISSUE #2112: Server-side cancellation does not promptly inform server
Both `context.isCancelled()` and `call.isCancelled()` only begin returning `true` _after_ `ServerCallListener.onCancel()` (in `ServerCallImpl.ServerStreamListenerImpl`, but that is already on application callback thread via `ServerImpl.JumpToApplicationThreadServerStreamListener`). That means callers must return from a listener event before the values can change _and_ they will have already seen `onCancel()`, which makes them broken for the very case they are there to solve.

If `!status.isOk()`, both the context and the call should return `true` as soon as the transport notifies the server. How to do that seems it will be a bit awkward with the current code organization.

zpencer:
When we cancel a stream, we must 1) set both ServerCall.isCancelled() and CancellableContext.isCancelled() to be true, as well 2) notify any listeners that the stream is now cancelled. Today, this happens in JumpToApplicationThreadServerStreamListener therefore both 1 and 2 happen in a serialized executor and may become queued. This diff moves 1 to be executed immediately on the cancelling thread's context (we are still required to run 2 in the executor for correctness reasons). The result is that anyone polling for isCancelled() in either of the two places will immediately see the update rather than waiting.

Fixes #2112

### DIFF
--- a/core/src/main/java/io/grpc/internal/ServerCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerCallImpl.java
@@ -39,6 +39,7 @@ import static io.grpc.internal.GrpcUtil.MESSAGE_ACCEPT_ENCODING_KEY;
 import static io.grpc.internal.GrpcUtil.MESSAGE_ENCODING_KEY;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.util.concurrent.MoreExecutors;
 import io.grpc.Attributes;
 import io.grpc.Codec;
 import io.grpc.Compressor;
@@ -221,6 +222,17 @@ final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
       this.call = checkNotNull(call, "call");
       this.listener = checkNotNull(listener, "listener must not be null");
       this.context = checkNotNull(context, "context");
+      // Wire ourselves up so that if the context is cancelled, our flag call.cancelled also
+      // reflects the new state. Use a DirectExecutor so that it happens in the same thread
+      // as the caller of {@link Context#cancel}.
+      this.context.addListener(
+          new Context.CancellationListener() {
+            @Override
+            public void cancelled(Context context) {
+              ServerStreamListenerImpl.this.call.cancelled = true;
+            }
+          },
+          MoreExecutors.directExecutor());
     }
 
     @SuppressWarnings("Finally") // The code avoids suppressing the exception thrown from try
--- a/core/src/main/java/io/grpc/internal/ServerCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerCallImpl.java
@@ -232,7 +232,8 @@ final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
               ServerStreamListenerImpl.this.call.cancelled = true;
             }
           },
-          MoreExecutors.directExecutor());
+          MoreExecutors.directExecutor()
+      );
     }
 
     @SuppressWarnings("Finally") // The code avoids suppressing the exception thrown from try
--- a/core/src/main/java/io/grpc/internal/ServerCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerCallImpl.java
@@ -232,8 +232,7 @@ final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
               ServerStreamListenerImpl.this.call.cancelled = true;
             }
           },
-          MoreExecutors.directExecutor()
-      );
+          MoreExecutors.directExecutor());
     }
 
     @SuppressWarnings("Finally") // The code avoids suppressing the exception thrown from try
--- a/core/src/main/java/io/grpc/internal/ServerImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerImpl.java
@@ -594,7 +594,10 @@ public final class ServerImpl extends io.grpc.Server implements WithLogId {
       callExecutor.execute(new ContextRunnable(context) {
         @Override
         public void runInContext() {
-            getListener().closed(status);
+          if (status.isOk()) {
+            context.cancel(status.getCause());
+          }
+          getListener().closed(status);
         }
       });
     }
--- a/core/src/main/java/io/grpc/internal/ServerImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerImpl.java
@@ -584,7 +584,8 @@ public final class ServerImpl extends io.grpc.Server implements WithLogId {
 
     @Override
     public void closed(final Status status) {
-      // Immediately inform any users of the context that their work should be aborted.
+      // For cancellations, promptly inform any users of the context that their work should be
+      // aborted. Otherwise, we can wait until pending work is done.
       if (!status.isOk()) {
         context.cancel(status.getCause());
       }
--- a/core/src/main/java/io/grpc/internal/ServerImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerImpl.java
@@ -39,7 +39,6 @@ import static io.grpc.internal.GrpcUtil.TIMEOUT_KEY;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 import io.grpc.Attributes;
 import io.grpc.CompressorRegistry;
@@ -83,8 +82,6 @@ public final class ServerImpl extends io.grpc.Server implements WithLogId {
 
   private final LogId logId = LogId.allocate(getClass().getName());
   private final ObjectPool<? extends Executor> executorPool;
-  /** Executor for running cancellation related tasks. Non-null only in unit tests. */
-  private final Executor cancelExecutor;
   /** Executor for application processing. Safe to read after {@link #start()}. */
   private Executor executor;
   private final InternalHandlerRegistry registry;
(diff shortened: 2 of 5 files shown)


core/src/main/java/io/grpc/internal/ServerCallImpl.java
```diff
@@ -221,6 +222,17 @@ public ServerStreamListenerImpl(
       this.call = checkNotNull(call, "call");
       this.listener = checkNotNull(listener, "listener must not be null");
       this.context = checkNotNull(context, "context");
+      // Wire ourselves up so that if the context is cancelled, our flag call.cancelled also
+      // reflects the new state. Use a DirectExecutor so that it happens in the same thread
+      // as the caller of {@link Context#cancel}.
+      this.context.addListener(
```
zpencer: I see that ClientCallImpl is slightly different, and instead implements cancelled(Context) by calling stream.cancel on the ClientStream. A few unit tests break if I do the analogous thing to the ServerStream. Is it worthwhile to follow the ClientCallImpl example and fix those unit tests? It didn't seem obvious to me, because it feels strange to receive a cancellation signal from the stream, surface it up to the level of a ServerCall, and then have the ServerCallImpl try to cancel the stream again. In the client case on the other hand, it seems sensible to initiate the cancellation on a ClientCall directly, and expect it to propagate it down to the ClientStream.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -584,16 +584,21 @@ public void runInContext() {
 
     @Override
     public void closed(final Status status) {
+      // For cancellations, promptly inform any users of the context that their work should be
+      // aborted. Otherwise, we can wait until pending work is done.
+      if (!status.isOk()) {
+        context.cancel(status.getCause());
```
ejona86: Since the context listeners may be using directExecutor, I'd like to avoid calling them from the network thread. Can you schedule this on `executor`? (not callExecutor)

@zhangkun83, since we've talked so much about which executor to use when, how does this sound to you? Using `callExecutor` is known-problematic, and not using any executor makes me a bit worried.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -584,16 +584,21 @@ public void runInContext() {
...
+      }
+      // The contract of this.listener is that it's always accessed from inside callExecutor.
```
ejona86: Isn't this comment true for the entire class? It seems it's the entire purpose for the class. It seems like either remove it or move it to getListener()/listener or some similar class-wide documentation.

core/src/main/java/io/grpc/internal/ServerCallImpl.java
```diff
@@ -221,6 +222,17 @@ public ServerStreamListenerImpl(
...
+          new Context.CancellationListener() {
+            @Override
+            public void cancelled(Context context) {
+              ServerStreamListenerImpl.this.call.cancelled = true;
```
ejona86: If this becomes a problem, we could move it to ServerStream by having ServerCall.isCancelled call ServerStream.isCancelled (a new method).

core/src/main/java/io/grpc/internal/ServerImpl.java
zpencer: I'm repeating the comment found higher up in this class:
```
// Only accessed from callExecutor.
private ServerStreamListener listener;
```

I hit a unit test case where I didn't serialize on the executor and following happened:
- ServerImpl#streamCreated runs
  - It creates a JumpToApplicationThreadServerStreamListener whose listener is initially null
  - It enqueues a ContextRunnable on callExecutor responsible for wiring up the method from the registry, but it has not yet run
- Now the unit test attempts to cancel the stream. Because the operation is not serialized on callExecutor, we find that JumpToApplicationThreadServerStreamListener.listener is still null, and raise an error.

core/src/main/java/io/grpc/internal/ServerImpl.java
zhangkun83: Why can't we use `callExecutor`?

zpencer: The callExecutor is the one doing the user supplied unit of work, the same one that is should be checking for `Context.current().isCancelled()` or `ServerCall.isCancelled()`. We want to be able to set the flag and have its effect seen by the callExecutor's runnable, rather than enqueuing it into callExecutor.

zhangkun83: Isn't the user supplied work `getListener().closed(status)` that is enqueued **after** what we can enqueue here?

zhangkun83: Yeah, I misunderstood the scenario. `executor` should be OK.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -584,16 +605,27 @@ public void runInContext() {
 
     @Override
     public void closed(final Status status) {
+      // For cancellations, promptly inform any users of the context that their work should be
+      // aborted. Otherwise, we can wait until pending work is done.
+      if (!status.isOk()) {
+        // The callExecutor might be busy doing user work. To avoid waiting, use the alternate
+        // cancelExecutor.
```
zhangkun83: To be more accurate, you may want to say "The callExecutor is serializing executor. In order to notify tasks that are already in the callExecutor, we use an executor that is not serializing".

core/src/test/java/io/grpc/internal/ServerImplTest.java
```diff
@@ -822,9 +825,55 @@ public void cancelled(Context context) {
...
+    ServerStreamListener streamListener = testClientClose_setup(callReference,
+        context, contextCancelled);
+
+    // For close status being non OK:
+    // isCancelled is expected to be true immediately after calling closed(), without needing
+    // to wait for the main executor to run any tasks.
+    assertFalse(callReference.get().isCancelled());
+    assertFalse(context.get().isCancelled());
     streamListener.closed(Status.CANCELLED);
+    assertEquals(1, cancelExecutor.runDueTasks());
```
zhangkun83: IIUC, this is the reason you added `cancelExecutor`, because you only want the cancellation runnable to run, and leave out the stream listeners. Instead of having for-test mechanism in the implementation, you may consider this alternative approach:

Give the cancellation runnable a class, and add `runDueTasks(TaskFilter)` to `FakeClock` that runs selected runnables:
```java
interface TaskFilter {
  boolean shouldRun(Runnable task);
}
```

core/src/main/java/io/grpc/internal/ServerImpl.java
zpencer: done

core/src/test/java/io/grpc/internal/ServerImplTest.java
```diff
@@ -766,20 +767,24 @@ private void checkContext() {
...
     verify(stream, times(0)).close(isA(Status.class), isNotNull(Metadata.class));
   }
 
-  @Test
-  public void testClientCancelTriggersContextCancellation() throws Exception {
+  private ServerStreamListener testClientClose_setup(
+      final AtomicReference<ServerCall<String, Integer>> callReference,
+      final AtomicReference<Context> context,
+      final AtomicBoolean contextCancelled
+  ) throws Exception {
```
zpencer: Changed this and other similar places

core/src/test/java/io/grpc/internal/ServerImplTest.java
zpencer: That's a much cleaner idea. I've made the change in FakeClock and added a test.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -123,6 +123,17 @@
...
+        rootContext, decompressorRegistry, compressorRegistry, transportFilters, null);
+  }
+
+  @VisibleForTesting
+  ServerImpl(ObjectPool<? extends Executor> executorPool,
+      ObjectPool<ScheduledExecutorService> timeoutServicePool,
+      InternalHandlerRegistry registry, HandlerRegistry fallbackRegistry,
+      InternalServer transportServer, Context rootContext,
+      DecompressorRegistry decompressorRegistry, CompressorRegistry compressorRegistry,
+      List<ServerTransportFilter> transportFilters, Executor cancelExecutor) {
```
zhangkun83: `cancelExecutor` is no longer needed.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -512,18 +524,23 @@ public void onReady() {}
   @VisibleForTesting
   static class JumpToApplicationThreadServerStreamListener implements ServerStreamListener {
     private final Executor callExecutor;
+    private final Executor cancelExecutor;
```
zhangkun83: This is no longer needed since in the test it's always the same as `callExecutor`.

core/src/main/java/io/grpc/internal/ServerImpl.java
zpencer: Ah missed this, thanks

core/src/main/java/io/grpc/internal/ServerImpl.java
zpencer: Synced offline about this being a static class

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -616,4 +626,19 @@ public void runInContext() {
       });
     }
   }
+
+  static class ContextCloser implements Runnable {
```
zhangkun83: `@VisibleForTesting`

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -584,16 +590,20 @@ public void runInContext() {
...
       callExecutor.execute(new ContextRunnable(context) {
         @Override
         public void runInContext() {
-          try {
-            getListener().closed(status);
-          } finally {
-            // Regardless of the status code we cancel the context so that listeners
-            // are aware that the call is done.
+          if (status.isOk()) {
             context.cancel(status.getCause());
```
ejona86: We want to call the listener before cancelling the context. Basically, put this back the way it was, but with the new condition.

Issue #2961, opened 2017-04-29
Maven dependency conversion issues


dconnelly:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?

1.3.0

### What JVM are you using (`java -version`)?

1.8.0_121

### What did you do?

    mvn compile

If possible, provide a recipe for reproducing the error.

Have the following dependencies in Maven:

    <dependency>
      <groupId>io.grpc</groupId>
      <artifactId>grpc-netty</artifactId>
      <version>1.3.0</version> 
    </dependency>
    <dependency>
      <groupId>io.grpc</groupId>
      <artifactId>grpc-protobuf</artifactId>
      <version>1.3.0</version>
    </dependency>
    <dependency>
      <groupId>io.grpc</groupId>
      <artifactId>grpc-stub</artifactId>
      <version>1.3.0</version>
    </dependency>

Looks like artifact ```com.google.api.grpc:grpc-google-common-protos``` has a dependency
on ```grpc-all:1.0.1``` which causes a convergence issue if you have the above dependencies and you use the dependency enforcer plugin in Maven.

The workaround is to add ```grpc-all``` as an exclusion for that dependency which is always a bit scary.

### What did you expect to see?

No dependency conversion issues.

### What did you see instead?

Dependency conversion failure in Maven.

pambrose:
I am seeing the same issue.

ejona86:
Your example did not include your Maven enforcer configuration. As an aside, Maven enforcer's `requireSameVersions` rule is silly. `requireUpperBoundDeps` is quite sensible though, because it verifies that Maven is doing what it should have done to begin with.

Eww... grpc-google-common-protos:0.1.6 depends on grpc-all. We should upgrade to the latest (0.1.9) which depends on grpc-stub instead. But even better, the grpc and proto artifacts have been split so we can use proto-google-common-protos instead, which has no grpc dependency.

ejona86:
Hmm... actually, we do have an exclusion in place:

```xml
    <dependency>
      <groupId>com.google.api.grpc</groupId>
      <artifactId>grpc-google-common-protos</artifactId>
      <version>0.1.6</version>
      <scope>compile</scope>
      <exclusions>
        <exclusion>
          <artifactId>*</artifactId>
          <groupId>io.grpc</groupId>
        </exclusion>
        <exclusion>
          <artifactId>*</artifactId>
          <groupId>com.google.api</groupId>
        </exclusion>
      </exclusions>
[...]

PR #2965, opened 2017-05-01
build,protobuf: update google_api_protos dependency

ericgribkoff:
Addresses https://github.com/grpc/grpc-java/issues/2961.

Thanks to @zhangkun83 who identified a similar issue with instrumentation-java, I've sent https://github.com/googleapis/api-common-java/pull/27 to make their dependency on auto-value compileOnly, which will enable remove the remaining `exclude group` from the dependency on `libraries.google_api_protos`. AFAIK the presence of this rule doesn't impact https://github.com/grpc/grpc-java/issues/2961.

### DIFF
--- a/protobuf/build.gradle
+++ b/protobuf/build.gradle
@@ -16,7 +16,6 @@ dependencies {
             libraries.protobuf_util
 
     compile (libraries.google_api_protos) {
-        exclude group: 'io.grpc'
         // 'com.google.api' transitively depends on 'com.google.auto.value:auto-value:1.1', which
         // breaks our annotations.
         exclude group: 'com.google.api'
--- a/protobuf/build.gradle
+++ b/protobuf/build.gradle
@@ -16,9 +16,9 @@ dependencies {
             libraries.protobuf_util
 
     compile (libraries.google_api_protos) {
-        // 'com.google.api' transitively depends on 'com.google.auto.value:auto-value:1.1', which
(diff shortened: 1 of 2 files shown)


ejona86:
:( Why does it depend on api-common?

protobuf/build.gradle
```diff
@@ -16,7 +16,6 @@ dependencies {
             libraries.protobuf_util
 
     compile (libraries.google_api_protos) {
-        exclude group: 'io.grpc'
         // 'com.google.api' transitively depends on 'com.google.auto.value:auto-value:1.1', which
         // breaks our annotations.
         exclude group: 'com.google.api'
```
ejona86: Also specify the artifact name, so that it avoids the wildcard in the pom.

ericgribkoff: Done

ericgribkoff:
I'm not sure why they depend on api-common. It's not clear to me from https://github.com/googleapis/googleapis how the jar is even produced or what it contains, without looking through the jar from maven itself. Should I open an issue on https://github.com/googleapis/googleapis about this dependency?

ejona86:
Yeah, please open an issue about it, and CC me on it.

ericgribkoff:
Opened https://github.com/googleapis/googleapis/issues/333

ejona86:
Backported to 1.3.x in #3155

PR #2967, opened 2017-05-02
android-helloworld: Ignore proguard warnings for javax.naming

ejona86:
I have no clue why this isn't needed for routeguide as well, but
routeguide is compiling fine without the dontwarn.


ejona86:
No need for backport. TXT support is not in 1.3.x, and this fix was merged to master before 1.4.x was cut.

PR #2974, opened 2017-05-03
core: record individual messages to Census tracing

zhangkun83:
- Outbound messages are recorded when they are fully serialized.
- Inbound messages are recorded when they are fully read from wire.
- Wire (compressed) sizes are recorded.
- Uses per-stream per-direction incremental serial number as the message ID.

### DIFF
--- a/core/src/main/java/io/grpc/StreamTracer.java
+++ b/core/src/main/java/io/grpc/StreamTracer.java
@@ -62,32 +62,45 @@ public abstract class StreamTracer {
   }
 
   /**
+   * An outbound message has been completely serialized, revealing its wire size.
+   */
+  public void outboundMessageSerialized(long wireSize) {
+  }
+
+  /**
+   * An inbound message has been completely read from the wire, revealing its wire size.
+   */
+  public void inboundMessageReceived(long wireSize) {
+  }
+
+  /**
    * The wire size of some outbound data is revealed. This can only used to record the accumulative
-   * outbound wire size. There is no guarantee wrt timing or granularity of this method.
+   * outbound wire size.  gRPC makes the best effort to call it as soon the information is
+   * available, although there is not guarantee wrt timing or granularity.
    */
   public void outboundWireSize(long bytes) {
   }
 
   /**
    * The uncompressed size of some outbound data is revealed. This can only used to record the
-   * accumulative outbound uncompressed size. There is no guarantee wrt timing or granularity of
-   * this method.
+   * accumulative outbound uncompressed size. gRPC makes the best effort to call it as soon as the
+   * information is available, although there is not guarantee wrt timing or granularity.
    */
   public void outboundUncompressedSize(long bytes) {
   }
 
   /**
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/grpc/StreamTracer.java
```diff
@@ -62,32 +62,45 @@ public void inboundMessage() {
   }
 
   /**
+   * An outbound message has been completely serialized, revealing its wire size.
+   */
+  public void outboundMessageSerialized(long wireSize) {
```
ejona86: We can't support this accurately with stream-wide compression, because we won't want to flush the compressor after every message. So we'll just know "the last three messages are X bytes on the wire."

core/src/main/java/io/grpc/StreamTracer.java
```diff
@@ -62,32 +62,45 @@ public void inboundMessage() {
...
+  }
+
+  /**
+   * An inbound message has been completely read from the wire, revealing its wire size.
+   */
+  public void inboundMessageReceived(long wireSize) {
```
ejona86: This is even more problematic with stream-wide compression, since we can't easily correlate compressed bytes to decompressed bytes (except if we decompressed one byte at a time, which is not acceptable).

core/src/main/java/io/grpc/StreamTracer.java
zhangkun83: @bogdandrutu for now we could provide accurate wire size and timing at the same time. But soon with stream-wide compression we will be left with two options:
 1. Provide accurate timing only. NetworkEvent could be added for each message, but the actual wire size is unknown. We could know the uncompressed size for outbound, but will have difficulty (but not impossible) in knowing the uncompressed size for inbound.
 2. Provide accurate wire size and timing for each HTTP2 frame.

zhangkun83: We may be able to implement Option 1 with uncompressed message size, if we can figure out a way to know when the message is fully read out in [SizeEnforcingInputStream](https://github.com/grpc/grpc-java/blob/v1.3.0/core/src/main/java/io/grpc/internal/MessageDeframer.java#L396). @bogdandrutu would that work for you?

bogdandrutu: I think what option we choose we should be consistent on both cases. Which one do you think it is easier.

I will also have to consider maybe changing the NetworkEvent to support 2 sizes (compressed and uncompressed) to make it clear to the users which one it is.

I am not sure I understand all the options, but I would say consistency is important (between inbound and outbound and between languages), so I would like to know what is the story in other languages (Go, C)

zhangkun83: I think other languages will have the same problem getting the message wire size with stream-based compression, because message boundaries have diminish in the compressed stream. Uncompressed size will be more feasible.

zhangkun83:
Closing in favor of #3461

PR #2987, opened 2017-05-09
netty: fix server keepalive not initialized bug

### LINKED ISSUE #2982: Fail to to enable server-side keepAlive on 1.3.0 RELEASE
GRPC 1.3.0

1.8.0_112

### What did you do?
Try to enable server-side keepAlive(added in gprc-java 1.3) but the feature didn't work properly.
After setting server-side keepAlive to 180s and keepAliveTimeout to 30s. what i expect is the server will sending a keepAlive to client if the connection is idle(didn't sent any data during 180s). however. i did't see such ping packet on network. 

Debugging the code(io.grpc.internal.KeepAliveManager), it seems the `onTransportActive`/`onTransportIdle` methods didn't be called by NettyServerHandler.  Unlike the NettyClientHandler, the keepAliveManager can't be triggered by stream's created/remove.

Only `onDataReceived` is called if data received by NettyServerHandler. 
image

So, simply if the client didn't sent any data, the server wouldn't sent back any ping.

In this case, another way to enable server-side KeepAlive is set `keepAliveDuringTransportIdle(permitKeepAliveWithoutCalls)` to true, so when the `io.grpc.internal.KeepAliveManager#onTransportStarted` is called, it will schedule a ping future immediately. 

However, the call of `io.grpc.internal.KeepAliveManager#onTransportStarted` didn't make effective too!
the only caller on server is `io.grpc.netty.NettyServerHandler#handleProtocolNegotiationCompleted`, which didn't invoke `onTransportStarted` because the `keepAliveManager` has not been initialized yet(the keepAliveManager always null during this method called).

image

dapengzhang0:
fixes #2982 

The problem is `grpcHandler.handleProtocolNegotiationCompleted` can happen before `ctx.pipeline().replace(this, null, grpcHandler)` in which `handlerAdded()` is called

Moved `keepAliveManager.onTransportStarted()` inside `handlerAdded()` to fix it.

Tested with manual interoptest and it's working now.

Removed the test case `keepAliveManagerStarted() ` that is not functioning. The test case `keepAliveManager_pingSent()` already covered it.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
@@ -334,6 +334,7 @@ class NettyServerHandler extends AbstractNettyHandler {
     if (keepAliveTimeInNanos != SERVER_KEEPALIVE_TIME_NANOS_DISABLED) {
       keepAliveManager = new KeepAliveManager(new KeepAlivePinger(ctx), ctx.executor(),
           keepAliveTimeInNanos, keepAliveTimeoutInNanos, true /* keepAliveDuringTransportIdle */);
+      keepAliveManager.onTransportStarted();
     }
     super.handlerAdded(ctx);
   }
@@ -441,9 +442,6 @@ class NettyServerHandler extends AbstractNettyHandler {
   @Override
   public void handleProtocolNegotiationCompleted(Attributes attrs) {
     attributes = transportListener.transportReady(attrs);
-    if (keepAliveManager != null) {
-      keepAliveManager.onTransportStarted();
-    }
   }
 
   @VisibleForTesting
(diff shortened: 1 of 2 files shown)


netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -706,7 +704,7 @@ public void onPingAckRead(ChannelHandlerContext ctx, ByteBuf data) throws Http2E
           logger.log(Level.FINE, String.format("Window: %d",
               decoder().flowController().initialWindowSize(connection().connectionStream())));
         }
-      } else {
+      } else if (data.getLong(data.readerIndex()) != KEEPALIVE_PING_BUF.getLong(0)) {
```
ejona86: Use `KEEPALIVE_PING_BUF.equals(data)`?

dapengzhang0: done

dapengzhang0:
PTAL

PR #2991, opened 2017-05-10
Minor updates to expand javadocs

zpencer:
This adds / clarifies javadocs for some classes.

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -746,9 +746,10 @@ public class Context {
      * Returns true if the Context is the current context.
      *
      * @deprecated This method violates some GRPC class encapsulation and should not be used.
-     *     If you must check whether a Context is the current context, check whether it is the same
+     *     If you must know whether a Context is the current context, check whether it is the same
      *     object returned by {@link Context#current()}.
      */
+    //TODO(spencerfang): The superclass's method is package-private, so this should really match.
     @Override
     @Deprecated
     public boolean isCurrent() {
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -742,8 +742,9 @@ public class Context {
       uncancellableSurrogate.detach(toAttach);
     }
 
+    // Visible for testing
     @Override
-    public boolean isCurrent() {
+    boolean isCurrent() {
       return uncancellableSurrogate.isCurrent();
     }
 
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -742,9 +742,16 @@ public class Context {
       uncancellableSurrogate.detach(toAttach);
     }
 
-    // Visible for testing
+    /**
+     * Returns true if the Context is the current context.
+     *
+     * @deprecated This method violates some GRPC class encapsulation and should not be used.
+     * If you must check whether a Context is the current context, check whether it is the same
+     * object returned by {@link Context#current()}.
+     */
     @Override
-    boolean isCurrent() {
+    @Deprecated
+    public boolean isCurrent() {
       return uncancellableSurrogate.isCurrent();
     }
 
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -746,8 +746,8 @@ public class Context {
      * Returns true if the Context is the current context.
      *
      * @deprecated This method violates some GRPC class encapsulation and should not be used.
-     * If you must check whether a Context is the current context, check whether it is the same
-     * object returned by {@link Context#current()}.
+     *     If you must check whether a Context is the current context, check whether it is the same
+     *     object returned by {@link Context#current()}.
      */
     @Override
     @Deprecated
--- a/core/src/main/java/io/grpc/CallOptions.java
+++ b/core/src/main/java/io/grpc/CallOptions.java
@@ -283,7 +283,7 @@ public final class CallOptions {
     }
 
     /**
-     * @return the user supplied default value for this key.
+     * Returns the user supplied default value for this key.
      */
     public T getDefault() {
       return defaultValue;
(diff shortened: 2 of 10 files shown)


context/src/main/java/io/grpc/Context.java
```diff
@@ -719,8 +719,9 @@ public void detach(Context toAttach) {
       uncancellableSurrogate.detach(toAttach);
     }
 
+    // Visible for testing
     @Override
-    public boolean isCurrent() {
+    boolean isCurrent() {
```
carl-mastrangelo: This would be an API breaking change, no?

core/src/main/java/io/grpc/Metadata.java
```diff
@@ -679,6 +691,11 @@ public final String name() {
       return nameBytes;
     }
 
+    /**
+     * @return True if the two objects are both Keys, and their names match.
```
ejona86: Don't use @return here.

ejona86: Don't use @return here. Just say what it does

core/src/main/java/io/grpc/CallOptions.java
```diff
@@ -171,7 +171,7 @@ public CallOptions withAffinity(Attributes affinity) {
 
   /**
    * Enables 'wait for ready' feature for the call.
-   * <a href="https://github.com/grpc/grpc/blob/master/doc/fail_fast.md">'Fail fast'</a>
+   * <a href="https://github.com/grpc/grpc/blob/master/doc/wait-for-ready.md">'wait for ready'</a>
```
zhangkun83: The default is fail-fast, not wait-for-ready.

context/src/main/java/io/grpc/Context.java
zhangkun83: It's not public on the parent class. Being public is a bug, and we should fix it.

core/src/main/java/io/grpc/CallOptions.java
zpencer: Good catch, I should only change the URL not the text

context/src/main/java/io/grpc/Context.java
carl-mastrangelo: The thing about that is that it is public, stable API.  

Even if we could change it, it would be a sign of badwill towards our users.  I vote for leaving it alone.

zpencer: Agreed that this may cause pain for users, will deprecate this instead and add a javadoc about what's happening here.

core/src/main/java/io/grpc/CallOptions.java
dapengzhang0: You can leave `Fail fast` without link and put the link at `wait for ready` in the next line instead:)

context/src/main/java/io/grpc/Context.java
```diff
@@ -719,7 +719,15 @@ public void detach(Context toAttach) {
...
 
+    /**
+     * Returns true if the Context is the current context.
+     *
+     * @deprecated This method violates some GRPC class encapsulation and should not be used.
+     *     If you must check whether a Context is the current context, check whether it is the same
+     *     object returned by {@link Context#current()}.
+     */
     @Override
+    @Deprecated
```
dapengzhang0: also add a `//TODO` for removing the `public` keyword?

core/src/main/java/io/grpc/Metadata.java
```diff
@@ -454,7 +461,12 @@ public void merge(Metadata other) {
     size += other.size;
   }
 
-  /** Merge values for the given set of keys into this set of metadata. */
+  /**
+   * Merge values from the given set of keys into this set of metadata. If a key is present in keys,
+   * then all of the associated values will be copied over.
+   * @param other The source of the new key values.
```
dapengzhang0: need a blank line before the first `@param`?

zpencer: done

core/src/main/java/io/grpc/CompressorRegistry.java
```diff
@@ -49,10 +49,16 @@
       new Codec.Gzip(),
       Codec.Identity.NONE);
 
+  /**
+   * Returns the default instance which supports gzip compression.
```
zhangkun83: I wonder if we really want to mention "gzip" here. Are we allowed to change the default compression in the future? @ejona86

core/src/main/java/io/grpc/ConnectivityStateInfo.java
```diff
@@ -94,6 +94,10 @@ public int hashCode() {
     return state.hashCode() ^ status.hashCode();
   }
 
+  /**
+   * If the {@link Status} is OK, then return the {@link ConnectivityState}. If there is an error,
+   * then return both.
+   */
```
zhangkun83: In general `toString()` is left undocumented, unless you want to strictly define its format so that it can be consumed programmaticly.

core/src/main/java/io/grpc/CompressorRegistry.java
ejona86: How about:
> Returns the default instance used by gRPC when the registry is not specified. Currently the registry just contains support for gzip.

core/src/main/java/io/grpc/Metadata.java
```diff
@@ -679,6 +692,11 @@ public final String name() {
       return nameBytes;
     }
 
+    /**
+     * Returns true if the two objects are both Keys, and their names match.
+     * Note: Upper case letters in the name are normalized to lower case when the Key is
```
ejona86: Instead of this note, could we say that it matches case-insensitively?

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -124,6 +124,10 @@
   private boolean enableStatsTagPropagation;
   private boolean enableTracing;
 
+  /**
+   * Sets the maximum message size allowed for a single gRPC frame. Any messages larger than this
+   * limit will result in a {@link io.grpc.StatusRuntimeException} of type RESOURCE_EXHAUSTED.
```
ejona86: StatusRuntimeException is actually an artifact of the stubs, so I'd rather not say it here.

How about:
If an inbound messages larger than this limit is received it will not be processed and the RPC will fail with RESOURCE_EXHAUSTED.

PR #3001, opened 2017-05-12
Javadoc improvements, especially to Context

ejona86:


context/src/main/java/io/grpc/Context.java
```diff
@@ -228,25 +229,25 @@ private Context(Context parent, Object[][] keyValueEntries, boolean isCancellabl
...
    *         while (!current.isCancelled()) {
    *           keepWorking();
    *         }
    *       }
    *     });
-   *     doSomethingRelatedWork();
-   *   } catch (Throwable t) {
-   *     withCancellation.cancel(t);
+   *   } finally {
+   *     withCancellation.cancel(null);
```
dapengzhang0: This is a burden for users. Why not just call `cancel(null)` in the implementation of `CancellableContext#detach()`?

core/src/main/java/io/grpc/Channel.java
```diff
@@ -34,15 +34,14 @@
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
- * A Channel provides an abstraction over the transport layer that is designed to be consumed
- * by stub implementations. Channel and its associated types {@link ClientCall} and
- * {@link ClientCall.Listener} exchange parsed request and response objects whereas the
- * transport layer only works with serialized data.
+ * A virtual connection to an endpoint, to perform RPCs. A channel is free to have zero or many
+ * actual connections to the endpoint based on configuration, load, etc. Applications are generally
```
zpencer: Is it the case that "endpoint" here can refer to many hosts in the case of load? Is it worth emphasizing that the many actual connections here need not be for the same destination?

context/src/main/java/io/grpc/Context.java
ejona86: Firstly, because that is the API and we can't change it. Secondly, because that would kill using it across threads or asynchronously. Thirdly, it would be quite subtle that if you call withValue(), that context wouldn't cancel on detach. We definitely don't want detach() to imply cancel() because they are commonly separate. In fact, gRPC's usage would break if detach() cancelled.

core/src/main/java/io/grpc/Channel.java
ejona86: Yeah. Done. I'm okay with the language I added, but I would be open to improvements.

PR #3002, opened 2017-05-12
doc: generate use page in javadoc

dapengzhang0:


ejona86:
Have you looked at the output and found it useful?

dapengzhang0:
@ejona86 yes, it will look like this
https://htmlpreview.github.io/?https://github.com/dapengzhang0/grpc-java/blob/temp/all/build/docs/javadoc/io/grpc/ClientInterceptor.html

https://htmlpreview.github.io/?https://raw.githubusercontent.com/dapengzhang0/grpc-java/temp/all/build/docs/javadoc/io/grpc/class-use/ClientInterceptor.html

dapengzhang0:
This is useful in that it allows reader to quickly find out how an instance of the class can be produced and where it can be used as a parameter across all public APIs in the gRPC library, without searching in the source code which includes non-public APIs.

PR #3006, opened 2017-05-15
Fixes for static checker failures

ejona86:
Each commit gives some details. Although normally changes like this are trivial, in this case they aren't, so this should be given normal attention during review.

### DIFF
--- a/grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
+++ b/grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
@@ -51,6 +51,7 @@ import io.grpc.ManagedChannel;
 import io.grpc.Metadata;
 import io.grpc.Status;
 import io.grpc.grpclb.GrpclbConstants.LbPolicy;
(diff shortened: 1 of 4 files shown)


grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -439,19 +439,28 @@ private void handleResponse(LoadBalanceResponse response) {
...
+              "[{0}] : Did not receive response with type initial response: {1}",
+              new Object[] {logId, response});
+          return;
         }
+        initialResponseReceived = true;
+        InitialLoadBalanceResponse initialResponse = response.getInitialResponse();
+        loadReportIntervalMillis =
+            Durations.toMillis(initialResponse.getClientStatsReportInterval());
+        scheduleNextLoadReport();
+        return
```
zhangkun83: missing semicolon

ejona86: Fixed. There were multiple other trivial problems as well.

Issue #2808, opened 2017-03-08
Inner class unit tests in GrpcServerRuleTest look so weird


dapengzhang0:
Should make the inner classes top level classes, and also add `@RunWith(Junit4.class)`

ejona86:
We may be able to set [`scanForTestClasses = false`](https://docs.gradle.org/3.5/dsl/org.gradle.api.tasks.testing.Test.html#org.gradle.api.tasks.testing.Test:scanForTestClasses) to prevent this from reoccuring.

PR #3015, opened 2017-05-17
testing: Add missing RunWith to GrpcServerRuleTest

ejona86:
CC @zhangkun83 @carl-mastrangelo


carl-mastrangelo:
I thought this test was going to be factored @dapengzhang0  ?

dapengzhang0:
#2808 mentioned some problems of `GrpcServerRuleTest`. Missing `RunWith` is one of them. The other is that test cases in inner classes can not run in some test frameworks by default, and it is very hard to config.

ejona86:
So you're saying two things needed to be done, and this is one of the two things to be done. What is the problem?

I'm most bothered this was done internally without being done externally. I see no reason to diverge.

Issue #2243, opened 2016-09-08
Detect zero-message response to unary request on server-side


ejona86:
It [appears](https://github.com/grpc/grpc-java/blob/v1.0.0/core/src/main/java/io/grpc/internal/ServerCallImpl.java#L177) that a server can respond OK to a unary RPC without any detection on the server-side. It'd be nice to throw an exception if close(OK) is called without first sending a message for a UNARY response, although that may be against our API stability. At the very least, we should convert the OK into a Status saying the server is broken.

As reported in grpc/grpc#8010.

PR #3016, opened 2017-05-17
core: ServerCallImpl error if >1 client message in SERVER_STREAMING, split up test suite

zpencer:
- Expand error checking to disallow sending more than 1 client msg if the method type is SERVER_STREAMING.
- ServerCallImplTest only tests against UNARY. Let's split in into sub classes so that different behaviors can be separately tested.

### DIFF
--- a/core/src/test/java/io/grpc/internal/ServerCallImplAbstractTest.java
+++ b/core/src/test/java/io/grpc/internal/ServerCallImplAbstractTest.java
@@ -32,7 +32,10 @@
 package io.grpc.internal;
 
 import static com.google.common.base.Charsets.UTF_8;
+import static org.junit.Assume.assumeTrue;
 
+import com.google.common.base.Function;
+import com.google.common.collect.Lists;
 import com.google.common.io.CharStreams;
 import io.grpc.CompressorRegistry;
 import io.grpc.Context;
@@ -46,15 +49,33 @@ import io.grpc.Status;
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import javax.annotation.Nullable;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.rules.ExpectedException;
+import org.junit.runners.Parameterized.Parameters;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 
+/**
+ * An abstract test class to help run subclass test suites against all applicable
+ * {@link MethodType} classes. Subclasses must implement {{@link #shouldRunTest(MethodType)}}.
+ */
 public abstract class ServerCallImplAbstractTest {
+  private static final List<Object[]> ALL = Lists.transform(
+      Arrays.asList(MethodType.values()),
+      new Function<MethodType, Object[]>() {
+        @Nullable
+        @Override
+        public Object[] apply(@Nullable MethodType input) {
+          return new Object[] {input};
+        }
+      });
 
   @Rule public final ExpectedException thrown = ExpectedException.none();
   @Mock protected ServerStream stream;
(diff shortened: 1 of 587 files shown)


core/src/test/java/io/grpc/internal/ServerCallImplCommonTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015, Google Inc. All rights reserved.
+ * Copyright 2017, Google Inc. All rights reserved.
```
ejona86: Don't update the dates. Updating to 2017 isn't right. If anything, this would be `2015-2017`. But it isn't worth our time. Just make sure that new files have the right date.

core/src/test/java/io/grpc/internal/ServerCallsImplClientSendsOneTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.mockito.Mockito;
+
+@RunWith(Parameterized.class)
+public class ServerCallsImplClientSendsOneTest extends ServerCallImplAbstractTest {
+
+  /**
+   * Unit tests for UNARY and SERVER_STREAMING methods.
+   */
+  @Parameters
+  public static Collection<Object[]> params() {
```
ejona86: All of this extending and param transforming seems convoluted and seems like it will be painful to maintain. Have you considered having a single test with `@Parameters` and do something like `assumeTrue(method.clientSendsOneMessage())` on tests that don't apply?

core/src/test/java/io/grpc/internal/ServerCallImplCommonTest.java
zpencer: I think git turned my file rm and file add into a rename here

ejona86: Yes, git automatically finds renames. In fact, git can't record a rename explicitly. Yes, sometimes it can be wrong, but in this case it seems pretty valid.

core/src/test/java/io/grpc/internal/ServerCallsImplClientSendsOneTest.java
zpencer: Done, was not previously aware of this junit feature. The test classes are much simpler now.

core/src/test/java/io/grpc/internal/ServerCallImplAbstractTest.java
```diff
@@ -0,0 +1,135 @@
...
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+/**
+ * An abstract class to help run subclass test suites against all applicable
+ * {@link MethodType} classes. Subclasses must implement {{@link #shouldRunTest(MethodType)}}.
+ */
+public abstract class ServerCallImplAbstractTest {
+  private static final List<Object[]> ALL = Lists.transform(
```
carl-mastrangelo: This could be shorter as a simple for loop

core/src/test/java/io/grpc/internal/ServerCallsImplClientSendsOneTest.java
```diff
@@ -0,0 +1,71 @@
...
+import io.grpc.MethodDescriptor.MethodType;
+import io.grpc.Status;
+import io.grpc.internal.ServerCallImpl.ServerStreamListenerImpl;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.mockito.Mockito;
+
+@RunWith(Parameterized.class)
+public class ServerCallsImplClientSendsOneTest extends ServerCallImplAbstractTest {
```
carl-mastrangelo: Can you describe a little more why you choose this structure?  It seems like for the change you are making to the SUT needs just one new test method, rather than splitting the whole test class.

core/src/test/java/io/grpc/internal/ServerCallImplAbstractTest.java
zpencer: Good point

core/src/test/java/io/grpc/internal/ServerCallsImplClientSendsOneTest.java
zpencer: The restructuring is motivated by #2243, and the change on SUT is to fix the failing test that is revealed after the refactor.

core/src/test/java/io/grpc/internal/ServerCallImplAbstractTest.java
carl-mastrangelo: Also, we prefer to avoid Guava classes since it makes for dependencies problems.  I would avoid anything in common.collect

zpencer:
Going with a different direction with #3068

PR #3018, opened 2017-05-18
Use periodic builds for release branches

ejona86:
We need to know of external breakages that would prevent testing release branches.

I moved the CI configuration for a new release branch to immediately after the branch is created, as seems appropriate. I also hope it will make it less likely to be forgotten. There's also unrelated link fixes.

I'll submit the changes as individual commits. Reviewing the commits individually is probably easiest due to the move.


ejona86:
@carl-mastrangelo friendly ping

Issue #2873, opened 2017-04-03
InternalSubChannel doesn't comply with the connectivity state spec


zhangkun83:
When receiving a GOAWAY (`transportShutdown()` callback), `InternalSubChannel` unconditionally [transitions to `IDLE`](https://github.com/grpc/grpc-java/blob/v1.2.0/core/src/main/java/io/grpc/internal/InternalSubchannel.java#L439). However, the [spec](https://github.com/grpc/grpc/blob/v1.2.1/doc/connectivity-semantics-and-api.md) says
 1. `READY` transitions to `TRANSIENT_FAILURE` on "any failure encountered while expecting successful communication on established channel."
 1. "Channels that receive a GOAWAY when there are no active or pending RPCs should also switch to IDLE".

`InternalSubchannel` needs to look at the status passed to `transportShutdown()`:
 - If it's not OK, according to Rule 1 above, the new state should be `TRANSIENT_FAILURE`
 - If it's OK, then it's a GOAWAY.
   - If the in-use state of the subchannel is `false`, according to Rule 2 above, the new state should be `IDLE`.
   - Otherwise, the spec doesn't say what the new state should be. Apparently it cannot stay in `READY`. It cannot go to `TRANSIENT_FAILURE` either because it's a GOAWAY, not an error. This part of the spec should be revisited.

zhangkun83:
Also, `InternalSubChannel` doesn't enforce `IDLE_TIMEOUT`. Strictly speaking, it should.

ejona86:
I'm not sure I agree InternalSubChannel should enforce IDLE_TIMEOUT. For all the current LBs, it seems it would just cause trouble, except for OobChannel.

ejona86:
Actually, I could maybe see the benefit for `IDLE_TIMEOUT` for something like GRPCLB, as convenience. However, I think it would probably be an option set explicitly by the LB policy, not inherited from the Channel's settings. That's because it is either useless/harmful for pick-first and round-robin, and for GRPCLB the user's setting of the timeout is probably pretty unrelated to the user's configuration. As far as I'm concerned, though, today the value of `IDLE_TIMEOUT` is infinity for subchannels (so they are complying, but in a useless way, because there has been no benefit to it).

zhangkun83:
The LoadBalancers have the full power to determine channel's connectivity state, making the transitions defined by the spec even less relevant.

PR #3019, opened 2017-05-18
POC: channel state API plumbing

dapengzhang0:
This is only implementing how subchannel state changes are hooked up to channel state APIs. Debates in state semantics and transition rules for loadbalanced channel such as #2873 are not addressed. Any changes in state semantics and transition rules in the future can be easily adapted.

### DIFF
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -22,6 +22,7 @@ import com.google.common.base.Preconditions;
 import java.net.SocketAddress;
 import java.util.ArrayList;
 import java.util.List;
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 import javax.annotation.concurrent.NotThreadSafe;
@@ -462,7 +463,33 @@ public abstract class LoadBalancer {
      * called again and a new picker replaces the old one.  If {@code updatePicker()} has never been
      * called, the channel will buffer all RPCs until a picker is provided.
      */
+    // TODO(zdapeng): add '@deprecated Please migrate ALL usages to {@link #updateBalancingState}'
+    // TODO(zdapeng): and add '@Deprecated'
     public abstract void updatePicker(SubchannelPicker picker);
+
+    /**
+     * Set a new state with a new picker to the channel.
+     *
+     * <p>When a new picker is provided via {@code updateBalancingState()}, the channel will apply
+     * the picker on all buffered RPCs, by calling {@link SubchannelPicker#pickSubchannel(
+     * LoadBalancer.PickSubchannelArgs)}.
+     *
+     * <p>The channel will hold the picker and use it for all RPCs, until {@code
+     * updateBalancingState()} is called again and a new picker replaces the old one.  If {@code
+     * updateBalancingState()} has never been called, the channel will buffer all RPCs until a
+     * picker is provided.
+     *
+     * <p>implSpec: This method must be overridden
+     */
+    public void updateBalancingState(
+        @Nonnull ConnectivityState newState, @Nonnull SubchannelPicker newPicker) {
+      deprecatedUpdatePicker(newPicker);
+    }
+
+    @SuppressWarnings("deprecation") // internal fallback method
+    private void deprecatedUpdatePicker(SubchannelPicker picker) {
+      updatePicker(picker);
+    }
 
     /**
      * Schedule a task to be run in the Channel Executor, which serializes the task with the
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -482,7 +482,10 @@ public abstract class LoadBalancer {
      * updateBalancingState()} has never been called, the channel will buffer all RPCs until a
      * picker is provided.
      *
-     * <p>implSpec: This method must be overridden
+     * <p>The passed state will be the channel's new state. The SHUTDOWN state should not be passed
+     * and its behavior is undefined.
+     *
+     * <p>This method must be overridden
      */
     public void updateBalancingState(
         @Nonnull ConnectivityState newState, @Nonnull SubchannelPicker newPicker) {
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -462,6 +462,9 @@ public abstract class LoadBalancer {
      * <p>The channel will hold the picker and use it for all RPCs, until {@code updatePicker()} is
      * called again and a new picker replaces the old one.  If {@code updatePicker()} has never been
      * called, the channel will buffer all RPCs until a picker is provided.
+     *
+     * <p>Using this method implies that this load balancer doesn't support channel state, and the
+     * application will get exception when trying to get the channel state.
      */
     // TODO(zdapeng): add '@deprecated Please migrate ALL usages to {@link #updateBalancingState}'
     // TODO(zdapeng): and add '@Deprecated'
@@ -483,12 +486,7 @@ public abstract class LoadBalancer {
      */
     public void updateBalancingState(
         @Nonnull ConnectivityState newState, @Nonnull SubchannelPicker newPicker) {
-      deprecatedUpdatePicker(newPicker);
-    }
-
-    @SuppressWarnings("deprecation") // internal fallback method
-    private void deprecatedUpdatePicker(SubchannelPicker picker) {
-      updatePicker(picker);
+      updatePicker(newPicker);
     }
 
     /**
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -484,8 +484,6 @@ public abstract class LoadBalancer {
      *
      * <p>The passed state will be the channel's new state. The SHUTDOWN state should not be passed
      * and its behavior is undefined.
-     *
-     * <p>This method must be overridden
      */
     public void updateBalancingState(
         @Nonnull ConnectivityState newState, @Nonnull SubchannelPicker newPicker) {
--- a/core/src/main/java/io/grpc/ManagedChannel.java
+++ b/core/src/main/java/io/grpc/ManagedChannel.java
@@ -97,7 +97,6 @@ public abstract class ManagedChannel extends Channel {
    *
    * @param source the assumed current state, typically just returned by {@link #getState}
    * @param callback the one-off callback
-   * @throws UnsupportedOperationException if not supported by implementation
    * @since 1.1.0
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/28")
--- a/core/src/main/java/io/grpc/ManagedChannel.java
+++ b/core/src/main/java/io/grpc/ManagedChannel.java
@@ -97,6 +97,7 @@ public abstract class ManagedChannel extends Channel {
    *
    * @param source the assumed current state, typically just returned by {@link #getState}
    * @param callback the one-off callback
+   * @throws UnsupportedOperationException if not supported by implementation
    * @since 1.1.0
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/28")
--- a/core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
+++ b/core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
@@ -16,66 +16,89 @@
 
 package io.grpc.internal;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
 import io.grpc.ConnectivityState;
+import io.grpc.ManagedChannel;
 import java.util.ArrayList;
 import java.util.concurrent.Executor;
+import javax.annotation.Nonnull;
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
- * Book-keeps the connectivity state and callbacks.
+ * Manages connectivity states of the channel. Used for {@link ManagedChannel#getState} to read the
+ * current state of the channel, for {@link ManagedChannel#notifyWhenStateChanged} to add
+ * listeners to state change events, and for {@link io.grpc.LoadBalancer.Helper#updateBalancingState
+ * LoadBalancer.Helper#updateBalancingState} to update the state and run the {@link #updateState}s.
  */
 @NotThreadSafe
-class ConnectivityStateManager {
-  private ArrayList<StateCallbackEntry> callbacks;
+public final class ConnectivityStateManager {
+  private ArrayList<Listener> listeners = new ArrayList<Listener>();
 
-  private ConnectivityState state;
+  private volatile ConnectivityState state = ConnectivityState.IDLE;
 
-  ConnectivityStateManager(ConnectivityState initialState) {
-    state = initialState;
-  }
-
-  void notifyWhenStateChanged(Runnable callback, Executor executor, ConnectivityState source) {
-    StateCallbackEntry callbackEntry = new StateCallbackEntry(callback, executor);
+  /**
+   * Adds a listener for state change event.
+   *
+   * <p>Must be synchronized with {@link #updateState}.
+   *
+   * <p>The {@code executor} must be one that can run RPC call listeners.
+   */
+  public void addListener(Runnable callback, Executor executor, ConnectivityState source) {
+    Listener stateChangeListener = new Listener(callback, executor);
     if (state != source) {
-      callbackEntry.runInExecutor();
+      stateChangeListener.runInExecutor();
     } else {
-      if (callbacks == null) {
-        callbacks = new ArrayList<StateCallbackEntry>();
-      }
-      callbacks.add(callbackEntry);
+      listeners.add(stateChangeListener);
     }
   }
 
-  // TODO(zhangkun83): return a runnable in order to escape transport set lock, in case direct
-  // executor is used?
-  void gotoState(ConnectivityState newState) {
+  /**
+   * Must be synchronized with {@link #addListener}.
+   */
+  public void updateState(@Nonnull ConnectivityState newState) {
+    checkNotNull(newState, "newState");
+    checkState(state != null, "ConnectivityStateManager is disabled");
+
     if (state != newState) {
-      if (state == ConnectivityState.SHUTDOWN) {
-        throw new IllegalStateException("Cannot transition out of SHUTDOWN to " + newState);
-      }
+      checkState(
+          state != ConnectivityState.SHUTDOWN,
+          "Cannot transition out of SHUTDOWN to " + newState);
+
       state = newState;
-      if (callbacks == null) {
+      if (listeners.isEmpty()) {
         return;
       }
       // Swap out callback list before calling them, because a callback may register new callbacks,
       // if run in direct executor, can cause ConcurrentModificationException.
-      ArrayList<StateCallbackEntry> savedCallbacks = callbacks;
-      callbacks = null;
-      for (StateCallbackEntry callback : savedCallbacks) {
-        callback.runInExecutor();
+      ArrayList<Listener> savedListeners = listeners;
+      listeners = new ArrayList<Listener>();
+      for (Listener listener : savedListeners) {
+        listener.runInExecutor();
       }
     }
   }
 
-  ConnectivityState getState() {
+  /**
+   * Gets the current connectivity state of the channel. This method is threadsafe.
+   */
+  public ConnectivityState getState() {
+    if (state == null) {
+      throw new UnsupportedOperationException("Channel state API is not implemented");
+    }
     return state;
   }
 
-  private static class StateCallbackEntry {
+  void disable() {
+    state = null;
+  }
+
+  private static final class Listener {
     final Runnable callback;
     final Executor executor;
 
-    StateCallbackEntry(Runnable callback, Executor executor) {
+    Listener(Runnable callback, Executor executor) {
       this.callback = callback;
       this.executor = executor;
     }
--- a/core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
+++ b/core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
@@ -68,11 +68,7 @@ final class ConnectivityStateManager {
   }
 
   private void gotoNullableState(@Nullable ConnectivityState newState) {
-    if (state != newState) {
-      checkState(
-          state != ConnectivityState.SHUTDOWN,
-          "Cannot transition out of SHUTDOWN to " + newState);
-
+    if (state != newState && state != ConnectivityState.SHUTDOWN) {
       state = newState;
       if (listeners.isEmpty()) {
         return;
--- a/core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
+++ b/core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
@@ -30,10 +30,10 @@ import javax.annotation.concurrent.NotThreadSafe;
  * Manages connectivity states of the channel. Used for {@link ManagedChannel#getState} to read the
  * current state of the channel, for {@link ManagedChannel#notifyWhenStateChanged} to add
  * listeners to state change events, and for {@link io.grpc.LoadBalancer.Helper#updateBalancingState
- * LoadBalancer.Helper#updateBalancingState} to update the state and run the {@link #updateState}s.
+ * LoadBalancer.Helper#updateBalancingState} to update the state and run the {@link #gotoState}s.
  */
 @NotThreadSafe
-public final class ConnectivityStateManager {
+final class ConnectivityStateManager {
   private ArrayList<Listener> listeners = new ArrayList<Listener>();
 
   private volatile ConnectivityState state = ConnectivityState.IDLE;
@@ -41,11 +41,15 @@ public final class ConnectivityStateManager {
   /**
    * Adds a listener for state change event.
    *
-   * <p>Must be synchronized with {@link #updateState}.
+   * <p>Must be synchronized with {@link #gotoState}.
    *
    * <p>The {@code executor} must be one that can run RPC call listeners.
    */
-  public void addListener(Runnable callback, Executor executor, ConnectivityState source) {
+  void notifyWhenStateChanged(Runnable callback, Executor executor, ConnectivityState source) {
+    checkNotNull(callback, "callback");
+    checkNotNull(executor, "executor");
+    checkNotNull(source, "source");
+
     Listener stateChangeListener = new Listener(callback, executor);
     if (state != source) {
       stateChangeListener.runInExecutor();
@@ -55,9 +59,9 @@ public final class ConnectivityStateManager {
   }
 
   /**
-   * Must be synchronized with {@link #addListener}.
+   * Must be synchronized with {@link #notifyWhenStateChanged}.
    */
-  public void updateState(@Nonnull ConnectivityState newState) {
+  void gotoState(@Nonnull ConnectivityState newState) {
     checkNotNull(newState, "newState");
     checkState(state != null, "ConnectivityStateManager is disabled");
 
@@ -90,6 +94,10 @@ public final class ConnectivityStateManager {
     return state;
   }
 
+  /**
+   * Call this method when the channel learns from the load balancer that channel state API is not
+   * supported.
+   */
   void disable() {
     state = null;
   }
--- a/core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
+++ b/core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
@@ -63,11 +63,11 @@ final class ConnectivityStateManager {
    */
   void gotoState(@Nonnull ConnectivityState newState) {
     checkNotNull(newState, "newState");
+    checkState(state != null, "ConnectivityStateManager is already disabled");
     gotoNullableState(newState);
   }
 
   private void gotoNullableState(@Nullable ConnectivityState newState) {
-    checkState(state != null, "ConnectivityStateManager is already disabled");
     if (state != newState) {
       checkState(
           state != ConnectivityState.SHUTDOWN,
@@ -103,9 +103,7 @@ final class ConnectivityStateManager {
    * supported.
    */
   void disable() {
-    if (state != null) {
-      gotoNullableState(null);
-    }
+    gotoNullableState(null);
   }
 
   private static final class Listener {
--- a/core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
+++ b/core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
@@ -41,8 +41,6 @@ final class ConnectivityStateManager {
   /**
    * Adds a listener for state change event.
    *
-   * <p>Must be synchronized with {@link #gotoState}.
-   *
    * <p>The {@code executor} must be one that can run RPC call listeners.
    */
   void notifyWhenStateChanged(Runnable callback, Executor executor, ConnectivityState source) {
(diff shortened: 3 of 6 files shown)


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -582,6 +592,53 @@ private void maybeTerminateChannel() {
...
+      if (!done) {
+        callback.run();
+        done = true;
+      }
+    }
+  }
+
+  @Override
+  public void notifyWhenStateChanged(ConnectivityState source, final Runnable callback) {
+    StateChangeCallBack stateChangCallBack = new StateChangeCallBack(source, callback);
```
ejona86: s/ChangCall/ChangeCall/

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -582,6 +592,53 @@ private void maybeTerminateChannel() {
...
+    if (channelState != source) {
+      executor.execute(stateChangCallBack);
+    } else {
+      stateChangeCallbacks.offer(stateChangCallBack);
+      // check again in case of race with updateChannelState()
```
ejona86: This race handling concerns me as it seems very easy to introduce a bug and not notice. Could you instead use a volatile "State" object. The object would contain the current ConnectivityState and a List of listeners. When the ConnectivityState changes, the object is "completed" and all listeners are run. If listeners are added in the future they would be run immediately.

dapengzhang0: With a volatile "State" object, you still have to take care of race condition of changing state and adding new listeners concurrently, which is still easy to introduce a bug and not notice, unless racy operations are synchronized with a lock.

ejona86: Using a lock for that sounds fine. In fact, getting rid of the volatile and using a lock for the entire state probably wouldn't be an issue. You can maybe also use channelExecutor.

zhangkun83: Since the state-keeping logic is basically the same among all `ManagedChannel` impls, we probably should create some `StateManager` class under `io.grpc.util`.

ejona86: Why expose the logic under `io.grpc.util`? What other implementation would that be for?

dapengzhang0: I feel not comfortable with a lock, if an application call notifyWhenStateChanged too frequently by mistake, the contention on the lock may be strong. If we choose to use channelExecutor, no volatile or lock will be needed, which seems fine.

zhangkun83: `OobChannel`, and any 3rd-party `ManagedChannel` (because we allow it) would need to implement the same logic.

zhangkun83: `io.grpc.util` might not be a good place. I meant any public place where 3rd-party can access.

dapengzhang0: Maybe a package-private `StateManager` in `io.grpc.internal` is sufficient

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: done

lukaszx0:
FWIW, I took a stab at it but unfortunately never managed to finish it: https://github.com/grpc/grpc-java/pull/2768

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -582,6 +592,53 @@ private void maybeTerminateChannel() {
...
+        }
+      }).drain();
+    }
+    return savedChannelState;
+  }
+
+  private static final class StateChangeCallBack implements Runnable {
+    final ConnectivityState source;
+    final Runnable callback;
+    boolean done;
```
rgolangh: is there a reason 'done' isn't private?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -582,6 +592,53 @@ private void maybeTerminateChannel() {
...
+
+    StateChangeCallBack(ConnectivityState source, Runnable callBack) {
+      this.source = source;
+      this.callback = callBack;
+    }
+
+    @Override
+    public synchronized void run() {
```
rgolangh: The same instance will run by 2 different threads?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: It's class is private and final, so any access specifier before `boolean done` effectively implies the same access scope.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: yes, one possibility is the application thread calling `notifyWhenStateChanged`, the other is the transport thread calling `updateChannelState`. Anyway, this Queue implementation was abandoned in favor of ArrayList.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -167,6 +170,46 @@ public void shutdown() {
     private void updatePicker(@Nullable Status error) {
       List<Subchannel> activeList = filterNonFailingSubchannels(getSubchannels());
       helper.updatePicker(new Picker(activeList, error));
+      updateChannelState();
+    }
+
+    private void updateChannelState() {
+      ConnectivityState newState;
```
dapengzhang0: this is a nice trick, but we can not assume any priority on State enum.

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -34,6 +34,7 @@
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
+import com.sun.istack.internal.NotNull;
```
zpencer: Is there anything special about this one versus javax.annotation.Nonnull

dapengzhang0: it was a mistake, should import from `javax.annotation`, thanks.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
rgolangh: Its a bit misleading. The class instance is private to ManageChannelImpl. But its internal state is exposed as well, so anywhere within ManagedChannelImpl you can do `instance.done = false;`

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
rgolangh: I'll catch up with the updates

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
zhangkun83: Even if the field were marked as `private`, it would still be accessible from `ManagedChannelImpl`. To make it worse, compiler will create an accessor for this private field to allow the outer class to access it, leading to larger code size.

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
```diff
@@ -32,65 +32,74 @@
...
-      if (callbacks == null) {
-        return;
-      }
       // Swap out callback list before calling them, because a callback may register new callbacks,
       // if run in direct executor, can cause ConcurrentModificationException.
-      ArrayList<StateCallbackEntry> savedCallbacks = callbacks;
-      callbacks = null;
-      for (StateCallbackEntry callback : savedCallbacks) {
-        callback.runInExecutor();
+      ArrayList<Listener> savedListeners = listeners;
```
dapengzhang0: done

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -582,6 +585,35 @@ private void maybeTerminateChannel() {
...
+              exitIdleMode();
+            }
+          })).drain();
+    }
+    return savedChannelState;
+  }
+
+  @Override
+  public void notifyWhenStateChanged(final ConnectivityState source, final Runnable callback) {
+    if (channelStateManager.getState() == null) {
```
zhangkun83: We don't need this. The caller ready gets the exception from `getState`.

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
```diff
@@ -31,66 +31,89 @@
...
-  void notifyWhenStateChanged(Runnable callback, Executor executor, ConnectivityState source) {
-    StateCallbackEntry callbackEntry = new StateCallbackEntry(callback, executor);
+  /**
+   * Adds a listener for state change event.
+   *
+   * <p>Must be synchronized with {@link #updateState}.
+   *
+   * <p>The {@code executor} must be one that can run RPC call listeners.
+   */
+  public void addListener(Runnable callback, Executor executor, ConnectivityState source) {
```
zhangkun83: I think `notifyWhenStateChanged` is more accurate, as it implies the listener is one-shot. `addLIstener` sounds like the listener will stay after being called.

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -477,9 +478,35 @@ public void updateOobChannelAddresses(ManagedChannel channel, EquivalentAddressG
...
+     *
+     * <p>implSpec: This method must be overridden
+     */
+    public void updateBalancingState(
+        @Nonnull ConnectivityState newState, @Nonnull SubchannelPicker newPicker) {
+      deprecatedUpdatePicker(newPicker);
+    }
+
+    @SuppressWarnings("deprecation") // internal fallback method
+    private void deprecatedUpdatePicker(SubchannelPicker picker) {
```
zhangkun83: `updatePicker()` is not deprecated yet, thus this is not necessary.

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -477,9 +478,35 @@ public void updateOobChannelAddresses(ManagedChannel channel, EquivalentAddressG
      * called again and a new picker replaces the old one.  If {@code updatePicker()} has never been
      * called, the channel will buffer all RPCs until a picker is provided.
      */
+    // TODO(zdapeng): add '@deprecated Please migrate ALL usages to {@link #updateBalancingState}'
```
zhangkun83: Please add a comment to say that using this method implies that this load balancer doesn't support channel state, and the application will get exception when trying to get the channel state.

zhangkun83:
Please also test the channel state plumbing in `ManagedChannelImplTest`, for both implemented and unimplemented cases.

core/src/main/java/io/grpc/LoadBalancer.java
dapengzhang0: done

core/src/main/java/io/grpc/LoadBalancer.java
dapengzhang0: done

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
```diff
@@ -31,66 +31,89 @@
...
+ * current state of the channel, for {@link ManagedChannel#notifyWhenStateChanged} to add
+ * listeners to state change events, and for {@link io.grpc.LoadBalancer.Helper#updateBalancingState
+ * LoadBalancer.Helper#updateBalancingState} to update the state and run the {@link #updateState}s.
  */
 @NotThreadSafe
-class ConnectivityStateManager {
-  private ArrayList<StateCallbackEntry> callbacks;
-
-  private ConnectivityState state;
+public final class ConnectivityStateManager {
```
dapengzhang0: done

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
dapengzhang0: done

dapengzhang0: done

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
```diff
@@ -31,66 +31,89 @@
...
+   */
+  public ConnectivityState getState() {
+    if (state == null) {
+      throw new UnsupportedOperationException("Channel state API is not implemented");
+    }
     return state;
   }
 
-  private static class StateCallbackEntry {
+  void disable() {
```
dapengzhang0: done

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: done

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
```diff
@@ -16,66 +16,97 @@
...
       }
     }
   }
 
-  ConnectivityState getState() {
+  /**
+   * Gets the current connectivity state of the channel. This method is threadsafe.
+   */
+  public ConnectivityState getState() {
+    if (state == null) {
```
ejona86: Only read `state` once; save it to a local variable.

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
```diff
@@ -16,66 +16,97 @@
...
-      callbacks.add(callbackEntry);
+      listeners.add(stateChangeListener);
     }
   }
 
-  // TODO(zhangkun83): return a runnable in order to escape transport set lock, in case direct
-  // executor is used?
-  void gotoState(ConnectivityState newState) {
+  /**
+   * Must be synchronized with {@link #notifyWhenStateChanged}.
```
ejona86: This is extraneous given `@NotThreadSafe`

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
```diff
@@ -16,66 +16,97 @@
...
 
-  ConnectivityStateManager(ConnectivityState initialState) {
-    state = initialState;
-  }
+  private volatile ConnectivityState state = ConnectivityState.IDLE;
 
+  /**
+   * Adds a listener for state change event.
+   *
+   * <p>Must be synchronized with {@link #gotoState}.
```
ejona86: This is extraneous given `@NotThreadSafe`

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -573,6 +577,32 @@ private void maybeTerminateChannel() {
     }
   }
 
+  @Override
+  public ConnectivityState getState(boolean requestConnection) {
+    ConnectivityState savedChannelState = channelStateManager.getState();
+    if (requestConnection && savedChannelState == IDLE) {
+      channelExecutor.executeLater(new LogExceptionRunnable(
```
ejona86: No need for LogExceptionRunnable. This isn't schedule().

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -573,6 +577,32 @@ private void maybeTerminateChannel() {
...
+              exitIdleMode();
+            }
+          })).drain();
+    }
+    return savedChannelState;
+  }
+
+  @Override
+  public void notifyWhenStateChanged(final ConnectivityState source, final Runnable callback) {
+    channelExecutor.executeLater(new LogExceptionRunnable(
```
ejona86: No need for LogExceptionRunnable. This isn't schedule().

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -644,6 +674,23 @@ public void run() {
     }
 
     @Override
+    public void updateBalancingState(
+        final ConnectivityState newState, final SubchannelPicker newPicker) {
+      checkNotNull(newState, "newState");
+      checkNotNull(newPicker, "newPicker");
+
+      runSerialized(new LogExceptionRunnable(
```
ejona86: No need for LogExceptionRunnable. This isn't schedule().

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
dapengzhang0: done

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
dapengzhang0: done

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
dapengzhang0: done

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: done

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: done

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: done

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -1148,6 +1150,86 @@ public void pickerReturnsStreamTracer_delayed() {
...
+
+    assertEquals(ConnectivityState.CONNECTING, channel.getState(false));
+    assertEquals(1, channel.channelExecutor.numPendingTasks());
+
+    assertEquals(ConnectivityState.CONNECTING, channel.getState(true));
+    assertEquals(1, channel.channelExecutor.numPendingTasks());
+  }
+
+  @Test
+  public void getState_loadBalancerDoesNotSupportChannelState() {
```
zhangkun83: You also need a test for this case:
```java
assertEquals(ConnectivityState.IDLE, channel.getState(false|true));
channel.notifyWhenStateChanged(ConnectivityState.IDLE, onStateChanged);
helper.updatePicker(mockPicker);
// verify onStateChanged is called, and getState() throws when called.
```

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -1148,6 +1150,86 @@ public void pickerReturnsStreamTracer_delayed() {
...
+      // do nothing, just adding a mark in the queue of channelExecutor
+      @Override public void run() {}
+    };
+
+    createChannel(new FakeNameResolverFactory(false), NO_INTERCEPTOR);
+    channel.channelExecutor.drain();
+    channel.channelExecutor.executeLater(runnable);
+
+    assertEquals(ConnectivityState.IDLE, channel.getState(false));
+    assertEquals(1, channel.channelExecutor.numPendingTasks());
```
dapengzhang0: Done.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
dapengzhang0: Done.

dapengzhang0:
@zhangkun83 any more comments?

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
```diff
@@ -16,66 +16,103 @@
...
+    }
+    return stateCopy;
+  }
+
+  /**
+   * Call this method when the channel learns from the load balancer that channel state API is not
+   * supported.
+   */
+  void disable() {
+    if (state != null) {
```
zhangkun83: This check doesn't seem necessary.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -1148,6 +1159,97 @@ public void pickerReturnsStreamTracer_delayed() {
...
+
+  @Test
+  public void notifyWhenStateChanged_loadBalancerDoesNotSupportChannelState() {
+    createChannel(new FakeNameResolverFactory(false), NO_INTERCEPTOR);
+    assertEquals(ConnectivityState.IDLE, channel.getState(false));
+
+    Runnable onStateChanged = mock(Runnable.class);
+    channel.notifyWhenStateChanged(ConnectivityState.IDLE, onStateChanged);
+    helper.updatePicker(mockPicker);
+    channel.channelExecutor.drain();
```
zhangkun83: Why need this (and all the other invocations)?

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -1148,6 +1159,97 @@ public void pickerReturnsStreamTracer_delayed() {
...
+    channel.getState(false);
+  }
+
+  @Test
+  public void notifyWhenStateChanged_loadBalancerDoesNotSupportChannelState() {
+    createChannel(new FakeNameResolverFactory(false), NO_INTERCEPTOR);
+    assertEquals(ConnectivityState.IDLE, channel.getState(false));
+
+    Runnable onStateChanged = mock(Runnable.class);
+    channel.notifyWhenStateChanged(ConnectivityState.IDLE, onStateChanged);
```
zhangkun83: You will want to verify that `onStateChanged.run()` is not run after this line.

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
```diff
@@ -16,66 +16,103 @@
...
+
+  private void gotoNullableState(@Nullable ConnectivityState newState) {
+    checkState(state != null, "ConnectivityStateManager is already disabled");
     if (state != newState) {
-      if (state == ConnectivityState.SHUTDOWN) {
-        throw new IllegalStateException("Cannot transition out of SHUTDOWN to " + newState);
-      }
+      checkState(
+          state != ConnectivityState.SHUTDOWN,
+          "Cannot transition out of SHUTDOWN to " + newState);
```
dapengzhang0: If this is the case, it implies that the lb is calling both `helper.updateBalancingState()` and `helper.updatePicker()`, which should be illegal, so it's okey to throw illegal state in this case.

```
     * <p>Using this method implies that this load balancer doesn't support channel state, and the
     * application will get exception when trying to get the channel state.
     */
    public abstract void updatePicker(SubchannelPicker picker);
```

core/src/main/java/io/grpc/ManagedChannel.java
```diff
@@ -97,7 +97,6 @@ public ConnectivityState getState(boolean requestConnection) {
    *
    * @param source the assumed current state, typically just returned by {@link #getState}
    * @param callback the one-off callback
-   * @throws UnsupportedOperationException if not supported by implementation
```
dapengzhang0: done

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
dapengzhang0: done

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
dapengzhang0: removed `channel.channelExecutor.drain();`

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
dapengzhang0: done

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
zhangkun83: What if the channel is shutdown immediately after created? `LoadBalancer` would not have been created, and channel would stay in `IDLE` forever. This is not right.

Since channel outlives the balancer, `SHUTDOWN` state should be triggered by channel, not the balancer. An interesting case is, if balancer reports `SHUTDOWN` but channel is not shutdown yet, what should we do?

dapengzhang0: Oh I haven't thought about that. Per offline discussion, seems lb is not supposed to report `SHUTDOWN` to the channel, and we should let `channel.shutdown()` update state manager with `SHUTDOWN` state.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -93,7 +95,8 @@
   private final ObjectPool<? extends Executor> oobExecutorPool;
   private final LogId logId = LogId.allocate(getClass().getName());
 
-  private final ChannelExecutor channelExecutor = new ChannelExecutor();
+  @VisibleForTesting
+  final ChannelExecutor channelExecutor = new ChannelExecutor();
```
zhangkun83: Please revert this.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -644,6 +681,27 @@ public void run() {
     }
 
     @Override
+    public void updateBalancingState(
+        final ConnectivityState newState, final SubchannelPicker newPicker) {
+      checkNotNull(newState, "newState");
+      checkNotNull(newPicker, "newPicker");
+      if (newState == SHUTDOWN) {
+        // It's unclear if it is appropriate to report SHUTDOWN state from lb. Ignore it for now.
```
zhangkun83: It's definitely inappropriate to report SHUTDOWN from LB. What's unclear is whether we should simply ignore it or do something else.

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -461,10 +462,34 @@ public void updateOobChannelAddresses(ManagedChannel channel, EquivalentAddressG
...
     public abstract void updatePicker(SubchannelPicker picker);
 
     /**
+     * Set a new state with a new picker to the channel.
+     *
+     * <p>When a new picker is provided via {@code updateBalancingState()}, the channel will apply
+     * the picker on all buffered RPCs, by calling {@link SubchannelPicker#pickSubchannel(
+     * LoadBalancer.PickSubchannelArgs)}.
+     *
+     * <p>The channel will hold the picker and use it for all RPCs, until {@code
```
zhangkun83: Please also say that the passed state will be the channel's new state, and that `SHUTDOWN` should not be passed and its behavior is undefined.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -644,6 +680,27 @@ public void run() {
     }
 
     @Override
+    public void updateBalancingState(
+        final ConnectivityState newState, final SubchannelPicker newPicker) {
+      checkNotNull(newState, "newState");
+      checkNotNull(newPicker, "newPicker");
+      if (newState == SHUTDOWN) {
+        // It's not appropriate to report SHUTDOWN state from lb. Ignore it for now.
+        return;
```
zhangkun83: We should probably still accept the new picker, just ignore the state update.

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -461,10 +462,37 @@ public void updateOobChannelAddresses(ManagedChannel channel, EquivalentAddressG
...
+     * updateBalancingState()} is called again and a new picker replaces the old one.  If {@code
+     * updateBalancingState()} has never been called, the channel will buffer all RPCs until a
+     * picker is provided.
+     *
+     * <p>The passed state will be the channel's new state. The SHUTDOWN state should not be passed
+     * and its behavior is undefined.
```
zhangkun83: Please add a test case for it, asserting that SHUTDOWN is ignored.

dapengzhang0: done

dapengzhang0:
Thank you so much for the review @zhangkun83 @ejona86 
ejona86, you have further comments?

dapengzhang0:
Ping @ejona86

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -466,6 +470,12 @@ public ManagedChannelImpl shutdown() {
     if (!shutdown.compareAndSet(false, true)) {
       return this;
     }
+    channelExecutor.executeLater(new Runnable() {
+      @Override
+      public void run() {
+        channelStateManager.gotoState(SHUTDOWN);
```
ejona86: Why isn't this in the `cancelIdleTimer()` Runnable later? It seems it should either be calling `drain()` or the code should just be moved.

core/src/main/java/io/grpc/internal/ConnectivityStateManager.java
```diff
@@ -16,66 +16,97 @@
...
+   * been registered earlier by {@link ManagedChannel#notifyWhenStateChanged}.
+   */
+  void gotoState(@Nonnull ConnectivityState newState) {
+    checkNotNull(newState, "newState");
+    checkState(state != null, "ConnectivityStateManager is already disabled");
+    gotoNullableState(newState);
+  }
+
+  private void gotoNullableState(@Nullable ConnectivityState newState) {
+    if (state != newState && state != ConnectivityState.SHUTDOWN) {
```
dapengzhang0: Otherwise, the current `state == SHUTDOWN` and `newState != SHUTDOWN`. `SHUTDOWN` is the final state, we shouldn't change the state out of `SHUTDOWN`

ejona86: Oh, sorry, it is `state`, not `newState`. Yeah, you have it correct.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
ejona86: `delayedTransport.shutdown()` can do lots of things; it seems strange to single-out adding anything to the `channelExecutor`. But if you add a comment pointing out the subtle usage of `drain()` (I'm unsure whether here or at the drain is better) I could live with it.

bobbymicroby:
@dapengzhang0 @zhangkun83 

Hello, 

I have compiled 1.6.0-SNAPSHOT in order to try the ChannelState plumbing.  I've ran the following code 

```java
//1
 channel.getState(false);

 Observable.interval(5, TimeUnit.SECONDS)
.subscribeOn(Schedulers.computation()).subscribe(l ->
    {
                       
    //2
    channel.getState(false);

    });
```

``//1`` call to channel.getState returns IDLE
``//2`` call throws UnsupportedOperationException in [ConnectivityStateManager.getState()](https://github.com/grpc/grpc-java/blob/7e9e5a55b16205d3acfc47f2247aaa4d634f7522/core/src/main/java/io/grpc/internal/ConnectivityStateManager.java#L92)


Do you have any idea why the ```ConnectivityState state``` in  ```ConnectivityStateManager``` is null during the second call, which leads to the UnsupportedOperationException branch in ```ConnectivityStateManager.getState``` ?

Regards, B

ejona86:
@bobbymicroby, all existing Load Balancers still have to be updated to call `LoadBalancer.Helper.updateBalancingState()` instead of `LoadBalancer.Helper.updatePicker()`. This change only enabled Load Balancers to report the state; it didn't update any Load Balancers to _actually_ report the state.

dapengzhang0:
@bobbymicroby
If you call `getState` at the very early stage of the channel, it will return `IDLE` regardless of whether channel state API is supported or not, because at this point of time, the channel has not interacting with nameresolver/loadbalancer yet and is not able to know if it is supported (by load balancer) yet.
The second call of `getState` throws because the channel state API is not supported.

bobbymicroby:
@ejona86 @dapengzhang0 Thanks for the information. I will wait for the complete channel state release in the future. Great work guys, thanks again.

PR #3021, opened 2017-05-18
core,netty,okhttp: detect proxy via ProxySelector

zpencer:
For DNS names: use the system wide ProxySelector to determine the proxy. GRPC_PROXY_EXP is treated as an override, but will be removed in the future.


core/src/main/java/io/grpc/internal/Proxies.java
```diff
@@ -0,0 +1,138 @@
...
+  /**
+   * Given a target address, returns which proxy address should be used. If no proxy should be
+   * used, then return null.
+   */
+  public static InetSocketAddress proxyFor(SocketAddress targetServerAddress) {
+    return proxyFor(
+        targetServerAddress,
+        ProxySelector.getDefault(),
+        System.getenv(GRPC_PROXY_ENV_VAR)
+    );
```
ejona86: We mostly put the parenthesis on the same line as the last argument. Since you can't have a comma after the getenv, the newline doesn't really seem to help anything.

core/src/main/java/io/grpc/internal/Proxies.java
```diff
@@ -0,0 +1,138 @@
...
+
+  /**
+   * Helper method to achieve stateless unit tests.
+   */
+  @VisibleForTesting
+  static InetSocketAddress proxyFor(
+      SocketAddress targetServerAddress,
+      ProxySelector proxySelector,
+      @Nullable String proxyEnvStr
+  ) {
```
ejona86: Ditto.

core/src/main/java/io/grpc/internal/Proxies.java
```diff
@@ -0,0 +1,138 @@
...
+
+  private static InetSocketAddress detectProxy(
+      ProxySelector proxySelector,
+      SocketAddress targetServerAddress
+  ) {
+    if (!(targetServerAddress instanceof InetSocketAddress)) {
+      return null;
+    }
+
+    String hostName = ((InetSocketAddress) targetServerAddress).getHostName();
```
ejona86: If the socket address was created with an IP address this can do I/O. If you only expect this class to work with `DnsNameResolver`, you could `assert address.isUnresolved()`. But this problem has been causing some havoc. There is getHostString(), but it's not available in Java 6.

core/src/main/java/io/grpc/internal/Proxies.java
```diff
@@ -0,0 +1,138 @@
...
+    URI uri;
+    try {
+      uri = new URI(String.format(URI_FORMAT, hostName));
```
ejona86: This string.format does not work for IPv6 addresses. Use `GrpcUtil.authorityFromHostAndPort()`

ejona86: This should include the port. The port may change the proxy decision.

core/src/main/java/io/grpc/internal/Proxies.java
```diff
@@ -0,0 +1,138 @@
...
+    } catch (final URISyntaxException e) {
+      log.warning(String.format(
```
ejona86: Always try to use the logger's formatter, as it is good habit to reduce chances of doing something like this with INFO and lower which will cause useless computation:

```java
log.log(Level.WARNING, "blah: {0}", hostName);
```

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -365,20 +366,16 @@ void overrideAuthorityChecker(@Nullable OverrideAuthorityChecker authorityChecke
...
+    SocketAddress proxy = Proxies.proxyFor(targetServerAddress);
     if (proxy != null) {
-      String[] parts = proxy.split(":", 2);
-      int port = 80;
-      if (parts.length > 1) {
-        port = Integer.parseInt(parts[1]);
-      }
-      InetSocketAddress proxyAddress = new InetSocketAddress(parts[0], port);
-      negotiator = ProtocolNegotiators.httpProxy(proxyAddress, null, null, negotiator);
+      negotiator = ProtocolNegotiators.httpProxy(proxy, null, null, negotiator);
```
ejona86: ProxySelector assumes you will be using Authenticator as well. We can break setups if we use ProxySelector without using Authenticator.

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -365,20 +366,16 @@ void overrideAuthorityChecker(@Nullable OverrideAuthorityChecker authorityChecke
...
+      SocketAddress targetServerAddress,
       String authority,
       NegotiationType negotiationType,
-      SslContext sslContext) {
+      SslContext sslContext
+  ) {
     ProtocolNegotiator negotiator =
         createProtocolNegotiatorByType(authority, negotiationType, sslContext);
-    String proxy = System.getenv("GRPC_PROXY_EXP");
+    SocketAddress proxy = Proxies.proxyFor(targetServerAddress);
```
ejona86: This makes me worried for tests, since this uses ProxySelector with its global auto-configuration.

core/src/main/java/io/grpc/internal/Proxies.java
zpencer: done

core/src/main/java/io/grpc/internal/Proxies.java
zpencer: Ah didn't realize this was the way to do it here, it's a bit different from log4j

core/src/main/java/io/grpc/internal/Proxies.java
zpencer: done, and audited rest of changes

core/src/main/java/io/grpc/internal/Proxies.java
zpencer: done

zpencer: done

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
zpencer: Updated diff to use dependency injection to avoid static calls

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
zpencer: Updated diff to include changes for Authenticator, which was originally intended separate diff.

core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java
```diff
@@ -0,0 +1,213 @@
...
+    String hostPort =
+        GrpcUtil.authorityFromHostAndPort(targetAddr.getHostName(), targetAddr.getPort());
+    URI uri;
+    try {
+      uri = new URI(String.format(URI_FORMAT, hostPort));
+    } catch (final URISyntaxException e) {
+      log.log(
+          Level.WARNING,
+          "Failed to construct URI for proxy lookup, proceeding without proxy: {0}",
+          hostPort);
```
zhangkun83: It will be useful to log the exception.

core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java
```diff
@@ -0,0 +1,213 @@
...
+
+    if (auth == null) {
+      return new ProxyParameters(proxyAddr, null, null);
+    }
+
+    try {
+      //todo(spencerfang): users of ProxyParameters should clear the password when done
+      return new ProxyParameters(proxyAddr, auth.getUserName(), new String(auth.getPassword()));
+    } finally {
+      Arrays.fill(auth.getPassword(), '0');
```
zhangkun83: Why is this necessary?

core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
+
+/**
+ * A utility class that detects proxies using {@link ProxySelector} and detects authentication
+ * credentials using {@link Authenticator}.
+ *
+ */
+public class ProxyDetectorImpl implements ProxyDetector {
```
zhangkun83: This can be package-private.

core/src/main/java/io/grpc/internal/ProxyDetector.java
```diff
@@ -0,0 +1,90 @@
...
+package io.grpc.internal;
+
+import com.google.common.base.Objects;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+import javax.annotation.Nullable;
+
+/**
+ * A utility class to detect which proxy, if any, should be used for a given
+ * {@link java.net.InetSocketAddress}.
```
zpencer: Done

core/src/main/java/io/grpc/internal/ProxyDetector.java
```diff
@@ -0,0 +1,90 @@
...
+  ProxyDetector NOOP_INSTANCE = new ProxyDetector() {
+    @Override
+    public ProxyParameters proxyFor(SocketAddress targetServerAddress) {
+      return null;
+    }
+  };
+
+  /**
+   * Given a target address, returns which proxy address should be used. If no proxy should be
+   * used, then return value will be an absent value.
```
zpencer: Done

core/src/main/java/io/grpc/internal/ProxyDetector.java
```diff
@@ -0,0 +1,90 @@
...
+   */
+  ProxyDetectorImpl.ProxyParameters proxyFor(SocketAddress targetServerAddress);
```
zpencer: Done

core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java
```diff
@@ -0,0 +1,213 @@
...
+    PasswordAuthentication requestPasswordAuthentication(
+        String host,
+        InetAddress addr,
+        int port,
+        String protocol,
+        String prompt,
+        String scheme);
+  }
+
+  static class AuthenticationProviderImpl implements AuthenticationProvider {
```
zpencer: Done, there's no reason why this can't be anonymous

core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java
zpencer: The javadoc for PasswordAuthentication recommends clearing the password after it is used. Even though gRPC doesn't treat passwords in a special way today, I think having one fewer place where passwords linger around is better.

zpencer:
There's some more changes coming to this branch to avoid using ProxyDetector twice, and instead plumb the ProxyParameters into the ClientTransportFactory via Attributes.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -160,9 +164,10 @@ public void run() {
           resolving = true;
         }
         try {
-          if (System.getenv("GRPC_PROXY_EXP") != null) {
-            EquivalentAddressGroup server =
-                new EquivalentAddressGroup(InetSocketAddress.createUnresolved(host, port));
+          InetSocketAddress destination = InetSocketAddress.createUnresolved(host, port);
+          ProxyParameters proxyParams = proxyDetector.proxyFor(destination);
+          if (proxyParams != null) {
+            EquivalentAddressGroup server = new EquivalentAddressGroup(destination);
```
zpencer: #3030 will change the flow here

zpencer:
PTAL: This diff now addresses case 1 and 2, while making it easy to implement 3 in the future. 

`ClientTransportFactory #newClientTransport` will now require a `ProxyParameters` argument. Implementations are responsible for correctly connecting to `serverAddress` via the specified proxy.

For case 1:
- This happens when the DNS name is associated with a HTTP proxy. `DnsNameResolver` will see that a proxy should be used, and there will be 1 EAG containing a single entry, the unresolved destination address. This is a bit different from c-core, because in c-core the resolved address is the proxy's address.
- In general, the name resolver implementation is responsible for figuring out whether a proxy is used. We can't do the check in MangedChannelImpl because the `target` string is opaque, and it is up to the NameResolver implementations to turn the string into host(s).
- Later when it is time for the `InternalSubchannel` to connect, the class will consult with the `ProxyDetector` to determine the proxy settings to pass into `newClientTransport`.

For case 2:
- This happens when the DNS name is not associated with any proxies, and `DnsNameResolver` is able to resolve the name to IPs and ports.
- The `ProxyDetector` must have the IP:PORT resolution results associated with the correct proxy values, so that `InternalSubchannel` can look up the proxy settings to pass into `newClientTransport`.

For case 3:
- Connecting to the LB: As long as `ProxyDetector` has a proxy associated with the LB's hostname, the name resolver will leave the name unresolved. `GrpcLoadBalancer` uses `InternalSubchannel` to create out of band channels with the LB, which means the proxy settings will be correctly autodetected.
- Connecting to the backends: As long as each backend server returned by the LB each have a proxy associated with it, `InternalSubchannel` will detect the proxy and pass it into `newClientTransport`.

zpencer:
ping @ejona86

core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java
```diff
@@ -0,0 +1,178 @@
...
+ */
+class ProxyDetectorImpl implements ProxyDetector {
+  private static final Logger log = Logger.getLogger(ProxyDetectorImpl.class.getName());
+  private static final AuthenticationProvider DEFAULT_AUTHENTICATOR = new AuthenticationProvider() {
+    @Override
+    public PasswordAuthentication requestPasswordAuthentication(
+        String host, InetAddress addr, int port, String protocol, String prompt, String scheme) {
+      // TODO(spencerfang): consider using java.security.AccessController here
+      return Authenticator.requestPasswordAuthentication(
+          host, addr, port, protocol, prompt, scheme);
```
ejona86: You should specify RequestorType.PROXY

core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java
```diff
@@ -0,0 +1,178 @@
...
+      log.warning("More than 1 proxy detected, gRPC will select the first one");
+    }
+    Proxy proxy = proxies.get(0);
+
+    if (proxy.type() == Proxy.Type.DIRECT) {
+      return null;
+    }
+    InetSocketAddress proxyAddr = (InetSocketAddress) proxy.address();
+    PasswordAuthentication auth = authenticationProvider.requestPasswordAuthentication(
+        proxyAddr.getHostName(),
```
ejona86: This may do a DNS request. I wonder if this is an appropriate time to do reflection and check whether `getHostString()` is available... Java 6 does `getHostName()` here, but Java 7 swaps to `getHostString()`.

core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java
```diff
@@ -0,0 +1,178 @@
...
+        proxyAddr.getAddress(),
+        proxyAddr.getPort(),
+        PROXY_SCHEME,
+        "proxy authentication required by gRPC",
```
ejona86: This is supposed to be the realm returned by the server. We won't have it since we are avoiding the full authentication handshake. We maybe should use `""`, but at the very least it deserves a comment.

PR #3022, opened 2017-05-19
protobuf: make toStatusProto return null when metadata key is absent

ericgribkoff:


protobuf/src/main/java/io/grpc/protobuf/StatusProto.java
```diff
@@ -158,7 +158,7 @@ private static Metadata toMetadata(com.google.rpc.Status statusProto, Metadata m
 
   @Nullable
   private static com.google.rpc.Status toStatusProto(Status status, Metadata trailers) {
-    if (trailers != null) {
+    if (trailers != null && trailers.containsKey(STATUS_DETAILS_KEY)) {
       com.google.rpc.Status statusProto = trailers.get(STATUS_DETAILS_KEY);
       checkArgument(
```
carl-mastrangelo: The check for null should go here.  metadata uses a lot of linear time accesses, so doing it twice is reaching.

ericgribkoff: Done, PTAL.

ericgribkoff:
Friendly ping...this fixes a null pointer exception, so would like to get this in before the branch cut.

protobuf/src/main/java/io/grpc/protobuf/StatusProto.java
```diff
@@ -160,10 +160,12 @@ private static Metadata toMetadata(com.google.rpc.Status statusProto, Metadata m
   private static com.google.rpc.Status toStatusProto(Status status, Metadata trailers) {
     if (trailers != null) {
       com.google.rpc.Status statusProto = trailers.get(STATUS_DETAILS_KEY);
-      checkArgument(
-          status.getCode().value() == statusProto.getCode(),
-          "com.google.rpc.Status code must match gRPC status code");
-      return statusProto;
+      if (statusProto != null) {
+        checkArgument(
+                status.getCode().value() == statusProto.getCode(),
```
carl-mastrangelo: indent -4.

protobuf/src/main/java/io/grpc/protobuf/StatusProto.java
```diff
@@ -160,10 +160,12 @@ private static Metadata toMetadata(com.google.rpc.Status statusProto, Metadata m
   private static com.google.rpc.Status toStatusProto(Status status, Metadata trailers) {
     if (trailers != null) {
```
carl-mastrangelo: Does there need to be a `checkNotNull(status, "status")` here?

ericgribkoff: No, it's not a nullable field of StatusException or StatusRuntimeException.

protobuf/src/main/java/io/grpc/protobuf/StatusProto.java
ericgribkoff: I can do this when I'm at a computer later tonight. I think the +8 was the result of running google format, and allowed by the style guide when there is more than two continuation lines.

ericgribkoff: Indentation fixed.

PR #3024, opened 2017-05-19
Fix incorrect assertEquals argument ordering

ejona86:
The expected value should be first.


ejona86:
@zpencer, friendly ping.

PR #2071, opened 2016-07-20
core: Hard-code a list of providers for Android

ejona86:
Class.getResource() is expensive on Android, which is used by
ServiceLoader. So we hard-code a list for Android instead.

Fixes #2037

PR #3029, opened 2017-05-23
core: Simplify ChannelImpl constructor; testable Builder

ejona86:
Add some example tests for easier fields in AbstractManagedChannelImplBuilder.

Many fields are no longer Nullable, in order to move logic from construction to
mutation, which eases testing and simplifies cross-class interactions.

The nameResolverFactory comment starting 'Avoid loading the provider unless
necessary' was outdated and has not been a concern since #2071 which swapped to
a hard-coded list on Android.

Consider this a working RFC. If we like it, there are many more places we can apply
the same pattern and apply more cleanups. Specific details can maybe be done
differently, but before harping on the details, it'd be good to determine if we like the
general approach.

This is low priority :smile:

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
@@ -16,7 +16,6 @@
 
 package io.grpc.internal;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -75,34 +74,45 @@ public abstract class AbstractManagedChannelImplBuilder
   @VisibleForTesting
   static final long IDLE_MODE_MIN_TIMEOUT_MILLIS = TimeUnit.SECONDS.toMillis(1);
 
-  @Nullable
-  private Executor executor;
+  private static final ObjectPool<? extends Executor> DEFAULT_EXECUTOR_POOL =
+      SharedResourcePool.forResource(GrpcUtil.SHARED_CHANNEL_EXECUTOR);
+
+  private static final NameResolver.Factory DEFAULT_NAME_RESOLVER_FACTORY =
+      NameResolverProvider.asFactory();
+
+  private static final LoadBalancer.Factory DEFAULT_LOAD_BALANCER_FACTORY =
+      PickFirstBalancerFactory.getInstance();
+
+  private static final DecompressorRegistry DEFAULT_DECOMPRESSOR_REGISTRY =
+      DecompressorRegistry.getDefaultInstance();
+
+  private static final CompressorRegistry DEFAULT_COMPRESSOR_REGISTRY =
+      CompressorRegistry.getDefaultInstance();
+
+  ObjectPool<? extends Executor> executorPool = DEFAULT_EXECUTOR_POOL;
 
   private final List<ClientInterceptor> interceptors = new ArrayList<ClientInterceptor>();
 
-  private final String target;
+  final String target;
 
   @Nullable
   private final SocketAddress directServerAddress;
 
   @Nullable
-  private String userAgent;
+  String userAgent;
 
   @Nullable
-  private String authorityOverride;
+  String authorityOverride;
 
-  @Nullable
-  private NameResolver.Factory nameResolverFactory;
+  NameResolver.Factory nameResolverFactory = DEFAULT_NAME_RESOLVER_FACTORY;
 
-  private LoadBalancer.Factory loadBalancerFactory;
+  LoadBalancer.Factory loadBalancerFactory = DEFAULT_LOAD_BALANCER_FACTORY;
 
-  @Nullable
-  private DecompressorRegistry decompressorRegistry;
+  DecompressorRegistry decompressorRegistry = DEFAULT_DECOMPRESSOR_REGISTRY;
 
-  @Nullable
-  private CompressorRegistry compressorRegistry;
+  CompressorRegistry compressorRegistry = DEFAULT_COMPRESSOR_REGISTRY;
 
-  private long idleTimeoutMillis = IDLE_MODE_DEFAULT_TIMEOUT_MILLIS;
+  long idleTimeoutMillis = IDLE_MODE_DEFAULT_TIMEOUT_MILLIS;
 
   private int maxInboundMessageSize = GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
 
@@ -162,7 +172,11 @@ public abstract class AbstractManagedChannelImplBuilder
 
   @Override
   public final T executor(Executor executor) {
-    this.executor = executor;
+    if (executor != null) {
+      this.executorPool = new FixedObjectPool<Executor>(executor);
+    } else {
+      this.executorPool = DEFAULT_EXECUTOR_POOL;
+    }
     return thisT();
   }
 
@@ -182,7 +196,11 @@ public abstract class AbstractManagedChannelImplBuilder
     Preconditions.checkState(directServerAddress == null,
         "directServerAddress is set (%s), which forbids the use of NameResolverFactory",
         directServerAddress);
-    this.nameResolverFactory = resolverFactory;
+    if (resolverFactory != null) {
+      this.nameResolverFactory = resolverFactory;
+    } else {
+      this.nameResolverFactory = DEFAULT_NAME_RESOLVER_FACTORY;
+    }
     return thisT();
   }
 
@@ -191,19 +209,31 @@ public abstract class AbstractManagedChannelImplBuilder
     Preconditions.checkState(directServerAddress == null,
         "directServerAddress is set (%s), which forbids the use of LoadBalancer.Factory",
         directServerAddress);
-    this.loadBalancerFactory = loadBalancerFactory;
+    if (loadBalancerFactory != null) {
+      this.loadBalancerFactory = loadBalancerFactory;
+    } else {
+      this.loadBalancerFactory = DEFAULT_LOAD_BALANCER_FACTORY;
+    }
     return thisT();
   }
 
   @Override
   public final T decompressorRegistry(DecompressorRegistry registry) {
-    this.decompressorRegistry = registry;
+    if (registry != null) {
+      this.decompressorRegistry = registry;
+    } else {
+      this.decompressorRegistry = DEFAULT_DECOMPRESSOR_REGISTRY;
+    } 
     return thisT();
   }
 
   @Override
   public final T compressorRegistry(CompressorRegistry registry) {
-    this.compressorRegistry = registry;
+    if (registry != null) {
+      this.compressorRegistry = registry;
+    } else {
+      this.compressorRegistry = DEFAULT_COMPRESSOR_REGISTRY;
+    }
     return thisT();
   }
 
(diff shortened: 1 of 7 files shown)


ejona86:
retest this please

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -177,7 +187,11 @@ public final T directExecutor() {
 
   @Override
   public final T executor(Executor executor) {
-    this.executor = executor;
+    if (executor != null) {
```
zpencer: MoreObjects#firstNonNull() can make this a bit less verbose

zpencer:
I support this change

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
ejona86: That's true, but it also then hides condition that needs testing. I'm in favor of that for checkNotNull, where the case is just a distraction. Or like the previous usages of `firstNonNull()`, since the code was so complex already. I could go either way, but I chose the boring approach that makes it more obvious when we're skipping test cases.

core/src/main/java/io/grpc/internal/FixedObjectPool.java
```diff
@@ -0,0 +1,55 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.grpc.internal;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * An object pool that always returns the same instance and does nothing when returned the object.
```
zhangkun83: nit: s/returned/returning

ejona86: Done.

ejona86:
@carl-mastrangelo, do you want to take a glance to see if you like this approach? I'm fine with just @zhangkun83 reviewing, but wanted to make sure we all thought this was an improvement.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -177,7 +187,11 @@ public final T directExecutor() {
...
+      this.executorPool = new FixedObjectPool<Executor>(executor);
```
ejona86: SharedResourcePool is somewhat similar to a pool, but I agree it still isn't really. The Pool API and semantics do actually match a pool pretty well, though, so I'm having trouble coming up with an alternative name. I'm open for a rename in the future.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -379,43 +379,49 @@ public void run() {
...
+      tmpNameResolverFactory = new OverrideAuthorityNameResolverFactory(
+          tmpNameResolverFactory, builder.authorityOverride);
+    }
+    this.nameResolverFactory = tmpNameResolverFactory;
+    this.nameResolverParams = checkNotNull(builder.getNameResolverParams(), "nameResolverParams");
     this.nameResolver = getNameResolver(target, nameResolverFactory, nameResolverParams);
-    this.loadBalancerFactory = checkNotNull(loadBalancerFactory, "loadBalancerFactory");
-    this.executorPool = checkNotNull(executorPool, "executorPool");
+    this.loadBalancerFactory =
+        checkNotNull(builder.loadBalancerFactory, "loadBalancerFactory");
```
carl-mastrangelo: These checkNotNulls are no longer necessary right?

core/src/main/java/io/grpc/internal/OverrideAuthorityNameResolverFactory.java
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+   * Constructor for the {@link NameResolver.Factory}
+   *
+   * @param delegate The actual underlying factory that will produce the a {@link NameResolver}
+   * @param authorityOverride The authority that will be returned by {@link
+   *   NameResolver#getServiceAuthority()}
+   */
+  OverrideAuthorityNameResolverFactory(NameResolver.Factory delegate,
+      String authorityOverride) {
```
carl-mastrangelo: fits on prev line?

core/src/main/java/io/grpc/internal/OverrideAuthorityNameResolverFactory.java
```diff
@@ -0,0 +1,90 @@
...
+  @Nullable
+  @Override
+  public NameResolver newNameResolver(URI targetUri, Attributes params) {
+    final NameResolver resolver = delegate.newNameResolver(targetUri, params);
+    // Do not wrap null values. We do not want to impede error signaling.
+    if (resolver == null) {
+      return null;
+    }
+    return new NameResolver() {
+      @Override
```
carl-mastrangelo: Maybe forward the other methods too, like `refresh`

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
ejona86: Without a nullability checker, I feel like they still serve a purpose. It seems like it would be pretty easy to change the builder without investigating whether this class supported null values.

core/src/main/java/io/grpc/internal/OverrideAuthorityNameResolverFactory.java
ejona86: This is pre-existing code, but you're right. That's a bug. Let's fix that separately to avoid hidden behavior changes in this PR. Filed #3061

core/src/main/java/io/grpc/internal/OverrideAuthorityNameResolverFactory.java
ejona86: Done.

ejona86:
I do see I need to fix the headers of the new files due to the license change. I'll do that after reviews though, so I can rebase. If I do it before, checkstyle will fail in my branch.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -90,34 +89,45 @@
...
+
+  private static final CompressorRegistry DEFAULT_COMPRESSOR_REGISTRY =
+      CompressorRegistry.getDefaultInstance();
+
+  ObjectPool<? extends Executor> executorPool = DEFAULT_EXECUTOR_POOL;
 
   private final List<ClientInterceptor> interceptors = new ArrayList<ClientInterceptor>();
 
-  private final String target;
+  final String target;
```
dapengzhang0: What about make these fields private and define a helper class for test, which puts their getters and setters together? This can prevent other src classes in the same package from using these fields.
````
class TestTool {
  String getTarget() {
    return target;
  }
  void setTarget(String newTarget) {
    target = newTarget;
  }
  ...
}

@VisibleForTesting final TestTool testTool = new TestTool();
````

Comment (ejona86) [inline core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java]: I don't follow. This field and the others are directly accessed by ManagedChannelImpl.

Also, I wouldn't want to add two methods per field to prevent other usages in the package. I would instead probably just put `@VisibleForTesting` on each.

Comment (dapengzhang0) [inline core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java]: my bad, I thought they were used only for test as before. As they are directly accessed by ManagedChannelImpl, then maybe they are not qualified for `@VisibleForTesting` any more.

Comment (ejona86) [inline core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java]: Right. They are not `@VisibleForTesting`.

**Comment by ejona86** [comment]
@carl-mastrangelo, @dapengzhang0 any more comments? If you want more time, that's fine. I'm just wanting to avoid delay if everyone is already fine with the change.

File: core/src/main/java/io/grpc/internal/FixedObjectPool.java [hunk 8]
```diff
@@ -0,0 +1,55 @@
...
+
+import com.google.common.base.Preconditions;
+
+/**
+ * An object pool that always returns the same instance and does nothing when returning the object.
+ */
+final class FixedObjectPool<T> implements ObjectPool<T> {
+  private final T object;
+
+  public FixedObjectPool(T object) {
```
carl-mastrangelo: s/public//

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -197,7 +211,11 @@ public final T nameResolverFactory(NameResolver.Factory resolverFactory) {
     Preconditions.checkState(directServerAddress == null,
         "directServerAddress is set (%s), which forbids the use of NameResolverFactory",
         directServerAddress);
-    this.nameResolverFactory = resolverFactory;
+    if (resolverFactory != null) {
```
dapengzhang0: Is it an unbannable usecase to call `builder.nameResolverFactory(null)`? Can we just throw if the argument `resolverFactory` is null?

core/src/main/java/io/grpc/internal/FixedObjectPool.java
ejona86: The two are equivalent, and I don't like to assume the class's visibility in individual fields/methods. The method is public because all users of the class are expected/allowed to use the method.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
ejona86: I'm purposefully avoiding changing the externally-visible behavior of this class. The previous code allowed passing in `null` as a way to choose the default, so the new code maintains that behavior.

core/src/main/java/io/grpc/internal/FixedObjectPool.java
carl-mastrangelo: It's confusing to have different visibility.  Also this version is more verbose.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
dapengzhang0: I think these experimental APIs are rarely used, and the javadoc didn't say what's the behavior for null argument, so if we change the behavior now, there shouldn't be impact.

core/src/main/java/io/grpc/internal/FixedObjectPool.java
ejona86: I feel like this is a style decision. It is not against the style guide and I find it more confusing to use the package-private visibility.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
ejona86: I strongly avoid unnecessary behavior changes in refactoring PRs, and hope others do the same. If you'd like to send out a PR or make an issue to change the behavior of `null`, that can be done.

dapengzhang0: That's fair.

core/src/main/java/io/grpc/internal/FixedObjectPool.java
carl-mastrangelo: I know, but don't you like shorter code?  Don't you like consistent code?

core/src/main/java/io/grpc/internal/OverrideAuthorityNameResolverFactory.java
```diff
@@ -0,0 +1,89 @@
...
+  private final String authorityOverride;
+
+  /**
+   * Constructor for the {@link NameResolver.Factory}
+   *
+   * @param delegate The actual underlying factory that will produce the a {@link NameResolver}
+   * @param authorityOverride The authority that will be returned by {@link
+   *   NameResolver#getServiceAuthority()}
+   */
+  OverrideAuthorityNameResolverFactory(NameResolver.Factory delegate, String authorityOverride) {
```
carl-mastrangelo: this is package private too, which I expect

core/src/test/java/io/grpc/internal/AbstractManagedChannelImplBuilderTest.java
```diff
@@ -50,6 +58,173 @@
 /** Unit tests for {@link AbstractManagedChannelImplBuilder}. */
 @RunWith(JUnit4.class)
 public class AbstractManagedChannelImplBuilderTest {
+  private Builder builder = new Builder("fake");
+  private Builder directAddressBuilder = new Builder(mock(SocketAddress.class), "fake");
```
carl-mastrangelo: new SocketAddress(){}

?

ejona86: Oh, nice. Done.

core/src/main/java/io/grpc/internal/OverrideAuthorityNameResolverFactory.java
ejona86: This is not new code; it was just a move. I left it as-is.

core/src/main/java/io/grpc/internal/FixedObjectPool.java
ejona86: I feel no need to defend myself and I don't think you are likely to change your mind. I'm okay with that.

I favor clearer code, and I find it clearer to consumers of the class as public. I also favor separation of concerns; if a class is to become public, why should it be necessary to change method visibility? I would not use public instead of protected nor package-private instead of private simply because it is shorter; the fact package-private is written with zero characters is irrelevant to me.

ejona86:
Thank you everyone! I agree there are still further improvements to be made, but we can iterate.

PR #3030, opened 2017-05-23
core,netty,okhttp: plumb Attributes from NameResolver to ClientTransportFactory

zpencer:
When DnsNameResolver resolves a name, it notifies NameResolver.Listener#onAddresses with a set of Attribute kv pairs. This diff plumbs the Attribute to ClientTransportFactory, so that we can easily add proxy related attributes in the near future. No behavior is changed here.

Tests:
Added a check to informationPropagatedToNewStreamAndCallCredentials to ensure the Attribute set is properly propagated.

### DIFF
--- a/core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
@@ -32,6 +32,7 @@
 package io.grpc.inprocess;
 
 import com.google.common.base.Preconditions;
+import io.grpc.Attributes;
 import io.grpc.ExperimentalApi;
 import io.grpc.Internal;
 import io.grpc.internal.AbstractManagedChannelImplBuilder;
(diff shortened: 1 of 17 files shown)


zpencer:
retest this please

zpencer:
retest this please

zpencer:
EquivalentAddressGroup already has Attributes that can be used.

zpencer:
My mistake, this diff is already using the correct Attributes (the one inside EquivalentAddressGroup which is the one we want, rather than NameResolver.Listener#onAddresses one)

zhangkun83:
To record our offline discussion: proxy selection seems to be NameResolver-neutral and transport-agnostic, thus it might be done entirely in channel impl, based on our current understanding (which may be wrong) of Case 1 and Case 2 from the [proxy spec](https://github.com/grpc/proposal/blob/master/A1-http-connect-proxy-support.md)
1. Channel impl may look up the target name in the proxy selector. If a proxy is returned, it can bypass `NameResolver` and pass the target name as an EAG to `LoadBalancer`. (Case 1).
2. `InternalSubChannel` may look up in proxy selector for the address before creating a transport, and may pass proxy information along with the address to `newClientTransport()` (Case 2).

Issue #3032, opened 2017-05-24
OkHttp is incompatible with newer C-core binaries


ejona86:
As reported by grpc/grpc#11258. This is caused by grpc/proposal#19 and [a bug](https://github.com/grpc/grpc-java/blob/v1.3.0/okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/framed/Http2.java#L304) in the OkHttp transport that doesn't ignore unknown settings frames (as required by the [HTTP/2 spec](http://httpwg.org/specs/rfc7540.html#rfc.section.6.5.2)).

It's unclear how this wasn't caught in the integration tests.

makdharma:
@ericgribkoff to test in the Android interop he's working on.

ejona86:
We don't run okhttp interop externally (full cross-product testing), but we do internally. We think the breaking change was delayed being brought internally for long enough that it was released without the test having failed yet.

ericgribkoff:
Created https://github.com/grpc/grpc-java/pull/3036 to address this, although I'm not sure what restrictions we have with modifying the third_party okhttp code. The bug was noticed on OkHttp3 (https://github.com/square/okhttp/pull/2299), although I believe their fix still fails for the experimental settings ids used by C++.

> We don't run okhttp interop externally (full cross-product testing), but we do internally. We think the breaking change was delayed being brought internally for long enough that it was released without the test having failed yet.

This appears to have been the case. 

I did some work to add a gRPC Java OkHttp client to the grpc/grpc interop test suite, but encountered some issues with getting our test certificates working with OkHttp. I will look further into this tomorrow, as it seems like we definitely should be testing OkHttp clients as part of our external interops, to (help) avoid issues like this.

ejona86:
This was fixed in #3036

PR #3036, opened 2017-05-25
okhttp: ignore unknown HTTP/2 settings

ericgribkoff:
Intended to address https://github.com/grpc/grpc-java/issues/3032.

This was addressed in OkHttp3 by https://github.com/square/okhttp/pull/2299, but I'm fairly certain their fix still contains a bug - they parse the settings id as a Java short, so anything in the experimental range (such as gRPC C++ servers are using) is a negative number, and the linked pull request's use of `break` instead of `continue` results in an index out of bounds exception in `Settings#set`. I'll follow up on that with the OkHttp repo separately. 

@ejona86 I'm not sure of the status of the code in okhttp/third_party - since this is "shaded" from OkHttp2, can we patch it, or do we need to move the modifications somehow out of the third_party/ subdirectory? As-is, I'm also not sure where to put tests for this, but just wanted to get the process started.

### DIFF
--- a/okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/framed/Http2.java
+++ b/okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/framed/Http2.java
@@ -301,7 +301,8 @@ public final class Http2 implements Variant {
           case 6: // SETTINGS_MAX_HEADER_LIST_SIZE
             break; // Advisory only, so ignored.
           default:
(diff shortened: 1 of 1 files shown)


carl-mastrangelo:
You might consider backporting this to 1.4

ejona86:
Backported to v1.4.x via #3043
Backported to v1.3.x via #3157

PR #3037, opened 2017-05-25
interop-testing: add alpnagent to support okhttp in test client

ericgribkoff:
In response to https://github.com/grpc/grpc/issues/11258 / https://github.com/grpc/grpc-java/issues/3032, we should run our full suite of interop tests using OkHttp. As best I can tell, OkHttp cannot use ALPN unless it has the ALPN agent, which was [removed](https://github.com/grpc/grpc-java/commit/9466eb5014bd4ed2fd0af2567c605ae79a27e86b#diff-dbbdfba18c8958c51b031ef112a6fa7c) a while ago from the build rule, back when we were still using the boot classpath. This adds it back in, so that the `--use_okhttp=true` option in the test client is able to use TLS.

### DIFF
--- a/interop-testing/build.gradle
+++ b/interop-testing/build.gradle
@@ -29,9 +29,16 @@ test {
 task test_client(type: CreateStartScripts) {
     mainClassName = "io.grpc.testing.integration.TestServiceClient"
     applicationName = "test-client"
-    defaultJvmOpts = ["-javaagent:" + configurations.alpnagent.asPath]
+    defaultJvmOpts = ["-javaagent:JAVAAGENT_APP_HOME" + configurations.alpnagent.singleFile.name]
     outputDir = new File(project.buildDir, 'tmp')
     classpath = jar.outputs.files + configurations.runtime
+    dependencies {
+        runtime configurations.alpnagent
+    }
+    doLast {
+        unixScript.text = unixScript.text.replace('JAVAAGENT_APP_HOME', '\$APP_HOME/lib/')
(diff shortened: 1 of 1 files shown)


ejona86:
This does not work if you move the install files to a different machine or clear the gradle cache. It has a hard-coded path:
```
$ grep javaagent interop-testing/build/install/grpc-interop-testing/bin/test-client
DEFAULT_JVM_OPTS='"-javaagent:/usr/local/google/home/ejona/.gradle/caches/modules-2/files-2.1/org.mortbay.jetty.alpn/jetty-alpn-agent/2.0.6/ecca008cfd140e088590c0f2e0f223523cdb18d2/jetty-alpn-agent-2.0.6.jar"'
```

ericgribkoff:
AFAIK, the only alternative to using Gradle's `defaultJvmOpts` with the hard-coded path in the generated shell script is to require the user to specify the path to Jetty ALPN on the command-line, e.g., `env JAVA_OPTS=-javaagent:pathTo/jetty-alpn-agent-2.0.6.jar build/install/grpc-interop-testing/bin/test-client --server_host_override=foo.test.google.fr --use_test_ca=true --use_okhttp=true`. Is this alternative the preferred approach? If so, I should at least add some message to the test client output explaining that this is required. Right now the combination of tls and okhttp never works.

ericgribkoff:
@ejona86 I updated `build.gradle` to avoid having hard-coded paths in the startup scripts. This requires a little bit of hackery to get the `$APP_PATH` and `%APP_PATH%` variables working correctly on *nix and Windows startup scripts.

PR #3042, opened 2017-05-30
Remove thrift-related code from grpc/grpc-java

vjpai:
This is a suggestion; grpc-ecosystem/grift is now the new home of the experimental grift (thrift marshaling over gRPC) code and so it doesn't make sense to keep this code in the mainline grpc repos. I have created grpc-ecosystem/grift#1 to move the C++ code and grpc-ecosystem/grift#7 to move over the Java code, so I believe the corresponding thing in both cases is to remove the thrift code from the grpc org repo. Thrift was already removed from the C repo a while ago in grpc/grpc#10911. Hopefully this PR doesn't break the world!


carl-mastrangelo:
See #3477 This will probably need to be rebased, which might as well be a new PR.

PR #3047, opened 2017-05-31
all: update to Apache2 licence

carl-mastrangelo:
@nicolasnoble @zinuga @jayantkolhe @hsaliak


nicolasnoble:
The Copyright lines have to change to "gRPC Authors", and an AUTHORS file needs to be added that contains Google Inc.

Issue #2488, opened 2016-12-06
GrpcServerRule is Experimental


joeyb:
This is a tracking issue that will be closed once `GrpcServerRule` exits experimental status.

dapengzhang0:
@ejona86 I'm taking it now

dapengzhang0:
oops, I was referencing to the wrong issue.

brunobuss:
What are the requirements for this to graduate from experimental?

dapengzhang0:
@brunobuss `GrpcServerRule` is superseded with `GrpcCleanupRule` which is more flexible and preferred.  The plan is  that `GrpcServerRule` will be marked `@Deprecated` while not be removed, and that `GrpcCleanupRule` will graduate from experimental.

PR #3053, opened 2017-06-01
examples: migrate unittest examples to GrpcServerRule

### LINKED ISSUE #2490: Migrate example unit tests to GrpcServerRule
GrpcServerRule from #2456 simplifies the boilerplate necessary to unit test gRPC clients and services. We should update the example unit tests to make use of it (as appropriate).

dapengzhang0:
migrated simple tests using `GrpcServerRule`.
Kept the low level in-process channel setup and tear down code for the RouteGuide example to show how users can use in-process directly to set more custom channel builder options when needed.

resolves #2490

### DIFF
--- a/examples/src/test/java/io/grpc/examples/helloworld/HelloWorldServerTest.java
+++ b/examples/src/test/java/io/grpc/examples/helloworld/HelloWorldServerTest.java
@@ -20,7 +20,6 @@ import static org.junit.Assert.assertEquals;
 
 import io.grpc.examples.helloworld.HelloWorldServer.GreeterImpl;
 import io.grpc.testing.GrpcServerRule;
-import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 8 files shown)


ejona86:
How does this resolve #2488?

examples/src/test/java/io/grpc/examples/helloworld/HelloWorldServerTest.java
```diff
@@ -70,7 +55,9 @@ public void tearDown() {
    */
   @Test
   public void greeterImpl_replyMessage() throws Exception {
-    GreeterGrpc.GreeterBlockingStub blockingStub = GreeterGrpc.newBlockingStub(inProcessChannel);
+    grpcServerRule.getServiceRegistry().addService(new GreeterImpl());
```
ejona86: Isn't the service already registered?

dapengzhang0:
Referenced to the wrong issue. Should be #2490.

PR #3062, opened 2017-06-05
core: OverrideAuthorityNameResolverFactory should forward refresh()

zpencer:
The current implementation has a bug where certain methods are not forwarded to the delegate.


zpencer:
Fixing license problem, then will squash merge. This is v1.4.x and needs the old styled license.

PR #3063, opened 2017-06-05
testing: move out inner class tests in GrpcServerRuleTest

### LINKED ISSUE #2808: Inner class unit tests in GrpcServerRuleTest look so weird

dapengzhang0:
Resolves #2808


ejona86:
What did you think about the `scanForTestClasses = false` suggestion?

dapengzhang0:
@ejona86 I'm not sure I understand how it works. If that is marked false, it will error out if some inner class tests are added, or just ignore them?

PR #3065, opened 2017-06-06
core: move NameResolverProvider to internal

carl-mastrangelo:
This is part one to enable grpclb to do dns queries


ejona86:
This doesn't look right. This is similar to the channel and server providers, and those we can't move to internal.

carl-mastrangelo:
@ejona86 the difference is that `NameResolverProvider` is used by AbstractManagedChannelBuilder impl, and not on the pubic api.  The other providers are used directly by the public builders.

ejona86:
Users will need to reference this class name for their own name resolver implementation. From the user's perspective I see its use in `AbstractManagedChannelBuilder` as an implementation detail.

So I guess the bigger issue is this should be `@ExperimentalApi` and not `@Internal`?

carl-mastrangelo:
NameResolver.Factory is the API to setting a name resolver, not this class.

As mentioned in the commit message, this needs to be internal because I need to expose some of its methods to be public for implementing srv records.  This class was marked `@Internal` which I need in order to make breaking API changes.  

If you want to suggest this class to be experimental, file an issue?

ejona86:
Users can provide their own NameResolvers though, and that is through this class. I saw the PR description, but it didn't really say much.

It seems this needs to be reverted. I'll send out a PR.

PR #3066, opened 2017-06-06
core: hide access to Status code and message keys

carl-mastrangelo:


core/src/main/java/io/grpc/InternalStatus.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package io.grpc;
+
+/**
+ * Accesses internal data.  Do not use this.
+ */
+@Internal
+public final class InternalStatus {
```
ejona86: I'd honestly be happy with these in just the `io.grpc.Internal` class. I'm not sure we need a ton of different Internal classes.

core/src/main/java/io/grpc/Status.java
```diff
@@ -377,8 +376,7 @@ public static Status fromCode(Code code) {
   /**
    * Key to bind status message to trailing metadata.
    */
-  @Internal
-  public static final Metadata.Key<String> MESSAGE_KEY =
+  static final Metadata.Key<String> MESSAGE_KEY =
```
ejona86: It looks like this and `StatusMessageMarshaller` could be moved to `io.grpc.internal`.

carl-mastrangelo: I tried it that way, but it ended up being ugly.  There are parts of Status.Code which need to be exposed too.  It pulled too much of Status and Code into internal, rather than just the internal bits.

ejona86: That sounds true for CODE_KEY, but not this description marshalling.

carl-mastrangelo: It is, but then its weird to not have them together, which is what Status does.

PR #3068, opened 2017-06-06
stub: detect invalid states on server side (eg zero responses for unary)

### LINKED ISSUE #2243: Detect zero-message response to unary request on server-side

### LINKED ISSUE #3059: ServerStreamListenerImpl closing ServerStream in unsafe way
This is not an actual problem because of the way stubs are implemented today, but the code can potentially allow the below to happen:
ServerStreamListenerImpl is executing in the network thread (in the case of the direct executor) or in a application thread (via JumpToApplicationThreadServerStreamListener in the case of an application thread pool). ServerStreamListenerImpl.messageRead can detect an inconsistent gRPC protocol state, and close the ServerStream. This is a problem because the user application responsible for sending data (a different application thread from the JumpToApplicationThreadServerStreamListener) can be in the middle of working with the stream.

A possible fix is to treat this particular invalid gRPC protocol state as a special case, because for clientSendsOne methods, we can require the client to halfClose before we trigger the application code. This allows us to detect the problem in the stub before the application gets triggered, and we avoid any synchronization concerns. We should investigate further into other inconsistent states we want to detect, and make sure this strategy of moving it to the stub also works.

zpencer:
The current check in ServerCallImpl is theoretically unsafe (#3059). Move that check into the stub, and expand the unit tests to cover other interesting edge cases on the server side:
- client sends one, but zero requests received at onHalfClose
- client sends one, but > 1 requests received at onHalfClose
- server sends one, but zero responses sent at onComplete
- server sends one, but > 1 responses sent via onNext

fixes #2243 
fixes #3059

### DIFF
--- a/core/src/main/java/io/grpc/internal/ServerCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerCallImpl.java
@@ -60,7 +60,6 @@ final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
   private boolean closeCalled;
   private Compressor compressor;
   private boolean messageSent;
-  private boolean internalClosed;
 
   ServerCallImpl(ServerStream stream, MethodDescriptor<ReqT, RespT> method,
       Metadata inboundHeaders, Context.CancellableContext context,
@@ -75,9 +74,6 @@ final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
 
   @Override
   public void request(int numMessages) {
-    if (internalClosed) {
-      return;
-    }
     stream.request(numMessages);
   }
 
@@ -85,11 +81,6 @@ final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
   public void sendHeaders(Metadata headers) {
     checkState(!sendHeadersCalled, "sendHeaders has already been called");
     checkState(!closeCalled, "call is closed");
-    // Check internalClosed for completeness, though today there's no way to trigger an internal
-    // close without already having sent headers.
-    if (internalClosed) {
-      return;
-    }
 
     headers.discardAll(MESSAGE_ENCODING_KEY);
     if (compressor == null) {
@@ -130,9 +121,6 @@ final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
   public void sendMessage(RespT message) {
     checkState(sendHeadersCalled, "sendHeaders has not been called");
     checkState(!closeCalled, "call is closed");
-    if (internalClosed) {
-      return;
-    }
 
     if (method.getType().serverSendsOneMessage() && messageSent) {
       internalClose(Status.INTERNAL.withDescription(TOO_MANY_RESPONSES));
@@ -155,10 +143,6 @@ final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
 
   @Override
   public void setMessageCompression(boolean enable) {
-    if (internalClosed) {
-      return;
-    }
-
     stream.setMessageCompression(enable);
   }
 
@@ -173,9 +157,6 @@ final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
 
   @Override
   public boolean isReady() {
-    if (internalClosed) {
-      return false;
-    }
     return stream.isReady();
   }
 
@@ -183,9 +164,6 @@ final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
   public void close(Status status, Metadata trailers) {
     checkState(!closeCalled, "call already closed");
     closeCalled = true;
-    if (internalClosed) {
-      return;
-    }
 
     if (status.isOk() && method.getType().serverSendsOneMessage() && !messageSent) {
       internalClose(Status.INTERNAL.withDescription(MISSING_RESPONSE));
@@ -225,7 +203,6 @@ final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
    * on.
    */
   private void internalClose(Status internalError) {
-    internalClosed = true;
     stream.close(internalError, new Metadata());
   }
 
--- a/core/src/main/java/io/grpc/internal/ServerCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerCallImpl.java
@@ -39,11 +39,8 @@ import io.grpc.Status;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.List;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
-  private static final Logger log = Logger.getLogger(ServerCallImpl.class.getName());
 
   @VisibleForTesting
   static String TOO_MANY_RESPONSES = "Too many responses";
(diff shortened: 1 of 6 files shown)


zpencer:
Synced offline, will move as much validation as possible into ServerCallImpl as possible, and only put things in stubs as a last resort. The reasoning is that the ServerCallImpl is lower level.

zpencer:
retest this please

stub/src/main/java/io/grpc/stub/ServerCalls.java
```diff
@@ -138,8 +157,11 @@ public void onHalfClose() {
                 onReady();
               }
             } else {
-              call.close(Status.INTERNAL.withDescription("Half-closed without a request"),
-                  new Metadata());
+              Status internalError = Status.INTERNAL.withDescription(MISSING_REQUEST);
+              StatusException exception = internalError.asException();
+              log.log(Level.FINE, internalError.getDescription(), exception);
```
ejona86: It seems fine to me not to log this, since the client will be told pretty clearly what went wrong and the backtrace in the exception probably isn't too useful.

However, if you are going to log it, don't create the exception unless logging is on. So surround the two lines here with  a `logger.isLoggable(FINE)` check. Ditto elsewhere.

core/src/main/java/io/grpc/internal/ServerCallImpl.java
```diff
@@ -55,6 +63,8 @@
   private boolean sendHeadersCalled;
   private boolean closeCalled;
   private Compressor compressor;
+  private boolean messageSent;
+  private boolean internalClosed;
```
ejona86: Did you discover `internalClosed` was necessary? I think it should be unnecessary.

core/src/main/java/io/grpc/internal/ServerCallImpl.java
```diff
@@ -156,14 +188,25 @@ public void setCompression(String compressorName) {
...
     closeCalled = true;
-    stream.close(status, trailers);
+    if (internalClosed) {
+      return;
+    }
+
+    if (status.isOk() && method.getType().serverSendsOneMessage() && !messageSent) {
+      internalClose(Status.INTERNAL.withDescription(MISSING_RESPONSE));
+    } else {
+      stream.close(status, trailers);
```
ejona86: Prefer making error cases return to avoid indentation in the common case. Sometimes it doesn't matter, but here it seems like it would improve the readability.

core/src/main/java/io/grpc/internal/ServerCallImpl.java
```diff
@@ -191,6 +234,17 @@ public String getAuthority() {
   }
 
   /**
+   * Close the {@link ServerStream} because an internal error occurred. Allow the application to
+   * run until completion, but silently ignore interactions with the {@link ServerStream} from now
+   * on.
+   */
+  private void internalClose(Status internalError) {
+    internalClosed = true;
+    log.log(Level.FINE, internalError.getDescription(), internalError.asException());
```
ejona86: I don't think we need to really be logging this. But even if we are, it should be protected with an if isloggable check to avoid creating the exception.

core/src/main/java/io/grpc/internal/ServerCallImpl.java
zpencer: The `ServerStream` API states that close() will close the stream for reading and writing, but did not specify what happens if we keep interacting with it after calling close(). `InProcessTransport` and `AbstractStream`  happen to both ignore all interactions after `close`. I didn't want to rely on what seems like behavior, and opted to never touch the stream again after calling `close()`. If you're ok with adding that behavior to the API, this flag can be removed.

stub/src/main/java/io/grpc/stub/ServerCalls.java
zpencer: removed

core/src/main/java/io/grpc/internal/ServerCallImpl.java
zpencer: removed

core/src/main/java/io/grpc/internal/ServerCallImpl.java
zpencer: restructured

core/src/main/java/io/grpc/internal/ServerCallImpl.java
ejona86: I'm pretty sure we're [relying on similar behavior client-side](https://github.com/grpc/grpc-java/blob/v1.4.0/core/src/main/java/io/grpc/internal/ClientCallImpl.java#L395). I would believe it may be slightly broken, but I'd be in favor of trying to make it a guarantee of the stream.

core/src/test/java/io/grpc/internal/ServerCallImplTest.java
```diff
@@ -159,6 +168,114 @@ public void sendMessage_closesOnFailure() {
...
+    verify(stream, times(1)).writeMessage(any(InputStream.class));
+    verify(stream, times(1)).close(any(Status.class), any(Metadata.class));
+
+    // App runs to completion but everything is ignored
+    serverCall.sendMessage(1L);
+    serverCall.close(Status.OK, new Metadata());
+    try {
+      serverCall.close(Status.OK, new Metadata());
+      fail("calling a second time should still cause an error");
+    } catch (IllegalStateException expected) {
```
ejona86: Instead of the try-catch, use `thrown.expect(IllegalStateException.class)` just before calling close the second time?

stub/src/main/java/io/grpc/stub/ServerCalls.java
```diff
@@ -122,24 +129,35 @@ private ServerCalls() {
           ReqT request;
           @Override
           public void onMessage(ReqT request) {
+            if (this.request != null) {
+              // Safe to close the call, because the application has not yet been invoked
```
ejona86: invoke() can still happen after this point. It seems you may need a boolean to ignore any future halfClose.

stub/src/test/java/io/grpc/stub/ServerCallsTest.java
```diff
@@ -374,15 +438,19 @@ public Integer parse(InputStream stream) {
   }
 
   private static class ServerCallRecorder extends ServerCall<Integer, Integer> {
-    private List<Integer> requestCalls = new ArrayList<Integer>();
+    private final MethodDescriptor<Integer, Integer> methodDescriptor;
```
ejona86: This looks unused.

stub/src/main/java/io/grpc/stub/ServerCalls.java
zpencer: Added the onHalfClose to the test and verified that this now passes the test with the boolean. Also realized that the server streaming tests should really be using the `ServerCalls.asyncServerStreamingCall` stub, even though it's equivalent to `asyncUnaryRequestCall`. I broke the tests into independent tests to be more precise here.

core/src/test/java/io/grpc/internal/ServerCallImplTest.java
zpencer: done

stub/src/test/java/io/grpc/stub/ServerCallsTest.java
zpencer: This field is used in ServerCalls.java:
```
Preconditions.checkArgument(
            call.getMethodDescriptor().getType().clientSendsOneMessage(),
```

This field was previously a "private" and was always null, I turned it into a "private final" and moved it to the rest of the "private final"s.

ejona86: Ah, it previously being always-null is what caused me to miss the usage.

PR #3069, opened 2017-06-06
V1.4.x

carl-mastrangelo:
Two commits, as always.  Readme follows after release on mvn


ejona86:
Readme needs to be updated on the release branch, because people look at the actual tag itself.

"start v1.4.0" is strange, since it was "started" earlier. How about just "Version 1.4.0"?

carl-mastrangelo:
@ejona86 done.

PR #3071, opened 2017-06-07
context: add Context.Builder.

zhangkun83:
Context.withValues() are useful only if the set of keys are known
upfront.  There are cases where the keys are only known at runtime.  A
builder pattern can prevent excessive creation and chaining of Contexts
in such cases.

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -348,6 +348,14 @@ public class Context {
   }
 
   /**
+   * Create a builder which builds contexts that cascade cancellation from their parent, which is
+   * this context.
+   */
+  public Builder newBuilder() {
+    return new Builder(this);
+  }
+
+  /**
    * Create a new context which propagates the values of this context but does not cascade its
    * cancellation.
    */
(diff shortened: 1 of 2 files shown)


context/src/main/java/io/grpc/Context.java
```diff
@@ -860,6 +868,38 @@ public String toString() {
...
+
+    /**
+     * Adds a value.
+     *
+     * <p>Like {@code withValues()} on the {@link Context}, the builder doesn't check duplicate
+     * keys. The outcome of duplicate keys given the same builder is unspecified.
+     *
+     * @return this builder
+     */
+    public <K,V> Builder withValue(Key<V> k, V v) {
```
carl-mastrangelo: `with` is the wrong word here.  Builders use `set` and `add`.  `with` implies that you get a forked copy.

zhangkun83: Sure. Renamed to `add`.

ejona86:
I think part of the point of the current API is to discourage lots of keys (i.e., don't use 5 keys when 1 will do). The builder concerns me because either 1) the calling code shouldn't be using lots of keys or 2) it seems like we are needing to make some optimizations.

carl-mastrangelo:
@ejona86 people already use the 4-key overload already, and it is a high source of slow down according to internal global  profiling.

zhangkun83:
> I think part of the point of the current API is to discourage lots of keys (i.e., don't use 5 keys when 1 will do). The builder concerns me because either 1) the calling code shouldn't be using lots of keys or 2) it seems like we are needing to make some optimizations.

The motivation is not to have lots of keys, but dynamic set of keys. I have a case where a Context is created with 4 keys: A, B, C1 or C2, D1 or D2. The `or`s are determined by two runtime conditions independently. Without the builder, I would have to either have repetitive code in 4 branches, or creating 2 more Contexts.

zhangkun83:
Can the performance problem be solved by using a hashmap?

zhangkun83:
I will try some refactor to eliminate my use case.

PR #3072, opened 2017-06-07
Revert "core: move NameResolverProvider to internal"

ejona86:
This reverts commit 9c6ea274fe93955236c9ef4c68d02b72c9f58ccc.

The class name is important for the service loaders.


carl-mastrangelo:
Name resolution falls in the same bucket as transport implementation as far as support goes: internal.  I don't think I can approve this request until it has gone under more thorough API review.

ejona86:
No. It does not. This is something users are providing.

PR #3073, opened 2017-06-07
Allow gRPC LB to override and delegate DNS name resolution

carl-mastrangelo:
This is part 2 of enabling balancer address lookup for DNS.

Since SRV queries add extra latency, and they are only useful to gRPBLB, only enable them when gRPC LB is present.  

There may be another env var to enable this behavior, but that will need ot be determined later.  for now, small PRs.

FYI: @ejona86 this change depends on the breaking API changes in NameResolverProvider.


grpclb/src/main/java/io/grpc/grpclb/SecretGrpclbDnsNameResolverProvider.java
```diff
@@ -0,0 +1,79 @@
...
+   * in the process.
+   */
+  public static final class GrpclbDnsNameResolverProvider extends NameResolverProvider {
+
+    /**
+     * Invoked reflectively by the service loader.
+     */
+    public GrpclbDnsNameResolverProvider() {}
+
+    private final DnsNameResolverProvider delegate =
```
ejona86: As an alternative to delegating, you may consider just extending DnsNameResolverProvider and call super(true) in a default constructor here. Yes, that means the other provider would no longer be a final class. Your choice on whether you want to make the methods public on DnsNameResolverProvider or extend. It doesn't matter to me.

carl-mastrangelo:
I have a yet to be sent PR that does send SRV.

grpclb/src/main/java/io/grpc/grpclb/SecretGrpclbDnsNameResolverProvider.java
carl-mastrangelo: Thinking about the approach here, I think delegating is a better long term approach.  When we add the Netty name resolver, I don't want to have to extend it, which may not even be possible.  (this also gives a reason why the NRP should have public methods, but that can be changed later).

grpclb/src/main/java/io/grpc/grpclb/SecretGrpclbDnsNameResolverProvider.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+package io.grpc.grpclb;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import io.grpc.Attributes;
+import io.grpc.NameResolver;
+import io.grpc.internal.DnsNameResolverProvider;
+import io.grpc.internal.NameResolverProvider;
```
zhangkun83: Please sync with master HEAD and remove `internal` from the package name.

carl-mastrangelo: Done

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -82,9 +84,13 @@
...
 
-  DnsNameResolver(@Nullable String nsAuthority, String name, Attributes params,
+  DnsNameResolver(
+      @Nullable String nsAuthority,
+      String name,
+      Attributes params,
       Resource<ScheduledExecutorService> timerServiceResource,
-      Resource<ExecutorService> executorResource) {
+      Resource<ExecutorService> executorResource,
+      boolean balancerLookup) {
```
zhangkun83: I thought we were going to have a GrpclbDnsNameResolver that would look up the balancer, and this DnsNameResolver should not know about balancers.

carl-mastrangelo: There will be only one resolver that does DNS, and it will optionally do SRV.  There should not be two.

zhangkun83: I agree we shouldn't duplicate implementations. But doing SRV and doing balancer is different. Are we saying that DnsNameResolver knows about GRPCLB balancers?

carl-mastrangelo: No, but it will know how to do SRV record lookups.  SRV is not grpclb specific.  Its more DNS than LB so it belongs there.

carl-mastrangelo:
@zhangkun83  ping

core/src/main/java/io/grpc/internal/DnsNameResolver.java
zhangkun83: Although SRV is not grpclb specific, that SRV is used for balancer lookup (as this boolean argument implies) is grpclb specific. Consider changing it to "srvLookup" and a generic SRV plumbing path (e.g., as attributes) that don't have any "balancer" indication.

PR #3079, opened 2017-06-08
android-interop-testing: instrumented interop tests

ericgribkoff:
This adds an instrumented test which runs the Android interops against our deployed test server. This will be used in Firebase test lab to get tests running on Android emulators and/or devices.

### DIFF
--- a/android-interop-testing/app/src/androidTest/java/io/grpc/android/integrationtest/InteropTesterTest.java
+++ b/android-interop-testing/app/src/androidTest/java/io/grpc/android/integrationtest/InteropTesterTest.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2015, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.android.integrationtest;
+
+import static junit.framework.Assert.assertTrue;
+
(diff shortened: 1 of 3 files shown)


ejona86:
@ericgribkoff, this breaks building of the app for me (proguard + junit).

ericgribkoff:
@ejona86 Should be fixed now, added the dontwarn rule for junit.

android-interop-testing/app/src/androidTest/java/io/grpc/android/integrationtest/InteropTesterTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(AndroidJUnit4.class)
+public class InteropTesterTest {
+  private final int TIMEOUT_SECONDS = 120;
+
+  @Test
+  public void interopTests() throws Exception {
+    final CountDownLatch latch = new CountDownLatch(1);
```
ejona86: Use a SettableFuture? That way you can also report the actual failure message, which looks like it is thrown away right now.

android-interop-testing/app/src/androidTest/java/io/grpc/android/integrationtest/InteropTesterTest.java
```diff
@@ -0,0 +1,57 @@
...
+    final AtomicBoolean passed = new AtomicBoolean(false);
+    new InteropTester(
+            "all",
```
dapengzhang0: nit: I believe the indentation here and below is not google-java-stylish

android-interop-testing/app/src/androidTest/java/io/grpc/android/integrationtest/InteropTesterTest.java
ericgribkoff: Done, thanks for the suggestion

android-interop-testing/app/src/androidTest/java/io/grpc/android/integrationtest/InteropTesterTest.java
ericgribkoff: I agree that it looks a little strange, but this is the result of running google-java-format (double- and triple-checked this :)) so will leave it as-is.

PR #3086, opened 2017-06-12
Introduce MethodDescriptorSupplier

lukaszx0:
Second iteration on providing an easier way to access service methods which I took stab at in #2904.

It introduces new `MethodDescriptorSupplier` and `MethodDescriptor#setSchemaDescriptor`.

cc @ejona86

### DIFF
--- a/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
+++ b/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
@@ -270,8 +270,10 @@ public final class BenchmarkServiceGrpc {
     }
   }
 
-  private static class BenchmarkServiceFileDescriptorSupplier
+  private static abstract class BenchmarkServiceBaseDescriptorSupplier
       implements io.grpc.protobuf.ProtoFileDescriptorSupplier, io.grpc.protobuf.ProtoServiceDescriptorSupplier {
+    BenchmarkServiceBaseDescriptorSupplier() {}
+
     @java.lang.Override
     public com.google.protobuf.Descriptors.FileDescriptor getFileDescriptor() {
       return io.grpc.benchmarks.proto.Services.getDescriptor();
@@ -283,12 +285,17 @@ public final class BenchmarkServiceGrpc {
     }
   }
 
+  private static final class BenchmarkServiceFileDescriptorSupplier
+      extends BenchmarkServiceBaseDescriptorSupplier {
+    BenchmarkServiceFileDescriptorSupplier() {}
+  }
+
   private static final class BenchmarkServiceMethodDescriptorSupplier
-      extends BenchmarkServiceFileDescriptorSupplier
-        implements io.grpc.protobuf.ProtoMethodDescriptorSupplier {
+      extends BenchmarkServiceBaseDescriptorSupplier
+      implements io.grpc.protobuf.ProtoMethodDescriptorSupplier {
     private final String methodName;
 
-    private BenchmarkServiceMethodDescriptorSupplier(String methodName) {
+    BenchmarkServiceMethodDescriptorSupplier(String methodName) {
       this.methodName = methodName;
     }
 
--- a/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
+++ b/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
@@ -270,7 +270,8 @@ public final class BenchmarkServiceGrpc {
     }
   }
 
-  private static class BenchmarkServiceFileDescriptorSupplier implements io.grpc.protobuf.ProtoFileDescriptorSupplier, io.grpc.protobuf.ProtoServiceDescriptorSupplier {
+  private static class BenchmarkServiceFileDescriptorSupplier
+      implements io.grpc.protobuf.ProtoFileDescriptorSupplier, io.grpc.protobuf.ProtoServiceDescriptorSupplier {
     @java.lang.Override
     public com.google.protobuf.Descriptors.FileDescriptor getFileDescriptor() {
       return io.grpc.benchmarks.proto.Services.getDescriptor();
@@ -282,10 +283,12 @@ public final class BenchmarkServiceGrpc {
     }
   }
 
-  private static final class BenchmarkServiceMethodDescriptorSupplier extends BenchmarkServiceFileDescriptorSupplier implements io.grpc.protobuf.ProtoMethodDescriptorSupplier {
+  private static final class BenchmarkServiceMethodDescriptorSupplier
+      extends BenchmarkServiceFileDescriptorSupplier
+        implements io.grpc.protobuf.ProtoMethodDescriptorSupplier {
     private final String methodName;
 
-    public BenchmarkServiceMethodDescriptorSupplier(String methodName) {
+    private BenchmarkServiceMethodDescriptorSupplier(String methodName) {
       this.methodName = methodName;
     }
 
--- a/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
+++ b/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
@@ -38,6 +38,7 @@ public final class BenchmarkServiceGrpc {
               io.grpc.benchmarks.proto.Messages.SimpleRequest.getDefaultInstance()))
           .setResponseMarshaller(io.grpc.protobuf.ProtoUtils.marshaller(
               io.grpc.benchmarks.proto.Messages.SimpleResponse.getDefaultInstance()))
+          .setSchemaDescriptor(new BenchmarkServiceMethodDescriptorSupplier("UnaryCall"))
           .build();
   @io.grpc.ExperimentalApi("https://github.com/grpc/grpc-java/issues/1901")
   public static final io.grpc.MethodDescriptor<io.grpc.benchmarks.proto.Messages.SimpleRequest,
@@ -50,6 +51,7 @@ public final class BenchmarkServiceGrpc {
               io.grpc.benchmarks.proto.Messages.SimpleRequest.getDefaultInstance()))
           .setResponseMarshaller(io.grpc.protobuf.ProtoUtils.marshaller(
               io.grpc.benchmarks.proto.Messages.SimpleResponse.getDefaultInstance()))
+          .setSchemaDescriptor(new BenchmarkServiceMethodDescriptorSupplier("StreamingCall"))
           .build();
 
   /**
@@ -268,10 +270,28 @@ public final class BenchmarkServiceGrpc {
     }
   }
 
-  private static final class BenchmarkServiceDescriptorSupplier implements io.grpc.protobuf.ProtoFileDescriptorSupplier {
+  private static class BenchmarkServiceFileDescriptorSupplier implements io.grpc.protobuf.ProtoFileDescriptorSupplier, io.grpc.protobuf.ProtoServiceDescriptorSupplier {
     @java.lang.Override
     public com.google.protobuf.Descriptors.FileDescriptor getFileDescriptor() {
       return io.grpc.benchmarks.proto.Services.getDescriptor();
+    }
+
+    @java.lang.Override
+    public com.google.protobuf.Descriptors.ServiceDescriptor getServiceDescriptor() {
+      return getFileDescriptor().findServiceByName("BenchmarkService");
+    }
+  }
+
+  private static final class BenchmarkServiceMethodDescriptorSupplier extends BenchmarkServiceFileDescriptorSupplier implements io.grpc.protobuf.ProtoMethodDescriptorSupplier {
+    private final String methodName;
+
+    public BenchmarkServiceMethodDescriptorSupplier(String methodName) {
+      this.methodName = methodName;
+    }
+
+    @java.lang.Override
+    public com.google.protobuf.Descriptors.MethodDescriptor getMethodDescriptor() {
+      return getServiceDescriptor().findMethodByName(methodName);
     }
   }
 
@@ -284,7 +304,7 @@ public final class BenchmarkServiceGrpc {
         result = serviceDescriptor;
         if (result == null) {
           serviceDescriptor = result = io.grpc.ServiceDescriptor.newBuilder(SERVICE_NAME)
-              .setSchemaDescriptor(new BenchmarkServiceDescriptorSupplier())
+              .setSchemaDescriptor(new BenchmarkServiceFileDescriptorSupplier())
               .addMethod(METHOD_UNARY_CALL)
               .addMethod(METHOD_STREAMING_CALL)
               .build();
--- a/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/WorkerServiceGrpc.java
+++ b/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/WorkerServiceGrpc.java
@@ -396,8 +396,10 @@ public final class WorkerServiceGrpc {
     }
   }
 
-  private static class WorkerServiceFileDescriptorSupplier
+  private static abstract class WorkerServiceBaseDescriptorSupplier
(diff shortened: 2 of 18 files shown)


lukaszx0:
(Splitted into two commit as per @ejona86's suggestion)

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
```diff
@@ -38,6 +38,7 @@ private BenchmarkServiceGrpc() {}
               io.grpc.benchmarks.proto.Messages.SimpleRequest.getDefaultInstance()))
           .setResponseMarshaller(io.grpc.protobuf.ProtoUtils.marshaller(
               io.grpc.benchmarks.proto.Messages.SimpleResponse.getDefaultInstance()))
+          .setSchemaDescriptor(new BenchmarkServiceFileDescriptorSupplier().getServiceDescriptor().findMethodByName("UnaryCall"))
```
ejona86: This should have just been `.setSchemaDescriptor(new BenchmarkServiceFileDescriptorSupplier())`

protobuf/src/main/java/io/grpc/protobuf/ProtoServiceDescriptorSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package io.grpc.protobuf;
+
+import com.google.protobuf.Descriptors.ServiceDescriptor;
+
+/**
+ * Provides access to the underlying proto service descriptor.
+ */
+public interface ProtoServiceDescriptorSupplier extends ProtoFileDescriptorSupplier {
```
ejona86: I had intended that this would be in the second commit as well.

ejona86:
This looks fine to me, other than issues with how the commits were split. Those can either be resolved by reworking what is in each commit or by squashing.

@zhangkun83, @carl-mastrangelo, @dapengzhang0, It's important to note that the new APIs introduced to MethodDescriptor, ProtoServiceDescriptorSupplier, and ProtoMethodDescriptorSupplier can't be `@ExperimentalApi` since they are used by the codegen. The change to `MethodDescriptor` is a copy of what is already in `ServiceDescriptor`.

CC @ericgribkoff

lukaszx0:
@ejona86 squashing will be easier if that works?

lukaszx0:
@ejona86 rebased and squashed into one commit.

protobuf/src/main/java/io/grpc/protobuf/ProtoMethodDescriptorSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package io.grpc.protobuf;
+
+import com.google.protobuf.Descriptors.MethodDescriptor;
+
+/**
+ * Provides access to the underlying proto service method descriptor.
+ */
+public interface ProtoMethodDescriptorSupplier {
```
ejona86: Extend `ProtoServiceDescriptorSupplier`?

dapengzhang0:
LGTM

compiler/src/java_plugin/cpp/java_generator.cpp
```diff
@@ -906,7 +917,40 @@ static void PrintGetServiceDescriptorMethod(const ServiceDescriptor* service,
...
+
+    p->Outdent();
+    p->Print(*vars, "}\n\n");
+
+    p->Print(
+        *vars,
+        "private static final class $proto_method_descriptor_supplier$ extends $proto_file_descriptor_supplier$ implements $ProtoMethodDescriptorSupplier$ {\n");
+    p->Indent();
+    p->Print(*vars, "private final String methodName;\n\n");
+    p->Print(*vars, "public $proto_method_descriptor_supplier$(String methodName) {\n");
```
carl-mastrangelo: private

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
```diff
@@ -268,11 +270,29 @@ public void invoke(Req request, io.grpc.stub.StreamObserver<Resp> responseObserv
...
       return io.grpc.benchmarks.proto.Services.getDescriptor();
     }
+
+    @java.lang.Override
+    public com.google.protobuf.Descriptors.ServiceDescriptor getServiceDescriptor() {
+      return getFileDescriptor().findServiceByName("BenchmarkService");
+    }
+  }
+
+  private static final class BenchmarkServiceMethodDescriptorSupplier extends BenchmarkServiceFileDescriptorSupplier implements io.grpc.protobuf.ProtoMethodDescriptorSupplier {
```
carl-mastrangelo: Why extend?  What if the service descriptor isn't available?  Services *have-a* set of methods.  Methods fail the `is-a` test.  So too with the supplier.  

Perhaps a method could be added to get the serviceDescriptorSupplier for the method?

core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -42,6 +42,7 @@
   private final String fullMethodName;
   private final Marshaller<ReqT> requestMarshaller;
   private final Marshaller<RespT> responseMarshaller;
+  private final Object schemaDescriptor;
```
carl-mastrangelo: `@Nullable`

core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -312,6 +315,20 @@ public InputStream streamResponse(RespT response) {
...
+   * Returns the schema descriptor for this method. A schema descriptor is an object that is not
+   * used by gRPC core but includes information related to the service method. The type of the
+   * object is specific to the consumer, so both the code setting the schema descriptor and the code
+   * calling {@link #getSchemaDescriptor()} must coordinate.  For example, protobuf generated code
+   * sets this value, in order to be consumed by the server reflection service.  See also:
+   * {@code io.grpc.protobuf.ProtoMethodDescriptorSupplier}.
+   *
+   * @since 1.5.0
+   */
+  public Object getSchemaDescriptor() {
```
carl-mastrangelo: `@Nullable`

protobuf/src/main/java/io/grpc/protobuf/ProtoMethodDescriptorSupplier.java
carl-mastrangelo: `@DoNotMock`

protobuf/src/main/java/io/grpc/protobuf/ProtoMethodDescriptorSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+  /**
+   * Returns method descriptor to the proto service method.
+   */
+  MethodDescriptor getMethodDescriptor();
```
carl-mastrangelo: `@CheckReturnValue`

compiler/src/java_plugin/cpp/java_generator.cpp
```diff
@@ -906,7 +917,40 @@ static void PrintGetServiceDescriptorMethod(const ServiceDescriptor* service,
...
+    p->Print(*vars, "return getFileDescriptor().findServiceByName(\"$service_name$\");\n");
+    p->Outdent();
+    p->Print(*vars, "}\n");
+
+    p->Outdent();
+    p->Print(*vars, "}\n\n");
+
+    p->Print(
+        *vars,
+        "private static final class $proto_method_descriptor_supplier$ extends $proto_file_descriptor_supplier$ implements $ProtoMethodDescriptorSupplier$ {\n");
```
carl-mastrangelo: So far grpc generated code is fairly readable.  One of the things that makes it that way is generally fitting within line limits.  Until we auto format code, could this be wrapped to try to fit? Google style is binary ops go on the following line.  Something like:

```java
class Foo
    extends Bar
    implements Baz
```

carl-mastrangelo:
I'm curious how this will interact with the existing schema objects for marshallers.  What should the relationship be?  (can one be null but not the other?  Vice-versa?, Should marshaller schema be removed?)

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
ejona86: protobuf's `MethodDescriptor.getService()` is guaranteed to work.

protobuf/src/main/java/io/grpc/protobuf/ProtoMethodDescriptorSupplier.java
ejona86: DoNotMock an interface? What would break?

ejona86:
@carl-mastrangelo, there aren't schema objects for marshallers. There are schema objects for services though. It seems necessary to have a separate places to store them. We could use the same schema object for both, though (although then it takes more work to lookup the proto MethodDescriptor).

compiler/src/java_plugin/cpp/java_generator.cpp
lukaszx0: Done

protobuf/src/main/java/io/grpc/protobuf/ProtoMethodDescriptorSupplier.java
lukaszx0: Done

core/src/main/java/io/grpc/MethodDescriptor.java
lukaszx0: Done

core/src/main/java/io/grpc/MethodDescriptor.java
lukaszx0: Done

compiler/src/java_plugin/cpp/java_generator.cpp
lukaszx0: Done

protobuf/src/main/java/io/grpc/protobuf/ProtoMethodDescriptorSupplier.java
lukaszx0: Done

lukaszx0:
@carl-mastrangelo addressed your comments, PTAL

ejona86:
@carl-mastrangelo, ping

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/WorkerServiceGrpc.java
```diff
@@ -392,11 +396,32 @@ public void invoke(Req request, io.grpc.stub.StreamObserver<Resp> responseObserv
     }
   }
 
-  private static final class WorkerServiceDescriptorSupplier implements io.grpc.protobuf.ProtoFileDescriptorSupplier {
+  private static class WorkerServiceFileDescriptorSupplier
```
carl-mastrangelo: I would prefer if this class could be final, since every other class in this generate code is final too.

What do you think about making this final, making a static instance on the `MethodDescriptorSupplier`, and having `MethodDescriptorSupplier` implement the interfaces here?  


(non final classes, even private ones can be mocked by spying instances of them.  This was the cause of a recent test breakage)

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/WorkerServiceGrpc.java
```diff
@@ -392,11 +396,32 @@ public void invoke(Req request, io.grpc.stub.StreamObserver<Resp> responseObserv
...
+      implements io.grpc.protobuf.ProtoFileDescriptorSupplier, io.grpc.protobuf.ProtoServiceDescriptorSupplier {
     @java.lang.Override
     public com.google.protobuf.Descriptors.FileDescriptor getFileDescriptor() {
       return io.grpc.benchmarks.proto.Services.getDescriptor();
     }
+
```
carl-mastrangelo: This need a package-private constructor.  A private ctor causes synthetic classes to be made, and a public one makes reflection too easy.

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/WorkerServiceGrpc.java
```diff
@@ -392,11 +396,32 @@ public void invoke(Req request, io.grpc.stub.StreamObserver<Resp> responseObserv
...
+      return getFileDescriptor().findServiceByName("WorkerService");
+    }
+  }
+
+  private static final class WorkerServiceMethodDescriptorSupplier
+      extends WorkerServiceFileDescriptorSupplier
+        implements io.grpc.protobuf.ProtoMethodDescriptorSupplier {
+    private final String methodName;
+
+    private WorkerServiceMethodDescriptorSupplier(String methodName) {
```
carl-mastrangelo: This should be package-protected too.

benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/WorkerServiceGrpc.java
```diff
@@ -392,11 +396,32 @@ public void invoke(Req request, io.grpc.stub.StreamObserver<Resp> responseObserv
...
+    public com.google.protobuf.Descriptors.ServiceDescriptor getServiceDescriptor() {
+      return getFileDescriptor().findServiceByName("WorkerService");
+    }
+  }
+
+  private static final class WorkerServiceMethodDescriptorSupplier
+      extends WorkerServiceFileDescriptorSupplier
+        implements io.grpc.protobuf.ProtoMethodDescriptorSupplier {
+    private final String methodName;
+
```
carl-mastrangelo: Another alternative would be to nest the `MethodDescriptorSupplier`  class as an inner class of the `FileDescriptorSupplier` class, and let it capture the fields and methods.

ejona86:
@lukaszx0, it seems there are only small changes remaining. Want to fix them up, or us to?

ejona86:
@carl-mastrangelo, PTAL. Does this work for you? I created a base abstract class and extend it for both file and method descriptors, which are then both final.

I also took the liberty to make java_generator.cpp use a giant string for the block of code, which makes it much more readable and easier to modify for me.

ejona86:

core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -312,6 +315,20 @@ public InputStream streamResponse(RespT response) {
...
 
   /**
+   * Returns the schema descriptor for this method. A schema descriptor is an object that is not
+   * used by gRPC core but includes information related to the service method. The type of the
+   * object is specific to the consumer, so both the code setting the schema descriptor and the code
+   * calling {@link #getSchemaDescriptor()} must coordinate.  For example, protobuf generated code
+   * sets this value, in order to be consumed by the server reflection service.  See also:
+   * {@code io.grpc.protobuf.ProtoMethodDescriptorSupplier}.
+   *
+   * @since 1.5.0
```
carl-mastrangelo: 1.7.0

protobuf/src/main/java/io/grpc/protobuf/ProtoMethodDescriptorSupplier.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package io.grpc.protobuf;
+
+import com.google.protobuf.Descriptors.MethodDescriptor;
+import javax.annotation.CheckReturnValue;
+
+/**
+ * Provides access to the underlying proto service method descriptor.
+ */
```
carl-mastrangelo: Would you mind adding a `@Since` here too?

protobuf/src/main/java/io/grpc/protobuf/ProtoServiceDescriptorSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package io.grpc.protobuf;
+
+import com.google.protobuf.Descriptors.ServiceDescriptor;
+
+/**
+ * Provides access to the underlying proto service descriptor.
+ */
```
carl-mastrangelo: And here?

ejona86:
@carl-mastrangelo, can you review #3451. It includes the fixes requested (plus another one I noticed). I couldn't push to lukaszx0's branch, so I needed to create a new PR.

ejona86:
Merged as part of #3451

PR #3087, opened 2017-06-12
update netty-tcnative version in SECURITY.md

xuwei-k:


ejona86:
This would have normally been handled when we cherry-pick the updates to our README.md to 1.4.0, but it seems 1.4.0 failed to have this change as well.

PR #3088, opened 2017-06-12
v1.4.x: update netty-tcnative version in SECURITY.md; and bump to 4.0.3

ejona86:
This cherry-picks #3087 to the v1.4.x branch, since updating SECURITY.md was missed before the release.

If we were strict about the versions, we would recommend 2.0.1, since that's what we were using at the time of the release, but it doesn't seem to matter much in this case and 2.0.3 is expected to be a faster and support shading.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -62,7 +62,7 @@ In Maven, you can use the [os-maven-plugin](https://github.com/trustin/os-maven-
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-tcnative-boringssl-static</artifactId>
-      <version>1.1.33.Fork26</version>
+      <version>2.0.3.Final</version>
     </dependency>
   </dependencies>
 </project>
@@ -80,7 +80,7 @@ buildscript {
 }
 
 dependencies {
-    compile 'io.netty:netty-tcnative-boringssl-static:1.1.33.Fork26'
+    compile 'io.netty:netty-tcnative-boringssl-static:2.0.3.Final'
 }
 ```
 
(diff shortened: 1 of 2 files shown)


ejona86:
CC @xuwei-k

ejona86:
@carl-mastrangelo ping

dapengzhang0:
also `SECURITY.md` in master branch should be update to 2.05 now.

ejona86:
No, we update SECURITY.md to what was being used at release time. We've not done any release with 2.0.5

SECURITY.md
```diff
@@ -62,7 +62,7 @@ In Maven, you can use the [os-maven-plugin](https://github.com/trustin/os-maven-
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-tcnative-boringssl-static</artifactId>
-      <version>1.1.33.Fork26</version>
+      <version>2.0.3.Final</version>
```
carl-mastrangelo: 2.0.5

ejona86: This is a cherry-pick. Making real changes in a cherry-pick is not a great idea. I wasn't "changing it anyways", in that I didn't even open up an editor for the file.

carl-mastrangelo: Considering that there is a known bug, and we can't tell people to just use the latest version (1.5 isn't out yet), it seems reasonable to point to the latest version.  (for 1.3 and earlier it doesn't matter so much, we just tell people to upgrade)

ejona86: Then make a new commit and apply it to both master and v1.4.x. It is really important not to change cherrypicks when unnecessary and it seems unrelated to this PR.

ejona86:
@carl-mastrangelo, are we at an impasse here?

PR #3092, opened 2017-06-12
core: ensure ServerStream and ClientStream become noops after closing

### LINKED ISSUE #3083: Add unit tests to verify reads and writes are ignored after ServerStream#close
After `ServerStream#close` is called, any attempts to read from or write to the server stream should be silently ignored. Add unit tests to `AbstractTransportTest` to enforce this behavior.

(See also #3068)

zpencer:
Add unit tests to enforce behavior that we already expect today.

fixes #3083

### DIFF
--- a/testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
+++ b/testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
@@ -1283,7 +1283,7 @@ public abstract class AbstractTransportTest {
   }
 
   @Test
-  public void interactionsAfterClientCancelAreNoops() throws Exception {
+  public void interactionsAfterClientStreamCancelAreNoops() throws Exception {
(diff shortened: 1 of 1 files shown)


testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
```diff
@@ -1265,6 +1338,28 @@ private void doPingPong(MockServerListener serverListener) throws InterruptedExc
     ServerStream serverStream = serverStreamCreation.stream;
     ServerStreamListener mockServerStreamListener = serverStreamCreation.listener;
 
+    verify(mockServerStreamListener, never()).messageRead(any(InputStream.class));
```
ejona86: Actually sending messages seems unnecessary. This is mainly just making sure the HTTP/2 connection is still working.

PR #2386, opened 2016-10-28
Proto reflection API

ericgribkoff:
protobuf: Proto reflection service

This includes the core and compiler changes from [https://github.com/grpc/grpc-java/pull/2380](https://github.com/grpc/grpc-java/pull/2380)

PR #3093, opened 2017-06-13
all: Filter out Internal* classes from Javadoc

ejona86:
core's sources already have filters applied, so it isn't necessary to
copy them to all.

This filters InternalNotifyOnServerBuild, which is marked experimental. What's up with that one @ericgribkoff?

I should probably add the Netty Internal* classes to this PR as well.

### DIFF
--- a/all/build.gradle
+++ b/all/build.gradle
@@ -46,7 +46,6 @@ javadoc {
         source subproject.javadoc.source
         options.links subproject.javadoc.options.links.toArray(new String[0])
     }
-    exclude 'io/grpc/internal/**'
 }
 
 task jacocoMerge(type: JacocoMerge) {
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -10,4 +10,8 @@ dependencies {
     signature "org.codehaus.mojo.signature:java16:+@signature"
 }
 
-javadoc.exclude 'io/grpc/internal/**'
+javadoc {
+    // We want io.grpc.Internal, but not io.grpc.Internal*
+    exclude 'io/grpc/Internal?*.java'
+    exclude 'io/grpc/internal/**'
+}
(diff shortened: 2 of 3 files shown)


dapengzhang0:
Does it exclude `io.grpc.netty.InternalNettyChannelBuilder`?

ericgribkoff:
@ejona86 I'm not sure exactly what you're asking in regards to InternalNotifyOnServerBuild - it's necessary for the reflection service, but there's no tracking issue for it itself as part of the API. Should there be?

ejona86:
> Does it exclude io.grpc.netty.InternalNettyChannelBuilder?

No. But the same thing can be done for other packages.

@ericgribkoff, why does InternalNotifyOnServerBuild have "Internal" in its name if it is marked `@ExperimentalApi`? That seems contradictory; which way is it supposed to be?

ericgribkoff:
It has internal in the name because this was [suggested](https://github.com/grpc/grpc-java/pull/2386#discussion_r86055879) during the code review. As far as I was aware at that time we didn't have a hard rule about what the class name prefix internal signified. Maybe the experimental annotation should be dropped; I thought we wanted to potentially allow experimental usage of this api, and the internal went in as a discouraging factor in case users ignored the annotations, but I agree this is confusing to have both labels.

ejona86:
@dapengzhang0, PTAL. It now filters the Netty `Internal*` classes.

ejona86:
@ericgribkoff, okay. I'm going to filter `InternalNotifyOnServerBuild` and send out a separate PR to make it `@Internal`. If it is to become a stable API it seems it needs to be renamed, so the current version is `@Internal` in every regard then.

dapengzhang0:
LGTM

PR #3096, opened 2017-06-15
Add config and build files to run Windows tests on Kokoro

matt-kwong:
This adds the necessary files to run the current Jenkins Windows tests on Kokoro.

### DIFF
--- a/buildscripts/kokoro/windows.bat
+++ b/buildscripts/kokoro/windows.bat
@@ -1,0 +1,42 @@
+@rem ##########################################################################
+@rem
+@rem Script to set up Kokoro worker and run Windows tests
+@rem
+@rem ##########################################################################
+
+@rem Enter repo root
+cd /d %~dp0\..\..
+
+set WORKSPACE=T:\src\github\grpc-java
+set ESCWORKSPACE=%WORKSPACE:\=\\%
+
+mkdir grpc-java-helper
+
+@rem Install 7za
+@rem TODO(mattkwong): After Windows tests are no longer running on Jenkins, 7za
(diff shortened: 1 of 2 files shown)


buildscripts/kokoro/windows.cfg
```diff
@@ -0,0 +1,11 @@
+# Config file for internal CI
+
+# Location of the continuous shell script in repository.
+build_file: "grpc-java/buildscripts/kokoro/windows.bat"
+timeout_mins: 360
```
ejona86: That's 6 hours? How about we limit it to something smaller, say 45 minutes?

buildscripts/kokoro/windows.bat
```diff
@@ -0,0 +1,40 @@
...
+cd /d %~dp0\..\..
+
+set WORKSPACE=T:\src\github\grpc-java
+set ESCWORKSPACE=%WORKSPACE:\=\\%
+
+mkdir grpc-java-helper
+
+@rem Install 7za
+wget http://www.7-zip.org/a/7za920.zip
+unzip 7za920.zip
```
ejona86: If we have unzip then we don't need 7za. Our other build machines don't have unzip. But it'd be annoying to support both in make_dependencies.bat. Add a TODO here to swap make_dependencies.bat to use unzip instead.

matt-kwong: Is Jenkins the only place `make_dependencies.bat` is used currently?

ejona86: Yeah, although it is a nice feature that `make_dependencies.bat` can be run manually. But depending on unzip vs 7za doesn't really make it any harder to run manually.

PR #3098, opened 2017-06-16
core,netty,okhttp: strip outbound headers with reserved names

### LINKED ISSUE #3017: Duplicate headers causes c-core failure

1.3.0

1.8.0_121

### What did you do?
If possible, provide a recipe for reproducing the error.
I made a server interceptor as follows:

```
    @Override
    public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(final ServerCall<ReqT, RespT> serverCall,
                                                                 final Metadata headers,
                                                                 final ServerCallHandler<ReqT, RespT> next) {
        final Optional<String> targetAlias = resolveRedirectTarget();
        if (!targetAlias.isPresent()) {
            return next.startCall(serverCall, headers);
        }

        final ManagedChannel channel;
        final ClientCall<ReqT, RespT> clientCall;
        try {
            channel = channelFactory.get(targetAlias.get());
            clientCall = channel.newCall(serverCall.getMethodDescriptor(), CallOptions.DEFAULT);
            clientCall.start(new ClientCall.Listener<RespT>() {
                @Override
                public void onHeaders(final Metadata headers) {
                    // Issue is here
                    serverCall.sendHeaders(headers);
                }

                @Override
                public void onMessage(final RespT message) {
                    serverCall.sendMessage(message);
                }

                @Override
                public void onClose(final Status status, final Metadata trailers) {
                    serverCall.close(status, trailers);
                }

                @Override
                public void onReady() {
                    serverCall.request(1);
                }

            }, headers);
            clientCall.request(2);
            serverCall.request(1);
        } catch (final Exception ignored) {
            return next.startCall(serverCall, headers);
        }

        return new ServerCall.Listener<ReqT>(){
            public void onMessage(final ReqT message) {
                clientCall.sendMessage(message);
            }

            public void onHalfClose() {
                clientCall.halfClose();
            }

            public void onCancel() {
                clientCall.cancel("Redirect call was cancelled upstream.", null);
            }

             public void onReady() {
                clientCall.request(1);
            }
        };
    }
```

A caller of this service to get a response that has been proxied through to a different backend target. 

### What did you see instead?
This works in Java, but not in c-core from the client side.  As a result of forwarding the proxy target response headers, there ends up being 2 entries for "content-type"  Java clients are okay with this, c based clients are not.  I'm not sure who is "right" here as the spec is ambiguous as to the desired behavior if there are multiple headers for the same key(and in this case, their values are identical as well).  My fix for now is to simply send a new Metadata() rather than forwarding headers, but this might lose some custom headers from the proxy target.  I also considered doing a headers.removeAll(GrpcUtil.CONTENT_TYPE/ENCODING_TYPE/ETC) but since those are in the internal package I didn't.  I'm wondering what your thoughts are on how I'm doing this.  Maybe there's even a trick where I could pull this off at a pure bytestream level so my proxy isn't parsing the bytes at all?

zpencer:
Today applications can put arbitrary key values in the Metadata field. This can lead to duplicates of certain key that have special meaning to gRPC. This PR discards the user specified keys so that there is only 1 key-value for each of the special reserved keys.

fixes #3017

### DIFF
--- a/core/src/main/java/io/grpc/internal/GrpcUtil.java
+++ b/core/src/main/java/io/grpc/internal/GrpcUtil.java
@@ -98,6 +98,12 @@ public final class GrpcUtil {
           Metadata.Key.of("content-type", Metadata.ASCII_STRING_MARSHALLER);
 
   /**
+   * {@link io.grpc.Metadata.Key} for the Transfer encoding.
+   */
+  public static final Metadata.Key<String> TE_HEADER =
+      Metadata.Key.of("te", Metadata.ASCII_STRING_MARSHALLER);
+
+  /**
    * {@link io.grpc.Metadata.Key} for the Content-Type request/response header.
    */
   public static final Metadata.Key<String> USER_AGENT_KEY =
@@ -112,12 +118,6 @@ public final class GrpcUtil {
    * The default port for SSL connections.
    */
   public static final int DEFAULT_PORT_SSL = 443;
-
-  /**
-   * {@link io.grpc.Metadata.Key} for the Transfer encoding.
-   */
-  public static final Metadata.Key<String> TE_HEADER =
-      Metadata.Key.of("te", Metadata.ASCII_STRING_MARSHALLER);
 
   /**
    * Content-Type used for GRPC-over-HTTP/2.
--- a/core/src/main/java/io/grpc/internal/GrpcUtil.java
+++ b/core/src/main/java/io/grpc/internal/GrpcUtil.java
@@ -619,20 +619,4 @@ public final class GrpcUtil {
     }
     return "";
   }
-
-  /**
-   * Strips away any reserved headers for client requests.
-   */
-  public static void discardHttp2RequestHeaders(Metadata headers) {
-    headers.discardAll(CONTENT_TYPE_KEY);
-    headers.discardAll(TE_HEADER);
-    headers.discardAll(USER_AGENT_KEY);
-  }
-
-  /**
-   * Strips away any reserved headers for server responses.
-   */
-  public static void discardHttp2ResponseHeaders(Metadata headers) {
-    headers.discardAll(CONTENT_TYPE_KEY);
-  }
 }
--- a/core/src/main/java/io/grpc/internal/GrpcUtil.java
+++ b/core/src/main/java/io/grpc/internal/GrpcUtil.java
@@ -23,6 +23,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
 import com.google.common.base.Stopwatch;
 import com.google.common.base.Supplier;
+import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import io.grpc.CallOptions;
@@ -40,6 +41,7 @@ import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.Charset;
+import java.util.List;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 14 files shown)


zpencer:
It would be nice to avoid checks once per send, though it will mean a bit of coupling between the Metadata which is transport agnostic and the HTTP transport. I see some pre-existing code in `validateName()` that finds keys starting with a single ':' and allows them, contrary to the API. Removing this would get us most of the way there, except for "content-type", "user-agent" and "te".

core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -618,4 +624,20 @@ private static String getImplementationVersion() {
     }
     return "";
   }
+
+  /**
+   * Strips away any reserved headers for client requests.
+   */
+  public static void discardHttp2RequestHeaders(Metadata headers) {
```
ejona86: I'm not excited about how far away from the usages these are. OkHttp in particular would be very easy to put the discards near where the other keys are added. With Netty, the filtering could happen in GrpcHttp2OutboundHeaders.

core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -92,26 +92,32 @@
   }
 
   /**
-   * {@link io.grpc.Metadata.Key} for the Content-Type request/response header.
+   * The default port for plain-text connections.
```
zpencer: Originally I wanted to make the header keys be closer in the file to the header values, but I don't feel strongly about this.

zpencer:
- Created a publicly accessible static `Metadata.HTTP2_STATUS_KEY` for reading `:status`. Creating pseudo headers is not possible using the public API.
- Added more restrictions to pseudo headers: they can never be put() into a Metadata, only parsed from raw byte[]s.

zpencer:
Synced to latest master where users can not create pseudo headers. This means we now only need to remove user duplicated grpc headers that are non pseudo. Also moved the remove code to be closer to where we set the header fields, getting rid of the util functions.

netty/src/main/java/io/grpc/netty/GrpcHttp2OutboundHeaders.java
```diff
@@ -81,6 +81,21 @@ public CharSequence status() {
     return new Itr();
   }
 
+  /**
+   * Returns the first value whose key matches the input.
+   */
+  @Override
```
carl-mastrangelo: im reluctant to add this method.  It really shouldn't exist since it isnt very efficient.  Can you just dump all the headers into a DefaultHttp2Headers if you want to check if they are present?

okhttp/src/test/java/io/grpc/okhttp/HeadersTest.java
```diff
@@ -0,0 +1,60 @@
...
+    metaData.put(GrpcUtil.CONTENT_TYPE_KEY, "to-be-removed");
+    metaData.put(GrpcUtil.USER_AGENT_KEY, "to-be-removed");
+    metaData.put(GrpcUtil.TE_HEADER, "to-be-removed");
+
+
+    Metadata.Key<String> userKey = Metadata.Key.of("user-key", Metadata.ASCII_STRING_MARSHALLER);
+    String userValue =  "user-value";
+    metaData.put(userKey, userValue);
+
+    final String path = "//testServerice/test";
```
carl-mastrangelo: nit: locals are usually not final.

okhttp/src/test/java/io/grpc/okhttp/HeadersTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    // Intentionally being explicit here rather than relying on any pre-defined lists of headers,
+    // since the goal of this test is to validate the correctness of such lists in the first place.
+    metaData.put(GrpcUtil.CONTENT_TYPE_KEY, "to-be-removed");
+    metaData.put(GrpcUtil.USER_AGENT_KEY, "to-be-removed");
+    metaData.put(GrpcUtil.TE_HEADER, "to-be-removed");
+
+
+    Metadata.Key<String> userKey = Metadata.Key.of("user-key", Metadata.ASCII_STRING_MARSHALLER);
+    String userValue =  "user-value";
```
carl-mastrangelo: extra space

okhttp/src/test/java/io/grpc/okhttp/HeadersTest.java
```diff
@@ -0,0 +1,60 @@
...
+    final String authority = "localhost";
+    final String userAgent = "useragent";
+
+    List<Header> headers = Headers.createRequestHeaders(metaData, path, authority, userAgent);
+
+    // 7 reserved headers, 1 user header
+    assertEquals(7 + 1, headers.size());
+    // Check the 3 reserved headers that are non pseudo
+    // Users can not create pseudo headers keys so no need to check for them here
+    assertTrue(headers.contains(Headers.CONTENT_TYPE_HEADER));
```
carl-mastrangelo: Optional: use Truth.assertThat(headers).contains(....)

netty/src/main/java/io/grpc/netty/GrpcHttp2OutboundHeaders.java
zpencer: Dumping it by iterating externally is not too bad. Done.

okhttp/src/test/java/io/grpc/okhttp/HeadersTest.java
zpencer: fixed

okhttp/src/test/java/io/grpc/okhttp/HeadersTest.java
zpencer: Was intending to show it's the golden value we are comparing against, but don't feel strongly about this.

okhttp/src/test/java/io/grpc/okhttp/HeadersTest.java
zpencer: changed

zpencer:
@carl-mastrangelo ping

netty/src/main/java/io/grpc/netty/Utils.java
```diff
@@ -118,6 +122,9 @@ public static Http2Headers convertClientHeaders(Metadata headers,
   }
 
   public static Http2Headers convertServerHeaders(Metadata headers) {
+    // Discard any application supplied duplicates of the reserved headers
+    headers.discardAll(CONTENT_TYPE_KEY);
```
carl-mastrangelo: remove te and user agent here too?

netty/src/test/java/io/grpc/netty/UtilsTest.java
```diff
@@ -43,6 +53,68 @@ public void testStatusFromThrowable() {
...
+    for (Map.Entry<CharSequence, CharSequence> entry : output) {
+      headers.add(entry.getKey(), entry.getValue());
+    }
+
+    // 7 reserved headers, 1 user header
+    assertEquals(7 + 1, headers.size());
+    // Check the 3 reserved headers that are non pseudo
+    // Users can not create pseudo headers keys so no need to check for them here
+    assertEquals(GrpcUtil.CONTENT_TYPE_GRPC,
+        headers.get(GrpcUtil.CONTENT_TYPE_KEY.name()).toString());
```
carl-mastrangelo: does toString() need to be here?

zpencer:
This is ready to be looked at

PR #3099, opened 2017-06-16
stub: add generic method handler registry

dapengzhang0:
This is the first of a sequence of PRs to make unit test easier.

Although we have `InProcess` transports and `GrpcServerRule`, unit testing is sometimes still very hard, especially when testing a user-defined generic interceptor without depending on an existing service or method, without proto codegen. 

To test this kind of interceptor, the user has to provide a `ClientCall` object. 
`ClientCall` is marked `@DoNotMock`, because a `ClientCall` object is not something that a user can provide, it is always produced by the grpc core library (and wrapped with interceptors). 

Without an existing service or method, it's hard to get a stub or method descriptor and so it's hard to use  `GrpcServerRule`. So it is hard to get a genuine `ClientCall` to test the interceptor. Users are now using `NoopClientCall` basically as an alternative of mocked client call, which is as bad as a mock. I've seen people writing wrong features in their interceptor and all tests passed with `NoopClientCall`, which would fail if using a genuine `ClientCall` instead.

So it would be helpful if the test framework can

- register generic methods to `GrpcServerRule`; (this PR)
- provide generic InProcess methods descriptors (`TestMethodDescriptors.noopMethod` or mocked method descriptor does not work). (next PR)
- make `NoopClientCall` for internal developer/advanced use only, not for application developers.

### DIFF
--- a/stub/src/main/java/io/grpc/stub/util/GenericMethodHandlerRegistry.java
+++ b/stub/src/main/java/io/grpc/stub/util/GenericMethodHandlerRegistry.java
@@ -41,9 +41,9 @@ import javax.annotation.Nullable;
  * @since 1.5.0
  */
 @ExperimentalApi() // TODO(zdapeng): tracking url
-public class GenericMethodHandlerRegistry extends HandlerRegistry {
+public final class GenericMethodHandlerRegistry extends HandlerRegistry {
 
-  private MutableHandlerRegistry delegate = new MutableHandlerRegistry();
+  private final MutableHandlerRegistry delegate = new MutableHandlerRegistry();
 
   /**
    * Registers a service with a generic unary method.
@@ -128,7 +128,7 @@ public class GenericMethodHandlerRegistry extends HandlerRegistry {
    *         otherwise {@code null}.
    */
   @Nullable
-  public final ServerServiceDefinition addService(ServerServiceDefinition service) {
+  public ServerServiceDefinition addService(ServerServiceDefinition service) {
     return delegate.addService(service);
   }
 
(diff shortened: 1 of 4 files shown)


stub/src/main/java/io/grpc/stub/util/GenericMethodHandlerRegistry.java
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * An implementation of {@link HandlerRegistry} which can add generic methods or services.
+ *
+ * @since 1.5.0
+ */
+@ExperimentalApi() // TODO(zdapeng): tracking url
+public class GenericMethodHandlerRegistry extends HandlerRegistry {
+
+  private MutableHandlerRegistry delegate = new MutableHandlerRegistry();
```
dapengzhang0: I forgot `final` here

ejona86:
I'm not excited about this approach. I agree things could be better, but I feel like this isn't very orthogonal and is too much/awkward for a very specific use-case. The UUIDs in #3100 make me feel that way even more.

A simpler approach would simply be to provide a service for unit testing interceptors. I'm not sure I want to do that either, but it seems better than adding additional special-purpose code with large APIs. I'm already annoyed with much of the test classes/methods and their limited usability and noise. I'd rather take a step back and reconsider the testing tools.

dapengzhang0:
UUID is not a big deal, it can be replaced by user providing the serviceName and methodName arguments.

We can not simply provide a service for unit testing interceptors, because we don't know if the user wants the most general service or protobuf only or json only or a fixed reqT and respT only. That would be too much. The motivation of this PR is I'm annoyed with `NoopClientCall` which is used just as another mock. It should be deprecated.

dapengzhang0:
filed issue #3105

ejona86:
> We can not simply provide a service for unit testing interceptors, because we don't know if the user wants the most general service or protobuf only or json only or a fixed reqT and respT only.

But interceptors shouldn't really care about the message type itself. If an interceptor cares, they should make their own test service for testing it.

PR #3100, opened 2017-06-16
testing: add inProcessGenericMethod descriptor

dapengzhang0:
This is the second of a sequence of PRs to make unit test easier.

The first is #3099 

test will be added after #3099.

### DIFF
--- a/testing/src/main/java/io/grpc/testing/TestMethodDescriptors.java
+++ b/testing/src/main/java/io/grpc/testing/TestMethodDescriptors.java
@@ -16,11 +16,17 @@
 
 package io.grpc.testing;
 
+import com.google.common.io.ByteStreams;
+import com.google.common.primitives.Longs;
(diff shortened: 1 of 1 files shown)


ejona86:
@dapengzhang0 and I spoke offline. We're thinking all that's really necessary for most interceptor users is a simple test service. We're planning on adding a .proto with such a test service, for users to use in their own tests. It will be _very_ simple though, with only 1 or 4 methods (because streaming). Users that need more should create their own test.proto.

This does mean interceptor _tests_ would depend on proto, although they wouldn't need to run protoc. We considered having a simple MethodDescriptor instead, but found it still a bit too annoying because you have to create your own ServerServiceDefinition. Not the end of the world, but the proto-based solution seems easier for our users and we don't except a superficial _test_ dependency on protobuf to be burdensome.

PR #3089, opened 2017-06-12
all: update readme to point to 1.4.0

ejona86:
This is a cherry-pick from 1.4.x

PR #3101, opened 2017-06-16
all: update readme to point to 1.4.0

carl-mastrangelo:


ejona86:
Why not #3089 (and #3088) that I sent out earlier this week?

PR #3103, opened 2017-06-16
core: Mark InternalNotifyOnServerBuilder @Internal

ejona86:
It is misnamed if it is @ExperimentalApi. Even being @Internal is
strange, since nothing in io.grpc actually uses it and so it could
actually be in io.grpc.internal.

I'm not trying to fix the API, but just make it slightly more
consistent.


ejona86:
@ericgribkoff, I feel like there needs to be a public API for users, otherwise they'll start using the Internal API, because that's what we did and they don't have any other option.

PR #3104, opened 2017-06-16
netty: rename HandlerSettings to InternalHandlerSettings

dapengzhang0:


netty/src/main/java/io/grpc/netty/HandlerSettings.java
```diff
@@ -23,9 +23,12 @@
 /**
  * Allows autoFlowControl to be turned on and off from interop testing and flow control windows to
  * be accessed. For internal use only.
+ *
+ * @deprecated renamed to {@link InternalHandlerSettings} and should not be used externally
  */
 @VisibleForTesting // Visible for tests in other packages.
 @Internal
+@Deprecated
 public final class HandlerSettings {
```
ejona86: This now looks unused. Using it does nothing. Why not delete?

netty/src/main/java/io/grpc/netty/InternalHandlerSettings.java
```diff
@@ -0,0 +1,76 @@
...
+import com.google.common.base.Preconditions;
+import io.grpc.Internal;
+
+/**
+ * Allows autoFlowControl to be turned on and off from interop testing and flow control windows to
+ * be accessed. For internal use only.
+ */
+@VisibleForTesting // Visible for tests in other packages.
+@Internal
+public final class InternalHandlerSettings {
```
ejona86: Why is this not detected as a copy/move?

netty/src/main/java/io/grpc/netty/HandlerSettings.java
dapengzhang0: I was thinking to delete it in the next release cycle.

netty/src/main/java/io/grpc/netty/InternalHandlerSettings.java
dapengzhang0: maybe because I just ran refactor, maybe the file named HandlerSettings.java is still there

netty/src/main/java/io/grpc/netty/HandlerSettings.java
dapengzhang0: It was a public API, the annotation `@Internal` was just added recently.

ejona86: Then let's make it clear this implementation is useless and calling it does nothing. Let's remove the package-private and private stuff and replace the public pieces with fakes. And then change the docs to say it does nothing.

dapengzhang0: as discussed offline, all public methods in `HandlerSettings` just call `InternalHandlerSettings`'s methods.

zhangkun83:
Internally `netty:internal` depends on `netty`. In this PR `HandlerSettings` depends on `InternalHandlerSettings`, which is the other way around. Now this is blocking the import. @ejona86 @carl-mastrangelo, which direction should we go with?

ejona86:
@zhangkun83, Just delete `HandlerSettings` (maybe just internally)?

zhangkun83:
It doesn't help. `NettyClientTransport` also uses `InternalHandlerSettings`. Shall we move the meat of `InternalHandlerSettings` into some package-private class, and make both `HandlerSettings` and `InternalHandlerSettings` depend on it?

dapengzhang0:
To unblock the import, temporarily exclude `InternalHandlerSettings.java` from `netty:internal`'s `srcs` and add it to `netty`'s?

ejona86:
This was resolved by moving the code from one alias to the other: https://github.com/grpc/grpc-java/pull/3143

PR #3107, opened 2017-06-19
Fix LICENSE file

jtattermusch:
See https://github.com/grpc/grpc-go/issues/1309

New LICENSE file is copied from http://www.apache.org/licenses/LICENSE-2.0.txt

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,13 +1,14 @@
 
                                  Apache License
                            Version 2.0, January 2004
-                        https://www.apache.org/licenses/
+                        http://www.apache.org/licenses/
 
    TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
    1. Definitions.
 
       "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
 
       "Licensor" shall mean the copyright owner or entity authorized by
       the copyright owner that is granting the License.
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -171,13 +176,24 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
mehrdada: Similarly to the grpc/grpc version, shouldn't this template parameter be specialized?

carl-mastrangelo: Other OSS projects don't seem to:

https://github.com/kubernetes/kubernetes/blob/master/LICENSE
https://github.com/netty/netty/blob/4.1/LICENSE.txt

mehrdada: The Go repo has it though, so I am not sure who is doing it right. Whatever we do we need to be consistent among ourselves at least: https://github.com/grpc/grpc-go/blob/master/LICENSE

jtattermusch:
Let's merge the current version as it is a significant improvement over what we have right now (and is consistent with grpc/grpc and grpc/grpc-go). Later we can remove the appendix.

LICENSE
jtattermusch: grpc/grpc won't have it and I'm pretty sure go has been updated ad-hoc without doing much research. Let's merge this and update go later.

PR #3109, opened 2017-06-19
Use CodedInputStream#newInstance(ByteBuffer) for non-segmented messages

### LINKED ISSUE #2937: Use CodedInputStream.newInstance(ByteBuffer) for non-segmented messages
This is to improve the decode performance of small messages (<16 KB) to avoid a message-sized allocation and copy.

Protobuf already has an optimized code path for [decoding a direct `ByteBuffer`](https://github.com/google/protobuf/blob/v3.2.0/java/core/src/main/java/com/google/protobuf/CodedInputStream.java#L141). We currently [copy to a `byte[]`](https://github.com/grpc/grpc-java/blob/v1.2.0/protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoLiteUtils.java#L139), but that could be avoided because small messages will commonly be in a single `ByteBuffer`.

This will require adding a new interface for retrieving the `ByteBuffer` from the `InputStream`. I'm uncertain whether the interface should support returning multiple `ByteBuffer`s.

amorken:
Punch a few holes in abstractions to permit direct access to ByteBuffers for CodedInputStream. This reduces copying and garbage production and enables an optimized code path in CodedInputStream for decoding direct ByteBuffers.

Fixes #2937

### DIFF
--- a/core/src/main/java/io/grpc/internal/NioBufferBackedStream.java
+++ b/core/src/main/java/io/grpc/internal/NioBufferBackedStream.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2017, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.internal;
+
+import java.nio.ByteBuffer;
+
+public interface NioBufferBackedStream {
+  ByteBuffer nioBuffer();
+}
--- a/protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoLiteUtils.java
+++ b/protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoLiteUtils.java
@@ -30,6 +30,8 @@ import io.grpc.MethodDescriptor.Marshaller;
 import io.grpc.MethodDescriptor.PrototypeMarshaller;
 import io.grpc.Status;
 import io.grpc.internal.GrpcUtil;
+import io.grpc.internal.NioBufferBackedStream;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.ref.Reference;
@@ -119,7 +121,15 @@ public class ProtoLiteUtils {
         }
         CodedInputStream cis = null;
         try {
-          if (stream instanceof KnownLength) {
+          if (stream instanceof NioBufferBackedStream) {
+            int size = stream.available();
+            if (size > 0) {
+              NioBufferBackedStream nioBufferBackedStream = (NioBufferBackedStream) stream;
+              cis = CodedInputStream.newInstance(nioBufferBackedStream.nioBuffer());
+            } else {
+              return defaultInstance;
+            }
+          } else if (stream instanceof KnownLength) {
             int size = stream.available();
             if (size > 0 && size <= GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE) {
               // buf should not be used after this method has returned.
--- a/core/src/main/java/io/grpc/internal/AbstractReadableBuffer.java
+++ b/core/src/main/java/io/grpc/internal/AbstractReadableBuffer.java
@@ -16,6 +16,8 @@
 
 package io.grpc.internal;
 
+import java.nio.ByteBuffer;
+
 /**
  * Abstract base class for {@link ReadableBuffer} implementations.
  */
@@ -61,6 +63,16 @@ public abstract class AbstractReadableBuffer implements ReadableBuffer {
 
   @Override
   public int arrayOffset() {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public boolean hasNioBuffer() {
+    return false;
+  }
+
+  @Override
+  public ByteBuffer nioBuffer() {
     throw new UnsupportedOperationException();
   }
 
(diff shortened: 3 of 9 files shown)


amorken:
@ejona86, this may be relevant to your interests.

No new tests, code path appears to be exercised by existing tests, but if you want a specific test, just ask.

Includes a drive-by fix for confusing BufferInputStream#close() behavior/javadoc. I hope this doesn't break anything, but it's required to avoid a Netty ByteBuf leak.

I'm also not sure whether retrieving the NIO ByteBuffer from CompositeReadableBuffer should mark the returned buffer as read (remove it) or not. It should perhaps be "read" (i.e. removed), as the NioBufferInputStream takes ownership of and closes the buffer. Or perhaps the NioBufferInputStream should not assume ownership? Guru guidance welcome.

Thanks,
-AndersM

ejona86:
Okay to test

amorken:
I admit that I have no idea why the Travis build failed. I don't see how the changes I made could cause any impact on a generated proto file - I haven't changed any protobuf definitions. Could anything else cause this?

```test -z "$(git status --porcelain)" || (git status && echo Error Working directory is not clean. Forget to commit generated files? && false)
# HEAD detached at FETCH_HEAD
# Untracked files:
#   (use "git add <file>..." to include in what will be committed)
#
#	benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/ReportQpsScenarioServiceGrpc.java
#
nothing added to commit but untracked files present (use "git add" to track)
Error Working directory is not clean. Forget to commit generated files?```

**Comment by ejona86** [comment]
@amorken, wasn't your fault. Fixed in #3113

**Comment by ejona86** [comment]
FYI: I have looked at this some. But need to look at it more.

**Comment by amorken** [comment]
Ack. I have a hunch about some possible issues, at least the question of who's in charge of closing the buffers, but I'm not in a particular hurry and won't touch anything until you've had a chance to make a complete pass. =)

Thanks,
-AndersM

File: protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoLiteUtils.java [hunk 1]
```diff
@@ -119,7 +121,15 @@ public T parse(InputStream stream) {
         }
         CodedInputStream cis = null;
         try {
-          if (stream instanceof KnownLength) {
+          if (stream instanceof NioBufferBackedStream) {
+            int size = stream.available();
+            if (size > 0) {
+              NioBufferBackedStream nioBufferBackedStream = (NioBufferBackedStream) stream;
+              cis = CodedInputStream.newInstance(nioBufferBackedStream.nioBuffer());
```
zpencer: Although in `ServerStreamListenerImpl` we can assume the `InputStream` is always closed at the end, this is not a part of the `Marshaller#parse` API. If we only read some of the stream and them pass the stream in again a second time to read more data, we would start over from the beginning because `nioBuffer` returns a new one every time. To be clean here we probably should probably maintain the read offset ourselves, so that whether the user reads from the returned ByteBuffer or from reading the InputStream directly, the read offset is consistent.

I also synced with @ejona86 and learned there's a rope based API soon coming to the protobuf CodedInputStream API. I think this is the javadoc from that project:
https://github.com/google/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/RopeByteString.java

The rope is essentially a list of byte arrays, which enables us to generalize your optimization to `CompositeReadableBuffer`s with more than 1 buffer, which would imply the return value of `NioBufferBackedStream#nioBuffer` will be a list of buffers.

core/src/main/java/io/grpc/internal/NioBufferBackedStream.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.internal;
+
+import java.nio.ByteBuffer;
+
+public interface NioBufferBackedStream {
```
zpencer: Missing javadoc

protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoLiteUtils.java
dapengzhang0: Maybe we can use `ByteBuffer.duplicate()` or `slice()` to create a new ByteBuffer that shares the same content. Because it only reads from the byte buffer here so it's safe to duplicate.

dapengzhang0: Oh, from the javadoc of ` CodedInputStream.newInstance(ByteBuffer buf)`, it seems this is not necessary.
```java
 /**
   * Create a new CodedInputStream wrapping the given ByteBuffer. The data starting from the
   * ByteBuffer's current position to its limit will be read. The returned CodedInputStream may or
   * may not share the underlying data in the ByteBuffer, therefore the ByteBuffer cannot be changed
   * while the CodedInputStream is in use. Note that the ByteBuffer's position won't be changed by
   * this function. Concurrent calls with the same ByteBuffer object are safe if no other thread is
   * trying to alter the ByteBuffer's status.
   */
public static CodedInputStream newInstance(ByteBuffer buf)
```

amorken: I think I agree with the principle that both kinds of reading should advance the underlying pointer. I did it the simple way here, as the scope of the optimization was limited to the simple case of a single buffer getting thrown over the fence to the ProtoLite parser, but I'll see if I can get enough of a sense of the whole to be consistent with it.

amorken:
Thanks for the feedback both of you, sorry I haven't gotten around to dealing with 
this yet. I'll see if I can scrape out some time to address it the next few days. Sorry about the forgotten Javadoc, silly mistake. (And I even mixed it into the wrong commit... Oh well. =)

dapengzhang0:
I think this PR is obsolete. @amorken feel free to reopen if you have update.

thelinuxfoundation:

* User @amorken isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

PR #3113, opened 2017-06-20
benchmarks: add missing ReportQpsScenarioServiceGrpc

ericgribkoff:
Looks like this was overlooked earlier


adelez:
May I ask what's the usage of this?

ericgribkoff:
This PR just adds in an auto-generated file that `gradle build` generated but was not checked in. The relevant proto file changes were in https://github.com/grpc/grpc-java/pull/3108, and I believe just a (re)sync of the benchmark protos with those in grpc/grpc.

PR #3114, opened 2017-06-20
core: hide access to pseudo headers

carl-mastrangelo:
Previously anyone could create metadata keys with a leading ":", due to not having a way to prevent it.  This change effectively only allows internal code to make use of this.

### DIFF
--- a/core/src/main/java/io/grpc/Metadata.java
+++ b/core/src/main/java/io/grpc/Metadata.java
@@ -599,11 +599,15 @@ public final class Metadata {
      *     <b>not</b> end with {@link #BINARY_HEADER_SUFFIX}
      */
     public static <T> Key<T> of(String name, AsciiMarshaller<T> marshaller) {
-      return new AsciiKey<T>(name, marshaller);
+      return of(name, false, marshaller);
     }
 
-    static <T> Key<T> of(String name, TrustedAsciiMarshaller<T> marshaller) {
-      return new TrustedAsciiKey<T>(name, marshaller);
+    static <T> Key<T> of(String name, boolean pseudo, AsciiMarshaller<T> marshaller) {
+      return new AsciiKey<T>(name, pseudo, marshaller);
+    }
+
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/grpc/Metadata.java
```diff
@@ -642,9 +645,9 @@ private static String validateName(String n) {
       return n;
     }
 
-    private Key(String name) {
+    private Key(String name, boolean pseudo) {
```
zpencer: Do we want to be extra safe and disallow calling .put() on any keys that are pseudo?
Also, there are users today who try to read ":status", so we should probably pre-define a Key<Integer> for reading ":status".

carl-mastrangelo: No for put.  if they cannot create keys that are pseudo , there is no need to check.

User who try to read `:status` may have a legitimate need, but I don't think it is currently supported.  If we changed the transport to inProcess, the result would not make sense.  Any other transport would reveal this problem.  Until we have thought about how to support this, I would like to prevent it.

ejona86: We have told one user in particular they can use `:status` in a supported state. It isn't _stable_ but they have a very legitimate use-case for when the RPC never gets started, needing to check the HTTP status. We'll need to figure out an alternative for them and then migrate them.

I see two alternative ways of supporting the :status use-case: an attribute on `ClientCall`, or just change the name to a `grpc-`-prefixed name like `grpc-http-status-code`. Attribute seems cleaner, but using a different name is probably easier.

carl-mastrangelo: It seems the user of that header is now no longer using it.  Can this PR proceed?

ejona86: Yes. There are no longer any known users using the pseudo header hack.

carl-mastrangelo: @ejona86  Synced to head, PTAL

Issue #3095, opened 2017-06-13
How to cancel a GRPC Server streaming call


varshanbp:
I made a Call to server. The server streams. I want to cancel the call from client.

How to cancel? No Method to cancel a call.

fanminshi:
@varshanbp I think you need to call `onCompleted()` of the streamObserver from client to terminate the stream call. 
http://www.grpc.io/docs/tutorials/basic/java.html

varshanbp:
I use "Server-side streaming RPC". I want to close the channel from client. If I shutdown channel from client, it throws error in Server and keeps returning value to closed channel. How to solve this problem?

**I want to close a channel from Client properly.**

fanminshi:
@varshanbp you probably don't want to kill channel right away. maybe follow the example here to properly shutdown channel.

```java
  public void stop() {
    if (server != null) {
      server.shutdown();
    }
  }

  /**
   * Await termination on the main thread since the grpc library uses daemon threads.
   */
  private void blockUntilShutdown() throws InterruptedException {
    if (server != null) {
      server.awaitTermination();
    }
  }
```
 https://github.com/grpc/grpc-java/blob/master/examples/src/main/java/io/grpc/examples/routeguide/RouteGuideServer.java#L83-L96

conorgriffin:
@varshanbp wants to shut down the stream from the client side, this example shuts it down from the server side @fanminshi

fanminshi:
@conorgriffin oops. wrong example. I meant to show this one.

```java
  public void shutdown() throws InterruptedException {
    channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
  }
```
https://github.com/grpc/grpc-java/blob/master/examples/src/main/java/io/grpc/examples/routeguide/RouteGuideClient.java#L62-L64

carl-mastrangelo:
@varshanbp In the future, this kind of question would be better posted on Stack Overflow or the mailing list.

To answer your question:  You can cancel a call on the client by calling `onError(Status.CANCELED)`.  The actual error doesn't matter too much since it won't actually be put on the wire.  gRPC will send an RST_STREAM packet which will show up on the server's `onError` handler.  

If you want a more graceful shutdown, you can modify your streaming RPC to just make `onCompleted` interpreted as a cancellation, which will make things close down more gracefully.

[...]

PR #3115, opened 2017-06-20
Implemented ClientCallStreamObserver.cancel()

rmichela:
Allow a client to gracefully cancel a one-to-many streaming service by signaling the server to stop sending messages.

Addresses #3095

### DIFF
--- a/stub/src/main/java/io/grpc/stub/ClientCallStreamObserver.java
+++ b/stub/src/main/java/io/grpc/stub/ClientCallStreamObserver.java
@@ -25,20 +25,20 @@ import javax.annotation.Nullable;
  * client calls.
  */
 @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1788")
-public abstract class ClientCallStreamObserver<V> extends CallStreamObserver<V> {
-    /**
-     * Prevent any further processing for this {@code ClientCallStreamObserver}. No further messages will be received.
-     * The server is informed of cancellations, but may not stop processing the call. Cancelling an already
-     * {@code cancel()}ed {@code ClientCallStreamObserver} has no effect.
-     *
-     * <p>No other methods on this class can be called after this method has been called.
-     *
-     * <p>It is recommended that at least one of the arguments to be non-{@code null}, to provide
-     * useful debug information. Both argument being null may log warnings and result in suboptimal
-     * performance. Also note that the provided information will not be sent to the server.
-     *
-     * @param message if not {@code null}, will appear as the description of the CANCELLED status
-     * @param cause if not {@code null}, will appear as the cause of the CANCELLED status
-     */
-    public abstract void cancel(@Nullable String message, @Nullable Throwable cause);
+public abstract class ClientCallStreamObserver<V> extends CallStreamObserver<V> {  /**
+  * Prevent any further processing for this {@code ClientCallStreamObserver}. No further messages
+  * will be received. The server is informed of cancellations, but may not stop processing the
+  * call. Cancelling an already
+  * {@code cancel()}ed {@code ClientCallStreamObserver} has no effect.
+  *
+  * <p>No other methods on this class can be called after this method has been called.
+  *
+  * <p>It is recommended that at least one of the arguments to be non-{@code null}, to provide
+  * useful debug information. Both argument being null may log warnings and result in suboptimal
+  * performance. Also note that the provided information will not be sent to the server.
+  *
+  * @param message if not {@code null}, will appear as the description of the CANCELLED status
+  * @param cause if not {@code null}, will appear as the cause of the CANCELLED status
+  */
+  public abstract void cancel(@Nullable String message, @Nullable Throwable cause);
 }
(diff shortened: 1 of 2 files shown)


varshanbp:
Was waiting for this.

dapengzhang0:
Isn't it already sufficient to call `ClientCallStreamObserver.onError(Status.CANCELLED.asRuntimeException())`? rather than introducing a new API?

dapengzhang0:
Isn't it already sufficient to call `ClientCallStreamObserver.onError(Status.CANCELLED.asRuntimeException())` rather than introducing a new API? In case of Unary call or ServerStreaming call, if users can not call `ClientCallStreamObserver.onError`, then they can not call `ClientCallStreamObserver.cancel` either.

carl-mastrangelo:
@dapengzhang0  +1

@rmichela does this work for you?

rmichela:
Sure. It works, but its inelegant. Signaling cancellation by reporting an error feels un-intuitive and isn't a discoverable API.

I submitted this PR in response to @ejona86 's comment "It'd be a good API to add to `ClientCallStreamObserver`" in the linked issue.

ejona86:
@dapengzhang0, Calling `ClientCallStreamObserver.onError()` after `onCompleted()` (that is, after the client has half-closed) is illegal. Whereas you can cancel at any time.

stub/src/main/java/io/grpc/stub/ClientCallStreamObserver.java
```diff
@@ -24,4 +24,10 @@
  */
 @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1788")
 public abstract class ClientCallStreamObserver<V> extends CallStreamObserver<V> {
+    /**
+     * Prevent any further processing for this {@code ClientCallStreamObserver}. No further messages will be received.
+     * The server is informed of cancellations, but may not stop processing the call. Cancelling an already
+     * {@code cancel()}ed {@code ClientCallStreamObserver} has no effect.
+     */
+    public abstract void cancel();
```
ejona86: I'll point out that adding the abstract method is technically API-breaking, but cancel() is really important (*cough*) and this is `@ExperimentalApi` so I do think it is appropriate.

stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -327,6 +327,11 @@ public void request(int count) {
     public void setMessageCompression(boolean enable) {
       call.setMessageCompression(enable);
     }
+
+    @Override
+    public void cancel() {
```
ejona86: We want the `String, Throwable` arguments here. We already decided to remove the no-arg version on `ClientCall` and I see no reason to re-add it here.

stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -327,6 +327,11 @@ public void request(int count) {
...
+      call.cancel("Call cancelled by request.", new StatusException(Status.CANCELLED));
```
ejona86: FYI: this exception will become the _cause_ of the Status, so this would produce a weird Status. We basically do the equivalent of `Status.CANCELLED.withDescription(passedString).withCause(passedThrowable)`.

And _yes_, that's really surprising given server-side usage of `onError()`, but with the old zero-arg cancel() we would throw the `Throwable` away.

stub/src/main/java/io/grpc/stub/ClientCallStreamObserver.java
```diff
@@ -24,4 +24,10 @@
  */
 @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1788")
 public abstract class ClientCallStreamObserver<V> extends CallStreamObserver<V> {
+    /**
+     * Prevent any further processing for this {@code ClientCallStreamObserver}. No further messages will be received.
```
ejona86: No further messages may be sent either. If someone calls `clientCall.onNext()` after this it will throw.

dapengzhang0:
@ejona86 It is legal to call `ClientCallStreamObserver.onError()` after `ClientCall.halfClose()`. 
Which `onCompletet()` do you mean, of `requestObserver` or `responseObserver`? Does server-streaming call call `ClientCallStreamObserver.onCompleted()`?

dapengzhang0:
It seems this added API make sense for bidi-streaming to cancel after `requestObserver.onCompleted()` but does not help for the case of server-streaming.

dapengzhang0:
It seems this added API makes sense for bidi-streaming to cancel after `requestObserver.onCompleted()` but does not help for the case of server-streaming.

stub/src/main/java/io/grpc/stub/ClientCalls.java
rmichela: `String` makes sense, but why `Throwable`? Why should I have to provide an exception class when I am intentionally canceling the stream? i.e. not an exceptional situation.

stub/src/main/java/io/grpc/stub/ClientCallStreamObserver.java
ejona86: This small doc change is still necessary.

rmichela: Updated

ejona86:
Checkstyle is failing.

ejona86:
okay to test

ejona86:
okay to test

ejona86:
okay to test

ejona86:
Thanks @rmichela! Sorry for the delay toward the end there.

sankate:
Since all these classes are marked as `experimental` , What is a stable way to allow a client to gracefully cancel streaming service by signaling the server to stop sending messages?

Also, if I decide to use these classes to achieve this functionality, can I assume that I will be fine unless I upgrade `grpc-java` version?

PR #3116, opened 2017-06-20
inprocess: mark all classes final

carl-mastrangelo:


rmichela:
I'm curious...why? Is there an explicit need to prevent extension of the inprocess classes? What's the harm? Why block future use cases just because you can't envision them right now?

Requiring someone to ask to remove finality stifles innovation.

carl-mastrangelo:
@rmichela See Effective Java Item 17: "Design and document for inheritance or else prohibit it".  Also Item 16: Favor composition over inheritance


Adding final is future proofing; easy to remove but hard to add.  These classes weren't designed for inheritance, so they should be final.   If you would like to propose how one could extend these, I don't think there would be pushback to remove it.

carl-mastrangelo:
@rmichela Also: do you have code extending these?  If so, what's the use case?

rmichela:
I took another look at the `InProcess*` classes. I see why this change was made.

I've had bad experiences in the past with frameworks being too rigidly locked down to the set of approved use cases, making unenvisioned usage impossible. (:cough: Microsoft)

ejona86:
@rmichela, lacking final _strongly_ limits what changes are backward-compatible. We are _strongly_ devoted to API stability, so we want public classes `final` "by default" to reduce accidents.

I am a user that has done quite interesting things to classes that weren't marked as final, so I know some of the benefit. But I also see how it makes it basically impossible to extend later unless you prepared and were _very_ thorough ahead-of-time (like marking certain methods final).

Anyway, must of the final classes seem to be pretty useless for most people to extend, so it's not been a concern. Other cases like the stubs... are more nuanced.

PR #3118, opened 2017-06-20
core: ServerBuilder.intercept().

zhangkun83:
This adds server-wide interceptors that applies to all call handlers.
Because ServerCallHandler is acquired per request, and can be dynamicly
provided by the fallback registry, the interceptors have to be installed
on a per-request basis.  This adds a few object allocations per request,
which is acceptable.

### DIFF
--- a/core/src/main/java/io/grpc/ServerBuilder.java
+++ b/core/src/main/java/io/grpc/ServerBuilder.java
@@ -98,7 +98,7 @@ public abstract class ServerBuilder<T extends ServerBuilder<T>> {
    * @since 1.5.0
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/3117")
-  public T addInterceptor(ServerInterceptor interceptor) {
+  public T intercept(ServerInterceptor interceptor) {
     throw new UnsupportedOperationException();
   }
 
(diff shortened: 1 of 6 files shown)


dapengzhang0:
To be in the same pattern with `ManagedChannelBuilder`, should it be named as `ServerBuilder.intercept(...)`?

core/src/main/java/io/grpc/ServerBuilder.java
```diff
@@ -89,6 +89,20 @@
...
+   * Adds a {@link ServerInterceptor} that is run for all services on the server.  Interceptors
+   * added through this method always run before per-service interceptors added through {@link
+   * ServerInterceptors}.  Interceptors run in the reverse order in which they are added.
+   *
+   * @param interceptor the all-service interceptor
+   * @return this
+   * @since 1.5.0
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/3117")
+  public T addInterceptor(ServerInterceptor interceptor) {
```
ejona86: While I wouldn't disagree that `addInterceptor()` may be a better name, it is probably better to copy `ManagedChannelBuilder`'s `intercept()` name.

carl-mastrangelo:
@zhangkun83  Fyi the commit message should be something slightly more meaty.  Reading the commitlog doesn't tell that this change did.

Issue #3105, opened 2017-06-16
Reviewing gRPC unit test framework


dapengzhang0:
There are some issues and limitation to the current unit test framework:

- It's hard to test a user-defined generic interceptor
- People are abusing `NoopClientCall`, using it as an alternative of `mock(ClientCall.class)`, which is as bad as a mock.
- `TestMethodDescriptors.noopMarshaller` does not gain much benefit. It violates the interface contract, it's equivalent to a `mock(Marshaller.class)`, and it does not work in InProcess transport or interop test.

dapengzhang0:
Quoting from another thread @ejona86: 
> @dapengzhang0 and I spoke offline. We're thinking all that's really necessary for most interceptor users is a simple test service. We're planning on adding a .proto with such a test service, for users to use in their own tests. It will be _very_ simple though, with only 1 or 4 methods (because streaming). Users that need more should create their own test.proto.
> 
> This does mean interceptor _tests_ would depend on proto, although they wouldn't need to run protoc. We considered having a simple MethodDescriptor instead, but found it still a bit too annoying because you have to create your own ServerServiceDefinition. Not the end of the world, but the proto-based solution seems easier for our users and we don't except a superficial _test_ dependency on protobuf to be burdensome. If we're wrong, we can add it in the future.

PR #3121, opened 2017-06-21
testing: move NoopClientCall & NoopServerCall to internal

dapengzhang0:
one step for #3105

Edited:
* It is very rare that users have to stub a ClientCall given that we have InProcess channel. It's even rare that they stub a ClientCall in a correct way. Saving user's a few lines of code for the rare rare case does not gain much benefit.
* Everybody who used or wants to use `mock(ClientCall.class)`, that's a lot of people, will switch to `NoopClientCall`, which makes the effort of marking `ClientCall` with `@DoNotMock` useless.


carl-mastrangelo:
I think these classes were for people to extend, instead of mocking them.  Since CC and SC are abstract, and people just want to stub out one method, that was the intended use.  I dont think they should be internal, they are supposed to be used.

dapengzhang0:
Basically `ClientCall` is not something users should provide, it is always produced by grpc core library (possibly wrapped with interceptors). People are abusing it, using it as an alternative of `mock(ClientCall.class)` which is as bad as a mock. If some people (external) really want to implement their own `ClientCall` in test, which is not recommended, they can just implement it from scratch, extending from the noop does not provide so much benefits compared to the harm of the abuse.

dapengzhang0:
Basically `ClientCall` is not something users should provide, it is always produced by grpc core library (possibly wrapped with interceptors). People are abusing it, using it as an alternative of `mock(ClientCall.class)` which is as bad as a mock. If some people (external) really want to implement their own `ClientCall` in test, which is not recommended, they can just implement it from scratch, extending from the noop does not provide so much benefit compared to the harm of the abuse.

dapengzhang0:
Edited description with rationale.

rmichela:
I can see the `NoOpServerCall` class being useful starting point for dynamically building a server using the fallback handler registry, but it is also trivial to replicate.

ejona86:
@rmichela, I'm also aware of some use-cases for interceptors. But in _testing_, it does seem to just be asking for trouble. For those other legit use-cases, I'm willing to let users replicate it today, but maybe we can figure out a solution that won't encourage misuse in tests. (Hmmm... like maybe if it was `final`...)

PR #3122, opened 2017-06-22
netty: fix race condition for listeners attaching to connect future

### LINKED ISSUE #1877: {Netty,OkHttp}TransportTest.serverNotListening is flaky
https://grpc-testing.appspot.com/job/gRPC-Java-PR-Windows/1439/console

```
io.grpc.okhttp.OkHttpTransportTest > serverNotListening FAILED
    Wanted but not invoked:
    listener.transportTerminated();
    -> at io.grpc.internal.testing.AbstractTransportTest.serverNotListening(AbstractTransportTest.java:178)
    Actually, there were zero interactions with this mock.
```

zpencer:
The listener to the connect future depends on the channel pipeline being intact. But the way it is attached allows the connect attempt to fail, and have the entire pipeline being torn down by netty before the `.addListener` actually runs. The result is that the listener will be attached to an already completed future, and the logic will be applied to an empty pipeline.

The fundamental problem is that there are two threads, the grpc thread, and the netty thread. This PR moves the listener attaching code into the netty thread, guaranteeing the listener is attached before any connection is made. It makes more sense for the code to live inside `AbstractBufferingHandler`, since handlers are generally free to swallow exceptions (the alternative is to make `NettyClientHandler` forward exceptions up the pipeline from itself). `AbstractBufferingHandler` needs the special guarantees, so it will be the one with special code.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
+++ b/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
@@ -27,6 +27,8 @@ import io.grpc.Internal;
 import io.grpc.Status;
 import io.grpc.internal.GrpcUtil;
 import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelFuture;
+import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerAdapter;
 import io.netty.channel.ChannelHandlerContext;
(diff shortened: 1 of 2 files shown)


netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -446,6 +448,25 @@ public void channelRegistered(ChannelHandlerContext ctx) throws Exception {
...
 
     /**
+     * Do not rely on channel handlers to propagate exceptions to us.
+     * {@link NettyClientHandler} is an example of a class that does not propagate exceptions.
+     * Add a listener to the connect future directly and do appropriate error handling.
+     */
+    @Override
+    public void connect(final ChannelHandlerContext ctx, SocketAddress remoteAddress,
+        SocketAddress localAddress, ChannelPromise promise) throws Exception {
+      promise.addListener(new ChannelFutureListener() {
```
ejona86: Add the promise after calling `connect()`. Always add the promise afterward for sane ordering.

zpencer:
Turns out this still fails if it is run 10,000s of times, whereas before it would fail reliably with even 100 iterations without this change. Will continue digging into this.

zpencer:
retest this please

PR #3130, opened 2017-06-23
all: migrate .proto files from testing-proto to interop-testing

dapengzhang0:
and provide a simple protobuf service for test in testing-proto instead.

Second step for #3105


testing-proto/src/main/proto/io/grpc/testing/protobuf/simpleservice.proto
```diff
@@ -0,0 +1,47 @@
...
+// limitations under the License.
+// An integration test service that covers all the method signature permutations
+// of unary/streaming requests/responses.
+
+syntax = "proto3";
+
+package grpc.testing;
+
+option java_package = "io.grpc.testing.protobuf";
+option java_outer_classname = "SimpleServiceProto";
```
ejona86: We tend to use `option java_multiple_files = true;` as well.

testing-proto/src/main/proto/io/grpc/testing/protobuf/simpleservice.proto
```diff
@@ -0,0 +1,47 @@
...
+
+  // Simple server-to-client streaming RPC.
+  rpc ServerStreamingRpc (SimpleRequest) returns (stream SimpleResponse) {}
+
+  // Simple bidirectional streaming RPC.
+  rpc BidiStreamingRpc (stream SimpleRequest) returns (stream SimpleResponse) {}
+}
+
+// A simple request message type for test.
+message SimpleRequest {
```
ejona86: I do think we'll eventually want to add a string or similar to the request/response, but this is fine at this point.

Issue #2514, opened 2016-12-14
Periodical refresh of DnsNameResolver


zhangkun83:
Please answer these questions before submitting your issue.

Original discussion:
https://groups.google.com/forum/#!topic/grpc-io/wxgLgjzkR30

Currently DnsNameResolver soles relies on `refresh()` being called to re-resolve. `refresh()` currently is called only when there is a connection failure, but there are cases where refreshing is needed without connection failure, e.g., rolling deploys, or simply adding a server.  We could introduce periodic refresh to DnsNameResolver.

carl-mastrangelo:
Briefly talking with @a11r, I believe the correct thing to do is re-resolve the name on GOAWAY in addition to connection failure.

zhangkun83:
We have already be doing so. What about the case of adding new servers while keeping the old servers?

mightyguava:
@carl-mastrangelo that sounds like it would solve the bug in reconnect code you referenced in the Google Group discussion, but I'm not seeing how that solves making sure that new servers added to the name get picked up. Right now, if I add new servers to a DNS name, I think they would never get picked up by the clients until one of its connected servers goes down.

While re-resolving on connection failure and GOAWAY does make sure we have HA, it doesn't allow us to fully utilize resources. We only run 2-3 replicas per service right now (hopefully will scale up soon...), so being at N-1 would mean running at 50%-67% capacity.

carl-mastrangelo:
DNS is not really designed to handle your use case.  Adding a method to force DNS refresh to pick up new servers would just be hacking around DNS shortcomings.

The proper solution would be to use a watchable name service like etc.d or something like that.

mightyguava:
We actually did go down that route and implemented resolvers for our internal name service in Go and Java, but we pulled back because it seems the current method of adding resolvers in core (we need it for Python) is quite challenging, requires us to fork core, and I'm not sure how far core or Python is from supporting adding your own resolver, if that's intended at all.

Got any timelines or intended direction on core or Python's side?

markdroth:
C-core does support adding third-party resolvers via the `grpc_register_resolver_type()` function, which can be called from a plugin function, which is registered with `grpc_register_plugin()`.

mightyguava:
@markdroth is there documentation on how to add plugins to core? I thought I had to fork and statically build core to get another resolver plugged in. I'm fairly poor at C so I could totally be missing the obvious.

markdroth:
[...]

PR #3133, opened 2017-06-25
Periodical Refresh of DnsNameResolver

sdorazio:
This pull request is to address issue #2514. The implementation uses the cache time configured with Java's [networkaddress.cache.ttl](http://docs.oracle.com/javase/6/docs/technotes/guides/net/properties.html) to periodically refresh the resolved addresses.

### DIFF
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -20,6 +20,8 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+import com.google.common.primitives.Longs;
 import io.grpc.Attributes;
 import io.grpc.EquivalentAddressGroup;
 import io.grpc.NameResolver;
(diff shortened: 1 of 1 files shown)


sdorazio:
I signed it!

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -120,6 +124,13 @@ public final synchronized void start(Listener listener) {
     executor = SharedResourceHolder.get(executorResource);
     this.listener = Preconditions.checkNotNull(listener, "listener");
     resolve();
+
+    Long addressTtl = Longs.tryParse(Strings.nullToEmpty(
+          java.security.Security.getProperty("networkaddress.cache.ttl")));
```
ejona86: Hmm... so if it is the default then we don't poll? The default varies, but is 30 seconds in many cases which means that polling would be useful (although I'm not saying polling every 30 seconds is what we want).

ejona86:
okay to test

ejona86:
@zhangkun83, it seems this slipped through the cracks. Who should review this?

zhangkun83:
I don't feel very confident on figuring out what the right behavior should be. Maybe you or @carl-mastrangelo ?

ejona86:
@sdorazio, I want to mention that this is actually a cross-language issue we've had trouble coming up with a design. I'd have to look at the DNS landscape a bit more, but I do have a concern that polling could always return different addresses (which would trigger unnecessary reconnects as part of the DNS server itself providing round-robin). It seems discussion should continue in #2514 to arrive at an agreed approach before reviewing an implementation.

ejona86:
Closing to let discussion in #2514 complete. If this is the route we want to go, we can reopen this PR.

PR #3136, opened 2017-06-26
benchmark: add missing server side workloads

zpencer:
Add missing workloads for missing streaming workloads. Client workloads coming on another PR.

Also, refactor away some redundant code.

### DIFF
--- a/benchmarks/src/main/java/io/grpc/benchmarks/qps/AsyncServer.java
+++ b/benchmarks/src/main/java/io/grpc/benchmarks/qps/AsyncServer.java
@@ -46,11 +46,13 @@ import java.util.concurrent.ForkJoinWorkerThread;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.logging.Logger;
 
 /**
  * QPS server using the non-blocking API.
  */
 public class AsyncServer {
+  private static final Logger log = Logger.getLogger(AsyncServer.class.getName());
 
   /**
    * checkstyle complains if there is no javadoc comment here.
@@ -249,11 +251,13 @@ public class AsyncServer {
         @Override
         public void onError(Throwable t) {
           // other side closed with non OK
+          responseObserver.onError(t);
         }
 
         @Override
         public void onCompleted() {
           // other side closed with OK
+          responseObserver.onCompleted();
         }
       };
     }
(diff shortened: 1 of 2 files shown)


benchmarks/src/main/java/io/grpc/benchmarks/qps/AsyncServer.java
```diff
@@ -184,36 +191,194 @@ public ForkJoinWorkerThread newThread(ForkJoinPool pool) {
...
+      this.shutdown.set(true);
+    }
+
+    /**
+     * Helper method to avoid polling the shutdown volatile in a tight loop. The server shutdown
+     * operation only happens once at the end of the test scenario, so there's no need to
+     * constantly check it.
+     */
+    private boolean shouldShutdown() {
+      Long latestPoll = shutdownPollNanos.get();
```
carl-mastrangelo: I don't think this code is necessary.  Polling a volatile is effectively free on x86.  Also the branch predictor is goign to get very good at guessing that shutdown is usually false, and will make it a noop.

ejona86: +1, this seems unnecessary. Also, `System.nanoTime()` involves some `volatile` operations itself and has other overheads.

zpencer:
Any remaining issues on this one?

benchmarks/src/main/java/io/grpc/benchmarks/qps/AsyncServer.java
```diff
@@ -184,36 +191,173 @@ public ForkJoinWorkerThread newThread(ForkJoinPool pool) {
...
+        @Override
+        public void onError(Throwable t) {
+          // other side closed with non OK
+          responseObserver.onError(t);
+        }
+
+        @Override
+        public void onCompleted() {
+          if (lastSeen != null) {
+            responseObserver.onNext(Utils.makeResponse(lastSeen));
```
carl-mastrangelo: If shutdown is true, should this be invoked?

zpencer: This would be the case where a shutdown races with an `onCompleted`, and I think either is fine here. The client should be streaming requests to us and `onNext` would have detected the shutdown. I'll change that case to an `onError` to be more clear.

benchmarks/src/main/java/io/grpc/benchmarks/qps/AsyncServer.java
```diff
@@ -184,36 +191,173 @@ public ForkJoinWorkerThread newThread(ForkJoinPool pool) {
...
+            return;
+          }
+          responseObserver.onNext(Utils.makeResponse(value));
         }
 
         @Override
         public void onError(Throwable t) {
-          System.out.println("Encountered an error in streamingCall");
-          t.printStackTrace();
+          // other side closed with non OK
```
zpencer: We may want to be a bit quieter, if the other side closes using `onError` when the load generator receives a shutdown signal.

zpencer:
@carl-mastrangelo

PR #3137, opened 2017-06-26
netty: client transport needs to flush fence operation

zpencer:
tldr: the NOOP_MESSAGE fence operation is racey; we are not flushing
the channel and the application thread is not waiting for the
operation to complete.

Details:

On windows, the serverNotListening test sometimes still fails with this
problem:

https://gist.github.com/zpencer/077da915c70e9bd75624932cfa21aa9d

The test fails on this assertion:
```
verify(mockClientTransportListener, timeout(TIMEOUT_MS)).transportTerminated();
```

The call being verified should have been called when the NOOP_MESSAGE
fails. Because the call to `NettyClientTransport.start` returns a null
for the success case, the caller immediately starts the verification
with a 1s timeout. We should return a Runnable that tells the caller
to block until the fence operation is done. We also need to flush the
channel to ensure it completes at all.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyClientTransport.java
+++ b/netty/src/main/java/io/grpc/netty/NettyClientTransport.java
@@ -41,7 +41,6 @@ import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelOption;
-import io.netty.channel.ChannelPromise;
 import io.netty.channel.EventLoop;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.socket.nio.NioSocketChannel;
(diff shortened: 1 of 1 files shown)


netty/src/main/java/io/grpc/netty/NettyClientTransport.java
```diff
@@ -248,7 +250,16 @@ public void operationComplete(ChannelFuture future) throws Exception {
       keepAliveManager.onTransportStarted();
     }
 
-    return null;
+    return new Runnable() {
+      @Override
+      public void run() {
+        try {
+          fenceOperation.await();
```
zpencer: This is still racey: we will need to block until the *listener* is done, not the future that the listener is waiting on. Update coming soon.

zpencer: done

netty/src/main/java/io/grpc/netty/NettyClientTransport.java
```diff
@@ -248,7 +252,16 @@ public void operationComplete(ChannelFuture future) throws Exception {
       keepAliveManager.onTransportStarted();
     }
 
-    return null;
+    return new Runnable() {
+      @Override
+      public void run() {
+        try {
+          listenerDone.sync();
```
ejona86: We can't block like this without risking deadlock. Let's look for an alternative.

zpencer: I suspect adding only the "flush" might be enough to make this test non flakey, given our 1second verification timeout. Let's try that.

PR #3140, opened 2017-06-26
core: update javadocs regarding ClientCall mocks

### LINKED ISSUE #3124: Update ClientCall and InProcessServer javadoc wrt mocking
The `@DoNotMock` annotation should really refer to `InProcessServerBuilder` instead of `InProcessTransport`, and `InProcessServer` should have a good example in its javadoc.

zpencer:
Change the `@DoNotMock` message to refer to `InProcessServerBuilder`, and include some usage examples that users may find useful.

fixes #3124

### DIFF
--- a/core/src/main/java/io/grpc/ClientCall.java
+++ b/core/src/main/java/io/grpc/ClientCall.java
@@ -97,7 +97,7 @@ import javax.annotation.Nullable;
  * @param <ReqT> type of message sent one or more times to the server.
  * @param <RespT> type of message received one or more times from the server.
  */
-@DoNotMock("Use InProcessTransport and make a fake server instead")
+@DoNotMock("Use InProcessServerBuilder and make a full featured in process server instead")
 public abstract class ClientCall<ReqT, RespT> {
   /**
    * Callbacks for receiving metadata, response messages and completion status from the server.
--- a/core/src/main/java/io/grpc/ClientCall.java
+++ b/core/src/main/java/io/grpc/ClientCall.java
@@ -97,7 +97,7 @@ import javax.annotation.Nullable;
  * @param <ReqT> type of message sent one or more times to the server.
  * @param <RespT> type of message received one or more times from the server.
  */
-@DoNotMock("Use InProcessServerBuilder and make a full featured in process server instead")
+@DoNotMock("Use InProcessServerBuilder and make a test server instead")
 public abstract class ClientCall<ReqT, RespT> {
   /**
    * Callbacks for receiving metadata, response messages and completion status from the server.
--- a/core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
@@ -43,7 +43,7 @@ import java.util.List;
  *   ExecutorService serverExecutor = Executors.newCachedThreadPool();
  *   Server server = InProcessServerBuilder.forName("unique-name")
  *       .executor(serverExecutor)
- *       .addService(service)
+ *       .fallbackHandlerRegistry(serviceRegistry)
  *       .build().start();
  *   ExecutorService clientExecutor = Executors.newCachedThreadPool();
  *   ManagedChannel channel = InProcessChannelBuilder.forName("unique-name")
--- a/core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
@@ -28,6 +28,47 @@ import java.util.List;
  * its name.
  *
  * <p>The server is intended to be fully-featured, high performance, and useful in testing.
+ *
+ * <h3>Using JUnit TestRule</h3>
+ * The class "GrpcServerRule" (from the "grpc-java/testing") is a JUnit TestRule that
+ * creates a {@link InProcessServer} and a {@link io.grpc.ManagedChannel ManagedChannel}. This
+ * test rule contains the boiler plate code shown below. The classes "HelloWorldServerTest" and
+ * "HelloWorldClientTest" (from "grpc-java/examples") demonstrate basic usage.
+ *
+ * <h3>Usage example</h3>
+ * <h4>Server and client channel setup</h4>
+ * <pre>
+ *   MutableHandlerRegistry serviceRegistry = new MutableHandlerRegistry();
+ *   addServicesHelper(serviceRegistry); // your code here
+ *   ExecutorService serverExecutor = Executors.newCachedThreadPool();
+ *   Server server = InProcessServerBuilder.forName("unique-name")
+ *       .executor(serverExecutor)
+ *       .addService(service)
+ *       .build().start();
+ *   ExecutorService clientExecutor = Executors.newCachedThreadPool();
+ *   ManagedChannel channel = InProcessChannelBuilder.forName("unique-name")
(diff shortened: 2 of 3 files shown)


core/src/main/java/io/grpc/ClientCall.java
```diff
@@ -97,7 +97,7 @@
  * @param <ReqT> type of message sent one or more times to the server.
  * @param <RespT> type of message received one or more times from the server.
  */
-@DoNotMock("Use InProcessTransport and make a fake server instead")
+@DoNotMock("Use InProcessServerBuilder and make a full featured in process server instead")
```
ejona86: I think the intention was "fake service", if we're being precise.

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
```diff
@@ -28,6 +28,47 @@
...
+ * The class "GrpcServerRule" (from the "grpc-java/testing") is a JUnit TestRule that
+ * creates a {@link InProcessServer} and a {@link io.grpc.ManagedChannel ManagedChannel}. This
+ * test rule contains the boiler plate code shown below. The classes "HelloWorldServerTest" and
+ * "HelloWorldClientTest" (from "grpc-java/examples") demonstrate basic usage.
+ *
+ * <h3>Usage example</h3>
+ * <h4>Server and client channel setup</h4>
+ * <pre>
+ *   MutableHandlerRegistry serviceRegistry = new MutableHandlerRegistry();
+ *   addServicesHelper(serviceRegistry); // your code here
```
ejona86: Move this line some point later? As it is it doesn't need the `MutableHandlerRegistry`.

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
```diff
@@ -28,6 +28,47 @@
  * its name.
  *
  * <p>The server is intended to be fully-featured, high performance, and useful in testing.
+ *
+ * <h3>Using JUnit TestRule</h3>
+ * The class "GrpcServerRule" (from the "grpc-java/testing") is a JUnit TestRule that
+ * creates a {@link InProcessServer} and a {@link io.grpc.ManagedChannel ManagedChannel}. This
+ * test rule contains the boiler plate code shown below. The classes "HelloWorldServerTest" and
```
ejona86: s/boiler plate/boilerplate/

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
```diff
@@ -28,6 +28,47 @@
...
+ * <h4>Server and client channel setup</h4>
+ * <pre>
+ *   MutableHandlerRegistry serviceRegistry = new MutableHandlerRegistry();
+ *   addServicesHelper(serviceRegistry); // your code here
+ *   ExecutorService serverExecutor = Executors.newCachedThreadPool();
+ *   Server server = InProcessServerBuilder.forName("unique-name")
+ *       .executor(serverExecutor)
+ *       .fallbackHandlerRegistry(serviceRegistry)
+ *       .build().start();
+ *   ExecutorService clientExecutor = Executors.newCachedThreadPool();
```
ejona86: Reuse the `ExecutorService` for both client and server-side? Really, for testing, I'd probably not bother setting the executors (or I'd use directExecutor()).

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
```diff
@@ -28,6 +28,47 @@
...
+ *   ManagedChannel channel = InProcessChannelBuilder.forName("unique-name")
+ *       .executor(clientExecutor)
+ *       .build();
+ * </pre>
+ *
+ * <h4>Client call usage</h4>
+ * <pre>
+ *   MethodDescriptor&lt;Integer, Integer&gt; methodDescriptor =
```
ejona86: For this example, I probably wouldn't bother showing this MethodDescriptor-based code. I'd probably just say something like "it's a channel to use like normal" and then show the `newBlockingStub()` example line.

core/src/main/java/io/grpc/ClientCall.java
carl-mastrangelo: Precision is the wrong thing here.  Naming the exact class they use is a better usability story.  People who run into this annotation probably don't know what they should do.  

Even better would be pointing to an example class, but the change as is is better than previously.

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
carl-mastrangelo: setting executors is conditioning for real production services.  Showing best practices here is a good thing.

core/src/main/java/io/grpc/ClientCall.java
ejona86: I was fine with the `InProcessServerBuilder` reference. I was talking about the "full featured in process".

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
ejona86: It's a test, not a real production service. It is a Good Idea™ to use directExecutor() in tests when possible; it's not generally good to do that in production.

And people probably aren't going to shutdown the executors, so it is sort of more harmful to specify them in addition to being additional, unnecessary boilerplate.

core/src/main/java/io/grpc/ClientCall.java
zpencer: The "full featured" wording is taken from the javadoc of `InProcessServerBuilder`, as I didn't want to use the word "fake" which feels contradictory. I'll use a more neutral word to avoid this altogether.

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
zpencer: done

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
zpencer: switched to `directExecutor()` with a comment regarding unit tests

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
zpencer: done

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
zpencer: Changed to use `addService()` instead, to avoid `MutableHandlerRegistry`

core/src/main/java/io/grpc/ClientCall.java
dapengzhang0: "full featured" isn't precise either. The `InProcessServer` does not feature load-balancing for example. Can we just say `Use InProcessServerBuilder and stub an in-process server instead`?

ejona86: The "fake server" was not referring to in-process transport. It was referring to the code service code registered via `builder.addService(new ServiceImplBase{} /* this part here */)`. That's why I said "fake service" was being more precise.

zpencer:
@ejona86 @dapengzhang0 any more outstanding issues with the PR?

PR #3145, opened 2017-06-27
Support for deframing in the application thread

ericgribkoff:
This pull request adds an option in `AbstractStream` (currently always disabled) to perform deframing in the application thread. The change is divided into three commits:

1. Introduces the `StreamListener.MessageProducer` interface and replaces `StreamListener.messageRead(InputStream)` with `StreamListener.messagesAvailable(MessageProducer)`.
2. Removes `MessageDeframer`'s tracking of the end-of-stream bit and checking for delivery stalled, and replaces them with a close request/callback mechanism. This is (arguably) simpler and better fits the model necessary for asynchronous client-thread deframing.
3. Adds the option for client-thread deframing, which places calls from the transport to the deframer inside of a `MessageProducer` that executes on the application thread. `MessageDeframer` remains entirely single-threaded, with the exception of a volatile boolean flag to signal an immediate close request.

This PR supersedes https://github.com/grpc/grpc-java/pull/2927 after offline discussion with @ejona86.

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractClientStream.java
+++ b/core/src/main/java/io/grpc/internal/AbstractClientStream.java
@@ -184,7 +184,8 @@ public abstract class AbstractClientStream extends AbstractStream
     private ClientStreamListener listener;
     private DecompressorRegistry decompressorRegistry = DecompressorRegistry.getDefaultInstance();
 
-    private Runnable deliveryStalledTask;
+    private boolean deframerClosed = false;
+    private Runnable deframerClosedTask;
 
     /**
      * Whether the stream is closed from the transport's perspective. This can differ from {@link
@@ -210,16 +211,12 @@ public abstract class AbstractClientStream extends AbstractStream
     }
 
     @Override
-    public final void deliveryStalled() {
-      if (deliveryStalledTask != null) {
-        deliveryStalledTask.run();
-        deliveryStalledTask = null;
+    public void deframerClosed(boolean hasPartialMessageIgnored) {
+      deframerClosed = true;
+      if (deframerClosedTask != null) {
+        deframerClosedTask.run();
+        deframerClosedTask = null;
       }
-    }
-
-    @Override
-    public final void endOfStream() {
-      deliveryStalled();
     }
 
     @Override
@@ -266,7 +263,7 @@ public abstract class AbstractClientStream extends AbstractStream
         }
 
         needToCloseFrame = false;
-        deframe(frame, false);
+        deframe(frame);
       } finally {
         if (needToCloseFrame) {
           frame.close();
@@ -314,18 +311,18 @@ public abstract class AbstractClientStream extends AbstractStream
       statusReported = true;
       onStreamDeallocated();
 
-      // If not stopping delivery, then we must wait until the deframer is stalled (i.e., it has no
-      // complete messages to deliver).
-      if (stopDelivery || isDeframerStalled()) {
-        deliveryStalledTask = null;
+      if (deframerClosed) {
+        deframerClosedTask = null;
         closeListener(status, trailers);
       } else {
-        deliveryStalledTask = new Runnable() {
-          @Override
-          public void run() {
-            closeListener(status, trailers);
-          }
-        };
+        deframerClosedTask =
+            new Runnable() {
+              @Override
+              public void run() {
+                closeListener(status, trailers);
+              }
+            };
+        closeDeframer(stopDelivery);
       }
     }
 
@@ -337,7 +334,6 @@ public abstract class AbstractClientStream extends AbstractStream
     private void closeListener(Status status, Metadata trailers) {
       if (!listenerClosed) {
         listenerClosed = true;
-        closeDeframer();
         statsTraceCtx.streamClosed(status);
         listener().closed(status, trailers);
       }
--- a/core/src/main/java/io/grpc/internal/AbstractServerStream.java
+++ b/core/src/main/java/io/grpc/internal/AbstractServerStream.java
@@ -213,9 +213,10 @@ public abstract class AbstractServerStream extends AbstractStream
         if (!immediateCloseRequested && hasPartialMessage) {
           // We've received the entire stream and have data available but we don't have
           // enough to read the next frame ... this is bad.
-          deframeFailed(Status.INTERNAL
-              .withDescription("Encountered end-of-stream mid-frame")
-              .asRuntimeException());
+          deframeFailed(
+              Status.INTERNAL
+                  .withDescription("Encountered end-of-stream mid-frame")
+                  .asRuntimeException());
           deframerClosedTask = null;
           return;
         }
--- a/core/src/main/java/io/grpc/internal/AbstractServerStream.java
+++ b/core/src/main/java/io/grpc/internal/AbstractServerStream.java
@@ -182,6 +182,11 @@ public abstract class AbstractServerStream extends AbstractStream
     private ServerStreamListener listener;
     private final StatsTraceContext statsTraceCtx;
 
+    private boolean endOfStream = false;
+    private boolean deframerClosed = false;
+    private boolean immediateCloseRequested = false;
+    private Runnable deframerClosedTask;
+
     protected TransportState(int maxMessageSize, StatsTraceContext statsTraceCtx) {
       super(maxMessageSize, statsTraceCtx);
       this.statsTraceCtx = Preconditions.checkNotNull(statsTraceCtx, "statsTraceCtx");
@@ -202,12 +207,24 @@ public abstract class AbstractServerStream extends AbstractStream
     }
 
     @Override
-    public void deliveryStalled() {}
-
-    @Override
-    public void endOfStream() {
-      closeDeframer();
-      listener().halfClosed();
+    public void deframerClosed(boolean hasPartialMessage) {
+      deframerClosed = true;
+      if (endOfStream) {
+        if (!immediateCloseRequested && hasPartialMessage) {
+          // We've received the entire stream and have data available but we don't have
+          // enough to read the next frame ... this is bad.
+          deframeFailed(Status.INTERNAL
+              .withDescription("Encountered end-of-stream mid-frame")
+              .asRuntimeException());
+          deframerClosedTask = null;
+          return;
+        }
+        listener.halfClosed();
+      }
+      if (deframerClosedTask != null) {
+        deframerClosedTask.run();
+        deframerClosedTask = null;
+      }
     }
 
     @Override
@@ -224,8 +241,13 @@ public abstract class AbstractServerStream extends AbstractStream
      * @param endOfStream {@code true} if no more data will be received on the stream.
      */
     public void inboundDataReceived(ReadableBuffer frame, boolean endOfStream) {
+      Preconditions.checkState(!this.endOfStream, "Past end of stream");
       // Deframe the message. If a failure occurs, deframeFailed will be called.
-      deframe(frame, endOfStream);
+      deframe(frame);
+      if (endOfStream) {
+        this.endOfStream = true;
+        closeDeframer(false);
+      }
     }
 
     /**
@@ -238,9 +260,22 @@ public abstract class AbstractServerStream extends AbstractStream
      *
      * @param status the error status. Must not be {@link Status#OK}.
      */
-    public final void transportReportStatus(Status status) {
+    public final void transportReportStatus(final Status status) {
       Preconditions.checkArgument(!status.isOk(), "status must not be OK");
-      closeListener(status);
+      if (deframerClosed) {
+        deframerClosedTask = null;
+        closeListener(status);
+      } else {
+        deframerClosedTask =
+            new Runnable() {
+              @Override
+              public void run() {
+                closeListener(status);
+              }
+            };
+        immediateCloseRequested = true;
+        closeDeframer(true);
+      }
     }
 
     /**
@@ -249,7 +284,20 @@ public abstract class AbstractServerStream extends AbstractStream
      * #transportReportStatus}.
      */
     public void complete() {
-      closeListener(Status.OK);
+      if (deframerClosed) {
+        deframerClosedTask = null;
+        closeListener(Status.OK);
+      } else {
+        deframerClosedTask =
+            new Runnable() {
+              @Override
+              public void run() {
+                closeListener(Status.OK);
+              }
+            };
+        immediateCloseRequested = true;
+        closeDeframer(true);
+      }
     }
 
     /**
@@ -263,7 +311,6 @@ public abstract class AbstractServerStream extends AbstractStream
         }
         listenerClosed = true;
         onStreamDeallocated();
-        closeDeframer();
         listener().closed(newStatus);
       }
     }
--- a/core/src/main/java/io/grpc/internal/AbstractStream.java
+++ b/core/src/main/java/io/grpc/internal/AbstractStream.java
@@ -24,6 +24,7 @@ import io.grpc.Codec;
 import io.grpc.Compressor;
 import io.grpc.Decompressor;
 import java.io.InputStream;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.GuardedBy;
 
 /**
@@ -142,7 +143,7 @@ public abstract class AbstractStream implements Stream {
 
     @Override
     public void messageRead(InputStream is) {
-      listener().messageRead(is);
+      listener().messagesAvailable(new SingleMessageProducer(is));
     }
 
     /**
@@ -285,6 +286,22 @@ public abstract class AbstractStream implements Stream {
       }
       if (doNotify) {
         listener().onReady();
+      }
+    }
+
+    public static class SingleMessageProducer implements StreamListener.MessageProducer {
+      private InputStream message;
+
+      private SingleMessageProducer(InputStream message) {
+        this.message = message;
+      }
+
+      @Nullable
+      @Override
+      public InputStream next() {
+        InputStream messageToReturn = message;
+        message = null;
+        return messageToReturn;
       }
     }
   }
--- a/core/src/main/java/io/grpc/internal/AbstractStream.java
+++ b/core/src/main/java/io/grpc/internal/AbstractStream.java
@@ -24,7 +24,6 @@ import io.grpc.Codec;
 import io.grpc.Compressor;
 import io.grpc.Decompressor;
 import java.io.InputStream;
-import javax.annotation.Nullable;
 import javax.annotation.concurrent.GuardedBy;
 
 /**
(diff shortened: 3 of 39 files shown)


netty/src/test/java/io/grpc/netty/NettyClientTransportTest.java
```diff
@@ -571,14 +573,16 @@ public void onReady() {
       this.method = method;
       this.headers = headers;
       stream.writeHeaders(new Metadata());
-      stream.request(1);
```
ejona86: Why did this move?

netty/src/test/java/io/grpc/netty/NettyStreamTestBase.java
```diff
@@ -129,11 +133,12 @@ public void inboundMessageShouldCallListener() throws Exception {
       ((NettyClientStream) stream).transportState()
           .transportDataReceived(messageFrame(MESSAGE), false);
     }
-    ArgumentCaptor<InputStream> captor = ArgumentCaptor.forClass(InputStream.class);
-    verify(listener()).messageRead(captor.capture());
+
+    InputStream message = listenerMessageQueue.poll();
```
ejona86: This being populated by the extending class is really subtle. I think we've instead made an abstract method here (say, `listenerMessageQueue()`) that is required to be implemented by extending classes.

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
```diff
@@ -177,6 +178,22 @@ public void setUp() {
     when(serverStreamTracerFactory.newServerStreamTracer(anyString(), any(Metadata.class)))
         .thenReturn(serverStreamTracer);
     callOptions = CallOptions.DEFAULT.withStreamTracerFactory(clientStreamTracerFactory);
+
+    doAnswer(
```
ejona86: Eww... each test needing to do something different is pretty ugly. It could be made better if you made the Answers non-anonymous. But I do wonder if there is something better we can do. Maybe something with Captors... Dunno.

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
```diff
@@ -523,6 +540,18 @@ public void newStream_duringShutdown() throws Exception {
       inOrder.verify(clientStreamTracerFactory).newClientStreamTracer(any(Metadata.class));
     }
     ClientStreamListener mockClientStreamListener2 = mock(ClientStreamListener.class);
+    doAnswer(
+          new Answer<Void>() {
+            @Override
+            public Void answer(InvocationOnMock invocation) throws Throwable {
+              StreamListener.MessageProducer producer =
+                  (StreamListener.MessageProducer) invocation.getArguments()[0];
+              while (producer.next() != null) {}
```
ejona86: I'd really hope we'd close the messages properly everywhere in AbstractTransportTest.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -466,22 +466,25 @@ public final void runInContext() {
...
-              observer.onMessage(method.parseResponse(message));
-            } finally {
-              message.close();
+            while ((message = producer.next()) != null) {
+              if (closed) {
+                return;
+              }
+              try {
+                observer.onMessage(method.parseResponse(message));
+              } finally {
```
ericgribkoff: Can you elaborate on what you're looking for here? This can fail to close some messages, if `producer.next()` throws an exception, but not sure what we should do about that here since the messages might not exist or have been deframed at that point. When the producer is a wrapper around a single message, as in this commit, isn't the try/finally here exactly the same as with the current code?

netty/src/test/java/io/grpc/netty/NettyClientTransportTest.java
ericgribkoff: When deframing in the client thread (later commit, so this could be moved out of this commit entirely), calls to `request` trigger a call to `messagesAvailable` on the listener. Calling `request(1)` here in the `EchoServerStreamListener` constructor causes a null pointer exception, because the stream doesn't have a listener set until line 615, after the `EchoServerStreamListener` is created and set on the stream. This can be avoided by doing a null check in `AbstractStream`, but it seemed cleaner just not to invoke these methods when the listener isn't set.

But I realize now that this either needs to be documented on the `{Client,Server,}Stream` API and/or still do the a null check in `AbstractStream.messagesAvailable`

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
ericgribkoff: Done

core/src/test/java/io/grpc/internal/ClientCallImplTest.java
```diff
@@ -820,5 +823,21 @@ void release() {
       }
     }
   }
+
+  private static class SingleMessageProducer implements StreamListener.MessageProducer {
```
ericgribkoff: Was waiting to make sure this was the right approach. Done.

core/src/test/java/io/grpc/internal/AbstractServerStreamTest.java
```diff
@@ -68,6 +74,22 @@ public WritableBuffer allocate(int capacityHint) {
   @Test
   public void frameShouldBeIgnoredAfterDeframerClosed() {
     ServerStreamListener streamListener = mock(ServerStreamListener.class);
```
ericgribkoff: Done. This would be a nicer approach in the other tests as well, but I'd prefer to hold off on that as a follow-up PR, after everything else is solidified.

netty/src/test/java/io/grpc/netty/NettyStreamTestBase.java
ericgribkoff: Done

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
ericgribkoff: Each test can do the same thing - store the messages it receives - but a few didn't need this, so I just loop over them (and don't close them, but correcting that). This can definitely be made nicer (using a concrete class instead of mocks, as you suggested in AbstractServerStreamTest), but I didn't want to invest too much additional time into redoing the tests until the API changes are approved/stabilized, as changes there could require completely different testing mechanism.

Incidentally, this change includes a couple fixes to the flow control pushback test that - in local testing - seem to deflake it. That's orthogonal to this, but hopefully a nice benefit once I invest the time to clean up the mocking/Answers used here.

core/src/main/java/io/grpc/internal/ServerCallImpl.java
```diff
@@ -237,26 +236,27 @@ public void cancelled(Context context) {
...
       try {
-        if (call.cancelled) {
-          return;
+        while ((message = producer.next()) != null) {
+          if (call.cancelled) {
+            return;
+          }
+          try {
+            listener.onMessage(call.method.parseRequest(message));
+          } finally {
```
ericgribkoff: Fixed

core/src/main/java/io/grpc/internal/ClientCallImpl.java
ericgribkoff: Fixed. We spoke offline and the problem would occur with exceptions in handling the message when the producer potentially (as in the subsequent parts of this PR) has additional queued messages.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -497,11 +497,14 @@ public LogId getLogId() {
...
-        value.close();
-      } catch (IOException e) {
-        throw new RuntimeException(e);
+    public void messagesAvailable(MessageProducer producer) {
+      InputStream message;
+      while ((message = producer.next()) != null) {
+        try {
+          message.close();
+        } catch (IOException e) {
+          throw new RuntimeException(e);
```
ericgribkoff: Fixed

core/src/main/java/io/grpc/internal/AbstractStream.java
```diff
@@ -162,30 +162,28 @@ public void messageRead(InputStream is) {
      * Closes this deframer and frees any resources. After this method is called, additional calls
```
ejona86: This documentation _really_ needs to be updated to note that closing may be queued behind deframing messages.

core/src/main/java/io/grpc/internal/AbstractServerStream.java
```diff
@@ -232,9 +244,21 @@ public void inboundDataReceived(ReadableBuffer frame, boolean endOfStream) {
...
+        closeListener(status);
+      } else {
+        deframerClosedTask =
+            new Runnable() {
+              @Override
+              public void run() {
+                closeListener(status);
+              }
+            };
+        closeDeframer(false);
```
ejona86: Previously the code notified the application immediately, right? Why isn't this `true`?

core/src/main/java/io/grpc/internal/AbstractServerStream.java
```diff
@@ -243,7 +267,19 @@ public final void transportReportStatus(Status status) {
...
+        closeListener(Status.OK);
+      } else {
+        deframerClosedTask =
+            new Runnable() {
+              @Override
+              public void run() {
+                closeListener(Status.OK);
+              }
+            };
+        closeDeframer(false);
```
ejona86: This seems like it should be `true` as well to match the previous code. Generally, on the server-side we don't continue deframing after we have a status. This seems like it could cause a deadlock (in that we make no further progress). I'd have sort of expected a test to catch this... maybe we're missing a test case here.

ejona86:
I got through the second part. I still need to look at the third.

core/src/main/java/io/grpc/internal/AbstractServerStream.java
ericgribkoff: Done, and added test case to AbstractServerStreamTest. It seemed to fit better there than in the abstract transport test, as it's fairly low level.

core/src/main/java/io/grpc/internal/AbstractServerStream.java
ericgribkoff: Done, and added test case to AbstractServerStreamTest.

netty/src/test/java/io/grpc/netty/NettyClientTransportTest.java
ericgribkoff: Done.

core/src/main/java/io/grpc/internal/MessageDeframer.java
```diff
@@ -253,26 +251,8 @@ private void deliver() {
        * be in unprocessed.
        */
       boolean stalled = unprocessed.readableBytes() == 0;
-
-      if (endOfStream && stalled) {
-        boolean havePartialMessage = nextFrame != null && nextFrame.readableBytes() > 0;
```
ericgribkoff: Added this back in, and added test cases in AbstractServerStreamTest and MessageDeframerTest. The deframer will notify the listener on close if it has a partial message - this is ignored by clients, since they don't set end-of-stream, but used by server to call deframeFailed when appropriate.

core/src/main/java/io/grpc/internal/AbstractStream.java
ericgribkoff: This is done in the third commit. If this is all eventually to be merged as three separate commits, I can redo the history later to get the comment in the second commit instead.

core/src/main/java/io/grpc/internal/MessageDeframer.java
ejona86: It being missed on clients is an oversight (probably lost in an earlier refactoring). I've filed https://github.com/grpc/grpc-java/issues/3264

ericgribkoff:
@ejona86 I squashed the updates to the first two commits and did a force-push to clean up the PR. As we discussed, I also updated the third commit (adding app-thread deframing support) to use ApplicationThreadDeframer, a new class which handles the interactions between the transport state and MessageDeframer.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -941,7 +943,8 @@ public void headers(boolean outFinished,
 
     @Override
     public void rstStream(int streamId, ErrorCode errorCode) {
-      finishStream(streamId, toGrpcStatus(errorCode).augmentDescription("Rst Stream"), null, null);
+      finishStream(
+          streamId, toGrpcStatus(errorCode).augmentDescription("Rst Stream"), false, null, null);
```
ejona86: This needs logic to determine whether to stopDelivery.

ericgribkoff: Done. I restored the original check for CANCELLED or DEADLINE_EXCEEDED status alongside the explicit stopDelivery flag.

core/src/main/java/io/grpc/internal/ApplicationThreadDeframer.java
```diff
@@ -0,0 +1,189 @@
...
+  ApplicationThreadDeframer(
+      MessageDeframer.Listener listener,
+      Decompressor decompressor,
+      int maxMessageSize,
+      StatsTraceContext statsTraceCtx,
+      String debugString,
+      TransportExecutor transportExecutor) {
+    this.deframer =
+        new MessageDeframer(this, decompressor, maxMessageSize, statsTraceCtx, debugString);
+    this.storedListener = listener;
```
ericgribkoff: Done

core/src/main/java/io/grpc/internal/ApplicationThreadDeframer.java
```diff
@@ -0,0 +1,189 @@
...
+          @Override
+          public void run() {
+            storedListener.bytesRead(numBytes);
+          }
+        });
+  }
+
+  @Override
+  public void messagesAvailable(StreamListener.MessageProducer producer) {
+    messageReadQueue.add(producer.next());
```
ericgribkoff: Done

core/src/main/java/io/grpc/internal/MessageDeframer.java
```diff
@@ -105,29 +116,20 @@ public MessageDeframer(Listener listener, Decompressor decompressor, int maxMess
...
-  /**
-   * Sets the decompressor available to use.  The message encoding for the stream comes later in
-   * time, and thus will not be available at the time of construction.  This should only be set
-   * once, since the compression codec cannot change after the headers have been sent.
-   *
-   * @param decompressor the decompressing wrapper.
-   */
+  @Override
   public void setDecompressor(Decompressor decompressor) {
+    if (isClosedOrScheduledToClose()) {
```
ericgribkoff: You're right. Removed.

core/src/main/java/io/grpc/internal/AbstractClientStream.java
```diff
@@ -174,7 +174,7 @@ public final boolean isReady() {
     private ClientStreamListener listener;
 
     private boolean deframerClosed = false;
-    private Runnable deframerClosedTask;
+    protected Runnable deframerClosedTask;
```
ericgribkoff: Did this slip in from an earlier review? You're right, but I don't think the commit where this was protected survives anymore - it was removed in the earlier rebase, and so this should no longer be an issue.

ejona86: Yeah, that looks like a comment that failed to be sent out earlier. Sorry for the noise.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -721,8 +732,14 @@ void finishStream(
...
         if (status != null) {
-          stream.transportState().transportReportStatus(status, stopDelivery,
-              trailers != null ? trailers : new Metadata());
+          boolean isCancelled =
+              (status.getCode() == Code.CANCELLED || status.getCode() == Code.DEADLINE_EXCEEDED);
+          stream
+              .transportState()
+              .transportReportStatus(
+                  status,
+                  isCancelled || stopDelivery,
```
ejona86: This is complicated and surprising. I had hoped only the RST stream method I mentioned earlier would have this logic.

core/src/main/java/io/grpc/internal/ServerCallImpl.java
```diff
@@ -237,27 +241,37 @@ public void cancelled(Context context) {
...
-    public void messageRead(final InputStream message) {
+    public void messagesAvailable(final MessageProducer producer) {
+      InputStream message;
       Throwable t = null;
       try {
-        if (call.cancelled) {
-          return;
+        while ((message = producer.next()) != null) {
+          if (call.cancelled) {
+            return;
```
ejona86: Getting here fails to close `message`.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -457,24 +458,36 @@ public final void runInContext() {
...
-            if (closed) {
-              return;
-            }
-            try {
-              observer.onMessage(method.parseResponse(message));
-            } finally {
-              message.close();
+            while ((message = producer.next()) != null) {
+              if (closed) {
+                return;
```
ejona86: Getting here fails to close `message`.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
ericgribkoff: The stop on CANCELLED or DEADLINE_EXCEEDED is the existing logic here. I should not have removed this entirely in part 3 of this PR - I now am just adding it back to preserve the current behavior, alongside the now-necessary explicit stopDelivery flag. 

Do you want this logic moved to the rst stream method? I was trying to minimize changes to other code paths, which would also potentially be impacted by moving the isCancelled check - this might be the appropriate place for the check, but didn't seem relevant to the changes here.

ejona86: I'm not excited about the mixing of implicit (via Status) and explicit `stopDelivery`. It should be one way or the other, unless there was a _really good_ reason. In auditing the existing calls to this method, it seemed that only the rst stream one was depending on this behavior. If you'd like we can sit down and pair program/audit the callers.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -150,7 +150,7 @@ public void request(final int numMessages) {
     @Override
     public void cancel(Status reason) {
       synchronized (state.lock) {
-        state.cancel(reason, false, null);
+        state.cancel(reason, true, null);
```
ejona86: Yeah, always stopDelivery. If it's not today, it should be.

ericgribkoff:
@ejona86 Just updated the handling of exceptions/closing in ClientCallImpl and ServerCallImpl. Should be somewhat improved now in terms of clarity, and maintains for both server and client that (a) all messages will be closed (b) if the application code throws an exception, that exception is preserved (e.g., an IOException during the clean-up phase's closes is ignored) and (c) if a message close fails after the application message handling returned normally, that IOException is propagated just like in the current code.

ericgribkoff:
Rebased onto master to fix the merge conflicts and squashed the fixup commits back into the original three commit structure.

PR #3150, opened 2017-06-28
examples: add bazel build example

dapengzhang0:
## Maven

If you prefer to use Maven:
```
$ mvn verify
$ # Run the server
$ mvn exec:java -Dexec.mainClass=io.grpc.examples.helloworld.HelloWorldServer
$ # In another terminal run the client
$ mvn exec:java -Dexec.mainClass=io.grpc.examples.helloworld.HelloWorldClient
```

## Bazel

If you prefer to use Bazel:
```
(With Bazel v0.4.5 or above.)
$ bazel build :hello-world-server :hello-world-client
$ # Run the server:
$ bazel-bin/hello-world-server
$ # In another terminal run the client
$ bazel-bin/hello-world-client
```

PR #3151, opened 2017-06-28
examples: use the official error_details.proto

dapengzhang0:
The `error_details.proto` file in `examples/src/main/proto/` was not originated from grpc-java. 
It was officially opensourced at https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto


examples/build.gradle
```diff
@@ -25,6 +25,7 @@ repositories {
 def grpcVersion = '1.5.0-SNAPSHOT' // CURRENT_GRPC_VERSION
 
 dependencies {
+  compile "com.google.api.grpc:grpc-google-common-protos:0.1.12"
```
ericgribkoff: I think this should use proto-google-common-protos, not grpc-google-common-protos. See https://github.com/grpc/grpc-java/issues/2961#issuecomment-298370880

dapengzhang0: Thanks! Updated.

ejona86:
Maven build is pretty broken. See the Travis logs.

dapengzhang0:
That's interesting, the 1st commit using `grpc-google-common-protos` worked.

ejona86:
FYI: this will conflict with #3150

dapengzhang0:
Is there any particular reason we use v0.1.9 in grpc-protobuf? Can we bump to v0.1.12 globally?

dapengzhang0:
`proto-google-common-protos` v0.1.9 pom.xml
```
<groupId>com.google.protobuf</groupId>
<artifactId>protobuf-java</artifactId>
<version>3.2.0</version>
```
`proto-google-common-protos` v0.1.12 pom.xml
```
<groupId>com.google.protobuf</groupId>
<artifactId>protobuf-java</artifactId>
<version>3.3.0</version>
```

ejona86:
Let's add maven-enforcer-plugin to the pom.xml to detect problems like this in the future. It's important for our users to do the same:

```xml
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-enforcer-plugin</artifactId>
        <version>1.4.1</version>
        <executions>
          <execution>
            <id>enforce</id>
            <goals>
              <goal>enforce</goal>
            </goals>
            <configuration>
              <rules>
                <requireUpperBoundDeps/>
              </rules>
            </configuration>
          </execution>
        </executions>
      </plugin>
```

ejona86:
Actually, I'm just going to do the maven enforcer part in #3158.

Issue #3149, opened 2017-06-28
proto message using map NoSuchMethodError


SeriousMa:
it seems invoke  CodedOutputStream.class in grpc-protobuf-lite 
![image](https://user-images.githubusercontent.com/464213/27632080-3a040c34-5c2d-11e7-8780-27aa1b111bc9.png)
![image](https://user-images.githubusercontent.com/464213/27632085-3e686b9e-5c2d-11e7-815a-3a9c32fb199c.png)
![image](https://user-images.githubusercontent.com/464213/27632086-411628f4-5c2d-11e7-8b05-58bde4582b3a.png)

```
  <dependency>
                <groupId>io.grpc</groupId>
                <artifactId>grpc-all</artifactId>
                <version>1.3.0</version>
            </dependency>
```

ejona86:
Maybe you're hitting grpc/grpc-java#2985, since `isSerializationDeterministic` is not in protobuf-lite. Could you define your deps without grpc-all and grpc-protobuf-lite and see if that resolves the problem?

SeriousMa:
i exclusion  protobuf-lite and it's ok now
```
 <dependency>
                <groupId>io.grpc</groupId>
                <artifactId>grpc-all</artifactId>
                <version>${grpc.version}</version>
                <exclusions>
                    <exclusion>
                        <artifactId>protobuf-lite</artifactId>
                        <groupId>com.google.protobuf</groupId>
                    </exclusion>
                </exclusions>
            </dependency>
```

ejona86:
Fixed by #3153

PR #3153, opened 2017-06-29
all: Don't depend directly on grpc-protobuf-lite

### LINKED ISSUE #2985: grpc-all brings in both protobuf-java and protobuf-lite
```
    <dependency>
      <groupId>io.grpc</groupId>
      <artifactId>grpc-protobuf</artifactId>
      <version>1.2.0</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>io.grpc</groupId>
      <artifactId>grpc-protobuf-lite</artifactId>
      <version>1.2.0</version>
      <scope>compile</scope>
    </dependency>
```

grpc-protobuf depends on protobuf-java and grpc-protobuf-lite depends on protobuf-lite. Although grpc-protobuf depends on grpc-protobuf-lite and excludes the protobuf-lite dependency, that does nothing for grpc-all which depends directly on both.

We should probably remove the explicit dependency of grpc-protobuf-lite from grpc-all, or add in an exclusion for protobuf-lite.

ejona86:
grpc-protobuf-lite brings in protobuf-lite as a dependency, which has
different versions of classes in protobuf. This causes duplicate classes
on the class path, which easily breaks things. Since lite the lite
runtime/codegen is incompatible with full protobuf, it makes sense to
only depend on one of the two.

grpc-protobuf actually already has a dependency on grpc-protobuf-lite,
but excludes the protobuf-lite dependency, making the dep safe. But this
is more of an implementation detail.

We only change deps so that JavaDoc, code coverage, and similar will
still include grpc-protobuf-lite.

Fixes #2985 

Doing this now since it seems a user (in #3149) has hit this problem.
Previously it was theoretical.

### DIFF
--- a/all/build.gradle
+++ b/all/build.gradle
@@ -32,7 +32,7 @@ for (subproject in rootProject.subprojects) {
 }
 
 dependencies {
-    compile subprojects
+    compile subprojects.minus(project(':grpc-protobuf-lite'))
 }
 
 javadoc {


dapengzhang0:
This should be backported to 1.4.x branch.

ejona86:
@dapengzhang0, why backport? I don't think we'd do a 1.4.1 just for this, and thus it would seem 1.5.0 will be released before any 1.4.1.

dapengzhang0:
Never mind, I misunderstood the release number.

PR #3154, opened 2017-06-29
build.gradle: bump proto-google-common-protos to v0.1.12

### LINKED ISSUE #3148: using protoc 3.3.0 generate wrong parseMethod？

grpc 1.4.0
using protoc generate version 3.2.0
image

using protoc generate version 3.3.0
image

dapengzhang0:
This might resolve #3148

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -180,7 +180,7 @@ subprojects {
                 hpack: 'com.twitter:hpack:0.10.1',
                 jsr305: 'com.google.code.findbugs:jsr305:3.0.0',
                 oauth_client: 'com.google.auth:google-auth-library-oauth2-http:0.4.0',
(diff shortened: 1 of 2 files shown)


ejona86:
Whoa, this shouldn't be necessary. If updating the proto is required, then there is a regression in protobuf that needs to be fixed. (Unless if the application wants to use the new method directly.)

dapengzhang0:
Agree.

ejona86:
For posterity's sake: I'm okay with updating the proto, but I'm not okay with _forced to_ because of protobuf library incompatibilities, since that wouldn't scale for our users. In this case, we discovered it was simply Maven choose the wrong protobuf library version.

PR #3159, opened 2017-06-30
core: Move client-side decompressor selection to stream

### LINKED ISSUE #2157: Flaky CompressionTest

### LINKED ISSUE #2865: Rare race condition in Client

ejona86:
Previously ClientCallImpl's stream listener would call
stream.setDecompressor(), but this has always been a bit strange as the
only case where the call listener calls the stream and forms a bit of a
loop. It also turned out to be racy in the presence of
DelayedClientStream since DelayedClientStream does not guarantee that
the listener has processed before returning.

Now we let the stream handle decompressor selection itself. Compressor
selection on client and server and decompressor selection on server
remain unchanged. Nothing prevents them from being changed, other than
it is currently unnecessary to fix the severe compressionTest flake.

Fixes #2865
Fixes #2157

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractClientStream.java
+++ b/core/src/main/java/io/grpc/internal/AbstractClientStream.java
@@ -16,9 +16,14 @@
 
 package io.grpc.internal;
 
+import static io.grpc.internal.GrpcUtil.MESSAGE_ENCODING_KEY;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import io.grpc.Codec;
 import io.grpc.Compressor;
+import io.grpc.Decompressor;
+import io.grpc.DecompressorRegistry;
 import io.grpc.Metadata;
 import io.grpc.Status;
 import java.io.InputStream;
@@ -111,6 +116,11 @@ public abstract class AbstractClientStream extends AbstractStream
     transportState().setMaxInboundMessageSize(maxSize);
   }
 
+  @Override
+  public final void setDecompressorRegistry(DecompressorRegistry decompressorRegistry) {
+    transportState().setDecompressorRegistry(decompressorRegistry);
+  }
+
   /** {@inheritDoc} */
   @Override
   protected abstract TransportState transportState();
@@ -172,6 +182,7 @@ public abstract class AbstractClientStream extends AbstractStream
     private final StatsTraceContext statsTraceCtx;
     private boolean listenerClosed;
     private ClientStreamListener listener;
+    private DecompressorRegistry decompressorRegistry = DecompressorRegistry.getDefaultInstance();
 
     private Runnable deliveryStalledTask;
 
@@ -184,6 +195,12 @@ public abstract class AbstractClientStream extends AbstractStream
     protected TransportState(int maxMessageSize, StatsTraceContext statsTraceCtx) {
       super(maxMessageSize, statsTraceCtx);
       this.statsTraceCtx = Preconditions.checkNotNull(statsTraceCtx, "statsTraceCtx");
+    }
+
+    private void setDecompressorRegistry(DecompressorRegistry decompressorRegistry) {
+      Preconditions.checkState(this.listener == null, "Already called start");
+      this.decompressorRegistry =
+          Preconditions.checkNotNull(decompressorRegistry, "decompressorRegistry");
     }
 
     @VisibleForTesting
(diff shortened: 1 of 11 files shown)


core/src/main/java/io/grpc/internal/AbstractClientStream.java
```diff
@@ -218,6 +235,19 @@ protected final ClientStreamListener listener() {
     protected void inboundHeadersReceived(Metadata headers) {
       Preconditions.checkState(!statusReported, "Received headers on closed stream");
       statsTraceCtx.clientInboundHeaders();
+
+      Decompressor decompressor = Codec.Identity.NONE;
+      if (headers.containsKey(MESSAGE_ENCODING_KEY)) {
+        String encoding = headers.get(MESSAGE_ENCODING_KEY);
```
carl-mastrangelo: yikes, wanna combine this with the if?

core/src/main/java/io/grpc/internal/AbstractClientStream.java
```diff
@@ -218,6 +235,19 @@ protected final ClientStreamListener listener() {
...
+        decompressor = decompressorRegistry.lookupDecompressor(encoding);
+        if (decompressor == null) {
+          deframeFailed(Status.INTERNAL.withDescription(
+              String.format("Can't find decompressor for %s", encoding)).asException());
```
carl-mastrangelo: RuntimeException() ?

core/src/main/java/io/grpc/internal/ClientStream.java
```diff
@@ -52,6 +53,14 @@
   void setAuthority(String authority);
 
   /**
+   * Sets the registry to find a decompressor for the framer. May only be called before {@link
+   * #start}.
```
carl-mastrangelo: Mention that this may be a noop, like for Inprocess.

core/src/main/java/io/grpc/internal/ServerStream.java
```diff
@@ -59,6 +60,13 @@
   void cancel(Status status);
 
   /**
+   * Sets the decompressor on the deframer.
+   *
+   * @param decompressor the decompressor to use.
+   */
+  void setDecompressor(Decompressor decompressor);
```
carl-mastrangelo: same comment about docs.

core/src/main/java/io/grpc/internal/AbstractClientStream.java
ejona86: `asRuntimeException()` produces identical behavior. The exception is thrown away and only the status is kept in all `deframeFailed()` handling. And `deframeFailed()` is also the only way the `TransportState` can kill the RPC.

core/src/main/java/io/grpc/internal/ClientStream.java
ejona86: Done

core/src/main/java/io/grpc/internal/ServerStream.java
ejona86: Done

core/src/main/java/io/grpc/internal/AbstractClientStream.java
ejona86: Done.

core/src/main/java/io/grpc/internal/AbstractClientStream.java
ejona86: Done.

PR #3160, opened 2017-06-30
all: Use fixed version number for java 6 signature

ejona86:
This is important for stable builds, as if the signature changes the old source
may no longer validate.


dapengzhang0:
Does netty need a fixed version number for java 7 signature?

ejona86:
@dapengzhang0, yes, you're right. Now also fixed.

Issue #2446, opened 2016-11-26
Generate interface for every service action


lukaszx0:
The generated service stubs evolved quite a lot in past few months and I've noticed that we removed service interface in favor of having just an abstract base class, it was indeed redundant and probably not very useful (I don't know if it was solve factor that led to removing it, if not please correct me).

While the per-service interface in the presence of equivalent abstract class might not provide a lot of value (in the context of generated services) I'd like to propose generating per-action interfaces which would be implemented by the base abstract service.

The value of per-action interfaces is that it would allow users to easily build custom service definitions which then would allow custom integrations for frameworks where single action is represented by class (as opposed to one class implementing all actions). This approach is used in many custom frameworks which relay heavily on dependency injection and the main benefit it provides is that in order to test single action, you don't need to provide dependencies for every action of a given service.

I'm open to work on implementation if we reach consensus. The change will be additive and won't break any existing APIs or implementations.

daicoden:
This would be really helpful. Something like: https://github.com/daicoden/grpc-reflection-request/pull/1/files

It's not just tests but also resource scoping.

As an example, a service with different permissions (like read/write) might have access to different bound resources. It's hard to explain with a straight face to other developers why there are two optional transactors available (read & write) and how, based on scoping at dispatch time, one of the transactors is present, and the other is not, and that they should read the proto definition in order to figure out which one to use.

ejona86:
It was removed because it is not API-safe. Interfaces can not receive new methods over time, whereas gRPC services can. If a new method is added to a gRPC service, it can break compilation of Java code which was deemed unacceptable (we can't just trust our users to "do the right thing" and avoid the interface when they don't own the service). https://github.com/grpc/grpc-java/issues/1469#issuecomment-226873511 had the discussion.

This is a case where a Java 8-specific API could do better, because it could provide default implementations. But it would probably need to be a separate API; it seems we'll need to provide Java 7 compatibility for a long time still (if for no other reason, Android). But such a thing causes trouble for toolchains, since we want to be able to provide pre-generated code to consumers. It's unclear how best to proceed and if it is worthwhile.

[...]

PR #3172, opened 2017-07-05
Generate interface for each rpc method.

daicoden:
Makes class per rpc style implementations easier for the developer.

Issue #2446

### DIFF
--- a/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
+++ b/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
@@ -39,6 +39,10 @@ public final class BenchmarkServiceGrpc {
           .setResponseMarshaller(io.grpc.protobuf.ProtoUtils.marshaller(
               io.grpc.benchmarks.proto.Messages.SimpleResponse.getDefaultInstance()))
           .build();
+  @io.grpc.ExperimentalApi("https://github.com/grpc/grpc-java/issues/2446")
+  public interface MethodUnaryCall extends
+      io.grpc.stub.ServerCalls.UnaryMethod<io.grpc.benchmarks.proto.Messages.SimpleRequest, io.grpc.benchmarks.proto.Messages.SimpleResponse> {}
+
   @io.grpc.ExperimentalApi("https://github.com/grpc/grpc-java/issues/1901")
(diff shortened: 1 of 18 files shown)


ejona86:
okay to test

daicoden:
To use these interfaces in the stub class, MethodHandler would implement these methods instead of the the different ServerCall interfaces. I don't think *ImplBase can manage invoke based on how it's structured right now.

The reason I chose to do it this way was to have the MethodType information available in annotation processors. You can only read constants in annotation processors, and I can see that an interface extends ServerCalls.UnaryCall, but I can't get access to the MethodDescriptor instance to ask it the type.

However, I can live without that. I can change this PR to generate implementations for *ImplBase instead if you think that better.

The main motivation for this PR is to have some API for implementing per-class RPCs without having to use ServerCalls.X<ReqT,RespT>.

ejona86:
It seems these changes would work well in a separate code generator. The interfaces are useless without an annotation system, and although other annotation systems could theoretically benefit from them as well it's unclear whether any actually will benefit. The interfaces also don't use the existing generated code nor does the existing generated code depend on the interfaces.

So I don't think we're likely to accept this at this point because it seems better served by a separate code generator. It seems like writing a generator for this case would be fairly straight-forward.

daicoden:
Alright, thanks for the consideration :).

Is the vision for extending GRPC to be through protoc extensions? I got everything working by forking and modifying, but it felt nasty, but maybe I can look at chaining generators.

I would put a plug in for supporting GRPC extension through annotation processors, more meta information is required in the generated code to make it a reality. Reason being there's a ton to wrap your head around without adding another language layer, there's no java poet for writing the java classes, you have to switch how you do tests/ set up different harness. Just a lot of infra to bring in to maintain it.

ejona86:
At present the gRPC plugin does not provide extension points. This is because: 1) in OSS there needs to be a canonical version of the proto; any extension points _must not_ be used in OSS, 2) we've not seen a strong case for them in gRPC; the cases we've seen for them work fine as a separate class.

I'd instead just write a plugin that assumes the grpc-java plugin has already run. Yes, it produces yet _another_ class, but such is life.

I'm not sure what you'd have to do to swap how you do tests. Unless you are talking about testing the generated code. For tests we've had good success with generating code as part of the test build process.

Yes, maintaining your own protoc plugin requires more infrastructure. You may consider writing the plugin [in Java](https://github.com/salesforce/grpc-java-contrib/tree/master/jprotoc) for easier maintenance. Unfortunately, writing it in Java does make some things harder, like learning the generated message name for a Protobuf message.

PR #3174, opened 2017-07-05
core: Force conscrypt for all AppEngine clients

kpayson64:


ejona86:
I thought Java 8 users were going to use tcnative. Is that not the case?

kpayson64:
It sounds like they should be using Conscrypt.  I've cc'ed you on a thread.

core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -58,8 +58,7 @@
   // AppEngine runtimes have constraints on threading and socket handling
   // that need to be accommodated.
   public static final boolean IS_RESTRICTED_APPENGINE =
-      System.getProperty("com.google.appengine.runtime.environment") != null
-          && "1.7".equals(System.getProperty("java.specification.version"));
+      System.getProperty("com.google.appengine.runtime.environment") != null;
```
carl-mastrangelo: If Using OkHTTP, it would seem to try harder to not use Conscrypt.  Looking in OkHttpChannelBuilder:

```
            if (GrpcUtil.IS_RESTRICTED_APPENGINE) {
              // The following auth code circumvents the following AccessControlException:
              // access denied ("java.util.PropertyPermission" "javax.net.ssl.keyStore" "read")
              // Conscrypt will attempt to load the default KeyStore if a trust manager is not
              // provided, which is forbidden on AppEngine
              sslContext = SSLContext.getInstance("TLS", Platform.get().getProvider());
              TrustManagerFactory trustManagerFactory =
                  TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
              trustManagerFactory.init((KeyStore) null);
              sslContext.init(
                  null,
                  trustManagerFactory.getTrustManagers(),
                  // Use an algorithm that doesn't need /dev/urandom
                  SecureRandom.getInstance("SHA1PRNG", Platform.get().getProvider()));

            } else {
              sslContext = SSLContext.getInstance("Default", Platform.get().getProvider());
            }
```

Appengine restricted looks like it is using intentionally weaker ciphers.  Are you sure that this change actually uses conscrypt?

ejona86:
From my reading of the thread, it is basically saying they need to include the binary component manually, which matched my current understanding. But that is tcnative normally. Conscrypt was used in Java 7 because OkHttp was being used. In Java 8 Netty should be used, and Conscrypt support in Netty is still stabilizing.

core/src/main/java/io/grpc/internal/GrpcUtil.java
ejona86: What weak ciphers do you see? I just see providing an alternative SecureRandom.

carl-mastrangelo: The cipher is weaker if the iv is weaker.  

Actually, from reading what SSLContext.getInstance does, neither of "Default" or "TLS" seems right.  TLS could pick 1.1 or lower, which is forbidden by http2.  Default could pick SSLv3.


I don't actually see a protocol that says 1.2 or higher.

https://docs.oracle.com/javase/6/docs/technotes/guides/security/StandardNames.html#SSLContext

ejona86: The ciphers and TLS version are dictated by [the ConnectionSpec](https://github.com/grpc/grpc-java/blob/v1.4.0/okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java#L79) which is [applied elsewhere](https://github.com/grpc/grpc-java/blob/v1.4.0/okhttp/src/main/java/io/grpc/okhttp/OkHttpTlsUpgrader.java#L75). ConnectionSpec [explicitly sets both tls versions and the suites](https://github.com/grpc/grpc-java/blob/v1.4.0/okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/ConnectionSpec.java#L128).

So the provider here will be required to support TLSv1.2 and a GCM cipher.

carl-mastrangelo:
@ejona86 are you saying that this PR is or isn't correct?

ejona86:
This PR was not correct because it triggered using OkHttp in Java 8 AppEngine. Java 8 AppEngine should be using Netty, and thus tcnative. Thus any discussion about Conscrypt was misguided, because that was going down the OkHttp road.

PR #3183, opened 2017-07-06
core: Stop using context ClassLoader in Providers

### LINKED ISSUE #2375: Use provider's class loader instead of context class loader
In a multi classloader application it may easily turn out Thread#contextClassloader contains something unrelated to a classloader that indeed loads grpc classes.

``` java
@Internal
public abstract class ServerProvider {
  private static final ServerProvider provider =
      load(Thread.currentThread().getContextClassLoader());
...
```

It's better to check current classloader for ServerProvider.class too.

I use JRE 1.8.0_66, macos 10.12

ejona86:
currentThread().getContextClassLoader() was originally used to match
ServiceLoader's default. ServiceLoader is mostly used with SPIs that
exist in the bootclasspath and so the class's ClassLoader would not
work. But in gRPC we have a library in the application's ClassLoader.

Context ClassLoader has been causing problems in any case more than one
application ClassLoader exists, like in Servlet containers, Android, and
plugins. To my knowledge, in every case the context class loader is
different from the provider's class loader, the context class loader has
caused breakage, not success. In addition, since we use static fields
for storing the provider results, using anything but the provider's
class loader is asking to have results that vary simply depending on
which thread called gRPC first.

Fixes #2375

All tests pass in Google with this change.

### DIFF
--- a/core/src/main/java/io/grpc/ManagedChannelProvider.java
+++ b/core/src/main/java/io/grpc/ManagedChannelProvider.java
@@ -34,7 +34,7 @@ import java.util.ServiceLoader;
 @Internal
 public abstract class ManagedChannelProvider {
   private static final ManagedChannelProvider provider
-      = load(getCorrectClassLoader());
+      = load(ManagedChannelProvider.class.getClassLoader());
 
   @VisibleForTesting
   static ManagedChannelProvider load(ClassLoader classLoader) {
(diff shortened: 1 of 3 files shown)


ejona86:
retest this please

zhangkun83:
> ServiceLoader is mostly used with SPIs that
exist in the bootclasspath and so the class's ClassLoader would not
work. But in gRPC we have a library in the application's ClassLoader.

Do you mean we didn't need context class loader in the first place?

ejona86:
> Do you mean we didn't need context class loader in the first place?

Yes, I think context class loader was always the wrong one. At the time, it wasn't clear why ServiceLoader would default to context class loader. I think the main realization is that it makes sense for things in the JDK, and so that means we now have a better understanding on why it isn't appropriate for us.

PR #3184, opened 2017-07-06
core: Server uses transport's ScheduledExecutorService

ejona86:
For Netty, this reduces the number of threads necessary for servers (although
until channel is converted, actual number of threads isn't impacted) and
naturally reduces contention and timeout latency.

For InProcess, this gets us closer to allowing applications to provide all
executors, which is especially useful during tests.

CC @zhangkun83

### DIFF
--- a/core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
@@ -22,7 +22,10 @@ import io.grpc.Internal;
 import io.grpc.internal.AbstractManagedChannelImplBuilder;
 import io.grpc.internal.ClientTransportFactory;
 import io.grpc.internal.ConnectionClientTransport;
+import io.grpc.internal.GrpcUtil;
+import io.grpc.internal.SharedResourceHolder;
 import java.net.SocketAddress;
+import java.util.concurrent.ScheduledExecutorService;
 
 /**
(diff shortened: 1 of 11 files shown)


ejona86:
I do have one pending question for this change: whether it is a good idea for the InProcess server to use the channel's scheduler. In one way, it is nicer, since it is one fewer things to configure. In another, why should the client's configuration be impacting the server's behavior? I think I am leaning toward changing it so the server has its own scheduler, but I'm still on the fence.

core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
```diff
@@ -98,13 +103,16 @@ public ConnectionClientTransport newClientTransport(
...
     }
 
     @Override
     public void close() {
+      if (closed) {
+        return;
+      }
       closed = true;
-      // Do nothing.
+      SharedResourceHolder.release(GrpcUtil.TIMER_SERVICE, timerService);
```
dapengzhang0: is test verifying the release?

ejona86:
@zpencer @dapengzhang0, PTAL. For the server to own the scheduler, it has to be refcounted. Not too hard, but a bit more interesting than earlier. I added a test for the server, but not InProcessTransport since there wasn't any existing infrastructure. The changes to InProcessChannelBuilder were reverted.

Issue #1981, opened 2016-06-25
Executor usage in `newCall()` races with channel shutdown and termination.


zhangkun83:
`ManagedChannelImpl` clear `scheduledExecutor` in `shutdown()`, and releases (which potentially closes) `executor` in `maybeTerminateChannel()`.

Neither `newCall()` nor `ClientCallImpl` checks the shutdown state of the channel. `ClientCallImpl` relies on `FailingClientTransport` for the expected behavior. However, `ClientCallImpl` uses the passed in executors anyway, for scheduling the deadline timer and invoking the call listener.

If `ClientCallImpl` tries to schedule a deadline timer after the channel is shut down, it will get a NPE. If it runs the call listener after the shared executor has been closed, which is 1 second (`SharedResourceHolder.DESTROY_DELAY_SECONDS`) after all references are gone, e.g., the application calls `Call.start()` that late, it will get a `RejectedExecutionException`. Our current tests are not testing for the two cases.

This doesn't seem to be a serious issue. It only affect people who try to use `Call`s after the channel has been shutdown. I am yet to figure out a solution.

Anyway, it seems `executor` should be cleared after being returned to the shared pool, like `scheduledExecutor`.

zhangkun83:
We should not miss out the case where a new call is created, the channel is terminated, then the call is started. Simply resetting the executors in channel upon termination doesn't help with this case, because the call holds on to the executors.

zhangkun83:
`OobChannel` has the same issue with `ManagedChannelImpl`.

zhangkun83:
@ejona86 @dapengzhang0, I wonder if the fix should be:
 1. Only release the executors when channel is terminated
 2. `newCall()` and `call.start()` throw if channel is shutdown. Calling them after `shutdown()` is application's fault. They should and can consciously avoid it. This might be considered an API-breaking change though.

ejona86:
@zhangkun83, I'm not sold on (2), but even if I was favorable toward it, it does seem like an API breakage.

PR #3185, opened 2017-07-06
core: Channel uses transport's ScheduledExecutorService

ejona86:
Coupled with the similar change on server-side, this removes the need for a
thread when using Netty. For InProcess and OkHttp, it would allow us to let the
user to provide the scheduler for tests or application-wide thread sharing.

The first commit is from #3184, and need not be reviewed here. This change
depends on it, but not too heavily; the dependency could be removed without
much trouble (although then at the price of merge conflicts and needing to
re-apply review fixes).

### DIFF
--- a/core/src/main/java/io/grpc/internal/ClientTransportFactory.java
+++ b/core/src/main/java/io/grpc/internal/ClientTransportFactory.java
@@ -18,6 +18,7 @@ package io.grpc.internal;
 
 import java.io.Closeable;
 import java.net.SocketAddress;
+import java.util.concurrent.ScheduledExecutorService;
 import javax.annotation.Nullable;
 
 /** Pre-configured factory for creating {@link ConnectionClientTransport} instances. */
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/grpc/internal/ClientTransportFactory.java
```diff
@@ -32,6 +33,17 @@ ConnectionClientTransport newClientTransport(SocketAddress serverAddress, String
       @Nullable String userAgent);
 
   /**
+   * Returns an executor for scheduling provided by the transport. The service should be configured
+   * to allow cancelled scheduled runnables to be GCed.
+   *
+   * <p>The executor should not be used after the factory has been closed. The caller should ensure
+   * any outstanding tasks are cancelled before the factory is closed. However, it is a known issue
+   * that ClientCallImpl may use this executor after close, so implementations should not go out of
```
zhangkun83: Add a link to #1981 ?

ejona86: Done.

ejona86:
Since the server-side change no longer includes the code this PR was depending on, it has been rebased to exclude those commits. Code in InProcessChannelBuilder.java was copied from the older server-side change to here, but everything else remains the same.

@zhangkun83, wanna take a look at InProcessChannelBuilder.java? It now mirrors OkHttpChannelBuilder.java.

ejona86:
@zhangkun83, ping for looking at InProcessChannelBuilder. It's pretty small.

zhangkun83:
Still LGTM

PR #3190, opened 2017-07-06
core: Stop using context ClassLoader in Providers (backport)

### LINKED ISSUE #2375: Use provider's class loader instead of context class loader

ejona86:
currentThread().getContextClassLoader() was originally used to match
ServiceLoader's default. ServiceLoader is mostly used with SPIs that
exist in the bootclasspath and so the class's ClassLoader would not
work. But in gRPC we have a library in the application's ClassLoader.

Context ClassLoader has been causing problems in any case more than one
application ClassLoader exists, like in Servlet containers, Android, and
plugins. To my knowledge, in every case the context class loader is
different from the provider's class loader, the context class loader has
caused breakage, not success. In addition, since we use static fields
for storing the provider results, using anything but the provider's
class loader is asking to have results that vary simply depending on
which thread called gRPC first.

Fixes #2375

### DIFF
--- a/core/src/main/java/io/grpc/ManagedChannelProvider.java
+++ b/core/src/main/java/io/grpc/ManagedChannelProvider.java
@@ -34,7 +34,7 @@ import java.util.ServiceLoader;
 @Internal
 public abstract class ManagedChannelProvider {
   private static final ManagedChannelProvider provider
-      = load(getCorrectClassLoader());
+      = load(ManagedChannelProvider.class.getClassLoader());
 
   @VisibleForTesting
   static ManagedChannelProvider load(ClassLoader classLoader) {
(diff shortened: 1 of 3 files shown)


ejona86:
Hmm... thinking about this more, the original intention was to backport this before the 1.5.0 release. However, it didn't get in 1.5.0 and I think this isn't something we want in a patch release. So I'm going to trash this instead. Sorry @ericgribkoff

PR #3191, opened 2017-07-06
context: stop using logger at all in static initialization.

zhangkun83:
Turns out avoiding log() alone in the static initialization is not
enough. isLoggable() can also be overridden and call back to Context.
We take a another approach, save the exception and log it outside of
the initialization block.


adriancole:
sometimes turtles all the way down is a bad thing I guess :P

PR #3193, opened 2017-07-06
context: avoid a lock on maybeLogStorageOverrideNotFound.

zhangkun83:
attach() is frequently called, and in most cases
storageOverrideNotFoundErrors is empty but poll() always acquires a
lock.  Use volatile to avoid the locking.


zhangkun83:
Closed in favor of #3198

PR #3198, opened 2017-07-07
context avoid maybeLogStorageOverrideNotFound by lazy initialization

dapengzhang0:
might not be the optimal solution, but demos an alternative approach to avoid logging promlem
cc @zhangkun83

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -113,7 +113,6 @@ public class Context {
   // problematic (e.g., NullPointerException) if the Context class has not done loading at this
   // point.
   private static volatile Storage storage;
-  private static Exception storageInitError;
 
   // For testing
   static Storage storage() {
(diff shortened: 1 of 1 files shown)


context/src/main/java/io/grpc/Context.java
```diff
@@ -109,37 +108,36 @@
...
+          try {
+            Class<?> clazz = Class.forName("io.grpc.override.ContextStorageOverride");
+            tmp = (Storage) clazz.getConstructor().newInstance();
+            storage = tmp;
+          } catch (ClassNotFoundException e) {
+            tmp = new ThreadLocalContextStorage();
+            storage = tmp;
+            log.log(Level.FINE, "Storage override doesn't exist. Using default.", e);
+          } catch (Exception e) {
+            storageInitError = e;
```
dapengzhang0: Without this field, next time it would fail but possibly would throw NPE from which user can not easily find the root cause if the original RuntimeExecption had happened quite long ago and only been logged.

ejona86:
@zhangkun83, when I approved the earlier commit, I already checked: LinkedBlockingQueue only does an atomic read after the queue is drained (it has a fast-path for count==0). The only deficiencies is it has two pointer-chases and the queue memory overhead.

I have no reservation swapping to something else though.

context/src/main/java/io/grpc/Context.java
dapengzhang0: seems I was wrong. removing this field.

ejona86:
We discussed locking concerns in-person and decided to swap to compare-and-set in #3203.

PR #3203, opened 2017-07-07
context: Lazy load storage

ejona86:
Using static initialization is possible, but quite complex to handle
logging and circular loading. Lazy loading prevents an entire class of
circular dependencies.

Given the various iterations involved here, it'd be good if all the
reviewers reviewed this change.

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -19,11 +19,11 @@ package io.grpc;
 import java.util.ArrayList;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
-import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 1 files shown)


context/src/main/java/io/grpc/Context.java
```diff
@@ -108,39 +109,38 @@
...
+  // they don't depend on storage, like key() and currentContextExecutor(). It also makes it easier
+  // to handle exceptions.
+  private static final AtomicReference<Storage> storage = new AtomicReference<Storage>();
 
   // For testing
   static Storage storage() {
-    if (storage == null) {
-      throw new RuntimeException("Storage override had failed to initialize", storageInitError);
+    Storage tmp = storage.get();
+    if (tmp == null) {
```
carl-mastrangelo: consider moving the creation logic into a slow path, to make inlining of `storage()` easier.  Also would separate concerns of getting storage from creating storage.

dapengzhang0:
@ejona86 can you rebase first?

zhangkun83:
@carl-mastrangelo I can confirm it works on all the known broken targets. I can run a global TAP to make sure no new breakages pop up.

context/src/main/java/io/grpc/Context.java
dapengzhang0: No need to declare `tmp`:
`if (storage.get() == null)`

ejona86:
@dapengzhang0, done. It is now rebased. I handled the merge conflicts within the first commit. The only real difference is now I delete more code.

context/src/main/java/io/grpc/Context.java
ejona86: Ah, good point. Done.

context/src/main/java/io/grpc/Context.java
```diff
@@ -109,37 +109,42 @@
...
+      tmp = createStorage();
+    }
+    return tmp;
+  }
+
+  private static Storage createStorage() {
+    // Note that this method may be run more than once
     try {
       Class<?> clazz = Class.forName("io.grpc.override.ContextStorageOverride");
```
adriancole: @raphw I always wondered if there's a bytebuddy way to test code like this. For example, a test class which declares a good and various naughty impls of this type. That way, we could test these defenses.

raphw: You could get the byte code of all involved classes, manipulate some or allof them and load them in an isolated class loader. Have a look at ByteArrayClassLoader.

ejona86:
Could someone review? It seems people are content with the change.

dapengzhang0:
LGTM. Are you going to add a test with custom logger handler or something that can break the old code?

zhangkun83:
The promised global TAP was delayed. I will run it now and see the results (hopefully there won't be any issue).

ejona86:
@dapengzhang0, I don't plan to add any test at this time. Since you can't simply rename a .class, it's pretty annoying. You either need to have pre-prepared (manually or via convoluted build) some .classes and have them stored somewhere out-of-the-way or you need to rewrite the .class's bytecode to have a different name. In either case, you'd probably then use something like `StaticTestingClassLoader` from #3241.

PR #3204, opened 2017-07-07
core: Change gRPC to use io.opencensus:opencensus-api:0.5.1.

bogdandrutu:


core/build.gradle
```diff
@@ -1,12 +1,15 @@
 description = 'gRPC: Core'
 
 dependencies {
-    compile libraries.guava,
+    compile project(':grpc-context'),
```
carl-mastrangelo: Could you please also update the Bazel files?

zhangkun83:
It doesn't seem opencensus has been imported yet. Shouldn't it happen first?

core/build.gradle
bogdandrutu: Done.

core/src/main/java/io/grpc/internal/CensusTracingModule.java
```diff
@@ -272,9 +272,8 @@ public ServerStreamTracer newServerStreamTracer(String fullMethodName, Metadata
     public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
         MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
       // New RPCs on client-side inherit the tracing context from the current Context.
```
adriancole: Add a comment here about impact of using an internal/unsafe type, or when it will not be unsafe?

core/src/main/java/io/grpc/internal/CensusTracingModule.java
```diff
@@ -60,14 +60,14 @@
...
-  private final BinaryPropagationHandler censusTracingPropagationHandler;
+  private final BinaryFormat censusTracingPropagationHandler;
   @VisibleForTesting
   final Metadata.Key<SpanContext> tracingHeader;
   private final TracingClientInterceptor clientInterceptor = new TracingClientInterceptor();
   private final ServerTracerFactory serverTracerFactory = new ServerTracerFactory();
 
   CensusTracingModule(
-      Tracer censusTracer, final BinaryPropagationHandler censusTracingPropagationHandler) {
+      Tracer censusTracer, final BinaryFormat censusTracingPropagationHandler) {
```
adriancole: Much nicer name

core/src/test/java/io/grpc/internal/CensusModulesTest.java
```diff
@@ -140,45 +140,22 @@ public String parse(InputStream stream) {
...
-      SpanContext.create(
-          TraceId.generateRandomId(random), SpanId.generateRandomId(random),
-          TraceOptions.builder().build());
-  private final Span fakeClientSpan = new FakeSpan(fakeClientSpanContext);
-  private final Span fakeServerSpan = new FakeSpan(fakeServerSpanContext);
-  private final Span fakeClientParentSpan = new FakeSpan(fakeClientParentSpanContext);
-  private final Span fakeServerParentSpan = new FakeSpan(fakeServerParentSpanContext);
-  private final Span spyClientSpan = spy(fakeClientSpan);
-  private final Span spyServerSpan = spy(fakeServerSpan);
+  private final Random random = new Random();
```
zhangkun83: A constant seed is preferred in tests as it yields reproducible results.

core/src/main/java/io/grpc/internal/CensusTracingModule.java
bogdandrutu: Done.

core/src/main/java/io/grpc/internal/CensusTracingModule.java
bogdandrutu: Done.

core/src/test/java/io/grpc/internal/CensusModulesTest.java
bogdandrutu: Done.

zhangkun83:
> It doesn't seem opencensus has been imported yet. Shouldn't it happen first?

It's imported, and I have verified the new APIs being called are already there.

@carl-mastrangelo @ejona86, any more concerns?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Jul 05 17:02:24 PDT 2017
+#Fri Jul 07 15:01:58 PDT 2017
```
carl-mastrangelo: Was this intentional?

ejona86: Please revert.

ejona86:
@zhangkun83, internally, it looks like they are still using the "instrumentation" target but with the new io.opencensus. That shouldn't cause us any problems, so I see no problem with this going in. (other than file that looks like it should be reverted)

gradle/wrapper/gradle-wrapper.properties
bogdandrutu: Done.

bogdandrutu: was not intentional, idea decided that it know better how to wrap using gradle.

PR #3205, opened 2017-07-07
Add support for specifying a custom hostname verifier when using on OkHttpChannelBuilder

JackOfMostTrades:
We use custom x509 extensions to encode multiple dimensions of server identity, and clients verify different dimensions based on individual client requirements. This is all coded up into HostnameVerifier implementations, but we need to ability to pass this to the the OkHttpChannelBuilder. I've implemented the ability to pass this through, defaulting to the existing OkHostnameVerifier implementation when unspecified.

### DIFF
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
@@ -48,6 +48,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
+import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSocketFactory;
 import javax.net.ssl.TrustManagerFactory;
@@ -103,6 +104,7 @@ public class OkHttpChannelBuilder extends
   private Executor transportExecutor;
 
   private SSLSocketFactory sslSocketFactory;
+  private HostnameVerifier hostnameVerifier;
   private ConnectionSpec connectionSpec = DEFAULT_CONNECTION_SPEC;
   private NegotiationType negotiationType = NegotiationType.TLS;
   private long keepAliveTimeNanos = KEEPALIVE_TIME_NANOS_DISABLED;
@@ -243,6 +245,31 @@ public class OkHttpChannelBuilder extends
   }
 
   /**
+   * Set the hostname verifier to use when using TLS negotiation. The hostnameVerifier is only used
+   * if using TLS negotiation. If the hostname verifier is not set, a default hostname verifier is
+   * used.
+   *
+   * <p>Be careful when setting a custom hostname verifier! By setting a non-null value, you are
+   * replacing all default verification behavior. If the hostname verifier you supply does not
+   * effectively supply the same checks, you may be removing the security assurances that TLS aims
+   * to provide.</p>
+   *
+   * <p>This method should not be used to avoid hostname verification, even during testing, since
+   * {@link #overrideAuthority} is a safer alternative as it does not disable any security checks.
+   * </p>
+   *
+   * @see io.grpc.okhttp.internal.OkHostnameVerifier
+   *
+   * @since 1.6.0
+   * @return this
+   *
+   */
+  public final OkHttpChannelBuilder hostnameVerifier(@Nullable HostnameVerifier hostnameVerifier) {
+    this.hostnameVerifier = hostnameVerifier;
+    return this;
+  }
+
+  /**
    * For secure connection, provides a ConnectionSpec to specify Cipher suite and
    * TLS versions.
    *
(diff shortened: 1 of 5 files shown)


ejona86:
okay to test

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -243,6 +245,17 @@ public final OkHttpChannelBuilder sslSocketFactory(SSLSocketFactory factory) {
   }
 
   /**
+   * Set the hostname verifier to use when using TLS negotiation. The hostnameVerifier is only used
+   * if an sslSocketFactory is set. If the hostname verifier is not set, a default hostname
+   * verifier is used.
+   * @see io.grpc.okhttp.internal.OkHostnameVerifier
+   */
+  public final OkHttpChannelBuilder hostnameVerifier(HostnameVerifier hostnameVerifier) {
```
carl-mastrangelo: Please consider adding `@Nullable`

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -243,6 +245,17 @@ public final OkHttpChannelBuilder sslSocketFactory(SSLSocketFactory factory) {
   }
 
   /**
+   * Set the hostname verifier to use when using TLS negotiation. The hostnameVerifier is only used
+   * if an sslSocketFactory is set. If the hostname verifier is not set, a default hostname
+   * verifier is used.
+   * @see io.grpc.okhttp.internal.OkHostnameVerifier
```
carl-mastrangelo: Would you please add a `@since 1.6.0`  and a `@returns this` ?

ejona86:
This is fair. What we don't want is for people to disable hostname verification, so during review I may ask for some docs strongly discouraging doing such things. I assume you could do things like certificate pinning to avoid user-aided MITM attacks and the like.

I'm leaving for the day, so not doing the actual review at the moment (although I see it is pretty trivial).

okhttp/src/test/java/io/grpc/okhttp/OkHttpClientTransportTest.java
```diff
@@ -183,7 +183,7 @@ public long read() {
   public void testToString() throws Exception {
     InetSocketAddress address = InetSocketAddress.createUnresolved("hostname", 31415);
     clientTransport = new OkHttpClientTransport(
-        address, "hostname", null /* agent */, executor, null,
+        address, "hostname", null /* agent */, executor, null, null,
```
carl-mastrangelo: please include an inline comment for what this null is.

JackOfMostTrades:
Yep, would definitely support some extra javadoc suggesting you avoid using this to shoot yourself in the foot... This let's you add a lot of extra security measures (certificate pinning being the most common one I've heard of), but unfortunately that means you can do silly things as well.

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -243,6 +245,27 @@ public final OkHttpChannelBuilder sslSocketFactory(SSLSocketFactory factory) {
   }
 
   /**
+   * Set the hostname verifier to use when using TLS negotiation. The hostnameVerifier is only used
+   * if an sslSocketFactory is set. If the hostname verifier is not set, a default hostname
+   * verifier is used.
+   *
+   * <p>Be careful when setting a custom hostname verifier! By setting a non-null value, you are
```
ejona86: Also add something like:

> `This method should not be used to avoid hostname verification, even during testing, since {@link #overrideAuthority} is a safer alternative as it does not disable any security checks.`

JackOfMostTrades: I liked your wording on that; went ahead and added that to the method doc verbatim.

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -243,6 +245,30 @@ public final OkHttpChannelBuilder sslSocketFactory(SSLSocketFactory factory) {
   }
 
   /**
+   * Set the hostname verifier to use when using TLS negotiation. The hostnameVerifier is only used
+   * if an sslSocketFactory is set. If the hostname verifier is not set, a default hostname
```
ejona86: > The hostnameVerifier is only used if an sslSocketFactory is set.

Actually, that's a bit misleading. You can use TLS without manually setting the `sslSocketFactory()`; grpc will just create one for you (in `createSocketFactory()`).

JackOfMostTrades: Oops, right you are; I missed that `createSocketFactory` makes one for you. I changed it to say that it's only used when using TLS negotiation, which I think is accurate.

JackOfMostTrades:
Thanks for the pointer to Http2OkHttpTest. Knowing where to utilize the existing testing framework is half the battle sometimes. ;) I got a couple of tests added in there.

ejona86:
@JackOfMostTrades, the CI tests are now failing. Sorry, no lambdas allowed :-).

Yeah, in this case I'd say finding testing framework is 90%. Http2OkHttpTest would have been hard for you to find, and the unit tests in okhttp/src/test would have been painful to adapt.

Thank you!

carl-mastrangelo:
@JackOfMostTrades   could you update the commit message to use the "package: what improved" format?  Something like:

oktthp:  add support for specifying a custom hostname verifier when using on OkHttpChannelBuilder

JackOfMostTrades:
> @JackOfMostTrades, the CI tests are now failing. Sorry, no lambdas allowed :-).
Oops. I have too much faith in my IDE to set the right language level on projects. ;) Tests updated.

> could you update the commit message to use the "package: what improved" format?

Should I collapse all the commits, or do you want me to just update the message first one?

carl-mastrangelo:
@JackOfMostTrades Squash, if you would.

dapengzhang0:
Jenkins, retest this please

Issue #3195, opened 2017-07-06
Implement compression interop tests


ericgribkoff:
There are four compression interop tests defined in [interop-test-descriptions.md](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md) but not implemented for Java:

- client_compressed_unary
- client_compressed_streaming
- server_compressed_unary
- server_compressed_streaming

These should be implemented and enabled in [run_interop_tests.py](https://github.com/grpc/grpc/blob/master/tools/run_tests/run_interop_tests.py).

ejona86:
server_compressed_unary and server_compressed_streaming cannot be implemented in Java, because the compressed message flag is not available and we have no want to expose such a feature. Last I heard, the cross-language expectation is that only C++ will implement those tests.

ericgribkoff:
The decision for now was to just implement these on the client-side, because, as @ejona86 points out, Java has no mechanism to implement the [CompressedRequest](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#compressedrequest) server-side feature of the interop test specification, which is necessary for server-side implementation of `client_compressed_unary` and `client_compressed_streaming` (in these tests, the client sends compressed/uncompressed messages and the server must verify the requests are compressed/uncompressed).

The client-side tests in PR #3211 include `server_compressed_unary` and `server_compressed_streaming`, designed to be run against a C++ server implementation that can respond appropriately to the `request_compressed` flag in the request messages. The Java clients cannot actually assert that the server's responses are compressed/uncompressed<sup>[1]</sup>, and https://github.com/grpc/grpc/pull/11739 updates the interop test descriptions to allow for this limitation.

***
[1]: Technically, the Java client could override  `MethodDescriptor.Marshaller#parse(InputStream stream)` to do a `stream instanceof KnownLength` to determine if the message was received compressed (does not implement `KnownLength`) or uncompressed (does implement `KnownLength`), but this is decidedly not an intended use of the API, and it does not add enough value to the test coverage to be worth supporting this type of use.

ericgribkoff:
Closed via #3211.

PR #3211, opened 2017-07-10
Implement client-side compression tests

ericgribkoff:
This adds the client-side compression tests for https://github.com/grpc/grpc-java/issues/3195.

The first commit brings `messages.proto` up to date with its corresponding file in grpc/grpc.

@ejona86 Filed https://github.com/grpc/grpc/pull/11739 to update the interop test descriptions to explicitly allow for implementations that cannot verify the received messages are truly compressed


ericgribkoff:
@makdharma  fyi

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -319,6 +320,107 @@ public void largeUnary() throws Exception {
...
+            .build();
+
+    // Send a non-compressed message with expectCompress=true. Servers supporting this test case
+    // should return INVALID_ARGUMENT.
+    try {
+      blockingStub.unaryCall(expectCompressedRequest);
+      fail("expected INVALID_ARGUMENT");
+    } catch (StatusRuntimeException e) {
+      assertEquals(Status.INVALID_ARGUMENT.getCode(), e.getStatus().getCode());
+      assertEquals("Compressed request expectation not met.", e.getStatus().getDescription());
```
ejona86: This is overly strict and should be removed. It is not specified to check this string, and the Server-side feature CompressedRequest does not mandate a particular status description. Ditto elsewhere.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -395,6 +497,90 @@ public void clientStreaming() throws Exception {
...
+    Throwable e = responseObserver.getError();
+    assertNotNull("expected INVALID_ARGUMENT", e);
+    assertEquals(Status.INVALID_ARGUMENT.getCode(), Status.fromThrowable(e).getCode());
+    assertEquals(
+        "Compressed request expectation not met.", Status.fromThrowable(e).getDescription());
+
+    // Start a new stream
+    responseObserver = StreamRecorder.create();
+    ClientCall<StreamingInputCallRequest, StreamingInputCallResponse> call =
+        channel.newCall(
```
ejona86: There should be no need to use `ClientCall` directly. Cast the stream observer to `ClientCallStreamObserver`, which has `setMessageCompression`.

ericgribkoff: Done

PR #3208, opened 2017-07-10
Make Kokoro show actual Gradle results

matt-kwong:
There were a couple of issues:
- `windows.bat` wasn't returning the Gradle exit code; it was returning an exit code related to the commands renaming the `*.xml` files
- `%%ERRORLEVEL%%` changed to `%ERRORLEVEL%`

After PR review, I'll remove the commit causing test failures, squash commits, and make sure the Kokoro build is green with these changes.

PR #3221, opened 2017-07-11
all: make Kokoro show actual Gradle results (backport)

matt-kwong:
Backport of #3208

PR #3224, opened 2017-07-11
all: make Kokoro show actual Gradle results (backport)

matt-kwong:
Backport of #3208


matt-kwong:
Could you also approve #3223 #3222 #3221 #3220 #3219? Only assigning one PR to avoid spamming notifcations

PR #3234, opened 2017-07-14
netty: quieter errors in NettyServerTransport

### LINKED ISSUE #1768: NettyServerTransport.notifyTransport() logs at SEVERE
[Logging at SEVERE](https://github.com/grpc/grpc-java/blob/a3c79e8/netty/src/main/java/io/grpc/netty/NettyServerTransport.java#L106) is probably an overkill most of the time. There are rare cases that something higher would be useful, but I don't think we could easily distinguish between those cases. We should probably reduce the severity to something lower, although I'm sure what makes the most sense.

zpencer:
Log all exceptions at INFO rather than SEVERE.
Known uninteresting exceptions should be FINE.

fixes #1768

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServerTransport.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerTransport.java
@@ -17,6 +17,7 @@
 package io.grpc.netty;
 
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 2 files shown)


zpencer:
@carl-mastrangelo please take a look

PR #3235, opened 2017-07-14
stub: clean up client calls to avoid synthetic class

carl-mastrangelo:


stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -47,7 +47,8 @@
  * that the runtime can vary behavior without requiring regeneration of the stub.
  */
 public final class ClientCalls {
-  private static final Logger log = Logger.getLogger(ClientCalls.class.getName());
+  // Non private to avoid synthetic class
+  static final Logger log = Logger.getLogger(ClientCalls.class.getName());
```
zhangkun83: Maybe just move this logger to the class where it is used?

stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -329,13 +332,15 @@ public void setMessageCompression(boolean enable) {
     }
   }
 
-  private static class StreamObserverToCallListenerAdapter<ReqT, RespT>
+  // Non private to avoid synthetic class
+  static final class StreamObserverToCallListenerAdapter<ReqT, RespT>
```
zhangkun83: This is only used in static methods of the outer class. I thought outer class have direct access to its private inner classes, thus only the constructor needed to be non-private. Isn't it the case?

stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -392,11 +397,12 @@ public void onReady() {
   /**
    * Complete a GrpcFuture using {@link StreamObserver} events.
    */
-  private static class UnaryStreamToFuture<RespT> extends ClientCall.Listener<RespT> {
+  // Non private to avoid synthetic class
+  static final class UnaryStreamToFuture<RespT> extends ClientCall.Listener<RespT> {
```
zhangkun83: ditto

stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -536,7 +544,8 @@ public void remove() {
       throw new UnsupportedOperationException();
     }
 
-    private class QueuingListener extends ClientCall.Listener<T> {
+    // Non private to avoid synthetic class
+    final class QueuingListener extends ClientCall.Listener<T> {
```
zhangkun83: Ditto. This is only used by its outer class.

stub/src/main/java/io/grpc/stub/ClientCalls.java
carl-mastrangelo: Done.

stub/src/main/java/io/grpc/stub/ClientCalls.java
carl-mastrangelo: You are right.  I have made this change.

carl-mastrangelo:
@zhangkun83 PTAL

carl-mastrangelo:
Jenkins retest this please

PR #3239, opened 2017-07-14
testing-proto: add string message to testing proto

dapengzhang0:
In migrating internal tests to use testing protos, I found it was not convenient to test in some scenarios without having a field in the message. All `SimpleResponse` messages are `equals` to each other. Users might not want to assume reference identity/inequality between server and client for equal/unequal messages even for InProcess. Even with reference identity, it would still be inconvenient if their were hundreds of messages.

To have a  non-proto based String to String service instead is another solution, but the usefulness of this proto based service would still be too limited.


ejona86:
Please include a _reason_ for the change in the commit message. The title and description should include what the change is, but it's equally important to include in the description _why_ the change is being made (although this is generally not best in the commit title (the first line)).

PR #3241, opened 2017-07-14
core: Use Class.forName(String) in provider for Android

### LINKED ISSUE #2633: Do not pass ClassLoader to Class.forName, for Android
See #2207. An easy workaround for #2207 was to specify -keep in ProGuard configuration. However, everyone would be happier if such configuration was unnecessary. Using `forName()` without passing `ClassLoader` should remove the need for configuration, as originally intended.

ejona86:
Class.forName(String) is understood by ProGuard, removing the need for
manual ProGuard configuration and allows ProGuard to rename the provider
classes. Previously the provider classes could not be renamed.

Fixes #2633

### DIFF
--- a/core/src/test/java/io/grpc/StaticTestingClassLoader.java
+++ b/core/src/test/java/io/grpc/StaticTestingClassLoader.java
@@ -39,10 +39,6 @@ public final class StaticTestingClassLoader extends ClassLoader {
     if (!pattern.matcher(name).matches()) {
       throw new ClassNotFoundException(name);
     }
-    return findClass0(name);
-  }
-
-  private Class<?> findClass0(String name) throws ClassNotFoundException {
     InputStream is = getResourceAsStream(name.replace('.', '/') + ".class");
     if (is == null) {
       throw new ClassNotFoundException(name);
@@ -58,10 +54,15 @@ public final class StaticTestingClassLoader extends ClassLoader {
 
   @Override
   protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
+    // Reverse normal loading order; check this class loader before its parent
     synchronized (getClassLoadingLock(name)) {
       Class<?> klass = findLoadedClass(name);
-      if (klass == null && pattern.matcher(name).matches()) {
-        klass = findClass0(name);
+      if (klass == null) {
+        try {
+          klass = findClass(name);
+        } catch (ClassNotFoundException e) {
+          // This ClassLoader doesn't know a class with that name; that's part of normal operation
+        }
       }
       if (klass == null) {
         klass = super.loadClass(name, false);
--- a/core/src/test/java/io/grpc/StaticTestingClassLoader.java
+++ b/core/src/test/java/io/grpc/StaticTestingClassLoader.java
@@ -27,16 +27,16 @@ import java.util.regex.Pattern;
  * instance is required per test.
  */
 public final class StaticTestingClassLoader extends ClassLoader {
-  private final Pattern pattern;
+  private final Pattern classesToDefine;
 
-  public StaticTestingClassLoader(ClassLoader parent, Pattern pattern) {
+  public StaticTestingClassLoader(ClassLoader parent, Pattern classesToDefine) {
     super(parent);
-    this.pattern = Preconditions.checkNotNull(pattern, "pattern");
+    this.classesToDefine = Preconditions.checkNotNull(classesToDefine, "classesToDefine");
   }
 
   @Override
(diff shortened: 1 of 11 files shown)


ericgribkoff:
Looking at this now, but FYI`NettyChannelProviderTest#getCandidatesViaHardCoded` and `OkHttpChannelProviderTest#getCandidatesViaHardCoded` need to have the class loader argument removed from their calls to `getCandidatesViaHardCoded`

core/src/test/java/io/grpc/StaticTestingClassLoader.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.regex.Pattern;
+
+/**
+ * A class loader that can be used to repeatedly trigger static initialization of a class. A new
+ * instance is required per test.
+ */
+public final class StaticTestingClassLoader extends ClassLoader {
+  private final Pattern pattern;
```
ericgribkoff: I think a more descriptive name than `pattern` would be useful here. Maybe `classesToDefine`?

core/src/test/java/io/grpc/StaticTestingClassLoader.java
```diff
@@ -0,0 +1,75 @@
...
+
+  public StaticTestingClassLoader(ClassLoader parent, Pattern pattern) {
+    super(parent);
+    this.pattern = Preconditions.checkNotNull(pattern, "pattern");
+  }
+
+  @Override
+  protected Class<?> findClass(String name) throws ClassNotFoundException {
```
ericgribkoff: Why override this method? The tests pass with this method removed, and its presence made the functionality here harder for me to understand.

core/src/test/java/io/grpc/StaticTestingClassLoader.java
```diff
@@ -0,0 +1,75 @@
...
+package io.grpc;
+
+import com.google.common.base.Preconditions;
+import io.grpc.internal.IoUtils;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.regex.Pattern;
+
+/**
+ * A class loader that can be used to repeatedly trigger static initialization of a class. A new
```
ericgribkoff: I'm not really sure of the significance of "static initialization" here versus just initialization. My understanding, which might be incomplete, is that this is just attempting class loading for classes matching `pattern`, and deferring everything else to `parent`. Would "TestingClassLoader" be a sufficient name for the class, or is there some importance of "Static" that I'm missing?

core/src/test/java/io/grpc/StaticTestingClassLoader.java
ejona86: ClassLoader (the base class) normally calls `findClass()` from `loadClass()`. Because our overridden `loadClass()` calls `findClass0()`, it's effectively unnecessary. However, calling `findClass0()` is just an optimization; I'd probably get rid of `findClass0()` (and inline the contents into `findClass()` instead of removing the `findClass()` override.

To most closely follow the normal "protocol", I could call `findClass()` with a try-catch and `catch (ClassNotFoundException)`; that's what's normally done (although in the opposite order, which would reduce the number of exceptions thrown). But throwing the exception just to catch it seemed annoying at the time. But that's probably what I should do; it simplifies things. Done.

core/src/test/java/io/grpc/StaticTestingClassLoader.java
ejona86: Done.

core/src/test/java/io/grpc/StaticTestingClassLoader.java
ejona86: "initialization" could simply mean what happens during creating a class instance: new Object() initializes an Object. So I'm really referring to "class initialization", but the only way that is really useful to test code is to trigger static code blocks and static field initializers to run. This can also be used if you are mutating static state after initialization. So that's why "Static" is important, as it describes its use-case.

I honestly tried pretty hard to come up with another name, but came up short. I mainly wanted to name it based on what it did, and only secondly based on how it was used. But I found it pretty hard. "TestingClassLoader" is a very weak name in my mind, since it would apply equally well to [ReplacingClassLoader](https://github.com/grpc/grpc-java/blob/v1.5.0/core/src/test/java/io/grpc/ReplacingClassLoader.java) (wow, it's Javadoc is bad...). I considered things like "RedefiningClassLoader" and "CopyingClassLoader" and more. And then I gave up and moved on with life.

PR #3242, opened 2017-07-15
benchmarks: don't block shutdown when using netty epoll

carl-mastrangelo:


benchmarks/src/main/java/io/grpc/benchmarks/qps/AsyncServer.java
```diff
@@ -109,10 +112,11 @@ static Server newServer(ServerConfiguration config) throws IOException {
     final EventLoopGroup boss;
     final EventLoopGroup worker;
     final Class<? extends ServerChannel> channelType;
+    ThreadFactory tf = new DefaultThreadFactory("server-elg-", true /*daemon */);
     switch (config.transport) {
       case NETTY_NIO: {
-        boss = new NioEventLoopGroup();
-        worker = new NioEventLoopGroup();
+        boss = new NioEventLoopGroup(0, tf);
```
ejona86: The bosses should probably just have one thread.

benchmarks/src/main/java/io/grpc/benchmarks/qps/AsyncServer.java
```diff
@@ -82,6 +84,7 @@ public void run() {
         }
       }
     });
+    server.awaitTermination();
```
ejona86: You can delete the awaitTermination in the shutdownHook.

carl-mastrangelo: Done.

benchmarks/src/main/java/io/grpc/benchmarks/qps/AsyncServer.java
carl-mastrangelo: Done

carl-mastrangelo:
I'll update the desc when I squash

PR #3247, opened 2017-07-17
Clarify docs/examples: ServerCall, interceptors, flow control

zpencer:
docs: small clarification on why blocking is fine
interceptors: create explicit variables to be a little more clear what's going on
RouteGuideServer: demonstrate how to use basic flow control on the server streaming method


zpencer:
No longer needed because we now have a detailed flow control example

PR #3255, opened 2017-07-18
all: update animal sniffer and error prone

carl-mastrangelo:


carl-mastrangelo:
Seems to add some caching: https://github.com/xvik/gradle-animalsniffer-plugin/releases/tag/1.4.0

ejona86:
Note that the animalsniffer caching is disabled by default.

PR #3257, opened 2017-07-19
Revert interop-testing: overrideAuthority breaks JWT

ejona86:
This reverts commit 57b9105c7f14b52a3ed132302af1af59fc869281.

Issue #2682 is fixed, so we can revert the commit as planned. This
re-applies a previously-reverted modernization.


ejona86:
retest this please

PR #3262, opened 2017-07-19
Context: add test for the scenario that Context's logger itself is using Context

dapengzhang0:
This will break if 2f6e2c87ab1aeda591fcb7764a8ad6bb20855a2a & a3a5420922ecf1ddc2c4ab77c273be3df47b74f5 are reverted:

```
io.grpc.ContextTest > initContextWithCustomClassLoaderWithCustomLogger FAILED
    java.lang.ExceptionInInitializerError
        at io.grpc.ContextTest$LoadMeWithStaticTestingClassLoader.run(ContextTest.java:789)
        at io.grpc.ContextTest.initContextWithCustomClassLoaderWithCustomLogger(ContextTest.java:760)

        Caused by:
        java.lang.RuntimeException: Storage override had failed to initialize
            at io.grpc.Context.storage(Context.java:134)
            at io.grpc.Context.current(Context.java:163)
            at io.grpc.ContextTest$LoadMeWithStaticTestingClassLoader$1.publish(ContextTest.java:773)
            at java.util.logging.Logger.log(Logger.java:738)
            at java.util.logging.Logger.doLog(Logger.java:765)
            at java.util.logging.Logger.log(Logger.java:875)
            at io.grpc.Context.<clinit>(Context.java:122)
            ... 2 more
```

### DIFF
--- a/context/build.gradle
+++ b/context/build.gradle
@@ -2,5 +2,6 @@ description = 'gRPC: Context'
 
 dependencies {
     testCompile project(':grpc-testing')
+    testCompile project(':grpc-core').sourceSets.test.output
     signature "org.codehaus.mojo.signature:java16:1.1@signature"
 }
--- a/context/src/test/java/io/grpc/ContextTest.java
+++ b/context/src/test/java/io/grpc/ContextTest.java
@@ -45,6 +45,7 @@ import java.util.logging.Handler;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import java.util.logging.Logger;
+import java.util.regex.Pattern;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
@@ -56,30 +57,6 @@ import org.junit.runners.JUnit4;
  */
 @RunWith(JUnit4.class)
 public class ContextTest {
-
-  // The following logger setup is to test the scenario when the logger itself is using Context,
-  // the Context/Storage initialization shouldn't fail.
-  private static final Logger logger = Logger.getLogger(Context.class.getName());
-
-  static {
-    logger.setLevel(Level.ALL);
-    logger.addHandler(new Handler() {
-      @Override
-      public void publish(LogRecord record) {
-        Context ctx = Context.current();
-        Context previous = ctx.attach();
-        ctx.detach(previous);
-      }
-
-      @Override
-      public void flush() {
-      }
-
-      @Override
-      public void close() throws SecurityException {
-      }
-    });
-  }
 
   private static final Context.Key<String> PET = Context.key("pet");
   private static final Context.Key<String> FOOD = Context.keyWithDefault("food", "lasagna");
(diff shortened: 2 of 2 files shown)


ejona86:
This test is brittle, as it assumes test functionality and order (not within the test class, but between test classes). You could use [`StaticTestingClassLoader`](https://github.com/grpc/grpc-java/blob/master/core/src/test/java/io/grpc/StaticTestingClassLoader.java) though.

dapengzhang0:
@ejona86, I try to use `StaticTestingClassLoader`:

```
  @Test
  public void test1() throws Exception {
    StaticTestingClassLoader classLoader =
        new StaticTestingClassLoader(
            getClass().getClassLoader(), Pattern.compile("io\\.grpc\\.Context*"));

    Class<?> contextClass = classLoader.loadClass("io.grpc.Context");
    assertNotNull(contextClass.getDeclaredField("ROOT").get(null));
  }
```

but I'm getting this error:
```
io.grpc.ContextTest > test1 FAILED
    java.lang.IllegalAccessError: tried to access method io.grpc.Context$Key.<init>(Ljava/lang/String;)V from class io.grpc.Context
        at io.grpc.Context.<clinit>(Context.java:101)
        at sun.misc.Unsafe.ensureClassInitialized(Native Method)
        at sun.reflect.UnsafeFieldAccessorFactory.newFieldAccessor(UnsafeFieldAccessorFactory.java:43)
        at sun.reflect.ReflectionFactory.newFieldAccessor(ReflectionFactory.java:142)
        at java.lang.reflect.Field.acquireFieldAccessor(Field.java:1088)
        at java.lang.reflect.Field.getFieldAccessor(Field.java:1069)
        at java.lang.reflect.Field.get(Field.java:393)
        at io.grpc.ContextTest.test1(ContextTest.java:757)

```

ejona86:
You need `.*` instead of `*`: `io\\.grpc\\.Context.*`

dapengzhang0:
oops, thanks @ejona86

context/build.gradle
```diff
@@ -2,5 +2,6 @@ description = 'gRPC: Context'
 
 dependencies {
     testCompile project(':grpc-testing')
+    testCompile project(':grpc-core').sourceSets.test.output
```
ejona86: Maybe move StaticTestingClassLoader to context and have core depend on context's test?

context/src/test/java/io/grpc/ContextTest.java
```diff
@@ -745,4 +747,53 @@ public void expiredDeadlineShouldCancelContextImmediately() {
...
+            getClass().getClassLoader(),
+            Pattern.compile("(io\\.grpc\\.Context.*)|(io\\.grpc\\.ThreadLocalContextStorage.*)"));
+    Class<?> runnable =
+        classLoader.loadClass("io.grpc.ContextTest$LoadMeWithStaticTestingClassLoader");
+
+    ((Runnable) runnable.getDeclaredConstructor().newInstance()).run();
+  }
+
+  // UsedReflectively
+  @VisibleForTesting
```
ejona86: VisibleForTesting is a strange annotation here. I'd get rid of it and just have the UsedReflectively.

context/build.gradle
dapengzhang0: `StaticTestingClassLoader` is primarily used for testing ServiceProvider which is not related to `Context`, maybe move it to testing/io.grpc.interal.testing later?

dapengzhang0:
I did a `commit --amend` and `push -f`.

context/build.gradle
ejona86: I sort of wish this didn't depend on grpc-testing already... The only reason it does now is for DeadlineSubject in a single test case.

PR #3263, opened 2017-07-21
benchmark: context benchmark

zpencer:
Here is a gist of a sample run:
https://gist.github.com/zpencer/b778a10563ff2d3958c5c503684dd749


benchmarks/src/jmh/java/io/grpc/context/ReadMissingKeyBenchmark.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  /**
+   * Missing keys are typically the slow path for Contexts.
+   */
+  @Benchmark
+  @BenchmarkMode(Mode.SampleTime)
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  public void readMissingKey() throws ExecutionException, InterruptedException {
+    missingKey.get(contextChain);
```
ejona86: You need to consume the return value of get(), otherwise the JIT can heavily optimize this. Easiest way is simply to return the value.

benchmarks/src/jmh/java/io/grpc/context/ReadMissingKeyBenchmark.java
```diff
@@ -0,0 +1,127 @@
...
+  public void setUp() throws ExecutionException, InterruptedException {
+    final List<Key<Integer>> keys = makeKeys(numKeys);
+    final List<List<Key<Integer>>> keyBatches = Lists.partition(keys, CONTEXT_MAX_BATCH_SIZE);
+    for (int i = 0; i < putIterations; i++) {
+      if (batched) {
+        for (Key<Integer> key : keys) {
+          contextChain = contextChain.withValue(key, DUMMY_VAL);
+        }
+      } else {
+        for (List<Key<Integer>> batch : keyBatches) {
```
carl-mastrangelo: inline `keyBatches` ?

benchmarks/src/jmh/java/io/grpc/context/ReadMissingKeyBenchmark.java
```diff
@@ -0,0 +1,127 @@
...
+            batch.get(1), val,
+            batch.get(2), val);
+      case 2:
+        return c.withValues(
+            batch.get(0), val,
+            batch.get(1), val);
+      case 1:
+        return c.withValue(batch.get(0), val);
+      default:
+        throw new RuntimeException();
```
carl-mastrangelo: AssertionError

benchmarks/src/jmh/java/io/grpc/context/ReadWriteBenchmark.java
```diff
@@ -0,0 +1,192 @@
...
+  /**
+   * Missing keys are typically the slow path for Contexts.
+   */
+  @Benchmark
+  @BenchmarkMode(Mode.SampleTime)
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  public int readWrite() throws ExecutionException, InterruptedException {
+    int dummy = 0;
+    Context c = Context.ROOT;
+    for (int i = 0; i < iterations; i++) {
```
ejona86: This is suspicious, in that it couples depth of tree with number of different tests (each iteration is actually its own test of sorts, and only the total sum is reported ta the end) with number of gets. I don't think the result of this allows us to make any conclusions.

zpencer:
Changed the code to be the context benchmark authored by @lukesandberg, to get a basic benchmark into the codebase

zpencer:
@ejona86 please take a look

benchmarks/src/jmh/java/io/grpc/context/WriteBenchmark.java
```diff
@@ -0,0 +1,86 @@
...
+      batchKeys = Lists.partition(keys, 4);
+    }
+  }
+
+  @Benchmark
+  @BenchmarkMode(Mode.SampleTime)
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  public void testContextWrite(ContextState state, Blackhole bh) {
+    // always reset Context to avoid unbounded memory usage
+    state.context = Context.ROOT;
```
ejona86: Why are you assigning to `state.context` at all in this method?

benchmarks/src/jmh/java/io/grpc/context/WriteBenchmark.java
```diff
@@ -0,0 +1,86 @@
...
+    if (state.batched) {
+      for (Context.Key<Object> key : state.keys) {
```
ejona86: I still find this looping strange, in that they are benchmarking something that doesn't quite seem to matter. It's seem much more valuable to know the cost of modification with a pre-existing context that has a chain of length X.

I feel like setup() should create the context and then this method simply modifies it once. That looks _very_ similar to ReadBenchmark; I'd expect one class that prepares a Context chain in @Setup and has a couple different methods to check the performance of operations with that chain.

benchmarks/src/jmh/java/io/grpc/context/WriteBenchmark.java
```diff
@@ -0,0 +1,86 @@
...
+        keys.add(Context.key("Key" + i));
+      }
+      batchKeys = Lists.partition(keys, 4);
+    }
+  }
+
+  @Benchmark
+  @BenchmarkMode(Mode.SampleTime)
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  public void testContextWrite(ContextState state, Blackhole bh) {
```
ejona86: It's weird to have a benchmark method starting with 'test'

zpencer: Renamed

benchmarks/src/jmh/java/io/grpc/context/WriteBenchmark.java
zpencer: Removed this part of the code after simplifying the test

benchmarks/src/jmh/java/io/grpc/context/WriteBenchmark.java
zpencer: Changed the benchmark to only do a simple mutation with each iteration

benchmarks/src/jmh/java/io/grpc/context/WriteBenchmark.java
```diff
@@ -17,70 +14,36 @@
...
-        return c.withValues(
-            batch.get(0), val,
-            batch.get(1), val);
-      case 1:
-        return c.withValue(batch.get(0), val);
-      default:
-        throw new AssertionError("Batch size should be between 1 and 4");
-    }
+  public void doWrite(ContextState state, Blackhole bh) {
+    Context result = state.context.withValues(
```
ejona86: Return the result instead of using blackhole.

PR #3270, opened 2017-07-24
core,netty,okhttp: remove 2-arg ClientTransport.newStream()

dapengzhang0:
to keep only one newStream() method in the `ClientTransport` interface


dapengzhang0:
Jenkins, retest this please

PR #3275, opened 2017-07-25
context: Create TimeoutException only when deadline occurs

### LINKED ISSUE #3272: Request with deadline cause to many GC overhead
1.4.0 Release

java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)

### What the problem?
screenshot from 2017-07-25 14 08 27

When uses deadline on each client call, gRPC creates TimeouOutExcepton before real deadline 
occurs https://github.com/grpc/grpc-java/blob/master/context/src/main/java/io/grpc/Context.java#L703 
it's very expensive operation that also allocate not necessary junk for GC. In screen above you can see
That during 60 second gRPC allocate 786mb memory on fillStackTrace. 

What do you think about move creation exception to place when it's really necessary? Yes in this case stacktrace will be different in case when deadline already expired and when deadline occurs after some time, but is it really make sense?

Gordiychuk:
Use deadline on client calls lead to create TimeoutException even
if deadline not occurs yet. `fillStacktrack` very expensive operation
that allocate many unnecessary objects in heap. Now exception
creates only when deadline occurs.

Close #3272

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -700,7 +700,6 @@ public class Context {
         ScheduledExecutorService scheduler) {
       super(parent, deriveDeadline(parent, deadline), true);
       if (DEADLINE_KEY.get(this) == deadline) {
-        final TimeoutException cause = new TimeoutException("context timed out");
         if (!deadline.isExpired()) {
           // The parent deadline was after the new deadline so we need to install a listener
           // on the new earlier deadline to trigger expiration for this context.
@@ -708,7 +707,7 @@ public class Context {
             @Override
             public void run() {
               try {
-                cancel(cause);
+                cancel(new TimeoutException("context timed out"));
               } catch (Throwable t) {
                 log.log(Level.SEVERE, "Cancel threw an exception, which should not happen", t);
               }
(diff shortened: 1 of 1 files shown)


ejona86:
okay to test

Gordiychuk:
I signed it!

context/src/main/java/io/grpc/Context.java
```diff
@@ -700,23 +700,22 @@ private CancellableContext(Context parent, Deadline deadline,
...
-        final TimeoutException cause = new TimeoutException("context timed out");
         if (!deadline.isExpired()) {
           // The parent deadline was after the new deadline so we need to install a listener
           // on the new earlier deadline to trigger expiration for this context.
           pendingDeadline = deadline.runOnExpiration(new Runnable() {
             @Override
             public void run() {
               try {
-                cancel(cause);
+                cancel(new TimeoutException("context timed out"));
```
carl-mastrangelo: If there were a lot of timeouts happening, this would slow things down.  Maybe this coudl be a helper function that laziliy creates the timeout?

ejona86: This creates strictly fewer `TimeoutExceptions` compared to before; it was use-once before as well.

Yes, we could use a shared `TimeoutException` with its call stack thrown away (like we've seen in Netty), but that is a different sort of optimization, is unclear is beneficial, and shouldn't block this change in my mind.

ejona86:
Thanks @Gordiychuk!

(I'll also note, that your duck's raised wings produced a nice effect with the "I signed it!" :smile:)

PR #3276, opened 2017-07-26
core: pass CallOptions to newClientStreamTracer().

### LINKED ISSUE #3256: Passing CallOptions to ClientStreamTracer.newClientStreamTracer
Any thoughts on passing the CallOptions into ClientStreamTracer.newClientStreamTracer?  We have an interceptor chain with roughly 10 interceptors.  The first interceptor sets a MetricsContext and our internal implementation of ClientStreamTracer.Factory on the CallOptions.  Each subsequent interceptor may add tags to the MetricsContext.  MetricsContext is hierarchical so we can have sub contexts for things like retries and hedged requests.  We'd need access the MetricsContext from our ClientStreamTracer implementation.  The contract would look something like, 

```java
public abstract ClientStreamTracer newClientStreamTracer(CallOptions callOptions, Metadata headers);
```

Alternatively, we could track our MetricsContext as described above but only specify the ClientStreamTracer.Factory in the last interceptor.  I'm not sure this approach will work long term as it sounds like your guys are leaning towards setting the ClientStreamTracer.Factory on the channel as opposed to the CallOptions (which I like as it would be more consistent with how things are done on the server).

zhangkun83:
Resolves #3256

/cc @elandau @ejona86

### DIFF
--- a/core/src/main/java/io/grpc/ClientStreamTracer.java
+++ b/core/src/main/java/io/grpc/ClientStreamTracer.java
@@ -43,10 +43,23 @@ public abstract class ClientStreamTracer extends StreamTracer {
     /**
      * Creates a {@link ClientStreamTracer} for a new client stream.
      *
+     * @deprecated Override/call {@link #newClientStreamTracer(CallOptions, Metadata)} instead. 
+     */
+    @Deprecated
+    public ClientStreamTracer newClientStreamTracer(Metadata headers) {
+      throw new UnsupportedOperationException("This method will be deleted. Do not call.");
+    }
+
+    /**
+     * Creates a {@link ClientStreamTracer} for a new client stream.
+     *
      * @param headers the mutable headers of the stream. It can be safely mutated within this
      *        method.  It should not be saved because it is not safe for read or write after the
      *        method returns.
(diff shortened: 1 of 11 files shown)


core/src/main/java/io/grpc/ClientStreamTracer.java
```diff
@@ -43,10 +43,23 @@ public void inboundHeaders() {
...
+    /**
+     * Creates a {@link ClientStreamTracer} for a new client stream.
+     *
+     * @param callOptions the effective CallOptions of the call
      * @param headers the mutable headers of the stream. It can be safely mutated within this
      *        method.  It should not be saved because it is not safe for read or write after the
      *        method returns.
      */
-    public abstract ClientStreamTracer newClientStreamTracer(Metadata headers);
+    public ClientStreamTracer newClientStreamTracer(CallOptions callOptions, Metadata headers) {
```
ejona86: Doesn't this need `@SuppressWarnings("deprecated")`?

zhangkun83: Done, although it's strange that `gradle check` didn't show any warning.

PR #3285, opened 2017-07-27
netty: support `status()` on Headers

carl-mastrangelo:
Recent Netty change https://github.com/netty/netty/commit/a91df58ca17d5b30c57c46dde5b1d60bb659b029
caused the `status()` method to be invoked, which AbstractHttp2Headers does not implement.
This change is necesary to upgrade to Netty 4.1.14

### DIFF
--- a/netty/src/main/java/io/grpc/netty/GrpcHttp2OutboundHeaders.java
+++ b/netty/src/main/java/io/grpc/netty/GrpcHttp2OutboundHeaders.java
@@ -70,8 +70,8 @@ final class GrpcHttp2OutboundHeaders extends AbstractHttp2Headers {
   @SuppressWarnings("ReferenceEquality") // STATUS.value() never changes.
   public CharSequence status() {
     // preHeaders is never null.  It has status as the first element or not at all.
-    if (preHeaders.length >= 2 && preHeaders[0] == Http2Headers.PseudoHeaderName.STATUS.value()) {
-      return preHeaders[1];
+    if (preHeaders.length >= 1 && preHeaders[0] == Http2Headers.PseudoHeaderName.STATUS.value()) {
+      return Utils.STATUS_OK;
     }
     return null;
   }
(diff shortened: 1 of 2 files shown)


netty/src/main/java/io/grpc/netty/GrpcHttp2OutboundHeaders.java
```diff
@@ -67,6 +67,16 @@ private GrpcHttp2OutboundHeaders(AsciiString[] preHeaders, byte[][] serializedMe
   }
 
   @Override
+  @SuppressWarnings("ReferenceEquality") // STATUS.value() never changes.
+  public CharSequence status() {
+    // preHeaders is never null.  It has status as the first element or not at all.
+    if (preHeaders.length >= 1 && preHeaders[0] == Http2Headers.PseudoHeaderName.STATUS.value()) {
+      return Http2Headers.PseudoHeaderName.STATUS.value();
```
ejona86: Shouldn't this be `preHeaders[1]`?

carl-mastrangelo: Interestingly yes, but it doesn't make the test fail.  The reason this method is called is to check for 1XX status codes, which we don't use.

netty/src/main/java/io/grpc/netty/GrpcHttp2OutboundHeaders.java
```diff
@@ -67,6 +67,16 @@ private GrpcHttp2OutboundHeaders(AsciiString[] preHeaders, byte[][] serializedMe
   }
 
   @Override
+  @SuppressWarnings("ReferenceEquality") // STATUS.value() never changes.
+  public CharSequence status() {
+    // preHeaders is never null.  It has status as the first element or not at all.
+    if (preHeaders.length >= 1 && preHeaders[0] == Http2Headers.PseudoHeaderName.STATUS.value()) {
+      return Utils.STATUS_OK;
```
ejona86: Please don't hard-code it to this degree.

carl-mastrangelo: Done.

PR #3287, opened 2017-07-27
core: fix the race between channel shutdown and clientCallImpl start

dapengzhang0:
This PR tries to fix the race described and updated in #1981 by introducing a ReadWriteLock.

Tests will be added if the approach is acceptable.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ClientCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ClientCallImpl.java
@@ -48,8 +48,7 @@ import io.grpc.Status;
 import java.io.InputStream;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.Executor;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -65,7 +64,7 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
   private final MethodDescriptor<ReqT, RespT> method;
   private final Executor callExecutor;
   private final Context context;
-  private volatile ScheduledFuture<?> deadlineCancellationFuture;
+  private volatile Future<?> deadlineCancellationFuture;
   private final boolean unaryRequest;
   private final CallOptions callOptions;
   private ClientStream stream;
@@ -74,14 +73,14 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
   private boolean halfCloseCalled;
   private final ClientTransportProvider clientTransportProvider;
   private final CancellationListener cancellationListener = new ContextCancellationListener();
-  private ScheduledExecutorService deadlineCancellationExecutor;
+  private final DeadlineHandler deadlineHandler;
   private DecompressorRegistry decompressorRegistry = DecompressorRegistry.getDefaultInstance();
   private CompressorRegistry compressorRegistry = CompressorRegistry.getDefaultInstance();
 
   ClientCallImpl(
       MethodDescriptor<ReqT, RespT> method, Executor executor, CallOptions callOptions,
       ClientTransportProvider clientTransportProvider,
-      ScheduledExecutorService deadlineCancellationExecutor) {
+      DeadlineHandler deadlineHandler) {
     this.method = method;
     // If we know that the executor is a direct executor, we don't need to wrap it with a
     // SerializingExecutor. This is purely for performance reasons.
@@ -95,7 +94,7 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
         || method.getType() == MethodType.SERVER_STREAMING;
     this.callOptions = callOptions;
     this.clientTransportProvider = clientTransportProvider;
-    this.deadlineCancellationExecutor = deadlineCancellationExecutor;
+    this.deadlineHandler = deadlineHandler;
   }
 
   private final class ContextCancellationListener implements CancellationListener {
@@ -115,6 +114,17 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
      * @param args object containing call arguments.
      */
     ClientTransport get(PickSubchannelArgs args);
+  }
+
+
+  /**
+   * Handler for RPC deadline.
+   */
+  interface DeadlineHandler {
+    /**
+     * Schedules a task for deadline cancellation.
+     */
+    Future<?> scheduleDeadlineCancellation(Runnable command, Deadline deadline);
   }
 
   ClientCallImpl<ReqT, RespT> setDecompressorRegistry(DecompressorRegistry decompressorRegistry) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -310,11 +330,52 @@ public void run() {
...
+        }
+
+        @Override
+        public int compareTo(Delayed delayed) {
+          throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public boolean cancel(boolean mayInterruptIfRunning) {
+          return true;
```
ejona86: should return `false` if the future is already cancelled.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -310,11 +330,52 @@ public void run() {
     }
   }
 
-  private ScheduledFuture<?> startDeadlineTimer(Deadline deadline) {
+  private ScheduledFuture<?> startDeadlineTimer(final Deadline deadline) {
```
ejona86: It looks like we don't use any part of the `ScheduledFuture` interface. Maybe just return `Future<?>` and then use a pre-existing Future implementation?

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -310,11 +330,52 @@ public void run() {
...
+        }
+
+        @Override
+        public boolean isCancelled() {
+          return true;
+        }
+
+        @Override
+        public boolean isDone() {
+          return false;
```
ejona86: should return `true` if the future is cancelled.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -310,11 +330,52 @@ public void run() {
...
+          throw new CancellationException();
+        }
+
+        @Override
+        public Void get(long timeout, TimeUnit unit) {
+          throw new CancellationException();
+        }
+      }
+
+      return new CancelledFuture();
```
ejona86: I don't feel comfortable silently ignoring the deadline. That seems just asking for undiagnosable run-away RPCs. The graceful way to handle this would be to fail the RPC.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -149,6 +151,26 @@ public void start(final Listener<RespT> observer, Metadata headers) {
     checkNotNull(observer, "observer");
     checkNotNull(headers, "headers");
 
+    if (deadlineCancellationExecutor == null) {
```
ejona86: I'm not excited about having a very different code path during the race vs after (with this case being after). It makes it all the more difficult to find bugs during the race.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -554,6 +555,32 @@ public String authority() {
     }
   }
 
+  private final class DirectFallbackExecutor implements Executor {
```
ejona86: This is quite an interesting idea, although coupled with the ClientCallImpl change, it will cause us to use the `executor` after shutdown _more_ than previously. Previously we would _never_ use the `executor` after shutdown; only the deadlineCancellationExecutor.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
dapengzhang0: I'm okay with removing this code path.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
dapengzhang0: The RPC would fail automatically?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: Seems previously we also would use the `executor` after shutdown: anything inside ClientCallImpl.start() may be happening after shutdown.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
dapengzhang0: what's the pre-existing Future implementation?

ejona86: At the very least there's `FutureTask` and `Futures.immediateCancelledFuture()`.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
ejona86: The RPC is not guaranteed to fail here. You're hoping that the rejection is related to channel termination. But it could have been that the queue was too long. Or it could be that the user shut down their executor before gRPC terminated.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -554,6 +555,32 @@ public String authority() {
...
+      final boolean[] accepted = new boolean[1];
+      try {
+        ManagedChannelImpl.this.executor.execute(new Runnable() {
+          @Override
+          public void run() {
+            accepted[0] = true;
+            command.run();
+          }
+        });
+      } catch (RejectedExecutionException e) {
```
ejona86: Instead of relying on the rejection, can we set `executor = null` when terminated and then check it here?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -535,7 +536,7 @@ public String authority() {
         CallOptions callOptions) {
       Executor executor = callOptions.getExecutor();
       if (executor == null) {
-        executor = ManagedChannelImpl.this.executor;
+        executor = new DirectFallbackExecutor();
```
ejona86: Let's not create this every time.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: In that case, we can check `terminated` directly, but still a race.
If we check `executor =! null`, and run with it, there could be a NPE because of the race.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
dapengzhang0: Now I got what you concerned.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
ejona86: NPE is not a concern:
```java
Executor saved = executor;
if (saved == null) {
  command.run();
} else {
  saved.execute(command);
}
```

I agree there is still a race between checking `executor` and running with it, but it is dramatically smaller and controlled by our code instead of the executor.

In my mind the race is _calling_ `executor` after the channel has terminated. None of these solutions actually fix that. They just mitigate it.

dapengzhang0: I agree that none of these solutions actually fix the race. But with the rejection check, the only unresolved race is very very benign as follows:
The `executor.execute()` is called and throws a `RejectedExecutionException` for whatever unlikely reason before the channel is terminated, and channel is terminated when we check it, so we don't throw this exception, and call `command.run()` directly.

Other cases are handled gracefully.

ejona86: I disagree. The rejection check continues to allow calls to `executor`. That's the race. `executor` may be provided by the application and it's not safe to call after termination. Or at least that's been my assumption.

I could believe the argument that we should always use the `executor`, even after termination, based on the theory that the application is expecting us to use `executor` (so it may be "magical") and it's the application's own fault for making an RPC after shutdown (not termination). But in that case we _mustn't_ use direct if there was a `RejectionExecutionException`. I'm not sure how many options we have, but we'd maybe need to propagate the exception to the caller and we never call their listener.

But this seems to be neither of these. I'm not sure what API theory this change is based on.

dapengzhang0:
obsolete

Issue #1914, opened 2016-06-09
Make CallCredentials non-experimental


zhangkun83:
This is a tracking issue for removing `@ExperimentalApi` from `CallCredentials`-related API.

dapengzhang0:
not stabilized yet, keep unscheduled

garrettjonesgoogle:
This seems critical for client stability - without this being non-experimental, there is no non-experimental way to make gRPC calls that require credentials, effectively making it so that gRPC doesn't have a "complete" GA surface.

ejona86:
Agreed, this is very important to stabilize. If push comes to shove, there is one trick we could do: we could stabilize the usages, but not the implementations. So that would mean CallOptions.withCallCredentials and MoreCallCredentials could be stable, even though the interface may change. We'd basically say that the name would remain the same and MoreCallCredentials would remain compatible. grpc-auth would need to be version-pinned (if it isn't already) to the same version of grpc-core, but that shouldn't cause much issue.

ejona86:
I'm assigning this to myself to remove most of the experimental api tags as I mentioned in the last comment, but I'm not going to end up stabilizing all of the API.

saturnism:
Would it be possible not to have Google credentials in the core gRPC libraries? gRPC as a core RPC framework probably don't need Google credentials.

ejona86:
@saturnism, only `grpc-auth` depends on `google-auth-library-credentials`. Also, "Google credentials" isn't in `google-auth-library-credentials`, but in `google-auth-library-oauth2-http`. `CallCredentials` is in `grpc-core`, so there should be no unnecessary dependencies.

saturnism:
it seems grpc-auth is only dealing w/ google credentials. @garrettjonesgoogle do you actually need this from gcloud-java? It was a discussion point of dependency conflicts.

ejona86:
@saturnism, your comments are off-topic for this issue (CallCredentials is in grpc-core and has no external dependencies). Let's move the conversation to the freshly-created #3280

ejona86:
Possibly need to fix: there is no way to cancel an async CallCredential. C has a [cancel in its API](https://github.com/grpc/grpc/blob/v1.8.0/src/core/lib/security/credentials/credentials.h#L174).

ejona86:
[...]

PR #3289, opened 2017-07-27
Stabilize auth flow using CallCredentials

ejona86:
As discussed in #1914, we need CallCredentials and MoreCallCredentials
to be stable, but there's less of a strong need for the contents of
CallCredentials to be stable. We're willing to commit to the name,
without needing to commit to the plumbing.

CC @garrettjonesgoogle

### DIFF
--- a/auth/src/main/java/io/grpc/auth/MoreCallCredentials.java
+++ b/auth/src/main/java/io/grpc/auth/MoreCallCredentials.java
@@ -27,8 +27,10 @@ public final class MoreCallCredentials {
   /**
    * Converts a Google Auth Library {@link Credentials} to {@link CallCredentials}.
    *
-   * <p>Although this is a stable API, note that the returned instance's API is not stable. See
-   * {@link CallCredentials}.
+   * <p>Although this is a stable API, note that the returned instance's API is not stable. You are
+   * free to use the class name {@code CallCredentials} and pass the instance to other code, but the
+   * instance can't be called directly from code expecting stable behavior. See {@link
+   * CallCredentials}.
    */
   public static CallCredentials from(Credentials creds) {
     return new GoogleAuthLibraryCallCredentials(creds);
--- a/auth/src/main/java/io/grpc/auth/MoreCallCredentials.java
+++ b/auth/src/main/java/io/grpc/auth/MoreCallCredentials.java
@@ -18,7 +18,6 @@ package io.grpc.auth;
 
 import com.google.auth.Credentials;
 import io.grpc.CallCredentials;
-import io.grpc.ExperimentalApi;
 
 /**
  * A utility class that converts other types of credentials to {@link CallCredentials}.
--- a/auth/src/main/java/io/grpc/auth/MoreCallCredentials.java
+++ b/auth/src/main/java/io/grpc/auth/MoreCallCredentials.java
@@ -23,10 +23,12 @@ import io.grpc.ExperimentalApi;
 /**
  * A utility class that converts other types of credentials to {@link CallCredentials}.
  */
-@ExperimentalApi("https//github.com/grpc/grpc-java/issues/1914")
 public final class MoreCallCredentials {
   /**
    * Converts a Google Auth Library {@link Credentials} to {@link CallCredentials}.
+   *
+   * <p>Although this is a stable API, note that the returned instance's API is not stable. See
+   * {@link CallCredentials}.
    */
   public static CallCredentials from(Credentials creds) {
     return new GoogleAuthLibraryCallCredentials(creds);
--- a/core/src/main/java/io/grpc/CallCredentials.java
+++ b/core/src/main/java/io/grpc/CallCredentials.java
@@ -28,14 +28,20 @@ import java.util.concurrent.Executor;
  * FooGrpc.FooStub stub = FooGrpc.newStub(channel);
  * response = stub.withCallCredentials(creds).bar(request);
  * </pre>
+ *
+ * <p>The contents and nature of this interface (and whether it remains an interface) is
+ * experimental, in that it can change. However, we are guaranteeing stability for the
+ * <em>name</em>. That is, we are guaranteeing stability for code to be returned a reference and
+ * pass that reference to gRPC for usage. However, code may not call or implement the {@code
+ * CallCredentials} itself if it wishes to only use stable APIs.
  */
-@ExperimentalApi("https//github.com/grpc/grpc-java/issues/1914")
 public interface CallCredentials {
   /**
    * The security level of the transport. It is guaranteed to be present in the {@code attrs} passed
    * to {@link #applyRequestMetadata}. It is by default {@link SecurityLevel#NONE} but can be
    * overridden by the transport.
    */
+  @ExperimentalApi("https//github.com/grpc/grpc-java/issues/1914")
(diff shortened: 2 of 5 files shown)


auth/src/main/java/io/grpc/auth/MoreCallCredentials.java
```diff
@@ -23,10 +23,12 @@
 /**
  * A utility class that converts other types of credentials to {@link CallCredentials}.
  */
-@ExperimentalApi("https//github.com/grpc/grpc-java/issues/1914")
 public final class MoreCallCredentials {
   /**
    * Converts a Google Auth Library {@link Credentials} to {@link CallCredentials}.
+   *
+   * <p>Although this is a stable API, note that the returned instance's API is not stable. See
+   * {@link CallCredentials}.
```
garrettjonesgoogle: Technically the type `CallCredentials` itself is now stable, but just none of the methods on it are.

core/src/main/java/io/grpc/CallCredentials.java
```diff
@@ -28,14 +28,20 @@
...
  */
-@ExperimentalApi("https//github.com/grpc/grpc-java/issues/1914")
 public interface CallCredentials {
   /**
    * The security level of the transport. It is guaranteed to be present in the {@code attrs} passed
    * to {@link #applyRequestMetadata}. It is by default {@link SecurityLevel#NONE} but can be
    * overridden by the transport.
    */
+  @ExperimentalApi("https//github.com/grpc/grpc-java/issues/1914")
   public static final Key<SecurityLevel> ATTR_SECURITY_LEVEL =
```
carl-mastrangelo: I think this should be moved off.   It's weird to inherit `ATTR_SECURITY_LEVEL` if you implement CallCredentials.

core/src/main/java/io/grpc/CallCredentials.java
```diff
@@ -65,6 +72,7 @@
    * @param applier The outlet of the produced headers. It can be called either before or after this
    *        method returns.
    */
+  @ExperimentalApi("https//github.com/grpc/grpc-java/issues/1914")
   void applyRequestMetadata(
```
carl-mastrangelo: This is not a valid.  You can't make part of an interface experimental.   All implementers have to implement it, which will break if it is ever removed.

ejona86: Did you not read the text added for the interface?

core/src/main/java/io/grpc/CallCredentials.java
ejona86: Discussion of that sort should be part of the rest of the stabilization effort. I'm not trying to change the API here, just make a bare-minimum of it stable so that users can authenticate themselves safely.

auth/src/main/java/io/grpc/auth/MoreCallCredentials.java
ejona86: @garrettjonesgoogle, I've reworded the Javadoc. See if it strikes you as more clear now.

core/src/main/java/io/grpc/CallCredentials.java
carl-mastrangelo: No, and neither will anyone else.   Please do not make our API support load based on the reading comprehension of our users.

If you want to make this stable and keep experimental methods, please make a base class that extends this.  That is the correct place to put methods that may change.

auth/src/main/java/io/grpc/auth/MoreCallCredentials.java
garrettjonesgoogle: That helps, I think it is more clear now.

core/src/main/java/io/grpc/CallCredentials.java
ejona86: They don't have to read the documentation, but it does explain what is going on. How can someone implement this API without touching Experimental API? How can someone call this API without touching Experimental API? They can't. But I do agree it is strange, thus the documentation.

> base class that extends this

I don't understand that. I'd have understood "base class than this extends" and "make this a base class that is extended".

> If you want to make this stable and keep experimental methods, please make a base class that extends this. That is the correct place to put methods that may change.

I don't know what you are talking about? That's not a thing we do.

And it doesn't help, since applyRequestMetadata could be completely changed.

carl-mastrangelo: They can implement it, but it needs to be a base class.  Experimental API and interface are all or nothing.  

> I don't understand that. I'd have understood "base class than this extends" and "make this a base class that is extended".

Something like: 
```
interface CallCredentials {
  void doNotImplementThisOrYouWillBreak();
}
public abstract class BaseCallCredentials implements CallCredentials {
public final void doNotImplementThisOrYouWillBreak();
} 
```

That way CallCredentials can mostly live on as is, Experimental methods can be moved down to the base class, where they can change.  If methods change, callers of experimental methods will change, but implementors of CallCredentials will not.

As the code stands with this PR, someone extending CallCredentials, a stable class, will suffer breakage when experimental methods come and go.  That is not stable.

ejona86: > If methods change, ... implementors of CallCredentials will not [change].

I don't want to provide that guarantee. And I don't see how that is true, unless you are saying there would be no implementors so it is vacuously true.

> As the code stands with this PR, someone extending CallCredentials, a stable class, will suffer breakage when experimental methods come and go. That is not stable.

It is impossible to extend CallCredentials without touching ExperimentalApi, so I feel like your sentence is a word-game that is skewing reality. Extending CallCredentials is not stable, because it is impossible with the current ExperimentalApi annotations and even explicitly mentioned in the documentation.

I'd be willing to make CallCredentials an abstract class or add a `doNotImplementThisOrYouWillBreak()`, but in either case I want to reserve the right to _completely_ change the contents, like adding a new abstract method and removing the existing method.

I could maybe be okay with leaving ExperimentalApi on the interface, but say that we guarantee its name as stable.

carl-mastrangelo: > but in either case I want to reserve the right to completely change the contents

Okay, but that isn't stable.  Leave the `@ExperimentalAPI` up top if thats what you want.  

> I don't want to provide that guarantee.  
If you remove the annotation, you provide that guarantee.  

>It is impossible to extend CallCredentials without touching ExperimentalApi, so I feel like your sentence is a word-game that is skewing reality.

This isn't a word game;  you just didn't understand.  Extendees of CallCredentials, after it has been unmarked ExperimentalApi, will expect their classes to not break.  Leaving ExperimentalApi methods on it implies they will break, because said methods may change or be removed.  That's contradicting signal, and __certain__ to confuse some users.  That causes __us__ future pain.  Do not cause us future pain.

ejona86: Talked offline. It seems we're both okay with adding a useless method that implementors have to implement to make things clear. Adding a `iAmUsingUnstableApi()` method.

garrettjonesgoogle: lol

carl-mastrangelo:
@ejona86 Please use the "package: thing that changed" style commit message when you squash merge.

zhangkun83:
LGTM

ejona86:
retest this please

PR #3292, opened 2017-07-28
context: change storage API to return a restoreable context

zpencer:
This API change allows storage implementations to put some state information into the return value, giving it the ability to act as a cookie. In environments where contexts can be replaced, the current `original context` can be stashed there and be restored when `detach` is called.

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -881,8 +881,10 @@ public class Context {
      * Implements {@link io.grpc.Context#attach}.
      *
      * @param toAttach the context to be attached
+     * @return A {@link Context} that should be passed back into {@link #detach(Context, Context)}
+     *   as the {@code toRestore} parameter.
      */
-    public abstract void attach(Context toAttach);
+    public abstract Context attach(Context toAttach);
 
     /**
      * Implements {@link io.grpc.Context#detach}
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -16,6 +16,7 @@
 
 package io.grpc;
 
+import com.sun.istack.internal.Nullable;
 import java.util.ArrayList;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
@@ -890,10 +891,16 @@ public class Context {
     /**
      * Implements {@link io.grpc.Context#attach}.
      *
+     * <p>Caution: {@link Context#attach()} interprets a return value of {@code null} to mean
+     * the same thing as {@link Context#ROOT}.
+     * <p>See also: {@link #current()}.
+
      * @param toAttach the context to be attached
      * @return A {@link Context} that should be passed back into {@link #detach(Context, Context)}
-     *        as the {@code toRestore} parameter.
+     *        as the {@code toRestore} parameter. {@code null} is a valid return value, but see
+     *        caution note.
      */
+    @Nullable
     public Context doAttach(Context toAttach) {
       // This is a default implementation to help migrate existing Storage implementations that
       // have an attach() method but no doAttach() method.
@@ -913,8 +920,16 @@ public class Context {
     public abstract void detach(Context toDetach, Context toRestore);
 
     /**
-     * Implements {@link io.grpc.Context#current}.  Returns the context of the current scope.
+     * Implements {@link io.grpc.Context#current}.
+     *
+     * <p>Caution: {@link Context} interprets a return value of {@code null} to mean the same
+     * thing as {@code Context{@link #ROOT}}.
+     * <p>See also {@link #doAttach(Context)}.
+     *
+     * @return The context of the current scope. {@code null} is a valid return value, but see
+     *        caution note.
      */
+    @Nullable
     public abstract Context current();
   }
 
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -385,9 +385,7 @@ public class Context {
    * }}</pre>
    */
   public Context attach() {
-    Context previous = current();
-    storage().attach(this);
-    return previous;
+    return storage().doAttach(this);
   }
 
   /**
@@ -878,13 +876,21 @@ public class Context {
    */
   public abstract static class Storage {
     /**
+     * @deprecated This is an old API that is no longer used.
+     */
+    @Deprecated
+    public void attach(Context toAttach) {
+      throw new UnsupportedOperationException("Deprecated. Do not call.");
+    }
+
+    /**
      * Implements {@link io.grpc.Context#attach}.
      *
      * @param toAttach the context to be attached
      * @return A {@link Context} that should be passed back into {@link #detach(Context, Context)}
      *        as the {@code toRestore} parameter.
      */
-    public abstract Context attach(Context toAttach);
+    public abstract Context doAttach(Context toAttach);
 
     /**
      * Implements {@link io.grpc.Context#detach}
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -882,7 +882,7 @@ public class Context {
      *
      * @param toAttach the context to be attached
      * @return A {@link Context} that should be passed back into {@link #detach(Context, Context)}
-     *   as the {@code toRestore} parameter.
+     *        as the {@code toRestore} parameter.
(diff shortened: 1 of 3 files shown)


context/src/main/java/io/grpc/Context.java
```diff
@@ -881,8 +881,10 @@ public String toString() {
      * Implements {@link io.grpc.Context#attach}.
      *
      * @param toAttach the context to be attached
+     * @return A {@link Context} that should be passed back into {@link #detach(Context, Context)}
+     *        as the {@code toRestore} parameter.
      */
-    public abstract void attach(Context toAttach);
+    public abstract Context attach(Context toAttach);
```
zhangkun83: IIRC this is a binary-incompatible change (you may want to verify it). Although the storage API is experimental thus allows us to break, our practice has been that we will provide smooth transition as long as it doesn't take much effort.

If this is binary-incompatible, a trick I usually use is:
```java
@Deprecated
public void attach(Context toAttach) {
  throw new UnsupportedOperationException("Deprecated. Do not call.");
}

public Context doAttach(Context toAttach) {
  attach(toAttach);
  return Context.current();
}
```

After one release, delete `attach()` and make `doAttach()` abstract.

zhangkun83: `Context.attach()` needs to use the returned Context.

adriancole:
Curious what the driving use case is for this. For example, it is neat to stain a context, more interesting why you might need to?

A breaking change, but to understand this better, if the api was like this, it would make this more obvious?:

Object attach(Context toAttach)
void detatch(Object token, Context toRestore)

Ex the main point is you are passing back the token you got (otherwise things could be bad)

Aside, but the context storage api doesnt really say what happens when you have an incorrect pairing.

ejona86:
> Curious what the driving use case is for this. For example, it is neat to stain a context, more interesting why you might need to?

We're using it to migrate a codebase from another context. The two ways we're using it: 1) to support replace() in the old API without losing the verification in detach() and 2) to save the old API's instance so we can return it later.  (1) is rarely used and was basically legacy before we started, but is hard to migrate. It's questionable whether we need (2), but since we're doing the trick already it's one less risk.

> A breaking change, but to understand this better, if the api was like this, it would make this more obvious?:
>
> Object attach(Context toAttach)
> void detach(Object token, Context toRestore)

Yep (although `void detach(Object token)`). Can't be done now though. That could have helped during earlier migration too. I'm not sure we would have done that, even if we knew then what we know now. But we would have considered it; we didn't at the time.

> Aside, but the context storage api doesnt really say what happens when you have an incorrect pairing.

If you have an incorrect pairing then the `current()` after the call isn't what is expected. [detach()](https://grpc.io/grpc-java/javadoc/io/grpc/Context.html#detach-io.grpc.Context-) talks about the SEVERE log that happens when we detect such a breakage, but it is only after-the-fact; the log location isn't likely the source of the bug.

So the concern is more this case, that works without issue before the hack:
```java
Context old = Context.current();
new.attach(); // throws away return :(
try {
  ...;
} finally {
  new.detach(old);
}
```

From an API perspective, it isn't clear if that is a bug, since `attach()` says:
> The previously current context is returned.

While `detach()` says:
> The provided replacement should be what was returned by the same attach() call.

We'll call it a bug internally; it seems to have no redeeming value and is easy to fix.

adriancole:
Thank you for the detailed explanation, Eric. This helps me understand and
I appreciate your taking the time.

context/src/main/java/io/grpc/Context.java
zpencer: Performed a simple test and verified that changing a void return type to a non avoid return type is not binary compatible.

context/src/main/java/io/grpc/ThreadLocalContextStorage.java
```diff
@@ -28,11 +28,18 @@
   /**
    * Currently bound context.
    */
-  private static final ThreadLocal<Context> localContext = new ThreadLocal<Context>();
+  private static final ThreadLocal<Context> localContext = new ThreadLocal<Context>() {
+    @Override
+    protected Context initialValue() {
+      return Context.ROOT;
```
ejona86: This would likely bite any implementation, because current and attach are working differently. I see two options:

1. We remove the `if (current == null)` check in `current()` and require the storage to handle that case. We would still want a `current == null` check, but it would throw an exception instead of choosing ROOT.

2. Have `Context.attach()` check the return and use `ROOT` if it is `null`. This is functionally equivalent to what was being done before.

context/src/main/java/io/grpc/Context.java
```diff
@@ -878,11 +876,21 @@ public String toString() {
...
+
+    /**
      * Implements {@link io.grpc.Context#attach}.
      *
      * @param toAttach the context to be attached
+     * @return A {@link Context} that should be passed back into {@link #detach(Context, Context)}
+     *        as the {@code toRestore} parameter.
      */
-    public abstract void attach(Context toAttach);
+    public abstract Context doAttach(Context toAttach);
```
ejona86: Adding a new method is not binary compatible. You can provide a default implementation that uses attach().

Note that because of the current definition of current(), you'd have to convert `null` to `ROOT` within doAttach().

zpencer: done

context/src/main/java/io/grpc/ThreadLocalContextStorage.java
zpencer: Let's go with option 2, since today implementations are allowed to return null and have `Context` handle it.

context/src/test/java/io/grpc/ContextTest.java
```diff
@@ -763,6 +765,22 @@ public void initContextWithCustomClassLoaderWithCustomLogger() throws Exception
     ((Runnable) runnable.getDeclaredConstructor().newInstance()).run();
   }
 
+  /**
+   * The context storage has per thread data structures.
+   * Ensure that newly created threads can attach/detach.
+   */
+  @Test
+  public void attachDetachNewThread() throws ExecutionException, InterruptedException {
+    Future<?> future = scheduler.submit(Context.current().wrap(new Runnable() {
```
zpencer: Good suggestion on the current value, I will update the tests. As for per thread data structures, I was previously just referring to initializing the ThreadLocal to ROOT, which is not necessary anymore with ejona's suggestions.

context/src/main/java/io/grpc/Context.java
```diff
@@ -878,11 +880,27 @@ public String toString() {
...
+
+    /**
      * Implements {@link io.grpc.Context#attach}.
      *
      * @param toAttach the context to be attached
+     * @return A {@link Context} that should be passed back into {@link #detach(Context, Context)}
+     *        as the {@code toRestore} parameter.
      */
-    public abstract void attach(Context toAttach);
+    public Context doAttach(Context toAttach) {
```
zhangkun83: We should document that the return value can be `null`, and what `Context.attach()` will use in that case. I should have had the same comment on `current()` too.

context/src/main/java/io/grpc/Context.java
zpencer: fixed

context/src/main/java/io/grpc/Context.java
```diff
@@ -16,6 +16,7 @@
 
 package io.grpc;
 
+import com.sun.istack.internal.Nullable;
```
adriancole: is this the nullable you want? :)

context/src/main/java/io/grpc/Context.java
```diff
@@ -895,8 +920,16 @@ public String toString() {
     public abstract void detach(Context toDetach, Context toRestore);
 
     /**
-     * Implements {@link io.grpc.Context#current}.  Returns the context of the current scope.
+     * Implements {@link io.grpc.Context#current}.
+     *
+     * <p>Caution: {@link Context} interprets a return value of {@code null} to mean the same
```
adriancole: maybe make a test that via reflection changes Storage to something that returns null? then test that ROOT is substituted

context/src/main/java/io/grpc/Context.java
zpencer: Ah I should be more careful about trusting my IDE

zpencer:
@ejona86 any more comments on this one?

zpencer:
restest this please

zpencer:
retest this please

PR #3294, opened 2017-07-31
core: enable Census tags propagation by default.

zhangkun83:
According to Census team, the tag wire format is now stable.

/cc @dinooliva @songy23


ejona86:
I assume that since Census impl isn't a dependency, "by default" the tags won't be propagated. It's only "by default" when you are already using Census.

zhangkun83:
> I assume that since Census impl isn't a dependency, "by default" the tags won't be propagated. It's only "by default" when you are already using Census.

That is correct.

PR #3295, opened 2017-07-31
Don't schedule multiple pings.

### LINKED ISSUE #3274: Pings being sent too often by KeepAliveManager

1.5.0

1.8

### What did you do?

* Enabled keepAliveTime of 20 seconds on the client, permitKeepAlive of 10 seconds on the server
* Start the server w/logging in KeepAliveEnforcer
* Start the client w/logging in KeepAliveManager
* After ~3 pings, KeepAliveManager sends a ping after < 10 seconds, which is marked invalid by KeepAliveEnforcer
* This happens a few more times and then more calm is sent

* Pings sent only every 20 seconds

You can checkout github.com/stephenh/mirror, the latest master has the reproducible error.

Here is the output of some logging that I added to KeepAliveManager:

```

12:37:54 STATE GOING TO PING_SCHEDULED from onTransportActive <-- 1st schedule
12:37:54 SCHEDULING sendPing in onTransportActive, delay=19545661745 (19.5 seconds)
12:38:06 STATE GOING TO PING_DELAYED from onDataReceived
12:38:07 STATE GOING TO IDLE from onTransportIdle

12:38:08 STATE GOING TO PING_SCHEDULED from onTransportActive <-- 2nd schedule
12:38:08 SCHEDULING sendPing in onTransportActive, delay=19630595429 (19.6 seconds)

12:38:13 RUNNING sendPing, state=PING_SCHEDULED <-- from 1st schedule (~20 seconds later)
12:38:13 STATE GOING TO PING_SENT from sendPing <-- 1st ping at :13
12:38:13 STATE GOING TO PING_SCHEDULED from onDataReceived
12:38:13 SCHEDULING sendPing in onDataReceived, delay=20000000000 <-- 3rd schedule

12:38:27 RUNNING sendPing, state=PING_SCHEDULED <-- from 2nd schedule (~20 seconds later)
12:38:27 STATE GOING TO PING_SENT from sendPing <-- 2nd ping at :27

12:38:28 STATE GOING TO PING_SCHEDULED from onDataReceived
12:38:28 SCHEDULING sendPing in onDataReceived, delay=20000000000

12:38:33 RUNNING sendPing, state=PING_SCHEDULED <-- from 3rd schedule
12:38:33 STATE GOING TO PING_SENT from sendPing <-- 3rd ping at :33, < 10 seconds later, marked invalid
12:38:33 STATE GOING TO PING_SCHEDULED from onDataReceived
12:38:33 SCHEDULING sendPing in onDataReceived, delay=2000000000
```

You can see the 3rd ping is sent only 6 seconds after the 2nd ping.

My guess is that KeepAliveManager is not handling sendPing being scheduled multiple times, and so should check nextKeepaliveTime instead of just the state.

stephenh:
If onTransportActive ran while SendPing was already scheduled, we would
schedule another SendPing, which seems fine, but the server might observe
us sending pings too quickly, and make us GOAWAY.

Fixes #3274.

### DIFF
--- a/core/src/main/java/io/grpc/internal/KeepAliveManager.java
+++ b/core/src/main/java/io/grpc/internal/KeepAliveManager.java
@@ -17,6 +17,7 @@
 package io.grpc.internal;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.MoreExecutors;
@@ -61,6 +62,7 @@ public class KeepAliveManager {
   private final Runnable sendPing = new LogExceptionRunnable(new Runnable() {
     @Override
     public void run() {
+      pingFuture = null;
       boolean shouldSendPing = false;
       synchronized (KeepAliveManager.this) {
         if (state == State.PING_SCHEDULED) {
@@ -170,8 +172,8 @@ public class KeepAliveManager {
       }
       // schedule a new ping
       state = State.PING_SCHEDULED;
-      pingFuture =
-          scheduler.schedule(sendPing, keepAliveTimeInNanos, TimeUnit.NANOSECONDS);
+      checkState(pingFuture == null, "There should be no outstanding pingFuture");
+      pingFuture = scheduler.schedule(sendPing, keepAliveTimeInNanos, TimeUnit.NANOSECONDS);
     }
   }
 
@@ -183,10 +185,12 @@ public class KeepAliveManager {
       // When the transport goes active, we do not reset the nextKeepaliveTime. This allows us to
       // quickly check whether the connection is still working.
       state = State.PING_SCHEDULED;
-      pingFuture = scheduler.schedule(
-          sendPing,
-          nextKeepaliveTime - ticker.read(),
-          TimeUnit.NANOSECONDS);
+      if (pingFuture == null) {
+        pingFuture = scheduler.schedule(
+            sendPing,
+            nextKeepaliveTime - ticker.read(),
+            TimeUnit.NANOSECONDS);
+      }
     } else if (state == State.IDLE_AND_PING_SENT) {
       state = State.PING_SENT;
     } // Other states are possible when keepAliveDuringTransportIdle == true
@@ -218,6 +222,7 @@ public class KeepAliveManager {
       }
       if (pingFuture != null) {
         pingFuture.cancel(false);
+        pingFuture = null;
       }
     }
   }
--- a/core/src/test/java/io/grpc/internal/KeepAliveManagerTest.java
+++ b/core/src/test/java/io/grpc/internal/KeepAliveManagerTest.java
@@ -105,6 +105,10 @@ public final class KeepAliveManagerTest {
 
   @Test
   public void keepAlivePingDelayedByIncomingData() {
+    ScheduledFuture<?> future = mock(ScheduledFuture.class);
+    doReturn(future)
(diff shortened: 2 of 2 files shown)


core/src/main/java/io/grpc/internal/KeepAliveManager.java
```diff
@@ -170,8 +171,10 @@ public synchronized void onDataReceived() {
       }
       // schedule a new ping
       state = State.PING_SCHEDULED;
-      pingFuture =
-          scheduler.schedule(sendPing, keepAliveTimeInNanos, TimeUnit.NANOSECONDS);
+      if (pingFuture == null) {
```
dapengzhang0: You can move this condition check `if (pingFuture == null)` up by replacing the condition check `if (state == State.PING_SCHEDULED)` above with it.

dapengzhang0:
ok to test

core/src/main/java/io/grpc/internal/KeepAliveManager.java
stephenh: Hm, that seems reasonable, but AFAIU `onDataReceived` can happen while we are in `State.IDLE/IDLE_WITH_PING_SENT` and so with your suggestion we would move to `PING_SCHEDULED` when really we might want to stay in `IDLE`. (Or something like that.)

I did make your suggested change and the tests still pass though...

core/src/main/java/io/grpc/internal/KeepAliveManager.java
```diff
@@ -69,6 +69,7 @@ public void run() {
           // Schedule a shutdown. It fires if we don't receive the ping response within the timeout.
           shutdownFuture = scheduler.schedule(shutdown, keepAliveTimeoutInNanos,
               TimeUnit.NANOSECONDS);
+          pingFuture = null;
```
ejona86: It seems like this should be done in all cases (but doesn't matter with DELAYED). If the state is IDLE, the current code would prevent all further PINGS (because it would never be rescheduled). Tying `pingFuture == null` to when there is no callback scheduled prevents having to consider most of those cases. So move clearing the field to before the if?

core/src/main/java/io/grpc/internal/KeepAliveManager.java
ejona86: I agree that we should leave this as-is. Mixing `pingFuture` checks with state machine checks makes the code complicated and easily broken (as is evidenced by the change breaking the `IDLE` case).

dapengzhang0: @stephenh Your argument makes sense. We should stay `IDLE` when we are in `IDLE` while ping had been scheduled, and then `onDataReceived`. But the check `if (pingFuture == null)` still looks strange, may be `Preconditions.checkState(pingFuture == null, "There should be no outstanding pingFuture")` at this place.

ejona86: Adding a check to guarantee `pingFuture == null` before the scheduling SGTM.

stephenh: Ah sure, that's a good idea.

core/src/main/java/io/grpc/internal/KeepAliveManager.java
stephenh: Great catch. I updated the `transportGoesIdle` test to reproduce this "will never reschedule" bug, and then fixed it by moving the `pingFuture = null` to the top of the Runnable.

ejona86:
Waiting for @dapengzhang0's review before merging.

dapengzhang0:
@stephenh thanks a lot for the fix!

PR #3300, opened 2017-08-02
core: implement Channel State API for PickFirst

dapengzhang0:


core/src/main/java/io/grpc/PickFirstBalancerFactory.java
```diff
@@ -70,7 +72,7 @@ public void handleResolvedAddressGroups(
       EquivalentAddressGroup newEag = flattenEquivalentAddressGroup(servers);
       if (subchannel == null) {
         subchannel = helper.createSubchannel(newEag, Attributes.EMPTY);
-        helper.updatePicker(new Picker(PickResult.withSubchannel(subchannel)));
+        helper.updateBalancingState(CONNECTING, new Picker(PickResult.withSubchannel(subchannel)));
```
zhangkun83: The new subchannel won't try to connect until an RPC is started.
The initial state of a new subchannel is `IDLE`. This should pass `IDLE`.

dapengzhang0: The initial state of `ManagedChannelImpl` is `IDLE`. I think the method `handleResolvedAddressGroups()` wouldn't happen before `NameResolver.start()`, which wouldn't happen before `ManagedChannelImpl.exitIdle()`. So it should not be `IDLE` when this method is called.

core/src/main/java/io/grpc/PickFirstBalancerFactory.java
```diff
@@ -92,9 +93,13 @@ public void handleNameResolutionError(Status error) {
     @Override
     public void handleSubchannelState(Subchannel subchannel, ConnectivityStateInfo stateInfo) {
       ConnectivityState currentState = stateInfo.getState();
-      if (subchannel != this.subchannel || currentState == SHUTDOWN) {
+      if (subchannel != this.subchannel) {
         return;
       }
+      if (currentState == SHUTDOWN) {
+        // Either IDLE_TIMEOUT or channel is shutdown. Won't have any effect for the latter case.
+        helper.updateBalancingState(IDLE, new Picker(PickResult.withNoResult()));
```
zhangkun83: IDLE_TIMEOUT is completely a channel impl logic. Even if we wanted to expose it to the application, it should be channel impl that would do that, i.e., transit to `IDLE` when `IdleModeTimer` expires and ignore updates from balancer if it's not the current one (which channel impl should be probably doing in the first place).

The balancer should not assume the channel's state. It should only provide state from its own perspective.

dapengzhang0: Good point. Fixed

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -250,6 +250,9 @@ public void run() {
       loadBalancer.shutdown();
       loadBalancer = null;
```
zhangkun83: In `LbHelperImpl.updateBalancingState()`, the state and the picker should be ignored if `loadBalancer != lb`, so that the channel ignores any updates from a balancer that has been shutdown and discarded.

dapengzhang0: I don't like this code in `ManagedChannelImpl`:
```
LbHelperImpl helper = new LbHelperImpl(nameResolver);
helper.lb = loadBalancerFactory.newLoadBalancer(helper);
this.loadBalancer = helper.lb;
```

Because `ManagedChannelImpl` does not own `loadBalancerFactory`, it does not know the behavior of `loadBalancerFactory.newLoadBalancer()`, which may just always recycle the old loadBalancer and restart it so that `loadBalancer == lb` is always true.

zhangkun83: A factory that recycles the instances is violating the API. If the factory is shared among multiple channels, worse things will happen anyway. I don't think we should worry about it.

dapengzhang0: If the factory is shared among multiple channels, it's not worse as long as the factory only recycles the completely shutdown ones.

What about letting `ManagedChannelImpl` have a `helper` field instead of `loadBalancer` field. And in `LbHelperImpl`, checking `this == ManagedChannelImpl.this.helper` instead of checking `loadBalancer == lb`?

zhangkun83: That sounds fine.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -678,6 +677,9 @@ public void updateBalancingState(
       checkNotNull(newState, "newState");
       checkNotNull(newPicker, "newPicker");
 
+      if (this != lbHelper) {
```
zhangkun83: `lbHelper` must be accessed from channelExecutor.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -767,6 +769,9 @@ public void runSerialized(Runnable task) {
     @Deprecated
     @Override
     public void updatePicker(final SubchannelPicker picker) {
+      if (this != lbHelper) {
```
zhangkun83: ditto

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: Corrected. PTAL

dapengzhang0:
retest this please

PR #3301, opened 2017-08-02
Document how to use Conscrypt on Android

ejona86:


kraghu:
@ejona86  can you update the size of the Conscrypt it will add to the app in read me ?

ericgribkoff:
I don't think we can put a size in the readme, as new releases of the Conscrypt jar may change the size. There are tools available to figure out the APK size for a library, such as http://www.methodscount.com/?lib=org.conscrypt%3Aconscrypt-android%3A1.0.0.RC9

SECURITY.md
```diff
@@ -10,9 +10,36 @@ On Android, use the [Play Services Provider](#tls-on-android). For non-Android s
...
+Provider](https://www.appfoundry.be/blog/2014/11/18/Google-Play-Services-Dynamic-Security-Provider/)
+to ensure your application has an up-to-date OpenSSL library with the necessary
+ciper-suites and a reliable ALPN implementation. This requires [updating the
+security provider at
+runtime](https://developer.android.com/training/articles/security-gms-provider.html).
 
-You may need to [update the security provider](https://developer.android.com/training/articles/security-gms-provider.html) to enable ALPN support, especially for Android versions < 5.0. If the provider fails to update, ALPN may not work.
+Although ALPN mostly works on newer Android releases (especially since 5.0),
+there are bugs that impact users that are only fixed by upgrading the security
+provider. In addition, for many users it is updated more quickly after
```
ericgribkoff: The "it" here is a little bit confusing. Would it convey the same message to combine this and the previous sentence, into something like this:

> Although ALPN mostly works on newer Android releases (especially since 5.0),
there are bugs and discovered security vulnerabilities that are only fixed by 
upgrading the security provider. Thus, we recommend using the Play Service 
Dynamic Security Provider for all Android versions.

ericgribkoff:
Quick update: as pointed out on google/conscrypt#276, our `OkHttpProtocolNegotiator.AndroidNegotiator#pickTlsExtensionType` code is deciding based on the detected API level not to turn on ALPN for older devices, unless the GMSCore OpenSSL provider is found. Adding Conscrypt doesn't change the result of this check, so it's only enabling NPN. I'll send out a fix for this.

Additionally, on the devices I've tested, the instructions here require removing the existing provider (`Security.removeProvider("AndroidOpenSSL");`) before installing Conscrypt to take its place, although I'll have to look into this a bit more as I'm not confident that this is the most robust way to handle this.

ejona86:
> Additionally, on the devices I've tested, the instructions here require removing the existing provider (Security.removeProvider("AndroidOpenSSL");) before installing Conscrypt to take its place, although I'll have to look into this a bit more as I'm not confident that this is the most robust way to handle this.

We should discuss this some, as I'm having to figure out how I want to detect Conscrypt with OpenJDK. But as a basic workaround you can do `Security.addProvider(Conscrypt.newProvider("SomeDifferentName"))`.

kraghu:
@ericgribkoff @ejona86 am still waiting on this update. Please keep me posted asap. I also noticed when i run debugger for some weird reason grpc calls work on api 19 devices. 

If it helps I also notices api 19 devices available in the market are really lower api devices got upgraded at least in UI terms they behave exactly like 15 or 16 .

ejona86:
@ericgribkoff, I just thought of a hack that I think will work for @kraghu until we fix the gRPC bug. Just name Conscrypt "GmsCore_OpenSSL". That will avoid the Android version detection.

```
Security.installProviderAt(Conscrypt.newProvider("GmsCore_OpenSSL"), 1);
```

(`installProviderAt(blah, 1)` makes sure it is clear it is the most-preferred provider. While the `addProvider(blah)` would work today, we feel we will need to make some changes in the search order that could break it in the future.)

kraghu:
@ejona86 @ericgribkoff am going to check out this option today or tomorrow and let you know how it goes

ejona86:
Closed in favor of #3971.

PR #3305, opened 2017-08-03
testing,core: don't use mocks for stream tracers

carl-mastrangelo:
This is a big, but mostly mechanical change.  The newly added Test*StreamTracer classes are designed to be extended which is why they are non final and have protected fields.  There are a few notable things in this:

1.  verifyNoMoreInteractions is gone.   The API for StreamTracers doesn't make this guarantee.  I have recovered this behavior by failing duplicate calls.  This has resulted in a few bugs in the test code being fixed.

2.  StreamTracers cannot be mocked anymore.  Tracers need to be thread safe, which mocks simply are not.  This leads to a HUGE number of reports when trying to find real races in gRPC.

3.  If these classes are useful, we can promote them out of internal.  I just put them here out of convenience.

### DIFF
--- a/core/src/main/java/io/grpc/StreamTracer.java
+++ b/core/src/main/java/io/grpc/StreamTracer.java
@@ -16,6 +16,7 @@
 
 package io.grpc;
 
+import com.google.errorprone.annotations.DoNotMock;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
@@ -23,6 +24,7 @@ import javax.annotation.concurrent.ThreadSafe;
  */
 @ExperimentalApi("https://github.com/grpc/grpc-java/issues/2861")
 @ThreadSafe
+@DoNotMock
 public abstract class StreamTracer {
   /**
    * Stream is closed.  This will be called exactly once.
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -25,10 +25,7 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
-import static org.mockito.Matchers.any;
-import static org.mockito.Mockito.atLeast;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.timeout;
 import static org.mockito.Mockito.times;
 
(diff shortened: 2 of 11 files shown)


core/src/main/java/io/grpc/StreamTracer.java
```diff
@@ -16,13 +16,15 @@
...
 
+import com.google.errorprone.annotations.DoNotMock;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
  * Listens to events on a stream to collect metrics.
  */
 @ExperimentalApi("https://github.com/grpc/grpc-java/issues/2861")
 @ThreadSafe
+@DoNotMock
```
adriancole: What a cool annotation!

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -139,21 +138,32 @@
...
-    final ServerStreamTracer tracer;
+    final InteropServerStreamTracer tracer;
 
-    ServerStreamTracerInfo(String fullMethodName, ServerStreamTracer tracer) {
+    ServerStreamTracerInfo(String fullMethodName, InteropServerStreamTracer tracer) {
       this.fullMethodName = fullMethodName;
       this.tracer = tracer;
     }
+
+    private static final class InteropServerStreamTracer extends TestServerStreamTracer {
```
adriancole: Nit i would probably call this ContextCaptorStreamTracer

testing/src/main/java/io/grpc/internal/testing/TestServerStreamTracer.java
```diff
@@ -0,0 +1,126 @@
...
+
+  @Override
+  public void outboundMessage() {
+    delegate.outboundMessage();
+  }
+
+  @Override
+  public void serverCallStarted(ServerCall<?, ?> call) {
+    if (!serverCall.compareAndSet(null, call) && delegate.failDuplicateCallbacks.get()) {
+      throw new AssertionError("serverCallStarted called more than once");
```
adriancole: Nice though wll this assertion error raise on the correct thread? Maybe use a queue and assert during the test method?

core/src/main/java/io/grpc/StreamTracer.java
carl-mastrangelo: It requires errorprone checking to actually enforce anything, fyi.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
carl-mastrangelo: yeah, i didn't have a good idea for this name.  If you want to send a PR i'd approve it.

testing/src/main/java/io/grpc/internal/testing/TestServerStreamTracer.java
carl-mastrangelo: We try to propagate errors around, even if they happen on the wrong thread.  I would prefer to know the call site.  Also, it would mean every test would need to check the error queue at the end, which might be easy to forget.   Even putting in the TearDown could be missed.

PR #3306, opened 2017-08-03
interop-testing: Only enable tracing when metrics are expected

kpayson64:
We can't use Mockito on Appengine, so this is a workaround.


kpayson64:
Looks like it

PR #3307, opened 2017-08-03
netty: Update netty to 4.1.14

ejona86:
EmbeddedChannel now runs all pending tasks when the Channel is closed.
This caused the Http2ConnectionHandler to clear deframer references (on
channelInactive) on errors when it previously didn't. Now that the
errors were handled more fully, it exposed bugs in tests.

FYI @normanmaurer, @Scottmitch


dapengzhang0:
You may want to update repositories.bzl as well?

normanmaurer:
@ejona86 thanks for the mention... LGTM

ejona86:
@dapengzhang0, PTAL

Issue #3007, opened 2017-05-15
Netty caused "onStreamAllocated was not called, but it seems the stream is active"


ejona86:
Reported by https://groups.google.com/d/topic/grpc-io/gpapLCscalk/discussion

Seen in grpc 1.3.0

```
22:42:37.609 |- WARN  i.n.u.c.DefaultPromise - An exception was thrown by io.grpc.netty.NettyClientStream$Sink$2.operationComplete()
java.lang.IllegalStateException: onStreamAllocated was not called, but it seems the stream is active
        at com.google.common.base.Preconditions.checkState(Preconditions.java:174)
        at io.grpc.internal.AbstractStream2$TransportState.onSentBytes(AbstractStream2.java:287)
        at io.grpc.netty.NettyClientStream$Sink$2.operationComplete(NettyClientStream.java:172)
        at io.grpc.netty.NettyClientStream$Sink$2.operationComplete(NettyClientStream.java:166)
        at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:507)
        at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:481)
        at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:420)
        at io.netty.util.concurrent.DefaultPromise.setSuccess(DefaultPromise.java:95)
        at io.netty.channel.DefaultChannelPromise.setSuccess(DefaultChannelPromise.java:76)
        at io.netty.channel.DefaultChannelPromise.setSuccess(DefaultChannelPromise.java:71)
        at io.netty.handler.codec.http2.StreamBufferingEncoder$Frame.release(StreamBufferingEncoder.java:309)
        at io.netty.handler.codec.http2.StreamBufferingEncoder$DataFrame.release(StreamBufferingEncoder.java:357)
        at io.netty.handler.codec.http2.StreamBufferingEncoder$PendingStream.close(StreamBufferingEncoder.java:292)
        at io.netty.handler.codec.http2.StreamBufferingEncoder.writeRstStream(StreamBufferingEncoder.java:184)
        at io.grpc.netty.NettyClientHandler.cancelStream(NettyClientHandler.java:470)
        at io.grpc.netty.NettyClientHandler.write(NettyClientHandler.java:244)
        at io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(AbstractChannelHandlerContext.java:739)
[...]

PR #3314, opened 2017-08-05
netty: check for stream before freeing flow control bytes on client

ericgribkoff:
Intended to address https://github.com/grpc/grpc-java/issues/3007.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyClientStream.java
+++ b/netty/src/main/java/io/grpc/netty/NettyClientStream.java
@@ -152,6 +152,8 @@ class NettyClientStream extends AbstractClientStream {
             channel.newPromise().addListener(new ChannelFutureListener() {
               @Override
               public void operationComplete(ChannelFuture future) throws Exception {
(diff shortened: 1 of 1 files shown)


netty/src/main/java/io/grpc/netty/NettyClientStream.java
```diff
@@ -152,7 +152,7 @@ public void writeFrame(WritableBuffer frame, boolean endOfStream, boolean flush)
             channel.newPromise().addListener(new ChannelFutureListener() {
               @Override
               public void operationComplete(ChannelFuture future) throws Exception {
-                if (future.isSuccess()) {
+                if (future.isSuccess() && transportState().http2Stream() != null) {
```
ejona86: This needs a clear comment. Maybe:

> If http2Stream is null, the stream has been cancelled before it was started and Netty is now purging pending writes from the flow-controller.

ericgribkoff: Done

johannesh:
Thanks for the fix! :)

PR #3315, opened 2017-08-05
core,netty,okhttp: add grpc-payload-bin in GET requests

ericgribkoff:
Once https://github.com/grpc/grpc/pull/12092 is checked in, this updates the Java client to use query strings of the form `?grpc-payload-bin=<base64 encoded gRPC message>`. This is a revert of https://github.com/grpc/grpc-java/commit/e36d229 plus deletion of a `TODO` message. 

Edit: And adding the key to OKHttp as well.

### DIFF
--- a/netty/src/test/java/io/grpc/netty/NettyClientStreamTest.java
+++ b/netty/src/test/java/io/grpc/netty/NettyClientStreamTest.java
@@ -17,6 +17,7 @@
 package io.grpc.netty;
 
 import static com.google.common.truth.Truth.assertThat;
+import static io.grpc.internal.AbstractClientStream.GRPC_PAYLOAD_BIN_KEY;
 import static io.grpc.internal.GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
 import static io.grpc.netty.NettyTestUtil.messageFrame;
 import static io.grpc.netty.Utils.CONTENT_TYPE_GRPC;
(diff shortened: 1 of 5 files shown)


netty/src/test/java/io/grpc/netty/NettyClientStreamTest.java
```diff
@@ -407,8 +408,13 @@ public void getRequestSentThroughHeader() {
...
     assertThat(ImmutableListMultimap.copyOf(cmdCap.getValue().headers()))
-        .containsEntry(AsciiString.of(":path"), AsciiString.of(
-            "//testService/test?" + BaseEncoding.base64().encode(msg)));
+        .containsEntry(
+            AsciiString.of(":path"),
+            AsciiString.of(
+                "//testService/test?"
+                    + GRPC_PAYLOAD_BIN_KEY
+                    + "="
+                    + BaseEncoding.base64().encode(msg)));
```
ejona86: Add an assert that the base64-encoded `msg` does not include `=`. Ditto for OkHttp

ericgribkoff: Picking this back up after ~7 months to clear PR backlog...

This was a good catch: the base64 encoder uses `=` for padding. The call here should actually be to [`base64Url()`](https://google.github.io/guava/releases/16.0/api/docs/com/google/common/io/BaseEncoding.html#base64Url()), to use `-` and `_` instead of `+` and `/`. grpc/grpc has an equivalent to `base64Url`, but it also uses `=` for padding. AFAICT using the `=` reserved character will still "work" but violates [RFC 1738](https://www.ietf.org/rfc/rfc1738.txt). We should just URL encode the padding character here, but this requires decoding capability on the server side. Will update the grpc/grpc PR and then update this again (sometime this week, hopefully; in <7 months, worst case ;))

ejona86: Virtually all base64 versions use `=` for padding. The difference is whether they use padding at all. For Guava, we just need to call `omitPadding()`. I don't think we should UrlEncode the padding character, but that's a bit of a spec-level decision.

ejona86:
@ericgribkoff, why wasn't this submitted? It looks like it just needed one line of assertion in a test. Do you remember if there was anything else pending for it?

ericgribkoff:
@ejona86 The necessary change on Java's side was one-line, but the base64 encoding method used in core does not support omitting the padding or using anything other than `=` (I don't think I revisited this after your followup comment in July). As I recall the changes needed to be coordinated between the repos, since this is mainly for communicating with a C++ server, and I never got around to resolving this difference between core and Java.

PR #2945, opened 2017-04-25
okhttp: Support GET in okhttp transport

zsurocking:

Issue #1775, opened 2016-05-03
Tracking Issue for Idempotence being Experimental.


carl-mastrangelo:
Specific usages:
- `MethodDescriptor.isIdempotent`
- `MethodDescriptor.withIdempotent`

jhspaybar:
How do I get code gen to set this field to true?  We have a custom Retry interceptor that we'd like to check this field to determine when to trigger it.  It looks like there isn't a proto option, or at least I haven't been able to dig one up.

ejona86:
The protoc changes haven't landed yet. grpc/grpc#7945 with Proposed Method Annotations discusses it some, but I don't think I've seen the protoc changes land for any language.

noel-yap:
I'm thinking `withSafe(boolean safe)` should also set `idempotent` to the value of `safe`.

rmichela:
@ejona86 Are we are still waiting for protoc to emit cache-control compliant grpc stubs? Any idea on an ETA?

ejona86:
@makdharma, is there anything to share about the current state of POST support (plumbing, how hard we're pushing)?

makdharma:
The support for GET (POST was always supported :) has landed in Java and C
stacks. Nobody is working on the protoc change for plumbing the idempotent
option all the way through to code gen. For now you'll have to manually add
the flag.

ejona86:
> POST was always supported

Oops. I meant PUT.

> The support for GET has landed in Java and C stacks.

That needs to be clarified: GET on client-side. Java does not support GET nor PUT on server-side.

It should also be clarified that the current GET protocol has known future breaking changes planned.

rmichela:
The ultimate purpose of these changes is to implement native request cache support in gRPC. 

Is native request cache support a near-term, long-term, or "eventually" feature? Should we pursue other application-tier or `ClientInterceptor`-based caching strategies in the interim?

makdharma:
If you are using Java or C++ client and C++ server and you have a HTTP/2
caching proxy then it should work currently. Any other combination is a
"eventually" feature. The HTTP/2 proxy (that speaks H2 on both client and
backend) could be a problem, as the nginx changes are still in the process.

ejona86:
[...]

PR #3316, opened 2017-08-06
Finish support for GET methods in OkHttp transport

ericgribkoff:
It turns out that https://github.com/grpc/grpc-java/pull/2945 did not fully implement GET support for the OkHttp transport. This PR modifies the outgoing headers for a GET request to specify `method: GET` and sets `endStream=true`, and also tells the frame writer to immediately flush the headers instead of waiting for additional data on the stream (which will never come).

Edit: This also adds a test for `method: GET` to the corresponding Netty test case. Netty was already doing things correctly, but did not have test coverage for this.

### DIFF
--- a/okhttp/src/main/java/io/grpc/okhttp/Headers.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/Headers.java
@@ -36,6 +36,7 @@ class Headers {
 
   public static final Header SCHEME_HEADER = new Header(Header.TARGET_SCHEME, "https");
   public static final Header METHOD_HEADER = new Header(Header.TARGET_METHOD, GrpcUtil.HTTP_METHOD);
+  public static final Header METHOD_GET_HEADER = new Header(Header.TARGET_METHOD, "GET");
   public static final Header CONTENT_TYPE_HEADER =
       new Header(CONTENT_TYPE_KEY.name(), GrpcUtil.CONTENT_TYPE_GRPC);
   public static final Header TE_HEADER = new Header("te", GrpcUtil.TE_TRAILERS);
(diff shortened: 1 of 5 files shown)


okhttp/src/main/java/io/grpc/okhttp/Headers.java
```diff
@@ -45,8 +46,8 @@
    * creating a stream. Since this serializes the headers, this method should be called in the
    * application thread context.
    */
-  public static List<Header> createRequestHeaders(Metadata headers, String defaultPath,
-      String authority, String userAgent) {
+  public static List<Header> createRequestHeaders(
```
dapengzhang0: This is a public API. I'm not sure it's appropriate to delete the old signature. cc @ejona86 . Even decided to delete it, should keep the old signature for at least one release cycle and mark it with deprecated. Better to mark the new one with `@ExperimentalApi`.

ericgribkoff: Thanks for catching this. Removal from the public API surface was inadvertent, and now fixed. I added the new method with @ExperimentalApi, but did not pick an existing issue - https://github.com/grpc/grpc-java/issues/1775 is related but not quite right, so should probably create a new issue for this.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -99,6 +101,10 @@ public int id() {
     return id;
   }
 
+  boolean useGet() {
```
dapengzhang0: `@GuardedBy("lock")`

okhttp/src/main/java/io/grpc/okhttp/Headers.java
ejona86: Whoa. This class needs to be made non-public or marked as `@Internal`. I'll send out a PR that makes it package-private. That's safe since Header is internal, so there's no way to use this API safely already. Eric, no need to maintain compatibility or add the annotations for now.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
ericgribkoff: Sorry for the delay in responding. This doesn't need synchronization here - the boolean can only change from false to true on the application thread, before the stream starts. Added Javadoc to clarify that the method's return value is only valid after the headers are written.

PR #3320, opened 2017-08-07
core: Providers should consider Robolectric as Android

ejona86:
null class loader means to use the bootstrap class loader, which would
normally make sense. However, Robolectric tests run on OpenJDK and
provide the Android environment as part of the application, so "Android"
won't be present in the bootstrap class loader.

### DIFF
--- a/core/src/main/java/io/grpc/ManagedChannelProvider.java
+++ b/core/src/main/java/io/grpc/ManagedChannelProvider.java
@@ -126,7 +126,8 @@ public abstract class ManagedChannelProvider {
   protected static boolean isAndroid() {
     try {
       // Specify a class loader instead of null because we may be running under Robolectric
-      Class.forName("android.app.Application", /*initialize=*/ false, ManagedChannelProvider.class);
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/ManagedChannelProvider.java
```diff
@@ -125,7 +125,8 @@ public static ManagedChannelProvider provider() {
    */
   protected static boolean isAndroid() {
     try {
-      Class.forName("android.app.Application", /*initialize=*/ false, null);
+      // Specify a class loader instead of null because we may be running under Robolectric
+      Class.forName("android.app.Application", /*initialize=*/ false, ManagedChannelProvider.class);
```
ericgribkoff: ManagedChannelProvider.class.getClassLoader()?

ejona86: Yeah, done. I manually copied a small part of an internal CL, and obviously I copied it incompletely.

PR #3321, opened 2017-08-08
core: Refactor Java code

Moriadry:
Remove unnecessary 'continue' statement and simplify conditional expression


dapengzhang0:
ok to test

Moriadry:
improve performance issues.

core/src/main/java/io/grpc/ServerProvider.java
```diff
@@ -39,7 +39,6 @@ static final ServerProvider load(ClassLoader cl) {
 
     for (ServerProvider current : providers) {
       if (!current.isAvailable()) {
-        continue;
```
ejona86: The continue is more clear to me, and it is weird to have the block with no contents. It's honestly fine either way, so I think we should just leave it the way that it is.

core/src/main/java/io/grpc/internal/AbstractServerStream.java
```diff
@@ -116,7 +116,7 @@ public final void writeHeaders(Metadata headers) {
   public final void deliverFrame(WritableBuffer frame, boolean endOfStream, boolean flush) {
     // Since endOfStream is triggered by the sending of trailers, avoid flush here and just flush
     // after the trailers.
-    abstractServerStreamSink().writeFrame(frame, endOfStream ? false : flush);
+    abstractServerStreamSink().writeFrame(frame, !endOfStream && flush);
```
ejona86: I find the previous version more clear that the `flush` is pass-through. I'm having trouble with this new version, although I do question why it's harder to comprehend.

core/src/main/java/io/grpc/ServerProvider.java
Moriadry: ok, your explanation makes sense, I will rever it.

PR #3324, opened 2017-08-08
ProtoLiteUtils fix infinity loop

### LINKED ISSUE #3323: Infinite cycle during parse client request
1.5.0 Release

java version "1.8.0"
Java(TM) SE Runtime Environment (build 1.8.0-b132)
Java HotSpot(TM) 64-Bit Server VM (build 25.0-b70, mixed mode)

### What the problem?
After minor change my proto request all of my request start failed with deadline.

Thread-dumps from server side show that root cause of this deadlines in parsing code
```
"stockpile-server-grpc-1-2" #14 prio=5 os_prio=0 tid=0x00007f455c170800 nid=0x90b runnable [0x00007f4589dbd000]
   java.lang.Thread.State: RUNNABLE
	at io.grpc.protobuf.lite.ProtoLiteUtils$2.parse(ProtoLiteUtils.java:133)
	at io.grpc.protobuf.lite.ProtoLiteUtils$2.parse(ProtoLiteUtils.java:79)
	at io.grpc.MethodDescriptor.parseRequest(MethodDescriptor.java:280)
	at io.grpc.internal.ServerCallImpl$ServerStreamListenerImpl.messageRead(ServerCallImpl.java:246)
	at io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1.runInContext(ServerImpl.java:571)
	at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
	at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:102)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:138)
	at java.lang.Thread.run(Thread.java:748)
```

This cycle never completed, because try read zero bytes from `InputStream` 
https://github.com/grpc/grpc-java/blob/master/protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoLiteUtils.java#L133

```java
              int chunkSize;
              int position = 0;
              while ((chunkSize = stream.read(buf, position, size - position)) != -1) {
                position += chunkSize;
              }
              if (size != position) {
                throw new RuntimeException("size inaccurate: " + size + " != " + position);
```

[InputSteam contract](https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#read(byte[],%20int,%20int)) say that read method will always return zero if reads zero bytes.
> If len is zero, then no bytes are read and 0 is returned; otherwise, there is an attempt to read at least one byte. If no byte is available because the stream is at end of file, the value -1 is returned; otherwise, at least one byte is read and stored into b.

Gordiychuk:
InputStream by contract can retun zero if requested length equal to zero.

```
If len is zero, then no bytes are read and 0 is returned;
otherwise, there is an attempt to read at least one byte.
If no byte is available because the stream is at end of file,
the value -1 is returned; otherwise, at least one byte is read
and stored into b.
```

Close #3323


protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoLiteUtils.java
```diff
@@ -132,6 +132,9 @@ public T parse(InputStream stream) {
               int position = 0;
               while ((chunkSize = stream.read(buf, position, size - position)) != -1) {
                 position += chunkSize;
+                if (chunkSize == 0 && position == size) {
```
ejona86: How about just have `if (position == size)` here and then outside of the loop:

```java
if (stream.read(buf, 0, 1) != -1) {
  throw ...;
}
```

ejona86:
okay to test

protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoLiteUtils.java
Gordiychuk: What about 
```java
  int remaining = size;
  while (remaining > 0) {
    int position = size - remaining;
    int count = stream.read(buf, position, remaining);
    if (count == -1) {
      break;
    }
    remaining -= count;
  }

  if (remaining != 0) {
    throw new RuntimeException("size inaccurate: " + size + " != " + (size - remaining));
  }
```

ejona86: That seems equally fine to me as to my suggestion of using `if (position == size)` (without the additional `read`); your choice.

The additional `read()` was just to notice when an `InputStream` lied. While it would be nice, I can live without it. (The additional call to `read()` was already happening in the working case before; so I wasn't too concerned about the "useless" read.)

Gordiychuk:
@ejona86 all review comments fixed

ejona86:
okay to test

PR #3325, opened 2017-08-08
okhttp,interop-testing: use forked OkHttp classes

ericgribkoff:
This reduces our dependency on the OkHttp 2.5.0 library by switching imports of three classes (`CipherSuite`, `ConnectionSpec`,  and `TlsVersion`) from `com.squareup.okhttp.*` to our already manually copied/forked classes in `io.grpc.okhttp.internal.*`.

For our Android Hello World app after Proguard, this change reduces the classes.dex size of our `com.squareup.okhttp.*` dependencies  from 12.3KB to 4.0KB, and a modest reduction in APK size of 4.3KB (from 710.8KB to 706.5KB).

The following classes are still direct dependencies of `OkHttpClientTransport`:

- com.squareup.okhttp.Credentials
- com.squareup.okhttp.HttpUrl
- com.squareup.okhttp.Request
- com.squareup.okhttp.internal.http.StatusLine

None of these are forked in `io.grpc.okhttp.internal`, although `StatusLine` brings in `com.squareup.okhttp.Protocol`, which is already forked as `io.grpc.okhttp.internal.Protocol`. We could save an additional 400B out of the 4KB of the `com.squareup.okhttp.*` dependencies by forking `StatusLine` (and forking its dependency `com.squareup.okhttp.Response`).

I manually downloaded the OkHttp 2.5.0 jar and verified that the implementations of `CipherSuite`, `ConnectionSpec`,  and `TlsVersion` match our own forked copies (we do have one method with increased visibility in our copy of `ConnectionSpec`).


okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -37,7 +34,10 @@
 import io.grpc.internal.KeepAliveManager;
 import io.grpc.internal.SharedResourceHolder;
 import io.grpc.internal.SharedResourceHolder.Resource;
+import io.grpc.okhttp.internal.CipherSuite;
+import io.grpc.okhttp.internal.ConnectionSpec;
```
ejona86: ConnectionSpec is part of grpc's API in this class. We don't want to expose our internal fork, so we let users specify okhttp's ConnectionSpec and then we use Utils.convertSpec() to copy it into an internal equivalent.

Feel free to eagerly convert in `connectionSpec()`; that would allow proguard to remove the references if it isn't specified.

ericgribkoff:
Superseded by https://github.com/grpc/grpc-java/pull/4267

PR #3329, opened 2017-08-09
grpclb: update load_balancer.proto to the latest version.

zhangkun83:
Combined drop for rate limiting and drop for load balancing, broken down by token.


grpclb/src/main/java/io/grpc/grpclb/GrpclbClientLoadRecorder.java
```diff
@@ -38,8 +40,9 @@
   private final AtomicLong callsFinished = new AtomicLong();
 
   // Specific finish types
-  private final AtomicLong callsDroppedForRateLimiting = new AtomicLong();
-  private final AtomicLong callsDroppedForLoadBalancing = new AtomicLong();
+  // Access to it should be protected by lock.  Contention is not an issue for these counts, because
+  // normally only a small portion of all RPCs are dropped.
+  private HashMap<String, AtomicLong> callsDroppedPerToken = new HashMap<String, AtomicLong>();
```
zhangkun83: I use AtomicLong for its mutability. If I used Long, each update would be two map operations, and one or more object allocation, which is presumably more expensive than incrementing AtomicLong.

zhangkun83: Done.

grpclb/src/main/java/io/grpc/grpclb/GrpclbClientLoadRecorder.java
```diff
@@ -56,34 +59,43 @@ public ClientStreamTracer newClientStreamTracer(CallOptions callOptions, Metadat
...
         .setTimestamp(Timestamps.fromMillis(time.currentTimeMillis()))
         .setNumCallsStarted(callsStarted.getAndSet(0))
         .setNumCallsFinished(callsFinished.getAndSet(0))
-        .setNumCallsFinishedWithDropForRateLimiting(callsDroppedForRateLimiting.getAndSet(0))
-        .setNumCallsFinishedWithDropForLoadBalancing(callsDroppedForLoadBalancing.getAndSet(0))
         .setNumCallsFinishedWithClientFailedToSend(callsFailedToSend.getAndSet(0))
-        .setNumCallsFinishedKnownReceived(callsFinishedKnownReceived.getAndSet(0))
-        .build();
+        .setNumCallsFinishedKnownReceived(callsFinishedKnownReceived.getAndSet(0));
+    synchronized (this) {
```
zhangkun83: Done.

grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancerTest.java
```diff
@@ -248,7 +247,7 @@ public void errorPicker() {
   public void roundRobinPicker() {
     GrpclbClientLoadRecorder loadRecorder = new GrpclbClientLoadRecorder(timeProvider);
     Subchannel subchannel = mock(Subchannel.class);
-    RoundRobinEntry r1 = new RoundRobinEntry(DropType.RATE_LIMITING, loadRecorder);
+    RoundRobinEntry r1 = new RoundRobinEntry(loadRecorder, "LBTOKEN0001");
```
dapengzhang0: In `RoundRobinEntry` class, it's better to define two static methods to create new instances, because the constructor does not have a meaningful name, and I don't realize it's a drop entry here unless I check its definition.

zhangkun83: Done.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -649,24 +649,28 @@ public PickResult pickSubchannel(PickSubchannelArgs args) {
     final String token;
     final boolean isDrop;
 
+    private RoundRobinEntry(
+        PickResult result, GrpclbClientLoadRecorder loadRecorder, String token, boolean isDrop) {
+      this.result = checkNotNull(result);
+      this.loadRecorder = checkNotNull(loadRecorder, "loadRecorder");
+      this.token = checkNotNull(token, "token");
+      this.isDrop = isDrop;
```
dapengzhang0: `this.isDrop = result == DROP_PICK_RESULT`; No need of the 4th argument.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -738,8 +722,8 @@ public PickResult pickSubchannel(PickSubchannelArgs args) {
           index = 0;
         }
         result.updateHeaders(args.getHeaders());
-        if (result.dropType != null) {
-          result.loadRecorder.recordDroppedRequest(result.dropType);
+        if (result.result == DROP_PICK_RESULT) {
```
dapengzhang0: `if (result.isDrop)`

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
zhangkun83: Good idea. I dropped the `isDrop` field and rely on `result == DROP_PICK_RESULT` for checking a drop entry.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java

PR #3330, opened 2017-08-09
okhttp: Make Headers package-private

ejona86:
There was no way to use it without touching internal APIs, and it should
never have been public to begin with as it serves no purpose to a user.

CC @ericgribkoff


ejona86:
retest this please

PR #3338, opened 2017-08-11
all: Fix mismatch in naming Bazel Maven jars

### LINKED ISSUE #3328: Mismatch in naming with bazel maven jars
I am trying to use grpc-java in an existing project which uses bazel, java, and a bunch of jars from maven. I added this to my WORKSPACE:

```
load("@grpc_java//:repositories.bzl", "grpc_java_repositories")

grpc_java_repositories(
  omit_com_google_guava=True,
)
```

The problem is that I use the [migration-tooling](https://github.com/bazelbuild/migration-tooling) project to add maven jars, and its naming for external jars does not match yours. For example, you call it `@com_google_guava`[1] but that project uses `@com_google_guava_guava`[2]. So if I try to use the omits, I get:

```
no such package '@com_google_guava//jar': The repository could not be resolved
```

And if I remove the omits, my project references two different versions of guava and I get runtime errors (since this project uses guava 19 and I want 21).

Is it possible to switch the naming of external maven jars to match bazel's tooling?

1.5.0

```
% java -version
java version "1.8.0_144"
Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)
```

[1] https://github.com/grpc/grpc-java/blob/master/repositories.bzl#L113
[2]
```
% bazel run //generate_workspace -- --output_dir /tmp --artifact=com.google.guava:guava:21.0
...
Wrote /tmp/generate_workspace.bzl

% cat /tmp/generate_workspace.bzl 
# The following dependencies were calculated from:
#
# generate_workspace --output_dir /tmp --artifact=com.google.guava:guava:21.0


def generated_maven_jars():
  native.maven_jar(
      name = "com_google_guava_guava",
      artifact = "com.google.guava:guava:21.0",
  )




def generated_java_libraries():
  native.java_library(
      name = "com_google_guava_guava",
      visibility = ["//visibility:public"],
      exports = ["@com_google_guava_guava//jar"],
  )
```

pgr0ss:
This commit aligns the naming of the Bazel Maven jars with the names
used by Bazel's migration-tooling project:

https://github.com/bazelbuild/migration-tooling

Unfortunately, we can't fix @com_google_protobuf_java because it's
required by Bazel itself.

Fixes #3328


ejona86:
Okay to test

ejona86:
The naming convention is actually defined clearly in [the documentation for maven_jar](https://docs.bazel.build/versions/master/be/workspace.html#maven_jar).

PR #3339, opened 2017-08-11
netty: fix keepalive test flakiness

### LINKED ISSUE #3326: NettyServerHandlerTest keepalive tests are flaky since netty 4.1.14
https://travis-ci.org/grpc/grpc-java/jobs/262360341

```
io.grpc.netty.NettyServerHandlerTest > maxConnectionAgeGrace_channelClosedAfterGracePeriod FAILED
    java.nio.channels.ClosedChannelException
        at io.netty.channel.embedded.EmbeddedChannel.checkOpen(EmbeddedChannel.java:644)
        at io.netty.channel.embedded.EmbeddedChannel.ensureOpen(EmbeddedChannel.java:656)
        at io.netty.channel.embedded.EmbeddedChannel.writeInbound(EmbeddedChannel.java:310)
        at io.grpc.netty.NettyHandlerTestBase.channelRead(NettyHandlerTestBase.java:140)
        at io.grpc.netty.NettyServerHandlerTest.createStream(NettyServerHandlerTest.java:735)
        at io.grpc.netty.NettyServerHandlerTest.maxConnectionAgeGrace_channelClosedAfterGracePeriod(NettyServerHandlerTest.java:709)
260 tests completed, 1 failed, 3 skipped
:grpc-netty:test FAILED
```

dapengzhang0:
In `NettyHandlerTestBase` class, extended Netty's `EmbeddedChannel` by overriding`eventLoop()` to return an `eventLoop` that uses `FakeClock.getScheduledExecutorService() to schedule tasks.

Resolves #3326

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
@@ -78,7 +78,6 @@ import io.netty.handler.logging.LogLevel;
 import io.netty.util.AsciiString;
 import io.netty.util.ReferenceCountUtil;
 import java.util.List;
-import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
@@ -290,8 +289,7 @@ class NettyServerHandler extends AbstractNettyHandler {
 
     // init max connection age monitor
     if (maxConnectionAgeInNanos != MAX_CONNECTION_AGE_NANOS_DISABLED) {
-      ScheduledExecutorService scheduledExecutorService = ctx.executor();
-      maxConnectionAgeMonitor = scheduledExecutorService.schedule(
+      maxConnectionAgeMonitor = ctx.executor().schedule(
           new LogExceptionRunnable(new Runnable() {
             @Override
             public void run() {
@@ -315,11 +313,6 @@ class NettyServerHandler extends AbstractNettyHandler {
               } finally {
                 gracefulShutdownTimeoutMillis(savedGracefulShutdownTime);
               }
-            }
-
-            @Override
-            public String toString() {
-              return NettyServerHandler.class.getName() + ".maxConnectionAgeMonitor";
             }
           }),
           maxConnectionAgeInNanos,
--- a/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
@@ -78,6 +78,7 @@ import io.netty.handler.logging.LogLevel;
 import io.netty.util.AsciiString;
 import io.netty.util.ReferenceCountUtil;
 import java.util.List;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
@@ -289,7 +290,8 @@ class NettyServerHandler extends AbstractNettyHandler {
 
     // init max connection age monitor
     if (maxConnectionAgeInNanos != MAX_CONNECTION_AGE_NANOS_DISABLED) {
-      maxConnectionAgeMonitor = ctx.executor().schedule(
+      ScheduledExecutorService scheduledExecutorService = ctx.executor();
+      maxConnectionAgeMonitor = scheduledExecutorService.schedule(
           new LogExceptionRunnable(new Runnable() {
             @Override
             public void run() {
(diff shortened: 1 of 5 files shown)


netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -289,7 +290,8 @@ public void handlerAdded(final ChannelHandlerContext ctx) throws Exception {
 
     // init max connection age monitor
     if (maxConnectionAgeInNanos != MAX_CONNECTION_AGE_NANOS_DISABLED) {
-      maxConnectionAgeMonitor = ctx.executor().schedule(
```
ejona86: Why this modification? Is this avoiding an issue with your formatter?

netty/src/test/java/io/grpc/netty/NettyHandlerTestBase.java
```diff
@@ -84,6 +92,17 @@
...
   protected void manualSetUp() throws Exception {}
 
+  private final FakeClock fakeClock = new FakeClock();
+  private String[] scheduleTheseWithFakeClock = new String[0];
+
+  FakeClock fakeClock() {
+    return fakeClock;
+  }
+
+  void useFakeClockWhenSchdulingThese(String... taskNames) {
```
ejona86: Eww... I'd really hope for a different solution.

netty/src/test/java/io/grpc/netty/NettyHandlerTestBase.java
```diff
@@ -94,12 +113,50 @@ protected final void initChannel(Http2HeadersDecoder headersDecoder) throws Exce
...
+          for (String taskName : scheduleTheseWithFakeClock) {
+            if (name.contains(taskName)) {
+              return true;
+            }
+          }
+          return false;
+        }
+      };
+      doAnswer(delegatesTo(fakeClock.getScheduledExecutorService())).when(eventLoop)
+          .schedule(argThat(isFakeClockRunnable), anyLong(), any(TimeUnit.class));
```
ejona86: Why don't we schedule everything on the fake clock?

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
dapengzhang0: Without this change, the `.schedule()` method is compiled as `EventLoopGroup.schedule()`. 
With this change,  the `.schedule()` method is compiled as  `ScheduleExecutorService.schedule()`.


In the runtime, if `ctx.executor()`'s value is a mockito-stubbed eventLoop, `eventLoop.schedule()` will again invokes the compiled method reflectively: `method.invoke(caller, arguments)`, where the `caller` can be the `fakeClock`. Then without this change:  it will throw `IllegalArgumentException` because the first argument `fakeClock`, is not a `EventLoopGroup` instance. With this change: `method.invoke(caller, argument)` won't fail in the runtime. 

If `ctx.executor()` is not a mockito-stubbed eventLoop, then the runtime behaviors are the same for the previous code and the changed code, because in both case it will call the `.schedule()` method of the caller's runtime type.

ejona86: So you're saying that `ctx.executor()` does not return the correct type when mocking. Let's fix that instead of working around it here.

netty/src/test/java/io/grpc/netty/NettyHandlerTestBase.java
dapengzhang0: I'm not sure if it's safe to do so, because I don't how much Netty code calls on the real EmbeddedEventLoop directly. The stubbed eventLoop will only work if Netty code is calling on the value of `channel.eventLoop()`. For example, if Netty schedules a task with the value of `channel.eventLoop()` and calls `runPendingTasks()` on the real EmbeddedEventLoop, the task will not run.

ejona86: Everything calling _schedule_ should be safe to delay, because it is based on time. It seems every usage of schedule should be able to use the fake clock, with the possible (unlikely) exception of when something is scheduled with 0 or negative delay (in the past).

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
dapengzhang0: Whe mocking, `ctx.executor()` returns the correct type,: `io.netty.channel.EventLoop$$EnhancerByMockitoWithCGLIB$$d741cc0d.schedule(<generated>`. But when it calls `.schedule()`, it will then invoke the delegate:

`method.invoke(fakeClock, arguments)`, where `method` is `EventLoopGroup.schedule()`, `fakeClock` is instance of `FakeClock` not `EventLoopGroup`.

netty/src/test/java/io/grpc/netty/NettyHandlerTestBase.java
dapengzhang0: Okay, let me try to schedule everything with fake clock.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
ejona86: Oh, then make your own answer to call fakeClock directly? (Don't use `delegatesTo()` in that case)

dapengzhang0: Done.

PR #3340, opened 2017-08-11
netty: not to ignore tests for window overriding

### LINKED ISSUE #1175: Connection window should auto-refill
Consider an application with 2 streams, A and B. A receives a stream of messages and the application pops off one message at a time and makes a request on stream B. However, if receiving of data on A has caused the connection window to collapse, B will not be able to receive any data and the application will deadlock. The only way (currently) to get around this is to use multiple connections for the streams, which would needlessly complicate the application code.

The C implementation already solves the problem by auto-refilling the connection window, so Java and the other languages should just follow suit here.

Created PR https://github.com/netty/netty/pull/4423 in Netty to support configuring the local flow controller to auto-refill.  Until that's in, we should probably just set the connection window to MAX_INT for now.

@louiscryan @ejona86 @ctiller @a11r

dapengzhang0:
As noted after `@Ignore`
> Re-enable once https://github.com/grpc/grpc-java/issues/1175 is fixed

#1175 has been closed for a long time, should re-enable the tests.

### DIFF
--- a/netty/src/test/java/io/grpc/netty/NettyClientHandlerTest.java
+++ b/netty/src/test/java/io/grpc/netty/NettyClientHandlerTest.java
@@ -76,7 +76,6 @@ import java.io.InputStream;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TestName;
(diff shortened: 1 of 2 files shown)


dapengzhang0:
test this please

dapengzhang0:
Jenkins, retest this please

dapengzhang0:
Changed the PR description (removing quotes " "), and then Jenkins test passed.

PR #3343, opened 2017-08-11
grpclb: use two-level drop behavior

zhangkun83:
Previously, the round-robin list that the client uses (effective
round-robin list, ERRL) was the received round-robin list (RRRL, from
remote balancer) excluding non-READY backends.  Drop and backend
entries are in the same list.

The problem with it is that when not all backends are READY, drop
entries take a larger proportion in ERRL than they do in the RRRL,
resulting a larger drop ratio than intended.

To fix this, we employ a two-list scheme:

- A _drop list_ (DL) that is out of the RRRL, with the same size and
  the same number of drop entries as the RRRL.

- A _backend list_ (BL) that contains only the backend entries from
  the RRRL, excluding non-READY ones.

For every pick, the client would round-robin on the DL to determine
whether the pick should be dropped.  Only when it's not dropped,
round-robin on the BL to pick the actual backend.

This way, the drop ratio is always equal to the proportion they take
in the RRRL.

### DIFF
--- a/grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
+++ b/grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
@@ -122,7 +122,7 @@ class GrpclbLoadBalancer extends LoadBalancer implements WithLogId {
   private Map<EquivalentAddressGroup, Subchannel> subchannels = Collections.emptyMap();
 
   // Has the same size as the round-robin list from the balancer.
-  // A drop entry from the round-robin list becomes a DropListEntry here.
+  // A drop entry from the round-robin list becomes a DropEntry here.
   // A backend entry from the robin-robin list becomes a null here.
   private List<DropEntry> dropList = Collections.emptyList();
   // Contains only non-drop, i.e., backends from the round-robin list from the balancer.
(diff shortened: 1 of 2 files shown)


zhangkun83:
@markdroth please review the behavior. @dapengzhang0 will take care of the Java part.

dapengzhang0:
Is there a spec doc on it? You forward email to me if it's internal.

zhangkun83:
No, there isn't a doc. Consider the PR description as the spec.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -120,8 +121,14 @@ public PickResult pickSubchannel(PickSubchannelArgs args) {
   private LbStream lbStream;
   private Map<EquivalentAddressGroup, Subchannel> subchannels = Collections.emptyMap();
 
-  private List<RoundRobinEntry> roundRobinList = Collections.emptyList();
-  private SubchannelPicker currentPicker = BUFFER_PICKER;
+  // Has the same size as the round-robin list from the balancer.
+  // A drop entry from the round-robin list becomes a DropListEntry here.
```
dapengzhang0: DropListEntry is not a type, you mean a non-null DropEntry?

zhangkun83: Yes. Fixed.

zhangkun83:
retest this please

PR #3346, opened 2017-08-14
android-interop-testing: remove gms version metadata, add minimum sdk

ericgribkoff:
The `com.google.android.gms.version` is no longer required, and some build tools rely on the minimum SDK being present in `AndroidManifest.xml` rather than solely in `build.gradle`.

### DIFF
--- a/android-interop-testing/app/src/main/AndroidManifest.xml
+++ b/android-interop-testing/app/src/main/AndroidManifest.xml
@@ -2,6 +2,10 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="io.grpc.android.integrationtest" >
 
+    <uses-sdk
+        android:minSdkVersion="9"
+        android:targetSdkVersion="22"/>
+
     <uses-permission android:name="android.permission.INTERNET" />
 
     <application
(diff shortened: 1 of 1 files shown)


android-interop-testing/app/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,10 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="io.grpc.android.integrationtest" >
 
+    <uses-sdk
+        android:minSdkVersion="9"
```
carl-mastrangelo: I think this is at 15 now.  We don't test with anything less.

ericgribkoff:
9 matches the min SDK of the google play services we are using for some
internal tests. Unless there's a good reason to bump this up now, I would
prefer to leave as is to keep the changes minimal here, then do a general
updating of this and all the associated dependencies as a later PR. What do
you think?

On Aug 14, 2017 3:39 PM, "Carl Mastrangelo" <notifications@github.com>
wrote:

*@carl-mastrangelo* commented on this pull request.
------------------------------

In android-interop-testing/app/src/main/AndroidManifest.xml
<https://github.com/grpc/grpc-java/pull/3346#discussion_r133079753>:

> @@ -2,6 +2,10 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="io.grpc.android.integrationtest" >

+    <uses-sdk
+        android:minSdkVersion="9"

I think this is at 15 now. We don't test with anything less.

—
You are receiving this because you authored the thread.

Reply to this email directly, view it on GitHub
<https://github.com/grpc/grpc-java/pull/3346#pullrequestreview-56223678>,
or mute the thread

PR #3351, opened 2017-08-15
core: pass status to ManagedClientTransport.shutdown()

zhangkun83:
This aligns with shutdownNow(), which is already accepting a status.
The status will be propagated to application when RPCs failed because
of transport shutdown, which will become useful information for debug.


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -904,15 +908,15 @@ public void shutdown() {
                   new Runnable() {
                     @Override
                     public void run() {
-                      subchannel.shutdown();
+                      subchannel.shutdown(SHUTDOWN_STATUS);
```
ejona86: This looks like it should have a different status, since the channel hasn't been shutdown.

zhangkun83: You are right. Using a different status.

Issue #3248, opened 2017-07-17
Context: alternative data structure


ejona86:
Right now Context uses a linked list of mutations for implementing its mapping structure. This is serving us reasonably well as it has low mutation cost and is simple to implement. However, it may be using a noticeable amount of CPU time[1] and it is fairly easy to accidentally create very long chains when using async APIs (as discussed in #1926). The main thing we'd want from an alternative data structure is to scale in terms of number of elements instead of number of mutations.

For fun, I investigated binary trees, b-trees, b+trees, and hash trees as possible alternative structures. An order 3 b+tree and HAMT both look pretty reasonable in terms of simplicity ([~200 lines](https://gist.github.com/ejona86/1b0b8f791b72d94abd3bd84d7dcab978)) and allocation size per mutation. I think I like the HAMT most as I found it to have fewer/easier cases to consider.

There's no urgency to this, or very much evidence that it is necessary. But we've been discussing it some, so wanted a place to discuss and report some alternatives.

1. It's unclear whether this is Context's fault. There was a user that was always adding 4 keys, even though only 1 was changed. The user was cleaned up, and while a rough view of the data would seem to imply this improved the performance, there's a lot of guesswork involved. We really need more data of typical Context chain lengths in order to produce a useful benchmark and to help determine per-call costs.

zpencer:
Fixed with #3368

PR #3354, opened 2017-08-17
context: move KeyValueEntries into an abstract class

zpencer:
This will make it a lot easier to plug in the alternate data
structures from #3248 and also has a nice side effect of making
Context easier to understand.

The bloom filter mask of the `Context.Key` is now `hashCode() %
64` because from the `LinkedArrayKeyValueEntries` class we no
longer have insight into how many `Context.Key`s were created.


zpencer:
Synced with team members and decided to just go directly to integrating the hash based struct from #3248 without this refactoring.

PR #3357, opened 2017-08-17
all: bump to 1.7.0-SNAPSHOT

zpencer:
all: bump to 1.7.0-SNAPSHOT

This bump changelist is applied a bit late with respect to the
1.6.0 branch cut. Look at the 1.6.0 to see the source of truth of
where it was cut. Do not assume it is the commit that preceeds
this one.


carl-mastrangelo:
please use standard commit format.

PR #3360, opened 2017-08-18
netty: Handle shutdown and failures during negotiator

ejona86:
NettyClientTransport needs to call close() on the Channel directly
instead of sending a message, since the message would typically be
delayed until negotiation completes.

The closeFuture() closes too early to be helpful, which is very
unfortunate. Using it squelches the negotiator's error handling. We now
rely on the handlers to report shutdown without any back-up. The
handlers error handling has matured, so maybe this is okay.

------

This work was initially triggered by useless error messages if using a Proxy failed. I _thought_ I had already submitted this. But then it bit me again when trying to get Conscrypt working, so I finished up the old commit.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/WriteQueue.java
+++ b/netty/src/main/java/io/grpc/netty/WriteQueue.java
@@ -100,6 +100,18 @@ class WriteQueue {
   }
 
   /**
+   * Enqueue the runnable. It is not safe for another thread to queue an Runnable directly to the
+   * event loop, because it will be out-of-order with writes. This method allows the Runnable to be
+   * processed in-order with writes.
+   */
+  void enqueue(Runnable runnable, boolean flush) {
+    queue.add(new RunnableCommand(runnable));
+    if (flush) {
+      scheduleFlush();
+    }
+  }
+
+  /**
    * Process the queue of commands and dispatch them to the stream. This method is only
    * called in the event loop
    */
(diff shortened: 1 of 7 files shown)


netty/src/main/java/io/grpc/netty/WriteQueue.java
```diff
@@ -100,6 +100,18 @@ ChannelFuture enqueue(QueuedCommand command, ChannelPromise promise, boolean flu
   }
 
   /**
+   * Enqueue the runnable. It is not safe for another thread to queue an Runnable directly to the
+   * event loop, because it will be out-of-order with writes. This method allows the Runnable to be
+   * processed in-order with writes.
+   */
+  void enqueue(Runnable runnable, boolean flush) {
+    queue.add(new RunnableCommand(runnable));
```
carl-mastrangelo: you might consider chaining this with the above enqueue and just use a voidPromise

ejona86: Hmm... I'd also need to "support" the setter `QueuedCommand.promise(ChannelPromise)`. It feels about the same to me. I'm going to leave it as-is; this is a pretty straight-forward method and would be easy to change in the future if it became a larger problem.

ejona86:
Rebased to absorb conflict with #3351. I changed `NettyCLientTransport.shutdownNow()` to pass `reason` to `lifecycleManager` instead of creating its own `Status`; that preserves the behavior after #3351.

PR #3362, opened 2017-08-18
core: ForwardingChannelBuilder

dapengzhang0:
Introduced `ForwardingChannelBuilder`, in the same pattern as `ForwardingClientCall`, `ForwardingServerCall`, `ForwardingNameResolver`, `ForwardingReadableBuffer`, `ForwardingClientCallListener`, and `ForwardingServerCallListener`.

### DIFF
--- a/core/src/main/java/io/grpc/ForwardingDelegateChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ForwardingDelegateChannelBuilder.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2017, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc;
+
+import java.util.List;
+import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A {@link ManagedChannelBuilder} that delegates all its builder method to another builder by
+ * default.
+ *
+ * <p>A subclass of it typically has the following signature:
+ *
+ * <pre>
+ * public class CustomNettyChannelBuilder
+ *     extends DelegatedChannelBuilder&lt;CustomNettyChannelBuilder, NettyChannelBuilder&gt;
+ * </pre>
+ *
+ * @param <T> The type of the subclass extending this abstract class.
+ * @param <DelegateT> The type of the delegate.
+ *
+ * @since 1.6.0
+ */
+@ExperimentalApi
+public abstract class ForwardingDelegateChannelBuilder<
+        T extends ForwardingDelegateChannelBuilder<T, DelegateT>,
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/grpc/ForwardingDelegateChannelBuilder.java
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * <pre>
+ * public class CustomNettyChannelBuilder
+ *     extends DelegatedChannelBuilder&lt;CustomNettyChannelBuilder, NettyChannelBuilder&gt;
+ * </pre>
+ *
+ * @param <T> The type of the subclass extending this abstract class.
+ * @param <DelegateT> The type of the delegate.
+ *
+ * @since 1.6.0
```
dapengzhang0: Done.

dapengzhang0:
Jenkins, retest this please

core/src/main/java/io/grpc/ForwardingDelegateChannelBuilder.java
```diff
@@ -0,0 +1,149 @@
...
+ *     extends ForwardingDelegateChannelBuilder&lt;CustomChannelBuilder, NettyChannelBuilder&gt;
+ * </pre>
+ *
+ * @param <T> The type of the subclass extending this abstract class.
+ * @param <DelegateT> The type of the delegate.
+ *
+ * @since 1.7.0
+ */
+@ExperimentalApi("https://github.com/grpc/grpc-java/issues/3363")
+public abstract class ForwardingDelegateChannelBuilder<
```
ejona86: To match the other forwarding classes, wouldn't this be `ForwardingChannelBuilder`?

core/src/main/java/io/grpc/ForwardingDelegateChannelBuilder.java
```diff
@@ -0,0 +1,149 @@
...
+        T extends ForwardingDelegateChannelBuilder<T, DelegateT>,
+        DelegateT extends ManagedChannelBuilder<DelegateT>>
```
ejona86: Why have `DelegateT` instead of `ManagedChannelBuilder<DelegateT>`?

core/src/main/java/io/grpc/ForwardingDelegateChannelBuilder.java
dapengzhang0: I also thought about `ForwardingChannelBuilder` first, but worried that it may not be as accurate in meaning. I don't mind to use `ForwardingChannelBuilder` though.

core/src/main/java/io/grpc/ForwardingDelegateChannelBuilder.java
dapengzhang0: If `DelegateT` is `NettyChannelBuilder` for the subclass, I can call `delegate().eventLoopGroup()` without a cast, this is useful if the subclass is still an abstract class.

Effective Java Item 28:  **Do not use wildcard types as return types** and **If the user of a class has to think about wildcard types, there is probably something wrong with the class’s API**.  If returning `ManagedChannelBuilder<?>` would be better in this situation, a reason for that should be provided.

ejona86: > If DelegateT is NettyChannelBuilder for the subclass, I can call delegate().eventLoopGroup() without a cast

You can do that already, by calling `delegate` instead of `delegate()` or making the return value more specific in the implementation: `NettyChannelBuilder delegate() { return delegate; }`.

I agree with Effective Java, but think it is talking about a different situation. That point was not talking about return values of abstract methods as used by super classes; in this case the `delegate()` method is really used more like a method argument, just with the super class being able to call it whenever it wishes.

I don't see anyone as benefiting from the `DelegateT`; anyone who wants something else can already use something else. But _everyone_ pays the cost of its presence.

PR #3368, opened 2017-08-21
context: hashtrie based keyvalue store

zpencer:
This is the hashtrie data structure authored by @ejona86 

The linked list key value store is known cause problems in pathological cases where users keep updating the same key(s) over and over. This copy on write tree will have reads that cost `O(lgN)` where N is the number of keys in the map, rather than `O(M)` where M is the total number of `put` operations in the case of the linked arrays.

Also:
- added some unit tests
- ran a test putting random keys into the map and comparing the result with a java.util.HashMap to verify sanity. The test passes but I won't check it into the repo because it 
 takes a long time to run: https://gist.github.com/zpencer/12cb435235d171c1fe09aef18825fad0

Benchmarks:
linked array: https://gist.github.com/zpencer/7455cb146d39cbf5628e0608a94ad70c
hashtrie: https://gist.github.com/zpencer/4204505ccac9a32d413b3f2353a1e564

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -97,7 +97,8 @@ public class Context {
 
   private static final Logger log = Logger.getLogger(Context.class.getName());
 
-  private static final Object[] EMPTY_ENTRIES = new Object[0];
+  private static final PersistentHashArrayMappedTrie<Key<?>, Object> EMPTY_ENTRIES =
+      new PersistentHashArrayMappedTrie<Key<?>, Object>();
 
   private static final Key<Deadline> DEADLINE_KEY = new Key<Deadline>("deadline");
 
@@ -108,7 +109,7 @@ public class Context {
    * <p>Never assume this is the default context for new threads, because {@link Storage} may define
    * a default context that is different from ROOT.
    */
-  public static final Context ROOT = new Context(null);
+  public static final Context ROOT = new Context(null, EMPTY_ENTRIES, false, false);
 
   // Lazy-loaded storage. Delaying storage initialization until after class initialization makes it
   // much easier to avoid circular loading since there can still be references to Context as long as
@@ -180,16 +181,11 @@ public class Context {
   }
 
   private final Context parent;
-  // A 64 bit bloom filter of all the Key.bloomFilterMask values in this Context and all the parents
-  // this will help us detect failed lookups faster.  In fact if there are fewer than 64 key objects
-  // this will be perfect (though we don't currently take advantage of that fact).
-  private final long keyBloomFilter;
-  // Alternating Key, Object entries
-  private final Object[] keyValueEntries;
   private final boolean cascadesCancellation;
   private ArrayList<ExecutableListener> listeners;
   private CancellationListener parentListener = new ParentListener();
   private final boolean canBeCancelled;
+  final PersistentHashArrayMappedTrie<Key<?>, Object> keyValueEntries;
 
   /**
    * Construct a context that cannot be cancelled and will not cascade cancellation from its parent.
@@ -197,8 +193,7 @@ public class Context {
   private Context(Context parent) {
     this.parent = parent;
     // Not inheriting cancellation implies not inheriting a deadline too.
-    keyValueEntries = new Object[] {DEADLINE_KEY, null};
-    keyBloomFilter = computeFilter(parent, keyValueEntries);
+    keyValueEntries = parent.keyValueEntries.put(DEADLINE_KEY, null);
     cascadesCancellation = false;
     canBeCancelled = false;
   }
@@ -207,10 +202,9 @@ public class Context {
    * Construct a context that cannot be cancelled but will cascade cancellation from its parent if
    * it is cancellable.
    */
-  private Context(Context parent, Object[] keyValueEntries) {
+  private Context(Context parent, PersistentHashArrayMappedTrie<Key<?>, Object> keyValueEntries) {
     this.parent = parent;
     this.keyValueEntries = keyValueEntries;
-    keyBloomFilter = computeFilter(parent, keyValueEntries);
     cascadesCancellation = true;
     canBeCancelled = this.parent != null && this.parent.canBeCancelled;
   }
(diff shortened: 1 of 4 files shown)


context/src/main/java/io/grpc/Context.java
```diff
@@ -706,21 +689,22 @@ private Object lookup(Key<?> key) {
...
      * or listen for its expiration as the parent context will already be listening for it.
      */
-    private static Object[] deriveDeadline(Context parent, Deadline deadline) {
+    private static PersistentHashArrayMappedTrie<Key<?>, Object> deriveDeadline(
+        Context parent, Deadline deadline) {
       Deadline parentDeadline = DEADLINE_KEY.get(parent);
       return parentDeadline == null || deadline.isBefore(parentDeadline)
-          ? new Object[] {DEADLINE_KEY, deadline}
-          : EMPTY_ENTRIES;
+          ? parent.keyValueEntries.put(DEADLINE_KEY, deadline) :
```
ejona86: `:` goes on next line

context/src/main/java/io/grpc/PersistentHashArrayMappedTrie.java
```diff
@@ -0,0 +1,287 @@
...
+ * linearly. Delete is not supported, but replacement is. The implementation
+ * favors simplicity and low memory allocation during insertion. Although the
+ * asymptotics are good, it is optimized for small sizes like less than 20;
+ * "unbelievably large" would be 100.
+ *
+ * <p>Inspired by popcnt-based compression seen in Ideal Hash Trees, Phil
+ * Bagwell (2000). The rest of the implementation is ignorant of/ignores the
+ * paper.
+ */
+public final class PersistentHashArrayMappedTrie<K,V> {
```
ejona86: package-private

context/src/main/java/io/grpc/PersistentHashArrayMappedTrie.java
```diff
@@ -0,0 +1,287 @@
...
+      return new PersistentHashArrayMappedTrie<K,V>(new Leaf<K,V>(key, value));
+    } else {
+      return new PersistentHashArrayMappedTrie<K,V>(root.put(key, value, key.hashCode(), 0));
+    }
+  }
+
+  // Not actually annotated to avoid depending on guava
+  // @VisibleForTesting
+  static final class Leaf<K,V> implements Node<K,V> {
+    final K key;
```
ejona86: private, for `key` and `value` (They don't appear used in the tests)

context/src/main/java/io/grpc/PersistentHashArrayMappedTrie.java
```diff
@@ -0,0 +1,287 @@
...
+ * favors simplicity and low memory allocation during insertion. Although the
+ * asymptotics are good, it is optimized for small sizes like less than 20;
+ * "unbelievably large" would be 100.
+ *
+ * <p>Inspired by popcnt-based compression seen in Ideal Hash Trees, Phil
+ * Bagwell (2000). The rest of the implementation is ignorant of/ignores the
+ * paper.
+ */
+public final class PersistentHashArrayMappedTrie<K,V> {
+  final Node<K,V> root;
```
ejona86: private (doesn't appear used in tests)

context/src/main/java/io/grpc/PersistentHashArrayMappedTrie.java
```diff
@@ -0,0 +1,287 @@
...
+    @SuppressWarnings("unchecked")
+    public CollisionLeaf(K key1, V value1, K key2, V value2) {
+      this((K[]) new Object[] {key1, key2}, (V[]) new Object[] {value1, value2});
+      assert key1 != key2;
+      assert key1.hashCode() == key2.hashCode();
+    }
+
+    // Not actually annotated to avoid depending on guava
+    // @VisibleForTesting
+    CollisionLeaf(K[] keys, V[] values) {
```
ejona86: private? It looks like the usages in tests could use the other constructor instead. (This one does less verification, so was intended to be used rarely.)

context/src/main/java/io/grpc/PersistentHashArrayMappedTrie.java
```diff
@@ -0,0 +1,287 @@
...
+      int indexBit1 = indexBit(hash1, bitsConsumed);
+      int indexBit2 = indexBit(hash2, bitsConsumed);
+      if (indexBit1 == indexBit2) {
+        Node<K,V> node = combine(node1, hash1, node2, hash2, bitsConsumed + BITS);
+        @SuppressWarnings("unchecked")
+        Node<K,V>[] values = (Node<K,V>[]) new Node<?,?>[] {node};
+        return new CompressedIndex<K,V>(indexBit1, values);
+      } else {
+        // Make node1 the smallest
+        if (uncompressedIndex(hash1, bitsConsumed) > uncompressedIndex(hash2, bitsConsumed)) {
```
ejona86: Maybe simplify this to `if (indexBit1 > indexBit2)`?

(If you do, remove uncompressedIndex() as well, by inlining it with indexBit again)

zpencer: If `indexBit1` is (1<<31) then we would mistakenly think it's a negative number of a large magnitude. I think we'd need to do some bit twiddling like `(indexBit1 - 1) & indexBit2 != 0` but it would be less readable.

zpencer:
Locked down visibility wherever we can.

context/src/test/java/io/grpc/ContextTest.java
```diff
@@ -759,7 +759,8 @@ public void initContextWithCustomClassLoaderWithCustomLogger() throws Exception
     StaticTestingClassLoader classLoader =
         new StaticTestingClassLoader(
             getClass().getClassLoader(),
-            Pattern.compile("(io\\.grpc\\.Context.*)|(io\\.grpc\\.ThreadLocalContextStorage.*)"));
+            Pattern.compile("(io\\.grpc\\.Context.*)|(io\\.grpc\\.ThreadLocalContextStorage.*)|"
+                + "(io\\.grpc\\.PersistentHashArrayMappedTrie)"));
```
ejona86: Why was this necessary? It actually looks partially broken (although in a way that probably doesn't cause any failure), since the nested classes wouldn't be copied.

context/src/main/java/io/grpc/PersistentHashArrayMappedTrie.java
ejona86: Right... Yeah. We could cast to `long` (`indexBit1 & 0xFFFFFFFFL`), or we could pass `bitmap` to `compressedIndex()`... but it's all a wash. LGTM

context/src/test/java/io/grpc/ContextTest.java
zpencer: Without this, I get this error after making the trie class package-private:
```

java.lang.IllegalAccessError: tried to access class io.grpc.PersistentHashArrayMappedTrie from class io.grpc.Context

	at io.grpc.Context.<clinit>(Context.java:100)
	at io.grpc.ContextTest$LoadMeWithStaticTestingClassLoader.run(ContextTest.java:904)
	at io.grpc.ContextTest.initContextWithCustomClassLoaderWithCustomLogger(ContextTest.java:766)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

ejona86: Oh, the package visibility kicks in because... I guess we are implicitly redefining the package. And it doesn't break further because you aren't adding any keys to the context. For sanity's sake, add `.*` after `Trie` or replace the entire string with `"io\\.grpc\\.[^.]*"` instead.

context/src/test/java/io/grpc/ContextTest.java
```diff
@@ -758,8 +758,7 @@ public void expiredDeadlineShouldCancelContextImmediately() {
   public void initContextWithCustomClassLoaderWithCustomLogger() throws Exception {
     StaticTestingClassLoader classLoader =
         new StaticTestingClassLoader(
-            getClass().getClassLoader(),
-            Pattern.compile("(io\\.grpc\\.Context.*)|(io\\.grpc\\.ThreadLocalContextStorage.*)"));
+            getClass().getClassLoader(), Pattern.compile("(io\\.grpc\\.[^.]+)"));
```
ejona86: nit: no need for the `()` in the regex.

adriancole:
ps while a pathological case, did you get to the bottom of why users were re-updating keys so many times that this became an issue?

zpencer:
I don't think we identified the exact code path causing this to happen. However, we're pretty certain it's an application specific issue in a pretty large pre-existing system that we integrated with inside Google, so I won't expect most users of gRPC to encounter this.

PR #3369, opened 2017-08-21
grpclb: refactor main GRPCLB logic out of GrpclbLoadBalancer.

zhangkun83:
GrpclbLoadBalancer can work in non-GRPCLB (delegate) mode according to
name resolution results.  Previously the policy selection, delegation
and GRPCLB logic are in the same file, which is not very readable.  It
will get worse as we going to implement policy fallback logic soon.
This PR refactors the GRPCLB logic out, and makes GrpclbLoadBalancer
focus on the policy selection and delegation logic.

__Notes for reviewers__: github doesn't recognize the copy of major chunks of code from `GrpclbLoadBalancer.java` to `GrpclbState.java`. You may want to fetch my branch
and view the diff with `-C`, e.g.:
```bash
$ git remote add zhangkun83 git@github.com:zhangkun83/grpc-java.git
$ git fetch zhangkun83 grpclb_fallback
$ git difftool --tool=meld -C 8634632019ee7ea35e44c6c8749343ec99ecc858 zhangkun83/grpclb_fallback
```

### DIFF
--- a/grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
+++ b/grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
@@ -132,6 +132,9 @@ final class GrpclbState {
     maybeUpdatePicker();
   }
 
+  /**
+   * Set the address of the balancer, and create connection if not yet connected.
+   */
   void setLbAddress(LbAddressGroup newLbAddressGroup) {
     lbAddressGroup = checkNotNull(newLbAddressGroup, "newLbAddressGroup");
     startLbComm();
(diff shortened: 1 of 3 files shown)


grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
```diff
@@ -0,0 +1,645 @@
...
+  private final String serviceName;
+  private final Helper helper;
+  private final TimeProvider time;
+  private final ScheduledExecutorService timerService;
+
+  private static final Attributes.Key<AtomicReference<ConnectivityStateInfo>> STATE_INFO =
+      Attributes.Key.of("io.grpc.grpclb.GrpclbLoadBalancer.stateInfo");
+
+  @Nullable
+  private LbAddressGroup lbAddressGroup;
```
zhangkun83: Good suggestion! Done.

PR #3379, opened 2017-08-23
Fix gson / opencensus bazel naming mixup

duelinmarkers:
In 72b9ee2 (see also #3328, #3338), gson kept its old name while opencensus got what should have
become gson's new name instead of a fixed opencensus name.

I think this resolves that.

### DIFF
--- a/core/BUILD.bazel
+++ b/core/BUILD.bazel
@@ -43,7 +43,7 @@ java_library(
         "@com_google_errorprone_error_prone_annotations//jar",
         "@com_google_guava_guava//jar",
         "@com_google_instrumentation_instrumentation_api//jar",
-        "@com_google_code_gson_gson//jar",
+        "@io_opencensus_opencensus_api//jar",
     ],
 )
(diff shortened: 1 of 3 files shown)


duelinmarkers:
I signed it!

ejona86:
Okay to test

duelinmarkers:
Could we have 5c902c938489a0f04642f1c7d206555e3b659ffe and 7d0229a1f589c886c1b4088bbb93c62cbd7ad1b6 cherry-picked onto the v1.6.x branch so the first 1.6 release will be useable from bazel projects conveniently? I can submit a PR for this if it would be helpful. Thanks.

ejona86:
@duelinmarkers, yeah, I think it makes sense to backport them. I've added the backport label for us to backport them before the release.

pgr0ss:
Thanks @duelinmarkers for fixing my mistake. Looks I screwed up my scripting for bulk updating.

PR #3380, opened 2017-08-23
testing: add FrozenClock

zpencer:
Move `FakeClock.java` from internal `core` to `testing`.  This achieves what #2571 was trying to do.


testing/src/main/java/io/grpc/FrozenClock.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.grpc.internal;
+package io.grpc;
```
ejona86: We don't want this in this package. (I know that you must have it in this package for visibility.)

testing/src/main/java/io/grpc/FrozenClock.java
```diff
@@ -180,28 +193,43 @@ public String toString() {
...
   /**
+   * Returns a {@link Deadline} whose remaining time is controlled by this class.
+   *
+   * <p>If the deadline is to be passed into
+   * {@link Context#withDeadline(Deadline, ScheduledExecutorService)}, then
+   * {@link #getScheduledExecutorService()} should be used as the executor to ensure deadline
+   * cancellations are not scheduled based on system time.
+   */
+  public Deadline createDeadlineAfter(int duration, TimeUnit unit) {
+    return Deadline.after(duration, unit, deadlineTicker);
```
ejona86: This is not safe, as mixing "normal" Deadlines and "fake" Deadlines will lead to undefined/weird results. This is why Ticker wasn't exposed earlier.

PR #3381, opened 2017-08-23
all: Keep artifacts dependencyConvergence-clean

ejona86:
Maven Enforcer's dependencyConvergence is commonly used in Spring
projects, as it is inherited by all starter projects[1]. While I find
that option to be crazy and harmful (and would instead support
requireUpperBoundDeps), it does bother people.

1. https://github.com/spring-projects/spring-boot/blob/v1.5.6.RELEASE/spring-boot-starters/pom.xml#L94

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -219,11 +219,20 @@ subprojects {
     // Define a separate configuration for managing the dependency on Jetty ALPN agent.
     configurations {
         alpnagent
+
+        compile {
+            // Detect Maven Enforcer's dependencyConvergence failures. We only
+            // care for artifacts used as libraries by others.
+            if (!(project.name in ["grpc-benchmarks"])) {
+                //resolutionStrategy.failOnVersionConflict()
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -219,11 +219,20 @@ subprojects {
     // Define a separate configuration for managing the dependency on Jetty ALPN agent.
     configurations {
         alpnagent
+
+        compile {
+            // Detect Maven Enforcer's dependencyConvergence failures. We only
+            // care for artifacts used as libraries by others.
+            if (!(project.name in ["grpc-benchmarks"])) {
+                //resolutionStrategy.failOnVersionConflict()
```
zhangkun83: Uncomment?

ejona86: Doh. Done.

When there's a failure you are sort of required to comment it out in order to debug where the failure is coming from. It's unfortunate.

PR #3384, opened 2017-08-24
interop-testing: fix dealineExceeded test by increasing time interval

dapengzhang0:
Increased interval from 20,000 micros (i.e. 20 millis) to 20 seconds.


dapengzhang0:
Test results before #3387 and after #3389

PR #3385, opened 2017-08-24
context: make Ticker a public API

zpencer:
make Ticker a public API

After discussing with @ejona86 offline, this approach makes more
sense than exposing the FakeClock class.


carl-mastrangelo:
The problem with this is that a lot of times you may not control where the ticker comes from.  gRPC code often creates deadlines that don't have access to the ticker, which makes having a custom Ticker not that valuable.  

This was the problem with my original PR that did pretty much the same thing here.

zpencer:
Updated code to illustrate the idea @carl-mastrangelo and I discussed offline: inject the ticket into the Context so that gRPC can use it.

ejona86:
I'm not excited about the `TICKER_KEY`. Among other things, I find it highly unexpected that these produce different results:
```java
context.withDeadlineAfter(1, SECONDS, scheduler);
context.withDeadline(Deadline.after(1, SECONDS), scheduler);
```

Maybe InProcess transport should allow the Ticker to be provided. Note that Netty doesn't behave properly because the `scheduler` will an event loop which uses real time.

PR #3390, opened 2017-08-24
netty: Update to Netty 4.1.15

ejona86:
We'll need to figure out what we want to do about the deprecated API.
We'll probably just drop the verification check, but need to look into
it a bit.

----

I'll wait for the release announcement before actually merging.


ejona86:
retest this please

PR #3392, opened 2017-08-24
core: advance past empty buffer in CompositeReadableBuffer

ericgribkoff:
If the initial buffer in a CompositeReadableBuffer has 0 readable bytes, invoking readUnsignedByte() on the composite triggers the following exception:

```
java.lang.IndexOutOfBoundsException
	at io.grpc.internal.AbstractReadableBuffer.checkReadable(AbstractReadableBuffer.java:72)
	at io.grpc.internal.ReadableBuffers$ByteArrayWrapper.readUnsignedByte(ReadableBuffers.java:158)
	at io.grpc.internal.CompositeReadableBuffer$1.readInternal(CompositeReadableBuffer.java:71)
	at io.grpc.internal.CompositeReadableBuffer$ReadOperation.read(CompositeReadableBuffer.java:220)
	at io.grpc.internal.CompositeReadableBuffer.execute(CompositeReadableBuffer.java:178)
	at io.grpc.internal.CompositeReadableBuffer.readUnsignedByte(CompositeReadableBuffer.java:74)
	at io.grpc.internal.CompositeReadableBufferTest.readUnsignedByteShouldSkipZeroLengthBuffer(CompositeReadableBufferTest.java:77)
```

This change avoids the exception by invoking `advanceBufferIfNecessary()` before the for loop in `execute`, which will close the empty buffer and skip forward.

It seems that empty buffers should probably not be put into a CompositeReadableBuffer, but I didn't see anything to that effect in the Javadoc, and it's currently allowed by the code. An alternative fix would be to disallow empty buffers being added to the composite.

### DIFF
--- a/core/src/main/java/io/grpc/internal/CompositeReadableBuffer.java
+++ b/core/src/main/java/io/grpc/internal/CompositeReadableBuffer.java
@@ -166,6 +166,10 @@ public class CompositeReadableBuffer extends AbstractReadableBuffer {
   private void execute(ReadOperation op, int length) {
     checkReadable(length);
 
+    if (!buffers.isEmpty()) {
+      advanceBufferIfNecessary();
+    }
+
     for (; length > 0 && !buffers.isEmpty(); advanceBufferIfNecessary()) {
       ReadableBuffer buffer = buffers.peek();
       int lengthToCopy = Math.min(length, buffer.readableBytes());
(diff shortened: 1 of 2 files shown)


ejona86:
Yeah, the code currently assumes the current buffer isn't empty. I think ignoring empty buffers (just closing them immediately when added) would be the least bug-prone.

ejona86:
Ah, meh. Yeah, this fits in fine with some of the other code. This is fine as-is.

ejona86:
@ericgribkoff, should this be backported to 1.6.x? Or did this just impact the new code you've sent out?

ericgribkoff:
I don't think this needs to be backported. It actually doesn't even impact the code in the PR I sent out, I just ran into this while working on an earlier iteration of the PR and it seemed worth checking for.

PR #3393, opened 2017-08-24
all: update to proto 3.4.0

carl-mastrangelo:


README.md
```diff
@@ -85,7 +85,7 @@ For protobuf-based codegen integrated with the Maven build system, you can use
       <artifactId>protobuf-maven-plugin</artifactId>
       <version>0.5.0</version>
       <configuration>
-        <protocArtifact>com.google.protobuf:protoc:3.3.0:exe:${os.detected.classifier}</protocArtifact>
+        <protocArtifact>com.google.protobuf:protoc:3.4.0:exe:${os.detected.classifier}</protocArtifact>
```
ejona86: We update this at release time. It will break as-is because it uses new protoc with old runtime, which isn't allowed.

carl-mastrangelo: Done.

Issue #3025, opened 2017-05-19
Stop telling people Jetty ALPN has not been configured


ejona86:
If both OpenSsl (tcnative) and Jetty ALPN are unavailable then we end up failing with ["Jetty ALPN/NPN has not been properly configured"](https://github.com/grpc/grpc-java/blob/v1.3.0/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java#L174). We should tell them to configure tcnative, and not encourage the use of Jetty ALPN.

conorgriffin:
I have configured tcnative (as far as I can tell) but I get this error.

```
Exception in thread "main" java.lang.IllegalArgumentException: Jetty ALPN/NPN has not been properly configured.
	at io.grpc.netty.GrpcSslContexts.selectApplicationProtocolConfig(GrpcSslContexts.java:174)
	at io.grpc.netty.GrpcSslContexts.configure(GrpcSslContexts.java:151)
	at io.grpc.netty.GrpcSslContexts.configure(GrpcSslContexts.java:139)
	at io.grpc.netty.GrpcSslContexts.forServer(GrpcSslContexts.java:130)
```

If I set the log level to debug I actually get something more helpful which is the following message prior to the exception above:

```
java.lang.UnsatisfiedLinkError: no provided in java.library.path
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1867) ~[na:1.8.0_121]
	at java.lang.Runtime.loadLibrary0(Runtime.java:870) ~[na:1.8.0_121]
	at java.lang.System.loadLibrary(System.java:1122) ~[na:1.8.0_121]
	at org.apache.tomcat.jni.Library.<init>(Library.java:80) ~[tomcat-embed-core-8.5.6.jar:8.5.6]
	at org.apache.tomcat.jni.Library.initialize(Library.java:180) ~[tomcat-embed-core-8.5.6.jar:8.5.6]
	at io.netty.handler.ssl.OpenSsl.initializeTcNative(OpenSsl.java:417) ~[netty-handler-4.1.8.Final.jar:4.1.8.Final]
	at io.netty.handler.ssl.OpenSsl.<clinit>(OpenSsl.java:101) ~[netty-handler-4.1.8.Final.jar:4.1.8.Final]
	at io.grpc.netty.GrpcSslContexts.defaultSslProvider(GrpcSslContexts.java:158) [grpc-netty-1.3.0.jar:1.3.0]
[...]

PR #3394, opened 2017-08-24
netty,okhttp: include more detail about ALPN missing

carl-mastrangelo:
Additionally: include info on why android classes are missing


carl-mastrangelo:
Updates #3025

PR #3395, opened 2017-08-25
core: add gzip inflater for compressed streams

ericgribkoff:
This creates a gzip parser capable of inflating streaming gzipped data on demand. In a follow-up PR, this class will be used for full-stream decompression, with `MessageDeframer` passing incoming gzipped data frames to `GzipInflatingBuffer#addGzippedBytes(CompositeReadableBuffer gzippedFrame)` and getting inflated data back via `GzipInflatingBuffer#inflateBytes(int bytesRequest, CompositeReadableBuffer nextFrame)`.


ericgribkoff:
@ejona86 Added a call to `inflater.end()` upon reaching the end of a gzipped stream without sufficient bytes to immediately begin processing a concatenated stream. This required tracking a bit of additional state (awareness of the bytes written to the inflater's input buffer) directly in `GzipInflatingBuffer`, as we can no longer rely on a potentially closed inflater to tell us what data it has and hasn't looked at.

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -0,0 +1,514 @@
...
+     * caching the current CRC value when necessary before invoking this method.
+     */
+    private int readUnsignedByte() {
+      int bytesRemainingInInflater = inflaterInputEnd - inflaterInputStart;
+      int b;
+      if (bytesRemainingInInflater > 0) {
+        b = inflaterInput[inflaterInputStart] & 0xFF;
+        if (inflater != null) {
+          inflater.reset();
+          inflater.setInput(inflaterInput, inflaterInputStart + 1, bytesRemainingInInflater - 1);
```
ejona86: Leaving the `inflaterBuf` in the `inflater` even when not inflating is weird to me. Why wouldn't you just have the inflater ready when you are inflating... That would remove the `setInput()` noise here.

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -0,0 +1,514 @@
...
+   *
+   * <p>This method may consume gzipped bytes without writing any data to {@code bufferToWrite}, and
+   * may also write data to {@code bufferToWrite} without consuming additional gzipped bytes (if the
+   * inflater on an earlier call consumed the bytes necessary to produce output).
+   *
+   * @param n max number of bytes to inflate
+   * @param bufferToWrite destination for inflated data
+   * @return gzipped bytes consumed by the call (NOT the number of inflated bytes written)
+   */
+  int inflateBytes(int n, CompositeReadableBuffer bufferToWrite)
```
ejona86: I'd really prefer the output buffer to be passed in here. `(byte[], int, int)`-style like `InputStream.read()` and `inflater.inflate()`.

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -0,0 +1,514 @@
...
+    } else {
+      // Empty buffer allocated on previous call
+      bytesToInflate = Math.min(n, inflaterOutput.length);
+    }
+
+    int bytesNeeded;
+    boolean madeProgress = true;
+    while (madeProgress && (bytesNeeded = bytesToInflate - inflaterOutputEnd) > 0) {
+      switch (state) {
+        case HEADER_MAGIC:
```
ejona86: That's a _lot_ of cases for the HEADER. Can we reduce it by a few?

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -0,0 +1,514 @@
...
+      }
+
+      return true;
+    } catch (DataFormatException e) {
+      // Wrap the exception so tests can check for a specific prefix
+      throw new DataFormatException("Inflater data format exception: " + e.getMessage());
+    }
+  }
+
+  private boolean initializeInflater() {
```
ejona86: Could you move this method to just before inflate(), so they are roughly in processing order/case order? (It doesn't need to be perfect, but this one is easy to fix)

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -0,0 +1,514 @@
...
+      inflater.setInput(inflaterInput, inflaterInputStart, bytesToAdd);
+      state = State.INFLATING;
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  private boolean processHeaderMagic() throws ZipException {
+    if (gzipMetadataReader.readableBytes() >= UNSIGNED_SHORT_SIZE) {
```
ejona86: It'd be nice if these checks were generally inverted. That would move the small case early (so it's easier to see what case it relates to) and it would allow avoiding the need for an else entirely (since the if will return).

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -0,0 +1,514 @@
...
+    while (gzipMetadataReader.readableBytes() > 0) {
+      if (gzipMetadataReader.readUnsignedByte() == 0) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /** Reads unsigned short in Little-Endian byte order. */
+  private int readUnsignedShort(GzipMetadataReader buffer) {
```
ejona86: Move these to `GzipMetadataReader`?

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -0,0 +1,514 @@
...
+    } else {
+      return false;
+    }
+  }
+
+  private boolean processTrailer() throws ZipException {
+    if (gzipMetadataReader.readableBytes() >= GZIP_TRAILER_SIZE) {
+      if (crc.getValue() != (readUnsignedInt(gzipMetadataReader))
+          ||
+          (readUnsignedInt(gzipMetadataReader) != expectedGzipTrailerIsize)) {
```
ejona86: Put on same line as || ? Can also remove extraneous parens.

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -0,0 +1,514 @@
...
+      state = State.INITIALIZE_INFLATER;
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  private boolean processTrailer() throws ZipException {
+    if (gzipMetadataReader.readableBytes() >= GZIP_TRAILER_SIZE) {
+      if (crc.getValue() != (readUnsignedInt(gzipMetadataReader))
```
ejona86: Remove extraneous parens around `readUnsignedInt`.

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ejona86: Put `readUnsignedInt` on the same side of the comparison? It's on the left here, but in the right above.

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ejona86: In earlier discussion I mentioned just allocating the entire message. I was only thinking about mobile. For us to turn this on for all users, we probably do want to do the chunking, granted it could be at a much larger size (like 2 MB).

core/src/test/java/io/grpc/internal/GzipInflatingBufferTest.java
```diff
@@ -0,0 +1,629 @@
...
+
+  private GzipInflatingBuffer gzipInflatingBuffer;
+  private CompositeReadableBuffer outputBuffer;
+
+  @Before
+  public void setUp() {
+    gzipInflatingBuffer = new GzipInflatingBuffer();
+    outputBuffer = new CompositeReadableBuffer();
+    try {
+      InputStream inputStream = getClass().getResourceAsStream(UNCOMPRESSABLE_FILE);
```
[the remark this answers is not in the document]
ericgribkoff: Done

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ericgribkoff: Done - the chunking will be done by message deframer

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ericgribkoff: Condensed the three cases for header magic, compression method, and remaining bytes. The rest of the cases are cleaner to keep separated, I think

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ericgribkoff: Done

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ericgribkoff: Done

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ericgribkoff: Done

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ericgribkoff: Done

ericgribkoff: Done

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ericgribkoff: Done

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -300,215 +287,162 @@ private int fillInflatedBuf(int n) throws DataFormatException, ZipException {
...
+      int bytesConsumedDelta = inflater.getTotalIn() - inflaterTotalIn;
+      bytesConsumed += bytesConsumedDelta;
+      inflaterInputStart += bytesConsumedDelta;
+      crc.update(b, off, n);
+
+      if (inflater.finished()) {
+        // Save bytes written to check against the trailer ISIZE
+        expectedGzipTrailerIsize = (inflater.getBytesWritten() & 0xffffffffL);
+        if (gzipMetadataReader.readableBytes() <= GZIP_HEADER_MIN_SIZE + GZIP_TRAILER_SIZE) {
+          // We don't have enough bytes to begin inflating a concatenated gzip stream, drop context
```
ejona86: Do this in the trailer handling.

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -300,215 +287,162 @@ private int fillInflatedBuf(int n) throws DataFormatException, ZipException {
...
-    return false;
   }
 
-  /** Reads unsigned short in Little-Endian byte order. */
-  private int readUnsignedShort(GzipMetadataReader buffer) {
-    return buffer.readUnsignedByte() | (buffer.readUnsignedByte() << 8);
+  private boolean fill() {
+    checkState(inflater != null, "inflater is null");
+    checkState(inflaterInputStart == inflaterInputEnd, "inflaterInput has unconsumed bytes");
+    int bytesToAdd = Math.min(gzippedData.readableBytes(), INFLATE_BUFFER_SIZE);
```
ejona86: s/INFLATE_BUFFER_SIZE/inflaterInput.length/

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -300,215 +287,162 @@ private int fillInflatedBuf(int n) throws DataFormatException, ZipException {
...
+    if (bytesToAdd > 0) {
```
ejona86: invert condition and return `false` early to avoid the indentation in the successful case.

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -199,6 +201,9 @@ boolean hasPartialData() {
    */
   void addGzippedBytes(ReadableBuffer buffer) {
     checkState(!closed, "GzipInflatingBuffer is closed");
+    if (buffer.readableBytes() > 0) {
```
ejona86: How about:
```
if (buffer.readableBytes() == 0) {
  return;
}
isStalled = false
```

Or just ignoring the case (everything should still function correctly, just be slightly less efficient. But we don't really expect that to happen, so no point optimizing.)

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -287,6 +292,10 @@ int inflateBytes(byte[] b, int offset, int length) throws DataFormatException, Z
           throw new AssertionError("Invalid state: " + state);
       }
     }
+    // If we finished a gzip block, check if we have enough bytes to read another header
```
ejona86: Why? If we ignore that case won't we just notice the next time we try to read?

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -230,6 +236,16 @@ int getAndResetBytesConsumed() {
   }
 
   /**
+   * Reports bytes consumed by the inflater since the last invocation of this method, then resets
+   * the count to zero.
+   */
+  int getAndResetDeflatedBytesConsumed() {
```
ejona86: This is used for the "wire size"? It seems fair to include the gzip overhead as part of the wire size (i.e., use `getAndResetBytesConsumed()`). We're doing that with per-message compression today.

I'm okay with excluding the gzip overhead (like this method does), but it looks like it just adds Yet More Stuff to track. So I'd rather just keep it simple.

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -190,7 +192,7 @@ boolean isStalled() {
    */
   boolean hasPartialData() {
     checkState(!closed, "GzipInflatingBuffer is closed");
-    return gzippedData.readableBytes() != 0 || (inflaterInputStart < inflaterInputEnd);
+    return gzipMetadataReader.readableBytes() != 0;
```
ejona86: `&& state == HEADER`

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ericgribkoff: I spoke with @zhangkun83 offline about this. Currently [MessageDeframer](https://github.com/grpc/grpc-java/blob/7e534ed704441cdef7f2af6c094148acf3dabd8b/core/src/main/java/io/grpc/internal/MessageDeframer.java#L292) only reports gRPC body bytes as the wire size, ignoring gRPC message header bytes. Separating out the gzip overhead makes it possible to continue doing this - otherwise we will sometimes get gzip bytes included in the counts and sometimes not, depending on the message deframer state when the gzip metadata gets processed.

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ericgribkoff: Done (`|| state != HEADER`), also added test case to cover this.

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ericgribkoff: Done

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ericgribkoff: Yes, this isn't necessary. But not having to do that extra read to detect stalled makes it simpler to handle end-of-stream in message deframer. Will check on the necessary changes in message deframer (if we dropped this case here) and update this comment.

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
```diff
@@ -215,59 +215,38 @@ public void close() {
...
-    int bytesToInflate;
-    if (inflaterOutput == null) {
-      bytesToInflate = Math.min(n, MAX_OUTPUT_BUFFER_SIZE);
-      inflaterOutput = new byte[bytesToInflate];
-    } else {
-      // Empty buffer allocated on previous call
-      bytesToInflate = Math.min(n, inflaterOutput.length);
-    }
+  /**
+   * Attempts to inflate {@code length} bytes of data into {@code bufferToWrite}.
```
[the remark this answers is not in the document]
ericgribkoff: Done

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ericgribkoff: Done

core/src/main/java/io/grpc/internal/GzipInflatingBuffer.java
ericgribkoff: Done

PR #3401, opened 2017-08-25
Add support for Conscrypt with Netty

ejona86:
The commits will be kept separate when merging.

~~Since Conscrypt has had some breaking changes, this should probably wait until Conscrypt 1.0 is released to avoid breakage. That implies there will be some changes still to this PR before it is merged, but they can't be done at this time and are relatively small in scope.~~

### DIFF
--- a/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
+++ b/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
@@ -20,6 +20,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import io.grpc.ExperimentalApi;
+import io.grpc.internal.MoreThrowables;
 import io.netty.handler.codec.http2.Http2SecurityUtil;
 import io.netty.handler.ssl.ApplicationProtocolConfig;
 import io.netty.handler.ssl.ApplicationProtocolConfig.Protocol;
@@ -30,9 +31,15 @@ import io.netty.handler.ssl.SslContextBuilder;
 import io.netty.handler.ssl.SslProvider;
 import io.netty.handler.ssl.SupportedCipherSuiteFilter;
 import java.io.File;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.security.Provider;
+import java.security.Security;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Utility for configuring SslContext for gRPC.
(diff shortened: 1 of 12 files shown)


netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
```diff
@@ -84,6 +93,22 @@ private GrpcSslContexts() {}
       SelectedListenerFailureBehavior.ACCEPT,
       NEXT_PROTOCOL_VERSIONS);
 
+  private static final String SUN_PROVIDER_NAME = "SunJSSE";
+  private static final Method IS_CONSCRYPT_PROVIDER;
+
+  static {
```
ericgribkoff: Should this be wrapped in a function to provide an easier to understand stack trace if this ends up throwing?

netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
```diff
@@ -131,49 +156,115 @@ public static SslContextBuilder configure(SslContextBuilder builder) {
...
+    if (OpenSsl.isAvailable()) {
+      logger.log(Level.FINE, "Selecting OPENSSL");
+      return SslProvider.OPENSSL;
+    }
+    Provider provider = findJdkProvider();
+    if (provider != null) {
+      logger.log(Level.FINE, "Selecting JDK with provider {0}", provider);
+      return SslProvider.JDK;
+    }
+    logger.log(Level.INFO, "netty-tcnative unavailable (this may be normal)",
```
ericgribkoff: Do we expect the `IllegalStateException` to be caught somewhere? It looks like in the current flow this propagates back through the NettyChannelBuilder, in which case it seems a little confusing to say "(this may be normal)" for all of these when we're about to fail. I think the intent is that it might be normal for two out of the three to be missing, but the current message seems to imply it's ok for all of them to be missing. Maybe add an initial "One of the following TLS ALPN providers must be available:" log statement and drop the "(this may be normal)" text?

netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
ejona86: I could, although I find it more confusing to have static methods that are called _before_ static initialization completes. It's necessary and helpful at times, but when unnecessary I try to avoid it as I find it to be bug-prone.

netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
ejona86: Spoke offline. The main reason I didn't do that is because then the error is reported in two places: the log message and the exception. Right now the log messages are just informational and it's only after you get the exception that you can look at them and think "these might be useful."

ejona86:
OkHttp+Conscrypt+Windows is flaky. I've reverted the change to use Conscrypt in interop-testing's unit tests and the test client; we continue to use Jetty ALPN there for the moment.

PR #3403, opened 2017-08-26
Part 2 - Client-side full-stream decompression

ericgribkoff:
This builds on https://github.com/grpc/grpc-java/pull/3395 and adds client-side support for full-stream decompression.

This is successfully interoperating with the C++ server sending gzipped streams. The next step on the Java side is to improve the test coverage for full-stream compression by adding (or mocking) a server that sends gzipped streams. This will enable adding something along the lines of `FullStreamCompressionTest extends AbstractInteropTest`, similar to our existing `TransportCompressionTest`, `AutoWindowSizingOnTest`, etc. I already have such a server working and these tests passing locally with the client-side code included here.

@ejona86 This PR doesn't include a hard-coded 'don't do full-stream compression' flag yet. It's easier to test this and future work with compression enabled by default, but a circuit-breaker can be added once this gets closer to merging to make sure the feature is rolled out properly.

### DIFF
--- a/core/src/main/java/io/grpc/internal/MessageDeframer.java
+++ b/core/src/main/java/io/grpc/internal/MessageDeframer.java
@@ -16,10 +16,11 @@
 
 package io.grpc.internal;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
 import io.grpc.Codec;
 import io.grpc.Decompressor;
 import io.grpc.Status;
@@ -27,6 +28,7 @@ import java.io.Closeable;
 import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.zip.DataFormatException;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 
@@ -41,6 +43,7 @@ public class MessageDeframer implements Closeable, Deframer {
   private static final int HEADER_LENGTH = 5;
   private static final int COMPRESSED_FLAG_MASK = 1;
   private static final int RESERVED_MASK = 0xFE;
+  private static final int MAX_BUFFER_SIZE = 1024 * 1024 * 2;
 
   /**
    * A listener of deframing events. These methods will be invoked from the deframing thread.
@@ -82,11 +85,14 @@ public class MessageDeframer implements Closeable, Deframer {
     HEADER, BODY
   }
 
-  private final Listener listener;
+  private Listener listener;
   private int maxInboundMessageSize;
   private final StatsTraceContext statsTraceCtx;
   private final String debugString;
   private Decompressor decompressor;
+  private GzipInflatingBuffer fullStreamDecompressor;
+  private byte[] inflatedBuffer;
+  private int inflatedIndex;
   private State state = State.HEADER;
   private int requiredLength = HEADER_LENGTH;
   private boolean compressedFlag;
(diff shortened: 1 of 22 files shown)


ericgribkoff:
@ejona86  the included tests here bump coverage up by 0.9%. I think this recovers from the reduction of 0.7% in the earlier client-thread deframing PR, https://github.com/grpc/grpc-java/pull/3145

core/src/test/java/io/grpc/internal/MessageDeframerTest.java
```diff
@@ -69,6 +89,27 @@
...
 
+  @Before
+  public void setUp() {
+    if (useGzipInflatingBuffer) {
+      deframer.setFullStreamDecompressor(new GzipInflatingBuffer() {
+        @Override
+        public void addGzippedBytes(ReadableBuffer buffer) {
+          try {
+            ByteArrayOutputStream gzippedOutputStream = new ByteArrayOutputStream();
+            OutputStream gzipCompressingStream = new GZIPOutputStream(
```
ejona86: unwrap line?

core/src/test/java/io/grpc/internal/MessageDeframerTest.java
```diff
@@ -139,7 +196,17 @@ public void payloadSplitBetweenBuffers() {
         Bytes.asList(new byte[] {3, 14, 1, 5, 9, 2, 6}), bytes(producer.getValue().next()));
     verify(listener, atLeastOnce()).bytesRead(anyInt());
     verifyNoMoreInteractions(listener);
-    checkStats(1, 7, 7);
+
+    if (useGzipInflatingBuffer) {
+      checkStats(
```
ejona86: @zhangkun83, can we remove these checks from all but a couple of tests? This test for instance really has nothing to do with stats?

zhangkun83: It can be argued either way. It's not unreasonable to say that for each case you want to test that the deframer produces the correct output, you would also want to test that it records the correct measurements.

If hard-coding GZIP-derived numbers is the issue, can we try running the input through GZIP and get the actual wire size on the fly?

ericgribkoff: Changed the stats reporting for compressed-streams to exclude gzip metadata (per offline discussion with @zhangkun83). Only the `payloadSplitBetweenBuffers` test still reports stats differently with full-stream compression turned on, as the deflate block adds 2 bytes of overhead for compressing the second buffer.

core/src/main/java/io/grpc/internal/MessageDeframer.java
```diff
@@ -311,11 +311,15 @@ private boolean readRequiredBytes() {
       while ((missingBytes = requiredLength - nextFrame.readableBytes()) > 0) {
         if (fullStreamDecompressor != null) {
           try {
-            totalBytesRead += fullStreamDecompressor.inflateBytes(missingBytes, nextFrame);
-            if (missingBytes == requiredLength - nextFrame.readableBytes()) {
+            byte[] b = new byte[Math.min(missingBytes, MAX_BUFFER_SIZE)];
```
ejona86: This may be woefully underutilized. Consider receiving 16K DATA frames for a 1 MB uncompressable message. Since this method is called every time deframe() is called, this will allocate 1 MB + (1 MB - 16 KB) + (1 MB - 32 KB) + ... + 16 KB. I'd suggest continuing to use the unused portion on subsequent reads.

ericgribkoff: Fixed.

zhangkun83:
@ericgribkoff you need to resolve the conflict with #3461.

ericgribkoff:
Rebased onto master and resolved conflicts with #3461 in `MessageDeframerTest`. 

I also had to add one additional bookkeeping value to `MessageDeframer` to record `inboundBodyWireSize`. #3461 assumed that inbound wire size = Message-Length from the gRPC header and just reported the header field's value (`requiredLength`) as the inbound size to `StreamTracer#inboundMessageRead`, but this equality not necessarily hold for compressed streams.

PR #3404, opened 2017-08-28
Move jmh benchmarks to their respective modules

ejona86:
The benchmarks should be close to the code they're benchmarking, like
we do with tests.

This includes a bugfix to SerializingExecutorBenchmark to let it run.

The io.grpc.benchmarks.netty benchmarks in benchmarks/ depend on
ByteBufOutputMarshaller from benchmarks's main, so they were not moved.


ejona86:
Merged as 6164b7b2e

PR #3416, opened 2017-08-29
Bump version to 1.6.0, then 1.6.1-SNAPSHOT

zpencer:


carl-mastrangelo:
There should be two commits in this PR, one bumping to 1.6.0 and another starting 1.6.1-SNAPSHOT

zpencer:
Added commit for 1.6.1-SNAPSHOT

PR #3417, opened 2017-08-30
core: save alloc in SerializingExecutor

carl-mastrangelo:
Atomic booleans cost a few extra bytes.  Save the object overhead and use an atomic field updater.

### DIFF
--- a/core/src/main/java/io/grpc/internal/SerializingExecutor.java
+++ b/core/src/main/java/io/grpc/internal/SerializingExecutor.java
@@ -22,7 +22,7 @@ import com.google.common.base.Preconditions;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/grpc/internal/SerializingExecutor.java
```diff
@@ -87,7 +92,7 @@ private void schedule(@Nullable Runnable removable) {
             // to execute don't succeed and accidentally run a previous runnable.
             runQueue.remove(removable);
           }
-          running.set(false);
+          runState = STOPPED;
```
ejona86: You must use `runStateUpdater` for mutations. From the `AtomicIntegerFieldUpdater` documentation:
> it can guarantee atomicity only with respect to other invocations of compareAndSet and set on the same updater

carl-mastrangelo:
TIL

core/src/main/java/io/grpc/internal/SerializingExecutor.java
dapengzhang0: So is it error prone? Future coder may assign values on it directly without noticing the existence of the updater.

carl-mastrangelo: Yeah, it will be error prone.  I suggested this as an ErrorProne warning.

PR #3419, opened 2017-08-30
core: {Server,Client}CallImpl should close InputStream from marshaller

### LINKED ISSUE #3318: InputStream from MethodDescriptor.Marshaller does not close
1.5.0 Release

java version "1.8.0"
Java(TM) SE Runtime Environment (build 1.8.0-b132)
Java HotSpot(TM) 64-Bit Server VM (build 25.0-b70, mixed mode)

### What the problem?
I have a gRPC method with custom marshaller, that deserialize protobuf request from off-heap ByteBuf.

```java
private static final MethodDescriptor<ByteBuf, ShardCommandResponse> BINARY_METHOD_BULK_SHARD_COMMAND = 
            MethodDescriptor.<ByteBuf, ShardCommandResponse>newBuilder()
                    .setType(MethodDescriptor.MethodType.UNARY)
                    .setFullMethodName(StockpileServiceGrpc.METHOD_BULK_SHARD_COMMAND.getFullMethodName())
                    .setRequestMarshaller(new ByteBufMarshaller())
                    .setResponseMarshaller(ProtoUtils.marshaller(ShardCommandResponse.getDefaultInstance()))
                    .build();
```

Marshaller looks like 
```java
import java.io.InputStream;

import javax.annotation.ParametersAreNonnullByDefault;
import javax.annotation.WillNotClose;

import io.grpc.MethodDescriptor;
import io.netty.buffer.ByteBuf;
import io.netty.buffer.ByteBufInputStream;

@ParametersAreNonnullByDefault
public class ByteBufMarshaller implements MethodDescriptor.Marshaller<ByteBuf> {

    @Override
    public InputStream stream(@WillNotClose ByteBuf value) {
        ByteBuf buffer = value.retain();
        // todo wrap to {@link io.grpc.KnownLength} for improve transport
        InputStream delegate = new ByteBufInputStream(buffer, true);
        return delegate;
    }

    @Override
    @WillNotClose
    public ByteBuf parse(InputStream stream) { 
        throw new UnsupportedOperationException("Not implemented yet!");
    }
}
```
Use netty ByteBuf as request required release resources. And I expected that gRPC close `InputStream` after send request, but it's doesn't happens. As a result of memory leak my service
that use gRPC client failed with OOM.

```
[Time-limited test] ERROR io.netty.util.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. See http://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 4
#4:
	io.netty.buffer.AdvancedLeakAwareByteBuf.getBytes(AdvancedLeakAwareByteBuf.java:246)
	io.netty.buffer.CompositeByteBuf.getBytes(CompositeByteBuf.java:861)
	io.netty.buffer.CompositeByteBuf.getBytes(CompositeByteBuf.java:44)
	io.netty.buffer.ReadOnlyByteBuf.getBytes(ReadOnlyByteBuf.java:260)
	io.netty.buffer.AbstractByteBuf.readBytes(AbstractByteBuf.java:851)
	io.netty.buffer.ByteBufInputStream.read(ByteBufInputStream.java:179)
	java.io.InputStream.read(InputStream.java:101)
	com.google.common.io.ByteStreams.copy(ByteStreams.java:110)
	io.grpc.internal.MessageFramer.writeToOutputStream(MessageFramer.java:254)
	io.grpc.internal.MessageFramer.writeUncompressed(MessageFramer.java:155)
	io.grpc.internal.MessageFramer.writePayload(MessageFramer.java:127)
	io.grpc.internal.AbstractStream.writeMessage(AbstractStream.java:52)
	io.grpc.internal.ClientCallImpl.sendMessage(ClientCallImpl.java:386)
	io.grpc.stub.ClientCalls.asyncUnaryRequestCall(ClientCalls.java:239)
	io.grpc.stub.ClientCalls.asyncUnaryRequestCall(ClientCalls.java:222)
	io.grpc.stub.ClientCalls.asyncUnaryCall(ClientCalls.java:62)
```

I also recheck it via implement empty InputStream that log close method. And it method was not execute. 

I think for resolve this issue need resolve this todo https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/ClientCallImpl.java#L387

zpencer:
Fixes #3318

### DIFF
--- a/core/src/main/java/io/grpc/MethodDescriptor.java
+++ b/core/src/main/java/io/grpc/MethodDescriptor.java
@@ -267,8 +267,8 @@ public final class MethodDescriptor<ReqT, RespT> {
   }
 
   /**
-   * Convert a request message to an {@link InputStream}. gRPC will close the InputStream
-   * when it is done reading.
+   * Convert a request message to an {@link InputStream}.
+   * The returned InputStream should be closed by the caller.
    *
    * @param requestMessage to serialize using the request {@link Marshaller}.
    * @return serialized request message.
@@ -290,8 +290,8 @@ public final class MethodDescriptor<ReqT, RespT> {
   }
 
   /**
-   * Serialize an outgoing response message. gRPC will close the InputStream
-   * when it is done reading.
+   * Serialize an outgoing response message.
+   * The returned InputStream should be closed by the caller.
    *
    * @param response the response message to serialize.
    * @return the serialized message as a byte stream.
--- a/core/src/main/java/io/grpc/MethodDescriptor.java
+++ b/core/src/main/java/io/grpc/MethodDescriptor.java
@@ -267,7 +267,8 @@ public final class MethodDescriptor<ReqT, RespT> {
   }
 
   /**
-   * Convert a request message to an {@link InputStream}.
+   * Convert a request message to an {@link InputStream}. When gRPC will close the InputStream
+   * when it is done reading.
    *
    * @param requestMessage to serialize using the request {@link Marshaller}.
    * @return serialized request message.
@@ -289,7 +290,8 @@ public final class MethodDescriptor<ReqT, RespT> {
   }
 
   /**
-   * Serialize an outgoing response message.
+   * Serialize an outgoing response message. When gRPC will close the InputStream
+   * when it is done reading.
    *
    * @param response the response message to serialize.
    * @return the serialized message as a byte stream.
--- a/core/src/main/java/io/grpc/MethodDescriptor.java
+++ b/core/src/main/java/io/grpc/MethodDescriptor.java
@@ -267,7 +267,7 @@ public final class MethodDescriptor<ReqT, RespT> {
   }
 
   /**
-   * Convert a request message to an {@link InputStream}. When gRPC will close the InputStream
+   * Convert a request message to an {@link InputStream}. gRPC will close the InputStream
    * when it is done reading.
    *
    * @param requestMessage to serialize using the request {@link Marshaller}.
@@ -290,7 +290,7 @@ public final class MethodDescriptor<ReqT, RespT> {
   }
 
   /**
-   * Serialize an outgoing response message. When gRPC will close the InputStream
+   * Serialize an outgoing response message. gRPC will close the InputStream
    * when it is done reading.
    *
    * @param response the response message to serialize.
--- a/core/src/main/java/io/grpc/internal/ClientCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ClientCallImpl.java
@@ -425,7 +425,6 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
     checkState(stream != null, "Not started");
     checkState(!cancelCalled, "call was cancelled");
     checkState(!halfCloseCalled, "call was half-closed");
-    InputStream messageIs = null;
     try {
       if (stream instanceof RetriableStream) {
         @SuppressWarnings("unchecked")
@@ -442,8 +441,6 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
     } catch (Error e) {
       stream.cancel(Status.CANCELLED.withDescription("Client sendMessage() failed with Error"));
       throw e;
-    } finally {
-      GrpcUtil.closeQuietly(messageIs);
     }
     // For unary requests, we don't flush since we know that halfClose should be coming soon. This
     // allows us to piggy-back the END_STREAM=true on the last message frame without opening the
--- a/core/src/main/java/io/grpc/internal/ClientCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ClientCallImpl.java
@@ -30,6 +30,8 @@ import static io.grpc.internal.GrpcUtil.TIMEOUT_KEY;
 import static java.lang.Math.max;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import com.google.common.io.Closeables;
 import io.grpc.Attributes;
 import io.grpc.CallOptions;
 import io.grpc.ClientCall;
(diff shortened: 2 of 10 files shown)


core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -383,13 +383,17 @@ public void sendMessage(ReqT message) {
     Preconditions.checkState(stream != null, "Not started");
     Preconditions.checkState(!cancelCalled, "call was cancelled");
     Preconditions.checkState(!halfCloseCalled, "call was half-closed");
+    InputStream messageIs = null;
     try {
-      // TODO(notcarl): Find out if messageIs needs to be closed.
-      InputStream messageIs = method.streamRequest(message);
+      messageIs = method.streamRequest(message);
```
carl-mastrangelo: A minor note: if the consumer retained a ref to this IS we might accidentally close it prematurely.  For example, they may push it to some queue for asynchronous processing.   

I don't *think* thats a problem, but it is an API change.   This probably warrants a javadoc comment in the MethodDecriptor.

core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -259,7 +259,8 @@ public RespT parseResponse(InputStream input) {
   }
 
   /**
-   * Convert a request message to an {@link InputStream}.
+   * Convert a request message to an {@link InputStream}. When gRPC will close the InputStream
```
ejona86: Remove the leading "When". Ditto below.

zpencer:
@carl-mastrangelo PTAL

core/src/main/java/io/grpc/internal/ClientCallImpl.java
zpencer: Done

core/src/main/java/io/grpc/internal/ServerCallImpl.java
```diff
@@ -139,6 +140,8 @@ public void sendMessage(RespT message) {
     } catch (Throwable t) {
       close(Status.fromThrowable(t), new Metadata());
       throw new RuntimeException(t);
+    } finally {
+      GrpcUtil.closeQuietly(resp);
```
carl-mastrangelo: if resp != null

core/src/test/java/io/grpc/internal/ClientCallImplTest.java
```diff
@@ -878,6 +880,34 @@ public void getAttributes() {
...
+      public InputStream stream(Void value) {
+        return mockInputStream;
+      }
+
+      @Override
+      public Void parse(InputStream stream) {
+        throw new UnsupportedOperationException();
+      }
+    };
+    MethodDescriptor<Void, Void> method = MethodDescriptor.<Void, Void>newBuilder()
```
carl-mastrangelo: TestMethodDescriptors.voidDescriptor()

zpencer: In this case I  need to wire up the mockInputStream to the methodDescriptor, so I can't use the canned MethodDescriptor.

carl-mastrangelo: TestMethodDescriptors.voidDescriptor().toBuilder().setRequestMarshaller().build()

core/src/main/java/io/grpc/internal/ServerCallImpl.java
zpencer: Guava's `Closeables` allows the input stream to be null. Moving the check into `GrpcUtil`

zpencer:
retest this please

core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -267,7 +267,8 @@ public RespT parseResponse(InputStream input) {
   }
 
   /**
-   * Convert a request message to an {@link InputStream}.
+   * Convert a request message to an {@link InputStream}. gRPC will close the InputStream
```
ejona86: This seems over-specifying, as that isn't true if someone else calls the method. You could say "The returned InputStream should be closed by the caller."

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -441,6 +442,8 @@ public void sendMessage(ReqT message) {
     } catch (Error e) {
       stream.cancel(Status.CANCELLED.withDescription("Client sendMessage() failed with Error"));
       throw e;
+    } finally {
+      GrpcUtil.closeQuietly(messageIs);
```
ejona86: So this is great, except it breaks DelayedStream. https://github.com/grpc/grpc-java/blob/5ac9ad0/core/src/main/java/io/grpc/internal/DelayedStream.java#L215

I see two options:
1. Have the stream take ownership of the InputStream. That's not really that hard, as it'd just be a message.close() in AbstractStream, DelayedStream, and InProcessStream.
2. Have the stream call `method.streamRequest()`. That would also remove the special-casing of RetriableStream.

zpencer: InProcessTransport already has a pattern of passing ownership of InputStream:
1. https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/inprocess/InProcessTransport.java#L378
1. https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/inprocess/InProcessTransport.java#L561

So option 1 would be more consistent with what we already do today: `AbstractStream` should close, `DelayedStream` already passes on ownership, and InProcessStream already wraps the InputStream in a `SingleMessageProducer` which passes ownership to the `StreamListener` which calls close.

core/src/main/java/io/grpc/MethodDescriptor.java
zpencer: fixed

PR #3424, opened 2017-08-30
core: use Closeables.closeQuietly from guava

zpencer:
Essentially does the same thing but also logs at WARNING.


dapengzhang0:
FYI `Closeables` is `@Beta`

zpencer:
In that case we can't use Closeables, closing this PR

PR #3429, opened 2017-09-01
core,okhttp: use Executor interface rather than SerializingExecutor

carl-mastrangelo:


ejona86:
> Wouldn't we get incorrect behavior if we use any Executor that is not SerializingExecutor though?

+1

ejona86:
> All the usages of instances are within the same class, so it's unlikely that the mistake will be made.

That's not true for AsyncFrameWriter, and it's hard to see for OkHttpClientTransport.

> Also, using the top level interface allows returning a different type. If you look at new SerializingExecutor(MoreExecutors.directExecutor());, it isnt using the SerializeReentrantCallsDirectExecutor.

There is no need for an interface there, just change the type. Done. In _all_ the cases of `new SerializingExecutor(MoreExecutors.directExecutor());` that we see here, the concrete type of Executor will always be the same.

PR #3430, opened 2017-09-01
core: name more anonymous classes in ServerImpl

carl-mastrangelo:
This makes it much easier to read stack traces since the name of the
event is in the frame.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ServerImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerImpl.java
@@ -302,7 +302,7 @@ public final class ServerImpl extends io.grpc.Server implements WithLogId {
     }
   }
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -612,17 +636,28 @@ public void closed(final Status status) {
...
-      });
+      }
+
+      callExecutor.execute(new Closed());
     }
 
     @Override
     public void onReady() {
-      callExecutor.execute(new ContextRunnable(context) {
+      final class OnReady extends ContextRunnable {
```
dapengzhang0: nit: `final` might not be needed inside a method

PR #3431, opened 2017-09-02
core: refactor transportProvider creating fewer transport instances

dapengzhang0:
Before the change:
The `ClientTransportProvider` or `GrpcUtil.getTransportFromPickResult()` produces a new wrapped transport instance for every single call. The wrapped transport instance only overrides the `newStream()` method by injecting a `streamTracerFactory` in the `callOptions`.

After the change:
Refactored `ClientTransportProvider` and `GrpcUtil.getTransportFromPickResult()` so that it does not wrap the underlying transport, and still injects `streamTracerFactory` in the `callOptions` when calling `newStream()`.

This refactoring is likely to make retry and resolving isssue #2562 easier.


dapengzhang0:
close it for the moment

PR #3433, opened 2017-09-02
protobuf-nano: avoid using @Beta API com.google.common.io.ByteStreams

zpencer:


protobuf-nano/src/main/java/io/grpc/protobuf/nano/NanoUtils.java
```diff
@@ -57,4 +62,28 @@ public T parse(InputStream stream) {
       }
     };
   }
+
+  // Copied from guava com.google.common.io.ByteStreams because its API is unstable (beta)
+  private static byte[] toByteArray(InputStream in) throws IOException {
```
ejona86: IOUtils.toByteArray()

Issue #3435, opened 2017-09-05
grpc multiple instances of same service different ports within same jvm ubuntu


sampleref:
### What did you do?
I have a service **EventServer** with one method **sendEvent**.  I have created a unit test client for **EventServerImpl1**

My use case is there are 2 instances of **EventServer** running on different ports for this test case where one propagates the event to other and response vice versa

Example: **EventTestClient** invokes **sendEvent** to **EventServerImpl1** and in **sendEvent** method of **EventServerImpl1** it propagates to **sendEvent** method of **EventServerImpl2**. 

Here 2 grpc servers **EventServerImpl1**(port **17123**) and **EventServerImpl2**(port **17124**) implements **EventServer** within **same JVM** for this. 

Now in 

**EventTestClient**: I have code as 

eventChannel = InProcessChannelBuilder.forAddress("localhost", 17123).directExecutor()
                .usePlaintext(true).build();
stub = EventsServerGrpc.newBlockingStub(eventChannel );
stub.sendEvent(request); **//This is fine**

**EventServerImpl1**:

eventChannel2 = ManagedChannelBuilder.forAddress("localhost", 17124).usePlaintext(true).build();
stub2 = EventsServerGrpc.newBlockingStub(eventChannel2 );
stub2.sendEvent(request); **//This is hung**

Both are blocking stubs only

### What did you expect to see?
I expect to see the request propagates to **EventServerImpl2** as **EventServerImpl1** being the **client** to it

### What did you see instead?
When **EventServerImpl1.sendEvent** is triggered( **stub2.sendEvent(request)** ) the process **hangs forever**. 

When I **directly query EventServerImpl2(port 17124)** from other java process during this hang over period it works fine. But within that same process its hung forever.

Again on Windows this is fine and test case runs well. But only on Ubuntu I see this issue

Please suggest if there is something I am missing here

carl-mastrangelo:
this is a bug on our part.  `InProcessChannelBuilder.forAddress` doesn't do what it purports to do.  It actually calls the static factory method on the super class, which creates a live channel.  In the mean time please use `InProcessChannelBuilder.forName`

sampleref:
[...]

PR #3438, opened 2017-09-06
inprocess,core: add ManagedChannelBuilder and ServerBuilder factory h…

carl-mastrangelo:
…iders

Because the factory for Channels and Servers resides on the builder
itself, it is easy for subclasses to accidentally inherit the
factory.  This causes confusion, because calling a static method on
a specific class may result in a different class.

This change adds hiding static factories to each builder, and a test
to enforce that each subclass hides the factory.  The test lives in
the interop tests, because it has a classpath dependency on all the
existing transports.

Minor note: the test scans the classpath using a Beta Guava API.
The test can be disabled if the API goes away.

### DIFF
--- a/core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
@@ -49,23 +49,17 @@ public final class InProcessChannelBuilder extends
   }
 
   /**
-   * Create a channel builder that will connect to the server with the given name.
-   *
-   * @since 1.7.0
-   * @param target the name of the in process server to connect to.
+   * Always fails.  Call {@link #forName} instead.
    */
   public static InProcessChannelBuilder forTarget(String target) {
-    return forName(target);
+    throw new UnsupportedOperationException("call forName() instead");
   }
 
   /**
-   * Create a channel builder that will connect to the server with the given name and port,
-   * converted into an authority.
-   *
-   * @since 1.7.0
+   * Always fails.  Call {@link #forName} instead.
    */
   public static InProcessChannelBuilder forAddress(String name, int port) {
-    return forTarget(GrpcUtil.authorityFromHostAndPort(name, port));
+    throw new UnsupportedOperationException("call forName() instead");
   }
 
   private final String name;
--- a/core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
@@ -48,6 +48,26 @@ public final class InProcessChannelBuilder extends
     return new InProcessChannelBuilder(name);
   }
 
+  /**
+   * Create a channel builder that will connect to the server with the given name.
+   *
+   * @since 1.7.0
+   * @param target the name of the in process server to connect to.
+   */
+  public static InProcessChannelBuilder forTarget(String target) {
+    return forName(target);
+  }
+
+  /**
+   * Create a channel builder that will connect to the server with the given name and port,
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
```diff
@@ -48,6 +48,26 @@ public static InProcessChannelBuilder forName(String name) {
...
   }
 
+  /**
+   * Create a channel builder that will connect to the server with the given name.
+   *
+   * @since 1.7.0
+   * @param target the name of the in process server to connect to.
+   */
+  public static InProcessChannelBuilder forTarget(String target) {
+    return forName(target);
```
ejona86: These should throw. It doesn't provide the target semantics: the name resolver would be ignored.

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
```diff
@@ -68,6 +68,17 @@ public static InProcessServerBuilder forName(String name) {
...
 
+  /**
+   * Create a server builder that will bind with the given port, converted to a string.
+   *
+   * @since 1.7.0
+   * @param port the identity of the server for clients to connect to
+   * @return a new builder
+   */
+  public static InProcessServerBuilder forPort(int port) {
+    return new InProcessServerBuilder(String.valueOf(port));
```
ejona86: throw.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -56,6 +56,14 @@
         <T extends AbstractManagedChannelImplBuilder<T>> extends ManagedChannelBuilder<T> {
   private static final String DIRECT_ADDRESS_SCHEME = "directaddress";
 
+  public static ManagedChannelBuilder<?> forAddress(String name, int port) {
+    throw new UnsupportedOperationException("Subclass failed to hide static factory");
```
ejona86: Why don't we just make this throw in a normal "unsupported" message and let subclasses override when it is supported?

carl-mastrangelo:
Updates #3435

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
carl-mastrangelo: Not sure what you mean, can you clarify?

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
ejona86: Don't blame the subclasses with "Subclass failed to hide static factory". Basically, have a message that is "okay" here and let subclasses override when they feel like it.

Although that would remove the need for the test (although a test like that could have found `ForwardingChannelBuilder` which doesn't extend from this class).

carl-mastrangelo: Since this is an internal class, I think it would be an error.  All implementations are owned by us, and we do want to override.

Issue #3293, opened 2017-07-31
Go to "permanent error mode" if ChannelExecutor throws exception


dapengzhang0:
As mentioned in discussion in #3288, it's questionable only catching instead of throwing exceptions in `ChannelExecutor`. Should we throw regardless or should we only log in production and throw in test evn?

dapengzhang0:
@zhangkun83 , @ejona86

ejona86:
I do think it tends to make sense to catch and log the exception, since propagating the exception will probably only cause additional failures and we know the current stack does not depend on the result of the call (since the execution is not guaranteed to be complete when drain returns.)

I don't think we should throw in a test environment; that's likely to produce confusion. Instead, you can inject a "exception handler" that is passed the exception. Normally, it would simply log. In tests it could save the exception and we can add an `@After` to assert that no exception was thrown. None of the tests should trigger the exception code path (except maybe a test that is testing the exception code path :) ).

zhangkun83:
> Instead, you can inject a "exception handler" that is passed the exception

Not sure if you meant injecting the exception handler to `ChannelExecutor`, but I am just thinking about that. `ManagedChannelImpl`, when capturing a throwable, may switch to an "permanent error mode" which would fail subsequent RPCs, which propagate the error very loudly.

carl-mastrangelo:
FYI, I have been looking at consolidateing our executors (we have many) including the channel exec.  If you make a change before I send out mine, please cc me.

ejona86:
I'm fine with going into a "permanent error mode" if we get an exception with ChannelExecutor. Although the implementation will need to be very simple.

ejona86:
@zhangkun83, assigning this to you since it seemed you were really interested in getting this done; feel free to unassign/reassign.

ejona86:
#4152 disabled the added panic mode. It doesn't delete it, but leaves it non-functional.

We discovered pre-existing tests that had failures cause by exitIdleMode scheduling a timer, which can trigger Thread creation. The timer is to schedule IDLE_TIMEOUT, which is enabled by default to 30 minutes. In certain test environments, Thread creation is black-listed and throws an exception. Previously the exception was caught and logged, but otherwise things worked fine for the test; the test would not last long enough such that the broken IDLE_TIMEOUT would matter.

[...]

PR #3439, opened 2017-09-06
grpclb: fallback to backend addresses

zhangkun83:
The GRPCLB client will use the backend addresses from resolver if it has not received any server list from any balancer after a certain timeout (10s). 

/cc @markdroth


markdroth:
Looks like we've decided that the timeout default should be 10 seconds.

grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
```diff
@@ -217,22 +258,92 @@ GrpclbClientLoadRecorder getLoadRecorder() {
...
+          @Override
+          public void run() {
+            stream.loadReportTask = null;
+            stream.sendLoadReport();
+          }
+        });
+    }
+  }
+
+  class LbStream implements StreamObserver<LoadBalanceResponse> {
```
dapengzhang0: Why is this becoming package private?

grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
```diff
@@ -245,6 +356,7 @@ public void run() {
       // stream.
       loadRecorder = new GrpclbClientLoadRecorder(time);
       lbRequestWriter = stub.withWaitForReady().balanceLoad(this);
```
dapengzhang0: Is this safe to call `balanceLoad(this)` before the constructor is finished? Is it better to do RPC call after `LbStream` is `new()`ed?

grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
```diff
@@ -217,22 +258,92 @@ GrpclbClientLoadRecorder getLoadRecorder() {
...
+          public void run() {
+            fallbackTimerExpired = true;
+            maybeUseFallbackBackends();
+          }
+        });
+    }
+  }
+
+  @VisibleForTesting
+  class LoadReportingTask implements Runnable {
```
dapengzhang0: This class can be inner class of LbStream, and in test, still accessible by using `command.getClass().getName()` instead of `command instanceof`

zhangkun83: I would prefer a direct reference to the class rather than its name. It's more robust and easier to navigate in IDEs.

grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
zhangkun83: Oops. Reverted.

grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
zhangkun83: It's totally safe. Normally `balancerLoad(this)` should not throw, unless there is a bug. If that happens, the exception will be caught by `ChannelExecutor`, and will hopefully be propagated in a more useful way once #3293 is resolved.

dapengzhang0: My point is the RPC may call methods of `this`, while `this` is not fully initialize, for example: `loadReportingTask` is null at this point, so `this.onNext()` may throw NPE if the RPC gets a response before a request being sent. So moving the line `loadReportingTask = new LoadReportingTask(this)` above the RPC can fix it, but this fix is based on the assumption you know how RPC may misbehave.

zhangkun83: I see. I have moved the lines that reference into `start()`, and create `LoadReportingTask` every time it's scheduled instead of re-using one.

grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
```diff
@@ -535,10 +657,18 @@ public boolean equals(Object other) {
       this.token = checkNotNull(token, "token");
     }
 
+    BackendEntry(Subchannel subchannel) {
```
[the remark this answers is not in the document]
zhangkun83: Done.

PR #3441, opened 2017-09-07
core: consolidate termination state in ManagedChannelImpl

carl-mastrangelo:


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -571,13 +590,13 @@ public String authority() {
...
   private void maybeTerminateChannel() {
-    if (terminated) {
+    if (isTerminated()) {
       return;
     }
     if (shutdown.get() && subchannels.isEmpty() && oobChannels.isEmpty()) {
       log.log(Level.FINE, "[{0}] Terminated", getLogId());
-      terminated = true;
-      terminatedLatch.countDown();
+      assert terminator.isTerminating();
```
zhangkun83: Since you have made the effort to encapsulate the relation between terminating and terminated in `Terminator`, this check, which assumes that relation, doesn't seem to belong here. I would add `goTerminating()` and `goTerminated()` in `Terminator` and do the check there.

Also, I don't find assert useful enough here. If the check fails, the channel may be terminated prematurely and may lead to user-visible issue. We want to know if this happens even in production environment where assert is usually disabled. A precondition might work better, or to be conservative, a logged warning may also do.

carl-mastrangelo: Encapsulated the logic.  

That said, I would prefer to leave the asserts as is.  They are just comments, not to be enforced.  It explains my assumptions about termination.

Issue #3273, opened 2017-07-25
Use service.load(class) instead of service.load(class, classloader) for OSGi compat


HuanLinying:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?
1.4.0

### What JVM are you using (`java -version`)?
1.8
****
### What did you do?
I'm trying to wrap GRPC jar to OSGI bundle. There is a problem of making java.util.service works in OSGI environment. The reason is GRPC use service.load api with classloader parameter, such as:

    @VisibleForTesting
    public static Iterable<ManagedChannelProvider> getCandidatesViaServiceLoader(
      ClassLoader classLoader) {
    return _ServiceLoader.load(ManagedChannelProvider.class,_ classLoader);
    }

I‘m using Apache Aries spi-fly to transferring  java.util.service to OSGI service. But unfortunately this plugin does not support the api service.load(class, classloader) now, and there is no other way to figure this issue out. Actually assigning  classloader is conflicts with OSGI in my opinion.

### What did you expect to see?
It seems no necessary to assign classloader parameter when use service.load(), could we just use service.load(class)?

ejona86:
Since #3183 (which didn't make it in 1.5), the argument is necessary, because we use a different class loader (getClass().getClassLoader) than the default (context class loader). It seems like that SPI Fly will be unnecessary in 1.6?

Are you needing SPI Fly for anything else, or only gRPC? If only for gRPC, then it seems you should just stop using SPI Fly in 1.6.

HuanLinying:
> Since #3183 (which didn't make it in 1.5), the argument is necessary, because we use a different class loader (getClass().getClassLoader) than the default (context class loader). It seems like that SPI Fly will be unnecessary in 1.6?
> 
> Are you needing SPI Fly for anything else, or only gRPC? If only for gRPC, then it seems you should just stop using SPI Fly in 1.6.

It sounds great. I'm gonna check it out and feed back soon.

ejona86:
We have no knowledge with OSGi, so it is hard to help.

> gRPC-core and gRPC-netty are separate OSGI bundles

Can you just put them in the same bundle?

[...]

PR #3453, opened 2017-09-11
core: allow to fallback to ServiceLoader.load withouth classloader to make GRCP more OSGi friendly.

pieterjanpintens:
See https://github.com/grpc/grpc-java/issues/3273 for more information

### DIFF
--- a/core/src/main/java/io/grpc/ManagedChannelProvider.java
+++ b/core/src/main/java/io/grpc/ManagedChannelProvider.java
@@ -70,7 +70,14 @@ public abstract class ManagedChannelProvider {
   @VisibleForTesting
   public static Iterable<ManagedChannelProvider> getCandidatesViaServiceLoader(
       ClassLoader classLoader) {
-    return ServiceLoader.load(ManagedChannelProvider.class, classLoader);
+    Iterable<ManagedChannelProvider> i
+        = ServiceLoader.load(ManagedChannelProvider.class, classLoader);
+    // Attempt to load using the context class loader and ServiceLoader.
+    // This allows frameworks like http://aries.apache.org/modules/spi-fly.html to plug in.
+    if (!i.iterator().hasNext()) {
+      i = ServiceLoader.load(ManagedChannelProvider.class);
+    }
+    return i;
   }
 
   /**
--- a/core/src/main/java/io/grpc/ServerProvider.java
+++ b/core/src/main/java/io/grpc/ServerProvider.java
@@ -35,6 +35,13 @@ public abstract class ServerProvider {
   @VisibleForTesting
   static final ServerProvider load(ClassLoader cl) {
     ServiceLoader<ServerProvider> providers = ServiceLoader.load(ServerProvider.class, cl);
+
+    // Attempt to load using the context class loader and ServiceLoader.
+    // This allows frameworks like http://aries.apache.org/modules/spi-fly.html to plug in.
+    if (!providers.iterator().hasNext()) {
+      providers = ServiceLoader.load(ServerProvider.class);
+    }
+
     ServerProvider best = null;
 
     for (ServerProvider current : providers) {
@@ -81,4 +88,3 @@ public abstract class ServerProvider {
    */
   protected abstract ServerBuilder<?> builderForPort(int port);
 }
-
--- a/core/src/test/java/io/grpc/NameResolverProviderTest.java
+++ b/core/src/test/java/io/grpc/NameResolverProviderTest.java
@@ -49,11 +49,17 @@ public class NameResolverProviderTest {
 
   @Test
   public void noProvider() {
-    ClassLoader cl = new ReplacingClassLoader(
-        getClass().getClassLoader(), serviceFile,
-        "io/grpc/NameResolverProviderTest-doesNotExist.txt");
-    List<NameResolverProvider> providers = NameResolverProvider.load(cl);
-    assertEquals(Collections.<NameResolverProvider>emptyList(), providers);
+    ClassLoader ccl = Thread.currentThread().getContextClassLoader();
+    try {
+      ClassLoader cl = new ReplacingClassLoader(
+          getClass().getClassLoader(), serviceFile,
+          "io/grpc/NameResolverProviderTest-doesNotExist.txt");
+      Thread.currentThread().setContextClassLoader(cl);
+      List<NameResolverProvider> providers = NameResolverProvider.load(cl);
+      assertEquals(Collections.<NameResolverProvider>emptyList(), providers);
+    } finally {
+      Thread.currentThread().setContextClassLoader(ccl);
+    }
   }
 
   @Test
@@ -143,11 +149,15 @@ public class NameResolverProviderTest {
       }
     };
     cl = new StaticTestingClassLoader(cl, Pattern.compile("io\\.grpc\\.[^.]*"));
+    ClassLoader ccl = Thread.currentThread().getContextClassLoader();
     try {
+      Thread.currentThread().setContextClassLoader(cl);
       invokeGetCandidatesViaHardCoded(cl);
       fail("Expected exception");
     } catch (RuntimeException ex) {
       assertSame(toThrow, ex);
+    } finally {
+      Thread.currentThread().setContextClassLoader(ccl);
     }
   }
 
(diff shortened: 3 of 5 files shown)


carl-mastrangelo:
@pieterjanpintens Hi, would you mind adding a simple test for the PR?  We don't often work with OSGi, so it would be good to know if we accidentally break it.

pieterjanpintens:
Yes I will add tests, I think I also have an idea why the tests for Android loader fails. I will fix that too. But that will be for tomorrow

core/src/test/java/io/grpc/NameResolverProviderTest.java
```diff
@@ -143,11 +143,15 @@ public void getCandidatesViaHardCoded_triesToLoadClasses() throws Exception {
       }
     };
     cl = new StaticTestingClassLoader(cl, Pattern.compile("io\\.grpc\\.[^.]*"));
```
pieterjanpintens: Also take the new load into account when testing

core/src/test/java/io/grpc/ServerProviderTest.java
```diff
@@ -49,6 +49,20 @@ public void unavailableProvider() {
     assertNull(ServerProvider.load(cl));
   }
```
pieterjanpintens: added extra test

pieterjanpintens:
@carl-mastrangelo added a test and fixed the failing ones of the nameserviceprovider

core/src/main/java/io/grpc/ServerProvider.java
```diff
@@ -46,6 +46,27 @@ static final ServerProvider load(ClassLoader cl) {
         best = current;
       }
     }
+    
+    if (best == null) {
+      best = loadOsgi();
```
ejona86: Instead, could you do the same `iterator().hasNext()` trick as elsewhere?

```java
ServiceLoader<ServerProvider> providers = ServiceLoader.load(ServerProvider.class, cl);
if (!providers.iterator().hasNext() && classLoader == NameResolverProvider.class.getClassLoader()) {
  providers = ServiceLoader.load(ServerProvider.class);
}
```

core/src/main/java/io/grpc/ManagedChannelProvider.java
```diff
@@ -70,7 +70,13 @@ public int compare(ManagedChannelProvider f1, ManagedChannelProvider f2) {
   @VisibleForTesting
   public static Iterable<ManagedChannelProvider> getCandidatesViaServiceLoader(
       ClassLoader classLoader) {
-    return ServiceLoader.load(ManagedChannelProvider.class, classLoader);
+    Iterable<ManagedChannelProvider> i
+        = ServiceLoader.load(ManagedChannelProvider.class, classLoader);
+    if (!i.iterator().hasNext()
+            && classLoader == ManagedChannelProvider.class.getClassLoader()) {
+      i = ServiceLoader.load(ManagedChannelProvider.class);
```
ejona86: A comment about SPI fly would be great here. Ditto elsewhere.

core/src/main/java/io/grpc/ManagedChannelProvider.java
```diff
@@ -70,7 +70,13 @@ public int compare(ManagedChannelProvider f1, ManagedChannelProvider f2) {
   @VisibleForTesting
   public static Iterable<ManagedChannelProvider> getCandidatesViaServiceLoader(
       ClassLoader classLoader) {
-    return ServiceLoader.load(ManagedChannelProvider.class, classLoader);
+    Iterable<ManagedChannelProvider> i
+        = ServiceLoader.load(ManagedChannelProvider.class, classLoader);
+    if (!i.iterator().hasNext()
+            && classLoader == ManagedChannelProvider.class.getClassLoader()) {
```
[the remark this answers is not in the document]
pieterjanpintens: Partially this also makes sure that no side effects occur if some one want to use a custom classloader. I wanted to keep the possible side effects of this patch as small as posible.

core/src/main/java/io/grpc/ServerProvider.java
pieterjanpintens: I think that should be possible, the code there is just structured differently so that would require a bit more refactoring.

core/src/main/java/io/grpc/ManagedChannelProvider.java
pieterjanpintens: OK i will add this.

core/src/main/java/io/grpc/ServerProvider.java
pieterjanpintens: Ok will do this

core/src/test/java/io/grpc/ServerProviderTest.java
pieterjanpintens: changed test to better resemble the OSGi path. The 'default' classloader will fail to work. The OSGi context classloader must take over its job.

pieterjanpintens:
Updated the pull request with according to the remarks. Improved the test a bit

core/src/test/java/io/grpc/ServerProviderTest.java
```diff
@@ -49,6 +54,39 @@ public void unavailableProvider() {
...
+        @Override
+        public int hashCode() {
+          return super.hashCode();
+        }
+      };
+
+      // test that the context classloader is used as fallback
+      TrackResourceClassLoader tracker = new TrackResourceClassLoader(ccl);
+      Thread.currentThread().setContextClassLoader(tracker);
+      ServerProvider.load(cl);
```
ejona86: What does this find? It seems there should be an assert here.

core/src/test/java/io/grpc/ServerProviderTest.java
```diff
@@ -49,6 +54,39 @@ public void unavailableProvider() {
...
+        }
+
+        @Override
+        public int hashCode() {
+          return super.hashCode();
+        }
+      };
+
+      // test that the context classloader is used as fallback
+      TrackResourceClassLoader tracker = new TrackResourceClassLoader(ccl);
```
ejona86: Instead of tracking which resources are loaded, how about using ReplacingClassLoader with ServerProviderTest-multipleProvider.txt (or similar) and asserting that the expected provider is returned?

core/src/main/java/io/grpc/ServerProvider.java
```diff
@@ -35,6 +35,13 @@
   @VisibleForTesting
   static final ServerProvider load(ClassLoader cl) {
     ServiceLoader<ServerProvider> providers = ServiceLoader.load(ServerProvider.class, cl);
+
+    // Attempt to load using the context class loader and ServiceLoader.
+    // This allows frameworks like http://aries.apache.org/modules/spi-fly.html to plug in.
+    if (!providers.iterator().hasNext() && cl.equals(ServerProvider.class.getClassLoader())) {
```
ejona86: Instead of doing the equals hack in the test, how about we just drop this part of the condition. Let's drop it in all the providers? If it makes the tests too painful then maybe we just pass in a `boolean fallbackToContextCl` for the test to simplify the behavior.

core/src/test/java/io/grpc/ServerProviderTest.java
pieterjanpintens: ok done :)

core/src/main/java/io/grpc/ServerProvider.java
pieterjanpintens: Actually i don't think it is needed for the tests at all. I will drop it.

core/src/test/java/io/grpc/ServerProviderTest.java
```diff
@@ -49,6 +49,40 @@ public void unavailableProvider() {
...
+  public void contextClassLoaderProvider() {
+    ClassLoader ccl = Thread.currentThread().getContextClassLoader();
+    try {
+      // make a classloader that equals ServerProvider.class.getClassLoader()
+      // but that wont find any provider
+      ClassLoader cl = new ReplacingClassLoader(
+              getClass().getClassLoader(), serviceFile,
+              "io/grpc/ServerProviderTest-empty.txt") {
+        @Override
+        public boolean equals(Object obj) {
```
ejona86: This looks unnecessary now.

ejona86:
@pieterjanpintens, to make the CLA checker happy, would you mind rewriting the author of the commit to be your email instead of based on the machine you committed on? (Are you a Raymond Feist fan?)

pieterjanpintens:
Sorry for the mess, working with two laptops sucks sometimes :)
I'm a fan of feist. He writes (wrote?) very good fantasy books. I find the worlds much more rich than anything I read before. I also like how everything works out, how characters come back. Also the books about the empire in which you see the other side of the 'conflict'. I think those books are my all time favourites. And the characters are so Iconic. Arutha, Roby, Puc, Thomas love them

ejona86:
@carl-mastrangelo, do things look good to you as well?

ejona86:
okay to test

ejona86:
:-/ This was opened against 1.6.x instead of master. I can change the PR to be against master instead, but it then brings in random commits from 1.6.x. @pieterjanpintens, could you rebase (and clean up with -i) on master?

pieterjanpintens:
Yeah sure, but it sleepy time here now. Will do that tomorrow

ejona86:
Closing in favor of rebased #3479

PR #3454, opened 2017-09-11
netty,core: add a TransportTracer class

zpencer:
This class will gather some basic stats on the transport.


zpencer:
I need to fix the netty FlowController windows, just realized the class is not thread safe

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,212 @@
...
+import io.grpc.StreamTracer;
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicLong;
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * A class for gathering statistics about a transport. This is an experimental feature.
+ */
+@ThreadSafe
+@ExperimentalApi
```
ejona86: This is internal, no need to label experimental.

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,212 @@
...
+public final class TransportTracer {
+  private final AtomicLong streamsStarted = new AtomicLong();
```
ejona86: Are these mutated from multiple threads? It seems like a volatile would work.

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,212 @@
...
+        streamsSucceeded.incrementAndGet();
+      } else {
+        streamsFailed.incrementAndGet();
+      }
+    }
+
+    @Override
+    public void outboundMessage() {
+      messagesSent.incrementAndGet();
+      updateMsecTimestamp(lastMessageReceivedTimeMsec);
```
ejona86: s/Received/Sent/

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: Done

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: Done

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: I originally had them as volatiles, and one of the linters flagged the increment operations as unsafe. I switched them to AtomicLongs for this reason.

zpencer:
@carl-mastrangelo PTAL

core/src/main/java/io/grpc/internal/TransportTracer.java
ejona86: You can generally `@SuppressWarnings` for false positives like that, along with a comment saying we only mutate from one thread.

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * A class for gathering statistics about a transport. This is an experimental feature.
+ */
+public final class TransportTracer {
+  // streamsStarted happens serially, so a volatile is sufficient
+  private volatile long streamsStarted;
+  private long lastStreamCreatedTimeMsec;
+  private final AtomicLong streamsSucceeded = new AtomicLong();
```
carl-mastrangelo: can you make these volatiles and use  a field updater?

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,212 @@
...
+  private final AtomicLong streamsFailed = new AtomicLong();
+  private final AtomicLong messagesSent = new AtomicLong();
+  private final AtomicLong messagesReceived = new AtomicLong();
+  private final AtomicLong keepAlivesSent = new AtomicLong();
+  private final AtomicLong lastMessageSentTimeMsec = new AtomicLong();
+  private final AtomicLong lastMessageReceivedTimeMsec = new AtomicLong();
+  private volatile Callable<Integer> localFlowControlPollable;
```
carl-mastrangelo: optional: Callable forces boxing, and throws Exception.  You might consider making a tiny interface for this purpose.

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,212 @@
...
+  private volatile Callable<Integer> remoteFlowControlPollable;
+
+  private final StreamTracer streamTracer = new StreamTracer() {
+    @Override
+    public void streamClosed(Status status) {
+      if (status.isOk()) {
+        streamsSucceeded.incrementAndGet();
```
carl-mastrangelo: consider using the getAndIncrement variant, which doesn't do a +1 on the result of the getAndAddLong call

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,212 @@
...
+public final class TransportTracer {
+  // streamsStarted happens serially, so a volatile is sufficient
+  private volatile long streamsStarted;
+  private long lastStreamCreatedTimeMsec;
+  private final AtomicLong streamsSucceeded = new AtomicLong();
+  private final AtomicLong streamsFailed = new AtomicLong();
+  private final AtomicLong messagesSent = new AtomicLong();
+  private final AtomicLong messagesReceived = new AtomicLong();
+  private final AtomicLong keepAlivesSent = new AtomicLong();
+  private final AtomicLong lastMessageSentTimeMsec = new AtomicLong();
```
carl-mastrangelo: Let's use nanoseconds, which are used more uniformly.

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  /**
+   * Updates an AtomicLong representing a msec timestamp. Avoids races and only allows the value
+   * to increase.
+   */
+  private static void updateMsecTimestamp(AtomicLong timestamp) {
+    long now = System.currentTimeMillis();
+    long oldVal = timestamp.get();
+    while (oldVal < now && !timestamp.compareAndSet(oldVal, now)) {
```
carl-mastrangelo: nit: you can probably just try once.

Alternatively, you could use lazySet.  I don't think the results need to be urgently written.

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * Called by the transport to report a stream has started. For clients, this happens when a header
+   * is sent. For servers, this happens when a header is received. This method must be called from
+   * only one thread, but the resulting stats may be read from any thread.
+   */
+  public void reportStreamStarted() {
+    // @SuppressWarnings is not allowed on an increment statement, using another var here.
+    long newStreamsStarted = streamsStarted + 1;
+    streamsStarted = newStreamsStarted;
```
carl-mastrangelo: You might consider using the atomic increment function here.  The read and the write count as two barriers, where the atomic increment counts as just one.  My gut says atomic would be about the same speed.

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,212 @@
...
+/**
+ * A class for gathering statistics about a transport. This is an experimental feature.
+ */
+public final class TransportTracer {
+  // streamsStarted happens serially, so a volatile is sufficient
+  private volatile long streamsStarted;
+  private long lastStreamCreatedTimeMsec;
+  private final AtomicLong streamsSucceeded = new AtomicLong();
+  private final AtomicLong streamsFailed = new AtomicLong();
+  private final AtomicLong messagesSent = new AtomicLong();
```
carl-mastrangelo: One idea for optimizing this tracer would be to make none of these atomics and just be plain ints.  On read, this class could return a container.  That would make the write case very fast, something we expect to happen often,  and the read case to be slow, which should be seldom.

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,212 @@
...
+  public long getLastMessageReceivedTimeMsec() {
+    return lastMessageReceivedTimeMsec.get();
+  }
+
+  /**
+   * Returns the remote flow control window as reported by the callback of
+   * {@link #setRemoteFlowControlWindowPollable}. Returns -1 if no callback was registered.
+   */
+  public int getRemoteFlowControlWindow() {
+    if (remoteFlowControlPollable == null) {
```
carl-mastrangelo: store the poller in the local?   that will avoid the double volatile read.

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * Updates an AtomicLong representing a msec timestamp. Avoids races and only allows the value
+   * to increase.
+   */
+  private static void updateMsecTimestamp(AtomicLong timestamp) {
+    long now = System.currentTimeMillis();
+    long oldVal = timestamp.get();
+    while (oldVal < now && !timestamp.compareAndSet(oldVal, now)) {
+      // CAS failed, read new timestamp and maybe try again
```
carl-mastrangelo: _technically_ the system clock could change between each iteration of the loop.  I don't think this is a serious issue.

core/src/main/java/io/grpc/internal/TransportTracer.java
carl-mastrangelo: A second alternative would be getAndSet.  Most of the time, the clock goes forward.  getAndSet would allow you to check if the most recent time was actually  greater than timestamp, and let you compareAndSet to roll it back.

That would avoid the initial read too.

core/src/test/java/io/grpc/internal/ServerImplTest.java
```diff
@@ -354,7 +355,7 @@ public void methodNotFound() throws Exception {
     Metadata requestHeaders = new Metadata();
     StatsTraceContext statsTraceCtx =
         StatsTraceContext.newServerContext(
-            streamTracerFactories, "Waiter/nonexist", requestHeaders);
+            streamTracerFactories, null, "Waiter/nonexist", requestHeaders);
```
carl-mastrangelo: please extract this to a member variable.  A lot of our functions have high parameter counts, so seeing `null` makes it hard to remember which param it was for.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -328,6 +352,36 @@ public void run() {
           keepAliveTimeInNanos, keepAliveTimeoutInNanos, true /* keepAliveDuringTransportIdle */);
       keepAliveManager.onTransportStarted();
     }
+    if (transportTracer != null) {
+      class FlowControlPollable implements Callable<Integer> {
```
carl-mastrangelo: final

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -328,6 +352,36 @@ public void run() {
...
+              public void run() {
+                promise.setSuccess(controller.windowSize(connection.connectionStream()));
+              }
+            });
+            return promise.get();
+          }
+        }
+      }
+
+      transportTracer.setRemoteFlowControlWindowPollable(
```
carl-mastrangelo: Since these are often called together, it would probably be a good idea to bundle the results and batch the calls.

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: Done

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: Done

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: Done

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: Added an unsychronized counter whose value we copy to the volatile, resulting in 1 volatile write.

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: Done

core/src/test/java/io/grpc/internal/ServerImplTest.java
zpencer: Done

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
zpencer: Done

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
zpencer: done

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: Done

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: Wouldn't we still need to have synchronization between the writes and reads to ensure that we are reading the correct values? I might be missing something here.

zpencer:
Note: Turns out there's 1 flakey test.

I've changed the tracer to only be callable from the transport thread. It looks like a bit of duplication with stream tracers, but stream tracers are required to be thread safe today. I know Kun filed a ticket to relax the guarantees a bit. When that's done this can be collapsed into that code path.

zpencer:
@carl-mastrangelo ping

zpencer:
I noticed that I missed the fact that the framer and deframers do not work on the transport thread. The transport stats can not be updated from there because it is no longer synchronized. Will update with changes.

core/src/main/java/io/grpc/internal/ShardedAtomicLongCounter.java
```diff
@@ -0,0 +1,48 @@
...
+  ShardedAtomicLongCounter(int numShards) {
+    counters = new AtomicLong[numShards];
+    for (int i = 0; i < numShards; i++) {
+      counters[i] = new AtomicLong();
+    }
+  }
+
+  @Override
+  public void add(long delta) {
+    counters[(int) (Thread.currentThread().getId() % counters.length)].getAndAdd(delta);
```
carl-mastrangelo: fyi thread ids are not evenly distributed.  this will still result in contention.  

Also, modulo is almost certainly a bad idea, more expensive than the atomic increment.  Consider forcing the size to be a power of two.

zpencer: I was originally avoiding hashCode because of its relationship with biased locking. But maybe it's ok to use that, since it doesn't make much sense to use the thread object as a lock.

zpencer:
PTAL

core/src/main/java/io/grpc/internal/ShardedAtomicLongCounter.java
```diff
@@ -0,0 +1,76 @@
...
+
+import com.google.common.annotations.VisibleForTesting;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * An implementation of {@link LongCounter} that works by sharded across an array of
+ * {@link AtomicLong} objects. Do not instantiate directly, instead use {@link LongCounterFactory}.
+ */
+final class ShardedAtomicLongCounter implements LongCounter {
+  final AtomicLong[] counters;
```
carl-mastrangelo: AtomicLongArray ?

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,180 @@
...
+  private long streamsStarted;
+  private long lastStreamCreatedTimeNanos;
+  private long streamsSucceeded;
+  private long streamsFailed;
+  private long keepAlivesSent;
+  private FlowControlReader flowControlWindowReader;
+
+  // TODO(zpencer): msg sent can piggyback on framer's writequeue and avoid syncing the update
+  private final LongCounter messagesSent = LongCounterFactory.create();
+  private volatile long lastMessageSentTimeNanos;
```
carl-mastrangelo: non volatile

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,180 @@
...
+  // deframing happens on the application thread, and there's no easy way to avoid synchronization
+  private final LongCounter messagesReceived = LongCounterFactory.create();
```
carl-mastrangelo: this should be a regular long right?  If its only ever called from one thread

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -0,0 +1,180 @@
...
+   */
+  public void setFlowControlWindowReader(FlowControlReader flowControlWindowReader) {
+    this.flowControlWindowReader = Preconditions.checkNotNull(flowControlWindowReader);
+  }
+
+  /**
+   * A container that holds the local and remote flow control window sizes.
+   */
+  public static final class FlowControlWindows {
+    public final int remoteBytes;
```
carl-mastrangelo: nit: these should probably be longs

netty/src/main/java/io/grpc/netty/NettyServerTransport.java
```diff
@@ -164,12 +170,35 @@ private void notifyTerminated(Throwable t) {
...
+    channel.eventLoop().execute(new Runnable() {
+      @Override
+      public void run() {
+        promise.setSuccess(transportTracer.getStats());
+      }
+    });
+    try {
+      return promise.get();
+    } catch (InterruptedException e) {
+      throw new RuntimeException(e);
```
carl-mastrangelo: probably shoudl set the interrupted bit here

netty/src/main/java/io/grpc/netty/NettyServerTransport.java
```diff
@@ -164,12 +170,35 @@ private void notifyTerminated(Throwable t) {
...
+    }
+    final Promise<TransportTracer.Stats> promise = channel.eventLoop().newPromise();
+    channel.eventLoop().execute(new Runnable() {
+      @Override
+      public void run() {
+        promise.setSuccess(transportTracer.getStats());
+      }
+    });
+    try {
+      return promise.get();
```
carl-mastrangelo: probably should set some timeout here, perhaps 1 minute or something just for safety

netty/src/main/java/io/grpc/netty/NettyServerTransport.java
```diff
@@ -164,12 +170,35 @@ private void notifyTerminated(Throwable t) {
...
 
+  @Override
+  @Nullable
+  public TransportTracer.Stats getTransportStats() {
+    if (channel.eventLoop().inEventLoop()) {
+      // This is necessary, otherwise we will block forever on the promise in the else case
+      return transportTracer.getStats();
+    }
+    final Promise<TransportTracer.Stats> promise = channel.eventLoop().newPromise();
+    channel.eventLoop().execute(new Runnable() {
```
carl-mastrangelo: Probably want to use `submit()` here.  If there is an exception, the promise may not be resolved.

netty/src/main/java/io/grpc/netty/SendResponseHeadersCommand.java
```diff
@@ -53,13 +68,13 @@ public boolean equals(Object that) {
...
         && thatCmd.headers.equals(headers)
-        && thatCmd.endOfStream == endOfStream;
+        && thatCmd.status.equals(status);
   }
 
   @Override
   public String toString() {
     return getClass().getSimpleName() + "(stream=" + stream.id() + ", headers=" + headers
-        + ", endOfStream=" + endOfStream + ")";
+        + ", status=" + status + ")";
```
carl-mastrangelo: for line 81:    If equals changes, the hashcode should change too.  I think status is immutable so it shouldn't change.

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
```diff
@@ -1346,6 +1347,203 @@ public void interactionsAfterClientStreamCancelAreNoops() throws Exception {
...
+        firstTimestamp = TimeUnit.NANOSECONDS.toMillis(after.lastStreamCreatedTimeNanos);
+        assertThat(System.currentTimeMillis() - firstTimestamp).isAtMost(50L);
+      }
+
+      ServerStream serverStream = serverStreamCreation.stream;
+      serverStream.close(Status.OK, new Metadata());
+    }
+
+    // start second stream
+    {
```
carl-mastrangelo: this is a pretty uncommon pattern, its a little jarring to see it.

ejona86: I find this to be a good but underutilized pattern. It is maybe a bit too aggressive in this case though: it's not helping much with `before` and `after`.

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: This is currently happening in the framer, which is in the application thread(s). I will follow up with a PR to move this counter increment to the transport thread, since this PR is already big enough.

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: Netty returns only `int`s here, but I don't have a problem upcasting it in case that changes.

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: Deframing happens on application threads so we can't easily avoid synchronization the same way we can for framing.

netty/src/main/java/io/grpc/netty/SendResponseHeadersCommand.java
zpencer: If we want the hashcode to be better, I think we can include all the fields since we're already considering them for `equals`.

core/src/main/java/io/grpc/internal/ShardedAtomicLongCounter.java
zpencer: Done.

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
zpencer: Changed to use `{}` blocks less aggressively.

netty/src/main/java/io/grpc/netty/NettyServerTransport.java
zpencer: Done

netty/src/main/java/io/grpc/netty/NettyServerTransport.java
zpencer: Done

netty/src/main/java/io/grpc/netty/NettyServerTransport.java
zpencer: Done.

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: Update: The http/2 spec states that the flow control window must always be below 2^31-1, which makes an int technically sufficient.

```
A sender MUST NOT allow a flow-control window to exceed 2^31-1 octets. If a sender receives a WINDOW_UPDATE that causes a flow-control window to exceed this maximum, it MUST terminate either the stream or the connection, as appropriate. For streams, the sender sends a RST_STREAM with an error code of FLOW_CONTROL_ERROR; for the connection, a GOAWAY frame with an error code of FLOW_CONTROL_ERROR is sent.
```

zpencer:
@carl-mastrangelo : PTAL

netty/src/main/java/io/grpc/netty/NettyServerTransport.java
```diff
@@ -177,19 +180,23 @@ private void notifyTerminated(Throwable t) {
...
+          public TransportTracer.Stats call() throws Exception {
+            return transportTracer.getStats();
+          }
+        });
     try {
-      return promise.get();
+      // Arbitrarily chosen timeout to avoid blocking forever
+      return future.get(1, TimeUnit.MINUTES);
     } catch (InterruptedException e) {
+      Thread.interrupted();
```
carl-mastrangelo: `Thread.currentThread().interrupt()`

netty/src/main/java/io/grpc/netty/SendResponseHeadersCommand.java
```diff
@@ -79,6 +80,6 @@ public String toString() {
 
   @Override
   public int hashCode() {
-    return stream.hashCode();
+    return Objects.hashCode(stream, headers, status);
```
carl-mastrangelo: sorry to be dumb, but I think header is mutable, and shouldn't go in here.  Just status should be.

zpencer: True, I suppose somebody can create a SendResponseHeadersCommand before filling out the content of the headers object.

netty/src/main/java/io/grpc/netty/NettyServerTransport.java
zpencer: Fixed

netty/src/main/java/io/grpc/netty/NettyServerTransport.java
```diff
@@ -164,12 +173,39 @@ private void notifyTerminated(Throwable t) {
...
+    final Future<TransportTracer.Stats> future = channel.eventLoop().submit(
+        new Callable<TransportTracer.Stats>() {
+          @Override
+          public TransportTracer.Stats call() throws Exception {
+            return transportTracer.getStats();
+          }
+        });
+    try {
+      // Arbitrarily chosen timeout to avoid blocking forever
+      return future.get(1, TimeUnit.MINUTES);
```
ejona86: This is dangerous, as it must not be on _any_ event loop (not just this transport's); otherwise you can forever-blocking between two different event loops (consider the cases that lock ordering tries to solve). I don't see any users of this API so I don't know what's best, but we should either: return a future or have a callback be passed in.

zpencer: This is meant to be called from the thread handling the channelz request. I agree making this a Future would better communicate the potential to block. We also get the nice side effect of being able to kick off many reads in parallel.

core/src/main/java/io/grpc/internal/ShardedAtomicLongCounter.java
```diff
@@ -0,0 +1,73 @@
...
+
+import com.google.common.annotations.VisibleForTesting;
+import java.util.concurrent.atomic.AtomicLongArray;
+
+/**
+ * An implementation of {@link LongCounter} that works by sharded across an array of
+ * {@link AtomicLong} objects. Do not instantiate directly, instead use {@link LongCounterFactory}.
+ */
+final class ShardedAtomicLongCounter implements LongCounter {
+  final AtomicLongArray counters;
```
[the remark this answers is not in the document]
zpencer: This was once visible for testing but no longer needed, making it private.

zpencer:
@ejona86 do you have any more concerns with this PR?

PR #3459, opened 2017-09-14
context: log severe warning if ancestry chain > 1000

zpencer:
If a context has an unreasonable number of ancestors, then
chances are this is an application error. Log the stack trace to
notify the user and aid in debugging.

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -100,10 +100,9 @@ public class Context {
       new PersistentHashArrayMappedTrie<Key<?>, Object>();
 
   // Long chains of contexts are suspicious and usually indicate a misuse of Context.
-  // The threshold and interval are arbitrarily chosen.
+  // The threshold is arbitrarily chosen.
   // VisibleForTesting
-  static final int CONTEXT_DEPTH_WARN_THRESH = 200;
-  private static final int CONTEXT_WARN_INTERVAL = 20;
+  static final int CONTEXT_DEPTH_WARN_THRESH = 1000;
 
   /**
    * The logical root context which is the ultimate ancestor of all contexts. This context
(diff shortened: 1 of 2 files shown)


adriancole:
would be nice to show the max ancestry in a "typical" grpc app, configured with metrics and tracing, just for comparison's sake

context/src/main/java/io/grpc/Context.java
```diff
@@ -99,6 +99,12 @@
   private static final PersistentHashArrayMappedTrie<Key<?>, Object> EMPTY_ENTRIES =
       new PersistentHashArrayMappedTrie<Key<?>, Object>();
 
+  // Long chains of contexts are suspicious and usually indicate a misuse of Context.
+  // The threshold and interval are arbitrarily chosen.
+  // VisibleForTesting
+  static final int CONTEXT_DEPTH_WARN_THRESH = 200;
+  private static final int CONTEXT_WARN_INTERVAL = 20;
```
ejona86: Let's not do the interval warning at this point. Since we've not been warning before, let's start slow.

context/src/main/java/io/grpc/Context.java
```diff
@@ -99,6 +99,12 @@
   private static final PersistentHashArrayMappedTrie<Key<?>, Object> EMPTY_ENTRIES =
       new PersistentHashArrayMappedTrie<Key<?>, Object>();
 
+  // Long chains of contexts are suspicious and usually indicate a misuse of Context.
+  // The threshold and interval are arbitrarily chosen.
+  // VisibleForTesting
+  static final int CONTEXT_DEPTH_WARN_THRESH = 200;
```
ejona86: @adriancole, I'm okay with bumping this higher. If you're concerned about 200, do you have a suggestion on where to start, like 1000 or 5000?

The main concern at the moment is infinite mutation, because finding those in real life applications can be hard.

adriancole:
1000 feels safer starting point thanks!

Issue #3460, opened 2017-09-14
Tracking issue for deleting deprecated methods in StreamTracer


zhangkun83:
`outboundMessage()` and `inboundMessage()` will be released as deprecated in 1.7.0, thus should be deleted in 1.8.0.

PR #3461, opened 2017-09-14
core: record individual messages with sizes to Census/tracing

zhangkun83:
Two methods, `outboundMessageSent()` and `inboundMessageRead()` are added to StreamTracer in order to associate individual messages with sizes. Both types of sizes are optional, as allowed by Census tracing.

Both methods accept a sequence number as the type ID as required by Census. The original `outboundMesage()` and `inboundMessage()` are also replaced by overrides that take the sequence number, to better match the new methods. The deprecation of the old overrides are tracked by #3460

### DIFF
--- a/core/src/main/java/io/grpc/StreamTracer.java
+++ b/core/src/main/java/io/grpc/StreamTracer.java
@@ -48,7 +48,8 @@ public abstract class StreamTracer {
    * about the message, but doesn't have further guarantee such as whether the message is serialized
    * or not.
    *
-   * @param seqNo the sequential number of the message within the stream, starting from 0
+   * @param seqNo the sequential number of the message within the stream, starting from 0.  It can
+   *              be used to correlate with {@link #outboundMessageSent} for the same message.
    */
   public void outboundMessage(int seqNo) {
   }
@@ -69,7 +70,8 @@ public abstract class StreamTracer {
    * about the message, but doesn't have further guarantee such as whether the message is
    * deserialized or not.
    *
-   * @param seqNo the sequential number of the message within the stream, starting from 0
+   * @param seqNo the sequential number of the message within the stream, starting from 0.  It can
+   *              be used to correlate with {@link #inboundMessageRead} for the same message.
    */
   public void inboundMessage(int seqNo) {
   }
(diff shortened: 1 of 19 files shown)


zhangkun83:
@ericgribkoff, just a heads up: this PR changes how sizes are reported to stats in the framer and deframer.

core/src/main/java/io/grpc/StreamTracer.java
```diff
@@ -36,19 +36,65 @@ public void streamClosed(Status status) {
...
   }
 
   /**
+   * An outbound message has been passed to the stream.  This is called as soon as the stream knows
+   * about the message, but doesn't have further guarantee such as whether the message is serialized
+   * or not.
+   *
+   * @param seqNo the sequential number of the message within the stream, starting from 0
+   */
+  public void outboundMessage(int seqNo) {
```
ejona86: Why is `seqNo` necessary? Wouldn't it be easy for the implementation to do `i++`? Or is this to make clear the correlation between outboundMessage events and the respective outboundMessageSent event?

core/src/main/java/io/grpc/internal/MessageDeframer.java
```diff
@@ -284,6 +285,12 @@ private boolean readRequiredBytes() {
         totalBytesRead += toRead;
         nextFrame.addBuffer(unprocessed.readBytes(toRead));
       }
+      // There is no reliable way to get the uncompressed size per message when it's compressed,
+      // because the uncompressed bytes are provided through an InputStream whose total size is
+      // unknown until all bytes are read, and we don't know when it happens.
+      if (state == State.BODY) {
+        statsTraceCtx.inboundMessageRead(currentMessageSeqNo, requiredLength, -1);
```
ejona86: `processBody()` seems a better place to do this.

core/src/main/java/io/grpc/StreamTracer.java
zhangkun83: The main reason is to make the correlation clear.

core/src/main/java/io/grpc/internal/MessageDeframer.java
zhangkun83: Done.

core/src/main/java/io/grpc/StreamTracer.java
ejona86: Make a note of that in the Javadoc.

PR #3463, opened 2017-09-14
core,netty,okhttp,protobuf-lite: avoid @Beta guava classes

zpencer:
Fix more of these usages so we can enable `@Beta` checks

For `Ticker`: replace with a factory for StopWatch. We can make
fake stop watches from unit tests, where `@Beta` doesn't matter.

### DIFF
--- a/core/src/main/java/io/grpc/internal/GrpcUtil.java
+++ b/core/src/main/java/io/grpc/internal/GrpcUtil.java
@@ -483,7 +483,9 @@ public final class GrpcUtil {
    */
   public static ThreadFactory getThreadFactory(String nameFormat, boolean daemon) {
     if (IS_RESTRICTED_APPENGINE) {
-      return MoreExecutors.platformThreadFactory();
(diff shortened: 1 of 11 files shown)


core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -483,7 +483,9 @@ public void close(ScheduledExecutorService instance) {
    */
   public static ThreadFactory getThreadFactory(String nameFormat, boolean daemon) {
     if (IS_RESTRICTED_APPENGINE) {
-      return MoreExecutors.platformThreadFactory();
+      @SuppressWarnings("BetaApi")
+      ThreadFactory factory = MoreExecutors.platformThreadFactory();
```
carl-mastrangelo: idea: in the future we could reflectively find this so that future versions of guava work with older versions of gRPC.

zpencer:
retest this please

PR #3465, opened 2017-09-14
okhttp: update caller of Headers.createRequestHeaders

zpencer:


okhttp/src/test/java/io/grpc/okhttp/HeadersTest.java
```diff
@@ -45,7 +45,7 @@ public void createRequestHeaders_sanitizes() {
     String authority = "localhost";
     String userAgent = "useragent";
 
-    List<Header> headers = Headers.createRequestHeaders(metaData, path, authority, userAgent);
+    List<Header> headers = Headers.createRequestHeaders(metaData, path, authority, userAgent, true);
```
ejona86: Let's use `false`, as that's the more common case.

PR #3467, opened 2017-09-15
core/compiler: register Span names for code-generated methods.

### LINKED ISSUE #3359: Census: enable requestz sampling for generated methods
Census has an open-source requestz implementation that can show traces of live, sampled and error RPCs. In order to sample RPCs, gRPC needs to call [SampledSpanStore.registerSpanNamesForCollection()](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java) with any known method name. This is fairly expensive, thus should not be called on a per-RPC basis. A few options:
1. In a static initializer in the generated code. If we don't want to call Census directly from the generated code, maybe call Census code in a helper function in `ClientCalls` and `ServerCalls` which is called from generated code.
2. In `MethodDescriptor` constructor. This will be called once per method, which might slow start-up speed. This would also all custom-created methods to be registered, @bogdandrutu is this desirable, or do you only expect generated methods to be registered?

Option 1 sounds like a better option. @ejona86 WDYT?

zhangkun83:
Resolves #3359

TODOs: switch to a released version of opencensus, and file issues for the added experimental methods.

### DIFF
--- a/testing-proto/src/test/java/io/grpc/testing/protobuf/SimpleServiceTest.java
+++ b/testing-proto/src/test/java/io/grpc/testing/protobuf/SimpleServiceTest.java
@@ -16,13 +16,21 @@
 
 package io.grpc.testing.protobuf;
 
+import static com.google.common.truth.Truth.assertThat;
 import static io.grpc.MethodDescriptor.MethodType.BIDI_STREAMING;
 import static io.grpc.MethodDescriptor.MethodType.CLIENT_STREAMING;
 import static io.grpc.MethodDescriptor.MethodType.SERVER_STREAMING;
 import static io.grpc.MethodDescriptor.MethodType.UNARY;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 
+import io.grpc.Grpc.Side;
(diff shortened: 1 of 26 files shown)


testing-proto/src/test/java/io/grpc/testing/protobuf/SimpleServiceTest.java
```diff
@@ -51,4 +59,28 @@ public void serviceMethodDescriotrs() {
...
+    ArrayList<String> expectedSpans = new ArrayList<String>();
+    for (String methodName : methodNames) {
+      for (Side side : Side.values()) {
+        expectedSpans.add(Grpc.makeTraceSpanName(side, methodName));
+      }
+    }
+
+    SampledSpanStore sampledStore = Tracing.getExportComponent().getSampledSpanStore();
+    Set<String> registeredSpans = sampledStore.getRegisteredSpanNamesForCollection();
+    System.err.println("XXX: " + sampledStore);
```
ejona86: Remove.

zhangkun83:
Closing it in favor of #3471

PR #3468, opened 2017-09-15
android-interop-testing: use Java Style enums

carl-mastrangelo:
This makes it easier to convert over to Proto lite, which will be
replacing nano in the near future.

The enum style will be going away too after the migration.


carl-mastrangelo:
This CL is a port originally from @arielbackenroth

carl-mastrangelo:
retest this please

PR #3471, opened 2017-09-15
core/compiler: register Span names for code-generated methods (take 2).

### LINKED ISSUE #3359: Census: enable requestz sampling for generated methods

zhangkun83:
Resolves #3359

This is a more favorable approach than #3467.  Doing the registration
in MethodDescriptor should allow us to deregister in case the
generated stub and its MethodDescriptors are garbage-collected
routinely, e.g., if they are loaded by a separate ClassLoader.

TODOs: switch to a released version of opencensus, and file issues for the added experimentals.


build.gradle
```diff
@@ -57,6 +57,9 @@ subprojects {
     repositories {
         mavenCentral()
         mavenLocal()
+        maven {
+          url "https://oss.sonatype.org/content/repositories/snapshots/"
+        }
```
carl-mastrangelo: is this temporary?

core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -530,12 +556,32 @@ private Builder() {}
     }
 
     /**
+     * Sets whether the new MethodDescriptor should be registered into the tracing system, so that
+     * RPCs on this method may be collected and stored.
+     *
+     * @since 1.7.0
+     */
+    public Builder<ReqT, RespT> setShouldRegisterForTracing(boolean value) {
```
carl-mastrangelo: s/Should//

core/src/main/java/io/grpc/Grpc.java
```diff
@@ -40,4 +40,11 @@ private Grpc() {
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1710")
   public static final Attributes.Key<SSLSession> TRANSPORT_ATTR_SSL_SESSION =
           Attributes.Key.of("ssl-session");
+
+  //TODO(zhangkun83): create tracking issue
+  @ExperimentalApi
+  public enum Side {
```
carl-mastrangelo: This should probably exist in the InternalMethodDescriptor class.

core/src/main/java/io/grpc/Grpc.java
```diff
@@ -40,4 +40,11 @@ private Grpc() {
...
+    CLIENT,
+    SERVER
```
carl-mastrangelo: Consider moving the prefix generating method to here.

ejona86:
+1 to Carl's comments.

core/src/main/java/io/grpc/MethodDescriptor.java
zhangkun83: Done.

core/src/main/java/io/grpc/Grpc.java
zhangkun83: MethodDescriptor can't reference InternalMethodDescriptor. Since generateTraceSpanName() is no longer public, it's not necessary to make it good-looking. I deleted the enum and switched to boolean instead.

zhangkun83:
retest this please

PR #3472, opened 2017-09-16
netty:  use channel writability as signal to flush

carl-mastrangelo:
The goal here is to flush when the channel is going to buffer.  In
order to accomplish this, the channel's isWritable() method is used
to check if the state changes after writes.

The tests have been modified to not use mocks, and more accurately
check the desired behavior.  That said, the tests do not yet
reflect the target behavior.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/WriteQueue.java
+++ b/netty/src/main/java/io/grpc/netty/WriteQueue.java
@@ -29,29 +29,18 @@ import java.util.concurrent.atomic.AtomicBoolean;
 /**
  * A queue of pending writes to a {@link Channel} that is flushed as a single unit.
  */
-class WriteQueue {
+class WriteQueue implements Runnable {
 
   // Dequeue in chunks, so we don't have to acquire the queue's log too often.
   @VisibleForTesting
-  static final int DEQUE_CHUNK_SIZE = 128;
-
-  /**
-   * {@link Runnable} used to schedule work onto the tail of the event loop.
-   */
-  private final Runnable later = new Runnable() {
-    @Override
-    public void run() {
-      flush();
-    }
-  };
+  static final int FLUSH_LIMIT = 128;
 
   private final Channel channel;
-  private final Queue<QueuedCommand> queue;
+  private final Queue<QueuedCommand> queue = new ConcurrentLinkedQueue<QueuedCommand>();
   private final AtomicBoolean scheduled = new AtomicBoolean();
 
   public WriteQueue(Channel channel) {
     this.channel = Preconditions.checkNotNull(channel, "channel");
-    queue = new ConcurrentLinkedQueue<QueuedCommand>();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


netty/src/main/java/io/grpc/netty/WriteQueue.java
```diff
@@ -115,26 +105,31 @@ void enqueue(Runnable runnable, boolean flush) {
...
+      do {
+        while ((cmd = queue.poll()) != null) {
+          cmd.run(channel);
+          // There are two reasons to leave the loop:
+          // 1.  The channel's writability changed.
+          // 2.  We've done a lot of writes.
+          // Either way, we want to try another flush.  Use the non shortcircuiting | operator to
+          // make sure
+          if ((wasWritable != (wasWritable = channel.isWritable())) | (++writes == FLUSH_LIMIT)) {
+            // Always reset buffered writes, even in the good case.  We may have just exited the
```
ejona86: good case and "bad flush" aren't really defined. How about just do the `writes = 0` immediately before/after the `flush()`?

netty/src/main/java/io/grpc/netty/WriteQueue.java
```diff
@@ -115,26 +105,31 @@ void enqueue(Runnable runnable, boolean flush) {
...
-          flushedOnce = true;
+      do {
+        while ((cmd = queue.poll()) != null) {
+          cmd.run(channel);
+          // There are two reasons to leave the loop:
+          // 1.  The channel's writability changed.
+          // 2.  We've done a lot of writes.
+          // Either way, we want to try another flush.  Use the non shortcircuiting | operator to
+          // make sure
+          if ((wasWritable != (wasWritable = channel.isWritable())) | (++writes == FLUSH_LIMIT)) {
```
ejona86: Hoist the `++writes` before the if so you can use `||`?

netty/src/test/java/io/grpc/netty/WriteQueueTest.java
```diff
@@ -184,19 +155,40 @@ void checkException() throws Throwable {
...
+    assertThat(recorder.msgs).hasSize(writes);
   }
 
   static class CuteCommand extends WriteQueue.AbstractQueuedCommand {
 
   }
+
+  private static class Recorder extends ChannelOutboundHandlerAdapter {
+    final List<Object> msgs = new ArrayList<Object>();
+    final List<Object> promises = new ArrayList<Object>();
```
ejona86: What is this used for?

netty/src/test/java/io/grpc/netty/WriteQueueTest.java
```diff
@@ -184,19 +155,40 @@ void checkException() throws Throwable {
...
-    verify(channel, times(writes)).write(isA(CuteCommand.class), eq(promise));
+    assertThat(recorder.msgs).hasSize(writes);
   }
 
   static class CuteCommand extends WriteQueue.AbstractQueuedCommand {
 
   }
+
+  private static class Recorder extends ChannelOutboundHandlerAdapter {
+    final List<Object> msgs = new ArrayList<Object>();
```
ejona86: How is this different than `EmbeddedChannel.outboundMessages()`?

PR #3475, opened 2017-09-18
Errorprone fixes: Avoid catching AssertionError in tests, use .class literal when able

ejona86:
You can review the two commits individually.


ejona86:
@zhangkun83, please review.

PR #3478, opened 2017-09-19
core: remove reference to Guava collect.

zhangkun83:
It was added by mistake in commit 2b1363d58600ac49b73116e00d17302063ef532a

@zpencer, core should not depend on Guava collect because it's not Android-friendly.

### DIFF
--- a/core/src/main/java/io/grpc/internal/GrpcUtil.java
+++ b/core/src/main/java/io/grpc/internal/GrpcUtil.java
@@ -19,6 +19,7 @@ package io.grpc.internal;
 import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
 import com.google.common.base.Stopwatch;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/ServerCallImpl.java
```diff
@@ -90,7 +89,7 @@ public void sendHeaders(Metadata headers) {
     } else {
       if (messageAcceptEncoding != null) {
         // TODO(carl-mastrangelo): remove the string allocation.
-        if (!Iterables.contains(
+        if (!GrpcUtil.iterableContains(
             ACCEPT_ENCODING_SPLITTER.split(new String(messageAcceptEncoding, GrpcUtil.US_ASCII)),
```
carl-mastrangelo: Just use the old splitToList().contains(...) ?

zhangkun83: `splitToList()` is marked as `@Beta` and 2b1363d58600ac49b73116e00d17302063ef532a was moving away from beta APIs.

carl-mastrangelo: then `new ArrayList<>(...).contains()`

zhangkun83: Although the additional allocation from ArrayList is not substantially more than that what's already there, what's wrong with the helper method?

core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -630,6 +631,16 @@ static void closeQuietly(InputStream message) {
     }
   }
 
+  /** Checks whether the given item exists in the iterable. */
+  static <T> boolean iterableContains(Iterable<T> iterable, T item) {
```
zpencer: Maybe add a comment about avoiding guava to explain why this helper exists

ejona86:
note: Good progress has been made to make Guava collect Android-friendly. So it is something that could be re-visited, but using it heavily would probably dramatically increase the number of methods we pull in from Guava.

PR #3480, opened 2017-09-20
core: pull out reusable forwarding test, refactoring forwarding tests

zpencer:
The reflection based test from ForwardingChannelBuilderTest can
be reused to test other forwarding classes. The advantage of this
approach is that the test can catch the case where we add a new
method to the super class but forget to update the forwarding
class.


core/src/test/java/io/grpc/AbstractForwardingTest.java
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * An abstract base class for testing forwarding classes. This automatically checks that
+ * all public methods of the {@link #delegateClass()} are forwarded by the forwarder to the
+ * delegate. This does NOT verify that arguments are forwarded properly. It only alerts
+ * the developer if a forward method is missing.
+ */
+@RunWith(JUnit4.class)
```
carl-mastrangelo: nit: this shouldn't have `@RunWith` on it.

core/src/test/java/io/grpc/AbstractForwardingTest.java
```diff
@@ -0,0 +1,82 @@
...
+      Object[] args = new Object[argTypes.length];
+      for (int i = 0; i < argTypes.length; i++) {
+        args[i] = Defaults.defaultValue(argTypes[i]);
+      }
+      method.invoke(forwarder(), args);
+      try {
+        method.invoke(verify(mockDelegate()), args);
+      } catch (InvocationTargetException e) {
+        // Print the method name at the top to be more readable
+        throw new InvocationTargetException(
```
carl-mastrangelo: AssertionError?

core/src/test/java/io/grpc/ForwardingChannelBuilderTest.java
```diff
@@ -32,11 +31,29 @@
...
+  @Override
+  public ManagedChannelBuilder<?> forwarder() {
+    return testChannelBuilder;
+  }
+
+  @Override
+  public Class<ManagedChannelBuilder<?>> delegateClass() {
+    @SuppressWarnings("unchecked")
+    Class<ManagedChannelBuilder<?>> ret =
+        (Class<ManagedChannelBuilder<?>>) ((Object) ManagedChannelBuilder.class);
```
carl-mastrangelo: Might as well just cast it to `Class`

core/src/test/java/io/grpc/ForwardingChannelBuilderTest.java
```diff
@@ -32,11 +31,29 @@
...
+    return mockDelegate;
+  }
+
+  @Override
+  public ManagedChannelBuilder<?> forwarder() {
+    return testChannelBuilder;
+  }
+
+  @Override
+  public Class<ManagedChannelBuilder<?>> delegateClass() {
```
carl-mastrangelo: If you are a super cool turbo hacker, you can actually extract the class from the generics on line 34, which allows you to move this method in the the parent.

core/src/test/java/io/grpc/AbstractForwardingTest.java
```diff
@@ -0,0 +1,82 @@
...
+   */
+  public abstract Class<T> delegateClass();
+
+  @Test
+  public void methodsForwarded() throws Exception {
+    assertTrue(mockingDetails(mockDelegate()).isMock());
+    assertFalse(mockingDetails(forwarder()).isMock());
+
+    for (Method method : delegateClass().getDeclaredMethods()) {
+      if (Modifier.isStatic(method.getModifiers()) || Modifier.isPrivate(method.getModifiers())) {
```
dapengzhang0: Define a protected method that returns a collection of methods, so that subclasses can override it in special cases (e.g., a special method throws IllegalArgumentException on non-positive int argument, this method should be filtered out and handled specially)?

core/src/test/java/io/grpc/AbstractForwardingTest.java
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * Returns the class of the object being forwarded. There is no easy way to find this
+   * reflectively from the value of {@link #mockDelegate()}.
+   * {@code mockDelegate().getClass().getSuperclass()} works at the moment but this would rely on
+   * mockito internals. It's easier to explicitly specify the class.
+   */
+  public abstract Class<T> delegateClass();
+
+  @Test
+  public void methodsForwarded() throws Exception {
```
carl-mastrangelo: nit: I think this should be a protected final method, and accept a set of methods to ignore.  The base class can call this test.  Sometimes a final method or a constructor (like Builder.build()) messes up this reflection test.

core/src/test/java/io/grpc/AbstractForwardingTest.java
```diff
@@ -0,0 +1,82 @@
...
+ * all public methods of the {@link #delegateClass()} are forwarded by the forwarder to the
+ * delegate. This does NOT verify that arguments are forwarded properly. It only alerts
+ * the developer if a forward method is missing.
+ */
+@RunWith(JUnit4.class)
+public abstract class AbstractForwardingTest<T> {
+  /**
+   * Returns a mock object that can be used with {@code Mockito.verify()}.
+   */
+  public abstract T mockDelegate();
```
dapengzhang0: `protected`? because its only for overriding, not for invocation outside this class.
Ditto for below.

core/src/test/java/io/grpc/AbstractForwardingTest.java
dapengzhang0: I think `@Test` can only be `public`.

core/src/test/java/io/grpc/ForwardingChannelBuilderTest.java
zpencer: TIL

core/src/test/java/io/grpc/AbstractForwardingTest.java
zpencer: Done

core/src/test/java/io/grpc/AbstractForwardingTest.java
zpencer: How about an `@Override`able method that returns a list of ignored methods? This way we minimize the risk that a developer subclass this test and fail to call the test method, but see that all tests are green.

Alternatively, make all these things be constructor args to the class. It may make initializing slightly annoying because it forces everything to be created by the time we call `super()`.

carl-mastrangelo: I'm slightly against putting tests in the parent class, mainly because it is annoying to run them manually.  You can't click on the super class method to run from the child class.  

As for people forgetting: if the super class became a utility class there would be no risk of forgetting.

zpencer: At least in intellij, right clicking the superclass method lets you choose running a child class or running all child classes.

zpencer:
Turned the abstract class into a util to address some of the concerns. Also lets us avoid some verbose reflection.

zpencer:
@dapengzhang0 ping on this one

core/src/test/java/io/grpc/ForwardingTestUtil.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * A util class to help test forwarding classes.
+ */
+public final class ForwardingTestUtil {
+  /**
+   * Same as {@link #testMethodsForwarded(Class, Object, Object, Collection)} but with no skipped
+   * methods.
+   */
+  public static <T> void testAllMethodsForwarded(
+      Class<?> delegateClass,
```
dapengzhang0: `Class<T>`

core/src/test/java/io/grpc/ForwardingTestUtil.java
```diff
@@ -0,0 +1,84 @@
...
+   * This does NOT verify that arguments or return values are forwarded properly. It only alerts
+   * the developer if a forward method is missing.
+   *
+   * @param delegateClass The class whose methods should be forwarded
+   * @param mockDelegate The mockito mock of the delegate class
+   * @param forwarder The forwarder class that forwards to the mockDelegate
+   * @param skippedMethods A collection of methods that are skipped by the test.
+   */
+  public static <T> void testMethodsForwarded(
+      Class<?> delegateClass,
```
dapengzhang0: `Class<T>`

core/src/test/java/io/grpc/ForwardingTestUtil.java
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Use reflection to perform a basic sanity test. The forwarding class should forward all public
+   * methods to the delegate, except for those in skippedMethods.
+   * This does NOT verify that arguments or return values are forwarded properly. It only alerts
+   * the developer if a forward method is missing.
+   *
+   * @param delegateClass The class whose methods should be forwarded
+   * @param mockDelegate The mockito mock of the delegate class
+   * @param forwarder The forwarder class that forwards to the mockDelegate
```
dapengzhang0: s/class/object

zpencer:
@dapengzhang0 PTAL

PR #3485, opened 2017-09-21
core,netty: InputStream support in SslContextBuilder

### LINKED ISSUE #1881: More flexible SSL configuration API
Today `ServerBuilder.useTransportSecurity` takes two files as parameter. However, sometimes only `InputStream`s for the key and cert files are available. Can we add an API to take two input streams?

zpencer:
Also update unit tests to stop writing certs to temp files.

Fixes #1881

### DIFF
--- a/core/src/main/java/io/grpc/ServerBuilder.java
+++ b/core/src/main/java/io/grpc/ServerBuilder.java
@@ -17,6 +17,7 @@
 package io.grpc;
 
 import java.io.File;
+import java.io.InputStream;
 import java.util.concurrent.Executor;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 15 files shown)


core/src/main/java/io/grpc/ServerBuilder.java
```diff
@@ -151,6 +152,18 @@ public T addStreamTracerFactory(ServerStreamTracer.Factory factory) {
...
+   * Makes the server use TLS.
+   *
+   * @param certChain InputStream containing the full certificate chain
+   * @param privateKey InputStream containing the private key
+   *
+   * @return this
+   * @throws UnsupportedOperationException if the server does not support TLS.
+   * @since 1.7.0
+   */
+  public abstract T useTransportSecurity(InputStream certChain, InputStream privateKey);
```
ejona86: We can't add new abstract methods to public classes. We need to convert to `File` or throw.

netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
```diff
@@ -107,6 +109,21 @@ public static SslContextBuilder forServer(File keyCertChainFile, File keyFile) {
...
    * Creates a SslContextBuilder with ciphers and APN appropriate for gRPC.
    *
+   * @see SslContextBuilder#forServer(InputStream, InputStream)
+   * @see #configure(SslContextBuilder)
+   */
+  public static SslContextBuilder forServer(InputStream keyCertChain, InputStream key) {
+    try {
+      return configure(SslContextBuilder.forServer(keyCertChain, key));
+    } finally {
+      GrpcUtil.closeQuietly(keyCertChain);
```
ejona86: Normally the creator of the `InputStream` closes (i.e., with try-with-resources). But also this is supposed to act effectively the same as Netty's SslContextBuilder. I don't think we should be closing here.

PR #3487, opened 2017-09-21
core: unit test for new abstract methods that break API

zpencer:
This only detects a very specific type of API breakage, so it's
not really a complete solution. Still, it can automatically catch
this one case during testing. If we like this approach we can add
similar tests to the rest of our API-level abstract classes.

What do you think?

### DIFF
--- a/core/src/test/java/io/grpc/ServerBuilderTest.java
+++ b/core/src/test/java/io/grpc/ServerBuilderTest.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2017, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/grpc/ServerBuilderTest.java
```diff
@@ -0,0 +1,117 @@
...
+ */
+@RunWith(JUnit4.class)
+public class ServerBuilderTest {
+  @Test
+  public void apiCompatible() throws Exception {
+    // The purpose of this test is to ensure that gRPC developers do not accidentally add new
+    // abstract methods to ServerBuilder, breaking existing ServerBuilder implementations.
+
+    // DO NOT add methods from gRPC API > 1.0 to this class
+    class Builder extends ServerBuilder<Builder> {
```
carl-mastrangelo: You can even name this BuilderApi100  to reduce accidents!

carl-mastrangelo: final

zpencer:
Decided to go with a binary jar differ instead: 
https://github.com/grpc/grpc-java/pull/3500

PR #3488, opened 2017-09-22
compiler: interfaces with java7 support

elandau:
Generated stubs cannot be mocked which makes testing very cumbersome.  This PR introduces an option to generate client interfaces that are not deprecated.  I've also added an option to specify java version compatibility so that generated interfaces can have default methods for java 7+ where the default implementation is to throw an exception. 

These options may be specified using the gradle plugin like so,

```
grpc {
  option 'enable_client_interface=true'
  option 'java_version=7'
}
```

### DIFF
--- a/compiler/src/java_plugin/cpp/java_generator.cpp
+++ b/compiler/src/java_plugin/cpp/java_generator.cpp
@@ -485,8 +485,7 @@ static void PrintDeprecatedDocComment(const ServiceDescriptor* service,
 static void PrintStub(
     const ServiceDescriptor* service,
     std::map<string, string>* vars,
-    Printer* p, StubType type, bool generate_nano,
-    bool enable_deprecated) {
+    Printer* p, StubType type, bool generate_nano, Options& options) {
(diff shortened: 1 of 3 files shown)


carl-mastrangelo:
The recommend way of testing your stubs is to use InProcessChannel, and implement a dummy service.   Mocking the stubs leads to code that is surreptitiously wrong, but appears to work.

ejona86:
Interfaces without `default` is a deal-breaker, as adding a method to a service is now a breaking change. That's unacceptable.

As time has gone on we've been happier and happier with prohibiting mocking stubs. It wasn't just a good or necessary decision, but a _great_ one. Mocking stubs is a dead-end. I don't see us reversing our decision and adding interfaces for stubs.

> Generated stubs cannot be mocked which makes testing very cumbersome.

Let's instead focus on how it is cumbersome and figure out improvements. Mocking is a possible solution, not necessarily the only solution. Feel free to open an issue about specific things that are cumbersome.

PR #3490, opened 2017-09-23
core: remove static field in ManagedChannelImplTest

ericgribkoff:
Several test cases were not properly shutting down the channel, requiring a static `unterminatedChannels` variable to track across all tests the number of remaining channels, which were then garbage collected as a preprocessing step in `#orphanedChannelsAreLogged`. This relies on a timeout to wait for the garbage collector to run, and failed twice on Travis for https://github.com/grpc/grpc-java/pull/3403 (was able to reproduce locally on master branch with 1500+ plus runs) with the following exception:

```
io.grpc.internal.ManagedChannelImplTest > orphanedChannelsAreLogged FAILED
    java.lang.AssertionError: Not true that <1> is at most <0>
        at com.google.common.truth.FailureStrategy.fail(FailureStrategy.java:24)
        at com.google.common.truth.FailureStrategy.fail(FailureStrategy.java:20)
        at com.google.common.truth.Subject.failComparingToStrings(Subject.java:295)
        at com.google.common.truth.Subject.fail(Subject.java:272)
        at com.google.common.truth.ComparableSubject.isAtMost(ComparableSubject.java:104)
        at io.grpc.internal.ManagedChannelImplTest.orphanedChannelsAreLogged(ManagedChannelImplTest.java:1425)
```

This PR cleans up the channels for the test cases that were not doing so, removes the static counter variable and replaces it with an assertion in the `@After` method that the channel was properly terminated.

### DIFF
--- a/core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
+++ b/core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
@@ -30,6 +30,7 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyObject;
 import static org.mockito.Matchers.anyString;
@@ -124,7 +125,6 @@ public class ManagedChannelImplTest {
           .build();
   private static final Attributes.Key<String> SUBCHANNEL_ATTR_KEY =
       Attributes.Key.of("subchannel-attr-key");
-  private static int unterminatedChannels;
   private final String serviceName = "fake.example.com";
   private final String authority = serviceName;
   private final String userAgent = "userAgent";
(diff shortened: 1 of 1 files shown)


ericgribkoff:
I think the `orphanedChannelsAreLogged` test actually depends on state across all of the test classes that may create and not shutdown a channel, not just channels created in the `ManagedChannelImplTest` class (unless I missed something in our test setup, all of the test classes run in a single JVM).  I think this was the reason that the test's "clean-up" block at the beginning had to assert `assertThat(remaining).isAtMost(0);` instead of just asserting equals to zero. 

Is there any reason to keep this test case as-is? There might be an easier alternative that I'm missing, but otherwise it seems this should either (1) be placed in its own class and somehow get junit to fork a new JVM just for this class or (2) have some modification of `ManagedChannelImpl` so that the reference queue can be injected for test purposes, so the checks here don't get impacted by stray references from other tests. 

@carl-mastrangelo what do you think?

ericgribkoff:
@carl-mastrangelo Unfortunately this isn't a complete fix. The test still runs in the same JVM as our other test classes, and can fail when garbage collection of stray references from other tests doesn't complete. I rebuilt the Travis test on this PR and it passed, but the first run failed on `orphanedChannelsAreLogged`. It remains flaky, and depending on what changes are made to other test classes, the flake rate can increase.

dapengzhang0:
@ericgribkoff If we exclude this test from core:test and run it in a separate gradle test task, then maybe no other tests run in the same JVM with it.

ericgribkoff:
No longer needed after https://github.com/grpc/grpc-java/pull/4093

PR #3496, opened 2017-09-25
Promote keepalive to ManagedChannelBuilder; stabilize

### LINKED ISSUE #3493: Need to set keep-alive through ManagedChannelBuilder
For Pub/Sub streaming pull, we need to set keep-alive to prevent connections from being dropped. Unfortunately, this seems to only be available on NettyChannelBuilder (https://github.com/grpc/grpc-java/blob/8ee9fb51837a96b18664241a6875ff155843152b/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java#L278) and not ManagedChannelBuilder. Can we get a way to set this through ManagedChannelBuilder?

ejona86:
Fixes #3493

CC @garrettjonesgoogle

### DIFF
--- a/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
@@ -139,6 +139,24 @@ public abstract class ForwardingChannelBuilder<T extends ForwardingChannelBuilde
     return thisT();
   }
 
+  @Override
+  public T keepAliveTime(long keepAliveTime, TimeUnit timeUnit) {
+    delegate().keepAliveTime(keepAliveTime, timeUnit);
+    return thisT();
+  }
+
+  @Override
+  public T keepAliveTimeout(long keepAliveTimeout, TimeUnit timeUnit) {
+    delegate().keepAliveTimeout(keepAliveTimeout, timeUnit);
+    return thisT();
+  }
+
+  @Override
+  public T keepAliveWithoutCalls(boolean enable) {
+    delegate().keepAliveWithoutCalls(enable);
+    return thisT();
+  }
+
   /**
    * Returns the {@link ManagedChannel} built by the delegate by default. Overriding method can
    * return different value.
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -265,6 +265,54 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   }
 
   /**
+   * Sets the time without read activity before sending a keepalive ping. An unreasonably small
+   * value might be increased, and {@code Long.MAX_VALUE} nano seconds or an unreasonably large
+   * value will disable keepalive. Defaults to infinite.
+   *
+   * <p>Clients must receive permission from the service owner before enabling this option.
+   * Keepalives can increase the load on services and are commonly "invisible" making it hard to
+   * notice when they are causing excessive load. Clients are strongly encouraged to use only as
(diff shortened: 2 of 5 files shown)


ejona86:
@carl-mastrangelo, @zhangkun83, @zpencer, do y'all have any complaints with the keepalive API or my defaulting to throw UnsupportedOperationException?

core/src/main/java/io/grpc/ForwardingChannelBuilder.java
```diff
@@ -139,6 +139,24 @@ public T maxInboundMessageSize(int max) {
     return thisT();
   }
 
+  @Override
```
carl-mastrangelo: An API Alternative: this might be cleaner as some sort of "connectionConfig"  where a user can set each of the three options in a single object.   There are a couple connection specific options that a user might want to set, and it would be nice to not have a method for each one.  Max concurrent streams would be one option.  Another one would be to eagerly create connections, something that the channel state API handles now.

ejona86: That's a substantially different API, and I don't think I'd feel comfortable with it. It would not be extensible on a per-transport basis and we'd be limited to ignoring unknown configuration. I guess it could be a sub-builder for per-transport configuration, but I'm not sure what the added value is. Also, for the user, the distinction between "channel" and "connection" is nebulous.

Max concurrent streams is only on server-side, although that doesn't mean we couldn't mirror the API here. The channel state API was designed to be the place that would trigger eager connection creation, so I'm not excited to adding something equivalent here.

carl-mastrangelo: I know its a different API, though it doesn't seem inherently worse.  The configuration object could be extensible, with each transport providing a child class.  This is the netty pattern, and it works pretty well.

The API in this PR is adding three additional methods that most users will not need to set.  I think we should be careful to keep the MCB API clean and trim.

zhangkun83:
LGTM... Although I would prefer a single `keepAliveSettings()` rather than three separate methods, they are already on Netty and OkHttp builders and won't go away.

ejona86:
> Although I would prefer a single keepAliveSettings()

Hopefully better than `enableKeepAlive`; that's proved to be a very ugly API. It also doesn't support keepAliveWithoutCalls. Part of the problem is that most people shouldn't need to set the keepAliveTimeout.

> they are already on Netty and OkHttp builders and won't go away.

The old methods will go away. The transports are ExperimentalApi.

Note these new methods already existed on the transports; they just got added to the ManagedChannelBuilder.

core/src/main/java/io/grpc/ForwardingChannelBuilder.java
ejona86: > This is the netty pattern, and it works pretty well.

Netty's still a bit different than that because bootstrap has all concrete classes. You have the option to pass an Option to it and you can also change the channel-specific Config after Channel creation.

Going with this approach would have left ManagedChannelBuilder concrete with nobody extending it. While I'm not arguing whether that may have been better, it seems a stark departure to what we have now and unnecessary given we are already extending ManagedChannelBuilder. If you're proposing we get rid of extending ManagedChannelBuilder, I guess that may be possible, but it'd be quite a redesign.

carl-mastrangelo: The reason I am thinking about alternatives is because the API you are adding here is not experimental (and not sure it can be).   I'd be less concerned if this API change wasn't permanent, but since it is I think we need to evaluate alternatives before proceeding.

ejona86:
While there are misgivings about this API, our hands are basically tied due to timelines involved. Merging as-is.

PR #3497, opened 2017-09-25
testing: reduce mocks in AbstractTransportTest to eliminate flakes

ericgribkoff:
The additional mocks that I added in https://github.com/grpc/grpc-java/pull/3145 made at least the `#basicStreams` test flaky: approximately 0.25% of local runs would hit a Mockito `UnfinishedStubbingException`, which I believe was due to the potential for the `doAnswer` blocks to trigger another mocked method. I didn't fully track down the cause, but rather just removed the use of mocks for `ClientStreamListener` and `ServerStreamListener` altogether, and no longer observe failures across many thousands of runs.

This also adds an `#awaitOutboundHeaders` method to `TestClientStreamTracer` (https://github.com/grpc/grpc-java/pull/3305), as without this a small percentage of runs of `#basicStreams` would also fail due to a race between the server receiving the headers and the Netty future on the client invoking the stream tracer context to report headers sent.

### DIFF
--- a/testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
+++ b/testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
@@ -1517,7 +1517,6 @@ public abstract class AbstractTransportTest {
         fail("onReady invoked after closed");
       }
       onReadyLatch.countDown();
-      ;
     }
 
     @Override
@@ -1565,7 +1564,6 @@ public abstract class AbstractTransportTest {
         fail("onReady invoked after closed");
       }
       onReadyLatch.countDown();
-      ;
     }
 
     @Override
--- a/testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
+++ b/testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
@@ -32,11 +32,9 @@ import static org.mockito.Matchers.anyBoolean;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Matchers.same;
-import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.timeout;
 import static org.mockito.Mockito.verify;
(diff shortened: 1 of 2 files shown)


testing/src/main/java/io/grpc/internal/testing/TestClientStreamTracer.java
```diff
@@ -26,6 +27,7 @@
  */
 public class TestClientStreamTracer extends ClientStreamTracer implements TestStreamTracer {
   private final TestBaseStreamTracer delegate = new TestBaseStreamTracer();
+  private final CountDownLatch outboundHeadersLatch = new CountDownLatch(1);
```
carl-mastrangelo: consider making this protected.  The rest of this class is designed to be overridable.  

For example the `outboundHeaders()` method below could be overridden,  but have no way to count down this latch.

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
```diff
@@ -1622,14 +1486,114 @@ public StreamCreation takeStreamOrFail(long timeout, TimeUnit unit)
...
+      }
+    }
+
+    @Override
+    public void onReady() {
+      if (status.isDone()) {
+        fail("onReady invoked after closed");
+      }
+      onReadyLatch.countDown();
+      ;
```
carl-mastrangelo: extra semi

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
```diff
@@ -1622,14 +1486,114 @@ public StreamCreation takeStreamOrFail(long timeout, TimeUnit unit)
...
```
carl-mastrangelo: same here

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
ericgribkoff: Fixed

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
ericgribkoff: Fixed

testing/src/main/java/io/grpc/internal/testing/TestClientStreamTracer.java
ericgribkoff: Done

PR #3501, opened 2017-09-25
netty: update to 4.1.16

carl-mastrangelo:


netty/src/main/java/io/grpc/netty/AbstractHttp2Headers.java
```diff
@@ -26,6 +26,11 @@
 abstract class AbstractHttp2Headers implements Http2Headers {
 
   @Override
+  public Iterator<CharSequence> valueIterator(CharSequence name) {
```
dapengzhang0: Do we have to change Bazel repositories as well?

carl-mastrangelo: Done

carl-mastrangelo:
It would be nice if there was a way to see that bazel would have broken :(

Issue #3506, opened 2017-09-26
Utility class for automatically cancelling a context after an operation


zpencer:
CancellableContext must always be cancelled at the end of the request. Let's add a utility that simplifies the common cases, for example:
- run a Runnable, and then cancel when done or exception thrown
- run a Callable, and then cancel when done or exception thrown
- run a Callable<ListenableFuture<T>>, and cancel when done, or exception thrown, or ListenableFuture cancelled by user

PR #3507, opened 2017-09-26
context: CancellableContext runAndCancel and callAndCancel

zpencer:
This helper utility is needed to migrate some common internal use cases.

ExperimentalApi ticket: https://github.com/grpc/grpc-java/issues/3506


core/src/main/java/io/grpc/Contexts.java
```diff
@@ -149,4 +159,163 @@ public static Status statusFromCancelled(Context context) {
     }
     return status.withCause(cancellationCause);
   }
+
+  /**
+   * A utility for performing an operation inside of a
+   * {@link io.grpc.Context.CancellableContext CancellableContext}, and then cancelling the context
+   * after it is done.
+   *
+   * <p>Note: Because gRPC must be compatible with Java 1.6, this class does not support the
```
ejona86: We should be able to support the idiom even without breaking Java 6 compatibility. I thought extending Closeable sufficed because AutoCloseable is its super-interface.

I think we may want the ListenableFuture helper to also create the CancellableContext so the API works cleanly with http://errorprone.info/bugpattern/MustBeClosedChecker

zhangkun83:
I don't think `grpc-core` is the right place to put these helpers. The internal use cases you are trying to migrate are not using gRPC, thus they shouldn't depend on gRPC.
If it's not for the Guava dependency, `grpc-context` would be the proper place. Or maybe we should just do this internally?

zpencer:
I failed to notice this file also has grpc related stuff and is not purely dealing with Context. My original intention was to create a class in 'core' that only relied on context and guava, and make a separate target internally that exposes only this util class. Externally, people will only be able to use this helper if they include core.

core/src/main/java/io/grpc/Contexts.java
zpencer: Added a new enclosing class (WithCancellableContext) that extends Closeable, and added `@MustBeClosed` to the factory method. From the javadocs, it looks like that annotation can only be used for things that are meant to be used for things initialized in a try-with-resources block. The CancellableContextApplier won't fit with that idiom.

zhangkun83:
Although we can conveniently shove the applier into core, it doesn't really have anything to do with the rest of the core. People who want to use the applier, even if they don't use gRPC, will be forced to pull in the core and its dependencies (most notably, census). The other option is to put the applier in its own subproject, but I am not excited about that either.  I am on the fence...

ejona86:
I was wanting to see everything but callAsync in grpc-context directly. I also strongly question whether the `WithCancellableContext` style applies here. For example, we could have `CancellableContext.callAndCancel()` instead.

Note that the example `WithCancellableContext.with(cancellableContext).run(...)` doesn't actually adhere to `@MustBeClosed`. I do think it would be good for us to have an API that works cleanly with try-with-resources, though.

zpencer:
@ejona86: updated

ejona86:
Since this is forever, I'd like others to review as well. @zhangkun83 @carl-mastrangelo @dapengzhang0 

The only real thing that could be argued though is whether to cancel with the exception if it is caught.

adriancole:
Stil hazy on the use case for this. Maybe in the javadoc suggest why you would use this? Maybe note one of the concrete internal use cases (not what but why.. ex which rpc pattern of use this supports)

Another way to help others understand is to phrase it backwards: what would happen if you didnt cancel after running in a context?

ejona86:
> Another way to help others understand is to phrase it backwards: what would happen if you didnt cancel after running in a context?

It's actually in the Javadoc already that you must cancel CancellableContexts. Basically, if you don't cancel them you'll likely leak memory (the context itself, callbacks, and a deadline's scheduled future).

@zpencer, I'm second-guessing myself now on how to handle exceptions...

zpencer:
@adriancole : FYI https://github.com/grpc/grpc-java/pull/3607

adriancole:
Ack on javadoc existing with good advice. other comments for #3607

On 21 Oct 2017 04:24, "Eric Anderson" <notifications@github.com> wrote:

Another way to help others understand is to phrase it backwards: what would
happen if you didnt cancel after running in a context?

It's actually in the Javadoc already that you must cancel
CancellableContexts. Basically, if you don't cancel them you'll likely leak
memory (the context itself, callbacks, and a deadline's scheduled future).

@zpencer <https://github.com/zpencer>, I'm second-guessing myself now on
how to handle exceptions...

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/grpc/grpc-java/pull/3507#issuecomment-338313562>, or mute
the thread

PR #3511, opened 2017-09-27
core,grpclb: use denser atomics for census

carl-mastrangelo:
These are created per call, so use volatiles to save memory

### DIFF
--- a/grpclb/src/main/java/io/grpc/grpclb/GrpclbClientLoadRecorder.java
+++ b/grpclb/src/main/java/io/grpc/grpclb/GrpclbClientLoadRecorder.java
@@ -23,11 +23,10 @@ import io.grpc.CallOptions;
 import io.grpc.ClientStreamTracer;
 import io.grpc.Metadata;
 import io.grpc.Status;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicLong;
-import javax.annotation.concurrent.GuardedBy;
+import java.util.HashSet;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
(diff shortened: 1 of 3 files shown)


grpclb/src/main/java/io/grpc/grpclb/GrpclbClientLoadRecorder.java
```diff
@@ -36,44 +35,55 @@
...
+  private volatile long callsFinished;
 
   // Specific finish types
-  // Access to it should be protected by lock.  Contention is not an issue for these counts, because
-  // normally only a small portion of all RPCs are dropped.
-  @GuardedBy("this")
-  private HashMap<String, AtomicLong> callsDroppedPerToken = new HashMap<String, AtomicLong>();
-  private final AtomicLong callsFailedToSend = new AtomicLong();
-  private final AtomicLong callsFinishedKnownReceived = new AtomicLong();
+  private final ConcurrentMap<String, Long> callsDroppedPerToken =
```
ejona86: Why this change?

I'm not excited to see a change like this among all the noise, without it being called out.

carl-mastrangelo: Avoids locks, and avoid the extra atomics.  After talking with @zhangkun83 the reason to not use it is because Longs are less dense than AtomicLongs.

grpclb/src/main/java/io/grpc/grpclb/GrpclbClientLoadRecorder.java
```diff
@@ -83,52 +100,55 @@ ClientStats generateLoadReport() {
...
+        .setNumCallsStarted(callsStartedUpdater.getAndSet(this, 0))
+        .setNumCallsFinished(callsFinishedUpdater.getAndSet(this, 0))
+        .setNumCallsFinishedWithClientFailedToSend(callsFailedToSendUpdater.getAndSet(this, 0))
+        .setNumCallsFinishedKnownReceived(callsFinishedKnownReceivedUpdater.getAndSet(this, 0));
+
+    Map<String, LongHolder> localCallsDroppedPerToken = Collections.emptyMap();
     synchronized (this) {
-      savedCallsDroppedPerToken = callsDroppedPerToken;
-      callsDroppedPerToken = new HashMap<String, AtomicLong>();
+      if (!callsDroppedPerToken.isEmpty()) {
```
ejona86: In a noisy CL like this, a change like this which changes the logic is _really_ hard to notice. It should be separate or called out for reviewers.

This change itself looks fine, but it needs to be seen by reviewers.

grpclb/src/main/java/io/grpc/grpclb/GrpclbClientLoadRecorder.java
ejona86: Ah, I missed the fact it swapped from AtomicLong to Long. I mainly saw the @GuardedBy removal and the ConcurrentMap API usages.

grpclb/src/main/java/io/grpc/grpclb/GrpclbClientLoadRecorder.java
carl-mastrangelo: Fair enough.  I was hoping that the behavior was identical, which is why I didn't bring it up.

PR #3512, opened 2017-09-27
Added methods in AbstractStub to identify type of Stub

ramaraochavali:
Currently once a Stub is generated extending from AbstractStub, there is no way to identify whether is a BlockingStub, FutureStub or simple AsnycStub except from the name which is NOT ideal. Identification of Stub, will help enforce any application level checks such as enforcing deadlines for BlockingStub in the enviornment.
Added the following methods in AbstractStub
isBlockingStub
isFutureStub
isAsyncStub
which return false by default. 

The generator code is changed in such a way that the specific method is overridden for that type of stub and returns true. For example for BlockingStub, isBlockingStub is overridden and true is returned.

Issue [3513](https://github.com/grpc/grpc-java/issues/3513)

### DIFF
--- a/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
+++ b/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/BenchmarkServiceGrpc.java
@@ -240,6 +240,11 @@ public final class BenchmarkServiceGrpc {
       return new BenchmarkServiceStub(channel, callOptions);
     }
 
+    @java.lang.Override
+    public boolean isAsyncStub() {
+      return true;
+    }
+
     /**
      * <pre>
      * One request followed by one response.
@@ -320,6 +325,11 @@ public final class BenchmarkServiceGrpc {
       return new BenchmarkServiceBlockingStub(channel, callOptions);
     }
 
+    @java.lang.Override
+    public boolean isBlockingStub() {
+      return true;
+    }
+
     /**
      * <pre>
      * One request followed by one response.
@@ -360,6 +370,11 @@ public final class BenchmarkServiceGrpc {
     protected BenchmarkServiceFutureStub build(io.grpc.Channel channel,
         io.grpc.CallOptions callOptions) {
       return new BenchmarkServiceFutureStub(channel, callOptions);
+    }
+
+    @java.lang.Override
+    public boolean isFutureStub() {
+      return true;
     }
 
     /**
--- a/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/ReportQpsScenarioServiceGrpc.java
+++ b/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/ReportQpsScenarioServiceGrpc.java
@@ -110,6 +110,11 @@ public final class ReportQpsScenarioServiceGrpc {
       return new ReportQpsScenarioServiceStub(channel, callOptions);
     }
 
+    @java.lang.Override
+    public boolean isAsyncStub() {
+      return true;
+    }
+
     /**
      * <pre>
      * Report results of a QPS test benchmark scenario.
@@ -140,6 +145,11 @@ public final class ReportQpsScenarioServiceGrpc {
       return new ReportQpsScenarioServiceBlockingStub(channel, callOptions);
     }
 
+    @java.lang.Override
+    public boolean isBlockingStub() {
+      return true;
+    }
+
     /**
      * <pre>
      * Report results of a QPS test benchmark scenario.
@@ -167,6 +177,11 @@ public final class ReportQpsScenarioServiceGrpc {
     protected ReportQpsScenarioServiceFutureStub build(io.grpc.Channel channel,
         io.grpc.CallOptions callOptions) {
       return new ReportQpsScenarioServiceFutureStub(channel, callOptions);
+    }
+
+    @java.lang.Override
+    public boolean isFutureStub() {
+      return true;
     }
 
     /**
--- a/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/WorkerServiceGrpc.java
+++ b/benchmarks/src/generated/main/grpc/io/grpc/benchmarks/proto/WorkerServiceGrpc.java
@@ -213,6 +213,11 @@ public final class WorkerServiceGrpc {
       return new WorkerServiceStub(channel, callOptions);
     }
 
+    @java.lang.Override
+    public boolean isAsyncStub() {
+      return true;
+    }
+
     /**
      * <pre>
      * Start server with specified workload.
@@ -286,6 +291,11 @@ public final class WorkerServiceGrpc {
       return new WorkerServiceBlockingStub(channel, callOptions);
     }
 
+    @java.lang.Override
+    public boolean isBlockingStub() {
+      return true;
+    }
+
     /**
      * <pre>
      * Just return the core count - unary call
@@ -323,6 +333,11 @@ public final class WorkerServiceGrpc {
     protected WorkerServiceFutureStub build(io.grpc.Channel channel,
         io.grpc.CallOptions callOptions) {
       return new WorkerServiceFutureStub(channel, callOptions);
+    }
+
+    @java.lang.Override
+    public boolean isFutureStub() {
+      return true;
     }
 
     /**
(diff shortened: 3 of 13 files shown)


dapengzhang0:
> return false by default.

I'm not sure whether these APIs should be added in the abstract class rather than adding in the generated code only, but if we do add them, I think `throw new UnsupportedOperationException("not implemented")` should be better than `return false`.

carl-mastrangelo:
@ramaraochavali We have more stringent requirements around stubs than most code.   In this case, the method count increase puts pressure on android code that has limitations around how many methods a jar can have.  Also, it increases the size of the class, which is a second android concern.  

If you can find a way to identify stubs without adding extra methods that would be ideal.

ramaraochavali:
@dapengzhang0 I am proposing to add them to AbstractStub because we can access them easily by just casting to it on any stub instance. I think make sense to throw UnsupportedOperationException instead of returning false. Agree with it

ramaraochavali:
@carl-mastrangelo  I see your point more methods in Stubs. I can think of following options.
**Option 1:** 
1. Add a StubType enum as shown below in AbstractStub
`StubType {
  Blocking,
  Future,
  Async
}`
2. Add one method in AbstractStub - instead of three methods 
`public abstract StubType stubType()`
3. The generated classes override and return the correct `StubType`

**Option 2** - Not my choice
1. Define enum as above
2 Add a `Public final StubType stubType` in AbstractStub
3.Add a new default constructor which takes StubType and sets in
4. Users can access AbstractStub.stubType

Can't think of any other approach with out adding any methods. I think approach 1 is reasonable. Let me know what you think. or let me know if you have any other suggestions.

I am happy to rework the PR based on that

ejona86:
I'm not sure how I feel about this feature, but I'd probably argue for a `String getStubType()` instead of an enum. An enum would prevent/confuse others from creating their stub extending AbstractStub. Another option may be a tagging interface (`BlahBlockingStub implements BlockingStub`).

The deadline requirement seems a bit arbitrary, but given a specific type of application it may make sense. We have seen the want to restrict blocking stubs before, so maybe we limit the feature to detecting blocking stubs.

carl-mastrangelo:
@ramaraochavali  Are you using Android, and if so, Progaurd?  

If not, then realistically we will never change the names of the Stubs (it would break too many people).   I'd be okay with making a static utility to check if a given class matches a name scheme.  It could use Class.name().endsWith, which would be pretty fast.  We could maintain it, so if we change it in the future to always work.

ramaraochavali:
@ejona86 Yes. It is mainly required for detecting Blocking Stubs and enforcing some restrictions.
So from your point of view, couple of options
1.`String getStubType()` -  Make this abstract in AbstractStub and every stub return it's name. Since  there is no bounded, known String names that apps can verify - it  is little difficult. I understand your point about extending Stubs though and agree enum is not right in that case.
2. Since we want to limit it to just blocking stubs, add a `isBlockingStub` in AbstractStub which returns false and only BlockingStub overrides it and returns true. No need to change any other stubs. 
Option 2 seems ok to me.

@carl-mastrangelo  We are not on Android.  I am OK with static utility as well. Where should that reside? Can we have the following methods in that Utility?
`public boolean isBlockingStub()
public boolean isAsyncStub()
public boolean isFutureStub()`
all of them check on `Class.name().endsWith` and return appropriately. We could maintain and change it when names change.

Let me know which is better. I will rework the PR

carl-mastrangelo:
@ramaraochavali The idea is right, but the reason I asked about proguard in my post above is because it will break this.

Progaurd renames classes and breaks this function.  It would only work if you dont use progaurd, but lots of Android apps do.  I was suggested that you write that function in your own code, since you have extra knowledge that you aren't on android.  gRPC doesn't have that knowledge, and thus wouldn't be able to make assumptions about the name.

ramaraochavali:
@carl-mastrangelo I see your point about progaurd.  I was thinking that if the static utility resides in grpc code we could change it when the names change, however rare it may - so that we do not have to worry about the generation part. Seems it is difficult for you to own because it changes in Android. One way I am thinking is, if you own the utility, we can throw UnSupportedOperationException if the platform is Android. For others we can return values. 

@ejona86 If static utility is not an option, are you OK going with Option 2 above just for BlockingStubs?

ramaraochavali:
@ejona86 just following up on this, what do you think we should do about this PR? Here is the summary of options
1.  Add `isBlockingStub()` in AbstractStub and override in Blocking Stub and return `true`
2. Add a `BlockingStub` marker interface and tag it for Blocking Stub in the code generation
3. Add `String getStubtype()` in AbstractStub and override in each of the generated stubs.

Please let me know which option do you prefer. I am OK with any as long as I am able to detect a blocking stub.

ejona86:
Yet another option: have a (Abstract?)BlockingStub that extends AbstractStub, instead of a tagging interface. All blocking stubs extend it. We can do the same with the other types. Unfortunately, nothing is actually provided by the class. (While we could move some things from ClientCalls there, I think that likely just adds confusion without any benefit.)

@ramaraochavali, one thing I don't understand is how you're able to use this. Are you doing the checks at runtime or statically? If at runtime, how do you know what stub instance the client is using?

thelinuxfoundation:

* One or more commits was unable to be associated with a GitHub user. The email used in the Git commit must match a verified email address in a GitHub profile. For more information, please see https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/

_Regards,

ramaraochavali:
@ejona86  We have a light weight ClientBuilder api that allows consumers to build grpc clients. There we enforce certain restrictions like if the stub is a blocking stub, we want to enforce deadlines. Currently, we are getting the class name and checking if its name `endsWith` BlockingStub to determine if it is a blocking stub. Would think better check with instanceof or some other way rather than relying on the name of the class

ejona86:
@ramaraochavali, so the user passes in the stub? Something like this?

```java
class MyClientBuilder {
  public static <T extends AbstractStub<T>> T configure(T stub) {...}
}
```

That doesn't seem quite right because then they'd need a Channel already.

ramaraochavali:
no. they would pass the stub class
`<T extends AbstractStub<T>> T build(@Nonnull Class<T> var1)`
We will instantiate the class using reflection and internally check if it's name ends with '`BlockingStub`' enforce, they have passed the required builder parameters

ejona86:
> We will instantiate the class using reflection

Ah. The constructor has no API guarantees, but I understand the need. I'd also be willing to change that; I don't like that people need to import FooGrpc just to construct the stub. So we could expose the constructors or add factory methods to the stub classes.

Grrr... too many options with no clear winners. So much flip-flopping. Since extending a class is mostly useless, my feeling-of-the-day is now Annotations.

```java
// TODO: figure out if Stub is the name we want, and where it would live
@Stub(serviceClass=FooGrpc.class, style="blocking")
public static final class FooBlockingStub extends io.grpc.stub.AbstractStub<SimpleServiceBlockingStub> {
```

ramaraochavali:
Even If we introduce a factory method in these classes, still we need to import xxxGrpc because these classes are enclosed by that class. Are you thinking of moving them out?

If we go via the Annotation approach, may be we should provide some utilities to determine the type. Thinking how this is better than `String getStubType()` - one benefit of annotation is, do not need to add additional method in `AbstractStub`

ejona86:
> Even If we introduce a factory method in these classes, still we need to import xxxGrpc because these classes are enclosed by that class.

Not really. Today:

```java
import MyServiceGrpc;
import MyServiceGrpc.MyServiceStub;

// What I see most commonly
MyServiceStub stub = MyServiceGrpc.newStub(blah);
// or to avoid one import above
MyServiceGrpc.MyServiceStub stub = MyServiceGrpc.newStub(blah);
```

That style would have been fine if we had named the stubs without the "MyService" prefix, as we would always expect the qualified `MyServiceGrpc.Stub` form.

But since the prefix _is_ there, this would be cleaner:
```java
import MyServiceGrpc.MyServiceStub;

MyServiceStub stub = MyServiceStub.newStub(blah);
```

Most occurrences of the `MyServiceGrpc` import seem just to be to create a stub. There are other methods on `MyServiceGrpc`, but they are for more advanced use cases, so the import isn't as big of a deal there.

ramaraochavali:
Got it. Makes sense. Thanks. So do you want me to try looking at introducing that factory method `newStub` in each of the stubs?
So what about Annotation? What do you think about it?

Issue #2284, opened 2016-09-19
Census with InProcessTransport


zhangkun83:
Message sizes are counted in MessageFramer and MessageDeframer, which InProcessTransport doesn't use. Right now we simply do not record metrics for InProcessTransport at all. This may turn out to be needed, so I use this issue to track this missing feature.

carl-mastrangelo:
IPT skips all other sorts of checks, like Metadata restrictions.

zhangkun83:
In-process transport still needs to propagate the stats and tracing context, and even record message events to `StreamTracer`s as per the contract of the tracer-related APIs.

On the other hand, it probably makes more sense to **not report** the in-process activities to Census, as Census will expose the stats to monitoring interfaces where in most cases only network activity is interesting.

PR #3517, opened 2017-09-28
core: make in-process transport support StreamTracer

zhangkun83:
This is needed for both completeness and stats/tracing contexts propagation.

Stats recording with Census is intentionally disabled (#2284), while the rest of the Census-related logic work the same as on the other transports.


testing/src/main/java/io/grpc/testing/TestUtils.java
```diff
@@ -92,6 +93,23 @@ public static ServerInterceptor recordServerCallInterceptor(
   }
 
   /**
+   * Capture the context seen by the server.  Useful for testing context propagation.
+   */
+  public static ServerInterceptor recordContextInterceptor(
```
ejona86: Let's put this in the one test that uses it. Users shouldn't need this, and it is visible to them here.

PR #3520, opened 2017-09-29
examples: fix Bazel build

dapengzhang0:
Refactor the proto file `helloworld_streaming.proto` because Bazel and Gradle have incompatible base directory for proto imports. Bazel's proto import is relative to WORKSPACE, whereas Gradle proto plugin's is relative to `${sourceSet}/proto/`. In `helloworld_streaming.proto` file, `import helloworld.proto` does not work for Bazel. If `import src/main/proto/helloworld.proto`, that works for Bazel, but Gradle and Maven would fail. Some workarounds are very hacky, so use independent proto without imports instead to avoid this issue.

### DIFF
--- a/examples/BUILD.bazel
+++ b/examples/BUILD.bazel
@@ -17,6 +17,36 @@ java_grpc_library(
 )
 
 proto_library(
+    name = "helloworld_streaming_proto",
+    srcs = [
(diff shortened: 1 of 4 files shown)


dapengzhang0:
retest this please

PR #3521, opened 2017-09-29
Android on-device cache example

ericgribkoff:
This adds an example Android app with on-device caching for safe unary RPCs, implemented as a client interceptor (`SafeMethodCachingInterceptor`). The cache can be backed by an arbitrary map, but is intended to be used with `android.util.LruCache`.


examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
```diff
@@ -0,0 +1,150 @@
...
+          reply =
+              ClientCalls.blockingUnaryCall(
+                  channelToUse, safeCacheableUnaryCallMethod, callOptions, message);
+        } else {
+          GreeterGrpc.GreeterBlockingStub stub = GreeterGrpc.newBlockingStub(channelToUse);
+          reply = stub.sayHello(message);
+        }
+        return reply.getMessage();
+      } catch (Exception e) {
+        StringWriter sw = new StringWriter();
```
carl-mastrangelo: Android uses LOG right?

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
```diff
@@ -0,0 +1,297 @@
...
+              private int maxAge = -1;
+
+              @Override
+              public void onHeaders(Metadata headers) {
+                Iterable<String> cacheControlHeaders = headers.getAll(CACHE_CONTROL_KEY);
+                if (cacheResponse && cacheControlHeaders != null) {
+                  for (String cacheControlHeader : cacheControlHeaders) {
+                    for (String directive : CACHE_CONTROL_SPLITTER.split(cacheControlHeader)) {
+                      if (directive.equalsIgnoreCase("no-cache")) {
+                        cacheResponse = false;
```
carl-mastrangelo: this is scary.   I don't know if I want people copy pasting this .  Doesn't OkHTTP have one you can re appropriate?

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
ericgribkoff: This is creating the string that gets displayed within the app, not writing to logcat. This is the approach used by both of our existing Android example apps: would you like to see something different here?

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
ericgribkoff: From our offline conversation with @ejona86 my takeaway was that this limited cache-control parsing is OK for inclusion in an example app (perhaps with a more prominent indication of what this does and does not handle in the documentation). There are obstacles to re-appropriating the (internal OkHttp API) cache used by OkHttp, so for this basic example I'd prefer to stick with the current manual parsing.

ericgribkoff:
friendly ping :)

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
```diff
@@ -0,0 +1,150 @@
...
+import io.grpc.MethodDescriptor;
+import io.grpc.examples.helloworld.GreeterGrpc;
+import io.grpc.examples.helloworld.HelloReply;
+import io.grpc.examples.helloworld.HelloRequest;
+import io.grpc.stub.ClientCalls;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.concurrent.TimeUnit;
+
+public class ClientCacheExampleActivity extends AppCompatActivity {
```
carl-mastrangelo: final

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
```diff
@@ -0,0 +1,150 @@
...
+    getCheckBox = (CheckBox) findViewById(R.id.get_checkbox);
+    noCacheCheckBox = (CheckBox) findViewById(R.id.no_cache_checkbox);
+    onlyIfCachedCheckBox = (CheckBox) findViewById(R.id.only_if_cached_checkbox);
+    mResultText = (TextView) findViewById(R.id.grpc_response_text);
+    mResultText.setMovementMethod(new ScrollingMovementMethod());
+
+    cache = SafeMethodCachingInterceptor.newLruCache(CACHE_SIZE_IN_BYTES);
+  }
+
+  public void sendMessage(View view) {
```
carl-mastrangelo: should this be public?  Is so, javadoc?

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
```diff
@@ -0,0 +1,150 @@
...
+    ((InputMethodManager) getSystemService(Context.INPUT_METHOD_SERVICE))
+        .hideSoftInputFromWindow(mHostEdit.getWindowToken(), 0);
+    mSendButton.setEnabled(false);
+    new GrpcTask().execute();
+  }
+
+  private class GrpcTask extends AsyncTask<Void, Void, String> {
+    private String mHost;
```
carl-mastrangelo: android style guide says no more `m` prefix

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
```diff
@@ -0,0 +1,150 @@
...
+      mResultText.setText("");
+    }
+
+    @Override
+    protected String doInBackground(Void... nothing) {
+      try {
+        mChannel = ManagedChannelBuilder.forAddress(mHost, mPort).usePlaintext(true).build();
+        Channel channelToUse =
+            ClientInterceptors.intercept(
+                ManagedChannelBuilder.forAddress(mHost, mPort).usePlaintext(true).build(),
```
carl-mastrangelo: this should be mChannel right?

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
```diff
@@ -0,0 +1,150 @@
...
+                SafeMethodCachingInterceptor.newSafeMethodCachingInterceptor(cache));
+        HelloRequest message = HelloRequest.newBuilder().setName(mMessage).build();
+        HelloReply reply;
+        if (getCheckBox.isChecked()) {
+          MethodDescriptor<HelloRequest, HelloReply> safeCacheableUnaryCallMethod =
+              GreeterGrpc.METHOD_SAY_HELLO.toBuilder().setSafe(true).build();
```
carl-mastrangelo: uppercase static field is deprecated now.

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
carl-mastrangelo: Well, I would expect to see LOG, or let the exception propagate up.  I don't think the existing android examples were written with best practices.  

That said, I would have just expected the exception to get thrown.

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
```diff
@@ -0,0 +1,297 @@
...
+ * <p>This interceptor respects the cache-control directives in the server's response: max-age
+ * determines when the cache entry goes stale. no-cache, no-store, and no-transform entirely skip
+ * caching of the response. must-revalidate is ignored, as the cache does not support returning
+ * stale responses.
+ *
+ * <p>Note: other response headers besides cache-control (such as Expiration, Varies) are ignored by
+ * this implementation.
+ */
+final class SafeMethodCachingInterceptor implements ClientInterceptor {
+  public static CallOptions.Key<Boolean> NO_CACHE_CALL_OPTION =
```
carl-mastrangelo: make these constants non public to match the class.

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
```diff
@@ -0,0 +1,297 @@
...
+      CallOptions.Key.of("no-cache", false);
+  public static CallOptions.Key<Boolean> ONLY_IF_CACHED_CALL_OPTION =
+      CallOptions.Key.of("only-if-cached", false);
+
+  public static final class Key {
+    private final String fullMethodName;
+    private final MessageLite request;
+
+    public Key(String fullMethodName, MessageLite request) {
```
carl-mastrangelo: Why not take a MethodDescriptor, and parameterize on the request type from the MD?

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
```diff
@@ -0,0 +1,297 @@
...
+   * Obtain a new cache with a least-recently used eviction policy and the specified size limit. The
+   * backing caching implementation is provided by {@link LruCache}. It is safe for a single cache
+   * to be shared across multiple {@link SafeMethodCachingInterceptor}s without synchronization.
+   */
+  public static Cache newLruCache(final int cacheSizeInBytes) {
+    return new Cache() {
+      private final LruCache<Key, Value> lruCache =
+          new LruCache<Key, Value>(cacheSizeInBytes) {
+            protected int sizeOf(Key key, Value value) {
+              return value.response.getSerializedSize();
```
carl-mastrangelo: micronit: the size from this method doesn't always match the size of the response.  This is particularly true for protos made with proto2 encoders, but parsed with proto3 definitions.  

I don't this its too damaging, but its going to be very subtly off.

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
```diff
@@ -0,0 +1,297 @@
...
+        next.newCall(method, callOptions)) {
+      private Listener<RespT> interceptedListener;
+      private Key requestKey;
+      private boolean cacheResponse = true;
+      private boolean cacheOptionsCausedCancellation;
+
+      @Override
+      public void start(Listener<RespT> responseListener, Metadata headers) {
+        interceptedListener =
+            new ForwardingClientCallListener.SimpleForwardingClientCallListener<RespT>(
```
carl-mastrangelo: optional: there is a ton of indenting in this function.  maybe extract this to a helper class?

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
```diff
@@ -0,0 +1,297 @@
...
+                      } else if (directive.equalsIgnoreCase("no-transform")) {
+                        cacheResponse = false;
+                        break;
+                      } else if (directive.toLowerCase().startsWith("max-age")) {
+                        String[] parts = directive.split("=");
+                        if (parts.length == 2) {
+                          try {
+                            maxAge = Integer.parseInt(parts[1]);
+                          } catch (NumberFormatException e) {
+                            continue;
```
carl-mastrangelo: this seems as least loggable.  If someone misformats the number, they should at least be able to find out about it

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
```diff
@@ -0,0 +1,297 @@
...
+                super.onMessage(message);
+              }
+
+              @Override
+              public void onClose(Status status, Metadata trailers) {
+                if (cacheOptionsCausedCancellation) {
+                  // UNAVAILABLE is the canonical gRPC mapping for HTTP response code 504 (as used
+                  // by the built-in Android HTTP request cache).
+                  super.onClose(
+                      Status.UNAVAILABLE.withDescription(status.getDescription()), new Metadata());
```
carl-mastrangelo: You'll need to preserve the cause too.

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
```diff
@@ -0,0 +1,297 @@
...
+          cacheResponse = false;
+          super.sendMessage(message);
+          return;
+        }
+
+        // Check the cache
+        requestKey = new Key(fullMethodName, (MessageLite) message);
+        Value cachedResponse = internalCache.get(requestKey);
+        if (cachedResponse != null) {
+          if (cachedResponse.maxAgeDeadline.isExpired()) {
```
carl-mastrangelo: if there are no requests for a long time, the cache could build up to be pretty big correct?

examples/android/clientcache/app/src/test/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptorTest.java
```diff
@@ -0,0 +1,533 @@
...
+  @Test
+  public void safeCallsAreCached() {
+    HelloReply reply1 =
+        ClientCalls.blockingUnaryCall(
+            channelToUse, safeGreeterSayHelloMethod, CallOptions.DEFAULT, message);
+    HelloReply reply2 =
+        ClientCalls.blockingUnaryCall(
+            channelToUse, safeGreeterSayHelloMethod, CallOptions.DEFAULT, message);
+
+    assertEquals(reply1, reply2);
```
carl-mastrangelo: assertSame?

examples/android/clientcache/app/src/test/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptorTest.java
```diff
@@ -0,0 +1,533 @@
...
+
+    HelloReply reply1 =
+        ClientCalls.blockingUnaryCall(
+            channelToUse, safeGreeterSayHelloMethod, CallOptions.DEFAULT, message);
+    HelloReply reply2 =
+        ClientCalls.blockingUnaryCall(
+            channelToUse, safeGreeterSayHelloMethod, CallOptions.DEFAULT, message);
+
+    assertNotEquals(reply1, reply2);
+    assertEquals(0, cache.internalCache.size());
```
carl-mastrangelo: use Truth.assertThat(blah).isEmpty()

This will include what was present in the error message.

examples/android/clientcache/app/src/test/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptorTest.java
```diff
@@ -0,0 +1,533 @@
...
+    HelloReply reply1 =
+        ClientCalls.blockingUnaryCall(
+            channelToUse, safeGreeterSayHelloMethod, CallOptions.DEFAULT, message);
+    HelloReply reply2 =
+        ClientCalls.blockingUnaryCall(
+            channelToUse, safeGreeterSayHelloMethod, CallOptions.DEFAULT, message);
+    assertEquals(reply1, reply2);
+
+    // Wait for cache entry to expire
+    Thread.sleep(1001);
```
carl-mastrangelo: needs to be done in a loop, it could sleep for less time.

carl-mastrangelo: Also, since the clock timer is not super accurate, you probably need to sleep for a bit longer.  For example, the windows currentTimeMillis is only accurate to about 16ms.

ejona86: https://github.com/grpc/grpc-java/blob/f9f603570d20d4507f4d90a3dc53915ac7db4d33/testing/src/main/java/io/grpc/internal/testing/TestUtils.java#L171

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
ericgribkoff: Done

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
ericgribkoff: It needs to be public (https://developer.android.com/guide/topics/ui/controls/button.html#HandlingEvents), Javadoc added.

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
ericgribkoff: m's removed

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
ericgribkoff: Fixed

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
ericgribkoff: Updated

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/ClientCacheExampleActivity.java
ericgribkoff: Added log. We'll get an exception here when e.g. the connection is UNAVAILABLE. This is printed to the screen, and now also logged.

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
ericgribkoff: Done

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
ericgribkoff: Would be nicer, but MethodDescriptor doesn't implement equals and reference equality isn't sufficient here. "Current practice" (e.g., in this PR and our tests) allow copying the MethodDescriptor and setting it to safe as a valid means of turning on GET. Added test for this case that fails if we key on MD by reference equality.

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
ericgribkoff: Acknowledged. I think it's impact is minor here

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
ericgribkoff: There's a lot of shared state; refactored to a helper class but it looks cleaner as-is so reverted.

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
ericgribkoff: Logged

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
ericgribkoff: Done

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
ericgribkoff: The LruCache used here has a built-in size limit

examples/android/clientcache/app/src/test/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptorTest.java
ericgribkoff: Done

examples/android/clientcache/app/src/test/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptorTest.java
ericgribkoff: Done

examples/android/clientcache/app/src/test/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptorTest.java
ericgribkoff: Done

examples/android/clientcache/app/src/test/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptorTest.java
carl-mastrangelo: Not done?

ericgribkoff: Hah, oops - I did manage to change one of these, but there were even more of them to be rooted out. Now it's really done.

ericgribkoff:
@carl-mastrangelo Addressed use of cancel() issue discussed in https://github.com/grpc/grpc-java/issues/3725. PTAL

ericgribkoff:
@carl-mastrangelo friendly bump...addressed the cancellation issue discussed in #3725 in the last commit, wanted to get another LGTM before merging. Thanks!

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
```diff
@@ -234,14 +234,11 @@ public void onMessage(RespT message) {
 
               @Override
               public void onClose(Status status, Metadata trailers) {
-                if (cacheOptionsCausedCancellation) {
+                if (cacheOptionsErrorMsg != null) {
```
carl-mastrangelo: this needs to be assigned to a local variable, as it could be different by the time you use it.

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
```diff
@@ -257,8 +254,8 @@ public void sendMessage(ReqT message) {
 
         if (noCache) {
           if (onlyIfCached) {
-            cacheOptionsCausedCancellation = true;
-            super.cancel("Unsatisfiable Request (no-cache and only-if-cached conflict)", null);
+            cacheOptionsErrorMsg = "Unsatisfiable Request (no-cache and only-if-cached conflict)";
+            super.cancel(cacheOptionsErrorMsg, null);
```
carl-mastrangelo: this also needs to be assigned to a local, same reason.

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
```diff
@@ -282,16 +279,17 @@ public void sendMessage(ReqT message) {
         }
 
         if (onlyIfCached) {
-          cacheOptionsCausedCancellation = true;
-          super.cancel("Unsatisfiable Request (only-if-cached set, but value not in cache)", null);
+          cacheOptionsErrorMsg =
```
carl-mastrangelo: and here

examples/android/clientcache/app/src/main/java/io/grpc/clientcacheexample/SafeMethodCachingInterceptor.java
ericgribkoff: It can only be set to a non-null value, so subsequent changes wouldn't really matter here (it will never go back to null). We'd want the latest value anyways.

ericgribkoff:
commenting to trigger cla bot

PR #3529, opened 2017-10-01
netty: add plain-text upgrade to server

xujianhai:
add HttpClientCodec in BufferingHttp2UpgradeHandler  for decode upgrade http request, reference
https://github.com/netty/netty/blob/4.1/example/src/main/java/io/netty/example/http2/helloworld/client/Http2ClientInitializer.java
configureClearText()

### DIFF
--- a/core/src/main/java/io/grpc/ServerBuilder.java
+++ b/core/src/main/java/io/grpc/ServerBuilder.java
@@ -143,12 +143,19 @@ public abstract class ServerBuilder<T extends ServerBuilder<T>> {
    *
    * @param certChain file containing the full certificate chain
    * @param privateKey file containing the private key
-   *
    * @return this
    * @throws UnsupportedOperationException if the server does not support TLS.
    * @since 1.0.0
    */
   public abstract T useTransportSecurity(File certChain, File privateKey);
+
+  /**
+   * Makes the server use CleartextUpgrade.
+   *
+   * @return this
+   * @throws UnsupportedOperationException if the server does not support CleartextUpgrade.
+   */
+  public abstract T usePlainTextUpgrade();
 
   /**
    * Set the decompression registry for use in the channel.  This is an advanced API call and
--- a/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
+++ b/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
@@ -40,8 +40,14 @@ import io.netty.handler.codec.http.DefaultHttpRequest;
 import io.netty.handler.codec.http.HttpClientCodec;
 import io.netty.handler.codec.http.HttpClientUpgradeHandler;
 import io.netty.handler.codec.http.HttpMethod;
+import io.netty.handler.codec.http.HttpServerCodec;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler;
+import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeCodec;
 import io.netty.handler.codec.http.HttpVersion;
+import io.netty.handler.codec.http2.CleartextHttp2ServerUpgradeHandler;
 import io.netty.handler.codec.http2.Http2ClientUpgradeCodec;
+import io.netty.handler.codec.http2.Http2CodecUtil;
+import io.netty.handler.codec.http2.Http2ServerUpgradeCodec;
 import io.netty.handler.proxy.HttpProxyHandler;
 import io.netty.handler.proxy.ProxyConnectionEvent;
 import io.netty.handler.proxy.ProxyHandler;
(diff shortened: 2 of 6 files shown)


core/src/main/java/io/grpc/ServerBuilder.java
```diff
@@ -143,14 +143,21 @@ public T addStreamTracerFactory(ServerStreamTracer.Factory factory) {
...
    */
   public abstract T useTransportSecurity(File certChain, File privateKey);
 
   /**
+   * Makes the server use CleartextUpgrade.
+   *
+   * @return this
+   * @throws UnsupportedOperationException if the server does not support CleartextUpgrade.
+   */
+  public abstract T usePlainTextUpgrade();
```
ejona86: Annotate with `@ExperimentalApi`. All new APIs start with the annotation.

Or, alternatively, just add the new method to `NettyServerBuilder`.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -103,6 +109,47 @@ public AsciiString scheme() {
   }
 
   /**
+   * Create a server plaintextUpgrade handler for gRPC.
+   */
+  public static ProtocolNegotiator serverPlaintextUpgrade() {
+    return new ProtocolNegotiator() {
+      public Handler newHandler(final GrpcHttp2ConnectionHandler handler) {
+        class PlaintextUpgradeHandler extends ChannelHandlerAdapter implements Handler {
+          PlaintextUpgradeHandler() {
```
ejona86: Constructor is unnecessary. Remove.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -103,6 +109,47 @@ public AsciiString scheme() {
...
+                    return new Http2ServerUpgradeCodec(handler);
+                  } else {
+                    return null;
+                  }
+              }
+            };
+
+            HttpServerCodec sourceCodec = new HttpServerCodec();
+            HttpServerUpgradeHandler upgradeCodec = new HttpServerUpgradeHandler(sourceCodec, upgradeCodecFactory);
+            CleartextHttp2ServerUpgradeHandler upgrader = new
```
ejona86: Word-wrap before `new`, not after (so keep `new CleartextHttp2ServerUpgradeHandler(` together on the same line).

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -103,6 +109,47 @@ public AsciiString scheme() {
...
+  public static ProtocolNegotiator serverPlaintextUpgrade() {
+    return new ProtocolNegotiator() {
+      public Handler newHandler(final GrpcHttp2ConnectionHandler handler) {
+        class PlaintextUpgradeHandler extends ChannelHandlerAdapter implements Handler {
+          PlaintextUpgradeHandler() {
+          }
+
+          public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
+
+            final HttpServerUpgradeHandler.UpgradeCodecFactory upgradeCodecFactory = new HttpServerUpgradeHandler.UpgradeCodecFactory() {
```
ejona86: We choose not to use `final` for local variables, unless necessary for closures. It looks like this `final` can be removed.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -103,6 +109,47 @@ public AsciiString scheme() {
...
+                  }
+              }
+            };
+
+            HttpServerCodec sourceCodec = new HttpServerCodec();
+            HttpServerUpgradeHandler upgradeCodec = new HttpServerUpgradeHandler(sourceCodec, upgradeCodecFactory);
+            CleartextHttp2ServerUpgradeHandler upgrader = new
+                CleartextHttp2ServerUpgradeHandler(sourceCodec, upgradeCodec, handler);
+            handler.handleProtocolNegotiationCompleted(Attributes.newBuilder().set(Grpc.TRANSPORT_ATTR_REMOTE_ADDR, ctx.channel().remoteAddress()).build());
+            ctx.channel().pipeline().replace(this, (String)null, upgrader);
```
ejona86: I don't think the `String` cast is necessary. If it's not, remove it.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -329,7 +376,7 @@ public Handler newHandler(GrpcHttp2ConnectionHandler handler) {
       HttpClientCodec httpClientCodec = new HttpClientCodec();
       final HttpClientUpgradeHandler upgrader =
           new HttpClientUpgradeHandler(httpClientCodec, upgradeCodec, 1000);
-      return new BufferingHttp2UpgradeHandler(upgrader);
+      return new BufferingHttp2UpgradeHandler(httpClientCodec, upgrader);
```
ejona86: So client-side Upgrade support was previously broken? That's what it looks like, but wanted to confirm.

ejona86:
@xujianhai, would you be willing to say why you're using plaintext Upgrade? It's a valid thing to do, but most people have been assuming HTTP/2 which avoids a round-trip. It's also likely to run into problems where the upgrade is blocked waiting for the initial request to complete.

The first line of the commit message/issue should be more descriptive. I'd suggest it to be "netty: add h2c upgrade support to server" (or similar)

ejona86:
@xujianhai, what do you want to do with this? The CLA was not signed, so we can't merge as-is. It would be nice for my review comments to be addressed, since they were pretty small. But we'd also need to resolve some merge conflicts because it has been so long since this was created. Note that @raboof is addressing the client-side breakage in #4612.

I'd also still be interested in why you're interested in plaintext Upgrade, just to confirm that you won't hit ugly problems.

thelinuxfoundation:

* User @xujianhai isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

ejona86:
Closing due to inactivity.

PR #3533, opened 2017-10-02
Make AbstractInteropTest runnable in GAE+jdk7

zpencer:
Mocks fail in GAE+jdk7 with this exception, so avoid using mocks.
```
Caused by: java.lang.NoClassDefFoundError: Could not initialize class com.google.apphosting.runtime.security.shared.stub.sun.reflect.ReflectionFactory
	at org.objenesis.instantiator.sun.SunReflectionFactoryInstantiator.<init>(SunReflectionFactoryInstantiator.java:24)
	at org.objenesis.strategy.StdInstantiatorStrategy.newInstantiatorOf(StdInstantiatorStrategy.java:65)
	at org.objenesis.ObjenesisBase.getInstantiatorOf(ObjenesisBase.java:76)
	at org.objenesis.ObjenesisBase.newInstance(ObjenesisBase.java:59)
	at org.mockito.internal.creation.jmock.ClassImposterizer.createProxy(ClassImposterizer.java:128)
	at org.mockito.internal.creation.jmock.ClassImposterizer.imposterise(ClassImposterizer.java:63)
	at org.mockito.internal.creation.jmock.ClassImposterizer.imposterise(ClassImposterizer.java:56)
	at org.mockito.internal.creation.CglibMockMaker.createMock(CglibMockMaker.java:23)
	at org.mockito.internal.util.MockUtil.createMock(MockUtil.java:26)
	at org.mockito.internal.MockitoCore.mock(MockitoCore.java:51)
	at org.mockito.Mockito.mock(Mockito.java:1243)
	at org.mockito.Mockito.mock(Mockito.java:1120)
	at io.grpc.testing.integration.AbstractInteropTest.gracefulShutdown(AbstractInteropTest.java:1213)
	... 30 more
```

Also, `sendsTimeoutHeader` assumes have a server in the same process, which we obviously don't if we're testing against grpc-test.sandbox.googleapis.com:443.

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -38,6 +38,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
+import com.google.common.util.concurrent.SettableFuture;
 import com.google.instrumentation.stats.RpcConstants;
 import com.google.instrumentation.stats.StatsContextFactory;
 import com.google.instrumentation.stats.TagValue;
@@ -102,6 +103,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLSession;
@@ -969,6 +971,7 @@ public abstract class AbstractInteropTest {
 
   @Test(timeout = 10000)
   public void sendsTimeoutHeader() {
+    Assume.assumeTrue("can capture request headers on server side", server != null);
     long configuredTimeoutMinutes = 100;
     TestServiceGrpc.TestServiceBlockingStub stub =
         blockingStub.withDeadlineAfter(configuredTimeoutMinutes, TimeUnit.MINUTES);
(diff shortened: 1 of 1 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -969,6 +971,7 @@ public void exchangeMetadataStreamingCall() throws Exception {
 
   @Test(timeout = 10000)
   public void sendsTimeoutHeader() {
+    Assume.assumeTrue("can not capture server side request headers", server != null);
```
carl-mastrangelo: out of curiosity, when would server be null?

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1208,22 +1211,46 @@ public void gracefulShutdown() throws Exception {
                 .setBody(ByteString.copyFrom(new byte[4])))
             .build());
 
-    @SuppressWarnings("unchecked")
-    StreamObserver<StreamingOutputCallResponse> responseObserver = mock(StreamObserver.class);
+    final ArrayBlockingQueue<StreamingOutputCallResponse> responses =
+        new ArrayBlockingQueue<StreamingOutputCallResponse>(3);
+    final SettableFuture<Void> completed = SettableFuture.create();
+    final AtomicBoolean errorSeen = new AtomicBoolean();
```
carl-mastrangelo: Optional: SettableFuture does what you want.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
zpencer: The server is only non-null if the test subclass uses `startStaticServer` to construct a server from the the local process. Setting it up this way lets the developer capture headers and other info on the server side for test assertions. For GAE, the server is already running on a remote host. This sort of test is not applicable and should be skipped.

PR #3535, opened 2017-10-03
gae-interop-testing: add GAE interop tests

zpencer:
Add gradle projects for running our interop tests in GAE+jdk7 and
GAE+jdk8.

For jdk8 we also make sure we can create a channel and reuse it
across requests.

Some problems discovered:
- ~~These 5 tests fail in both environments, 4 of them seem related to
  the interop server not behaving being as expected, so it may not be
  the fault of gRPC Java.~~ maxInboundSize_exact shows a potential
  problem, while the other 4 are confirmed to be unsupported on the
  interop server process.
  https://gist.github.com/zpencer/0d5c94bbb608c5d3997001e78cc454ff
- Some log spam related to unclean channel shutdowns (jdk7)
  https://gist.github.com/zpencer/c33a1607377abbbaadbb42ed2697feb8
- Some log spam related to creating threads after the request is
  complete. This *may* be an artifact of the test itself. (possibly both)
  https://gist.github.com/zpencer/1dd2382f39266f146191fabcdcb79e7b
  https://gist.github.com/zpencer/e7979d0fa0d5534c3ed887d31e7cc9af
- Unit tests that require lots of memory are silently skipped because
  we use `org.junit.Assume` to check for sufficient memory before
  starting, and if there isn't enough memory, Junit treats the test as
  successful. (jdk7+jdk8)

Let's merge the test runner and file separate issues for the problems.


gae-interop-testing/README.md
```diff
@@ -0,0 +1,50 @@
+Google App Engine interop tests
```
carl-mastrangelo: This dir should not be exported when cutting a release I think.

gae-interop-testing/gae-jdk8/src/main/java/io/grpc/testing/integration/LongLivedChannel.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * A test class that checks we can reuse a channel across requests.
+ *
+ * <p>This servlet communicates with {@code grpc-test.sandbox.googleapis.com}, which is a server
+ * managed by the gRPC team. For more information, see
+ * <a href="https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md">
+ *   Interoperability Test Case Descriptions</a>.
+ */
+public class LongLivedChannel extends HttpServlet {
```
carl-mastrangelo: final

gae-interop-testing/README.md
zpencer: Does anything special need to be done to avoid including this dir?

gae-interop-testing/gae-jdk8/src/main/java/io/grpc/testing/integration/LongLivedChannel.java
zpencer: done

gae-interop-testing/README.md
carl-mastrangelo: Yes, see https://github.com/grpc/grpc-java/commit/4988d8be6b47eb65b3d2c3628dba50c37c991ce6

zpencer:
The above googlebot comment was due to a git command that merged upstream/master to the tip of this branch rather than rebasing. The comment no longer applies now that I did a rebase and force pushed.

PR #3543, opened 2017-10-05
core: allow attributes to be built multiple times

carl-mastrangelo:
Nice to have when using a base set of attributes multiple times.


zhangkun83:
Now I remember the reason why I forbid multiple builds -- to avoid allocating the map twice.
Since you have made so much effort trying to bring down allocations, you might want to consider using `newBuilder(Attributes)` for your use case.

carl-mastrangelo:
@zhangkun83 good point.  I change it to avoid the extra alloc.

PR #3544, opened 2017-10-05
core: add type param to attributes key

carl-mastrangelo:


ejona86:
Can you provide more justification for this change?

carl-mastrangelo:
The original reason was due a java style suggestion that the param should be derived from the argument type.  Key<T> narrowly avoids the original reason for the suggestion.

The reason for the change now is that it makes for a nicer toString.  If two keys have the same name, it will be easier to distinguish them (assuming they have different types).

ejona86:
If you're having trouble with name collisions with key descriptions, why not use a different descriptor?

I'd have expected confusion to only really happen when "I made two instances of the same key yet they aren't equal" and this doesn't address that case. So I'm not really seeing this change worth it as it also introduces problems with generics for container classes.

PR #3549, opened 2017-10-06
Revert core/compiler: register Span names for code-generated methods

zhangkun83:
This reverts commit a6653bb135ccf8352815f41036dd6cdf76f38fb8.

Census will provide a better API that doesn't require global registration, switching to it will slightly change the semantics of setRegisterForTracing(). For the sake of API stability, we exclude the current implementation from the 1.7.0 release.


zhangkun83:
retest this please

PR #3551, opened 2017-10-06
Backport Bazel build fixes to 1.7.x

ejona86:


dapengzhang0:
Also backport https://github.com/grpc/grpc-java/commit/f56be5423e5c901611be3dcdc4fc335c92bb48be

ejona86:
I actually did and then undid it because it is non-essential and just a cleanup. It doesn't seem like anyone would benefit from the backport.

PR #3552, opened 2017-10-06
core: add first-class drop support in LoadBalancer.

### LINKED ISSUE #3355: GRPCLB: drop doesn't work with wait-for-ready RPCs
`GrpclbLoadBalancer` may drop (fail immediately) certain RPCs in order to shed load for the servers. It is done by returning a `PickResult.withError()`. However, `withError()` only fail fail-fast (which is default) RPCs, and will buffer wait-for-ready RPCs, because it is meant to signal transient failures. "Dropped" RPCs should fail immediately regardless of wait-for-ready, and should not be conflated with transient failures.

Options considered:
1. Add `createFailingSubchannel()` to `Helper` to be used for dropped RPCs. This would work but feel awkward, because nearly all methods on `Subchannel` don't apply to the failing Subchannel.
2. Add a new type of `PickResult` in addition to `withError()`, and will fail the RPC regardless of its wait-for-ready option. This sounds like a better option.

/cc @ejona86

zhangkun83:
Resolves #3355

GrpclbLoadBalancer should use PickResult.withDrop() instead of
PickResult.withError() when dropping RPCs.

### DIFF
--- a/core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
+++ b/core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
@@ -653,6 +653,65 @@ public class ManagedChannelImplTest {
         any(Metadata.class), any(CallOptions.class));
   }
 
+  @Test
+  public void testFailFastRpcFailFromErrorFromBalancer() {
+    subtestFailRpcFromBalancer(false, false, true);
+  }
+
+  @Test
+  public void testFailFastRpcFailFromDropFromBalancer() {
+    subtestFailRpcFromBalancer(false, true, true);
+  }
+
+  @Test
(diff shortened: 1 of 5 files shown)


core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -653,6 +653,65 @@ public void firstResolvedServerFailedToConnect() throws Exception {
         any(Metadata.class), any(CallOptions.class));
   }
 
+  @Test
+  public void testFailFastRpcFailFromErrorFromBalancer() {
```
carl-mastrangelo: junit tests should not begin with "test".  It is redundant.

PR #3554, opened 2017-10-07
Use Java 9 ALPN

jjteo74:
Netty `4.1.16` supports Java 9 ALPN, gRPC should take advantage of that to ease deployment (especially on platforms without `tcnative` and/or `openssl`).

This PR determines whether Java 9 ALPN and Netty `4.1.16` is available and enables Netty Java 9 ALPN.


jjteo74:
I signed it!

Teo Jia Jiun
jjteo74@gmail.com

PR #3555, opened 2017-10-07
Use java 9 ALPN if available

jjteo74:
Netty `4.1.16` supports Java 9 ALPN, gRPC should take advantage of that to ease deployment (especially on platforms without tcnative and/or openssl).

This PR determines whether Java 9 ALPN and Netty 4.1.16 is available and enables Netty Java 9 ALPN.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/JettyTlsUtil.java
+++ b/netty/src/main/java/io/grpc/netty/JettyTlsUtil.java
@@ -19,7 +19,6 @@ package io.grpc.netty;
 import java.lang.reflect.Method;
 import java.security.AccessController;
 import java.security.PrivilegedExceptionAction;
-
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLEngine;
 
@@ -27,7 +26,6 @@ import javax.net.ssl.SSLEngine;
  * Utility class for determining support for Jetty TLS ALPN/NPN.
  */
 final class JettyTlsUtil {
-
   private JettyTlsUtil() {
   }
 
@@ -56,7 +54,6 @@ final class JettyTlsUtil {
         return t;
       }
     }
-
   }
 
   /**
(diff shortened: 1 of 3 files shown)


ejona86:
okay to test

ejona86:
This is related to #3401 in some ways, because it revamps how we handle Providers.

netty/src/main/java/io/grpc/netty/JettyTlsUtil.java
```diff
@@ -67,4 +96,17 @@ static synchronized Throwable getJettyNpnUnavailabilityCause() {
     }
     return jettyNpnUnavailabilityCause;
   }
+
+  /**
+   * Indicates whether Netty Java 9 ALPN is available.
+   */
+  static boolean isNettyJava9AlpnAvailable() {
+    try {
+      Class.forName("io.netty.handler.ssl.Java9SslEngine");
```
ejona86: We don't tend to do this sort of thing. We instead just upgrade to the netty that supports the feature we need.

netty/src/main/java/io/grpc/netty/JettyTlsUtil.java
```diff
@@ -16,10 +16,39 @@
...
+      SSLEngine engine = context.createSSLEngine();
+      getApplicationProtocol =
+          AccessController.doPrivileged(new PrivilegedExceptionAction<Method>() {
+            @Override
+            public Method run() throws Exception {
+              return SSLEngine.class.getMethod("getApplicationProtocol");
+            }
+          });
+      getApplicationProtocol.invoke(engine);
+    } catch (Throwable t) {
```
ejona86: Could we just catch Exception, or similar? I'd rather not catch every Error and throw it away, as it can hide legitimate problems.

ejona86:
okay to test

thelinuxfoundation:

* User @jjteo74 isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,

jjteo74:
CLA signed

ejona86:
Thank you!

PR #3559, opened 2017-10-09
Added GrpcContextRule for testing context related code

rmichela:
`GrpcContextRule` is a JUnit `TestRule` that forcibly resets the gRPC `Context` to `Context.ROOT` between every unit test.

This rule makes it easier to correctly implement correct unit tests by preventing the accidental leakage of context state between tests.


adriancole:
Only suggestion is to restore the context after the statement as usually mistakes breaks other tests who might not run this rule. Hunting down which test leaked (in the case they arent leak detecting) isnt fun.

zpencer:
The io.grpc.Context.Storage API allows implementations to define what the default context is, and it is not necessarily Context.ROOT:
https://github.com/grpc/grpc-java/blob/master/context/src/main/java/io/grpc/Context.java#L114

It is legal for a unit test to depend on a context storage implementation that has its own special root context. To allow that, the rule's constructor can read the current context before any tests are run and stash it. For each test, we can try to validate that the current context before and after the test are the same. If they are not the same, then we can forcibly attach the storage specific root context.

rmichela:
@zpencer 
I think I captured the intent of your comments about Context.Storage implementations overriding what ROOT means. Please take a look.

Out of curiosity, are alternate Context.Storage implementations a thing? Googling didn't turn up any results besides the implementation bundled with gRPC.

ejona86:
We're using a custom storage implementation in google to migrate from a pre-existing context. Others could easily do the same. Also, it could allow integrating with platform-specific contexts (say, J2EE's request scope or AppEngine's context).

Instead of doing anything fancy, can this rule:
1. Fail at end of test if context is different (unconditionally; it is not acceptable to fail to restore the context)
2. Always restore the context (simply by detaching)

This is effectively what you have in apply today, but instead of using `root`, I think you'd just use `Context.current()` at the time apply is called.

Note: Context.detach will already do the log.severe. Note that you generally won't see it though because logging has long been broken in JUnit runners. (I dealt with it [long ago](https://github.com/googlegsa/library/commit/344e643af4f06a0ed2efa13729a1362a367a7427) in Ant, and I think virtually the same code is used with Maven and Gradle; I've seen that it is still broken in Gradle; I really should have tried to get my workaround merged upstream...).

thelinuxfoundation:

* User @rmichela isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,

ejona86:
@rmichela, [the CLA has changed](https://groups.google.com/d/topic/grpc-io/8U6-f4LRfNA/discussion). You'll need to deal with the new one.

rmichela:
I signed it!

ejona86:
okay to test

rmichela:
@ejona86 I'm not clear on what is left to do on this PR. Do I need to make changes?

ejona86:
In my previous comment I'd mentioned some things: https://github.com/grpc/grpc-java/pull/3559#issuecomment-340124576

But it effectively is just change the evaluate to:
```java
    public void evaluate() throws Throwable {
        // Make sure tests restore the context
        Context context = Context.current();
        Context prev = current.attach();
        try {
          base.evaluate();
          if (Context.current() != current) {
              Assert.fail("Test is leaking context state between tests! Ensure proper "
                      + "attach()/detach() pairing.");
          }
        } finally {
          context.detach(prev);
        }
      }
```

I will note that `Context.current() != current` may be broken in some storage implementations, but we'll try to fix that if it becomes a problem.

rmichela:
@ejona86 GrpcContextRule changes made as requested.

ejona86:
@rmichela, when would someone use `failOnLeak == false`? They frequently won't see the log statement. At the very least, I'd hope `failOnLeak == true` was the default. But I'd also like to know when it's a good idea to disable.

rmichela:
@ejona86 The original idea for `GrpcContextRule` was to codify the pattern used in the existing context-related unit tests. See: https://github.com/grpc/grpc-java/blob/255643bed94c5a763d1228d4796b0bc7d85b27c4/context/src/test/java/io/grpc/ContextTest.java#L92

I want a way to guarantee a clean context for each test case because I am testing code that depends on context contents. Detecting and failing on context leaks is also important, but was not my initial aim. 

I context leak detector and a context test initializer serve different purposes. Perhaps we need two different test rules?

ejona86:
@rmichela, ah. I was going to delete that code and with the Rule style I was suggesting. :smile: That code in the ContextTest I consider a hack and would rather just guarantee that the tests don't mess up the Context.

But I could see how it might be useful to have a pre-chosen Context for all the tests. Normally you'd rely on `@Before` and `@After`, but that feels weird/wrong for Context. So a Rule does make sense. I do think the user should provide the Context they want, and the code doesn't care whether that is ROOT or something else.

I'm not sure about names/API, though. I feel like mutation can be confusing because if you mutate the Rule during the test it has no effect (or a wrong effect). And any mutation should work more like ExpectedException and TemporaryFolder instead. So I think that leaves: two separate rules, builder, using a `with*()` method, or static factory methods (one to create and use whatever is ambient, one to create and use a specific Context). I think my preference is toward factory methods or the `with*()`, with my suspicion others would probably favor factory methods.

So what about `GrpcContextRule.verify()` and `GrpcContextRule.runWith(Context)` (or runUnder? or withScope?)? Both would verify. I'm not at all sold on this exact API, but it's at least a starting point.

PR #3565, opened 2017-10-10
Release 1.7.0

dapengzhang0:


zpencer:
General question: Do we always use the latest available protoc available when we do a release?

ejona86:
Before merging this we should make sure that we didn't miss any backports and determine what's happening with the unfixed issues in the [1.7 milestone](https://github.com/grpc/grpc-java/milestone/20) (normally, it's to just move it to 1.8, but we may notice we've mixed something important).

We don't update protobuf on the release branch, unless there is a strong reason to. So if it wasn't updated on master before the branch cut then it's likely to stay slightly old.

dapengzhang0:
@zpencer I think we may try the latest available protoc in master before cutting the branch, but when we do a release/make a tag, just use the same protoc as in the branch.

dapengzhang0:
made a mistake

PR #3568, opened 2017-10-11
core: Remove deprecated load balancer and name resolver APIs

carl-mastrangelo:


core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -511,7 +460,7 @@ public void updateOobChannelAddresses(ManagedChannel channel, EquivalentAddressG
      */
     public void updateBalancingState(
         @Nonnull ConnectivityState newState, @Nonnull SubchannelPicker newPicker) {
-      updatePicker(newPicker);
+      throw new UnsupportedOperationException();
```
zhangkun83: ditto

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -128,20 +106,9 @@ public void handleResolvedAddresses(
...
-    for (EquivalentAddressGroup eag : servers) {
-      ResolvedServerInfoGroup.Builder serverInfoGroupBuilder =
-          ResolvedServerInfoGroup.builder(eag.getAttributes());
-      for (SocketAddress addr : eag.getAddresses()) {
-        serverInfoGroupBuilder.add(new ResolvedServerInfo(addr));
-      }
-      serverInfoGroups.add(serverInfoGroupBuilder.build());
-    }
-    handleResolvedAddresses(serverInfoGroups, attributes);
+    throw new UnsupportedOperationException();
```
zhangkun83: This should be made abstract, since it must be implemented.

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -100,26 +98,6 @@
...
-   * @deprecated Implement {@link #handleResolvedAddressGroups} instead.  As it is deprecated, the
-   *             {@link ResolvedServerInfo}s from the passed-in {@link ResolvedServerInfoGroup}s
-   *             lose all their attributes.
-   *
-   * @param servers the resolved server addresses, never empty.
-   * @param attributes extra metadata from naming system.
-   */
-  @Deprecated
-  public void handleResolvedAddresses(
-      List<ResolvedServerInfoGroup> servers, Attributes attributes) {
```
zhangkun83: ResolvedServerInfoGroup should also be removed.

carl-mastrangelo: Done

core/src/main/java/io/grpc/LoadBalancer.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/LoadBalancer.java
carl-mastrangelo: Done

core/src/main/java/io/grpc/LoadBalancer.java
zhangkun83: Not done yet?

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -450,9 +419,8 @@ public boolean equals(Object other) {
      * @throws IllegalArgumentException if {@code subchannel} was not returned from {@link
      *     #createSubchannel}
      */
-    public void updateSubchannelAddresses(Subchannel subchannel, EquivalentAddressGroup addrs) {
-      throw new UnsupportedOperationException();
-    }
+    public abstract void updateSubchannelAddresses(
+        Subchannel subchannel, EquivalentAddressGroup addrs);
```
zhangkun83: This change is unrelated to the deletion of the deprecated methods, and it doesn't seem necessary.

carl-mastrangelo: Sorry, abstracted the wrong method

PR #3570, opened 2017-10-11
core: add drop to hashCode() and equals() for PickResult

zhangkun83:


zhangkun83:
@carl-mastrangelo, test added.

PR #3579, opened 2017-10-13
util: improve scalabiltiy of RR load balancer

carl-mastrangelo:
In relative order of importance:

* Don't acquire lock when picking subchannel.
* Use O(1) lookup for updating channel state
* Use non synchronized ref instead of AtomicReference
* Dont store size in picker.
* make class final
* remove test that was not valid

### DIFF
--- a/core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
+++ b/core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
@@ -37,7 +37,6 @@ import io.grpc.NameResolver;
 import io.grpc.Status;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -76,15 +86,17 @@ public LoadBalancer newLoadBalancer(LoadBalancer.Helper helper) {
   }
 
   @VisibleForTesting
-  static class RoundRobinLoadBalancer extends LoadBalancer {
+  static final class RoundRobinLoadBalancer extends LoadBalancer {
+    @VisibleForTesting
+    static final Attributes.Key<Ref<ConnectivityStateInfo>> STATE_INFO =
+        Attributes.Key.of("state-info");
+    private static final Attributes.Key<Ref<EquivalentAddressGroup>> ADDRESS_GROUP =
```
zhangkun83: You don't need this, because `Subchannel.getAddresses()` is exactly what you need.

carl-mastrangelo: Done

Issue #2169, opened 2016-08-17
Provide hook for network availability notification on Android


HugeTimor:
I new a stub with a multiplexed channel like this:

``` java
ServiceGrpc.ServiceBlockingStub stub = ServiceGrpc.newBlockingStub(getChannel());
```

``` java
 public static ManagedChannel getChannel() {
        if (mChannel == null) {
            mChannel = ManagedChannelBuilder.forAddress(Config.HOST, Config.PORT)
                    .usePlaintext(true)
                    .build();
        }
        return mChannel;
    }
```

Launch the application when the network is close , then open the network . Invoking method of stub that is built with a multiplex  don`t work immediately. While invoking method of stub that built with new channel  can get response from server immediately . It will happen also when the network was unavailable long time .and then network turn available .

How does channel judge network state ？How to reconnect immediately with multiplexed channel ?

Build and platform infomation:

```
            compileSdkVersion: 23,
            buildToolsVersion: "23.0.2",
            minSdkVersion    : 14,
            targetSdkVersion : 19,
```

application is running in Android 5.1.1 and 6.0.1 

Thanks !

ejona86:
There is [exponential backoff](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md) in play. We'd like to provide a hook that allows interacting with Android's network notification, but that isn't available right now.

CC @zsurocking

HugeTimor:
Thanks! I new a channel each time to avoid exponential backoff.  And there are other problems confuse me when i new channel ：

When i new a channel ，is it reference to same transport ? Dose the approach has a performance overhead？

Will the hook that allows interacting with Android's network notification provide in v1.0.x?

zsurocking:
It will create a new transport and a new connection. And yes, it has performance overhead.
I don't think we have any plans yet. It's not likely this quarter though.

A quick workaround might be monitoring the network yourself and starts a new channel when wifi connects.

HugeTimor:
Thanks your help .

bubenheimer:
Having this functionality would be immensely useful for my Android use case as well. Have implementation plans become more concrete?

Could there be an option to offer a more aggressive backoff strategy for Android/OkHttp in the meantime? Is backoff carried over to future connection attempts right now, instead of resetting the backoff to 0 on success?

[...]

PR #3580, opened 2017-10-13
core: add resetConnectBackoff() method to ManagedChannel

ericgribkoff:
This addresses https://github.com/grpc/grpc-java/issues/2169. Currently users of OkHttp may have to wait up to 60 seconds (for DNS resolution to be re-attempted) or longer (if in exponential backoff to reconnect after `TRANSIENT_FAILURE`) for RPCs to succeed after the network connection is re-established. This provides a hook for apps to tell gRPC that the network is available and name resolution and reconnection attempts may occur immediately.

### DIFF
--- a/core/src/main/java/io/grpc/ManagedChannel.java
+++ b/core/src/main/java/io/grpc/ManagedChannel.java
@@ -104,4 +104,22 @@ public abstract class ManagedChannel extends Channel {
   public void notifyWhenStateChanged(ConnectivityState source, Runnable callback) {
     throw new UnsupportedOperationException("Not implemented");
   }
+
+  /**
+   * Notifies the channel that a reconnect attempt should be made. This will attempt to invoke
+   * {@link NameResolver#refresh} and short-circuit the backoff timer for subchannel reconnection
+   * attempts.
+   *
+   * <p>This is primarily intended for Android users, where the network may experience frequent
+   * temporary drops. Rather than waiting for gRPC's name resolution and reconnect timers to elapse
+   * before reconnecting, the app may use this method as a mechanism to notify gRPC that the network
+   * is now available and a reconnection attempt may occur immediately.
+   *
+   * @throws UnsupportedOperationException if not supported by implementation
+   * @since 1.8.0
+   */
+  @ExperimentalApi
+  public void reconnectNow() {
+    throw new UnsupportedOperationException("Not implemented");
+  }
 }
--- a/core/src/main/java/io/grpc/ManagedChannel.java
+++ b/core/src/main/java/io/grpc/ManagedChannel.java
@@ -106,20 +106,19 @@ public abstract class ManagedChannel extends Channel {
   }
 
   /**
-   * Notifies the channel that a reconnect attempt should be made. This will attempt to invoke
-   * {@link NameResolver#refresh} and short-circuit the backoff timer for subchannel reconnection
-   * attempts.
+   * Notifies the channel that a reconnect attempt should be made. The implementation may attempt to
+   * invoke {@link NameResolver#refresh} and short-circuit the backoff timer for subchannel
+   * reconnection attempts.
    *
    * <p>This is primarily intended for Android users, where the network may experience frequent
    * temporary drops. Rather than waiting for gRPC's name resolution and reconnect timers to elapse
    * before reconnecting, the app may use this method as a mechanism to notify gRPC that the network
    * is now available and a reconnection attempt may occur immediately.
    *
-   * @throws UnsupportedOperationException if not supported by implementation
+   * <p>No-op if not supported by the implementation.
+   *
    * @since 1.8.0
    */
   @ExperimentalApi
-  public void reconnectNow() {
-    throw new UnsupportedOperationException("Not implemented");
-  }
+  public void reconnectNow() {}
 }
--- a/core/src/main/java/io/grpc/ManagedChannel.java
+++ b/core/src/main/java/io/grpc/ManagedChannel.java
@@ -120,5 +120,5 @@ public abstract class ManagedChannel extends Channel {
    * @since 1.8.0
    */
   @ExperimentalApi
-  public void reconnectNow() {}
+  public void resetConnectBackoff() {}
 }
--- a/core/src/main/java/io/grpc/ManagedChannel.java
+++ b/core/src/main/java/io/grpc/ManagedChannel.java
@@ -106,9 +106,8 @@ public abstract class ManagedChannel extends Channel {
   }
 
   /**
-   * Notifies the channel that a reconnect attempt should be made. The implementation may attempt to
-   * invoke {@link NameResolver#refresh} and short-circuit the backoff timer for subchannel
-   * reconnection attempts.
+   * For subchannels that are in TRANSIENT_FAILURE state, short-circuit the backoff timer and make
+   * them reconnect immediately. May also attempt to invoke {@link NameResolver#refresh}.
    *
    * <p>This is primarily intended for Android users, where the network may experience frequent
    * temporary drops. Rather than waiting for gRPC's name resolution and reconnect timers to elapse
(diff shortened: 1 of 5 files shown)


dapengzhang0:
+ @zhangkun83

core/src/main/java/io/grpc/ManagedChannel.java
```diff
@@ -104,4 +104,22 @@ public ConnectivityState getState(boolean requestConnection) {
...
+   * <p>This is primarily intended for Android users, where the network may experience frequent
+   * temporary drops. Rather than waiting for gRPC's name resolution and reconnect timers to elapse
+   * before reconnecting, the app may use this method as a mechanism to notify gRPC that a
+   * reconnection attempt may occur immediately.
+   *
+   * @throws UnsupportedOperationException if not supported by implementation
+   * @since 1.8.0
+   */
+  @ExperimentalApi
+  public void networkAvailable() {
```
zhangkun83: This method should be named in an imperative tone like the other methods. Maybe "reconnectNow"?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -627,6 +631,26 @@ public void run() {
         }).drain();
   }
 
+  @Override
+  public void networkAvailable() {
+    channelExecutor.executeLater(
+        new Runnable() {
+          @Override
+          public void run() {
+            if (!nameResolverStarted || shutdown.get()) {
```
zhangkun83: nameResolverStarted should be reset to false when nameResolver is shutdown.

You are skipping the subchannels for the case `!nameResolverStarted` because we know that the channel is in idle mode and all subchannels are shut down. Because it won't harm to call the subchannels anyway, I would prefer not relying on that implication, which is not obvious at first glance. This would be easier to reason about:

```java
if (shutdown.get()) {
  return;
}
if (nameResolverStarted) {
  nameResolver.refresh();
}
for (...
...
```

thelinuxfoundation:

* User @ericgribkoff isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,

ericgribkoff:
I signed it

core/src/main/java/io/grpc/ManagedChannel.java
ericgribkoff: Done

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java

ericgribkoff:
@zhangkun83 Addressed comments, and rebased onto master to fix the merge conflicts that had popped up in the last two weeks

core/src/main/java/io/grpc/ManagedChannel.java
```diff
@@ -104,4 +104,22 @@ public ConnectivityState getState(boolean requestConnection) {
...
+   * temporary drops. Rather than waiting for gRPC's name resolution and reconnect timers to elapse
+   * before reconnecting, the app may use this method as a mechanism to notify gRPC that the network
+   * is now available and a reconnection attempt may occur immediately.
+   *
+   * @throws UnsupportedOperationException if not supported by implementation
+   * @since 1.8.0
+   */
+  @ExperimentalApi
+  public void reconnectNow() {
+    throw new UnsupportedOperationException("Not implemented");
```
zhangkun83: Since this is only advisory, it's safe to use no-op as the default implementation.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -253,12 +257,32 @@ public void run() {
...
         new LogExceptionRunnable(new EndOfCurrentBackoff()),
         delayNanos,
         TimeUnit.NANOSECONDS);
   }
 
+  /**
+   * Immediately attempt to reconnect if the current state is TRANSIENT_FAILURE. Otherwise this
+   * method has no effect.
+   */
+  void networkAvailable() {
```
zhangkun83: Rename this too

core/src/main/java/io/grpc/ManagedChannel.java
ericgribkoff: Done

core/src/main/java/io/grpc/internal/InternalSubchannel.java
ericgribkoff: Oops, fixed

ejona86:
API discussion notes:

ejona86:
API discussion notes
Names:
networkAvailable(). Sort of wrong; should be imperitive. respondToNetworkAvailable()
connectNow(). Doesn't force a connection though. connectNowIfNecessary(). verbose.
reconnectNow() is a bit confusing when we are already connected, since we won't reconnect. reconnectNowIfNecessary()
resetBackoff()/resetConnectBackoff()
ensureConnectivity(). Doesn't force a connect (getState() does that).

Verdict: +5 votes for resetConnectBackoff() "So far the best"

ericgribkoff:
Renamed to `resetConnectBackoff()`. Also changed`InternalSubchannel#cancelReconnectTask` and added test coverage to ensure that the backoff timer is actually reset, for the case where the reconnection attempt triggered by `resetConnectBackoff()` itself fails (thanks @zpencer for asking about this case).

ericgribkoff:
From @ejona86's comment on #2169:

> The API in #3580 may be a bit annoying for applications to use, as users will have to plumb the notification through their code to all channels. We may consider alternative approaches.

One alternative discussed yesterday was a global `resetAllConnectBackoff()` method that would reset the connect backoffs for all channels created by the app. I am not opposed to this type of global method, but I don't think we need to go down that route until we have user feedback suggesting it's necessary. In gRPC right now, users must not only keep a reference to all open channels when handling network switches, but also keep track of the creation parameters for each one so that the channel can be shutdown and recreated when the network comes back. With this PR, only the reference to the channel is necessary, and this is, I think, strictly an improvement and not *that* onerous - particularly since it seems like the per-channel API method added in this PR does not preclude also offering the `resetAllConectBackoff()` helper method in the future.

Additionally (and this may or may not be a compelling use case), I can imagine situations where it's desirable to reset the connection on some channels but not all. Some local services may become available when the device connects to a local network without internet connectivity, but there'd be no need to also reset the connect backoff for connections to external hosts. This per-channel functionality is much more straightforward with the API here than it would be with a global `resetAllConnectBackoff(exceptTheseChannels)` type of API.

core/src/test/java/io/grpc/internal/InternalSubchannelTest.java
```diff
@@ -884,6 +884,64 @@ public ProxyParameters proxyFor(SocketAddress targetServerAddress) {
...
+
+  @Test
+  public void resetConnectBackoff_noopOnIdleTransport() throws Exception {
+    SocketAddress addr = mock(SocketAddress.class);
+    createInternalSubchannel(addr);
+    assertEquals(IDLE, internalSubchannel.getState());
+
+    internalSubchannel.resetConnectBackoff();
+
+    assertNoCallbackInvoke();
```
zhangkun83: Probably want to also verify that no tasks are scheduled on fakeClock.

core/src/main/java/io/grpc/ManagedChannel.java
```diff
@@ -104,4 +104,21 @@ public ConnectivityState getState(boolean requestConnection) {
   public void notifyWhenStateChanged(ConnectivityState source, Runnable callback) {
     throw new UnsupportedOperationException("Not implemented");
   }
+
+  /**
+   * Notifies the channel that a reconnect attempt should be made. The implementation may attempt to
```
zhangkun83: Please also update the wording here. May be remove the first sentence and organize the rest as:
`For subchannels that are in TRANSIENT_FAILURE status, short-circuit the backoff timer and make them reconnect immediatly . May also attempt to invoke {@link NameResolver#refresh}`

ericgribkoff: Done

core/src/test/java/io/grpc/internal/InternalSubchannelTest.java
ericgribkoff: This is already done (for all tests) in the `@After noMorePendingTasks()` method

PR #3588, opened 2017-10-18
core: refresh name resolution when an OOB connection is closed.

zhangkun83:
This is required by an internal use case.  We have already been doing
so for Subchannels.


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -639,6 +639,13 @@ public void run() {
       this.nr = checkNotNull(nr, "NameResolver");
     }
 
+    // Must be called from channelExecutor
+    private void handleInternalSubchannelState(ConnectivityStateInfo newState) {
+      if ((newState.getState() == TRANSIENT_FAILURE || newState.getState() == IDLE)) {
```
carl-mastrangelo: remove double parens

PR #3590, opened 2017-10-19
kokoro: add a placeholder script for GAE testing

zpencer:
This empty script should return with a success immediately, so that the GAE kokoro config is effectively a NOOP for master branch.
After this, I can experiment with creating/deploying/destroying GAE apps with a private branch by modifying `gae.sh`.


buildscripts/kokoro/gae.cfg
```diff
@@ -0,0 +1,10 @@
+# Config file for internal CI
+
+# Location of the continuous shell script in repository.
+build_file: "grpc-java/buildscripts/kokoro/gae.sh"
+timeout_mins: 45
+action {
+  define_artifacts {
+    regex: "**/build/test-results/**/*.xml"
```
matt-kwong: This should probably be updated (or deleted) to fit what artifacts your test will generated. This can be done as your work on `gae.sh`

Issue #3575, opened 2017-10-12
On Windows, when doing a heartbeat check of a Grpc server using Socket, Grpc has very noisy logging


nddipiazza:
### What version of gRPC are you using?
grpcVersion = 1.6.1
protocVersion = 3.3.0
*Issue only reproduces on Windows operating systems.*

### What JVM are you using (`java -version`)?
java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)

### What did you do?
1) git clone this project https://github.com/nddipiazza/grpc-java-bidirectional-streaming-example
2) Open cmd prompt and run `gradlew startServer`
3) Open another cmd prompt and run `gradlew runHeartbeat`

Notice each time you do a heartbeat on Grpc's port you get this INFO stack trace:

```
Oct 12, 2017 12:58:21 AM io.grpc.netty.NettyServerTransport notifyTerminated
INFO: Transport failed
java.io.IOException: An established connection was aborted by the software in your host machine
        at sun.nio.ch.SocketDispatcher.read0(Native Method)
        at sun.nio.ch.SocketDispatcher.read(SocketDispatcher.java:43)
        at sun.nio.ch.IOUtil.readIntoNativeBuffer(IOUtil.java:223)
        at sun.nio.ch.IOUtil.read(IOUtil.java:192)
        at sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:380)
        at io.netty.buffer.PooledUnsafeDirectByteBuf.setBytes(PooledUnsafeDirectByteBuf.java:288)
        at io.netty.buffer.AbstractByteBuf.writeBytes(AbstractByteBuf.java:1106)
        at io.netty.channel.socket.nio.NioSocketChannel.doReadBytes(NioSocketChannel.java:372)
        at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:123)
        at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:645)
        at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:580)
        at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:497)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:459)
[...]

PR #3591, opened 2017-10-19
netty: separate connection log

zpencer:
Move netty connection log info to a separate logger:
  io.grpc.netty.NettyServerTransport.connections

Users can redirect or disable this log using the usual way:
  -Djava.util.logging.config.file="logging.properties"

Related to #1768, #3575

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServerTransport.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerTransport.java
@@ -39,6 +39,8 @@ import java.util.logging.Logger;
  */
 class NettyServerTransport implements ServerTransport {
   private static final Logger log = Logger.getLogger(NettyServerTransport.class.getName());
+  private static final Logger connectionLog = Logger.getLogger(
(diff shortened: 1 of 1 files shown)


netty/src/main/java/io/grpc/netty/NettyServerTransport.java
```diff
@@ -39,6 +39,8 @@
  */
 class NettyServerTransport implements ServerTransport {
   private static final Logger log = Logger.getLogger(NettyServerTransport.class.getName());
```
dapengzhang0: Is the original `log` being used anymore after the change?

zpencer: I want to have both so it's more obvious to developers in the future that there's a difference between them.

dapengzhang0: Then you might comment on the two fields what they are used for, and maybe add `@SuppressWarnings("unused")` to the first one.

PR #3598, opened 2017-10-21
Replace IntelliJ IDEA specific sourceDirs with sourceSets

vorburger:
The sourceSets also works with Eclipse, and presumably other IDEs (OOB).

see also https://github.com/google/protobuf-gradle-plugin/issues/109

### DIFF
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -51,17 +51,18 @@ protobuf {
   }
 }
 
-// Inform IntelliJ projects about the generated code.
-apply plugin: 'idea'
-
-idea {
-  module {
-    // Not using generatedSourceDirs because of
-    // https://discuss.gradle.org/t/support-for-intellij-2016/15294/8
-    sourceDirs += file("${projectDir}/build/generated/source/proto/main/java");
-    sourceDirs += file("${projectDir}/build/generated/source/proto/main/grpc");
-  }
+// Inform IDEs like IntelliJ IDEA, Eclipse or NetBeans about the generated code.
+sourceSets {
+    main {
+        java {
+            srcDirs 'build/generated/source/proto/main/grpc'
+            srcDirs 'build/generated/source/proto/main/java'
+        }
+    }
 }
+
+// Generate IntelliJ IDEA's .idea & .iml project files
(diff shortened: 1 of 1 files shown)


ejona86:
@zhangkun83 @zpencer, if this is really all that's necessary, it seems like the protobuf plugin can handle it.  I'm having a hard time remembering how the source sets are put together, but should we change the code [here-ish](https://github.com/google/protobuf-gradle-plugin/blob/a636d320099838217ea673ec472162603cc01420/src/main/groovy/com/google/protobuf/gradle/ProtobufPlugin.groovy#L298)?

zhangkun83:
The protobuf plugin gets to know the locations of generated code and add them to the Java compile srcDirs at the `afterEvaluate` phase (I had a hard time trying to move it to an earlier point). Apparently it's not recognized by IDEA.

vorburger:
@ejona86 @zhangkun83 yeah, if the protoc-gen-grpc-java Gradle plugin could somehow correctly add the folder of the generated sources, then this would not be neccessary... But given that it currently does not, and that you have explicit instructions for IDEA in the example build.gradle, in this PR I'm just suggesting to replace those with more general non-IDEA specific ones, which work e.g. for Eclipse as well.

I'm not much of a Gradle expert, but perhaps someone knows anyway with more Gradle plugin dev know-how who can point out how to correctly add the folder of the generated sources in the gRPC Gradle Plugin code? Android team? 

@oehme may have some input into this: When I use [your org.xtext.xtend Gradle plugin](https://github.com/xtext/xtext-gradle-plugin), I don't think I ever have to explicitly add the `build/xtend/main` location as I have to here for gRPC.. you somehow made it "just work" - perhaps @oehme you could point @zhangkun83 here to how to do that right?

oehme:
The protobuf plugin should add its proto source directory set to the "allSource" source directory set. That's what the IDEs query.

ejona86:
okay to test

ejona86:
The gradle plugin discussion will take more effort. This seems strictly better than what was there previously. Once VS 2013 finishes running I'll merge.

PR #3599, opened 2017-10-22
kokoro: Initial config files for macos

zpencer:
An initial noop task, so that our internal build system always succeeds for osx.

The end goal is to move macos builds from travis to kokoro.


zpencer:
retest this please

PR #3602, opened 2017-10-23
core: Use literal host:port instead of computing

ejona86:
Using HostAndPort just complicates the code. HostAndPort is also @Beta,
which is generally fine to use in tests, but is needless here.


ejona86:
retest this please

Issue #3593, opened 2017-10-20
netty SSL is wrong


linking12:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?
grpc 1.7.0

### What JVM are you using (`java -version`)?
java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)

### What did you do?
```
               <dependency>
			<groupId>io.netty</groupId>
			<artifactId>netty-tcnative</artifactId>
			<version>${netty-tcnative.version}</version>
			<classifier>${os.detected.classifier}</classifier>
		</dependency>
```
I  think do not need jetty alpn.

but:
```
Caused by: java.lang.ClassNotFoundException: org/eclipse/jetty/alpn/ALPN
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:348)
	at io.grpc.netty.JettyTlsUtil.isJettyAlpnConfigured(JettyTlsUtil.java:34)
```

netty has not load native library 

In netty OpenSSL

 ```
  private static void loadTcNative() throws Exception {
        String os = PlatformDependent.normalizedOs();
        String arch = PlatformDependent.normalizedArch();

        Set<String> libNames = new LinkedHashSet<String>(4);
        // First, try loading the platform-specific library. Platform-specific
        // libraries will be available if using a tcnative uber jar.
        libNames.add("netty_tcnative_" + os + '_' + arch);
        if ("linux".equalsIgnoreCase(os)) {
            // Fedora SSL lib so naming (libssl.so.10 vs libssl.so.1.0.0)..
            libNames.add("netty_tcnative_" + os + '_' + arch + "_fedora");
        }
        // finally the default library.
        libNames.add("netty_tcnative");

        NativeLibraryLoader.loadFirstAvailable(SSL.class.getClassLoader(),
            libNames.toArray(new String[libNames.size()]));
    }
```
   //mybe is  
     <font face="#FF7F50">libNames.add("netty-tcnative-" + os + '-' + arch);</font>
  (https://repo1.maven.org/maven2/io/netty/netty-tcnative/2.0.6.Final/netty-tcnative-2.0.6.Final-osx-x86_64.jar)

the library name is:

netty-tcnative-2.0.6.Final-osx-x86_64.jar

linking12:
<dependencies>
    <dependency>
[...]

PR #3604, opened 2017-10-23
netty: Updated ALPN error message and added troubleshooting procedures

saturnism:
For #3025

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -287,7 +287,7 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<RespT> c
 
 ## Troubleshooting
 
-If you received an error message "ALPN is not configured properly", it most likely means that:
+If you received an error message "ALPN is not configured properly" or "Jetty ALPN/NPN has not been properly configured", it most likely means that:
  - ALPN related dependencies are either not present in the classpath
  - or that there is a classpath conflict
  - or that a wrong version is used due to dependency management.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -306,6 +306,8 @@ If you have both `io.grpc:grpc-netty` and `io.grpc:grpc-okhttp`, you may also ha
 
 If you have both `netty-codec-http2` and `netty-tcnative-boringssl-static` dependencies, then check the versions carefully. These versions could've been overridden by dependency management from another BOM. You would receive the "ALPN is not configured properly" exception if there are using incompatible versions.
 
+If you have other `netty` dependencies, such as `netty-all`, that are pulled in from other libraries, then ultimately you should make sure only one `netty` dependency is used to avoid classpath conflict. The easiest way is to exclude transitive Netty dependencies from all the immediate dependencies, e.g., in Maven use `<exclusions>`, and then add an explict Netty dependency in your project along with the corresponding `tcnative` versions. See the versions table below.
+
 If you are using `musl` libc (e.g., with Alpine Linux), then `netty-tcnative-boringssl-static` won't work either. There are several alternatives:
  - Use [netty-tcnative-alpine] (https://github.com/pires/netty-tcnative-alpine)
  - Use a distribution w/ `glibc`
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -295,6 +295,8 @@ If you received an error message "ALPN is not configured properly" or "Jetty ALP
 ### Netty
 If you aren't using gRPC on Android devices, you are most likely using `grpc-netty` transport.
 
+If you are developing for Android and have a dependency on `grpc-netty`, you should remove it as `grpc-netty` is unsupported on Android. Use `grpc-okhttp` instead.
+
 Find the dependency tree (e.g., `mvn dependency:tree`), and look for versions of:
  - `io.grpc:grpc-netty`
  - `io.netty:netty-codec-http2`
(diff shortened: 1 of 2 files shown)


zpencer:
I recall mistakenly including both grpc-okhttp and grpc-netty at the same time when I only wanted grpc-okhttp, and getting this error. Might be worth calling out that if a user wants grpc-okhttp then grpc-netty should not appear in the classpath.

saturnism:
@zpencer does `grpc-okhttp` require some kind of tcnative/ALPN library as well? If so, which ones? :)  Thanks,

saturnism:
@zpencer also, are there any troubleshooting techniques specific to okhttp on android? e.g., grpc / okhttp / android versions?

zpencer:
okhttp requires only `io.grpc:grpc-okhttp:${grpcVersion}`, but if for some strange reason grpc-netty is also in the classpath, then I noticed grpc-netty will end up producing the ALPN error. I think the general troubleshooting take away is to always make sure only one of (netty, okhttp) is in the class path.

zpencer:
We should probably also remove the version numbers from the code snippets and point people to the table:
https://github.com/grpc/grpc-java/issues/3593

SECURITY.md
```diff
@@ -285,6 +285,74 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<RespT> c
...
+
+Find the dependency tree (e.g., `mvn dependency:tree`), and look for versions of:
+ - `io.grpc:grpc-okhttp`
+ - `com.squareup.okhttp:okhttp`
+
+If you don't have `grpc-okhttp`, you should add it as a dependency.
+
+If you have both `io.grpc:grpc-netty` and `io.grpc:grpc-okhttp`, you may also have issues. Remove `grpc-netty` if you are on Android.
+
+If you have `okhttp` version below 2.5.0, or, if you have okhttp 3.x, then it may not work with gRPC.
```
ejona86: All versions 2.5.0 and later should work fine with gRPC. OkHttp3 has a different group name and package name, so you'd simply have both OkHttp 2 and OkHttp 3 in your classpath. Version incompatibilities (other than downgrading) with okhttp have not been a problem since v0.12.0 (yes, in 2015).

SECURITY.md
```diff
@@ -285,6 +285,74 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<RespT> c
...
+If you aren't using gRPC on Android devices, you are most likely using `grpc-netty` transport.
+
+Find the dependency tree (e.g., `mvn dependency:tree`), and look for versions of:
+ - `io.grpc:grpc-netty`
+ - `io.netty:netty-codec-http2`
+ - `io.netty:netty-tcnative-boringssl-static:jar` 
+
+If `netty-tcnative-boringssl-static` is missing, then you either need to add it as a dependency, or use alternative methods of providing ALPN capability by reading the *Transport Security (TLS)* section carefully.
+
+If you have both `io.grpc:grpc-netty` and `io.grpc:grpc-okhttp`, you may also have issues. Remove `grpc-okhttp` if you are not on Android.
```
ejona86: I don't think this is true.

ejona86:
OkHttp requires Conscrypt, but supports Jetty ALPN for testing on JDK. There is no conflict with having both OkHttp and Netty, the only problem is that if OkHttp is being used then "diagnosing" Netty's ALPN is useless.

SECURITY.md
```diff
@@ -285,6 +285,74 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<RespT> c
...
+
+If you have both `netty-codec-http2` and `netty-tcnative-boringssl-static` dependencies, then check the versions carefully. These versions could've been overridden by dependency management from another BOM. You would receive the "ALPN is not configured properly" exception if there are using incompatible versions.
```
ejona86: Grammar: "there are using" seems wrong

SECURITY.md
```diff
@@ -285,6 +285,74 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<RespT> c
...
+
+If you have other `netty` dependencies, such as `netty-all`, that are pulled in from other libraries, then ultimately you should make sure only one `netty` dependency is used to avoid classpath conflict. The easiest way is to exclude transitive Netty dependencies from all the immediate dependencies, e.g., in Maven use `<exclusions>`, and then add an explict Netty dependency in your project along with the corresponding `tcnative` versions. See the versions table below.
+
+If you are using `musl` libc (e.g., with Alpine Linux), then `netty-tcnative-boringssl-static` won't work either. There are several alternatives:
```
ejona86: Remove "either"

SECURITY.md
```diff
@@ -285,6 +285,74 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<RespT> c
...
+ - Use [netty-tcnative-alpine] (https://github.com/pires/netty-tcnative-alpine)
+ - Use a distribution w/ `glibc`
```
ejona86: Spell out "with".

SECURITY.md
```diff
@@ -285,6 +285,74 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<RespT> c
...
+
+If you are running inside of an embedded Tomcat runtime (e.g., Spring Boot), then some versions of `netty-tcnative-boringssl-static` will have conflicts and won't work.
```
ejona86: Simply say they must use grpc-java 1.4.0 or later and remove "Conflicts w/ Embedded Tomcat" from the table.

SECURITY.md
```diff
@@ -285,6 +285,74 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<RespT> c
...
+1.1.0              | 4.1.8.Final              | 1.1.33.Fork26                           | Yes
+1.2.0              | 4.1.8.Final              | 1.1.33.Fork26                           | Yes
+1.3.0              | 4.1.8.Final              | 1.1.33.Fork26                           | Yes
+1.4.0              | 4.1.11.Final             | 2.0.1.Final                             | No
+1.5.0              | 4.1.12.Final             | 2.0.5.Final                             | No
+1.6.1              | 4.1.14.Final             | 2.0.5.Final                             | No
+1.7.0              | 4.1.16.Final             | 2.0.6.Final                             | No
+
+### OkHttp
+If you are using gRPC on Android devices, you are most likely using `grpc-netty` transport.
```
ejona86: s/grpc-netty/grpc-okhttp/

SECURITY.md
```diff
@@ -285,6 +285,74 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<RespT> c
...
+
+Find the dependency tree (e.g., `mvn dependency:tree`), and look for versions of:
+ - `io.grpc:grpc-okhttp`
+ - `com.squareup.okhttp:okhttp`
+
+If you don't have `grpc-okhttp`, you should add it as a dependency.
+
+If you have both `io.grpc:grpc-netty` and `io.grpc:grpc-okhttp`, you may also have issues. Remove `grpc-netty` if you are on Android.
```
ejona86: How about in the netty section:
> If you are developing for Android and have a dependency on `io.grpc:grpc-netty`, you should remove it as grpc-netty is unsupported on Android. Use `io.grpc:grpc-okhttp` instead.

SECURITY.md
```diff
@@ -285,6 +285,74 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<RespT> c
 
 [Mutual authentication]: http://en.wikipedia.org/wiki/Transport_Layer_Security#Client-authenticated_TLS_handshake
 
+## Troubleshooting
+
+If you received an error message "ALPN is not configured properly" or "Jetty ALPN/NPN has not been properly configured", it most likely means that:
+ - ALPN related dependencies are either not present in the classpath
+ - or that there is a classpath conflict
+ - or that a wrong version is used due to dependency management.
```
ejona86: For this case, we should tell Maven users to use Maven Enforcer with requireUpperBoundDeps (probably preferred) or dependencyConvergence.

saturnism: Even w/ dependencyConvergence, it may still converge to a wrong version, and/or worse yet, enforcing a newer version of netty, but w/ an incompatible version of tcnative.

SECURITY.md
saturnism: @zpencer ?

SECURITY.md
saturnism: ack & renamed to "if you are using"

SECURITY.md
saturnism: done

SECURITY.md
saturnism: done

SECURITY.md
saturnism: done

SECURITY.md
saturnism: great catch. done!

SECURITY.md
saturnism: done

SECURITY.md
saturnism: ack; updated verbiage

saturnism:
@ejona86 I fixed a couple of things based on comment. ptal, thnx!

SECURITY.md
zpencer: @ejona86 is right, looks like this was a mistake. It turns out that ManagedChannelBuilder prefers netty, except in the case of appengine+jdk7 and the case of android. If users create the okhttp channel via `OkHttpChannelBuilder` then there is no problem. If ManagedChannelBuilder selects netty, then the usual prerequisite of netty-tcnative applies and may lead to this ALPN error.

ejona86:
Okay to test

ejona86:
@zpencer, do you want to take another look?

saturnism:
Thank all!

PR #3606, opened 2017-10-23
stub: add withExecutor API

dapengzhang0:


stub/src/main/java/io/grpc/stub/AbstractStub.java
```diff
@@ -121,6 +122,18 @@ public final S withDeadlineAfter(long duration, TimeUnit unit) {
   }
 
   /**
+   * Returns a new stub with the given executor that is to be used instead of the default one
+   * specified with {@link ManagedChannelBuilder#executor}. Note that setting this option may not
+   * take effect for blocking unary calls.
```
zhangkun83: Not just for unary calls, it's also the case for blocking server-streaming calls.
You may just say "for blocking calls", or "on blocking stubs".

PR #3607, opened 2017-10-24
context: Make CancellableContext implement Closeable

zpencer:
This achieves two things:
* communicates that `CancellableContext` is a resource that must be freed
* provides the ability to use the try-with-resources idiom to cancel the context when we finish up some blocking work. For users, this may be a better API than `runAndCancel` and `callAndCancel`.

@ejona86 and I considered adding a try-with-resources API that attaches the CancellableContext before the `try`, and *both* cancels as well as detaches at the `close`. But the problem is that the close runs before any catch statements, so users will not be able to use the exception as the cancellation cause unless they use a nested try/catch.

We also considered adding a try-with-resources API to attach/detach a context without cancelling, so that the two `try` statements in the example will look more symmetrical. But the syntactic sugar it provides is very minor. 

Internal use cases that kick off work asynchronously and use `ListenableFuture` will use an internal helper utility. We can not easily open source that code because `grpc-context` must have minimal dependencies and can not use guava.

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -683,7 +683,7 @@ public class Context {
    * }
    * </pre>
    *
-   * Asynchronous code will have to manually track the end of the CancellableContext's lifetime,
+   * <p>Asynchronous code will have to manually track the end of the CancellableContext's lifetime,
    * and call either {@link #cancel} or {@link #close} at that time.
    */
   public static final class CancellableContext extends Context implements Closeable {
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -16,6 +16,7 @@
 
 package io.grpc;
 
+import java.io.Closeable;
 import java.util.ArrayList;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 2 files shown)


zpencer:
CC: @adriancole

adriancole:
Tldr I like this because errorprone can warn on lack of close. Wrt how to
handle exceptions, the example is honest.

Asides:

I know @pauldraper attempted to layer capabilities to handle exceptions
neatly (in a different lib), cleverly with multiple "throws arity"
https://github.com/lucidsoftware/java-thread-context/tree/master/thread-context/src/main/java/com/github/threadcontext/function
cc also @nicmunroe

Ack grpc context isnt java 8 but the approach could be layered with non
java 8 types here or elsewhere. I feel we are getting closer to something
ideal, but try/resources idiom is limited.

One idea is a compiler or codegen which "fixes" the catch vs autoclose
problem. Ex possibly similar to autovalue memoize or retrolambda. In this
way we could eliminate the boilerplate by hiding it somehow. Cc @orfjackal
@raphw @eamonnmcmanus

context/src/main/java/io/grpc/Context.java
```diff
@@ -657,8 +658,35 @@ private Object lookup(Key<?> key) {
    * cancelled and which will propagate cancellation to its descendants. To avoid leaking memory,
    * every CancellableContext must have a defined lifetime, after which it is guaranteed to be
    * cancelled.
+   *
+   * <p>This class must be cleaned up by either calling {@link #close} or {@link #cancel}.
+   * {@link #close} is equivalent to calling {@code cancel(null)}. It is safe to call the methods
```
zhangkun83: Maybe also in `cancel()`'s javadoc refer back to `close()`.

context/src/main/java/io/grpc/Context.java
```diff
@@ -657,8 +658,35 @@ private Object lookup(Key<?> key) {
...
+   * more than once, but only the first call will have any effect.
+   *
+   * <p>Blocking code can use the try-with-resources idiom:
+   * <pre>
```
ejona86: Get rid of the extra newlines. Take a look at the generated documentation to see how it looks.

ejona86:
Concerning catching the exception: I think I may encourage people not to (and just `cancel(null)`/`close()`).

Basically there's two different operations at play here: cancellation (cancel with error) and cleanup (cancel with null). My hypothesis is that generally when using try-with-resources (aka, you're using blocking-style APIs), you have no need to cancel, even if an exception was thrown. The only time "you're cancelling" is when the Deadline is reached, which provides an exception.

Now, that's not at all true when doing async work. But if you're doing async work you're much less likely to use try-with-resources.

adriancole:
One note is that at least with a trace context (usually not using this form
but as an example) we often use try with because you can end up with an
exception creating or starting a deferred task (ex latter could be a
rejection cause). Not sure if such a pattern applies elsewhere or if there
is a risk in not knowing the cause.

PR #3610, opened 2017-10-24
SECURITY.md: fix tcnative version for v1.7.x

zpencer:
Reported here: #3593

I verified the problem by hand.

Related: #3604


zpencer:
Yes, this will be fixed on both v1.7.x as well as master.

PR #3614, opened 2017-10-25
testing: move TestUtils.recordServerCallInterceptor internal

dapengzhang0:


testing/src/main/java/io/grpc/internal/testing/TestUtils.java
```diff
@@ -49,6 +54,24 @@
   public static final String TEST_SERVER_HOST = "foo.test.google.fr";
 
   /**
+   * Captures the request attributes. Useful for testing ServerCalls.
+   * {@link ServerCall#getAttributes()}
+   */
+  public static ServerInterceptor recordServerCallInterceptor(
```
ejona86: Since this is only needed in AbstractInteropTest, move it there?

testing/src/main/java/io/grpc/testing/TestUtils.java
```diff
@@ -76,7 +76,10 @@ public static ServerInterceptor recordRequestHeadersInterceptor(
   /**
    * Capture the request attributes. Useful for testing ServerCalls.
    * {@link ServerCall#getAttributes()}
+   *
+   * @deprecated Not for public use
```
carl-mastrangelo: are you intending to remove this?

dapengzhang0: Yes, can not think of a usecase of it for testing from user's perspective.

Issue #1791, opened 2016-05-03
Tracking Issue for Testing utilities being Experimental. 


carl-mastrangelo:
Specific usages;

* `StreamRecorder`
* `TestUtils`

dapengzhang0:
not stabilized yet, keep unscheduled.

trustin:
I realized that all members of `TestUtils` have been deprecated. We use them for our gRPC interop tests:

https://github.com/line/armeria/blob/master/it/src/test/java/com/linecorp/armeria/server/grpc/interop/ArmeriaGrpcServerInteropTest.java#L80

I also see a similar problem on the client side. We need `TestUtils.newSslSocketFactoryForCa()`.

What would be the recommended way to run this sort of interop tests?

dapengzhang0:
@trustin I think you can define/fork your own function similar to `TestUtils.newSslSocketFactoryForCa()`, it's not hard because the function is not dependent on gRPC library.

The source code is at
https://github.com/grpc/grpc-java/blob/master/testing/src/main/java/io/grpc/internal/testing/TestUtils.java

ejona86:
If certain methods are useful, have a reasonable API, and are maintainable, we can keep them.

@trustin, details about your test:
* It is not using GrpcSslContexts. It can override `trustManager` after configuring the SslContext via GrpcSslContexts, but it should pass through GrpcSslContexts.
* `preferredTestCiphers()` really should have been internal. Although you are using it the same as we are (testing OkHttp on the JDK). This was needed when the JDK's GCM went at 2 MB/s. Now it's 10x that and JDK 9 will be 10x again. So maybe we just drop it. Alternatively we can kill it in favor of Conscrypt. It is only helpful with Jetty ALPN. Conscrypt is _very_ close to 1.0 with pre-built binaries for multiple platforms. But OkHttp will need some fixes to make use of it off-Android.
* The usage of `newSslSocketFactoryForCa` is coupled with a usage of `io.grpc.okhttp.internal.Platform`. That's _obviously_ not supported. But I understand the need.
* There's a usage of `GrpcUtil.authorityFromHostAndPort` which is also internal. It could have been just `"example.com:" + port`. A big reason for the utility is to handle IPv6 IPs which need `[]`. That's unnecessary in the test

I'm fine to work out utility functions for using test certificates. I understand that is a major PITA and using grpc's pre-generated test certs makes things much easier. But I expect they could be simplified.

[...]

PR #3616, opened 2017-10-25
testing: add TlsTesting.loadCert

dapengzhang0:
addressing comments in #1791


testing/src/main/java/io/grpc/testing/TestUtils.java
```diff
@@ -180,6 +180,17 @@ public static File loadCert(String name) throws IOException {
   }
 
   /**
+   * Returns the given certificate resource in src/main/resources/certs/ as an {@code InputStream}.
+   *
+   * @param name name of a file in src/main/resources/certs/, e.g., {@code "ca.key"}.
+   *
+   * @since 1.8.0
+   */
+  public static InputStream loadCertFromResource(String name) throws IOException {
```
ejona86: I'd like to move this out of `TestUtils`, because `TestUtils` is a dumping ground. I was sort of thinking we could create a new class `TlsTesting` or similar for it: `TlsTesting.loadCert`.

PR #3618, opened 2017-10-25
 compiler: add methods for accessing method descriptors

carl-mastrangelo:
Several commits so you can see the changes

Key things:
* MethodDescriptor is lazy loaded, so protobuf loading only happens on demand.  This also means tracing registration happens  on demand.  
* The names of the getters all being with `method`.  This makes it harder for autocomplete to pick them up.
* A new field is used, which matches the getter name.  Rather than make the new-getters reference the old-fields, make the old-fields reference the new getters.  This makes removal of the old-fields a simple operation.  
* The getters may not be inlineable, but thats an easy fix if it ends up being a problem.  Not worth premature optimization (but is worth future work).

The expected timeline for this is adding this to the 1.8 cut, and deprecating the old-fields.  They will be removed in 1.9.


carl-mastrangelo:
Changes made as discussed.

PR #3619, opened 2017-10-25
testing: favor voidMethod and deprecate noopMethod

dapengzhang0:


testing/src/main/java/io/grpc/testing/TestMethodDescriptors.java
```diff
@@ -40,20 +40,29 @@ private TestMethodDescriptors() {}
...
+        .setRequestMarshaller(TestMethodDescriptors.voidMarshaller())
+        .setResponseMarshaller(TestMethodDescriptors.voidMarshaller())
+        .build();
   }
 
   /**
    * Creates a new method descriptor that always creates zero length messages, and always parses to
    * null objects.
    *
+   * @deprecated Use {@link #voidMethod()} instead or build one with {@link
```
ejona86: I thought we wanted them to use a descriptor from SimpleService if voidMethod didn't suffice.

dapengzhang0: Right, we wanted them
1. if they don't care about message, use `voidMethod`
2. if they care about message, but don't care about its type, use the one from `SimpleService`
3. If they care about message and also its type, they should be able to build one for their own message type.

Will update the comment a little bit.

ejona86: And instead of building their own descriptor, I think we thought to suggest they just make a custom .proto for testing. Although it's up to them whatever they find easiest.

testing/src/main/java/io/grpc/testing/TestMethodDescriptors.java
```diff
@@ -40,20 +40,29 @@ private TestMethodDescriptors() {}
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/2600")
   public static MethodDescriptor<Void, Void> voidMethod() {
-    return TestMethodDescriptors.<Void, Void>noopMethod();
+    return MethodDescriptor.<Void, Void>newBuilder()
```
carl-mastrangelo: Are the types needed here?

testing/src/main/java/io/grpc/testing/TestMethodDescriptors.java
```diff
@@ -40,20 +40,29 @@ private TestMethodDescriptors() {}
...
   /**
    * Creates a new method descriptor that always creates zero length messages, and always parses to
    * null objects.
    *
+   * @deprecated Prefer to use {@link #voidMethod()} instead or use one MethodDescriptor from {@code
+   *             io.grpc.testing.protobuf.SimpleServiceGrpc} or from other generated classes.
    * @since 1.1.0
    */
+  @Deprecated
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/2600")
```
carl-mastrangelo: Please mark the issue with the targeted removal date

PR #3620, opened 2017-10-25
core: remove unnecessary type params in ServerStreamTracer

dapengzhang0:


zhangkun83:
Will it break if it's removed from the interface but kept on the implementation? It's preferable to not break any implementations.

dapengzhang0:
If overriding with the full signature including the type params `<ReqT, RespT>`, the subclass will break. But I don't find any external subclass overriding it with the full signature, they are either without type params or not overriding it. After all it is an experimental API added not long ago.

thelinuxfoundation:

* User @dapengzhang0 isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,

PR #3608, opened 2017-10-24
kokoro: port .travis.yaml to kokoro

zpencer:

PR #3626, opened 2017-10-26
travis: Remove OS X build

### LINKED ISSUE #3466: Migrate OS X CI to Kokoro
Travis works quickly. For Linux. But not OS X.

OS X tends to have a long build queue (right now it is at least 3 hours long). This could be fine, except if 5 OS X jobs are queued they starve out Linux builds. This has been a problem since OS X support was added to Travis, so waiting for things to get better doesn't seem a good strategy. Since the "Backlog macOS Builds for Open Source projects" tells a good story on https://www.traviscistatus.com/ , we could notice when things have improved in the future.

In principle it'd be fine to only run on master (not per-PR), but I don't know of any option to do that for Travis. Kokoro shouldn't have any trouble handling our rate of PR builds.

ejona86:
With #3608 Kokoro is now able to handle OS X. I'm not removing the
OS X-specific parts of the .travis.yml in case we need to revert back to
using Travis.

Fixes #3466

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,6 @@ after_success:
 
 os:
   - linux
-  - osx
 
 notifications:
   email: false


adriancole:
Smart. Build slaves are funny.. sometimes they take turns flaking out. Being able to quickly turn travis back on might be helpful one day.

PR #3627, opened 2017-10-26
core: set sampled for local span per MethodDescriptor.

zhangkun83:
This moves away from the global String-based Span name registry which
is not as flexible as we desire.

Also renamed the option name to be more accurate.  This is not
API-breaking because the origianl addition to MethodDescriptor and
code-gen didn't make it into the 1.7.0 release.

This is using a snapshot of opencensus, and needs to switch to a
release before merging.


thelinuxfoundation:

* User @zhangkun83 isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,

zhangkun83:
Resolves #3359

zhangkun83:
Retest this please

PR #3628, opened 2017-10-27
buildscripts: make the script fail for the right reason with set -e

zpencer:
Remove the `&&`s so that the script will fail due to the build command failing, instead of `pushd` failing


zpencer:
retest this please

Issue #3633, opened 2017-10-30
Java 9: cannot find symbol javax.annotation.Generated


fcnn:
/root/grpc-kotlin-hello/proto/src/main/java/io/grpc/examples/helloworld/GreeterGrpc.java:23: error: cannot find symbol
@javax.annotation.Generated(
                 ^
  symbol:   class Generated
  location: package javax.annotation
1 error

FAILURE: Build failed with an exception.

ejona86:
That class exists in Java 9: https://docs.oracle.com/javase/8/docs/api/javax/annotation/Generated.html . It seems like the problem is with an example that isn't maintained here.

ejona86:
Okay... I've now run into this as well for our example after #3637.

ejona86:
@onesea, I'd suggest adding a dependency on `javax.annotation:jsr250-api:1.0` for the moment. It's possible to pass `--add-modules java.xml.ws.annotation`, but it seems it might cause more problems than it solves.

lcb1208:
java 8  build , error not find class "javax.annotation.concurrent.ThreadSafe"

davido:
I run into the same problem while working on Bazel build with Java 9: #4709.

davido:
The way it was fixed in other java based libraries, like auto-common, auto-value and dagger, is to dynamically select the right annotation:

- Java 9: javax.annotation.processing.Generated
- Java 8: javax.annotation.Generated

See e.g.: https://github.com/google/auto/commit/f04406c1f11d3796bd93ed89ae160f47bcb4b7e4,

ejona86:
@davido, dynamic selection doesn't really seem to be a solution except for annotation processors.

davido:
I was able to patch gRpc compiler in Bazel tree, with this one liner:

```
diff --git a/third_party/grpc/compiler/src/java_plugin/cpp/java_generator.cpp b/third_party/grpc/compiler/src/java_plugin/cpp/java_generator.cpp
index 7cb01238ac..3264f15ecf 100644
--- a/third_party/grpc/compiler/src/java_plugin/cpp/java_generator.cpp
+++ b/third_party/grpc/compiler/src/java_plugin/cpp/java_generator.cpp
@@ -1203,7 +1203,7 @@ void GenerateService(const ServiceDescriptor* service,
   vars["NanoUtils"] = "io.grpc.protobuf.nano.NanoUtils";
[...]

PR #3638, opened 2017-10-30
Fix compilation in Java 9

ejona86:
~~Tests still fail; at the very least because of missing Jetty ALPN.~~ I
don't know why the dependency on javax.annotations is unnecessary in the
compiler tests and the example when using Maven.

This depends on #3637. This isn't quite enough for #3633, because we've
not yet clear in our README what's necessary. The annotations stuff is still
very confusing to me; I think our users will suffer unless we have a pretty good
understanding of what's going on and clearly document what they need to do.

CC @carl-mastrangelo

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -192,6 +192,7 @@ subprojects {
                 gson: "com.google.code.gson:gson:2.7",
                 guava: "com.google.guava:guava:${guavaVersion}",
                 hpack: 'com.twitter:hpack:0.10.1',
+                javax_annotation: 'javax.annotation:javax.annotation-api:1.2',
                 jsr305: 'com.google.code.findbugs:jsr305:3.0.0',
                 oauth_client: 'com.google.auth:google-auth-library-oauth2-http:0.9.0',
                 google_api_protos: 'com.google.api.grpc:proto-google-common-protos:1.0.0',
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,11 +36,16 @@ script:
   - ./gradlew check :grpc-all:jacocoTestReport
 
 after_success:
-  - if \[ "$TRAVIS_OS_NAME" = linux \]; then ./gradlew :grpc-all:coveralls; fi
+    # Upload to coveralls once, instead of for each job in the matrix
+  - if \[\[ "$TRAVIS_JOB_NUMBER" == *.1 \]\]; then ./gradlew :grpc-all:coveralls; fi
   - bash <(curl -s https://codecov.io/bash)
 
 os:
   - linux
+
+jdk:
+  - oraclejdk8
+  - oraclejdk9
 
 notifications:
   email: false
--- a/benchmarks/build.gradle
+++ b/benchmarks/build.gradle
@@ -37,6 +37,7 @@ dependencies {
             libraries.netty_tcnative,
             libraries.netty_epoll,
             libraries.math
+    compileOnly libraries.javax_annotation
 }
 
 compileJava {
--- a/core/src/test/java/io/grpc/AttributesTest.java
+++ b/core/src/test/java/io/grpc/AttributesTest.java
@@ -53,11 +53,20 @@ public class AttributesTest {
   }
 
   @Test
-  @SuppressWarnings("BoxedPrimitiveConstructor")
   public void valueEquality() {
-    Attributes.Key<Integer> key = Attributes.Key.of("ints");
-    Integer v1 = new Integer(100000);
-    Integer v2 = new Integer(100000);
+    class EqualObject {
+      @Override public boolean equals(Object o) {
+        return o instanceof EqualObject;
+      }
+
+      @Override public int hashCode() {
+        return 42;
+      }
+    }
+
+    Attributes.Key<EqualObject> key = Attributes.Key.of("ints");
+    EqualObject v1 = new EqualObject();
+    EqualObject v2 = new EqualObject();
 
     assertNotSame(v1, v2);
     assertEquals(v1, v2);
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -37,6 +37,7 @@ dependencies {
   compile "io.grpc:grpc-netty:${grpcVersion}"
   compile "io.grpc:grpc-protobuf:${grpcVersion}"
   compile "io.grpc:grpc-stub:${grpcVersion}"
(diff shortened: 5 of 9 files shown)


ejona86:
There's quite a bit of confusion on what to do:
 * `--add-modules java.xml.ws.annotation`. Based on [the Java documentation](https://docs.oracle.com/javase/9/migrate/toc.htm#JSMIG-GUID-F640FA9D-FB66-4D85-AD2B-D931174C09A3) that seems like a temporary solution because the module will be removed. This is also incompatible with `-target` and `-source`, but it may be compatible with `-release` (I didn't try). Also, when using a module, I think it breaks `javax.annotation.Nullable` (which is probably another issue we need to fix independent of this)
 * There's plenty of references to `javax.annotation:jsr250-api:1.0` (note search.maven.org thinks 1.0-20050927.133100 is a later version).
 * Then there's also `javax.annotation:javax.annotation-api`. The versions should match [jsr250](https://jcp.org/en/jsr/detail?id=250), but as with most JSRs it's hard to figure out what version to use when. 1.3 uses `-target 1.8` and 1.2 uses `-target 1.6`
 * Then there's `org.glassfish:javax.annotation`. The JSR references Glassfish as the "Reference Implementation", but doesn't provide any version information

javax.annotation-api seems the thing to use, but I'd really like to find some clear documentation saying to use it and which version to use (without becoming an expert on the JSR). I would have also hoped for clear documentation (like, "we deploy to Maven Central with artifact id BLAH") at https://github.com/javaee/javax.annotation, but alas. I'll note, it mentions module `java.annotation` "to match the module name used in JDK 9", but a module by that name doesn't exist in JDK 9; I feel reasonably confident it was renamed several times during JDK 9 development, as [in a bug](https://bugs.openjdk.java.net/browse/JDK-8152842) it's called `java.annotations.common` instead.

Because of the poor state of dependency clarity, I think it's _very_ safe to believe all the deps will be mixed. Luckily, that may not break anything because `@Generated` is a source-only annotation.

In other news, I continue to dislike the poor usability of the JSR website.

build.gradle
```diff
@@ -214,6 +215,8 @@ subprojects {
                 netty_proxy_handler: "io.netty:netty-handler-proxy:${nettyVersion}",
                 netty_tcnative: 'io.netty:netty-tcnative-boringssl-static:2.0.7.Final',
 
+                conscrypt: 'org.conscrypt:conscrypt-openjdk-uber:1.0.0',
```
carl-mastrangelo: intended?

interop-testing/src/test/java/io/grpc/ChannelAndServerBuilderTest.java
```diff
@@ -48,19 +49,39 @@
...
+    Collection<ClassInfo> infos = ClassPath.from(loader).getTopLevelClassesRecursive("io.grpc");
+    // If infos is empty, then we can't verify our hard-coded list. Assume that's due to Java 9.
+    if (!infos.isEmpty()) {
+      List<Class<?>> classes = new ArrayList<Class<?>>();
+      for (ClassInfo classInfo : infos) {
+        Class<?> clazz = Class.forName(classInfo.getName(), false /*initialize*/, loader);
+        if (ServerBuilder.class.isAssignableFrom(clazz) && clazz != ServerBuilder.class) {
+          classes.add(clazz);
+        } else if (ManagedChannelBuilder.class.isAssignableFrom(clazz)
+            && clazz != ManagedChannelBuilder.class ) {
```
carl-mastrangelo: nit: extra space

interop-testing/src/test/java/io/grpc/ChannelAndServerBuilderTest.java
```diff
@@ -48,19 +49,39 @@
    */
   @Parameters(name = "class={0}")
   public static Collection<Object[]> params() throws Exception {
+    @SuppressWarnings("unchecked")
```
carl-mastrangelo: This somewhat defeats the purpose of this test.   It showed that every class that gRPC knows about overrides the static ctors, but since you enumerate it, new builders will be missed.    (Like the recently added Alts Channel)

build.gradle
ejona86: Yes. Jetty ALPN doesn't exist in Java 9. So for OkHttp TLS testing it will use Conscrypt instead.

interop-testing/src/test/java/io/grpc/ChannelAndServerBuilderTest.java
ejona86: Not so. We have a hard-coded list, but it is _verified_ when possible (JDK < 9). So that means the list will be kept up-to-date. It is true that if we only ran on JDK 9 the list would likely get out-of-date, but in that case it seems like the best we could do is use the hard-coded list or delete the test. Looking at ClassPath.Scanner.getClassLoaderUrls, it seems impossible to fix for Java 9.

interop-testing/src/test/java/io/grpc/ChannelAndServerBuilderTest.java
ejona86: It was pre-existing, but done.

interop-testing/src/test/java/io/grpc/ChannelAndServerBuilderTest.java
carl-mastrangelo: If this doesn't work for Java 9, it might as well not run.    The list will get out of date, and it will give false confidence.

It would be better to check the java version and call Assume.assumeTrue(javaVersion() <= 8) in a BeforeClass.  If that isn't possible you could make a dummy test that asserts there is more than one class to check and throw an assumption exception otherwise.  

If this test doesn't work on java 9 then so be it.  That's okay.

ejona86: `@BeforeClass` runs after `params()`, so an assume in it doesn't work. Similarly, a failing assume in `params()` causes a test failure. And a dummy test doesn't work because no tests are run if `params()` returns an empty list.

I'm detecting when `getTopLevelClassesRecursive()` returns nothing and returning no params in that case. Note that returning an empty list of params causes the test class to be excluded from test reports, since by definition there were no tests within the class. So there isn't any "ignored" count or similar.

ejona86:
@carl-mastrangelo, PTAL

PR #3646, opened 2017-10-31
SECURITY.md: provider must be installed before channel creation

ericgribkoff:
Our static initialization in `OkHttpProtocolNegotiator` has impacted a couple of Android users recently when trying to use the Play Services Dynamic Security provider, so just calling out this requirement explicitly in our docs.


ericgribkoff:
retest this please? (waiting for cla/google)

PR #3647, opened 2017-10-31
core: use new OpenCensus stats/tagging API.

sebright:
This commit updates gRPC core to use io.opencensus:opencensus-api and
io.opencensus:opencensus-contrib-grpc-metrics instead of
com.google.instrumentation:instrumentation-api for stats and tagging. The gRPC
Monitoring Service continues to use instrumentation-api.

The main changes affecting gRPC:

- The StatsContextFactory is replaced by three objects, StatsRecorder, Tagger,
  and TagContextBinarySerializer.

- The StatsRecorder, Tagger, and TagContextBinarySerializer are never null,
  but the objects are no-ops when the OpenCensus implementation is not
  available.

This commit includes changes written by @songy23 and @sebright.

_________________________________________________________________

I added two TODOs for parts that I didn't know how to update.  This PR is still using the snapshot of OpenCensus.

/cc @zhangkun83

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -192,6 +192,7 @@ subprojects {
                 okhttp: 'com.squareup.okhttp:okhttp:2.5.0',
                 okio: 'com.squareup.okio:okio:1.6.0',
                 opencensus_api: 'io.opencensus:opencensus-api:0.8.0',
+                opencensus_contrib_grpc_metrics: 'io.opencensus:opencensus-contrib-grpc-metrics:0.8.0',
                 opencensus_impl: 'io.opencensus:opencensus-impl:0.8.0',
                 instrumentation_api: 'com.google.instrumentation:instrumentation-api:0.4.3',
                 protobuf: "com.google.protobuf:protobuf-java:${protobufVersion}",
--- a/core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
@@ -20,8 +20,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.MoreExecutors;
-import com.google.instrumentation.stats.Stats;
-import com.google.instrumentation.stats.StatsContextFactory;
 import io.grpc.BindableService;
 import io.grpc.CompressorRegistry;
 import io.grpc.Context;
@@ -36,6 +34,11 @@ import io.grpc.ServerMethodDefinition;
 import io.grpc.ServerServiceDefinition;
 import io.grpc.ServerStreamTracer;
 import io.grpc.ServerTransportFilter;
+import io.opencensus.stats.Stats;
+import io.opencensus.stats.StatsRecorder;
+import io.opencensus.tags.Tagger;
+import io.opencensus.tags.Tags;
+import io.opencensus.tags.propagation.TagContextBinarySerializer;
 import io.opencensus.trace.Tracing;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -97,7 +100,13 @@ public abstract class AbstractServerImplBuilder<T extends AbstractServerImplBuil
   CompressorRegistry compressorRegistry = DEFAULT_COMPRESSOR_REGISTRY;
 
   @Nullable
-  private StatsContextFactory statsFactory;
+  private Tagger tagger;
+
+  @Nullable
+  private TagContextBinarySerializer tagCtxSerializer;
+
+  @Nullable
+  private StatsRecorder statsRecorder;
 
   private boolean statsEnabled = true;
   private boolean recordStats = true;
@@ -184,8 +193,13 @@ public abstract class AbstractServerImplBuilder<T extends AbstractServerImplBuil
    * Override the default stats implementation.
    */
   @VisibleForTesting
-  protected T statsContextFactory(StatsContextFactory statsFactory) {
-    this.statsFactory = statsFactory;
+  protected T statsImplementation(
+      final Tagger tagger,
+      TagContextBinarySerializer tagCtxSerializer,
+      StatsRecorder statsRecorder) {
+    this.tagger = tagger;
+    this.tagCtxSerializer = tagCtxSerializer;
+    this.statsRecorder = statsRecorder;
     return thisT();
   }
 
@@ -228,13 +242,26 @@ public abstract class AbstractServerImplBuilder<T extends AbstractServerImplBuil
     ArrayList<ServerStreamTracer.Factory> tracerFactories =
         new ArrayList<ServerStreamTracer.Factory>();
     if (statsEnabled) {
-      StatsContextFactory statsFactory =
-          this.statsFactory != null ? this.statsFactory : Stats.getStatsContextFactory();
-      if (statsFactory != null) {
-        CensusStatsModule censusStats =
-            new CensusStatsModule(statsFactory, GrpcUtil.STOPWATCH_SUPPLIER, true, recordStats);
-        tracerFactories.add(censusStats.getServerTracerFactory());
-      }
+      Tagger tagger = this.tagger != null ? this.tagger : Tags.getTagger();
+      TagContextBinarySerializer tagCtxSerializer =
+          this.tagCtxSerializer != null
+              ? this.tagCtxSerializer
+              : Tags.getTagPropagationComponent().getBinarySerializer();
+      StatsRecorder statsRecorder =
+          this.statsRecorder != null ? this.statsRecorder : Stats.getStatsRecorder();
+      // // TODO: How do we check whether stats is enabled, now that the StatsRecorder is always
+      // // non-null? Uncommenting this line causes test failures.
+      // if (Stats.getState() == StatsCollectionState.ENABLED) {
+      CensusStatsModule censusStats =
+          new CensusStatsModule(
+              tagger,
+              tagCtxSerializer,
+              statsRecorder,
+              GrpcUtil.STOPWATCH_SUPPLIER,
+              true,
+              recordStats);
+      tracerFactories.add(censusStats.getServerTracerFactory());
+      // }
     }
     if (tracingEnabled) {
       CensusTracingModule censusTracing =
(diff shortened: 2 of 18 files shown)


thelinuxfoundation:

* User @sebright isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,

sebright:
testing Linux Foundation CLA

core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
```diff
@@ -228,13 +242,26 @@ public Server build() {
...
-      }
+      Tagger tagger = this.tagger != null ? this.tagger : Tags.getTagger();
+      TagContextBinarySerializer tagCtxSerializer =
+          this.tagCtxSerializer != null
+              ? this.tagCtxSerializer
+              : Tags.getTagPropagationComponent().getBinarySerializer();
+      StatsRecorder statsRecorder =
+          this.statsRecorder != null ? this.statsRecorder : Stats.getStatsRecorder();
+      // // TODO: How do we check whether stats is enabled, now that the StatsRecorder is always
+      // // non-null? Uncommenting this line causes test failures.
```
zhangkun83: Most tests pass in the fake implementations from `StatsTestUtils`, but `Stats.getState()` returns `DISABLED`. If you don't initialize `CensusStatsModule`, these tests will fail.

A possibly better approach, is to make `CensusStatsModule` public, and change the three-argument Census impl setter of the builder to `overrideCensusStatsModule(CensusStatsModule)`. Tests will create `CensusStatsModule` out of the impls from `StatsTestUtils`, and here you can do:
```java
if (statsEnabled) {
  CensusStatsModule censusStats = this.censusStatsOverride;
  if (censusStats == null && Stats.getStats() == StatsCollectionState.ENABLED) {
    CensusStatsModule censusStats =
        new CensusStatsModule(...);
  }
  if (censusStats != null) {
    tracerFactories.add(censusStats.getServerTracerFactory());
  }
}
```

core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
```diff
@@ -228,13 +242,26 @@ public Server build() {
...
         new ArrayList<ServerStreamTracer.Factory>();
     if (statsEnabled) {
-      StatsContextFactory statsFactory =
-          this.statsFactory != null ? this.statsFactory : Stats.getStatsContextFactory();
-      if (statsFactory != null) {
-        CensusStatsModule censusStats =
-            new CensusStatsModule(statsFactory, GrpcUtil.STOPWATCH_SUPPLIER, true, recordStats);
-        tracerFactories.add(censusStats.getServerTracerFactory());
-      }
+      Tagger tagger = this.tagger != null ? this.tagger : Tags.getTagger();
```
zhangkun83: Just curious, are the three components coupled? For example, does one Tagger implementation only works with a certain StatsRecorder, or can they mix and match?

sebright: The Tagger and TagContextBinarySerializer objects are coupled.  The StatsRecorder is independent, except that it also uses the gRPC Context key that is used by the tagging classes.

build.gradle
```diff
@@ -191,8 +191,9 @@ subprojects {
                 google_auth_credentials: 'com.google.auth:google-auth-library-credentials:0.4.0',
                 okhttp: 'com.squareup.okhttp:okhttp:2.5.0',
                 okio: 'com.squareup.okio:okio:1.6.0',
-                opencensus_api: 'io.opencensus:opencensus-api:0.7.0',
-                opencensus_impl: 'io.opencensus:opencensus-impl:0.7.0',
+                opencensus_api: 'io.opencensus:opencensus-api:0.8.0',
+                opencensus_contrib_grpc_metrics: 'io.opencensus:opencensus-contrib-grpc-metrics:0.8.0',
+                opencensus_impl: 'io.opencensus:opencensus-impl:0.8.0',
```
bogdandrutu: Why do you depend on the implementation?

core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
sebright: @zhangkun83 I refactored the code for overriding the OpenCensus implementation.  There are two parts that I'm not sure about.

- CensusStatsModule no longer caches the StatsClientInterceptor and ServerTracerFactory.  It creates a new object whenever getServerTracerFactory or getClientInterceptor is called.  Is that correct?
- I'm not completely sure I used the right values for `recordStats`, especially in CensusModulesTest.

sebright:
I rebased.  I didn't squash any commits, so there is still a commit to upgrade to OpenCensus 0.8.0, even though that change was made on master.

build.gradle
sebright: It looks like it is only needed for the interop tests: https://github.com/grpc/grpc-java/blob/37a67bf04111c0a9d42a7722b89bab028c882000/interop-testing/build.gradle#L29

ejona86:
Okay to test.

bogdandrutu:
@ejona86 sorry for the wrong "contrib" name. We used it for "other" artifacts that are not in the core library. We will work in the future to maybe change that name. Thanks for the feedback, we are committed to keep backwards compatibility for "contrib" as well.

ejona86:
retest this please

core/src/test/java/io/grpc/internal/AbstractManagedChannelImplBuilderTest.java
```diff
@@ -61,15 +67,56 @@
         }
       };
 
-  private static final StatsContextFactory DUMMY_STATS_FACTORY =
-      new StatsContextFactory() {
+  private static final Tagger DUMMY_TAGGER =
```
zhangkun83: It probably won't hurt if this test uses the fake implementations from StatsTestUtils. Then you can save a lot of lines of the dummy implementations.

core/src/test/java/io/grpc/internal/AbstractServerImplBuilderTest.java
```diff
@@ -33,15 +39,57 @@
 /** Unit tests for {@link AbstractServerImplBuilder}. */
 @RunWith(JUnit4.class)
 public class AbstractServerImplBuilderTest {
-  private static final StatsContextFactory DUMMY_STATS_FACTORY =
-      new StatsContextFactory() {
+
+  private static final Tagger DUMMY_TAGGER =
```
zhangkun83: ditto

testing/src/main/java/io/grpc/internal/TestingAccessor.java
```diff
@@ -16,26 +16,38 @@
...
  */
 public final class TestingAccessor {
   /**
-   * Sets a custom {@link StatsContextFactory} for tests.
+   * Sets a custom stats implementation for tests.
    */
-  public static void setStatsContextFactory(
-      AbstractManagedChannelImplBuilder<?> builder, StatsContextFactory factory) {
-    builder.statsContextFactory(factory);
+  public static void setStatsImplementation(
```
zhangkun83: These two methods should take the form of the corresponding methods in the builders.

testing/src/main/java/io/grpc/internal/testing/StatsTestUtils.java
```diff
@@ -93,39 +107,29 @@ public long getMetricAsLongOrFail(MeasurementDescriptor metricName) {
   }
 
   /**
-   * This tag will be propagated by {@link FakeStatsContextFactory} on the wire.
+   * This tag will be propagated by {@link FakeTagger} on the wire.
    */
   public static final TagKey EXTRA_TAG = TagKey.create("/rpc/test/extratag");
 
   private static final String EXTRA_TAG_HEADER_VALUE_PREFIX = "extratag:";
   private static final String NO_EXTRA_TAG_HEADER_VALUE_PREFIX = "noextratag";
```
zhangkun83: I wonder if this and its related logic can be removed.
It was needed because the old StatsContext didn't provide a way to tell if it's empty, thus the metadata was always sent. Now that you have implemented the optimization to skip sending metdata if TagContext is empty, we should no longer need the metadata form that represents "no tag".

zhangkun83:
> @zhangkun83 I refactored the code for overriding the OpenCensus implementation. There are two parts that I'm not sure about.
> * CensusStatsModule no longer caches the StatsClientInterceptor and ServerTracerFactory. It creates a new object whenever getServerTracerFactory or getClientInterceptor is called. Is that correct?

Yes it's correct. The caching wasn't necessary because it's only called per channel/server.

> * I'm not completely sure I used the right values for recordStats, especially in CensusModulesTest.

That sounds right. `recordStats` should be `true` in most of the tests.

testing/src/main/java/io/grpc/internal/testing/StatsTestUtils.java
sebright: I'm not sure I understand.  I didn't try to change any of the logic for serializing and deserializing tags.

zhangkun83: I meant if you remove this field and the if-branches that references this field, tests should not fail. Because of the following `if` you added to `CensusStatsModule`, those branches will never be executed now.
```java
        if (!module.tagger.empty().equals(parentCtx)) {
          headers.put(module.statsHeader, parentCtx);
        }
```

testing/src/main/java/io/grpc/internal/TestingAccessor.java
sebright: Done

testing/src/main/java/io/grpc/internal/testing/StatsTestUtils.java
zhangkun83: You are right ... the optimization was added before.

To keep it simple, `FakeTagContextBinarySerializer` is supposed to only handle `EXTRA_TAG`, and tests are supposed to only add `EXTRA_TAG`. Now `fromByteArray()` will fail if the metadata contains anything other than `EXTRA_TAG`. `toByteArray()` will fail too, by throwing NullPointerException -- it's probably better to make it throw `UnsupportedOperationException` if `extraTagValue==null`.

sebright: I made toByteArray throw an UnsupportedOperationException.

zhangkun83:
LGTM. Please merge master onto your branch and resolve the conflict.

sebright:
Thanks.  I merged my branch.

ejona86:
okay to test

sebright:
Thanks!

dinooliva:
Excellent!

PR #3649, opened 2017-11-01
stub: clean up call cancellation

carl-mastrangelo:
Changes:

* No longer catch throwable, and don't wrap Error
* log an error if call.cancel() throws
* Unwrap ExecutionExcetion  always
* include a description with cancellation when interrupted

### DIFF
--- a/stub/src/main/java/io/grpc/stub/ClientCalls.java
+++ b/stub/src/main/java/io/grpc/stub/ClientCalls.java
@@ -190,7 +190,6 @@ public final class ClientCalls {
    *     or an {@link InterruptedException}.
    */
   private static <V> V getUnchecked(Future<V> future) {
-
     try {
       return future.get();
     } catch (InterruptedException e) {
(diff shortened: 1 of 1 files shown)


stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -182,13 +190,17 @@ private ClientCalls() {}
    *     or an {@link InterruptedException}.
    */
   private static <V> V getUnchecked(Future<V> future) {
+
```
ejona86: nit: Remove line addition?

stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -214,6 +226,28 @@ private static StatusRuntimeException toStatusRuntimeException(Throwable t) {
...
+  /**
+   * Cancels a call, and throws the exception.
+   * 
+   * @param t must be a RuntimeException or Error
+   */
+  private static RuntimeException cancelThrow(ClientCall<?, ?> call, Throwable t) {
+    try {
+      call.cancel(null, t);
+    } catch (RuntimeException e) {
+      logger.log(Level.SEVERE, "RuntimeException encountered while closing call", e);
```
ejona86: Why not just catch Throwable? These code paths are basically the same.

carl-mastrangelo: I was hoping to (eventually) remove all the catch(throwable) in our code, and limit it only to the exceptions that can actually be thrown.  I can put it back the way it was if youd like, but it makes it harder to catch places where we do it wrong.  For example, we were wrapping Error above, previously.

ejona86: Previously we weren't try-catching around cancel(). That honestly still sounds fine, but if you want the try-catch... sure. But I don't see any need for catching each type individually here.

stub/src/main/java/io/grpc/stub/ClientCalls.java
carl-mastrangelo: Done

stub/src/main/java/io/grpc/stub/ClientCalls.java
carl-mastrangelo: Done, but added an assert comment.

PR #3652, opened 2017-11-02
stub: update ServerCalls to include status messages

carl-mastrangelo:
Additionally:

* Prefer RuntimeException from status
* remove redundant "static" on interfaces
* use checkState to shorten code

### DIFF
--- a/stub/src/main/java/io/grpc/stub/ServerCalls.java
+++ b/stub/src/main/java/io/grpc/stub/ServerCalls.java
@@ -17,7 +17,9 @@
 package io.grpc.stub;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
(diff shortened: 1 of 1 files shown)


stub/src/main/java/io/grpc/stub/ServerCalls.java
```diff
@@ -413,7 +406,7 @@ public static void asyncUnimplementedUnaryCall(MethodDescriptor<?, ?> methodDesc
     responseObserver.onError(Status.UNIMPLEMENTED
         .withDescription(String.format("Method %s is unimplemented",
             methodDescriptor.getFullMethodName()))
-        .asException());
+        .asRuntimeException());
```
dapengzhang0: What's the difference here? Should we always onError SRE?

carl-mastrangelo: I prefer SRE

dapengzhang0:
PS: better put reformat changes in a separate commit for review

carl-mastrangelo:
Changes that are too small for single commits are not a good idea IMO.  Especially since there is at least a day between sending out a PR and it actually getting reviewed.

ejona86:
> Changes that are too small for single commits are not a good idea IMO.

But having a 3 line change become 30 is noisy and the real change gets lost in the noise.

> Especially since there is at least a day between sending out a PR and it actually getting reviewed.

Just send them both out at the same time. No additional latency.

PR #3653, opened 2017-11-02
protobuf-lite: add javalite generated sources to intellij path

zpencer:
This seems to be an intellij-only problem, everything compiles and tests fine from the gradle CLI. Intellij complains that in `ProtoLiteUtilsTest`  com.google.protobuf.{Empty, Enum, Type} can not be found unless this is added.

Not sure if this one is related: https://github.com/grpc/grpc-java/pull/3598

### DIFF
--- a/protobuf-lite/build.gradle
+++ b/protobuf-lite/build.gradle
@@ -18,6 +18,11 @@ dependencies {
             libraries.guava
 
     testProtobuf libraries.protobuf
+    // TODO(zpencer): fix proto plugin wrt testProtobuf and intellij
+    // Using testProtobuf to add the protobuf libriaries did not add the built-in
+    // types like com.google.protobuf.Empty to intellij's class path.
+    // Adding testCompile is a workaround.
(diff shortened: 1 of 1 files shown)


protobuf-lite/build.gradle
```diff
@@ -18,6 +18,11 @@ dependencies {
             libraries.guava
 
     testProtobuf libraries.protobuf
+    // TODO(zpencer): fix proto plugin wrt testProtobuf and intellij
+    // Using testProtobuf to add the protobuf libriaries did not add the built-in
+    // types like com.google.protobuf.Empty to intellij's class path.
+    // Adding testCompile is a workaround.
+    testCompile libraries.protobuf
```
ejona86: This is dangerous, as protobuf and protobuf_lite contain some of the same classes, but with different versions.

zpencer: Valid point. I just discovered that these classes exist in `build/generated` so we should probably just be including them.

PR #3654, opened 2017-11-02
Allow tests to run on IPv6-less machines

ejona86:
Our Travis-CI builds are failing with `Protocol family unavailable` due
to the usage of ::1. Although it's 2017 and we'd expect to have ipv6
_loopback_ anywhere that mattered, apparently that's not the case.

The tests now work equally well on IPv4-only and IPv6-only machines.


ejona86:
retest this please

PR #3655, opened 2017-11-02
stub: make it harder to construct special client observers

carl-mastrangelo:
This is an alternative way to expose an abstract class.  The user is not expected to create these (instead they implement SteamObserver directly).  This allows us to add abstract methods and more extensions without risking breakage.  The API to construct it is intentionally difficult, using dates, words, and passing by object.

If other projects want to (correctly) construct these objects, we can expose an InternalCallStreamObserver, which will expose the key.

This pattern is used by JMH for Blackholes.

### DIFF
--- a/stub/src/main/java/io/grpc/stub/CallStreamObserver.java
+++ b/stub/src/main/java/io/grpc/stub/CallStreamObserver.java
@@ -95,4 +95,18 @@ public abstract class CallStreamObserver<V> implements StreamObserver<V> {
    * @param enable whether to enable compression.
    */
   public abstract void setMessageCompression(boolean enable);
+
+  static final Object KEY_FOR_20171102 =
+      "I know what I'm doing and make my own observers. Today's password is actinide.";
+
+
+  /**
+   * Do not call this, it is for the gRPC library to construct.
+   */
+  @Deprecated
+  public CallStreamObserver(Object key) {
+    if (!KEY_FOR_20171102.equals(key)) {
+      throw new IllegalArgumentException("Don't create your own CallStreamObservers");
+    }
+  }
 }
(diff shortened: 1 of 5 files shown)


carl-mastrangelo:
This is needed to make CallStreamObservers stable.  Their names and methods are stable, just not subclasses.

ejona86:
How is this different from just having a package-private constructor?

carl-mastrangelo:
I mentioned this:

> If other projects want to (correctly) construct these objects, we can expose an InternalCallStreamObserver, which will expose the key.

stub/src/main/java/io/grpc/stub/CallStreamObserver.java
```diff
@@ -95,4 +95,18 @@
...
+  static final Object KEY_FOR_20171102 =
+      "I know what I'm doing and make my own observers. Today's password is actinide.";
+
+
+  /**
+   * Do not call this, it is for the gRPC library to construct.
+   */
+  @Deprecated
+  public CallStreamObserver(Object key) {
+    if (!KEY_FOR_20171102.equals(key)) {
```
dapengzhang0: If use `==` here, then users are not able to create it outside io.grpc.stub package.

ejona86:
I don't really see the point. The abstract class structure we have now seems fine. Yes, we want to put final and lock down everything we can, but I'd rather it just be open instead of doing something like this. After all, we would use an interface here instead of abstract class if we were on Java 8.

carl-mastrangelo:
@ejona86 making this an interface, marking all the methods as non-experimental would have the same effect.  If we want to add more methods, we can extend the hierarchy, make the new class experimental, and add the methods there.

The problem with expecting users to cast is that if any of *their* APIs depend on these types, they need to construct them for tests and what not.  I'd rather them not get in the habit of casting, and let gRPC be the sole creators of these objects.  

This PR:

* backwards compatible with existing casting code
* prevents users from creating their own instances 
* allows us to safely add abstract methods 
* allows wrapper libraries to construct instances (via Internal accessor, not part of this PR).

ejona86:
I don't see the point of locking this down. If you're okay with an interface, then why aren't you okay with an abstract class? With Java 8 we'd be able to add methods with default implementations, which is the functionality we're trying to get with this abstract class.

carl-mastrangelo:
The goal is make the API stable.  There needs to be some way of making it maintainable.  This PR and the interface version are two ways of doing so.  I prefer the former, but recognize the latter.

PR #3656, opened 2017-11-02
core: retry support part 1, RetriableStream

dapengzhang0:
With an approach using a variant of `DelayedStream` - `RetriableStream`.

`RetriableStream` is a logical stream per call that can create a sequence of `substreams`, the physical streams for the retry attempts. It could also create a family of substreams in parallel in hedging case.
 
`RetriableStream` buffers, delays and executes stream operations before `passThrough`, where `passThrough` is meant by `commit` as well as the buffer `drained` by the winning substream.

This PR only implements buffering messages and replaying buffered messages when retry.
Retry policy, hedging, transparent retry, backoff, max buffer size and APIs that users can enable retry are not included in this PR.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ClientCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ClientCallImpl.java
@@ -113,6 +113,7 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
   /**
    * Provider of {@link ClientTransport}s.
    */
+  // TODO(zdapeng): replace the two APIs with a single API: newStream()
   interface ClientTransportProvider {
     /**
      * Returns a transport for a new call.
@@ -121,7 +122,12 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
      */
     ClientTransport get(PickSubchannelArgs args);
 
-    DelayedClientTransport getDelayedTransport();
+    <ReqT> RetriableStream<ReqT> newRetriableStream(
+        MethodDescriptor<ReqT, ?> method,
+        CallOptions callOptions,
+        Metadata headers,
+        Context context);
+
   }
 
   ClientCallImpl<ReqT, RespT> setFullStreamDecompression(boolean fullStreamDecompression) {
@@ -225,8 +231,7 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
       updateTimeoutHeaders(effectiveDeadline, callOptions.getDeadline(),
           context.getDeadline(), headers);
       if (retryEnabled()) {
-        stream = new RetriableStream<ReqT>(
-            method, callOptions, headers, clientTransportProvider, context);
+        stream = clientTransportProvider.newRetriableStream(method, callOptions, headers, context);
       } else {
         ClientTransport transport = clientTransportProvider.get(
             new PickSubchannelArgsImpl(method, headers, callOptions));
--- a/core/src/main/java/io/grpc/internal/ClientCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ClientCallImpl.java
@@ -120,6 +120,8 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
      * @param args object containing call arguments.
      */
     ClientTransport get(PickSubchannelArgs args);
+
+    DelayedClientTransport getDelayedTransport();
   }
 
   ClientCallImpl<ReqT, RespT> setFullStreamDecompression(boolean fullStreamDecompression) {
@@ -222,13 +224,18 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
     if (!deadlineExceeded) {
       updateTimeoutHeaders(effectiveDeadline, callOptions.getDeadline(),
           context.getDeadline(), headers);
-      ClientTransport transport = clientTransportProvider.get(
-          new PickSubchannelArgsImpl(method, headers, callOptions));
-      Context origContext = context.attach();
-      try {
-        stream = transport.newStream(method, headers, callOptions);
-      } finally {
-        context.detach(origContext);
+      if (retryEnabled()) {
+        stream = new RetriableStream<ReqT>(
+            method, callOptions, headers, clientTransportProvider, context);
+      } else {
+        ClientTransport transport = clientTransportProvider.get(
+            new PickSubchannelArgsImpl(method, headers, callOptions));
+        Context origContext = context.attach();
+        try {
+          stream = transport.newStream(method, headers, callOptions);
+        } finally {
+          context.detach(origContext);
+        }
       }
     } else {
       stream = new FailingClientStream(DEADLINE_EXCEEDED);
@@ -267,6 +274,11 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
       // was cancelled.
       removeContextListenerAndCancelDeadlineFuture();
     }
+  }
+
+  // TODO: API plumbing to enable retry.
+  private boolean retryEnabled() {
+    return false;
   }
 
   /**
@@ -405,9 +417,15 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
     checkState(!cancelCalled, "call was cancelled");
     checkState(!halfCloseCalled, "call was half-closed");
     try {
-      // TODO(notcarl): Find out if messageIs needs to be closed.
-      InputStream messageIs = method.streamRequest(message);
-      stream.writeMessage(messageIs);
+      if (stream instanceof RetriableStream) {
+        @SuppressWarnings("unchecked")
+        RetriableStream<ReqT> retriableStream = ((RetriableStream<ReqT>) stream);
+        retriableStream.sendMessage(message);
+      } else {
+        // TODO(notcarl): Find out if messageIs needs to be closed.
+        InputStream messageIs = method.streamRequest(message);
+        stream.writeMessage(messageIs);
+      }
     } catch (Throwable e) {
       stream.cancel(Status.CANCELLED.withCause(e).withDescription("Failed to stream message"));
       return;
--- a/core/src/main/java/io/grpc/internal/DelayedClientTransport.java
+++ b/core/src/main/java/io/grpc/internal/DelayedClientTransport.java
@@ -279,6 +279,9 @@ final class DelayedClientTransport implements ManagedClientTransport {
 
   final void addUncommittedRetriableStream(RetriableStream<?> retriableStream) {
     synchronized (lock) {
+      if (shutdownStatus != null) {
+        return;
+      }
       uncommittedRetriableStreams.add(retriableStream);
       if (getPendingStreamsCount() == 1) {
         channelExecutor.executeLater(reportTransportInUse);
--- a/core/src/main/java/io/grpc/internal/DelayedClientTransport.java
+++ b/core/src/main/java/io/grpc/internal/DelayedClientTransport.java
@@ -277,15 +277,21 @@ final class DelayedClientTransport implements ManagedClientTransport {
     }
   }
 
-  final void addUncommittedRetriableStream(RetriableStream<?> retriableStream) {
+  /**
+   * Registers a RetriableStream and return null if not shutdown, otherwise just returns the
+   * shutdown Status.
+   */
+  @Nullable
+  final Status addUncommittedRetriableStream(RetriableStream<?> retriableStream) {
     synchronized (lock) {
       if (shutdownStatus != null) {
-        return;
+        return shutdownStatus;
       }
       uncommittedRetriableStreams.add(retriableStream);
       if (getPendingStreamsCount() == 1) {
         channelExecutor.executeLater(reportTransportInUse);
       }
+      return null;
     }
   }
 
--- a/core/src/main/java/io/grpc/internal/DelayedClientTransport.java
+++ b/core/src/main/java/io/grpc/internal/DelayedClientTransport.java
@@ -29,6 +29,7 @@ import io.grpc.Status;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
@@ -62,8 +63,11 @@ final class DelayedClientTransport implements ManagedClientTransport {
   @GuardedBy("lock")
   private Collection<PendingStream> pendingStreams = new LinkedHashSet<PendingStream>();
 
+  @GuardedBy("lock")
+  private Collection<ClientStream> uncommittedRetriableStreams = new HashSet<ClientStream>();
+
   /**
-   * When shutdownStatus != null and pendingStreams.isEmpty(), then the transport is considered
+   * When {@code shutdownStatus != null && !hasPendingStreams()}, then the transport is considered
    * terminated.
    */
   @GuardedBy("lock")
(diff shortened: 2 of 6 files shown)


core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -0,0 +1,675 @@
...
+  private boolean shouldRetry() {
+    return false;
+  }
+
+  private interface BufferEntry {
+    /** Replays the buffer entry with the given stream. */
+    void runWith(ClientStream substream);
+  }
+
+  private static final class RequestEntry implements BufferEntry {
```
ejona86: :-/ This is a lot of boilerplate and splits the two copies of the method call. If you're doing it just for giving names to these classes, you can still do that in-line in each method. Defining the class in-line (and not anonymous) in the method still allows you to capture final fields.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -0,0 +1,675 @@
...
+
+    /** True if all existing buffered callbacks are drained. */
+    volatile boolean upToDate;
+
+    volatile boolean cancelled;
+
+    /** Number of the callbacks in the buffer that have been replayed. */
+    int index = 0;
+
+    List<BufferEntry> buffer = sharedBuffer;
```
ejona86: I didn't think this was going to be necessary. We should maybe discuss why it's needed.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -0,0 +1,675 @@
...
+  }
+
+  /** Requests if committed, or buffers it first and then requests on the current substream. */
+  @Override
+  public void request(int numMessages) {
+    synchronized (lock) {
+      if (!committed) {
+        sharedBuffer.add(new RequestEntry(numMessages));
+      }
+      for (Substream substream : substreams) {
```
ejona86: I don't think we should be calling the substream with the lock held, as it's asking for a deadlock. We should discuss the locking probably in person.

dapengzhang0: For normal retry, it can be pulled out of the sync block easily by saving a copy ref of the collection in the sync block. For hedging, there are some annoying race conditions.

```
  public void request(int numMessages) {
    // Works for non-hedging case

    Collection<Substream> substreamsCopy;

    synchronized (lock) {
      substreamsCopy = substreams;
      if (!committed) {
        sharedBuffer.add(new RequestEntry(numMessages));
      }
    }

    for (Substream substream : substreamsCopy) {
      substream.request(numMessages);
    }
  }
```

dapengzhang0: uh... I take it back. It is subtle even in normal retry case.

dapengzhang0: Maybe run `substream.delayOrExecute(bufferEntry)` directly and check the tail of the buffer can solve the problem?

```java
  public void request(int numMessages) {
    BufferEntry bufferEntry = new RequestEntry(numMessages);
    Collection<Substream> substreamsCopy;

    synchronized (lock) {
      substreamsCopy = substreams;
      if (!committed) {
        sharedBuffer.add(bufferEntry);
      }
    }

    for (Substream substream : substreamsCopy) {
      substream.delayOrExecute(bufferEntry);
    }
  }
``` 

```java
    void delayOrExecute(BufferEntry bufferEntry) {
      synchronized (lock) {
        if (!upToDate) { // Draining is going on.
          if (committed) {
            buffer.add(bufferEntry);
          }
          // else bufferEntry is added in the sharedBuffer already
          return;
        }
        if (buffer.tail() == bufferEntry) {
          return;
        }
      }
      bufferEntry.runWith(stream);
    }
```

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -120,6 +120,8 @@ public void cancelled(Context context) {
      * @param args object containing call arguments.
      */
     ClientTransport get(PickSubchannelArgs args);
+
+    DelayedClientTransport getDelayedTransport();
```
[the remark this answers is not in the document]
dapengzhang0: This was to minimize the change on old code path. `newRetriableStream(params...)` is a good idea if we completely switch to the retry code path.

dapengzhang0:
PLTA

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -0,0 +1,555 @@
...
+        if (savedState.winningSubstream != null && savedState.winningSubstream != substream) {
+          // committed but not me
+          substream.cancel(Status.CANCELLED.withDescription(CANCELLED_BECAUSE_HEDGE_COMMITTED));
+          return;
+        }
+        bufferEntry.runWith(substream);
+      }
+
+      synchronized (lock) {
+        if (index != state.buffer.size()) { // I'm not drained
```
ejona86: Re-load savedState and use it.

I actually think this synchronized block can integrated with the earlier initial synchronized block.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -0,0 +1,555 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import javax.annotation.CheckReturnValue;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
+
+/** A logical {@link ClientStream} that is retriable. */
+final class RetriableStream<ReqT> implements ClientStream {
+  private static final String CANCELLED_BECAUSE_HEDGE_COMMITTED =
```
ejona86: Make this a `Status`?

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
```diff
@@ -249,14 +266,40 @@ public final void shutdownNow(Status status) {
...
 
   @VisibleForTesting
   final int getPendingStreamsCount() {
     synchronized (lock) {
-      return pendingStreams.size();
+      return pendingStreams.size() + uncommittedRetriableStreams.size();
+    }
+  }
+
+  final void addUncommittedRetriableStream(RetriableStream<?> retriableStream) {
```
ejona86: Somewhere you need to handle the shutdown case.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -0,0 +1,555 @@
...
+      public void runWith(ClientStream substream) {
+        substream.setMaxOutboundMessageSize(maxSize);
+      }
+    }
+
+    delayOrExecute(new MaxOutboundMessageSizeEntry());
+  }
+
+  @Override
+  public Attributes getAttributes() {
```
ejona86: This may [only be called after the application is called with onHeaders or onClose](https://github.com/grpc/grpc-java/blob/v1.7.0/core/src/main/java/io/grpc/ClientCall.java#L242). So this should simply use winningSubstream, and either throw or use EMPTY otherwise.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -0,0 +1,555 @@
...
+      }
+    }
+
+    @Override
+    public void closed(Status status, Metadata trailers) {
+      if (shouldRetry()) {
+        // TODO(zdapeng): backoff and schedule; retry() should run in an executor
+        retry();
+      } else if (!hasHedging()) {
+        delayedClientTransport.removeUncommittedRetriableStream(RetriableStream.this);
```
ejona86: Why not `commit(substream)`? It seems like `state` should be updated.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -0,0 +1,555 @@
...
+
+    ClientStream substream = newSubstream();
+    drain(substream);
+
+    // TODO(zdapeng): schedule hedging if needed
+  }
+
+  @Override
+  public void cancel(Status reason) {
+    delayedClientTransport.removeUncommittedRetriableStream(this);
```
ejona86: I think this needs to be delayed until after `state` is updated.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -0,0 +1,555 @@
...
+    }
+
+    ClientStream substream = newSubstream();
+    drain(substream);
+
+    // TODO(zdapeng): schedule hedging if needed
+  }
+
+  @Override
+  public void cancel(Status reason) {
```
ejona86: It seems like this should close masterListener, at least in some cases (like if we are between retry-attempts).

core/src/main/java/io/grpc/internal/ClientCallImpl.java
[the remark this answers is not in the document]
ejona86: I don't really understand this. Although I'd also accept adding a `newRetriableStream` method to `DelayedClientTransport`. It's just weird how the two are glued together.

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
dapengzhang0: Done

core/src/main/java/io/grpc/internal/ClientCallImpl.java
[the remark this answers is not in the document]
dapengzhang0: Done.

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
ejona86: Just adding the if is not enough, as it doesn't impact the stream's state at all. When the channel is shutdown new RPCs should fail. The current `if` just ignores the fact that the channel is shutdown. Instead you need some behavior that will trigger `masterListener.closed()`. As a starting point you could return a boolean here.

core/src/main/java/io/grpc/internal/OobChannel.java
```diff
@@ -72,8 +74,9 @@ public ClientTransport get(PickSubchannelArgs args) {
     }
 
     @Override
-    public DelayedClientTransport getDelayedTransport() {
-      return delayedTransport;
+    public <ReqT> RetriableStream<ReqT> newRetriableStream(MethodDescriptor<ReqT, ?> method,
+        CallOptions callOptions, Metadata headers, Context context) {
+      return null;
```
ejona86: Not throws? And comment saying why it isn't implemented (like how it's impossible to enable)?

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -0,0 +1,541 @@
...
+
+    @Override
+    public void headersRead(Metadata headers) {
+      if (commit(substream)) {
+        masterListener.headersRead(headers);
+      }
+    }
+
+    @Override
+    public void closed(final Status status, final Metadata trailers) {
```
ejona86: Why the `final`s?

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -0,0 +1,540 @@
...
+  @CheckReturnValue
+  @Nullable
+  abstract Status prestart();
+
+  /** Starts the first PRC attempt. */
+  @Override
+  public final void start(ClientStreamListener listener) {
+    Status shutdownStatus = prestart();
+
+    masterListener = listener;
```
ejona86: Assign this first, before prestart(). Today, if shutdownNow() is called between prestart() and masterListener being set, then cancel() as called by DelayedClientTransport will throw a NPE

core/src/main/java/io/grpc/internal/OobChannel.java
dapengzhang0: Throwed.

core/src/main/java/io/grpc/internal/RetriableStream.java
dapengzhang0: Removed.

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
dapengzhang0: Return the shutdown and propagate it to `prestart()`. Didn't implement `RetriableStream` inside `DelayedClientTransport` because `newSubstream()` seems not easy here.

core/src/main/java/io/grpc/internal/RetriableStream.java
dapengzhang0: I'm not calling `cancel()` in `DelayedClientTransport`. I call `cancel()` after `masterListener` being set.

ejona86: What? Isn't this a call here:
https://github.com/grpc/grpc-java/pull/3656/files#diff-188e8f126b986cb2849a7f4c9230ecbcR259

dapengzhang0: oh, right. I misunderstood you.

dapengzhang0:
Test added. PTALA

dapengzhang0:
TODO: Tests for covering changes in `ManagedChannelImpl` and `DelayedClientTransport` including shutdown cases will be postponed until retry policy is implemented because the API to enable retry is subject to change.

core/src/test/java/io/grpc/internal/RetriableStreamTest.java
```diff
@@ -0,0 +1,557 @@
...
+  private final MethodDescriptor<String, String> method =
+      MethodDescriptor.<String, String>newBuilder().setType(MethodType.BIDI_STREAMING)
+          .setFullMethodName(MethodDescriptor.generateFullMethodName("service_foo", "method_bar"))
+          .setRequestMarshaller(new StringMarshaller())
+          .setResponseMarshaller(new StringMarshaller()).build();
+  private final ClientStreamListener masterListener = mock(ClientStreamListener.class);
+
+  // TODO(zdapeng): for mockito 2.7+: instead of spy(an_impl),
+  // use mock(RetriableStream.class, withSettings().useConstructorArgs(method))
+  private final RetriableStream<String> retriableStream = spy(new RetriableStream<String>(method) {
```
ejona86: Whoa. That's scary. We should not use a spy on the code under test. Spy does evilness that means I wouldn't trust results.

core/src/test/java/io/grpc/internal/RetriableStreamTest.java
```diff
@@ -0,0 +1,557 @@
...
+
+/**
+ * Unit tests for {@link RetriableStream}.
+ */
+@RunWith(JUnit4.class)
+public class RetriableStreamTest {
+  private static final String CANCELLED_BECAUSE_COMMITTED =
+      "Stream thrown away because RetriableStream committed";
+  private static final String AUTHORITY = "fakeAuthority";
+  private static final Compressor COMPRESSOR = mock(Compressor.class);
```
ejona86: Codec.Identity.NONE?

core/src/test/java/io/grpc/internal/RetriableStreamTest.java
```diff
@@ -0,0 +1,557 @@
...
+
+    // retry1
+    ClientStream mockStream2 = mock(ClientStream.class);
+    setNextSubstream(mockStream2);
+    inOrder = inOrder(retriableStream, masterListener, mockStream1, mockStream2);
+    setShouldRetry(true);
+    sublistenerCaptor1.getValue().closed(Status.UNAVAILABLE, new Metadata());
+
+    inOrder.verify(retriableStream).shouldRetry();
+    // TODO(zdapeng): send more messages dureing backoff, then forward backoff ticker
```
ejona86: /dureing/during/. Ditto elsewhere

dapengzhang0:
PTALA thanks

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -0,0 +1,542 @@
...
+
+    @Override
+    public void headersRead(Metadata headers) {
+      if (commit(substream)) {
+        masterListener.headersRead(headers);
+      }
+    }
+
+    @Override
+    public void closed(Status status, Metadata trailers) {
```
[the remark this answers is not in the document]
dapengzhang0: Will do it in a follow-up PR.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -0,0 +1,542 @@
...
+      if (shouldRetry()) {
```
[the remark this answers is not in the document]

PR #3658, opened 2017-11-02
core: ShardedAtomicLongCounter randomized thread hashes

zpencer:
Randomly pick array elements similar to Striped64:
https://github.com/google/guava/blob/0c434ff088d811e34064c6a0720d45962492f090/guava/src/com/google/common/cache/Striped64.java#L126


ejona86:
I think you're misunderstanding the purpose of the value. It is used to _change_ a thread's hash when there is a conflict. This isn't re-choosing a thread's hash code, so the usefulness is questionable.

PR #3659, opened 2017-11-03
 core: add reflection based LongAdder

zpencer:


core/src/main/java/io/grpc/internal/ReflectionLongAdderCounter.java
```diff
@@ -0,0 +1,111 @@
...
+import com.google.common.base.Preconditions;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+/**
+ * A {@link LongCounter} that is implemented with a JDK8 {@link LongAdder}. Instantiates the object
+ * and invokes methods reflectively to avoid a compile time dependency on LongAdder.
+ */
+public class ReflectionLongAdderCounter implements LongCounter {
```
carl-mastrangelo: final

core/src/main/java/io/grpc/internal/ReflectionLongAdderCounter.java
```diff
@@ -0,0 +1,111 @@
...
+        if (ctor.getParameterTypes().length == 0) {
+          defaultConstructorLookup = ctor;
+          break;
+        }
+      }
+    } catch (ClassNotFoundException e) {
+      caught = e;
+    } catch (NoSuchMethodException e) {
+      caught = e;
+    }
```
carl-mastrangelo: You need to catch Error, RuntimeException and the other exceptions here.  NoClassDefError can be thrown too.  Same with SecurityException.

carl-mastrangelo: Also, you should probably store the error, for throwing later in the ReflectionLongAdderCounter ctor

core/src/test/java/io/grpc/internal/ReflectionLongAdderCounterTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class ReflectionLongAdderCounterTest {
+  private ReflectionLongAdderCounter counter;
+
+  @Before
+  public void setUp() {
+    counter = new ReflectionLongAdderCounter();
```
carl-mastrangelo: init it inline above?

core/src/main/java/io/grpc/internal/ReflectionLongAdderCounter.java
ejona86: NoClassDefError can't be thrown here. That's when compiled code references a missing class.

core/src/main/java/io/grpc/internal/ReflectionLongAdderCounter.java
zpencer: done

core/src/main/java/io/grpc/internal/ReflectionLongAdderCounter.java
zpencer: now catching all Throwables

zpencer: done

core/src/test/java/io/grpc/internal/ReflectionLongAdderCounterTest.java
zpencer: This `@Before` creates a new counter at each test iteration

carl-mastrangelo: I think there is a new instance created each test too

zpencer: I didn't know junit does that, TIL

core/src/main/java/io/grpc/internal/ReflectionLongAdderCounter.java
carl-mastrangelo: Hmm, I dont see the changes up here

zpencer: Ah my mistake, my `git push` must have pushed it to my workstation rather than github.

core/src/test/java/io/grpc/internal/ReflectionLongAdderCounterTest.java
ejona86: That's the number one benefit to JUnit 4 to me. JUnit 3 didn't behave that way.

core/src/main/java/io/grpc/internal/ReflectionLongAdderCounter.java
```diff
@@ -0,0 +1,114 @@
...
+
+      Constructor<?>[] constructors = klass.getConstructors();
+      for (Constructor<?> ctor : constructors) {
+        if (ctor.getParameterTypes().length == 0) {
+          defaultConstructorLookup = ctor;
+          break;
+        }
+      }
+    } catch (Throwable e) {
+      caught = e;
```
carl-mastrangelo: optional: add a fine level log here.  We normally expect this to exist, even on Android.

PR #3661, opened 2017-11-03
core,netty,okhttp: move transport tracer outbound counters to transport thread

zpencer:
This lets us remove synchronization for `lastMessageSentTimeNanos` and `messagesSent` in the TransportTracer.

### DIFF
--- a/core/src/main/java/io/grpc/internal/MessageFramer.java
+++ b/core/src/main/java/io/grpc/internal/MessageFramer.java
@@ -55,8 +55,13 @@ public class MessageFramer implements Framer {
      *              closed and there is no data to deliver.
      * @param endOfStream whether the frame is the last one for the GRPC stream
      * @param flush {@code true} if more data may not be arriving soon
+     * @param numMessages the number of messages that this series of frames represents
      */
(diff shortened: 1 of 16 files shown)


core/src/main/java/io/grpc/internal/MessageFramer.java
```diff
@@ -249,11 +247,12 @@ private void writeBufferChain(BufferChainOutputStream bufferChain, boolean compr
     // Note that we are always delivering a small message to the transport here which
     // may incur transport framing overhead as it may be sent separately to the contents
     // of the GRPC frame.
-    sink.deliverFrame(writeableHeader, false, false);
+    sink.deliverFrame(writeableHeader, false, false, messagesBuffered);
```
ejona86: This mis-accounts the message, since messagesBuffered was already incremented for this message but this message won't be completed now. Although that doesn't really matter, it could be confusing. Seems like it deserves a comment or pass `messagesBuffered - 1` and reset with `messagesBuffered = 1`.

zpencer: I'll go with the latter, since this is meant to aid in debugging we want to be more precise.

PR #3662, opened 2017-11-03
core,netty,okhttp: Add getTransportStats method to ClientTransport

zpencer:
This merely adds the method but does not do anything yet on the implementations.

### DIFF
--- a/core/src/main/java/io/grpc/internal/DelayedClientTransport.java
+++ b/core/src/main/java/io/grpc/internal/DelayedClientTransport.java
@@ -17,6 +17,7 @@
 package io.grpc.internal;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.util.concurrent.SettableFuture;
 import io.grpc.CallOptions;
 import io.grpc.Context;
 import io.grpc.LoadBalancer.PickResult;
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/grpc/internal/DelayedClientTransport.java
```diff
@@ -185,6 +187,13 @@ public final void ping(final PingCallback callback, Executor executor) {
     throw new UnsupportedOperationException("This method is not expected to be called");
   }
 
+  @Override
+  public Future<TransportTracer.Stats> getTransportStats() {
+    SettableFuture<TransportTracer.Stats> ret = SettableFuture.create();
+    ret.set(null);
+    return ret;
```
carl-mastrangelo: Futures.immediateFuture(null)

core/src/main/java/io/grpc/internal/FailingClientTransport.java
```diff
@@ -51,4 +53,11 @@ public void ping(final PingCallback callback, Executor executor) {
         }
       });
   }
+
+  @Override
+  public Future<TransportTracer.Stats> getTransportStats() {
+    SettableFuture<TransportTracer.Stats> ret = SettableFuture.create();
+    ret.set(null);
```
carl-mastrangelo: same comment

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
zpencer: Futures is `@Beta` class so we can't use it.

zpencer: @carl-mastrangelo PTAL if you feel strongly about using immediateFuture, it can be copied into our internal utils class.

Issue #3547, opened 2017-10-06
Interoptest failing in GAE


zpencer:
Fails for both jdk7 and jdk8:
```
  maxInboundSize_exact(io.grpc.testing.integration.NettyClientInteropServlet$Tester)
  io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: io.grpc.netty.NettyClientTransport$3: Frame size 7 exceeds maximum: 5. 
        at io.grpc.Status.asRuntimeException(Status.java:526)
        at io.grpc.stub.ClientCalls$BlockingResponseStream.hasNext(ClientCalls.java:523)
        at io.grpc.stub.ClientCalls$BlockingResponseStream.next(ClientCalls.java:530)
        at io.grpc.testing.integration.AbstractInteropTest.maxInboundSize_exact(AbstractInteropTest.java:1116)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
        at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
        at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
        at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.lang.Thread.run(Thread.java:745)
```

ejona86:
CC @carl-mastrangelo

zpencer:
@carl-mastrangelo I recall you said the output sizes are not deterministic. Is it valid to disable the test for this environment?

PR #3663, opened 2017-11-03
interop-testing:  make max size tests more deterministic

carl-mastrangelo:
This change removes some of the non deterministic size outputs for
proto.  Instead of serializing the message and measuring the size,
measure before serialization.  As long as the remote always returns
using the same encoder, this should be stable.

Fixes: #3547

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -1187,7 +1187,7 @@ public abstract class AbstractInteropTest {
         md.toBuilder(mar, md.getResponseMarshaller()).build(),
         blockingStub.getCallOptions(),
         request)
-            .next();
+        .next();
 
     int size = mar.lastOutSize;
 
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -40,6 +40,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
+import com.google.common.io.ByteStreams;
 import com.google.common.util.concurrent.SettableFuture;
 import com.google.instrumentation.stats.RpcConstants;
 import com.google.instrumentation.stats.StatsContextFactory;
@@ -99,6 +100,7 @@ import io.grpc.testing.integration.Messages.StreamingOutputCallResponse;
 import io.opencensus.trace.Span;
 import io.opencensus.trace.unsafe.ContextUtils;
 import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.security.cert.Certificate;
@@ -1128,7 +1130,7 @@ public abstract class AbstractInteropTest {
         md.toBuilder(md.getRequestMarshaller(), mar).build(),
         blockingStub.getCallOptions(),
         request)
-            .next();
+        .next();
 
     int size = mar.lastInSize;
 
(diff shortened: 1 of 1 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1966,4 +2013,53 @@ private static ServerInterceptor recordContextInterceptor(
...
+    volatile int lastInSize;
+
+    ByteSizeMarshaller(MethodDescriptor.Marshaller<T> delegate) {
+      this.delegate = delegate;
+    }
+
+    @Override
+    public InputStream stream(T value) {
+      InputStream is = delegate.stream(value);
+      byte[] dst = new byte[16 * 1024 * 1024];
```
ejona86: Use Guava (like CountingInputStream) or our own IoUtils.toByteArray?

(Note that the volatiles aren't necessary since the RPC has completed and the response was returned.)

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1116,7 +1118,19 @@ public void maxInboundSize_exact() {
...
+    MethodDescriptor<StreamingOutputCallRequest, StreamingOutputCallResponse> md =
+        TestServiceGrpc.getStreamingOutputCallMethod();
+    ByteSizeMarshaller<StreamingOutputCallResponse> mar =
+        new ByteSizeMarshaller<StreamingOutputCallResponse>(md.getResponseMarshaller());
+    blockingServerStreamingCall(
+        blockingStub.getChannel(),
+        md.toBuilder(md.getRequestMarshaller(), mar).build(),
+        blockingStub.getCallOptions(),
+        request)
+            .next();
```
zpencer: nit: the whitespace here and on 1188 don't match 1156 and 1215. My IDE seems to think 1156 is the correct one.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
zpencer: CountingInputStream is beta, but it since it's a testing package it should be fine.

carl-mastrangelo: The ideal thin would be ByteStreams.copy, but thats beta too.  I can use it if its okay.

ejona86: It's fine to use `@Beta` APIs in tests.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
carl-mastrangelo: Done.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
carl-mastrangelo: Done.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1146,29 +1174,50 @@ public void maxInboundSize_tooBig() {
...
+    MethodDescriptor<StreamingOutputCallRequest, StreamingOutputCallResponse> md =
+        TestServiceGrpc.getStreamingOutputCallMethod();
+    ByteSizeMarshaller<StreamingOutputCallRequest> mar =
+        new ByteSizeMarshaller<StreamingOutputCallRequest>(md.getRequestMarshaller());
+    blockingServerStreamingCall(
+        blockingStub.getChannel(),
+        md.toBuilder(mar, md.getResponseMarshaller()).build(),
+        blockingStub.getCallOptions(),
+        request)
+            .next();
```
zpencer: This one still looks off

carl-mastrangelo: Done

PR #3667, opened 2017-11-06
netty: bump correct value of outbound msg counter

zpencer:
Bugfix for 9fac15d4f8fb09106fc4a80d761afbe1594b4d0c which always bumps the value by 1. Also remove msgCount from SendGrpcFrameCommand because it is no longer necessary.

### DIFF
--- a/core/src/main/java/io/grpc/internal/TransportTracer.java
+++ b/core/src/main/java/io/grpc/internal/TransportTracer.java
@@ -76,7 +76,7 @@ public final class TransportTracer {
   }
 
   /**
-   * Reports that some messages were successfully sent. {@code numMessages} must be at least 1.
+   * Reports that some messages were successfully sent. {@code numMessages} must be at least 0.
    */
   public void reportMessageSent(int numMessages) {
     if (numMessages == 0) {
--- a/core/src/main/java/io/grpc/internal/TransportTracer.java
+++ b/core/src/main/java/io/grpc/internal/TransportTracer.java
@@ -76,10 +76,11 @@ public final class TransportTracer {
   }
 
   /**
-   * Reports that a message was successfully sent. This method is thread safe.
+   * Reports that some messages were successfully sent. {@code numMessages} must be at least 1.
    */
-  public void reportMessageSent() {
-    messagesSent++;
+  public void reportMessageSent(int numMessages) {
+    Preconditions.checkArgument(numMessages > 0);
+    messagesSent += numMessages;
     lastMessageSentTimeNanos = currentTimeNanos();
   }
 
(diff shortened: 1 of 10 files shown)


netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -585,6 +585,7 @@ private void sendGrpcFrame(ChannelHandlerContext ctx, SendGrpcFrameCommand cmd,
     }
     // Call the base class to write the HTTP/2 DATA frame.
     encoder().writeData(ctx, cmd.streamId(), cmd.content(), 0, cmd.endStream(), promise);
+    transportTracer.reportMessageSent(cmd.getNumMessages());
```
ejona86: It is not guaranteed numMessages will be positive here. For large messages it will be zero for many frames. I'd suggest requiring non-negative instead. It looks like the previous code was simply broken.

ejona86: Note that the message has not been sent here. It is simply queued. If the remote is applying push-back, it could be a long time until the message is sent (which is why the previous code was delaying until the future completed).

zpencer: Changed to checking for `>= 0`

zpencer: Moved the incr back to the future callback.

zpencer:
@ejona86 PTAL

netty/src/main/java/io/grpc/netty/NettyClientStream.java
```diff
@@ -148,13 +149,14 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
       final int numBytes = bytebuf.readableBytes();
       if (numBytes > 0) {
         // Add the bytes to outbound flow control.
         onSendingBytes(numBytes);
         writeQueue.enqueue(
-            new SendGrpcFrameCommand(transportState(), bytebuf, endOfStream, numMessages),
+            new SendGrpcFrameCommand(transportState(), bytebuf, endOfStream),
             channel.newPromise().addListener(new ChannelFutureListener() {
               @Override
               public void operationComplete(ChannelFuture future) throws Exception {
```
ejona86: Forget to reportMessageSent in here?

core/src/main/java/io/grpc/internal/TransportTracer.java
```diff
@@ -76,10 +76,13 @@ public void reportStreamClosed(Status status) {
   }
 
   /**
-   * Reports that a message was successfully sent. This method is thread safe.
+   * Reports that some messages were successfully sent. {@code numMessages} must be at least 1.
```
ejona86: s/1/0/

netty/src/main/java/io/grpc/netty/NettyClientStream.java
zpencer: That will be added as a part of the diff to add TransportTracer to client side, right now the field is not defined yet.

core/src/main/java/io/grpc/internal/TransportTracer.java
zpencer: Fixed

PR #3669, opened 2017-11-06
protobuf-lite: add mark/reset support to ProtoInputStream

zpencer:
This will be used for binary logging, where we want the read the first
N bytes of the stream (where N is possibly all the
bytes). ProtoInputStream is backed by a ByteArrayInputStream which
supports mark/reset. This lets us avoid creating an extra buffer to
store the bytes we read.

### DIFF
--- a/protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoInputStream.java
+++ b/protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoInputStream.java
@@ -33,11 +33,12 @@ import javax.annotation.Nullable;
 class ProtoInputStream extends InputStream implements Drainable, KnownLength {
 
   // ProtoInputStream is first initialized with a *message*. *partial* is initially null.
-  // Once there has been a read operation on this stream, *message* is serialized to *partial* and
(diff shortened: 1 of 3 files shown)


protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoInputStream.java
```diff
@@ -112,6 +113,35 @@ public int available() throws IOException {
...
+
+  @Override
+  public void reset() throws IOException {
+    if (!marked) {
+      throw new IOException("mark was never called");
+    }
+    if (partial != null) {
+      partial.reset();
+    }
+    // If partial is null, then there is nothing left on the stream at the time mark() was called.
```
ejona86: `marked = false`

ejona86:
mark/reset is generally temperamental (can only be used one place at a time), and it may be hard/slower for other marshallers (and even protobuf in the future) to support it. Since there's not strict performance goals for binary logging, it may be better to avoid this for the while.

zpencer:
Closing for now.

PR #3670, opened 2017-11-06
core,netty: Add server negotiation timeout before client preface

ejona86:
The two commits here are easiest to review separately and will be committed separately. The first adds support for a negotiation timeout to core and the second improves transportReady callback placement in Netty so the negotiation timeout applies to more phases.

Before merging I'll reserve an issue for `negotiationTimeout`'s `@ExperimentalApi` and update the annotation.

### DIFF
--- a/core/src/main/java/io/grpc/ServerBuilder.java
+++ b/core/src/main/java/io/grpc/ServerBuilder.java
@@ -18,6 +18,7 @@ package io.grpc;
 
 import java.io.File;
 import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 
 /**
@@ -171,6 +172,20 @@ public abstract class ServerBuilder<T extends ServerBuilder<T>> {
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1704")
   public abstract T compressorRegistry(@Nullable CompressorRegistry registry);
+
+  /**
+   * Sets the permitted time for new connections to complete negotiation handshakes before being
+   * killed.
+   *
+   * @return this
+   * @throws IllegalArgumentException if timeout is negative
+   * @throws UnsupportedOperationException if unsupported
+   * @since 1.8.0
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/3706")
+  public T handshakeTimeout(long timeout, TimeUnit unit) {
+    throw new UnsupportedOperationException();
+  }
 
   /**
    * Builds a server using the given parameters.
--- a/core/src/main/java/io/grpc/internal/ServerImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerImpl.java
@@ -49,6 +49,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.concurrent.Executor;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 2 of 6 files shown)


core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -308,7 +311,9 @@ public ServerTransportListener transportCreated(ServerTransport transport) {
       synchronized (lock) {
         transports.add(transport);
       }
-      return new ServerTransportListenerImpl(transport);
+      ServerTransportListenerImpl stli = new ServerTransportListenerImpl(transport);
+      stli.init();
```
dapengzhang0: Is this the precise point that protocol negotiation starts, and transportReady the precise point that it ends? InProcessServer also does this? Does it make sense?

core/src/main/java/io/grpc/ServerBuilder.java
```diff
@@ -173,6 +174,18 @@ public T addStreamTracerFactory(ServerStreamTracer.Factory factory) {
...
   /**
+   * Sets the permitted time for new connections to complete negotiation before being killed. Does
+   * nothing if unsupported.
+   *
+   * @return this
+   * @since 1.8.0
+   */
+  @ExperimentalApi // FIXME
+  public T negotiationTimeout(long timeout, TimeUnit unit) {
+    return thisT();
```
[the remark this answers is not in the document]
ejona86: I'd either want to throw or do nothing. I'd be fine swapping to throwing.

core/src/main/java/io/grpc/internal/ServerImpl.java
ejona86: At this point a TCP connection has been established, but nothing else. transportReady is called before any RPCs are triggered.

core/src/main/java/io/grpc/ServerBuilder.java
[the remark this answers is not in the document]
ejona86: Swapped to throwing. Doesn't really change much since all our transports receive an implementation from AbstractServerImplBuilder.

ejona86:
PTAL

In addition to throwing, I renamed to `handshakeTimeout` to align with c-core.

netty/src/test/java/io/grpc/netty/NettyServerHandlerTest.java
```diff
@@ -196,6 +199,19 @@ public Void answer(InvocationOnMock invocation) throws Throwable {
   }
 
   @Test
+  public void transportReadyDelayedUntilConnectionPreface() throws Exception {
+    initChannel(new GrpcHttp2ServerHeadersDecoder(GrpcUtil.DEFAULT_MAX_HEADER_LIST_SIZE));
+
+    handler().handleProtocolNegotiationCompleted(Attributes.EMPTY);
+    verify(transportListener, never()).transportReady(any(Attributes.class));
+
```
dapengzhang0: s/any/isA/ ?

ejona86: They're equivalent here. But isA would actually be worse, because if I had the wrong class I could miss callbacks that I wanted to trigger failure. I'm not trying to verify it is an Attributes; I just checking the method which happens to have certain arguments.

dapengzhang0: Oh, this `verify` checks `never()`, I should have meant the second `verify`.

ejona86: Yeah, the second verify would fall into "I'm not trying to verify it is an Attributes".

PR #3674, opened 2017-11-07
java_grpc_library: add test for remote proto in srcs

ejona86:
I'm planning to wait for when we update to the next protobuf release before merging this.


carl-mastrangelo:
Can this proceed?

ejona86:
@carl-mastrangelo, we can proceed with this. Bazel build is broken until #3814, but I've tested this works after that change.

Issue #3676, opened 2017-11-07
GAE+jdk7 regression introduced by ProxyDetector


zpencer:
RPCs will just hang because `ProxyDetectorImpl` tries to use `ProxySelector`, resulting in this error:

```
  Caused by: java.lang.NoClassDefFoundError: java.net.ProxySelector is a restricted class. Please see the Google App Engine developer's guide for more details.
  	at com.google.apphosting.runtime.security.shared.stub.java.net.ProxySelector.<clinit>(ProxySelector.java)
  	at io.grpc.internal.ProxyDetectorImpl$2.get(ProxyDetectorImpl.java:69)
  	at io.grpc.internal.ProxyDetectorImpl$2.get(ProxyDetectorImpl.java:65)
  	at io.grpc.internal.ProxyDetectorImpl.detectProxy(ProxyDetectorImpl.java:140)
  	at io.grpc.internal.ProxyDetectorImpl.proxyFor(ProxyDetectorImpl.java:118)
  	at io.grpc.internal.InternalSubchannel.startNewTransport(InternalSubchannel.java:202)
  	at io.grpc.internal.InternalSubchannel.obtainActiveTransport(InternalSubchannel.java:183)
  	at io.grpc.internal.ManagedChannelImpl$SubchannelImpl.requestConnection(ManagedChannelImpl.java:928)
  	at io.grpc.PickFirstBalancerFactory$PickFirstBalancer.handleResolvedAddressGroups(PickFirstBalancerFactory.java:79)
  	at io.grpc.internal.ManagedChannelImpl$NameResolverListenerImpl$1.run(ManagedChannelImpl.java:834)
  	at io.grpc.internal.ChannelExecutor.drain(ChannelExecutor.java:72)
  	at io.grpc.internal.ManagedChannelImpl$LbHelperImpl.runSerialized(ManagedChannelImpl.java:800)
  	at io.grpc.internal.ManagedChannelImpl$NameResolverListenerImpl.onAddresses(ManagedChannelImpl.java:826)
  	at io.grpc.internal.DnsNameResolver$1.run(DnsNameResolver.java:173)
  	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1168)
  	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:605)
[...]

PR #3677, opened 2017-11-07
core: no ProxyDetector for GAE+JDK7

zpencer:
This is a workaround of the security restrictions, but effectively turns off proxy support in GAE+JDK7.

I can't seem to get AccessController to work in GAE, the checkPermission API fails:
```
Can not use checkPermission:
Caused by: java.lang.SecurityException: AccessController.checkPermission is unsupported.
```

And doPriviledged also fails:
```
  Caused by: io.grpc.StatusRuntimeException: INTERNAL: Thrown from handleResolvedAddresses(): java.lang.NoClassDefFoundError: java.net.ProxySelector is a restricted class. Please see the Google App Engine developer's guide for more details.
```

Let's use the usual technique of checking for the `IS_RESTRICTED_APPENGINE` flag.

Fixes https://github.com/grpc/grpc-java/issues/3676

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
@@ -116,8 +116,6 @@ public abstract class AbstractManagedChannelImplBuilder
   String authorityOverride;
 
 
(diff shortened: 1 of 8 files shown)


zpencer:
I do not see the proxy detector changes in 1.7.x so it's only 1.8.x that needs the backport.

PR #3682, opened 2017-11-07
netty: Avoid clearing SSLParameters

### LINKED ISSUE #3532: ProtocolNegotiators.TlsNegotiator overwrites existing SSLParameters

1.6.1

```
java version "9"
Java(TM) SE Runtime Environment (build 9+181)
Java HotSpot(TM) 64-Bit Server VM (build 9+181, mixed mode)
```

### What did you do?
If possible, provide a recipe for reproducing the error.

I'm using a Netty `ManagedChannelBuilder` and I set it to use an `SslContext` configured for `h2`. Netty's `SslContextBuilder` will configure the `Java9SslEngineWrapper` that calls the `SSLParameters#setApplicationProtocols` with the `{h2}` array value to enable the client to use the ALPN extension. The ALPN extension is not sent by the client because the `SSLParameters` is overwritten by  `ProtocolNegotiators.TlsNegotiator` and the existing `SSLParameters` are lost.

I expect that the current `SSLParameters` to be preserved.

### What did you see instead?

The client is not able to perform negotiate the `h2` protocol during the TLS handshake.

### Advocated fix

I installed locally a version of `io.grpc:grpc-netty` with the following change to `ProtocolNegotiators#TlsNegotiator` and it worked fine:

```
    @Override
    public Handler newHandler(GrpcHttp2ConnectionHandler handler) {
      ChannelHandler sslBootstrap = new ChannelHandlerAdapter() {
        @Override
          SSLEngine sslEngine = sslContext.newEngine(ctx.alloc(), host, port);
          SSLParameters sslParams = sslEngine.getSSLParameters();
          sslParams.setEndpointIdentificationAlgorithm("HTTPS");
          sslEngine.setSSLParameters(sslParams);
          ctx.pipeline().replace(this, null, new SslHandler(sslEngine, false));
        }
      };
      return new BufferUntilTlsNegotiatedHandler(sslBootstrap, handler);
    }
```

This implementation is more conservative and modifies the current `SSLParameters` which preserve the application protocols for Java 9.

ejona86:
Since Netty may have set some parameters already, we should modify the
existing SSLParameters instead of starting from scratch.

This may fix ALPN with JDK9, but full support for ALPN with JDK9 is
still later work and we're not supporting it yet.

Fixes #3532 

CC @vietj

### DIFF
--- a/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
+++ b/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
@@ -304,7 +304,7 @@ public final class ProtocolNegotiators {
         @Override
         public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
           SSLEngine sslEngine = sslContext.newEngine(ctx.alloc(), host, port);
-          SSLParameters sslParams = new SSLParameters();
+          SSLParameters sslParams = sslEngine.getSSLParameters();
           sslParams.setEndpointIdentificationAlgorithm("HTTPS");
           sslEngine.setSSLParameters(sslParams);
           ctx.pipeline().replace(this, null, new SslHandler(sslEngine, false));


vietj:
great, let me know when this is released!

ejona86:
@vietj, this will be in the 1.8 release scheduled for November 21st.

vietj:
good to know! thanks

PR #3683, opened 2017-11-08
core,netty: correctly count streams that ended due to client cancel

zpencer:
There is now a single place that bumps the counter.


zpencer:
synced offline: will update this PR to only change the counter in `TransportState.closeListener()`

Issue #3587, opened 2017-10-18
Open source the cronet unit tests


zpencer:
Today, we only have the cronet production classes but no unit tests.

ericgribkoff:
Closed by #3690 (and replaced by https://github.com/grpc/grpc-java/issues/3694).

PR #3690, opened 2017-11-08
cronet: add tests

ericgribkoff:
Addresses https://github.com/grpc/grpc-java/issues/3587. Manually verified the roboelectric tests build and run with Android Studio.


zpencer:
Can these run as a part of the usual `./gradlew build` without too much pain?

ericgribkoff:
Not via the approaches I've tried: we need to copy the Cronet binaries from Cronet's GCS account and these (AFAIK) only run on Android, and I haven't had any luck getting the Android library plugin to play nicely with our build system (i.e., mixing `apply plugin 'java'` and `apply plugin: 'com.android.library'`). Getting this building in open-source is planned but not high priority.

PR #3691, opened 2017-11-08
codecov: Restore GitHub status notifications

ejona86:
I'm not sure when they stopped, but Travis OS X builds being delayed is
likely the reason we didn't notice. 'changes: no' is apparently now the
default, so it doesn't need to be explicitly disabled any more. And
removing the config for 'changes' seems to restore the old statuses.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,8 +7,4 @@ coverage:
       default:
         target: auto
         threshold: 1
-    changes:
-      default:
-        branches:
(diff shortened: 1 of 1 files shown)


ejona86:
:-/ Why no statuses! It worked on my clone... https://github.com/ejona86/grpc-java/pull/22

ejona86:
Closing because it seems to be a github permission problem.

PR #3692, opened 2017-11-08
RELEASING.md: add test compiler outputs to VERSION_FILES

ericgribkoff:


RELEASING.md
```diff
@@ -64,6 +64,9 @@ $ MAJOR=1 MINOR=7 PATCH=0 # Set appropriately for new release
 $ VERSION_FILES=(
   build.gradle
   android-interop-testing/app/build.gradle
+  compiler/src/test/golden/TestService.java.txt
+  compiler/src/testLite/golden/TestService.java.txt
+  compiler/src/testNano/golden/TestService.java.txt
```
carl-mastrangelo: This probably wont help because they don't contain the magic string `CURRENT_GRPC_VERSION`

ericgribkoff: Good catch. Gave it another try for these files

PR #3695, opened 2017-11-09
cronet: allow multiple annotation objects attached to stream

ericgribkoff:
Cronet allows attaching multiple annotation objects to a bidirectional stream ([javadoc](https://cs.chromium.org/chromium/src/components/cronet/android/api/src/org/chromium/net/ExperimentalUrlRequest.java?type=cs&q=addRequestAnnotation&sq=package:chromium&l=40)). This PR deprecates the `CRONET_ANNOTATION_KEY`, which allowed at most one annotation object, and instead adds a `CronetCallOptions#withAnnotation` which lets users add annotation objects to a set of annotations attached to the Cronet stream, more closely matching the semantics exposed by Cronet.

This (or some alternative) API change is helpful to accommodate existing users of `CRONET_ANNOTATION_KEY` alongside a mechanism for exporting gRPC-specific metrics to Cronet's `RequestFinishedInfo.Listener` implementations using the same annotation mechanism, which seems to require multiple annotations attached to the Cronet stream.


cronet/src/main/java/io/grpc/cronet/CronetCallOptions.java
```diff
@@ -27,8 +29,34 @@ private CronetCallOptions() {}
...
+   * annotation objects. When an RPC is made using a {@link CallOptions} instance returned by this
+   * method, the annotation objects will be attached to the underlying Cronet bidirectional stream.
+   * When the stream finishes, the user can retrieve the annotation object(s) via {@link
+   * org.chromium.net.RequestFinishedInfo.Listener}.
+   *
+   * @param annotation the object to attach to the Cronet stream
+   */
+  public static CallOptions withAnnotation(CallOptions callOptions, Object annotation) {
+    Set<Object> annotations = callOptions.getOption(CRONET_ANNOTATIONS_KEY);
+    if (annotations != null) {
```
ejona86:
```java
if (annotations == null) {
  annotations = new HashSet<Object>();
}
annotations.add(annotation);
```

However, since options are mostly immutable, it may be strange this modifies the set. Notice that the `withOption` at the end isn't necessary if the key was already present.

cronet/src/main/java/io/grpc/cronet/CronetCallOptions.java
```diff
@@ -27,8 +29,34 @@ private CronetCallOptions() {}
...
+      annotations.add(annotation);
+    } else {
+      annotations = new HashSet<Object>();
```
ejona86: What happens if you don't use a set? (say, a list instead?)

cronet/src/main/java/io/grpc/cronet/CronetCallOptions.java
ericgribkoff: Switched to Collection backed by ArrayList (also matches how [Cronet handles the annotations](https://cs.chromium.org/chromium/src/components/cronet/android/java/src/org/chromium/net/impl/BidirectionalStreamBuilderImpl.java?type=cs&q=addRequestAnnotation&l=123)) and recreate a new ArrayList on each call. I think this better respects the immutability of call options; is this type of change what you had in mind?

cronet/src/main/java/io/grpc/cronet/CronetCallOptions.java
```diff
@@ -38,25 +38,26 @@ private CronetCallOptions() {}
...
+    if (existingAnnotations == null) {
+      newAnnotations = new ArrayList<Object>();
     } else {
-      annotations = new HashSet<Object>();
-      annotations.add(annotation);
+      newAnnotations = new ArrayList<Object>(existingAnnotations);
     }
-    return callOptions.withOption(CronetCallOptions.CRONET_ANNOTATIONS_KEY, annotations);
+    newAnnotations.add(annotation);
+    return callOptions.withOption(CronetCallOptions.CRONET_ANNOTATIONS_KEY, newAnnotations);
```
ejona86: You may consider wrapping with unmodifiableList.

cronet/src/main/java/io/grpc/cronet/CronetCallOptions.java
ejona86: Yeah, that's what I was thinking.

cronet/src/main/java/io/grpc/cronet/CronetCallOptions.java
ericgribkoff: Done

PR #3592, opened 2017-10-20
[do not review] Using this PR to execute GAE kokoro check

zpencer:

PR #3698, opened 2017-11-09
gae-interop-testing: create new instance per test for okhttp

zpencer:
Without this, the test is flakey for some test methods. The flakiness
is probably caused by AbstractInteropTest and not due to gRPC itself.


zpencer:
Possibly still flakey:
https://sponge.corp.google.com/target?id=3009f84b-ac47-488b-a364-fb686d9cc51b&target=grpc/java/master/gae-interop/presubmit/gae-interop&searchFor=&show=ALL&sortBy=STATUS

zpencer:
I ran this a bunch of times on kokoro and I can't seem to reproduce the flake again: https://github.com/grpc/grpc-java/pull/3592

PR #3699, opened 2017-11-09
core: initial binary log class

zpencer:
ServerCallImpl and ClientCallImpl will each have a BinaryLog
field. Each time the constructor is called, we will do a lookup
for the log. Performance is not critical for binlog, but if it
becomes a problem MethodDescriptor can have a setter to store
it as an Object (to avoid a depdency on `internal`).

The binary log class is a skeleton class at the moment, but does
contain the max header and message length info. The limits are
determined by parsing the shell variable GRPC_BINARY_LOG_CONFIG.

### DIFF
--- a/core/src/main/java/io/grpc/MethodDescriptor.java
+++ b/core/src/main/java/io/grpc/MethodDescriptor.java
@@ -19,7 +19,6 @@ package io.grpc;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.base.Preconditions;
-import io.grpc.internal.BinaryLog;
 import java.io.InputStream;
 import java.util.concurrent.atomic.AtomicReferenceArray;
 import javax.annotation.CheckReturnValue;
@@ -47,7 +46,6 @@ public final class MethodDescriptor<ReqT, RespT> {
   private final boolean idempotent;
   private final boolean safe;
   private final boolean sampledToLocalTracing;
-  private final BinaryLog binaryLog;
 
   // Must be set to InternalKnownTransport.values().length
   // Not referenced to break the dependency.
@@ -213,8 +211,7 @@ public final class MethodDescriptor<ReqT, RespT> {
       Marshaller<RequestT> requestMarshaller,
       Marshaller<ResponseT> responseMarshaller) {
     return new MethodDescriptor<RequestT, ResponseT>(
-        type, fullMethodName, requestMarshaller, responseMarshaller, null, false, false, false,
-        BinaryLog.getLog(fullMethodName));
+        type, fullMethodName, requestMarshaller, responseMarshaller, null, false, false, false);
   }
 
   private MethodDescriptor(
@@ -225,8 +222,7 @@ public final class MethodDescriptor<ReqT, RespT> {
       Object schemaDescriptor,
       boolean idempotent,
       boolean safe,
-      boolean sampledToLocalTracing,
-      BinaryLog binaryLog) {
+      boolean sampledToLocalTracing) {
 
     this.type = Preconditions.checkNotNull(type, "type");
     this.fullMethodName = Preconditions.checkNotNull(fullMethodName, "fullMethodName");
@@ -238,8 +234,6 @@ public final class MethodDescriptor<ReqT, RespT> {
     this.sampledToLocalTracing = sampledToLocalTracing;
     Preconditions.checkArgument(!safe || type == MethodType.UNARY,
         "Only unary methods can be specified safe");
-    // The binlog may be a noop log
-    this.binaryLog = Preconditions.checkNotNull(binaryLog);
   }
 
   /**
@@ -574,8 +568,7 @@ public final class MethodDescriptor<ReqT, RespT> {
           schemaDescriptor,
           idempotent,
           safe,
-          sampledToLocalTracing,
-          BinaryLog.getLog(fullMethodName));
+          sampledToLocalTracing);
     }
   }
 }
--- a/core/src/main/java/io/grpc/MethodDescriptor.java
+++ b/core/src/main/java/io/grpc/MethodDescriptor.java
@@ -19,6 +19,7 @@ package io.grpc;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.base.Preconditions;
+import io.grpc.internal.BinaryLog;
 import java.io.InputStream;
 import java.util.concurrent.atomic.AtomicReferenceArray;
 import javax.annotation.CheckReturnValue;
@@ -46,6 +47,7 @@ public final class MethodDescriptor<ReqT, RespT> {
   private final boolean idempotent;
   private final boolean safe;
   private final boolean sampledToLocalTracing;
+  private final BinaryLog binaryLog;
 
   // Must be set to InternalKnownTransport.values().length
   // Not referenced to break the dependency.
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -19,6 +19,7 @@
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.base.Preconditions;
+import io.grpc.internal.BinaryLog;
```
carl-mastrangelo: core cannot depend on internal

zpencer: Updated code and description.

core/src/main/java/io/grpc/internal/BinaryLog.java
```diff
@@ -0,0 +1,208 @@
...
+import java.util.Map;
+import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+
+/**
+ * A binary log class that is configured for a specific {@link MethodDescriptor}.
+ */
+public final class BinaryLog {
```
ejona86: There doesn't seem to be a need yet for public (since the change to MethodDescriptor). package-private?

core/src/main/java/io/grpc/internal/BinaryLog.java
```diff
@@ -0,0 +1,208 @@
...
+    // Regexes to extract per-binlog options
+    // The form: {m:256}
+    private static final Pattern msgRe = Pattern.compile("\\{m(?::(\\d+))?}");
+    // The form: {h:256}
+    private static final Pattern headerRe = Pattern.compile("\\{h(?::(\\d+))?}");
+    // The form: {h:256,m:256}
+    private static final Pattern bothRe = Pattern.compile("\\{h(?::(\\d+))?;m(?::(\\d+))?}");
+
+    private final boolean enabled;
+    private final Map<String, BinaryLog> perMethodLogs = new HashMap<String, BinaryLog>();
```
ejona86: Make these unmodifiable? At the very least it'd be nice to assign them at the end of the constructor, so it's more obvious no code is modifying them after they're set.

core/src/main/java/io/grpc/internal/BinaryLog.java
```diff
@@ -0,0 +1,208 @@
...
+        globalLog = NOOP_LOG;
+        enabled = false;
+        return;
+      }
+      String[] configurations = configurationString.split(",");
+      BinaryLog globalLog = null;
+      for (String configuration : configurations) {
+        Matcher configMatcher = configRe.matcher(configuration);
+        if (!configMatcher.matches()) {
+          throw new IllegalArgumentException("Bad input: " + configuration);
```
ejona86: Exceptions in this constructor that is initialized statically seem a bad idea. It seems it should be logged and the entry ignored.

core/src/main/java/io/grpc/internal/BinaryLog.java
```diff
@@ -0,0 +1,208 @@
...
+        throw new IllegalArgumentException("Illegal log config pattern: " + logConfig);
+      }
+      return new BinaryLog(maxHeaderBytes, maxMsgBytes);
+    }
+
+    /**
+     * Returns true if the input string is a glob of the form: {@code <package-service>/*}.
+     */
+    static boolean isServiceGlob(String input) {
+      return input.substring(input.lastIndexOf('/')).equals("/*");
```
ejona86: endsWith("/*") ?

core/src/main/java/io/grpc/internal/BinaryLog.java
```diff
@@ -0,0 +1,208 @@
...
+      }
+      Matcher headerMatcher;
+      Matcher msgMatcher;
+      Matcher bothMatcher;
+      final int maxHeaderBytes;
+      final int maxMsgBytes;
+      if ((headerMatcher = headerRe.matcher(logConfig)).matches()) {
+        maxMsgBytes = 0;
+        String maxHeaderStr = headerMatcher.group(1);
+        maxHeaderBytes = maxHeaderStr != null ? Integer.parseInt(maxHeaderStr) : Integer.MAX_VALUE;
```
ejona86: Also handle the parse exception cleanly. (Have a helper that returns -1, and then check at the end? Or catch outside this method.)

core/src/main/java/io/grpc/internal/BinaryLog.java
```diff
@@ -0,0 +1,208 @@
...
+      if (logConfig == null) {
+        return new BinaryLog(Integer.MAX_VALUE, Integer.MAX_VALUE);
+      }
+      Matcher headerMatcher;
+      Matcher msgMatcher;
+      Matcher bothMatcher;
+      final int maxHeaderBytes;
+      final int maxMsgBytes;
+      if ((headerMatcher = headerRe.matcher(logConfig)).matches()) {
+        maxMsgBytes = 0;
```
ejona86: Would you mind assigning these in the same order each time? So this would be the last line of this block.

core/src/main/java/io/grpc/internal/BinaryLog.java
zpencer: Originally I didn't think it was too wrong here to just let the constructor fail in static initialization and prevent app startup if the config string is malformed, since this naturally surfaces the error to the user. Same for the ambiguous case of duplicate log patterns. I can change it to be more graceful though.

core/src/main/java/io/grpc/internal/BinaryLog.java
zpencer: Done

core/src/main/java/io/grpc/internal/BinaryLog.java
zpencer: Done

core/src/main/java/io/grpc/internal/BinaryLog.java
zpencer: Now ignoring duplicate entries

core/src/main/java/io/grpc/internal/BinaryLog.java
zpencer: Switched to an unmodifiable map

core/src/main/java/io/grpc/internal/BinaryLog.java
zpencer: Changed to package-private

zpencer:
@ejona86 friendly ping

core/src/main/java/io/grpc/internal/BinaryLog.java
```diff
@@ -64,14 +67,31 @@ public String toString() {
         + "maxMessageBytes=" + maxMessageBytes + "]";
   }
 
-  private static final Factory DEFAULT_FACTORY =
-      new Factory(System.getenv("GRPC_BINARY_LOG_CONFIG"));
+  @Nullable
+  private static final Factory DEFAULT_FACTORY;
+
+  static {
+    Factory defaultFactory = null;
```
ejona86: Assign NOOP_LOG instead of `null`? (Also would need changing in the Throwable). Would remove the `if` from `getLog()` without introducing any complexity.

core/src/main/java/io/grpc/internal/BinaryLog.java
```diff
@@ -115,39 +133,44 @@ public static BinaryLog getLog(String fullMethodName) {
         String methodOrSvc = configMatcher.group(1);
         String binlogOptionStr = configMatcher.group(2);
         BinaryLog binLog = createBinaryLog(binlogOptionStr);
+        if (binLog == null) {
+          logger.log(Level.SEVERE, "Can not create binary log for: " + configuration);
```
ejona86: This will be a duplicate log entry, as `createBinaryLog()` also logs before returning `null`. Choose one.

core/src/main/java/io/grpc/internal/BinaryLog.java
ejona86: I'm fine with failing the user. But failing static initialization yields _weird_ results. Far too much confusion.

ejona86:
FYI: after you push changes, make a comment to say it's ready for re-review. 1) Just because you pushed changes doesn't mean it's ready for re-review and 2) pushed changes are emailed differently from comments, so it is filtered differently for me.

PR #3705, opened 2017-11-09
core,netty: wire TransportTracer to netty client

zpencer:
Also updated unit tests to verify client side stats in `AbstractTransportTest`

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractClientStream.java
+++ b/core/src/main/java/io/grpc/internal/AbstractClientStream.java
@@ -104,6 +104,9 @@ public abstract class AbstractClientStream extends AbstractStream
       Metadata headers,
       boolean useGet) {
     Preconditions.checkNotNull(headers, "headers");
+    if (transportTracer != null) {
+      transportTracer.reportStreamStarted();
+    }
     this.useGet = useGet;
     if (!useGet) {
       framer = new MessageFramer(this, bufferAllocator, statsTraceCtx);
@@ -193,6 +196,8 @@ public abstract class AbstractClientStream extends AbstractStream
   protected abstract static class TransportState extends AbstractStream.TransportState {
     /** Whether listener.closed() has been called. */
     private final StatsTraceContext statsTraceCtx;
+    // TODO(zpencer): remove @Nullable when okhttp supports transport tracing
+    @Nullable private final TransportTracer transportTracer;
     private boolean listenerClosed;
     private ClientStreamListener listener;
     private boolean fullStreamDecompression;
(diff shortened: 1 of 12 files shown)


core/src/main/java/io/grpc/internal/AbstractClientStream.java
```diff
@@ -193,6 +196,8 @@ public final boolean isReady() {
   protected abstract static class TransportState extends AbstractStream.TransportState {
     /** Whether listener.closed() has been called. */
     private final StatsTraceContext statsTraceCtx;
+    // TODO(zpencer): remove @Nullable when okhttp supports transport tracing
+    @Nullable private final TransportTracer transportTracer;
```
ejona86: Provide a getter for this, to prevent needing the field in subclasses?

netty/src/main/java/io/grpc/netty/NettyClientTransport.java
```diff
@@ -295,9 +309,20 @@ public Attributes getAttributes() {
 
   @Override
   public Future<TransportTracer.Stats> getTransportStats() {
-    SettableFuture<TransportTracer.Stats> ret = SettableFuture.create();
-    ret.set(null);
-    return ret;
+    if (channel.eventLoop().inEventLoop()) {
+      // This is necessary, otherwise we will block forever if we get the future from inside
```
ejona86: This comment seems incorrect since the swap to Future.

zpencer: The comment is referring to how this workaround lets us do Future.get() on the return value while on the event loop. I suppose we can attach a listener to the Future to work around this but it may not be very obvious why the Future blocks forever.

core/src/main/java/io/grpc/internal/AbstractClientStream.java
zpencer: The superclass (`AbstractClientStream.TransportState`) does not need it, it's only passing it down to the Deframer. Would we want to punch a hole all the way into the deframer to get this value?

ejona86: I was thinking just having the getter here, for use by Netty+OkHttp.

core/src/main/java/io/grpc/internal/AbstractClientStream.java
```diff
@@ -104,6 +104,9 @@ protected AbstractClientStream(
       Metadata headers,
       boolean useGet) {
     Preconditions.checkNotNull(headers, "headers");
+    if (transportTracer != null) {
+      transportTracer.reportStreamStarted();
```
ejona86: Should this be done in the transportstate instead?

core/src/main/java/io/grpc/internal/AbstractClientStream.java
zpencer: Ah understood, moving it.

core/src/main/java/io/grpc/internal/AbstractClientStream.java
zpencer: `onStreamAllocated` looks like a good place to move this

ejona86:
Would it be better to plumb in a Ticker to inject the timestamps?

zpencer:
I'll bring in the fake clock on separate PR, since this one is getting a bit large.

PR #3711, opened 2017-11-10
core: backport pull #3677 to v1.8.x

zpencer:
https://github.com/grpc/grpc-java/pull/3677

### DIFF
--- a/core/src/main/java/io/grpc/internal/GrpcUtil.java
+++ b/core/src/main/java/io/grpc/internal/GrpcUtil.java
@@ -44,6 +44,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.HttpURLConnection;
 import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.net.URI;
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -226,6 +227,33 @@
...
 
   /**
+   * The default proxy detector.
+   */
+  public static final ProxyDetector DEFAULT_PROXY_DETECTOR = new ProxyDetectorImpl();
+
+  /**
+   * A proxy detector that always claims no proxy is needed.
+   */
+  public static final ProxyDetector NOOP_PROXY_DETECTOR = new ProxyDetector() {
```
carl-mastrangelo: optional:

this only appears to be used in tests.  You might consider pulling this out into its own static class, putting the instance inside of it, and referencing it from tests like that.  

Then the regular code won't have to initialize this anonymous class in prod code.

zpencer: This is no longer true because GAE+jdk7 will use this noop detector.

PR #3715, opened 2017-11-10
netty: propagate exception caught in BufferingHandler

carl-mastrangelo:


carl-mastrangelo:
The thing is that if there are no buffers in its cumulation, the error message goes nowhere and the trace is lost.  I can override this method in the sub classes.  The thing is that this class doesnt always know if its the last one in the pipeline.

ejona86:
> The thing is that if there are no buffers in its cumulation, the error message goes nowhere and the trace is lost.

Do you have a case where this happens?

> The thing is that this class doesnt always know if its the last one in the pipeline.

It _must_ be the last in the pipeline. Subtle things break if it's not. 99% of the reason it exists is to be the last thing in the pipeline.

Even if sometimes it isn't the last in the pipeline, in the common case it is, so we _can't_ propagate the exception here without other changes. We'd more pieces to come into play for a solution.

PR #3724, opened 2017-11-13
core,netty: plumb fake clock into transport tracer tests

zpencer:
The fake clock lets us advance the clock deterministically and avoid
`Thread.sleep` in unit tests.

### DIFF
--- a/netty/src/test/java/io/grpc/netty/NettyTransportTest.java
+++ b/netty/src/test/java/io/grpc/netty/NettyTransportTest.java
@@ -18,8 +18,10 @@ package io.grpc.netty;
 
 import io.grpc.ServerStreamTracer;
 import io.grpc.internal.ClientTransportFactory;
+import io.grpc.internal.FakeClock;
 import io.grpc.internal.InternalServer;
 import io.grpc.internal.ManagedClientTransport;
+import io.grpc.internal.TransportTracer;
 import io.grpc.internal.testing.AbstractTransportTest;
 import java.net.InetSocketAddress;
 import java.util.List;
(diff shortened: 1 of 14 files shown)


testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
```diff
@@ -190,6 +190,21 @@ public void tearDown() throws InterruptedException {
     }
   }
 
+  /**
+   * Moves the clock foward, for tests that require moving the clock fowrard. It is the transport
```
ejona86: s/fowrard/forward/

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
```diff
@@ -190,6 +190,21 @@ public void tearDown() throws InterruptedException {
...
+   * subclass's responsibility to implement this method.
+   */
+  protected void advanceClock(long nanos) {
```
ejona86: Include Nanos in the name, or require a TimeUnit. It was unexpected it uses different units than `currentTimeMillis()` below.

netty/src/test/java/io/grpc/netty/NettyTransportTest.java
```diff
@@ -32,15 +34,28 @@
...
-      .flowControlWindow(65 * 1024)
-      .negotiationType(NegotiationType.PLAINTEXT)
-      .buildTransportFactory();
+  private final ClientTransportFactory clientFactory;
+
+  /** Creates a test object. */
+  public NettyTransportTest() {
+    // Although specified here, address is ignored because we never call build.
+    NettyChannelBuilder builder = NettyChannelBuilder.forAddress("localhost", 0);
+    InternalNettyChannelBuilder.setTransportTracerFactorty(builder, fakeClockTransportTracer);
```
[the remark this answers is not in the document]
zpencer: I think I copied this pattern from a place that was doing it for other reasons. Reverted

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
zpencer: fixed

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
zpencer: changed to accept a timeunit

PR #3726, opened 2017-11-13
netty: remove unneeded TransportTracer null checks

zpencer:
There is already a null check at initialization, and the field is final:
`this.transportTracer = checkNotNull(transportTracer, "transportTracer");`


netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -351,22 +351,20 @@ public void run() {
...
 
-    if (transportTracer != null) {
-      assert encoder().connection().equals(decoder().connection());
-      final Http2Connection connection = encoder().connection();
-      transportTracer.setFlowControlWindowReader(new TransportTracer.FlowControlReader() {
-        private final Http2FlowController local = connection.local().flowController();
-        private final Http2FlowController remote = connection.remote().flowController();
+    assert encoder().connection().equals(decoder().connection());
+    final Http2Connection connection = encoder().connection();
+    transportTracer.setFlowControlWindowReader(new TransportTracer.FlowControlReader() {
```
carl-mastrangelo: it is recommended that you put this in a named class local to this method.  It makes fort better stack trace messages, rather than $3 or other hard to read names.  It usually only costs 1 line too.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -351,22 +351,20 @@ public void run() {
...
 
 
-    if (transportTracer != null) {
-      assert encoder().connection().equals(decoder().connection());
-      final Http2Connection connection = encoder().connection();
-      transportTracer.setFlowControlWindowReader(new TransportTracer.FlowControlReader() {
-        private final Http2FlowController local = connection.local().flowController();
-        private final Http2FlowController remote = connection.remote().flowController();
+    assert encoder().connection().equals(decoder().connection());
+    final Http2Connection connection = encoder().connection();
```
carl-mastrangelo: this could be non final if you extracted local and remove up a scope level.

zpencer: done

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
zpencer: moved to a class in `Util` so that it can be reused in `NettyClientHandler`

ejona86 (approved):

PR #3728, opened 2017-11-13
okhttp: injectable SerializingExecutor for unit tests

zpencer:
This is needed for wiring TransportTracer to okhttp:
For the pings sent stat, the AsyncFrameWriter performs the
ping on the SerializingExecutor. We need to block until the ping
runnable actually ran before checking the stats counters.


ejona86:
I'd like to discuss in-person. I'm a bit confused about what is happening where and what guarantees we need.

zpencer:
No longer needed, the stat will be bumped when a keepalive is attempted.

PR #3729, opened 2017-11-13
okhttp,netty: move StreamIdHolder to internal

carl-mastrangelo:
This makes streamId holder common between okhttp and netty.  In a
subsequent PR, I am going to make it used by ClientCallImpl callbacks
as a debugging identifier in toString()

The reason is that sometimes a callback throws an exception and is
logged, but the client call is still alive for some reason.  It's nice
to be able to associate an exception from a callback with a particular
RPC.

One other note: this makes the `id` in Okhttp non volatile.  I looked
at all the accesses, and think this was a hold over from before
the transport state change.


ejona86:
> It's nice to be able to associate an exception from a callback with a particular RPC.

How would that look? 1) Where is the identifier exposed when the exception occurs, and 2) how would you determine which RPC that identifier corresponded with?

PR #3731, opened 2017-11-14
gae-interop-testing: kokoro script to run interop tests in GAE

zpencer:
GAE interop tests:
- add runtime configs to jdk7 and jdk8 projects 

kokoro scripts: 
- make sure a 'dummy-default' version always exists
- when the tests are done, divert all traffic to dummy-default and delete the version deployed for tests

### DIFF
--- a/buildscripts/kokoro/gae-interop.sh
+++ b/buildscripts/kokoro/gae-interop.sh
@@ -35,11 +35,7 @@ echo "<?xml version='1.0' encoding='utf-8'?>
 cat ./gae-interop-testing/gae-jdk8/src/main/webapp/WEB-INF/appengine-web.xml
 # Deploy the dummy 'default' version. It doesn't matter if we race with other kokoro runs.
 # We only require that it exists when cleanup() is called.
-set +e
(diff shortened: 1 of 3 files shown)


zpencer:
@jtattermusch @matt-kwong the script is stable now, PTAL

buildscripts/kokoro/gae-interop.sh
```diff
@@ -1,3 +1,75 @@
-#!/bin/sh
+#!/bin/bash
 
-# An empty placeholder shellscript
+set -exu -o pipefail
+
+pwd
+hostname
```
matt-kwong: It looks like these are leftover commands from exploring Kokoro. These should be removed.

zpencer: Fixed

zpencer:
The test servlet running in GAE aggregates all the failed test case names and exceptions and puts it in the HTTP response body. The gradle script prints the `result?.body()?.string()` so we print that in the test log. It still has the problem of external users not being able to read them by default, but it does not require that people have access to the GAE logs.

zpencer:
Hi @jtattermusch , can I get an approval on this if there are no issues?

Issue #3370, opened 2017-08-21
Netty server poorly handles unknown content type


ejona86:
If the content-type is missing or unknown, the [Netty transport sends RST_STREAM with REFUSED_STREAM](https://github.com/grpc/grpc-java/blob/v1.5.0/netty/src/main/java/io/grpc/netty/NettyServerHandler.java#L586). That's bonkers on multiple levels.

We should probably respond with HTTP 415 instead.

lbensaad:
I was about to open an issue about this. I want a grpc-java server to serve a grpc-web client. 
When the server receives a connection, it gives an exception:
`SEVERE: Transport failed
io.netty.handler.codec.http2.Http2Exception: Unexpected HTTP/1.x request`

I want to write a wrapper the same way as in [grpc-web wrapper.go](https://github.com/improbable-eng/grpc-web/tree/master/go/grpcweb)

So I think the only way to add it, is to expose the netty pipeline and add a codec that transforms requests and responses between grpc and grpc-web.

If there is another way to do it please let me know.

ejona86:
@lbensaad, your problem seems unrelated. Note that the older HTTP/2 API in Netty that gRPC is using doesn't let you "add a codec that transforms": it's monolithic and avoids the pipeline. gRPC needs to migrate to "h2childchan": https://github.com/netty/netty/issues/3667

lbensaad:
Thanks @ejona86. 
Should I open another issue on how to due this wrapper?

ejona86:
@lbensaad, I feel like we had an issue for that already, but I can't find it. Yeah, open up a new issue.

ramaraochavali:
@ejona86 are you actively working on this?

ejona86:
On a recent plane flight I messed with it some in https://github.com/ejona86/grpc-java/tree/dont-misuse-refused . But I've not had time to circle back to it. I'd be fine if you wanted to pick it up (as it stands, I'll have to re-discover what my TODOs and FIXMEs mean :smile:).

ramaraochavali:
@ejona86 I looked at it and could not find what you meant by couple of TODOs though. Otherwise the implementation looks fine.

ejona86:
FIXME: https://github.com/ejona86/grpc-java/commit/ad930414292374e3215cbcc1d7c4d016a1b5cdc3#diff-dde99a3f12e631de9f1605b24abd9c2bR371
[...]

PR #3735, opened 2017-11-14
netty: Netty server poorly handles unknown content type

ramaraochavali:
PR for https://github.com/grpc/grpc-java/issues/3370

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
@@ -668,19 +668,6 @@ class NettyServerHandler extends AbstractNettyHandler {
     });
   }
 
-  private void verifyContentType(int streamId, Http2Headers headers) throws Http2Exception {
-    CharSequence contentType = headers.get(CONTENT_TYPE_HEADER);
-    if (contentType == null) {
-      throw Http2Exception.streamError(streamId, Http2Error.REFUSED_STREAM,
-          "Content-Type is missing from the request");
-    }
-    String contentTypeString = contentType.toString();
-    if (!GrpcUtil.isGrpcContentType(contentTypeString)) {
-      throw Http2Exception.streamError(streamId, Http2Error.REFUSED_STREAM,
-          "Content-Type '%s' is not supported", contentTypeString);
-    }
-  }
-
   private void respondWithHttpError(
       ChannelHandlerContext ctx, int streamId, int code, Status.Code statusCode, String msg) {
     Metadata metadata = new Metadata();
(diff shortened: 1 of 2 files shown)


netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -375,15 +378,40 @@ private void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
...
+
+      if (!HTTP_METHOD.equals(headers.method())) {
+        respondWithHttpError(ctx, streamId, 405, Status.Code.INTERNAL,
+            String.format("Method '%s' is not supported", headers.method()));
+        return;
+      }
+      // Remove the leading slash of the path and get the fully qualified method name
+      CharSequence path = headers.path();
+      if (path.charAt(0) != '/') {
+        respondWithHttpError(ctx, streamId, 404, Status.Code.UNIMPLEMENTED,
```
ejona86: The 404 case should be first. If both content-type and path are wrong, we should respond with 404.

ramaraochavali:
moved the path check up to return 404 early. Added explicit `path==null` to give a better message otherwise we could have combined both.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -375,15 +378,49 @@ private void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
...
+            String.format("Expected path to start with /: %s", path));
+        return;
+      }
+
+      String method = path.subSequence(1, path.length()).toString();
+
       // Verify that the Content-Type is correct in the request.
-      verifyContentType(streamId, headers);
```
ejona86: Oops. I didn't delete these methods. They're now unused: `determineMethod`, `verifyContentType`

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -375,15 +378,49 @@ private void onHeadersRead(ChannelHandlerContext ctx, int streamId, Http2Headers
...
+        return;
+      }
+
+      if (path.charAt(0) != '/') {
+        respondWithHttpError(ctx, streamId, 404, Status.Code.UNIMPLEMENTED,
+            String.format("Expected path to start with /: %s", path));
+        return;
+      }
+
+      String method = path.subSequence(1, path.length()).toString();
```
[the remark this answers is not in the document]
ejona86: `path` is a CharSequence, so it doesn't have `substring`.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
ejona86: I went ahead and deleted them, since it was trivial.

ejona86:
@ramaraochavali, thank you!

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
ramaraochavali: @ejona86  great..thanks..I should have done it my self. Did not pay attention. Sorry :-)

PR #3748, opened 2017-11-15
netty: add internal test accessor to builder

dapengzhang0:
Some forwarding channel builder that wraps a `NettyChannelBuilder` configs a custom protocol negotiator. It is very difficult to unit test that the config is really set up after `builder.build()`, because currently it is very hard to access the protocol negotiator from the builder, or the channel being built.

An internal test accessor is added to address the above difficulty.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/InternalNettyChannelBuilder.java
+++ b/netty/src/main/java/io/grpc/netty/InternalNettyChannelBuilder.java
@@ -16,9 +16,11 @@
 
 package io.grpc.netty;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.grpc.Internal;
 import io.grpc.internal.ProxyParameters;
 import java.net.SocketAddress;
+import javax.annotation.Nullable;
 
 /**
  * Internal {@link NettyChannelBuilder} accessor.  This is intended for usage internal to the gRPC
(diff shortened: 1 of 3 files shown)


netty/src/main/java/io/grpc/netty/InternalNettyChannelBuilder.java
```diff
@@ -75,4 +77,25 @@ public static void setStatsRecordStartedRpcs(NettyChannelBuilder builder, boolea
...
+
+    /**
+     * Gets the {@link TransportCreationParamsFilterFactory} that is dynamically set to the
+     * specified NettyChannelBuilder at the moment.
+     */
+    @Nullable
+    public static TransportCreationParamsFilterFactory getDynamicTransportParamsFactory(
+        NettyChannelBuilder builder) {
+      if (builder.dynamicParamsFactory instanceof TransportCreationParamsFilterFactory) {
+        return (TransportCreationParamsFilterFactory) builder.dynamicParamsFactory;
```
carl-mastrangelo: I would have to be convinced of the usefulness.  It only works if the tcpff was also set with this class, since it can only return the type from this class.  

Can you show some example usage of this class?

dapengzhang0: User can only set tcpff by `InternalNettyChannelBuilder` class. I cc'ed you a CL the builder in which is hard to test without an accessor.

dapengzhang0: Any insight making test easier?

PR #3752, opened 2017-11-15
Add ChannelStats and tests

zpencer:
The call begin stat is bumped when the application calls `.start` and call end stats are bumped when the  `ClientStreamListener.closed` callback is called. This means the begin stat on the channel will be bumped immediately but the transport level stats may come later because of  `DelayedClientTransport`. The call end stats may be bumped on the transport first, and will finally be bumped on the channel when the callback gets triggered.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ClientCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ClientCallImpl.java
@@ -68,7 +68,7 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
 
   private final MethodDescriptor<ReqT, RespT> method;
   private final Executor callExecutor;
-  private final ChannelTraceStats channelStats;
+  private final ChannelStats channelStats;
   private final Context context;
   private volatile ScheduledFuture<?> deadlineCancellationFuture;
   private final boolean unaryRequest;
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -267,6 +270,7 @@ public void runInContext() {
       // was cancelled.
       removeContextListenerAndCancelDeadlineFuture();
     }
+    channelStats.reportCallStarted();
```
carl-mastrangelo: this should be up higher, before the listener can be invoked.   Otherwise the number of calls failed might be higher than the number of calls started; a bug.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -523,6 +527,7 @@ private void close(Status status, Metadata trailers) {
       cancelListenersShouldBeRemoved = true;
       try {
         closeObserver(observer, status, trailers);
+        channelStats.reportCallEnded(status.isOk());
```
carl-mastrangelo: Probably worth nesting the try finally.  Exceptions in the users interceptors can cause an exception to be thrown in `closeObserver`

core/src/main/java/io/grpc/internal/OobChannel.java
```diff
@@ -54,6 +54,7 @@
   private SubchannelPicker subchannelPicker;
 
   private final LogId logId = LogId.allocate(getClass().getName());
+  private final ChannelStats channelStats = new ChannelStats();
```
carl-mastrangelo: nit: This is effectively untestable since you are using the system clock.  End to end tests will be harder to implement.

zpencer:
@carl-mastrangelo please review updated changes

zpencer:
FYI ChannelStats be turned into an interface and will need to be moved out of `io.grpc.internal` so that `io.grpc.ManagedChannel` can refer to it.

carl-mastrangelo:
Please make sure the commit message is in the right format

PR #3753, opened 2017-11-15
core: workaround for Atomic*FieldUpdater bug on some Android devices

ericgribkoff:
"This is a large but kinda sorta mechanical change..."

This adds a synchronized fallback update mechanism when use of Atomic*FieldUpdater fails. This ignores the use of atomic field updaters in grpclb and round robin load balancer. The bug in the JDK reflection API  seems to only be an issue on some Samsung Android 5.0.x devices, but there are enough of these out there that a workaround is necessary.

**Question:** The try-catch blocks in this PR catch any throwable from initialization of `FieldUpdaterAtomicHelper` objects, logs the throwable, and falls back to the synchronized implementation. Should we only catch `NoSuchFieldException`? This is the only known exception caused by the bug, so it could/should be ok to specialize the catch blocks. Guava did not do this, however. See prior art: https://github.com/google/guava/blob/fc992745150d95544fec68b9c6e89b5c3bc44b15/guava/src/com/google/common/util/concurrent/AbstractFuture.java#L147

### DIFF
--- a/core/src/main/java/io/grpc/internal/CensusStatsModule.java
+++ b/core/src/main/java/io/grpc/internal/CensusStatsModule.java
@@ -153,19 +153,28 @@ public final class CensusStatsModule {
   }
 
   private static final class ClientTracer extends ClientStreamTracer {
+    // When using Atomic*FieldUpdater, some Samsung Android 5.0.x devices encounter a bug in the JDK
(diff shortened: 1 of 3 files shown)


ericgribkoff:
Replaced by https://github.com/grpc/grpc-java/pull/3754

PR #3754, opened 2017-11-16
core: workaround for Atomic*FieldUpdater bug on some Android devices

ericgribkoff:
Replaces https://github.com/grpc/grpc-java/pull/3753. Leaves the synchronized fallback in place for SerializingExecutor, but just falls back to (racy) volatile updates in the stats modules.

### DIFF
--- a/core/src/main/java/io/grpc/internal/CensusStatsModule.java
+++ b/core/src/main/java/io/grpc/internal/CensusStatsModule.java
@@ -192,7 +192,7 @@ public final class CensusStatsModule {
         tmpInboundUncompressedSizeUpdater =
             AtomicLongFieldUpdater.newUpdater(ClientTracer.class, "inboundUncompressedSize");
       } catch (Throwable t) {
-        logger.log(Level.WARNING, "Creating atomic field updaters failed", t);
+        logger.log(Level.SEVERE, "Creating atomic field updaters failed", t);
         tmpOutboundMessageCountUpdater = null;
         tmpInboundMessageCountUpdater = null;
         tmpOutboundWireSizeUpdater = null;
@@ -301,7 +301,7 @@ public final class CensusStatsModule {
         tmpCallEndedUpdater =
             AtomicIntegerFieldUpdater.newUpdater(ClientCallTracer.class, "callEnded");
       } catch (Throwable t) {
-        logger.log(Level.WARNING, "Creating atomic field updaters failed", t);
+        logger.log(Level.SEVERE, "Creating atomic field updaters failed", t);
         tmpStreamTracerUpdater = null;
         tmpCallEndedUpdater = null;
       }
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/grpc/internal/CensusStatsModule.java
```diff
@@ -154,18 +155,58 @@ ClientInterceptor getClientInterceptor(boolean recordStartedRpcs, boolean record
...
+            AtomicLongFieldUpdater.newUpdater(ClientTracer.class, "outboundWireSize");
+        tmpInboundWireSizeUpdater =
+            AtomicLongFieldUpdater.newUpdater(ClientTracer.class, "inboundWireSize");
+        tmpOutboundUncompressedSizeUpdater =
+            AtomicLongFieldUpdater.newUpdater(ClientTracer.class, "outboundUncompressedSize");
+        tmpInboundUncompressedSizeUpdater =
+            AtomicLongFieldUpdater.newUpdater(ClientTracer.class, "inboundUncompressedSize");
+      } catch (Throwable t) {
+        logger.log(Level.WARNING, "Creating atomic field updaters failed", t);
+        tmpOutboundMessageCountUpdater = null;
```
carl-mastrangelo: initialize all these to null above?

core/src/main/java/io/grpc/internal/CensusStatsModule.java
```diff
@@ -154,18 +155,58 @@ ClientInterceptor getClientInterceptor(boolean recordStartedRpcs, boolean record
...
+        tmpOutboundWireSizeUpdater =
+            AtomicLongFieldUpdater.newUpdater(ClientTracer.class, "outboundWireSize");
+        tmpInboundWireSizeUpdater =
+            AtomicLongFieldUpdater.newUpdater(ClientTracer.class, "inboundWireSize");
+        tmpOutboundUncompressedSizeUpdater =
+            AtomicLongFieldUpdater.newUpdater(ClientTracer.class, "outboundUncompressedSize");
+        tmpInboundUncompressedSizeUpdater =
+            AtomicLongFieldUpdater.newUpdater(ClientTracer.class, "inboundUncompressedSize");
+      } catch (Throwable t) {
+        logger.log(Level.WARNING, "Creating atomic field updaters failed", t);
```
carl-mastrangelo: This is probably SEVERE worthy.  A core part of the Java API doesn't work.

ericgribkoff: Done

core/src/main/java/io/grpc/internal/CensusStatsModule.java
ericgribkoff: Leaving as-is so that all field updater references are assigned a value or they are all null

PR #3757, opened 2017-11-16
auth: Treat IOExceptions as UNAVAILABLE

### LINKED ISSUE #3267: UNAUTHORIZED if Auth Server is down?
We are using `GoogleAuthLibraryCredentials` to authenticate our gRPC calls and we sometimes have the case where the authentication fails, because the auth server is not reachable and then the call fails with `UNAUTHENTICATED` status code.

Some people on my team argue that it should fail with `UNAVAILABLE`, I personally think `UNAUTHENTICATED` makes sense, so that a user can distinquish between a call and auth having gone wrong. However, arguably [1] is not clear about this.

The discussion arose, because of retries. In Bazel, we want to retry certain status codes i.e. `UNAUTHENTICATED` if the server could not be reached, but not if it failed due to wrong credentials. Any thoughts?

cc: @ejona86 @zhangkun83 

[1] https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

ejona86:
Fixes #3267 

Should we backport this to 1.8?

CC @buchgr

### DIFF
--- a/auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
+++ b/auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
@@ -102,7 +102,9 @@ final class GoogleAuthLibraryCallCredentials implements CallCredentials {
               metadata = creds.getRequestMetadata(uri);
             } catch (IOException e) {
               // Since it's an I/O failure, let the call be retried with UNAVAILABLE.
-              applier.fail(Status.UNAVAILABLE.withCause(e));
+              applier.fail(Status.UNAVAILABLE
+                  .withDescription("Credentials failed to obtain metadata")
+                  .withCause(e));
               return;
             }
             // Re-use the headers if getRequestMetadata() returns the same map. It may return a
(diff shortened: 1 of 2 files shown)


auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
```diff
@@ -96,7 +97,14 @@ public void run() {
             // https://github.com/google/google-auth-library-java/issues/3 is resolved.
             //
             // Some implementations may return null here.
-            Map<String, List<String>> metadata = creds.getRequestMetadata(uri);
+            Map<String, List<String>> metadata;
+            try {
+              metadata = creds.getRequestMetadata(uri);
+            } catch (IOException e) {
+              // Since it's an I/O failure, let the call be retried with UNAVAILABLE.
+              applier.fail(Status.UNAVAILABLE.withCause(e));
```
carl-mastrangelo: Please include a message

ejona86: Ah, good point. Done.

PR #3758, opened 2017-11-16
auth: Use async version of getRequestMetadata

ejona86:
This avoids using DelayedStream and a thread hop when the credentials
are known immediately.

This is dependent on #3757.


auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
```diff
@@ -83,37 +85,42 @@ public void applyRequestMetadata(MethodDescriptor<?, ?> method, Attributes attrs
...
+        // Re-use the headers if getRequestMetadata() returns the same map. It may return a
+        // different map based on the provided URI, i.e., for JWT. However, today it does not
+        // cache JWT and so we won't bother tring to save its return value based on the URI.
+        Metadata headers;
+        synchronized (GoogleAuthLibraryCallCredentials.this) {
+          if (lastMetadata == null || lastMetadata != metadata) {
+            try {
+              lastHeaders = toHeaders(metadata);
+            } catch (Throwable t) {
+              applier.fail(Status.UNAUTHENTICATED.withCause(t));
```
zhangkun83: If you put try-catch outside of synchronized, you should be able to call applier.fail() without holding the lock.

core/src/main/java/io/grpc/internal/CallCredentialsApplyingTransportFactory.java
```diff
@@ -86,8 +87,12 @@ public ClientStream newStream(
         if (callOptions.getAuthority() != null) {
           effectiveAttrsBuilder.set(CallCredentials.ATTR_AUTHORITY, callOptions.getAuthority());
         }
-        creds.applyRequestMetadata(method, effectiveAttrsBuilder.build(),
-            firstNonNull(callOptions.getExecutor(), appExecutor), applier);
+        try {
+          creds.applyRequestMetadata(method, effectiveAttrsBuilder.build(),
+              firstNonNull(callOptions.getExecutor(), appExecutor), applier);
```
zhangkun83: applyRequestMetadata() is not supposed to throw. Errors should be passed to the applier.

ejona86: Since this is user-provided code, I don't feel like that negates the need to handle exceptions from the call. It's too easy for a NPE or similar to leak through. I've adding a message to the status to make it clearer the code did something wrong.

auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
ejona86: Done.

ejona86:
I'm going to need to rebase on top of #3757. When I do that I'll also update the Statuses to make sure they descriptions.

Issue #1767, opened 2016-05-03
Tracking Issue for authority being Experimental.


carl-mastrangelo:
In regards to #67.

Specific usages:

* ``

ejona86:
API review:
* Can't stabilize the CallOptions until #67 is fixed.
* Stabilize ManagedChannelBuilder.overrideAuthority, but should also tweak its documentation to say it is "okay" to use outside of tests, but intended for tests. It doesn't break any security assumptions.

larry-safran:
API review:  Needs to move forward or backward - tech debt requiring significant work

PR #3761, opened 2017-11-16
core: Stabilize ManagedChannelBuilder.overrideAuthority

ejona86:
The CallOptions from #1767 are not included, so this is not enough to resolve
the issue.


core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -140,7 +140,6 @@
    * @return this
```
dapengzhang0: `* <p>Should only used by tests.` need be revised? The grammar is also incorrect.

ejona86:
PTAL

PR #3764, opened 2017-11-16
core: Disable handshakeTimeout for InProcess

ejona86:
handshakeTimeout is unnecessary for InProcess, and the scheduling is causing
Thread creation that is breaking restrictive test environments. Those
environments are mostly broken already because client-side will try to create
Threads as well, but they are currently lucking out that the exception on
client-side doesn't break much.


core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
```diff
@@ -83,6 +84,9 @@ private InProcessServerBuilder(String name) {
     // https://github.com/grpc/grpc-java/issues/2284
     setStatsRecordStartedRpcs(false);
     setStatsRecordFinishedRpcs(false);
+    // Disable handshake timeout because it is unnecessary, and can trigger Thread creation that can
+    // break some environemnts (like tests).
```
carl-mastrangelo: s/environemnts/environments/

PR #3765, opened 2017-11-17
Bumping proto-google-common-protos to 1.0.0

garrettjonesgoogle:
We recently released proto-google-common-protos to 1.0.0.


thelinuxfoundation:

* User @garrettjonesgoogle isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,

garrettjonesgoogle:
Reverify me, linuxfoundation!

ejona86:
@garrettjonesgoogle, would this be good to backport to the 1.8 release next week?

garrettjonesgoogle:
It's really just about guarantee of surface stability - there might be some extra protos in there but I'm not sure. I think it really depends if you want to "hand over" this surface stability to users of 1.8; there is very little risk to backporting it.

ejona86:
Okay. We didn't stabilize the API that uses it in 1.8, and we'd still need to figure out Java 6 vs Java 7 bytecode before we could. So I won't bother backporting, since it's not really a bug and there's no need.

PR #3769, opened 2017-11-17
netty: Include more stace trace information for ClosedChannelExceptions

ejona86:
There's been another sighting, but it's very hard to narrow down what is
triggering it.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/Utils.java
+++ b/netty/src/main/java/io/grpc/netty/Utils.java
@@ -154,7 +154,13 @@ class Utils {
       // ClosedChannelException is used any time the Netty channel is closed. Proper error
       // processing requires remembering the error that occurred before this one and using it
       // instead.
-      return Status.UNKNOWN.withCause(t);
+      //
+      // Netty uses an exception that has no stack trace, while we would never hope to show this to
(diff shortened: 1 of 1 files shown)


netty/src/main/java/io/grpc/netty/Utils.java
```diff
@@ -154,7 +154,13 @@ public static Status statusFromThrowable(Throwable t) {
...
       // processing requires remembering the error that occurred before this one and using it
       // instead.
-      return Status.UNKNOWN.withCause(t);
+      //
+      // Netty uses an exception that has no stack trace, while we would never hope to show this to
+      // users, if it happens having the extra information may provide a small hint of where to
+      // look.
+      ClosedChannelException extraT = new ClosedChannelException();
+      extraT.initCause(t);
+      return Status.UNKNOWN.withCause(extraT);
```
carl-mastrangelo: Optional: include a message in either the exception or the Status.

ejona86: Yeah, I was looking at that. We don't have much information here. We should probably add a String argument to the method for the description. But anything along those lines isn't appropriate in this PR.

PR #3782, opened 2017-11-20
V1.8.x

carl-mastrangelo:
Two commits as usual


ejona86:
Shouldn't there be another commit updating README?

carl-mastrangelo:
@ejona86 right you, ptal

PR #3779, opened 2017-11-20
testing: replace deprecated truth.FailureStrategy with truth.FailureMetadata.

JiangJi:
Truth.FailureStrategy & SubjectFactory is being deprecated and will be removed in next release. 
This changed reference of those classes to use 'FailureMetadata' and 'Subject.Factory' respectively. 
New api is added in Truth-0.36, dependency is in build.gradle is also updated to use this
version of Truth.

PR #3786, opened 2017-11-22
testing: Remove DeadlineSubject

ejona86:
The class is still used internally, so we move it to context's tests for
it to be reused. To avoid a circular dependency with context's tests
depending on core's tests, StaticTestingClassLoader was also moved to
context's tests.

This is driven by a need to modernize DeadlineSubject for newer versions
of Truth, but the newer versions of Truth update Guava. To avoid leaking
the Guava update to all users of grpc-testing, we're removing the
Subject. In our internal tests we can update the Truth dependency with
less issue.

CC @JiangJi 

Some discussion in #3779

### DIFF
--- a/context/build.gradle
+++ b/context/build.gradle
@@ -1,7 +1,6 @@
 description = 'gRPC: Context'
 
 dependencies {
-    testCompile project(':grpc-testing')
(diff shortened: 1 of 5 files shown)


ejona86:
@carl-mastrangelo, friendly ping

PR #3789, opened 2017-11-22
core,okhttp: handle unresolved proxy addresses

ericgribkoff:
This fixes a NullPointerException in OkHttpClientTransport when the proxy returns an unresolved address, in which case `InetSocketAddress#getAddress` will return null. This also changes the `GRPC_PROXY_EXP` handling to explicitly return an unresolved `InetSocketAddress` to avoid I/O at channel creation. Our other uses of the `InetSocketAddress(String, int)` are all in tests or interop/benchmark clients, and as far as I can tell not problematic.

@ejona86 Should this fix (modified to the pre-`ProxySelector` code) be backported? If so, to which releases?


ejona86:
Yeah, let's backport to 1.7 and 1.8

PR #3795, opened 2017-11-23
netty,okhttp: handle unresolved proxy addresses (backport to v1.7.x)

ericgribkoff:
Backport of #3789, albeit modified to the pre-ProxyDetector implementation


ejona86:
Did we decide this was unnecessary?

ericgribkoff:
The channel builder changes are unnecessary (and wrong for Netty); the null check in OkHttpClientTransport still looks useful, as the proxy address can be unresolved and the existing code can throw an NPE. Should I revert the channel builder changes and continue with the backport?

ejona86:
That sounds fine, but revert Netty change as a separate commit (otherwise it'll be far to hard to notice the difference from master)

ericgribkoff:
The incorrect ChannelBuilder changes are reverted. PTAL

PR #3803, opened 2017-11-28
test: de-flake Census tag checking in AbstractInteropTest.

### LINKED ISSUE #3777: largeUnary interop test flaky: wrong method name in stats check
```
org.junit.ComparisonFailure: method names match expected:<...testing.TestService/[Unary]Call> but was:<...testing.TestService/[StreamingOutput]Call>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at io.grpc.testing.integration.AbstractInteropTest.checkStartTags(AbstractInteropTest.java:1916)
	at io.grpc.testing.integration.AbstractInteropTest.assertServerStatsTrace(AbstractInteropTest.java:1855)
	at io.grpc.testing.integration.AbstractInteropTest.assertStatsTrace(AbstractInteropTest.java:1779)
	at io.grpc.testing.integration.AbstractInteropTest.largeUnary(AbstractInteropTest.java:373)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.lang.Thread.run(Thread.java:748)
```

Seen here: https://kokoro2.corp.google.com/job/grpc/job/java/job/master/job/windows/job/presubmit/job/windows/386/ on PR https://github.com/grpc/grpc-java/pull/3776.

@zhangkun83 Any thoughts on this? My first thought was to blame my https://github.com/grpc/grpc-java/pull/3754 PR, but it looks like this is on the server side, and the server stream doesn't have to worry about multiple streams per call so it shouldn't be impacted even if there were a mistake in https://github.com/grpc/grpc-java/pull/3754.

zhangkun83:
Tests in AbstractInteropTest share the same server instance.  Test
methods, being on the client-side, have no way to reliably tell when a
server-side end-of-RPC record is registered into Census.  Therefore,
we try polling the server records until the expected record is found,
or there is no more record.

When I made the change to record start-of-RPCs to Census, I updated
this test to also check for such records.  I made an assumption that
the end-of-RPC record of an RPC immediately followed the corresponding
start-of-RPC record, which is not always the case.  Thus the flakiness.

Before the fix, Http2NettyLocalChannelTest fails 18 out of 1000 runs
when run internally.  After the fix, there is no more failure.

Resolves #3777

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -1830,50 +1830,54 @@ public abstract class AbstractInteropTest {
       return;
     }
     AssertionError checkFailure = null;
-    boolean passed = false;
 
     if (metricsExpected()) {
       // Because the server doesn't restart between tests, it may still be processing the requests
       // from the previous tests when a new test starts, thus the test may see metrics from previous
       // tests.  The best we can do here is to exhaust all records and find one that matches the
       // given conditions.
+      boolean startRecordFound = false;
+      boolean endRecordFound = false;
       while (true) {
-        MetricsRecord serverStartRecord;
-        MetricsRecord serverEndRecord;
+        MetricsRecord serverRecord;
         try {
-          // On the server, the stats is finalized in ServerStreamListener.closed(), which can be
-          // run after the client receives the final status.  So we use a timeout.
-          serverStartRecord = serverStatsRecorder.pollRecord(5, TimeUnit.SECONDS);
-          serverEndRecord = serverStatsRecorder.pollRecord(5, TimeUnit.SECONDS);
+          serverRecord = serverStatsRecorder.pollRecord(5, TimeUnit.SECONDS);
         } catch (InterruptedException e) {
           throw new RuntimeException(e);
         }
-        if (serverEndRecord == null) {
+        if (serverRecord == null) {
           break;
         }
-        try {
-          checkStartTags(serverStartRecord, method);
-          checkEndTags(serverEndRecord, method, code);
-          if (requests != null && responses != null) {
-            checkCensus(serverEndRecord, true, requests, responses);
+        if (!startRecordFound) {
+          try {
+            checkStartTags(serverRecord, method);
+            startRecordFound = true;
+          } catch (AssertionError e) {
+            // May be the fallout from a previous test, continue trying
+            checkFailure = e;
           }
-          passed = true;
+        } else if (!endRecordFound) {
+          try {
+            checkEndTags(serverRecord, method, code);
+            endRecordFound = true;
+          } catch (AssertionError e) {
+            checkFailure = e;
+          }
+        }
+        if (startRecordFound && endRecordFound) {
           break;
-        } catch (AssertionError e) {
-          // May be the fallout from a previous test, continue trying
-          checkFailure = e;
         }
       }
-      if (!passed) {
-        if (checkFailure == null) {
-          throw new AssertionError("No record found");
-        }
-        throw checkFailure;
+      if (!(startRecordFound && endRecordFound)) {
+        throw new AssertionError(
+            "Insufficient record. startRecordFound=" + startRecordFound
+            + ", endRecordFound=" + endRecordFound,
+            checkFailure);
       }
     }
 
     // Use the same trick to check ServerStreamTracer records
-    passed = false;
+    boolean passed = false;
     while (true) {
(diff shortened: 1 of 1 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1830,50 +1830,54 @@ private void assertServerStatsTrace(String method, Status.Code code,
...
-          checkEndTags(serverEndRecord, method, code);
-          if (requests != null && responses != null) {
-            checkCensus(serverEndRecord, true, requests, responses);
+        if (!startRecordFound) {
+          try {
+            checkStartTags(serverRecord, method);
+            startRecordFound = true;
+          } catch (AssertionError e) {
+            // May be the fallout from a previous test, continue trying
+            checkFailure = e;
```
carl-mastrangelo: I don't understand this line.  What happens if checkFailure is already not null from a previous iteration?   The original error will be lost.

zhangkun83: I only save the last error.
Because the error may be just because we get a record leaked from a previous test, the error is not always useful. If this test does produced a wrong record, the last error will be useful.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1830,50 +1830,54 @@ private void assertServerStatsTrace(String method, Status.Code code,
...
           break;
         }
-        try {
-          checkStartTags(serverStartRecord, method);
-          checkEndTags(serverEndRecord, method, code);
-          if (requests != null && responses != null) {
-            checkCensus(serverEndRecord, true, requests, responses);
+        if (!startRecordFound) {
+          try {
+            checkStartTags(serverRecord, method);
```
carl-mastrangelo: I don't think this approach is correct.  If its possible old tags could be in here, they shouldn't be validated, but discarded.  This is using exceptions as control flow, and it doesn't *look* right.

zhangkun83: This is a trial-and-error approach. This is not an usual approach, but it's the best I can find for this case.  Being trial-and-error implies involving error handling (in Java, it's the exceptions) in the control flow.

carl-mastrangelo: This doesn't work.  It does the wrong thing.  And it's confusing.  There are a number of right ways to do it, so please don't treat this as the only one.  For example:

* If there could be data from multiple calls, then shutdown the server for each test method.   Stats are per server, right?  

* If that's impossible, then you need to collect all the record tags until there are no more, and then scan through to see if the method you are looking for exists.  `checkStartTags`  implies that `method` MUST be in there.   That is false in this case.

* Tags that WERE found are relevant because if we dont find it, we would like to know what was present.   Right now your code just keeps the exception of one of the tags, rather than all of them.  How are you going to debug that??

* The method name is not enough to find the correct tags.   We call the same method multiple times in the test, so finding the tags for a method may not actually correspond to the correct invocation.

* If you do consume multiple tags, you may accidentally prevent a subsequent test from passing.  If testA gets B's and A's tags, and then testB is not going to find its own.  

Please don't make this already complicated test class worse.

zhangkun83:
Closing in favor of #3816

PR #3812, opened 2017-11-30
core,netty,okhttp: move TransportTracer.Stats to io.grpc

zpencer:
io.grpc can not refer to io.grpc.internal so this needs to be moved
out. The end state will have the channelz class in io.grpc, which
needs to refer to stats like this.

### DIFF
--- a/core/src/main/java/io/grpc/InternalTransportStats.java
+++ b/core/src/main/java/io/grpc/InternalTransportStats.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2017, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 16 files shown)


core/src/main/java/io/grpc/InternalTransportStats.java
```diff
@@ -0,0 +1,65 @@
...
+  public final long keepAlivesSent;
+  public final long lastMessageSentTimeNanos;
+  public final long lastMessageReceivedTimeNanos;
+  public final long localFlowControlWindow;
+  public final long remoteFlowControlWindow;
+
+  /**
+   * Creates an instance.
+   */
+  public InternalTransportStats(
```
ejona86: We will probably end up wanting a builder here. But this is fine for now.

PR #3808, opened 2017-11-29
Correct sha1 for common-protos

drigz:
Fixes #3802.

PR #3814, opened 2017-11-30
bazel: Fix protobuf checksums

ejona86:
The protobuf checksums were simply not updated when changing the
version.


ejona86:
Rebased on #3808, since otherwise the commit message appears weird because it includes things from that other PR, but the diff won't show them. Waiting for CI to complete, then will merge.

PR #3816, opened 2017-11-30
interop-testing: start a new server for each test method.

zhangkun83:
By doing this we can isolate the Census records for each test, and
eliminate the trial-and-error workaround in AbstractInteropTest.

This is a preferred fix for #3777 and supersedes #3803

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -214,7 +214,6 @@ public abstract class AbstractInteropTest {
         .add(TestUtils.recordRequestHeadersInterceptor(requestHeadersCapture))
         .add(recordContextInterceptor(contextCapture))
         .addAll(TestServiceImpl.interceptors())
-        .addAll(getServerInterceptors())
         .build();
 
     builder
@@ -309,10 +308,6 @@ public abstract class AbstractInteropTest {
   @Nullable
   protected AbstractServerImplBuilder<?> getServerBuilder() {
     return null;
-  }
-
-  protected List<? extends ServerInterceptor> getServerInterceptors() {
-    return Collections.emptyList();
   }
 
   protected final CensusStatsModule createClientCensusStatsModule() {
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -1858,7 +1858,7 @@ public abstract class AbstractInteropTest {
       assertNotNull(serverEndRecord);
       checkStartTags(serverStartRecord, method);
       checkEndTags(serverEndRecord, method, code);
-      if (requests != null && responses != null) {
+      if (server != null && requests != null && responses != null) {
         checkCensus(serverEndRecord, true, requests, responses);
       }
     }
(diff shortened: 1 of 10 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -270,15 +281,15 @@ public ClientStreamTracer newClientStreamTracer(CallOptions callOptions, Metadat
    */
   @Before
   public void setUp() {
+    clientStatsRecorder = new FakeStatsRecorder();
```
ejona86: Initialize the stats recorders at their declaration (like before)? It doesn't seem like this needs to be delayed until `@Before`

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -287,10 +298,24 @@ public void tearDown() throws Exception {
...
+  /**
+   * Returns the server builder used to create server for each test run.  Return {@code null} if
+   * it shouldn't start a server in the same process.
+   */
+  @Nullable
+  protected AbstractServerImplBuilder<?> getServerBuilder() {
+    return null;
+  }
+
+  protected List<? extends ServerInterceptor> getServerInterceptors() {
```
ejona86: Could we just let subclasses call `serverBuilder.intercept(...)` before returning in `getServerBuilder()`?

carl-mastrangelo: wouldn't that mean sharing the server between this class and the subclasses?

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1414,7 +1432,7 @@ public void customMetadata() throws Exception {
 
   @Test(timeout = 10000)
   public void censusContextsPropagated() {
-    if (!serverInProcess()) {
+    if (server == null) {
```
carl-mastrangelo: This should be `Assume.assumeTrue(server != null)`.  That way it will be clear from the test results that this was skipped rather than passing.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1975,7 +1955,7 @@ private void checkCensus(MetricsRecord record, boolean server,
     for (MessageLite response : responses) {
       uncompressedResponsesSize += response.getSerializedSize();
     }
-    if (server && serverInProcess()) {
+    if (isServer && server != null) {
```
carl-mastrangelo: can server be null is isServer is true?

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
zhangkun83: Done.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
zhangkun83: Done.

ejona86: @carl-mastrangelo, I don't know what you mean. Kun made the change, so you can point out an issue with it if you see one.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
zhangkun83: Done.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
zhangkun83: Yes. I did a refactor to check `server != null` before calling `checkCensus()`. This should be better.

PR #3817, opened 2017-11-30
core: InternalInstrumented<T> for instrumented classes

zpencer:
This allows the channelz class to keep collections of type `Map<Long, InternalInstrumented<T>>`, where `T` is an `@Internal` class that lives in `io.grpc`.

### DIFF
--- a/core/src/main/java/io/grpc/InternalInstrumented.java
+++ b/core/src/main/java/io/grpc/InternalInstrumented.java
@@ -16,15 +16,9 @@
 
 package io.grpc;
 
-import java.util.concurrent.Future;
-
 /**
  * This is an gRPC internal interface. Do not use this.
- *
- * <p>An interface for types that <b>may</b> support instrumentation. If the actual type does not
- * support instrumentation, then the future will return a {@code null}.
  */
 @Internal
-public interface InternalInstrumented<T> extends InternalWithLogId {
-  Future<T> getStats();
+public interface InternalInstrumented<T> extends Instrumented<T>, InternalWithLogId {
 }
--- a/core/src/main/java/io/grpc/InternalInstrumented.java
+++ b/core/src/main/java/io/grpc/InternalInstrumented.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2017, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc;
+
+import javax.annotation.Nullable;
+
+/**
+ * This is an gRPC internal interface. Do not use this.
+ *
+ * <p>An interface for types that <b>may</b> support instrumentation. If the concrete
+ * type does not support instrumentation, then the stat passed to {@link Consumer}
+ * will be {@code null}. The actual consumption of the stat T does not necessarily happen
+ * on the current thread.
+ */
+@Internal
+public interface InternalInstrumented<T> extends InternalWithLogId {
+  void produceStat(Consumer<T> consumer);
+
+  /**
+   * An interface for consuming an arbitrary stat of type T.
+   */
+  interface Consumer<T> {
+    void consume(@Nullable T stat);
+  }
+}
--- a/core/src/main/java/io/grpc/InternalInstrumented.java
+++ b/core/src/main/java/io/grpc/InternalInstrumented.java
@@ -16,24 +16,15 @@
 
 package io.grpc;
 
-import javax.annotation.Nullable;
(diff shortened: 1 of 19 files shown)


ejona86:
Would this be used for anything other than `InternalInstrumented`? If it is used for anything else it'd be `Map<InternalWithLogId, InternalInstrumented<?>>` which is a bit scary, since you wouldn't know the `?`. And wouldn't we need a `Map<Integer, ...>` anyway since we won't have the `InternalWithLogId` instance to do the lookup?

zpencer:
The description had an error, the key should be a Long. I was trying to be generic in the description but in usage it would be things like `private final Map<Long, InternalInstrumented<InternalTransportStats>> allTransports`, and not be a `?`.

I've also decided to keep the interface as a Future for now, because I think there are some edge cases with channel shutdown that I'd like to discuss in person. I wonder if it's possible to race with removing the transport from the event loop, or run into some other internal netty error, and have the callback never be called.

ejona86:
> I wonder if it's possible to race with removing the transport from the event loop, or run into some other internal netty error, and have the callback never be called.

It seems Future doesn't help. You would still need to complete the future in that case. And if you can complete the future then you can call the callback.

Netty actually does do some ugly goo in some of those cases to make sure to complete its futures, so we should be able to handle this. When we do the write to Netty we just need to make sure we listen to the Future; in most of those cases it will fail with ChannelClosedException.

core/src/main/java/io/grpc/InternalInstrumented.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * This is an gRPC internal interface. Do not use this.
+ *
+ * <p>An interface for types that <b>may</b> support instrumentation. If the actual type does not
+ * support instrumentation, then the future will return a {@code null}.
+ */
+@Internal
+public interface InternalInstrumented<T> extends InternalWithLogId {
+  Future<T> getStats();
```
ejona86: Future scares me, because there's really no way to use it without blocking. This would need to be ListenableFuture.

core/src/main/java/io/grpc/InternalInstrumented.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.concurrent.Future;
+
+/**
+ * This is an gRPC internal interface. Do not use this.
+ *
+ * <p>An interface for types that <b>may</b> support instrumentation. If the actual type does not
+ * support instrumentation, then the future will return a {@code null}.
+ */
+@Internal
+public interface InternalInstrumented<T> extends InternalWithLogId {
```
carl-mastrangelo: FYI, nothing in `^io.grpc$` can reference this class, since it would cause dependency cycles.  Can you give a sample usage of this inside `io.grpc` ?

zpencer: Moved the classes around to achieve proper visibility. This will be a sketch of the final result:

```
"core": package io.grpc
===============
// pkg-private so they are hidden from users

class StatsServerImpl {
  private static final StatsServerImpl INSTANCE = new StatsServerImpl();

  private StatsServerImpl() { /* do not instantiate */ }

  // all methods package-private
  static StatsServerImpl getInstance() { return INSTANCE; }
  void registerServer(Instrumented<ServerStats> server) { /*TODO*/ }
  void registerChannel(Instrumented<ChannelStats> server) { /*TODO*/ }
  void registerTransport(Instrumented<TransportStats> server) { /*TODO*/ }
}

class LogId {
  public long getId() { /* TODO */ }
}

interface WithLogId {
  LogId getId();
}

interface Instrumented<T> {
  T getStat();
}
class ServerStats { /*TODO*/ }
class ChannelStats { /*TODO*/ }
class TransportStats { /*TODO*/ }

"core_internal_accessor": package io.grpc
=====================================

// hole punch so that io.grpc.internal can implement LogId and WithLogId

public class InternalLogId extends LogId { }
public interface InternalWithLogId extends WithLogId {
  @Override
  public InternalLogId getId();
}

// hole punch so that netty and others can return stats

public interface InternalInstrumented<T> extends Instrumented<T>, InternalWithLogId { /*TODO*/ }
public class InternalServerStats extends ServerStats { /*TODO*/ }
public class InternalChannelStats extends ChannelStats { /*TODO*/ }
public class InternalTransportStats extends TransportStats { /*TODO*/ }

"netty": package io.grpc.netty
=====================

class NettyServer implements InternalInstrumented<ServerStats> { /*TODO*/ }
class NettyClientTransport implements InternalInstrumented<InternalTransportStats> { /*TODO*/ }
class NettyServerTransport implements InternalInstrumented<InternalTransportStats> { /*TODO*/ }

"core_internal": package io.grpc.internal
========================
public class ManagedChannelImpl implements InternalInstrumented<InternalChannelStats> {
  public ManagedChannelImpl(/* args */) {
    /* initialize it  */

    // in practice, we should not pass 'this' around before ctor is done
    InternalStatsServer.registerChannel(this);
  }
}
```

zpencer:
@carl-mastrangelo friendly ping to take a look at the updated PR, and the end state: https://github.com/grpc/grpc-java/pull/3817#discussion_r155683847

core/src/main/java/io/grpc/LogId.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+package io.grpc;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * A loggable ID, unique for the duration of the program.
+ */
+class LogId {
```
carl-mastrangelo: final

PR #3818, opened 2017-12-01
core: retry support part 1 followup, drainedSubstreams should not contain closed ones

dapengzhang0:
2 commits:
- the first one is pure refactoring to add Substream class,
- the second is not to add closed substreams to drainedSubstreams.

### DIFF
--- a/core/src/main/java/io/grpc/internal/RetriableStream.java
+++ b/core/src/main/java/io/grpc/internal/RetriableStream.java
@@ -529,8 +529,7 @@ abstract class RetriableStream<ReqT> implements ClientStream {
     State drained(Substream substream) {
       checkState(!passThrough, "Already passThrough");
 
-      Set<Substream> drainedSubstreams = new HashSet<Substream>();
-      drainedSubstreams.addAll(this.drainedSubstreams);
+      Set<Substream> drainedSubstreams = new HashSet<Substream>(this.drainedSubstreams);
 
       if (!substream.closed) {
         drainedSubstreams.add(substream);
@@ -554,8 +553,7 @@ abstract class RetriableStream<ReqT> implements ClientStream {
     State closed(Substream substream) {
       substream.closed = true;
       if (this.drainedSubstreams.contains(substream)) {
-        Set<Substream> drainedSubstreams = new HashSet<Substream>();
-        drainedSubstreams.addAll(this.drainedSubstreams);
+        Set<Substream> drainedSubstreams = new HashSet<Substream>(this.drainedSubstreams);
         drainedSubstreams.remove(substream);
         return new State(buffer, drainedSubstreams, winningSubstream, cancelled, passThrough);
       } else {
--- a/core/src/main/java/io/grpc/internal/RetriableStream.java
+++ b/core/src/main/java/io/grpc/internal/RetriableStream.java
@@ -89,14 +89,14 @@ abstract class RetriableStream<ReqT> implements ClientStream {
   abstract void postCommit();
 
   private void retry() {
-    Substream substream = newSubstream();
+    Substream substream = createSubstream();
 
     // TODO(zdapeng): update "grpc-retry-attempts" header
 
     drain(substream);
   }
 
-  private Substream newSubstream() {
+  private Substream createSubstream() {
     Substream sub = new Substream();
     // NOTICE: This set _must_ be done before stream.start() and it actually is.
     sub.stream = newStream();
@@ -189,7 +189,7 @@ abstract class RetriableStream<ReqT> implements ClientStream {
       state.buffer.add(new StartEntry());
     }
 
-    Substream substream = newSubstream();
+    Substream substream = createSubstream();
     drain(substream);
 
     // TODO(zdapeng): schedule hedging if needed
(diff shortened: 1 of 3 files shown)


dapengzhang0:
cc @ericgribkoff

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -536,6 +548,21 @@ State drained(Substream substream) {
       return new State(buffer, drainedSubstreams, winningSubstream, cancelled, passThrough);
     }
 
+    /** The given substream is closed. */
+    @CheckReturnValue
+    @GuardedBy("lock")
+    State closed(Substream substream) {
+      substream.closed = true;
+      if (this.drainedSubstreams.contains(substream)) {
+        Set<Substream> drainedSubstreams = new HashSet<Substream>();
```
ejona86: Use the `HashSet(Collection)` copy constructor.

core/src/test/java/io/grpc/internal/RetriableStreamTest.java
```diff
@@ -683,6 +688,48 @@ public void messageAvailable() {
...
+    final ClientStream mockStream3 = mock(ClientStream.class);
+    final ClientStream mockStream2 =
+        mock(
+            ClientStream.class,
+            delegatesTo(
+                new NoopClientStream() {
+                  @Override
+                  public void start(ClientStreamListener listener) {
+                    // closed while draning and retry
+                    doReturn(true).when(retriableStreamRecorder).shouldRetry();
```
ejona86: Changing stub return values within this start seems weird, as I'm not sure why they are related. It also makes the test harder to follow. Why isn't this just done early in `closedWhileDraining()`? It doesn't look like it ever will return `false` now.

core/src/test/java/io/grpc/internal/RetriableStreamTest.java
```diff
@@ -683,6 +688,48 @@ public void messageAvailable() {
     verify(masterListener).messagesAvailable(messageProducer);
   }
 
+  @Test
+  public void closedWhileDraining() {
+    ClientStream mockStream1 = mock(ClientStream.class);
+    doReturn(mockStream1).when(retriableStreamRecorder).newSubstream();
```
ejona86: Use `doReturn(mockStream1, mockStream2, mockStream3).when(retriableStreamRecorder).newSubstream()` instead?

dapengzhang0: Seems the current version mockito we used does not support vargs.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -166,46 +176,48 @@ public final void start(ClientStreamListener listener) {
...
+    Substream substream = newSubstream();
     drain(substream);
 
     // TODO(zdapeng): schedule hedging if needed
   }
 
   @Override
   public final void cancel(Status reason) {
-    if (commit0(new NoopClientStream())) {
+    Substream noopSubstream = newSubstream();
```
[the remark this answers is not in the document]
dapengzhang0: Uhh, good catch! The name is so errorprone.

core/src/test/java/io/grpc/internal/RetriableStreamTest.java
dapengzhang0: I wouldn't do this if mockito doReturn supported varargs. And I don't like `when(a.method()).thenReturn(b).thenReturn(c)`

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -536,6 +548,21 @@ State drained(Substream substream) {
       return new State(buffer, drainedSubstreams, winningSubstream, cancelled, passThrough);
     }
 
+    /** The given substream is closed. */
+    @CheckReturnValue
+    @GuardedBy("lock")
+    State closed(Substream substream) {
```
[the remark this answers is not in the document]

core/src/main/java/io/grpc/internal/RetriableStream.java

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -550,7 +550,7 @@ State drained(Substream substream) {
     /** The given substream is closed. */
     @CheckReturnValue
     @GuardedBy("lock")
-    State closed(Substream substream) {
+    State sbustreamClosed(Substream substream) {
```
ejona86: s/sbustream/substream/

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -505,32 +526,49 @@ State cancelled() {
...
         buffer = null;
       }
 
       return new State(buffer, drainedSubstreams, winningSubstream, cancelled, passThrough);
     }
 
+    /** The given substream is closed. */
+    @CheckReturnValue
+    @GuardedBy("lock")
+    State sbustreamClosed(Substream substream) {
```
ejona86: s/sbustream/substream/

(since github thinks my other comment is resolved)

PR #3830, opened 2017-12-05
Release commits for 1.7.1

ejona86:


dapengzhang0:
In v1.7.x thrift was not dead. So there are some version numbers in thrift.

ejona86:
Fixed.

PR #3834, opened 2017-12-05
buildscripts: Remove dependency on 7za

ejona86:
We already depend on Powershell, so this should make it easier to use the
script with less set-up.


ejona86:
With this and porting the Windows version of check-artifact.sh (in compiler/build.gradle) from https://github.com/ejona86/grpc-java/commit/147d12fe407448e879855c5b645b02e94f2f371e , maybe we can simplify our Windows release build setup.

PR #3837, opened 2017-12-06
netty: add test for NettyServerBuilder

jyane:


thelinuxfoundation:

* User @jyane isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,

netty/src/test/java/io/grpc/netty/NettyServerBuilderTest.java
```diff
@@ -69,4 +69,52 @@ public void failIfKeepAliveTimeoutNegative() {
 
     NettyServerBuilder.forPort(8080).keepAliveTimeout(-10L, TimeUnit.HOURS);
   }
+
+  @Test
+  public void failIfMaxConcurrentCallsPerConnectionNegative() {
+    thrown.expect(IllegalArgumentException.class);
+    thrown.expectMessage("max must be positive");
+
+    NettyServerBuilder.forPort(8080).maxConcurrentCallsPerConnection(0);
```
dapengzhang0: nit: Declare `builder = NettyServerBuilder.forPort(8080)` prior to `thrown.expect`

ejona86: Actually... we should probably just make that a field at the top, since so many of these tests will use that. That said, these new tests are working like the old tests, so it seems fair.

jyane:
I have fixed all test cases to declare `builder` at a top in NettyServerBuilderTest, PTAL.

netty/src/test/java/io/grpc/netty/NettyServerBuilderTest.java
```diff
@@ -21,6 +21,8 @@
 
 import io.netty.handler.ssl.SslContext;
 import java.util.concurrent.TimeUnit;
+
```
ejona86: We only add extra lines between static and non-static imports. I'll remove this.

netty/src/test/java/io/grpc/netty/NettyServerBuilderTest.java
```diff
@@ -36,9 +38,15 @@
 
   @Rule public final ExpectedException thrown = ExpectedException.none();
 
+  private NettyServerBuilder builder = null;
+
+  @Before
+  public void setUp() {
+    builder = NettyServerBuilder.forPort(8080);
```
ejona86: When possible, initialize variables at their declaration. There's no need for a `@Before` here. I'll fix this.

ejona86:
Thank you!

netty/src/test/java/io/grpc/netty/NettyServerBuilderTest.java
jyane: I'm sorry for this extra line. I have set the Google Style xml to my IDE.

PR #3840, opened 2017-12-06
core: retry support part 1 followup, channel.shutdown should not fail retry in backoff

dapengzhang0:
As pointed out by @zhangkun83 delayedTransport `shutdown` will make retry attempts that are in backoff unable to create new stream. So wrap delayedTransport with `retriableTransport`, which will keep uncommitted retriable streams alive when `shutdown`.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -56,6 +56,8 @@ import java.lang.ref.SoftReference;
 import java.lang.ref.WeakReference;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -160,6 +162,7 @@ public final class ManagedChannelImpl extends ManagedChannel implements Internal
 
   // reprocess() must be run from channelExecutor
   private final DelayedClientTransport delayedTransport;
+  private final RetriableTransport retriableTransport = new RetriableTransport();
 
   // Shutdown states.
   //
@@ -400,12 +403,12 @@ public final class ManagedChannelImpl extends ManagedChannel implements Internal
       return new RetriableStream<ReqT>(method) {
         @Override
         Status prestart() {
-          return delayedTransport.addUncommittedRetriableStream(this);
+          return retriableTransport.addUncommittedRetriableStream(this);
         }
 
         @Override
         void postCommit() {
-          delayedTransport.removeUncommittedRetriableStream(this);
+          retriableTransport.removeUncommittedRetriableStream(this);
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -694,6 +697,85 @@ public void run() {
         }).drain();
   }
 
+  /**
+   * A logical "transport" that prevents channel shutdown from killing existing retry attempts that
+   * are in backoff.
+   */
+  private final class RetriableTransport {
```
zhangkun83: This could use a better name. Maybe "PendingStreamRegistry"

dapengzhang0:
Renamed, PTAL
Test will be added once retry backoff is implemented.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -694,6 +698,87 @@ public void run() {
...
+  /**
+   * A registry that prevents channel shutdown from killing existing retry attempts that are in
+   * backoff.
+   */
+  // TODO(zdapeng): add test coverage for shutdown during retry backoff once retry backoff is
+  //                implemented.
+  private final class UncommittedRetriableStreamsRegistry {
+    final Object lock = new Object();
+
+    @GuardedBy("lock")
```
zhangkun83: This means we would acquire a lock for each new retry-able stream.
The non-retry-able stream has a fast path that is lock-free. Maybe we should spend some time to see if we can make the retry-able path lock-free. Please add a TODO.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -694,6 +698,87 @@ public void run() {
...
+        if (uncommittedRetriableStreams.isEmpty()) {
+          shutdownStatusCopy = shutdownStatus;
+          // Because retriable transport is long-lived, we take this opportunity to down-size the
+          // hashmap.
+          uncommittedRetriableStreams = new HashSet<ClientStream>();
+        }
+      }
+
+      if (shutdownStatusCopy != null) {
+        delayedTransport.shutdown(shutdownStatusCopy);
```
zhangkun83: Please add some comments here for why we keep the delayedTransport open until there is no more uncommitted streams. Also add a TODO to write test for this specific case.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -694,6 +697,85 @@ public void run() {
...
+        shutdownStatus = reason;
+        if (uncommittedRetriableStreams.isEmpty()) {
+          shouldShutdownDelayedTransport = true;
+        }
+      }
+
+      if (shouldShutdownDelayedTransport) {
+        delayedTransport.shutdown(reason);
+      }
+      channelExecutor.drain();
```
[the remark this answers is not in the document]
dapengzhang0: Removed this line.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: Added TODO.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: Comments added.
TODO added at the top of the class.

PR #3844, opened 2017-12-07
netty: temp disable flow control tests for #3856

carl-mastrangelo:
The previous getter is not thread safe, since it does an unsynchronized access to the netty thread.   The test is flaky, but I think this is just a part of it.


netty/src/main/java/io/grpc/netty/NettyHandlerSettings.java
```diff
@@ -56,17 +63,23 @@ public static synchronized void autoWindowOn(boolean autoFlowControl) {
...
   }
 
-  public static synchronized int getLatestServerWindow() {
-    return getLatestWindow(serverHandler);
+  public static synchronized Future<Integer> getLatestServerWindow() {
+    return getLatestWindow(serverLoop, serverHandler);
   }
 
-  private static synchronized int getLatestWindow(AbstractNettyHandler handler) {
+  private static synchronized Future<Integer> getLatestWindow(
```
dapengzhang0: As this is run in the loop, `synchronized` seems not necessary.

carl-mastrangelo: I guess this method doesn't need it, but the calls do.   The need to synchronize getting the loop and the handler with `setAutoWindow`

netty/src/main/java/io/grpc/netty/NettyHandlerSettings.java
```diff
@@ -56,17 +63,23 @@ public static synchronized void autoWindowOn(boolean autoFlowControl) {
     autoFlowControlOn = autoFlowControl;
   }
 
-  public static synchronized int getLatestClientWindow() {
-    return getLatestWindow(clientHandler);
+  public static synchronized Future<Integer> getLatestClientWindow() {
+    return getLatestWindow(clientLoop, clientHandler);
```
dapengzhang0: `checkNotNull(clientLoop, "setAutoWindow must be called first")` ? 

Also I don't see the point to synchronized it, unless `setAutoWindow` can be reset for multiple times in different threads which does not make sense and is make trouble.

carl-mastrangelo: since this method doesn't actually use the method, I am reluctant to put the check here.  I put it down where it is used.

carl-mastrangelo:
@ejona86 ptal

interop-testing/src/test/java/io/grpc/testing/integration/AutoWindowSizingOnTest.java
```diff
@@ -31,8 +30,11 @@
 
   @BeforeClass
   public static void turnOnAutoWindow() {
-    InternalHandlerSettings.enable(true);
-    InternalHandlerSettings.autoWindowOn(true);
+    // TODO(carl-mastrangelo): Reenable this when https://github.com/grpc/grpc-java/issues/3856 is
```
ejona86: This bothers me, as appearance is quite different from actuality. Could we do an Assume.assumeTrue(false) instead?

interop-testing/src/test/java/io/grpc/testing/integration/NettyFlowControlTest.java
```diff
@@ -76,6 +77,8 @@
 
   @BeforeClass
   public static void setUp() {
+    // TODO(carl-mastrangelo): reenabled after https://github.com/grpc/grpc-java/issues/3856 is
```
ejona86: Remove this comment?

PR #3852, opened 2017-12-09
core/grpclb: resolve TXT records in DNS name resolver and include balancer addresses

carl-mastrangelo:
This is part one of handling service config, and also enabled grpclb to work.  I manually tested this with some complex DNS data, but I dont have many unit tests because we dont have a dummy DNS server.   

Handling or the SRV records logs failures and tries to keep going.

### DIFF
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -190,12 +190,12 @@ final class DnsNameResolver extends NameResolver {
           for (InetAddress inetAddr : resolvedInetAddrs.addresses) {
             servers.add(new EquivalentAddressGroup(new InetSocketAddress(inetAddr, port)));
           }
-          servers.addAll(resolvedInetAddrs.srvRecords);
+          servers.addAll(resolvedInetAddrs.balancerAddresses);
 
           Attributes.Builder attrs = Attributes.newBuilder();
           if (!resolvedInetAddrs.txtRecords.isEmpty()) {
             attrs.set(
-                GrpcAttributes.NAME_RESOLVER_ATTR_TXT,
+                GrpcAttributes.NAME_RESOLVER_ATTR_DNS_TXT,
                 Collections.unmodifiableList(new ArrayList<String>(resolvedInetAddrs.txtRecords)));
           }
           savedListener.onAddresses(servers, attrs.build());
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -59,8 +63,13 @@
...
   private static final boolean JNDI_AVAILABLE = jndiAvailable();
 
+  // From https://github.com/grpc/proposal/blob/master/A2-service-configs-in-dns.md
+  private static final String SERVICE_CONFIG_NAME_PREFIX = "_grpc_config.";
+  // From https://github.com/grpc/proposal/blob/master/A5-grpclb-in-dns.md
+  private static final String GRPCLB_NAME_PREFIX = "_grpclb._tcp.";
+
   @VisibleForTesting
-  static boolean enableJndi = false;
+  static boolean enableJndi = true;
```
ejona86: We should keep this off until it is stable.

carl-mastrangelo: Put it behind a flag so that it can be tested before release.

markdroth:
For C-core, we have some tests that use a DNS server.  You can talk to @apolcyn about how this was set up.

carl-mastrangelo:
@zhangkun83  ping

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -278,10 +298,15 @@ static boolean jndiAvailable() {
   static final class ResolutionResults {
     final List<InetAddress> addresses;
     final List<String> txtRecords;
+    final List<EquivalentAddressGroup> srvRecords;
 
-    ResolutionResults(List<InetAddress> addresses, List<String> txtRecords) {
+    ResolutionResults(
+        List<InetAddress> addresses,
+        List<String> txtRecords,
+        List<EquivalentAddressGroup> srvRecords) {
```
zhangkun83: These are actually balancer addresses. They are not intended to be generic SRV records. Please name the variables (including those in CompositeResolver) accordingly.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -174,11 +186,19 @@ public void run() {
...
           for (InetAddress inetAddr : resolvedInetAddrs.addresses) {
             servers.add(new EquivalentAddressGroup(new InetSocketAddress(inetAddr, port)));
           }
-          savedListener.onAddresses(servers, Attributes.EMPTY);
+          servers.addAll(resolvedInetAddrs.srvRecords);
+
+          Attributes.Builder attrs = Attributes.newBuilder();
+          if (!resolvedInetAddrs.txtRecords.isEmpty()) {
+            attrs.set(
+                GrpcAttributes.NAME_RESOLVER_ATTR_TXT,
```
zhangkun83: TXT record is DNS-specific. Let's add "DNS" to the name of this attribute.

carl-mastrangelo: Done.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
carl-mastrangelo: Done.

PR #3853, opened 2017-12-10
core: support user-agent in InProcess

jyane:
Currently, InProcess(io.grpc.inprocess) does not support the use of a user-agent, but it may be useful for testing of `LoggingInterceptor`(It is just an example, it records the status, responses, user-agent, etc to log files).

### DIFF
--- a/core/src/main/java/io/grpc/inprocess/InProcessTransport.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessTransport.java
@@ -34,6 +34,7 @@ import io.grpc.Status;
 import io.grpc.internal.ClientStream;
 import io.grpc.internal.ClientStreamListener;
 import io.grpc.internal.ConnectionClientTransport;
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/grpc/inprocess/InProcessTransport.java
```diff
@@ -143,6 +146,9 @@ public void start(ClientStreamListener listener) {
         }
       };
     }
+    headers.put(
+        GrpcUtil.USER_AGENT_KEY,
+        GrpcUtil.getGrpcUserAgent("inprocess", userAgent));
```
carl-mastrangelo: nit: the agent should be a final var for this instance.

ejona86: I'll do that as a follow-up.

ejona86:
I created #3860 for the optimization.

PR #3858, opened 2017-12-11
core: Use hard-coded string for version

### LINKED ISSUE #2098: Android has version "0.0" in user agent
It should have the actual version.

ejona86:
Using META-INF for loading the version is broken on Android and
frequently broken when gRPC is shaded. Having hard-coded strings be
replaced on version bumps has been working well.

Fixes #2098

### DIFF
--- a/core/src/main/java/io/grpc/internal/GrpcUtil.java
+++ b/core/src/main/java/io/grpc/internal/GrpcUtil.java
@@ -194,7 +194,7 @@ public final class GrpcUtil {
 
   public static final Splitter ACCEPT_ENCODING_SPLITTER = Splitter.on(',').trimResults();
 
-  private static final String IMPLEMENTATION_VERSION = getImplementationVersion();
+  private static final String IMPLEMENTATION_VERSION = "1.9.0-SNAPSHOT"; // CURRENT_GRPC_VERSION
 
   /**
    * The default delay in nanos before we send a keepalive.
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -194,7 +194,7 @@
 
   public static final Splitter ACCEPT_ENCODING_SPLITTER = Splitter.on(',').trimResults();
 
-  private static final String IMPLEMENTATION_VERSION = getImplementationVersion();
+  private static final String IMPLEMENTATION_VERSION = "1.9.0-SNAPSHOT"; // CURRENT_GRPC_VERSION
```
carl-mastrangelo: according to `Package`, the version needs to be numerics and dots, no snapshot version.

ejona86: What is the implication?

core/src/main/java/io/grpc/internal/GrpcUtil.java
```diff
@@ -736,12 +736,4 @@ static void closeQuietly(InputStream message) {
   }
 
   private GrpcUtil() {}
-
-  private static String getImplementationVersion() {
-    String version = GrpcUtil.class.getPackage().getImplementationVersion();
-    if (version != null) {
-      return "/" + version;
```
xuwei-k: missing `"/"` ?

`grpc-java-netty/1.8.0` => `grpc-java-netty1.9.0`


https://github.com/grpc/grpc/blob/v1.8.4/doc/PROTOCOL-HTTP2.md#user-agents

> User-Agent → "grpc-" Language ?("-" Variant) "/" Version ?( " ("  *(AdditionalProperty ";") ")" )

ejona86: :cry: Sent out #3974

Issue #3857, opened 2017-12-11
io.grpc.Attributes - reference vs. object equality


nicktrav:
`io.grpc.Attributes` was modified in 291f17059eee0fee0bf0d0d6c3ddc6fc40e2a09c to use an `IdentityMap` to store the underlying attributes. This map uses reference equality in place of the object equality that was used in prior versions, when the class used `HashMap`.

This is more of a question around the _expected_ behavior of `Attributes#equals` and the notion of "attribute equality" as opposed to a bug or an issue. For context, we have some tests internally that are failing in 1.8 as they were written with object equality in mind.

For example, the following fails in 1.8 but passes in prior versions, which upon immediate inspection wasn't intuitive, as `InetSocketAddress#equals` returns `true` when comparing the two different objects:

```java
  @Test
  public void equals() {
    Attributes.Key<InetSocketAddress> attr = Attributes.Key.of("foo");

    InetSocketAddress addr1 = new InetSocketAddress(80);
    Attributes attrs1 = Attributes.newBuilder()
        .set(attr, addr1)
        .build();

    InetSocketAddress addr2 = new InetSocketAddress(80);
    Attributes attrs2 = Attributes.newBuilder()
        .set(attr, addr2)
        .build();

    assertEquals(addr1, addr2);   // passes
    assertEquals(attrs1, attrs2); // fails
  }
```

I'm interested in the motivation for the change in semantics (talking with @lukaszx0 in person, he recalls some talk about this previously). If reference equality was the intention, perhaps we could tighten up the javadoc a little to make it more explicit that this class now uses reference equality as opposed to object equality?

Or, perhaps this wasn't the original intention and it would be possible to revert to using a `HashMap`, although the benchmarks seem to indicate this would be slightly slower than the current impl.

carl-mastrangelo:
I don't believe that side effect was intended, only the keys were supposed to be compared by reference equality.   We could change it back, because it seems like a legit use case.  

Do you mainly use it in tests or in prod code?

ejona86:
We should change it back. Yeah, we should use object-equality for the values themselves.

nicktrav:
@carl-mastrangelo @ejona86 - from what I can tell we're only relying on this in tests right now.

Thanks for the prompt reply / fix nonetheless! 🙏

nicktrav:
(ignore my trigger finger on the close button)

PR #3859, opened 2017-12-11
core: fix Attributes value equality

carl-mastrangelo:
Fixes: #3857

### DIFF
--- a/core/src/main/java/io/grpc/Attributes.java
+++ b/core/src/main/java/io/grpc/Attributes.java
@@ -129,7 +129,18 @@ public final class Attributes {
       return false;
     }
     Attributes that = (Attributes) o;
-    return Objects.equal(data, that.data);
+    if (data.size() != that.data.size()) {
+      return false;
+    }
+    for (Entry<Key<?>, Object> e : data.entrySet()) {
+      if (!that.data.containsKey(e.getKey())) {
+        return false;
+      }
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/Attributes.java
```diff
@@ -129,7 +129,18 @@ public boolean equals(Object o) {
       return false;
     }
     Attributes that = (Attributes) o;
-    return Objects.equal(data, that.data);
+    if (data.size() != that.data.size()) {
+      return false;
+    }
+    for (Entry<Key<?>, Object> e : data.entrySet()) {
```
ejona86: This only handles one direction. `that.data` may contain an extra element, but this would incorrectly say they were equal.

carl-mastrangelo: the sizes are checked above

carl-mastrangelo:
Identity hashmap is slightly faster.  Equals is an uncommon operation, while get and put are common.

core/src/main/java/io/grpc/Attributes.java
ejona86: Oh, you're right. kk

ejona86:
While a microbenchmark can possibly notice a minor difference in performance, that level of optimization seems unlikely to matter at all in the greater scheme of things. Instead, it just makes more complicated code and bugs. But whatever.

carl-mastrangelo:
jenkins retest this please

PR #3861, opened 2017-12-12
Add a service config file

carl-mastrangelo:
This is the next part to enabling service config.  The files added try to represent the JSON as closely as possible.  There are currently no getters since the data isn't used yet.  The next step will be changing MCI to use this data, but that depends on The TXT records being returned from the DNS name resolver.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ServiceConfigFile.java
+++ b/core/src/main/java/io/grpc/internal/ServiceConfigFile.java
@@ -1,0 +1,436 @@
+/*
+ * Copyright 2017, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/grpc/internal/ServiceConfigFile.java
```diff
@@ -0,0 +1,436 @@
...
+import com.google.common.base.MoreObjects;
+import com.google.common.base.Objects;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Represents a ServiceConfig as described at
+ * https://github.com/grpc/grpc/blob/master/doc/service_config.md
+ */
+public final class ServiceConfigFile {
```
zhangkun83: Please document that all fields are nullable.

carl-mastrangelo: Done.   All fields are nullable so that this class can represent the json appropriately.   That said, some of the nullable fields are actually required

PR #3872, opened 2017-12-15
core: introduce io.grpc.internal.BinaryLogProvider

zpencer:


core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -0,0 +1,138 @@
...
+        public int compare(BinaryLogProvider f1, BinaryLogProvider f2) {
+          return f1.priority() - f2.priority();
+        }
+      });
+    }
+  }
+
+  private static ServiceLoader<BinaryLogProvider> getCandidatesViaServiceLoader(
+      ClassLoader classLoader) {
+    ServiceLoader<BinaryLogProvider> i = ServiceLoader.load(BinaryLogProvider.class, classLoader);
```
ejona86: This is _very_ expensive on Android because it requires loading/validating the apk zip and then it can leave considerable amount of memory left-over for the zip table. If we detect Android we could just use NULL_PROVIDER.

Issue #3864, opened 2017-12-14
Freshly checked out code does not build in Intellij IDEA


rmichela:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?
1.9.0-SNAPSHOT

Also reproduced with 1.7.0 and 1.8.0 tagged versions, and the past three major releases of Intellij Idea.

### Steps to reproduce
1. Check out the latest grpc-java code to a fresh directory
2. Run `./gradlew idea` to create the Intellij Idea project files
3. Open grpc-java with Idea
4. Attempt to build the project using the `Build -> Build Project` menu

### Expected
Grpc-java builds in Idea

### Actual
1. The `protobuf/test` module does not compile, complaining about missing methods and type mismatches.
2. `io.grpc.netty.shaded.ShadingTest` does not compile, because it cannot locate the shaded netty jar on the Idea classpath.

### Notes
* I was able to fix the `protobuf/test` module compilation by changing `sourcesRoot` to `testSourcesRoot` in protobuf-lite/build.gradle line 63.
* I was able to manually fix `io.grpc.netty.shaded.ShadingTest` by explicitly adding the generated `grpc-netty-shaded-1.9.0-SNAPSHOT.jar` file to the `grpc-netty-shaded_testShadow` module using Intellij's Project Structure screen. This is sub-optimal. I was unable to find a solution using gradle.

rmichela:
Looks like the second part of this defect is a know bug in Idea: https://youtrack.jetbrains.com/issue/IDEA-163411

ejona86:
Thanks for the PR. It looks like there's no fix for the second problem except for waiting for a fix on the shadow or IDEA side :frowning_face:

rmichela:
**Workaround for IDEA users.**

1. `./gradlew :grpc-netty-shaded:build`
2. In IDEA, expand the project tree to find `netty/shaded/src/testShadow`
3. Right-click `testShadow` module and select "Open Module Settings"
4. Find the `grpc-netty-shaded_testShadow` sub-project and select the "Dependencies" tab
5. Click the ➕ at the bottom and select "Jar or directories..."
6. Select the file `netty/shaded/build/libs/grpc-netty-shaded-[version].jar`
7. From the "Build" menu, select "Build Project"

zpencer:
@rmichela out of curiosity, have you tried loading the project into intellij as a gradle project, and then delegating build/run operations to gradle?
[...]

PR #3875, opened 2017-12-18
Fix interop tests in IDEA

rmichela:
Interop tests fail to run in IDEA due to a missing dependency on Conscrypt. Without the Conscrypt dependency, you cannot debug these tests.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -212,6 +212,7 @@ subprojects {
                 junit: 'junit:junit:4.12',
                 mockito: 'org.mockito:mockito-core:1.9.5',
                 truth: 'com.google.truth:truth:0.36',
+                conscrypt: 'org.conscrypt:conscrypt-openjdk-uber:1.0.0.RC14',
 
                 // Benchmark dependencies
                 hdrhistogram: 'org.hdrhistogram:HdrHistogram:2.1.10',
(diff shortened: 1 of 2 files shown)


rmichela:
Helps with #3864

rmichela:
@ejona86 Can you take a look at this PR?

rmichela:
Specifically, the `Http2OkHttpTest` is fixed by this PR.

ejona86:
We're not using Conscrypt at all for testing. I expect what's actually broken is the `jvmArgs "-javaagent:" + configurations.alpnagent.asPath` line isn't being applied/working.

If this was in okhttp/ it wouldn't be much of a concern, but I think [TlsTest](https://github.com/grpc/grpc-java/blob/master/interop-testing/src/test/java/io/grpc/testing/integration/TlsTest.java) might accidentally stop testing Jetty ALPN and use Conscrypt instead. That's a bit dangerous because we do still want to test Jetty ALPN. As part of adding full Conscrypt support I was going to be teaching it how to distinguish between them.

rmichela:
We can abandon this change by using gradle for IntelliJ's build and test actions instead of the default IntelliJ compiler.

https://github.com/grpc/grpc-java/pull/3958

ejona86:
Okay. Sounds good. That'll probably have fewer breakages long-term.

PR #3876, opened 2017-12-18
Preserve ManagedChannel when stub.withInterceptors() is invoked

rmichela:
Creating a stub around a `ManagedChannel` creates a stub who's `getChannel()` method returns a `ManagedChannel`. 

While `getChannel()`'s return type is only `Channel`, it is convenient to downcast `getChannel()`'s result to a `ManagedChannel` to shut down the `Channel`'s supporting `ManagedChannel`. However, if the stub's user adds an interceptor to the stub by calling `withInterceptors()`, `getChannel()` returns a `Channel` instead of a `ManagedChannel` (technically, an `InterceptorChannel`).

Technically, changing a `ManagedChannel` to a `Channel` complies with the public API for `AbstractStub`. While `stub.getChannel()` makes no guarantees about the type of `Channel` it returns, it does make sense that the type would be the same as the `Channel` that started the stub. 

This PR makes `getChannel()` behave more consistently. If a stub is started around a `ManagedChannel`, `stub.getChannel()` will return a `ManagedChannel`, even after one or more interceptors are added with `stub.withInterceptors()`.

### DIFF
--- a/core/src/main/java/io/grpc/ClientInterceptors.java
+++ b/core/src/main/java/io/grpc/ClientInterceptors.java
@@ -17,6 +17,8 @@
 package io.grpc;
 
 import com.google.common.base.Preconditions;
+import io.grpc.internal.ManagedChannelChain;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -83,10 +85,16 @@ public class ClientInterceptors {
    */
   public static Channel intercept(Channel channel, List<? extends ClientInterceptor> interceptors) {
     Preconditions.checkNotNull(channel, "channel");
+    Channel newChannel = channel;
     for (ClientInterceptor interceptor : interceptors) {
-      channel = new InterceptorChannel(channel, interceptor);
+      newChannel = new InterceptorChannel(newChannel, interceptor);
     }
-    return channel;
+
+    if (channel instanceof ManagedChannel) {
+      return new ManagedChannelChain(newChannel, (ManagedChannel) channel);
+    } else {
+      return newChannel;
+    }
   }
 
   private static class InterceptorChannel extends Channel {
--- a/core/src/main/java/io/grpc/internal/ManagedChannelChain.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelChain.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2016, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.internal;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.InternalChannelStats;
+import io.grpc.InternalLogId;
+import io.grpc.ManagedChannel;
+import io.grpc.MethodDescriptor;
+
+import java.util.concurrent.TimeUnit;
+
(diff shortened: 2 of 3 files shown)


rmichela:
@ramaraochavali

carl-mastrangelo:
This would still be kind of off if the MC passed in was a sub class.  For example:

```
class MyManagedChannel extends ManagedChannel {
  String toString() { return "My Custom Channel" }
}
```

Calling getChannel will return a ManagedChannel different than the one passed in.

rmichela:
I'm not sure it's possible to get perfect type preservation of `getChannel()` without significantly changing how interceptors work. This PR at least preserves the "ManagedChannel-ness" of the supporting channel after wrapping it in an interceptor.

rmichela:
@carl-mastrangelo @ejona86 I'm following up on this PR. Is it enough to preserve the "ManagedChannel-ness" of a channel after wrapping it in an interceptor? Is there any more you think that needs doing?

rmichela:
@carl-mastrangelo @ejona86 I'm following up on this PR. Is it enough to preserve the "ManagedChannel-ness" of a channel after wrapping it in an interceptor? Is there any more you think that needs doing?

core/src/main/java/io/grpc/ClientInterceptors.java
```diff
@@ -17,6 +17,7 @@
 package io.grpc;
 
 import com.google.common.base.Preconditions;
+import io.grpc.internal.ManagedChannelChain;
```
ejona86: FYI: io.grpc can't depend on io.grpc.internal. It seems ManagedChannelChain should just be package-private in io.grpc.

core/src/main/java/io/grpc/ClientInterceptors.java
```diff
@@ -83,10 +84,16 @@ public static Channel intercept(Channel channel, ClientInterceptor... intercepto
    */
   public static Channel intercept(Channel channel, List<? extends ClientInterceptor> interceptors) {
     Preconditions.checkNotNull(channel, "channel");
+    Channel newChannel = channel;
     for (ClientInterceptor interceptor : interceptors) {
-      channel = new InterceptorChannel(channel, interceptor);
+      newChannel = new InterceptorChannel(newChannel, interceptor);
+    }
+
+    if (channel instanceof ManagedChannel) {
```
ejona86: A ManagedChannel overload _may_ be better. Although it is sort of nice we're not advertising this too much in this current implementation. Hmm...

ejona86:
I don't have much problem with only preserving the ManagedChannel-ness.

We went through this with the initial interceptor design and we were _very_ scared of the confusion caused by having multiple things that can be shutdown that each look different but are actually the same resource. That is still a fundamental problem of a design that tries to re-discover the ManagedChannel: there has to be implicit ownership requirements and sharing disallowed. This is _why_ we have separate `Channel` and `ManagedChannel` and no way to recover the original `ManagedChannel`.

I'm really happy with how `ManagedChannelBuilder.intercept` has turned out, in that it solves the common case. But you're right, if you're adding interceptors to a stub, then you lose the original ManagedChannel.

ejona86:
The various code changes involved here isn't much. I think we need to focus on the use-case and what makes it hard to pass the `ManagedChannel` explicitly or to save the `ManagedChannel` before adding interceptors to the stub (or any other similar pain).

ejona86:
> I think we need to focus on the use-case and what makes it hard to pass the ManagedChannel explicitly or to save the ManagedChannel before adding interceptors to the stub (or any other similar pain).

@rmichela, can you help paint that picture for us?

rmichela:
* In an effort to simplify application developer interaction with gRPC, we've created a builder that builds service stubs given an address and a port. (address, port) -> stub
* The stub builder sets all the configuration options needed to interoperate with our service mesh.
* Using Spring, application developers have a simple idiom for instantiating a stub using our builder, and injecting it into their business logic as needed.
* A simple Spring-based idiom for gracefully shutting down a client stub with the application is to hook the Spring bean shutdown event and call `((ManagedChannel) stub.getChannel()).shutdown();`
    * We encapsulate the `shutdown()` call in a library method, so application developers are not exposed to `ManagedChannel` itself.
    * We ensure our builder returns stubs built around `ManagedChannel`, however
    * Our shutdown library method breaks if the user has added their own interceptor to the stub.

ejona86:
@rmichela, thank you. That's really useful. I still need to read it more carefully, and may glance at some of the Spring glue. But I can agree that seems disappointing.

dapengzhang0:
Ping @ejona86. What's the status of this PR?

sanjaypujare:
Is there any interest in this PR? Or can we close it?

PR #3879, opened 2017-12-18
core: Install the ServerInterceptor from BinaryLogProvider

zpencer:
The interceptor is installed such that the method appears to be
`Method<InputStream, InputStream>` to the interceptor code. The
boilerplate in this PR is responsible for converting to/from
`<ReqT, RespT>`.


core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
```diff
@@ -16,20 +16,89 @@
...
 import org.junit.runners.JUnit4;
 
 /** Unit tests for {@link BinaryLogProvider}. */
 @RunWith(JUnit4.class)
 public class BinaryLogProviderTest {
+  private static final MethodDescriptor<Integer, Integer> INCREMENT_BY_ONE =
+      MethodDescriptor.<Integer, Integer>newBuilder()
+          .setType(MethodDescriptor.MethodType.UNARY)
+          .setFullMethodName("some/incrementbyone")
+          .setRequestMarshaller(new IntegerMarshaller())
```
ejona86: Use `io.grpc.IntegerMarshaller` for these marshallers.

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
```diff
@@ -92,4 +289,21 @@ protected int priority() {
       return priority;
     }
   }
+
+  private static final class IntegerMarshaller implements MethodDescriptor.Marshaller<Integer> {
```
ejona86: Rename this to `CountingMarshaller` or `InvocationCountMarshaller` or similar to avoid confusion?

ejona86: Feel free to extend `io.grpc.IntegerMarshaller`. But it is fine as-is.

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
```diff
@@ -50,6 +119,134 @@ public void unavailableProvider() {
...
+      this.methods = methods;
+    }
+
+    @Override
+    public <ReqT, RespT> Listener<ReqT> interceptCall(
+        final ServerCall<ReqT, RespT> call,
+        Metadata headers,
+        ServerCallHandler<ReqT, RespT> next) {
+      methods.add(call.getMethodDescriptor());
+      ForwardingServerCall<ReqT, RespT> wCall = new ForwardingServerCall<ReqT, RespT>() {
```
ejona86: Extend `SimpleForwardingServerCall`?

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
```diff
@@ -50,6 +119,134 @@ public void unavailableProvider() {
...
+        @Override
+        public void sendMessage(RespT message) {
+          resp.add(message);
+          super.sendMessage(message);
+        }
+
+        @Override
+        public MethodDescriptor<ReqT, RespT> getMethodDescriptor() {
```
ejona86: Delete?

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
```diff
@@ -50,6 +119,134 @@ public void unavailableProvider() {
...
+          return call.getMethodDescriptor();
+        }
+
+        @Override
+        protected ServerCall<ReqT, RespT> delegate() {
+          return call;
+        }
+      };
+      final Listener<ReqT> oListener = next.startCall(wCall, headers);
+      return new ForwardingServerCallListener<ReqT>() {
```
ejona86: `SimpleForwardingServerCallListener`?

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
```diff
@@ -16,20 +16,89 @@
...
+                  new ServerCalls.UnaryMethod<Integer, Integer>() {
+                    @Override
+                    public void invoke(Integer request, StreamObserver<Integer> responseObserver) {
+                      responseObserver.onNext(request + 1);
+                      responseObserver.onCompleted();
+                    }
+                  }
+              ))
+          .build();
+  private final List<Object> req = new ArrayList<Object>();
```
ejona86: These four fields (this and below) are confusing, as they aren't used directly by any infrastructure while also being non-obvious just by looking at them. Since only one method is looking at them, could you move them into that test method for now?

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
```diff
@@ -50,6 +119,134 @@ public void unavailableProvider() {
     assertSame(BinaryLogProvider.NullProvider.class, BinaryLogProvider.load(cl).getClass());
   }
 
+  @Test
+  public void serverInterceptorFirstAndIsInputstream() throws Exception {
+    String serverName = getClass().getName() + "-serverInterceptor";
+    ServerInterceptor interceptor = new TracingServerInterceptor(req, resp, methods);
+    InProcessServerBuilder builder = InProcessServerBuilder
+        .forName(serverName)
```
ejona86: Add `directExecutor()`. This code is not thread-safe and it's simply easier to not think about it than proving it is correct.

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
```diff
@@ -50,6 +119,134 @@ public void unavailableProvider() {
...
+        .addService(service)
+        .intercept(interceptor);
+    TestBinaryLogProvider binlog = new TestBinaryLogProvider(req, resp, binaryMethods);
+    InternalInProcessServerBuilder.setBinaryLogProvider(builder, binlog);
+    server = builder.build();
+    server.start();
+    channel = InProcessChannelBuilder.forName(serverName).build();
```
ejona86: Add `directExecutor()`.

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
```diff
@@ -50,6 +119,134 @@ public void unavailableProvider() {
...
+
+    @Override
+    public ServerInterceptor getServerInterceptor(String fullMethodName) {
+      return serverInterceptor;
+    }
+
+    @Nullable
+    @Override
+    public ClientInterceptor getClientInterceptor(String fullMethodName) {
+      return clientInterceptor;
```
ejona86: `return null`

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
```diff
@@ -50,6 +119,134 @@ public void unavailableProvider() {
     assertSame(BinaryLogProvider.NullProvider.class, BinaryLogProvider.load(cl).getClass());
   }
 
+  @Test
+  public void serverInterceptorFirstAndIsInputstream() throws Exception {
```
ejona86: This test is in the wrong test class. This is really just testing ServerImpl today, as "first" and "InputStream" are mostly decided by ServerImpl (yes, InputStream isn't, but _converting_ the type is, and forgetting to convert is much more likely than choosing the wrong type).

Granted, I've provided some suggestions that would shift that around a bit, but I think a lot of the infrastructure here would still move to ServerImplTest. And then you will have a simpler test here.

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
```diff
@@ -99,4 +101,10 @@ protected InProcessServer buildTransportServer(
   public InProcessServerBuilder useTransportSecurity(File certChain, File privateKey) {
     throw new UnsupportedOperationException("TLS not supported in InProcessServer");
   }
+
+  @Override
+  @VisibleForTesting
+  protected void setBinaryLogProvider(BinaryLogProvider value) {
```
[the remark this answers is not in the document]
zpencer: Made `binlogProvider` a protected field so this can be a package-private non-override method.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -117,7 +120,8 @@
   ServerImpl(
       AbstractServerImplBuilder<?> builder,
       InternalServer transportServer,
-      Context rootContext) {
+      Context rootContext,
+      BinaryLogProvider binlogProvider) {
```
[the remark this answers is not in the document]
zpencer: Done

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -43,6 +58,14 @@ public static BinaryLogProvider provider() {
     return PROVIDER;
   }
 
+  /**
+   * Creates a new {@link MethodDescriptor} where both marshallers are {@code InputStream} based.
+   */
+  static <ReqT, RespT> MethodDescriptor<InputStream, InputStream> wrapMethod(
+      MethodDescriptor<ReqT, RespT> method) {
```
carl-mastrangelo: These shouldn't be captures, but `<?, ?>`

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -510,15 +515,45 @@ public void cancelled(Context context) {
...
+      ServerInterceptor binlogInterceptor = binlogProvider.getServerInterceptor(fullMethodName);
+      if (binlogInterceptor == null) {
+        return createListener(
+            stream, fullMethodName, methodDef.getMethodDescriptor(), headers, context,
+            statsTraceCtx, callHandler);
+      } else {
+        // This modified MethodDescriptor is not visible to users:
+        // The HandlerRegistry and ServerInterceptor APIs return the original MethodDescriptor.
+        MethodDescriptor<InputStream, InputStream> binaryMethod =
+            BinaryLogProvider.wrapMethod(methodDef.getMethodDescriptor());
```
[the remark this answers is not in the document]
zpencer: Moved to BinaryLogProvider

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -510,15 +515,45 @@ public void cancelled(Context context) {
...
+        ServerCallHandler<InputStream, InputStream> binaryCallHandler =
+            InternalServerInterceptors.wrapHandler(
+                callHandler, methodDef.getMethodDescriptor(), binaryMethod);
+        binaryCallHandler = InternalServerInterceptors.interceptCallHandler(
+            binlogInterceptor, binaryCallHandler);
+        return createListener(
```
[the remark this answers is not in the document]
zpencer: Moved

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
zpencer: Refactored

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
zpencer: Refactored

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
zpencer: Moved

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
zpencer: No longer applicable.

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -61,11 +64,32 @@ public static BinaryLogProvider provider() {
...
-      MethodDescriptor<ReqT, RespT> method) {
+  private static MethodDescriptor<InputStream, InputStream> wrapMethod(
+      MethodDescriptor<?, ?> method) {
     return method.toBuilder(IDENTITY_MARSHALLER, IDENTITY_MARSHALLER).build();
   }
 
+  /**
+   * Wraps a {@link ServerMethodDefinition} such that it performs binary logging if needed.
+   */
+  final <ReqT, RespT> ServerMethodDefinition<?, ?> wrapMethodDefinition(
```
ejona86: These ReqT, RespT look unused. Use `<?, ?>` instead?

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -513,52 +512,41 @@ public void cancelled(Context context) {
...
+      ServerCallImpl<WReqT, WRespT> call = new ServerCallImpl<WReqT, WRespT>(
+          stream,
+          wMethodDef.getMethodDescriptor(),
+          headers,
+          context,
           decompressorRegistry, compressorRegistry);
-      statsTraceCtx.serverCallStarted(call);
-      ServerCall.Listener<ReqT> listener = callHandler.startCall(call, headers);
+
+      // We must not reveal the modified method descriptor to public APIs
```
ejona86: This is already done in `ServerInterceptors.wrapHandler`

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -63,7 +64,8 @@
  * <pre><code>public class TcpTransportServerFactory {
  *   public static Server newServer(Executor executor, HandlerRegistry registry,
  *       String configuration) {
- *     return new ServerImpl(executor, registry, new TcpTransportServer(configuration));
+ *     return new ServerImpl(
+ *         executor, registry, new TcpTransportServer(configuration), BinaryLogProvider.provider());
```
ejona86: Revert or fix.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -508,21 +512,41 @@ public void cancelled(Context context) {
     private <ReqT, RespT> ServerStreamListener startCall(ServerStream stream, String fullMethodName,
         ServerMethodDefinition<ReqT, RespT> methodDef, Metadata headers,
         Context.CancellableContext context, StatsTraceContext statsTraceCtx) {
-      // TODO(ejona86): should we update fullMethodName to have the canonical path of the method?
```
ejona86: Why was this comment deleted?

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
zpencer: This calls a helper function that uses it later on:
https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/ServerInterceptors.java#L243

core/src/main/java/io/grpc/internal/ServerImpl.java
zpencer: Removed stale comment

core/src/main/java/io/grpc/internal/ServerImpl.java
zpencer: This got mangled when fixing the rebase conflicts, fixed.

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
ejona86: But it doesn't seem like they are necessary here.

core/src/main/java/io/grpc/InternalServerInterceptors.java
```diff
@@ -26,6 +26,12 @@
     return ServerInterceptors.InterceptCallHandler.create(interceptor, callHandler);
   }
 
+  public static <OReqT, ORespT, WReqT, WRespT> ServerMethodDefinition<WReqT, WRespT> wrapMethod(
```
ejona86: It looks like `?` would also work here.

core/src/main/java/io/grpc/ServerInterceptors.java
```diff
@@ -230,7 +230,7 @@ private InterceptCallHandler(
     }
   }
 
-  private static <OReqT, ORespT, WReqT, WRespT> ServerMethodDefinition<WReqT, WRespT> wrapMethod(
+  static <OReqT, ORespT, WReqT, WRespT> ServerMethodDefinition<WReqT, WRespT> wrapMethod(
```
ejona86: And OReqT and ORespT here look unnecessary.

core/src/main/java/io/grpc/InternalServerInterceptors.java
zpencer: I think we can remove it from this accessor class.

core/src/main/java/io/grpc/ServerInterceptors.java
zpencer: This one looks necessary, because `wrapHandler` expects the first two argument's types to be bound, and the `?` from `getServerCallHandler` and the `?` from `getMethodDescriptor` would be treated as different even if they came from the same `ServerMethodDefinition<?,?>`.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -509,20 +511,37 @@ public void cancelled(Context context) {
...
+        handler = InternalServerInterceptors.interceptCallHandler(interceptor, handler);
       }
-      ServerCall.Listener<ReqT> listener = callHandler.startCall(call, headers);
+      ServerMethodDefinition<ReqT, RespT> interceptedDef = methodDef.withServerCallHandler(handler);
+      ServerMethodDefinition<?, ?> wMethodDef = binlogProvider == null
+          ? interceptedDef : binlogProvider.wrapMethodDefinition(interceptedDef);
+      return startWrappedCall(fullMethodName, wMethodDef, stream, headers, context);
+    }
+
+    private <WReqT, WRespT> ServerStreamListener startWrappedCall(
```
zpencer: Bazel complains unless this method is bound to differently named types:
```
core/src/main/java/io/grpc/internal/ServerImpl.java:525: error: method startCall0 in class ServerImpl.ServerTransportListenerImpl cannot be applied to given types;
      return startCall0(
             ^
  required: String,ServerMethodDefinition<ReqT#1,RespT#1>,ServerStream,Metadata,CancellableContext
  found: String,binlogProv[...]dDef),ServerStream,Metadata,CancellableContext
  reason: inference variable ReqT#1 has incompatible equality constraints CAP#1,ReqT#2
  where ReqT#1,RespT#1,ReqT#2,RespT#2 are type-variables:
    ReqT#1 extends Object declared in method <ReqT#1,RespT#1>startCall0(String,ServerMethodDefinition<ReqT#1,RespT#1>,ServerStream,Metadata,CancellableContext)
    RespT#1 extends Object declared in method <ReqT#1,RespT#1>startCall0(String,ServerMethodDefinition<ReqT#1,RespT#1>,ServerStream,Metadata,CancellableContext)
    ReqT#2 extends Object declared in method <ReqT#2,RespT#2>startCall(ServerStream,String,ServerMethodDefinition<ReqT#2,RespT#2>,Metadata,CancellableContext,StatsTraceContext)
    RespT#2 extends Object declared in method <ReqT#2,RespT#2>startCall(ServerStream,String,ServerMethodDefinition<ReqT#2,RespT#2>,Metadata,CancellableContext,StatsTraceContext)
  where CAP#1 is a fresh type-variable:
    CAP#1 extends Object from capture of ?
```

zpencer:
I think the interceptors API is actually insufficient, as binary logging requires logging the incoming data before decompression, and outgoing data post decompression. The hooks will need to happen inside the transport / framer / deframer instead.

core/src/main/java/io/grpc/internal/ServerImpl.java
ejona86: That error message... what is this? C++?

Guess it doesn't matter much. Choosing a different name isn't too annoying.

PR #3881, opened 2017-12-19
Remove com_google_protobuf_java

drigz:
This was deprecated with Bazel 0.8.0, which now uses
@com_google_protobuf instead.

This change will break users that use
grpc_java_repositories(omit_com_google_protobuf_java=True), so I've
added a custom error message to make the resolution clearer.


repositories.bzl
```diff
@@ -42,8 +42,8 @@ def grpc_java_repositories(
     com_google_instrumentation_api()
   if not omit_com_google_protobuf:
     com_google_protobuf()
-  if not omit_com_google_protobuf_java:
-    com_google_protobuf_java()
+  if omit_com_google_protobuf_java:
+    fail("omit_com_google_protobuf_java is deprecated and must be not be passed to grpc_java_repositories()")
```
ejona86: s/deprecated/no longer supported/

dapengzhang0:
@drigz What's the minimum supported version of bazel? The minimum version of bazel to build `examples` was 0.4.5 (https://github.com/grpc/grpc-java/tree/master/examples#bazel) but it will break with this PR.

drigz:
Travis failure looks like a flake. I'll try closing and reopening to trigger a rebuild.

drigz:
@dapengzhang0 Good catch, thanks!

ejona86:
It seems the community is picking up new Bazel versions pretty rapidly, so expecting 0.8.0 or later doesn't seem a deal-breaker. But this also doesn't prevent using older versions; if using an older version the user would just need to define `@com_google_protobuf_java`, which they have to do anyway for the java_proto_library support. The only "gotcha" is that `@com_google_protobuf` could be a different version than `@com_google_protobuf_java`.

PR #3882, opened 2017-12-19
 services: utility methods to create BinaryLog protos

zpencer:


services/src/main/java/io/grpc/services/BinaryLog.java
```diff
@@ -244,4 +265,101 @@ public BinaryLog getLog(String fullMethodName) {
...
+        peerType = PeerType.PEER_IPV6;
+      } else {
+        logger.log(Level.SEVERE, "unknown type of InetSocketAddress: {}", address);
+      }
+      int port = ((InetSocketAddress) address).getPort();
+      byte[] portBytes = new byte[IP_PORT_BYTES];
+      portBytes[0] = (byte) (port & IP_PORT_UPPER_MASK);
+      portBytes[1] = (byte) (port & IP_PORT_LOWER_MASK);
+      peerAddress = Bytes.concat(inetAddress.getAddress(), portBytes);
+    } else if (address.getClass().getName().equals("io.netty.channel.unix.DomainSocketAddress")) {
```
carl-mastrangelo: The toString() method seems right for all sockets that aren't ISA.

services/src/main/java/io/grpc/services/BinaryLog.java
```diff
@@ -244,4 +265,101 @@ public BinaryLog getLog(String fullMethodName) {
...
     }
   }
+
+  /**
+   * Returns a {@link Uint128} by interpreting the first 8 bytes as the high int64 and the second
+   * 8 bytes as the low int64.
+   */
+  // TODO(zpencer): verify int64 representation with other gRPC languages
+  static Uint128 callIdToProto(byte[] bytes) {
+    Preconditions.checkArgument(bytes.length == 16);
```
carl-mastrangelo: include the actual length of the number of bytes in the error message.

services/src/main/java/io/grpc/services/BinaryLog.java
```diff
@@ -244,4 +265,101 @@ public BinaryLog getLog(String fullMethodName) {
...
+      if (inetAddress instanceof Inet4Address) {
+        peerType = PeerType.PEER_IPV4;
+      } else if (inetAddress instanceof Inet6Address) {
+        peerType = PeerType.PEER_IPV6;
+      } else {
+        logger.log(Level.SEVERE, "unknown type of InetSocketAddress: {}", address);
+      }
+      int port = ((InetSocketAddress) address).getPort();
+      byte[] portBytes = new byte[IP_PORT_BYTES];
+      portBytes[0] = (byte) (port & IP_PORT_UPPER_MASK);
```
carl-mastrangelo: doesnt there need to be a `>>>` in here?

services/src/main/java/io/grpc/services/BinaryLog.java
```diff
@@ -244,4 +265,101 @@ public BinaryLog getLog(String fullMethodName) {
...
+        byte[] key = serialized[i];
+        byte[] value = serialized[i + 1];
+        if (written + key.length + value.length <= maxHeaderBytes) {
+          builder.addEntry(
+              MetadataEntry
+                  .newBuilder()
+                  .setKeyBytes(ByteString.copyFrom(key))
+                  .setValue(ByteString.copyFrom(value))
+                  .build());
+          written += key.length;
```
carl-mastrangelo: check for overflow?

services/src/main/proto/io/grpc/binarylog.proto
```diff
@@ -0,0 +1,109 @@
...
+
+option java_multiple_files = true;
+option java_package = "io.grpc.binarylog";
+option java_outer_classname = "BinaryLogProto";
+
+// Log entry we store in binary logs
+message GrpcLogEntry {
+  // Enumerates the type of logs
+  enum Type {
+    SEND_INITIAL_METADATA = 0;
```
carl-mastrangelo: always include and `UNKNOWN` type at the beginning of each enum. If this needs to be upstreamed please do so.

services/src/main/proto/io/grpc/binarylog.proto
```diff
@@ -0,0 +1,109 @@
...
+    SEND_TRAILING_METADATA = 1;
+    SEND_MESSAGE = 2;
+    RECV_INITIAL_METADATA = 3;
+    RECV_TRAILING_METADATA = 4;
+    RECV_MESSAGE = 5;
+  };
+
+  // Enumerates the entity that generates the log entry
+  enum Logger {
+    CLIENT = 0;
```
carl-mastrangelo: same here.

services/src/main/proto/io/grpc/binarylog.proto
```diff
@@ -0,0 +1,109 @@
...
+// SERVER_INIT_METADATA and TRAILING_METADATA
+// Implementations may omit some entries to honor the header limits
+// of GRPC_BINARY_LOG_CONFIG. Implementations can choose which entries are logged.
+message Metadata {
+  repeated MetadataEntry entry = 1;
+}
+
+// A metadata key value pair
+message MetadataEntry {
+  string key = 1;
```
carl-mastrangelo: keys may be binary, and my not always be utf-8.  This is true for invalid calls, which should still be possible to repesent  in logs.

services/src/test/java/io/grpc/services/BinaryLogTest.java
```diff
@@ -19,22 +19,79 @@
...
+  private static final Metadata.Key<String> KEY_C =
+      Metadata.Key.of("c", Metadata.ASCII_STRING_MARSHALLER);
+  private static final MetadataEntry ENTRY_A;
+  private static final MetadataEntry ENTRY_B;
+  private static final MetadataEntry ENTRY_C;
+  private static final Metadata metadata;
+  private static final boolean IS_COMPRESSED = true;
+  private static final boolean IS_UNCOMPRESSED = false;
+
+  static {
```
carl-mastrangelo: avoid static init blocks.   I think you can just set these to their vars directly.

services/src/test/java/io/grpc/services/BinaryLogTest.java
```diff
@@ -203,6 +260,192 @@ public void configBinLog_ignoreDuplicates_method() throws Exception {
     assertEquals(HEADER_256, factory.getLog("p.other2/m"));
   }
 
+  @Test
+  public void callIdToProto() {
+    assertEquals(
```
carl-mastrangelo: CALL_ID should be defined inside the test.  As written, I can't read the test to know if its correct.  I have to scroll elsewhere to see it's right.   For example How do I know where 0x1112131415161718L came from?

services/src/test/java/io/grpc/services/BinaryLogTest.java
```diff
@@ -203,6 +260,192 @@ public void configBinLog_ignoreDuplicates_method() throws Exception {
...
+        Uint128
+            .newBuilder()
+            .setHigh(0x1112131415161718L)
+            .setLow(0x19101a1b1c1d1e1fL)
+            .build(),
+        BinaryLog.callIdToProto(CALL_ID));
+
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
carl-mastrangelo: `expected` should probably not be used in favor of ExpectedException.  You can't check the error message with expected.

services/src/main/java/io/grpc/services/BinaryLog.java
zpencer: done

services/src/main/java/io/grpc/services/BinaryLog.java
zpencer: This is an error, fixed the unit test to catch this

services/src/main/java/io/grpc/services/BinaryLog.java
zpencer: done

services/src/test/java/io/grpc/services/BinaryLogTest.java
zpencer: done

services/src/test/java/io/grpc/services/BinaryLogTest.java
zpencer: I was intending to use CALL_ID for the next diff but I agree right now it's not buying us anything. Moved to test method body.

services/src/test/java/io/grpc/services/BinaryLogTest.java
zpencer: Done

services/src/main/proto/io/grpc/binarylog.proto
zpencer: For the proto comments, I made the changes to this branch directly and will upstream the changes.

services/src/main/java/io/grpc/services/BinaryLog.java
zpencer: `maxHeaderBytes` should take care of int overflow concerns, though it does not account for message overheads. Is there another overflow you're thinking of?

PR #3883, opened 2017-12-19
core/grpclb: resolve TXT records in DNS name resolver and include balancer addresses

carl-mastrangelo:
backport of #3852


carl-mastrangelo:
gae is a flake

PR #3886, opened 2017-12-20
core: generifiy ServiceProvider class

zpencer:
Moving all the common code among different service provider
classes to one place. If we like this, then we can make things
like `ManagedChannelProvider` implement `ServiceProvider` and
strip down their static methods and unit tests.


core/src/main/java/io/grpc/ServiceProviders.java
```diff
@@ -0,0 +1,149 @@
...
+  @VisibleForTesting
+  static <T> Iterable<T> getCandidatesViaHardCoded(
+      Class<T> klass, List<String> classNames, ClassLoader cl) {
+    // Class.forName(String) is used to remove the need for ProGuard configuration. Note that
+    // ProGuard does not detect usages of Class.forName(String, boolean, ClassLoader):
+    // https://sourceforge.net/p/proguard/bugs/418/
+    List<T> list = new ArrayList<T>();
+    for (String name : classNames) {
+      try {
+        list.add(create(klass, cl.loadClass(name)));
```
ejona86: This doesn't work with ProGuard. It needs to be the exact literals like `Class.forName("io.grpc.okhttp.OkHttpChannelProvider")` we're using today, otherwise ProGuard won't detect the string and won't keep needed classes and won't rename properly. This is also why the hard-coded code doesn't use a `ClassLoader` (which is what that comment above is about).

core/src/main/java/io/grpc/ServiceProvider.java
```diff
@@ -0,0 +1,40 @@
...
+ * or {@code interface}.
+ * The {@link ServiceProviders} utility class should be used to create concrete implementations
+ * of the class.
+ */
+interface ServiceProvider {
+  /**
+   * Whether this provider is available for use, taking the current environment into consideration.
+   * If {@code false}, no other methods are safe to be called.
+   */
+  boolean isAvailable();
```
ejona86: I'd really hope these would be `protected`. That does require making this an abstract class though. I'm not sure how the docs work look in either case...

core/src/main/java/io/grpc/ServiceProviders.java
```diff
@@ -0,0 +1,149 @@
...
+    Iterator<T> iter = candidates.iterator();
+    while (iter.hasNext()) {
+      try {
+        T current = iter.next();
+        if (!current.isAvailable()) {
+          continue;
+        }
+        list.add(current);
+      } catch (Throwable t) {
+        // The iterator from ServiceLoader may throw ServiceConfigurationError
```
ejona86: This was purposefully being propagated before. I don't think we can safely continue. Different implementations have different behavior, so continuing when the user expected a certain implementation seems dangerous.

Note that this catch is also incomplete, because `hasNext()` can throw.

core/src/main/java/io/grpc/ServiceProvider.java
zpencer: Not 100% sure what you're referring to, but I tried an experiment where I created a `public abstract class NettyFoo extends InternalServiceProvider` and a `public class NettyFooImpl extends NettyFoo`. In both cases, `isAvailable` and `priority` show up in the output of `./gradlew javadoc`.

core/src/main/java/io/grpc/ServiceProviders.java
zpencer: This makes sense, I'll change it so that we just give up if any classes fail.

core/src/main/java/io/grpc/ServiceProviders.java
zpencer: Changed back to old behavior.

core/src/main/java/io/grpc/ServiceProvider.java
ejona86: Check in the output of all/. It will filter the InternalServiceProvider

core/src/main/java/io/grpc/ServiceProviders.java
```diff
@@ -0,0 +1,155 @@
...
+
+      // Sort descending based on priority.
+      Collections.sort(list, Collections.reverseOrder(new Comparator<ServiceProvider>() {
+        @Override
+        public int compare(ServiceProvider f1, ServiceProvider f2) {
+          return f1.priority() - f2.priority();
+        }
+      }));
+      return Collections.unmodifiableList(list);
+    } catch (Throwable t) {
```
ejona86: Can we please not have unnecessary behavior changes in a refactoring change like this? This can be done easily as a follow-up and avoids it being lost in the noise and conflated in review.

zpencer: Updated PR with this error handling code removed

zpencer:
While working on the usage commit, I realized that `NameResolverProvider extends NameResolver.Factory`, so it can not extend another abstract class. I ended up changing the interface so that classes using ServiceProviders will implement a:
```
public interface PriorityAccessor<T> {
    boolean isAvailable(T provider);
    int getPriority(T provider);
  }
```

This should also avoid the javadoc problem.

core/src/main/java/io/grpc/InternalServiceProviders.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+package io.grpc;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.util.List;
+
+@Internal
+public final class InternalServiceProviders {
+  /**
```
carl-mastrangelo: needs a private ctor

zpencer:
This is an example commit using this utility: https://github.com/zpencer/grpc-java/compare/serviceproviders-introduce...zpencer:serviceproviders-migrate-nameresolverprovider

zpencer:
@ejona86 ping

core/src/main/java/io/grpc/InternalServiceProviders.java
```diff
@@ -0,0 +1,74 @@
...
+
+@Internal
+public final class InternalServiceProviders {
+  private InternalServiceProviders() {
+  }
+
+  /**
+   * Accessor for method.
+   */
+  public static <T> T load(
```
ejona86: load, loadAll, isAndroid, and PriorityAccessor don't seem like they need to be exposed.

I'd sort of prefer this remained an implementation detail and so we don't expose getCandidatesViaServiceLoader and getCandidatesViaHardCoded as well (putting those methods on each provider instead), but I can live with this.

PR #3889, opened 2017-12-20
Make ManagedChannel honor Service config LB

carl-mastrangelo:
I *think* the idle state transitions are correct.  I have looked at them and tried tracing a few paths through.  This doesn't go full idle because the name resolver doesn't need to be restarted.  Also, the LB transition happens inside of a NR callback, so it would be odd to have the NR terminate itself upon successful resolution.  (Might this cause recursion?  I think it may).

### DIFF
--- a/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
+++ b/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright 2018, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.internal;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.gson.JsonObject;
+import io.grpc.Attributes;
+import io.grpc.ConnectivityState;
+import io.grpc.ConnectivityStateInfo;
+import io.grpc.EquivalentAddressGroup;
+import io.grpc.LoadBalancer;
+import io.grpc.LoadBalancer.Helper;
+import io.grpc.LoadBalancer.PickResult;
+import io.grpc.LoadBalancer.PickSubchannelArgs;
+import io.grpc.LoadBalancer.SubchannelPicker;
+import io.grpc.PickFirstBalancerFactory;
+import io.grpc.Status;
+import java.lang.reflect.Method;
+import java.util.List;
+import java.util.Locale;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
+
+final class AutoConfiguredLoadBalancerFactory extends LoadBalancer.Factory {
+  private static final Logger logger =
+      Logger.getLogger(AutoConfiguredLoadBalancerFactory.class.getName());
+
+  @VisibleForTesting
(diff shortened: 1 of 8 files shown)


zhangkun83:
This balancer selection logic is already implemented in GrpclbLoadBalancer, although with a different input format (`GrpclbConstants.ATTR_LB_POLICY` etc). It delegates to pick-first or round-robin in cases where the spec says pick-first or round-robin should be used. The internal EnvelopeNameResolver produces the format that GrpclbLoadBalancer recognizes, and our internal channel builder defaults to GrpclbLoadBalancer.

Granted we can't default to GrpclbLoadBalancer in OSS because is not in core. I still believe a delegating balancer is a better approach than implementing this directly in ManagedChannelImpl, because ManagedChannelImpl is already very complicated, and I don't see any benefit of adding this to ManagedChannelImpl.

My proposal is that we move the delegating logic from GrpclbLoadBalancer to, e.g., "AutoConfiguredLoadBalancer", and use it as the default balancer. Unify its input format so that it works with both DNS and Envelope. No changes needed in ManagedChannelImpl. The users are still able to specify their own balancers, which will override the default balancer.

carl-mastrangelo:
Done, PTAL.   I'd like to update the gRPCLB delegation behavior in a followup PR

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -0,0 +1,176 @@
...
+    //  Must be run inside ChannelExecutor.
+    @Override
+    public void handleResolvedAddressGroups(
+        List<EquivalentAddressGroup> servers, Attributes attributes) {
+      if (attributes.keys().contains(GrpcAttributes.NAME_RESOLVER_ATTR_SERVICE_CONFIG)) {
+        Factory newlbf = decideLoadBalancerFactory(
+            servers, attributes.get(GrpcAttributes.NAME_RESOLVER_ATTR_SERVICE_CONFIG));
+        if (newlbf != null && newlbf != delegateFactory) {
+          delegate.shutdown();
+          helper.updateBalancingState(ConnectivityState.CONNECTING, new EmptySubchannelPicker());
```
zhangkun83: I think updateBalancingState() should happen before delegate.shutdown().

delegate.shutdown() will shutdown all subchannels. RPCs started between shutdown() and updateBalancingState() will be assigned to shutdown'ed subchannels and will fail.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -0,0 +1,176 @@
...
+          ClassNotFoundException caught = null;
+          try {
+            Class<?> lbFactoryClass;
+            lbFactoryClass = Class.forName(ROUND_ROUND_LOAD_BALANCER_FACTORY_NAME);
+            Method getInstance = lbFactoryClass.getMethod("getInstance");
+            return (LoadBalancer.Factory) getInstance.invoke(null);
+          } catch (RuntimeException e) {
+            throw e;
+          } catch (Exception e) {
+            throw new RuntimeException("Can't get gRPC LB, but balancer addresses were present", e);
```
zhangkun83: The message is incorrect.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -0,0 +1,176 @@
...
+      if (haveBalancerAddress) {
+        try {
+          Class<?> lbFactoryClass;
+          lbFactoryClass = Class.forName(GRPCLB_LOAD_BALANCER_FACTORY_NAME);
+          Method getInstance = lbFactoryClass.getMethod("getInstance");
+          return (LoadBalancer.Factory) getInstance.invoke(null);
+        } catch (RuntimeException e) {
+          throw e;
+        } catch (Exception e) {
+          throw new RuntimeException("Can't get gRPC LB, but balancer addresses were present", e);
```
zhangkun83: We use GRPCLB or grpclb as a special term. It should not be split.

core/src/test/java/io/grpc/internal/AutoConfiguredLoadBalancerFactoryTest.java
```diff
@@ -0,0 +1,188 @@
+/*
+ * Copyright 2018, gRPC Authors All rights reserved.
```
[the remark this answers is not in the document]
carl-mastrangelo: done

Issue #2132, opened 2016-08-04
Filter mechanism for server transport creation and deletion


zhangkun83:
To support a per-connection throttling mechanism used inside google, gRPC needs to allow user to add filters to the server that:
1. Gets called when a server transport is created and handshake-completed, and has access to transport-specific attributes such as client address and identity.
2. Adds per-transport attributes to ServerCall attributes.
3. Gets called when a server transport is terminated.

I discussed a few options with @ejona86:
#### ServerInterceptor

The filter returns a `ServerInterceptor` when transport is created. The interceptor is transport-scoped, unlike ordinary interceptors that are service-scoped.

``` java
interface ServerTransportFilter {
  ServerInterceptor transportCreated(SomePublicServerTransportType transport);
  void transportTerminated(SomePublicServerTransportType transport);
}
```

Pros:
1. Doesn't touch transport API (ServerTransport, ServerStream) which are internal.
2. Re-uses `ServerInterceptor` for purposes that `ServerInterceptor` is designed for (modifying ServerCall attributes etc).

Cons:
1. Changes the scope of `ServerInterceptor`. @ejona86 thinks bonding `ServerInterceptor` to a transport is seemingly generic but actually an ad-hoc solution for a particular problem. I don't necessarily agree though.
#### Decorating transport

The filter returns a decorated `ServerListener` at registration. Through a few more layers of wrapping, the filter would be able to decorate `ServerListener.transportCreated(),`ServerTransportListener.transportTerminated() and `ServerStream.attributes()`.

Pros:
1. Minimal API change

Cons:
1. Heavily coupled with transport API which is internal.
2. Duplicates the existing interceptor mechanism, but on `ServerStream` instead of `ServerCall`. It would be ideal to just decorate `ServerCall`, which is what `ServerInterceptor` does. `ServerInterceptor` currently is per-service, but our use case need it to be per-connection. 
3. Even more layers of decoration than the interceptors. The filter can be annoying to write.
#### Ad-hoc

Just for our original requirements, the filter would return per-transport attributes that are to be merged into ServerCall attributes.

``` java
interface ServerTransportFilter {
  Attributes transportCreated(SomePublicServerTransportType transport);
  void transportTerminated(SomePublicServerTransportType transport);
}
```

It probably won't work in this form, because the attributes such as peer security identity may not be available immediately after the transport is created, if the security handshake has not completed yet. On the other hand, the first two options can work around this situation by reading the attributes only when the first stream is created.

[...]

PR #3892, opened 2017-12-25
introduce addTransportFilter to NettyChannel

xiaoshuang-lu:


xiaoshuang-lu:
I will offer another patch to enable this feature of okhttp later. :-)

xiaoshuang-lu:
Hi @dapengzhang0, what about this patch?

dapengzhang0:
Thanks for the PR. What is the motivation of adding this feature? In my opinion, giving users the power of filtering transport attributes (which is pretty a grpc internal thing) might make the library fragile.

xiaoshuang-lu:
@dapengzhang0, thanks for your comments. Appears grpc server offers the ability to filter transport attributes, so I port this feature to client. In fact, with transport filter functionality, users can insert their business logic to grpc framework, e.g. connection statistics.

dapengzhang0:
cc @zhangkun83 who might be familiar about `ServerTransportFilter`

dapengzhang0:
@xiaoshuang-lu can you elaborate more on usecase of connection statistics, and why will it be preferable at `Transport` level rather than [`LoadBalancer.Subchannel`](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/LoadBalancer.java#L533) level? If it is a useful/reasonable usecase, we will consider adding your API. Thanks.

xiaoshuang-lu:
https://github.com/grpc/grpc-java/issues/2132

zhangkun83:
@xiaoshuang-lu, before we add something we must be sure that it will be useful. Do you have an actual use case?

xiaoshuang-lu:
@zhangkun83, thanks for your comments.

1. I wanna count connections for clients indeed.
2. It would be beautiful if both server and client have addTransportFilter. Only server has this interface seems a little bit weird. :-)

zhangkun83:
I'd like to understand your use case better. Please elaborate whether you want to count currently active connections or connections created so far, and most importantly, what you are going to use this count for.

Client-server API symmetry is a non-goal. In fact, client and server have fundamental differences, so in a lot of cases it doesn't make sense to have the same API. Server needs TransportFilter because server doesn't create transports, and it needs a mechanism to know when transports are created. On the client-side, `LoadBalancer` initiates the creation of Subchannels, which in turn creates transports -- if you implemented your own `LoadBalancer`, you could listen to `handleSubchannelState()` and get an idea of how many transports are created and how many are currently alive.

xiaoshuang-lu:
Hi @zhangkun83,

> Please elaborate whether you want to count currently active connections or connections created so far

TransportFilter can handle both scenarios: 1. increase counter in `transportReady` and decrease counter in `transportTerminated,` 2. only accumulate in `transportReady`

> what you are going to use this count for

do connection statistics and report data to some metric systems :-)

> if you implemented your own LoadBalancer, you could listen to handleSubchannelState() and get an idea of how many transports are created and how many are currently alive.

`handleSubchannelState()` of `LoadBalancer` may be an alternative way to count connections. I am afraid that this solution seems unclear, inconvenient, and slightly ugly.

ejona86:
`ServerTransportFilter` is one of those things we did because we thought we required it, and then haven't really used it. Yes, we think we still may use it, but...

I'd rather open something like this up as a stats API. Something like a one-level up version of ServerStreamTracer.Factory, which works on transports instead of streams. On client-side we're using an interceptor today, but we aren't really using it for much. We could probably combine this theoretical transport-level stats API with configuring streams. We could maybe even remove the current ClientStreamTracer CallOption, which would restrict the ability to configure it and would encourage using it for its intended purpose.

(@zhangkun83, I know that last statement sort of conflicts with another statement I've said about relying on it being an interceptor more heavily; we'll have to sort things out.)

Part of my problem is the entire "filter" aspect of `ServerTransportFilter` remains awkward. While it tries to be generic, filtering has really only been useful for one use-case which we thought forced to begin with: initializing a per-transport object and making it available to each call. In https://github.com/grpc/grpc-java/issues/2132#issuecomment-279894851, @rmichela mentioned it felt weird as well. Maybe we could do some other approach there, like some identifier+WeakHashMap.

A stats API would be "look but don't touch," would fit with the other "be careful; you're on the network thread" stats APIs, and align to what people appear to be using it for: stats.

Although I do feel like I could be convinced another way easily enough.

rmichela:
The only use I've found for the current `ServerTransportFilter` API is to attach a unique identity to each server transport session, allowing gRPC services to become "session aware".

xiaoshuang-lu:
@rmichela I leveraged `ServerTransportFilter` to count connections... $facepalm$

ejona86:
It seems `ServerTransportFilter` solves the use-case involved here: connection statistics. It doesn't seem there's a use-case to actually filter the attributes. Closing.

PR #3894, opened 2017-12-27
core: fix unimplemented API in ManageChannel

dapengzhang0:
Two APIs were added to ManageChannel. This would be API breaking if they are still abstract.

### DIFF
--- a/core/src/main/java/io/grpc/ManagedChannel.java
+++ b/core/src/main/java/io/grpc/ManagedChannel.java
@@ -16,6 +16,7 @@
 
 package io.grpc;
 
+import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/grpc/ManagedChannel.java
```diff
@@ -115,4 +123,10 @@ public void notifyWhenStateChanged(ConnectivityState source, Runnable callback)
    */
   @ExperimentalApi
   public void resetConnectBackoff() {}
+
+  @Internal
+  @Override
+  public InternalLogId getLogId() {
+    return InternalLogId.allocate(getClass().getName());
```
zhangkun83: The logId should be a final field.

Also, there doesn't seem to be a case where a subclass needs to override this method. Can we make this method final? All current overrides could be simply removed.

dapengzhang0: Done.

PR #3896, opened 2017-12-27
core,netty,okhttp: Collect stats on subchannels

zpencer:
Today, the ChannelTracer inside ClientCall is the subchannel's
tracer.

This PR makes it so that the ClientCall ChannelTracer is the one
from the ManagedChannelImpl. The subchannels' ChannelTracers are
linked to the TransportTracer. The result is that as soon as the
ClientCall starts or finishes, the ManagedChannel stats are
updated. When the corresponding event takes place on the
transport, then the subchannel stats are updated.

The "update test class method signatures" contains mostly mechanical
changes.


zpencer:
We should not wait until the transport sees the event before reporting it to the subchannel.

PR #3897, opened 2017-12-27
[DONOT merge] core: remove circular dependency on Internal

dapengzhang0:


dapengzhang0:
Closed as @zpencer will send a PR to fix it.

Issue #3688, opened 2017-11-08
Support smooth network changes


bubenheimer:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?
1.7.0

### What JVM are you using (`java -version`)?
1.8

### What did you do?
On Android, create channel with non-streaming and streaming calls. Change data network from cellular to WiFi. (standard mobile use case)

### What did you expect to see?
New calls are started over WiFi. Existing calls receive an advisory notification, enabling them to finish up while the cellular network is still available and smoothly start a new call over WiFi, possibly concurrently to the existing call.

### What did you see instead?
All call activity continues on cellular data network until the OS yanks it (30 seconds on Nexus 6, Nougat). Communication is disrupted until channel transport is reestablished over WiFi.

bubenheimer:
I imagine that the brunt of this can be addressed via a mobile-specific load balancer with subchannels for cellular & WiFi. As this is a standard mobile use case, it would seem appropriate for the grpc-java framework to provide a standard load balancer with a hook for the application to indicate a network change. Or maybe it's trivial enough to custom-implement in the application, based on a standard example? I have not worked with the grpc-java load balancing mechanism yet.

The advisory notification would be triggered at the application level, and could be propagated at that level, too, but it somewhat feels like a general mechanism that the grpc framework might provide; not sure if there is something for this already.

ejona86:
This can be done in the ManagedChannel; no need to involve the load balancer. Although there will be a race where RPCs may be assigned to a shutdown transport, as mentioned in #2562. The workaround we did for LBs was to delay the shutdown for a fixed amount of time, hiding the race. That wouldn't work here.

ericgribkoff:
Spoke with @ejona86 about this. We can offer some kind of `#connectToNewNetwork()` (or `#networkDropped()`?) method on `ManagedChannel` that will invoke `#shutdown()` on the channel's existing client transports. This will allow existing RPCs to continue on the cell connection, but new RPCs will trigger creation of a new transport which will connect over wifi.

Re: an advisory notification that existing calls are going to be shut down - gRPC as a whole lacks this type of mechanism, e.g., we also cannot notify applications when we know the server is shutting down.

ejona86:
> gRPC as a whole lacks this type of mechanism, e.g., we also cannot notify applications when we know the server is shutting down.

We don't have it today, but it is possible for us to add. It's probably a separate feature request; I've created #3689 for it.

[...]

PR #3904, opened 2017-12-28
core: add shutdownTransports() method to ManagedChannel

ericgribkoff:
This is to address https://github.com/grpc/grpc-java/issues/3688

### DIFF
--- a/core/src/main/java/io/grpc/ManagedChannel.java
+++ b/core/src/main/java/io/grpc/ManagedChannel.java
@@ -127,7 +127,7 @@ public abstract class ManagedChannel extends Channel {
    *
    * <p>No-op if not supported by implementation.
    *
-   * @since 1.10.0
+   * @since 1.11.0
    */
   @ExperimentalApi
   public void shutdownTransports() {}
--- a/core/src/main/java/io/grpc/ManagedChannel.java
+++ b/core/src/main/java/io/grpc/ManagedChannel.java
@@ -112,12 +112,13 @@ public abstract class ManagedChannel extends Channel {
    *
    * @since 1.8.0
    */
-  @ExperimentalApi
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/4056")
   public void resetConnectBackoff() {}
 
   /**
-   * Shutdown existing transports while keeping the channel active. This allows on-going RPCs on the
-   * channel to continue while new RPCs will trigger creation of a new transport.
+   * Invoking this method moves the channel into the IDLE state and triggers tear-down of the
+   * channel's name resolver and load balancer, while still allowing on-going RPCs on the channel to
+   * continue. New RPCs on the channel will trigger creation of a new connection.
    *
    * <p>This is primarily intended for Android users when a device is transitioning from a cellular
    * to a wifi connection. Initially the device will maintain both the cellular and wifi
@@ -129,6 +130,6 @@ public abstract class ManagedChannel extends Channel {
    *
    * @since 1.11.0
    */
-  @ExperimentalApi
-  public void shutdownTransports() {}
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/4056")
+  public void prepareToLoseNetwork() {}
 }
--- a/core/src/main/java/io/grpc/ManagedChannel.java
+++ b/core/src/main/java/io/grpc/ManagedChannel.java
@@ -114,4 +114,21 @@ public abstract class ManagedChannel extends Channel {
    */
   @ExperimentalApi
   public void resetConnectBackoff() {}
+
+  /**
+   * Shutdown existing transports while keeping the channel active. This allows on-going RPCs on the
+   * channel to continue while new RPCs will trigger creation of a new transport.
+   *
+   * <p>This is primarily intended for Android users when a device is transitioning from a cellular
+   * to a wifi connection. Initially the device will maintain both the cellular and wifi
+   * connections, but the OS also issues a notification that after a short time the cellular
+   * connection will be terminated. Apps may invoke this method to ensure that new RPCs are created
+   * using the wifi connection, rather than the soon-to-be-disconnected cellular network.
+   *
+   * <p>No-op if not supported by implementation.
+   *
+   * @since 1.10.0
+   */
+  @ExperimentalApi
+  public void shutdownTransports() {}
 }
--- a/core/src/main/java/io/grpc/internal/InternalSubchannel.java
+++ b/core/src/main/java/io/grpc/internal/InternalSubchannel.java
@@ -292,35 +292,6 @@ final class InternalSubchannel implements InternalWithLogId {
     }
   }
 
-  void shutdownTransports(Status reason) {
-    ManagedClientTransport savedActiveTransport;
-    ConnectionClientTransport savedPendingTransport;
-    try {
-      synchronized (lock) {
-        if (state.getState() == SHUTDOWN) {
-          return;
-        }
-        savedActiveTransport = activeTransport;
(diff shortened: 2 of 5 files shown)


core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -286,6 +286,28 @@ void resetConnectBackoff() {
     }
   }
 
+  void shutdownTransports(Status reason) {
```
dapengzhang0: Using this new method may break the contract of `ManagedClientTransport` interface:
```java
  /**
   * ... ...  This method may only be called once.
   */
  void shutdown(Status reason);
```
Why not just use the existing `InternalSubchannel.shutdown()` method?

ericgribkoff: The current API assumes that `InternalSubchannel.shutdown()` only occurs when the LB has discarded the subchannel or the managed channel itself is shutting down. See the Javadoc on `io.grpc.LoadBalancer#handleSubchannelState`. `InternalSubchannel.shutdown()` is purposefully ignored by the LB, so just invoking this method means that outgoing RPCs will still be sent to the shutdown subchannel.

We can change this behavior in the LB/somewhere else to allow just shutting down the `InternalSubchannel` directly and recovering properly, but this is a more extensive change than just shutting down the transports as done here. Thoughts?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -721,6 +725,27 @@ public void run() {
...
+  @Override
+  public void shutdownTransports() {
+    channelExecutor
+        .executeLater(
+            new Runnable() {
+              @Override
+              public void run() {
+                if (shutdown.get()) {
+                  return;
+                }
```
dapengzhang0: Add
```java
if (nameResolverStarted) {
  nameResolver.refresh();
}
```

? Otherwise the LB will still route new RPCs to the same **EAG** of the shutdown InternalSubchannel, which might have been resolved by the DNS for cellular nextwork.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
dapengzhang0: One option may be: keep your code and add 
```java
activeTransport = null;
pendingTransport = null;
addressIndex = 0;
```
 
after
```java
savedActiveTransport = activeTransport;
savedPendingTransport = pendingTransport;
```

Another option: call `InternalSubchannel.shutdown()` and  change the behavior of LB so that is does not ignore it completely (LB ignored it because of the assumptions you mentioned which were the only two cases we knew at that time, this is never a hard requirement, now that we have a new case, maybe it should be changed.)

ericgribkoff: Went with option 1. When the `InternalSubchannel` itself is shutting down, we can ignore the resulting `TransportListener.transportShutdown` callback, but here we still want to handle the callback (as if the transport itself caused the shutdown), so saved an additional reference to the shutdown channels for use in the callback.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -721,6 +725,27 @@ public void run() {
         }).drain();
   }
 
+  @Override
+  public void shutdownTransports() {
+    channelExecutor
+        .executeLater(
+            new Runnable() {
```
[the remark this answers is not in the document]

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -286,6 +292,34 @@ void resetConnectBackoff() {
...
+    ManagedClientTransport savedActiveTransport;
+    ConnectionClientTransport savedPendingTransport;
+    try {
+      synchronized (lock) {
+        if (state.getState() == SHUTDOWN) {
+          return;
+        }
+        savedActiveTransport = activeTransport;
+        savedPendingTransport = pendingTransport;
```
dapengzhang0: (The logic of `InternalSubchannel` is pretty involved, so I'm not 100% sure, but) It seems that:
If current state is `CONNECTING`, we should call `gotoNonErrorState(IDLE)` here. Otherwise `obtainActiveTransport()` will return null all the time and never start a new transport or return an active transport.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -504,14 +538,16 @@ public void transportShutdown(Status s) {
...
             return;
           }
-          if (activeTransport == transport) {
+          if (activeTransport == transport || activeTransportToShutdown == transport) {
             gotoNonErrorState(IDLE);
             activeTransport = null;
+            activeTransportToShutdown = null;
             addressIndex = 0;
-          } else if (pendingTransport == transport) {
+          } else if (pendingTransport == transport || pendingTransportToShutdown == transport) {
```
dapengzhang0: Keeping the reference `pendingTransportToShutdown` is a nice idea!

core/src/main/java/io/grpc/internal/InternalSubchannel.java
dapengzhang0: I take it back. `startNewTransport()` will be called in `TransportListener.transportShutdown()` of the pending transport. So there is no problem.

core/src/main/java/io/grpc/ManagedChannel.java
```diff
@@ -114,4 +114,21 @@ public void notifyWhenStateChanged(ConnectivityState source, Runnable callback)
...
+   *
+   * <p>This is primarily intended for Android users when a device is transitioning from a cellular
+   * to a wifi connection. Initially the device will maintain both the cellular and wifi
+   * connections, but the OS also issues a notification that after a short time the cellular
+   * connection will be terminated. Apps may invoke this method to ensure that new RPCs are created
+   * using the wifi connection, rather than the soon-to-be-disconnected cellular network.
+   *
+   * <p>No-op if not supported by implementation.
+   *
+   * @since 1.10.0
```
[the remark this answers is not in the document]
ericgribkoff: Done

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -286,6 +292,34 @@ void resetConnectBackoff() {
...
+    ManagedClientTransport savedActiveTransport;
+    ConnectionClientTransport savedPendingTransport;
+    try {
+      synchronized (lock) {
+        if (state.getState() == SHUTDOWN) {
+          return;
+        }
+        savedActiveTransport = activeTransport;
+        savedPendingTransport = pendingTransport;
+        // Save a reference to the transports for the TransportListener.transportShutdown() callback
```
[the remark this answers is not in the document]
ericgribkoff: Added `cancelReconnectTask()`.

Re: IDLE mode: I'm not sure. The javadoc on `ConnectivityState.IDLE` (i.e.,  https://github.com/grpc/grpc/blob/master/doc/connectivity-semantics-and-api.md and its transition diagram) implied to me that the channel state should only IDLE there are no ongoing RPCs. Here we are allowing pending RPCs to continue, so it seemed to me like we should wait for the underlying transports to report shutdown before moving into IDLE.

ericgribkoff:
> Would it be better to just enter idle mode?

@ejona86 Done. PTAL

(new method name `prepareToLoseNetwork()` modeled after Android's `onLosing` network [callback](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback.html) ,which is when apps would want to invoke this method)

PR #3912, opened 2018-01-04
core,testing: deprecate passing ServerCall to StatsTraceContext

zpencer:
Instead, pass a ServerCallInfo object containing the interesting bits
of info. This lets us modify the call handler for binary logging, but
still provide the original info to the StatsTraceContext API.
 
Updated unit tests and existing code to use new API
Modified existing tests that were overspecified, and failed because of new but harmless mock interactions.


zpencer:
This PR is required for https://github.com/grpc/grpc-java/pull/3879

zpencer:
@zhangkun83 PTAL

core/src/main/java/io/grpc/ServerStreamTracer.java
```diff
@@ -54,4 +69,116 @@ public void serverCallStarted(ServerCall<?, ?> call) {
...
+    }
+
+    public String getAuthority() {
+      return authority;
+    }
+
+    public boolean isCancelled() {
+      return isCancelled;
+    }
+
```
zhangkun83: ServerCallInfo is passed only to `serverCallStarted()`. `isReady()` and `isCancelled()` don't seem to be useful at that point. We should probably remove them.

zpencer:
@zhangkun83 ping

The GAE failure is due to an unrelated issue https://github.com/grpc/grpc-java/pull/4000

core/src/main/java/io/grpc/ServerStreamTracer.java
```diff
@@ -54,4 +70,103 @@ public void serverCallStarted(ServerCall<?, ?> call) {
...
+  /**
+   * A data class with info about the started {@link ServerCall}.
+   */
+  public static final class ServerCallInfo<ReqT, RespT> {
+    private final MethodDescriptor<ReqT, RespT> methodDescriptor;
+    private final Attributes attributes;
+    private final String authority;
+
+    ServerCallInfo(
```
zhangkun83: Hmm... if you lock down the constructor, how do people outside of gRPC unit-test their own ServerStreamTracers?

Maybe this should be an abstract class, and put the impl in `io/grpc/internal`.

core/src/main/java/io/grpc/internal/ServerCallInfoImpl.java
```diff
@@ -0,0 +1,73 @@
...
+import com.google.common.base.Objects;
+import io.grpc.Attributes;
+import io.grpc.MethodDescriptor;
+import io.grpc.ServerStreamTracer.ServerCallInfo;
+import javax.annotation.Nullable;
+
+/**
+ * An implementation of {@link ServerCallInfo}.
+ */
+public final class ServerCallInfoImpl<ReqT, RespT> extends ServerCallInfo<ReqT, RespT> {
```
zhangkun83: This can be package-private.

Issue #368, opened 2015-05-01
Don't use SerializingExecutor when running with a direct executor.


buchgr:
I wanted to know what the impact of the `SerializingExecutor` is when running with a direct executor.

So I did 3 benchmarks, choosing the best out of 3 runs.

##### Direct Executor + Serializing (current master)
```
buchgr@buchgr0:~/Code/grpc-java/benchmarks/build/install/grpc-benchmarks/bin$ ./qps_client --port=33333 --host=localhost --channels=8 --outstanding_rpcs_per_channel=10 --warmup_duration=10s --duration=30s --server_payload=1 --client_payload=1 --directexecutor
Channels:                       8
Outstanding RPCs per Channel:   10
Server Payload Size:            1
Client Payload Size:            1
50%ile Latency (in micros):     669
90%ile Latency (in micros):     1817
95%ile Latency (in micros):     3231
99%ile Latency (in micros):     5727
99.9%ile Latency (in micros):   9255
QPS:                            78642
```

##### Direct Executor + Serializing Executor without synchronized blocks.
```
buchgr@buchgr0:~/Code/grpc-java/benchmarks/build/install/grpc-benchmarks/bin$ ./qps_client --port=33333 --host=localhost --channels=8 --outstanding_rpcs_per_channel=10 --warmup_duration=10s --duration=30s --server_payload=1 --client_payload=1 --directexecutor
Channels:                       8
Outstanding RPCs per Channel:   10
Server Payload Size:            1
Client Payload Size:            1
50%ile Latency (in micros):     655
90%ile Latency (in micros):     1647
95%ile Latency (in micros):     3083
99%ile Latency (in micros):     5679
99.9%ile Latency (in micros):   9607
QPS:                            81500
```

##### Direct Executor only, no Serializing Executor
```
buchgr@buchgr0:~/Code/grpc-java/benchmarks/build/install/grpc-benchmarks/bin$ ./qps_client --port=33333 --host=localhost --channels=8 --outstanding_rpcs_per_channel=10 --warmup_duration=10s --duration=30s --server_payload=1 --client_payload=1 --directexecutor
Channels:                       8
Outstanding RPCs per Channel:   10
Server Payload Size:            1
Client Payload Size:            1
50%ile Latency (in micros):     619
90%ile Latency (in micros):     1096
95%ile Latency (in micros):     1132
99%ile Latency (in micros):     3999
99.9%ile Latency (in micros):   11407
QPS:                            99904
```

[...]

PR #3914, opened 2018-01-04
core: Don't re-wrap SerializingExecutors in ClientCallImpl and ServerImpl

njhill:
If the user provides a `SerializingExecutor`, there's an unnecessary performance overhead in wrapping it in another `SerializingExecutor` (see older related PR https://github.com/grpc/grpc-java/issues/368).

This is useful in particular for cases where users manage their own `SerializingExecutor`-based event loops and want RPC callbacks to run in the same context.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ClientCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ClientCallImpl.java
@@ -95,7 +95,8 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
     // See https://github.com/grpc/grpc-java/issues/368
     this.callExecutor = executor == directExecutor()
         ? new SerializeReentrantCallsDirectExecutor()
-        : new SerializingExecutor(executor);
+        : (executor instanceof SerializingExecutor
+          ? executor : new SerializingExecutor(executor));
     this.channelTracer = channelTracer;
     // Propagate the context from the thread which initiated the call to all callbacks.
     this.context = Context.current();
--- a/core/src/main/java/io/grpc/internal/ServerImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerImpl.java
@@ -421,6 +421,8 @@ public final class ServerImpl extends io.grpc.Server implements InternalWithLogI
       // that comes with SerializingExecutor.
       if (executor == directExecutor()) {
         wrappedExecutor = new SerializeReentrantCallsDirectExecutor();
+      } else if (executor instanceof SerializingExecutor) {
+        wrappedExecutor = executor;
       } else {
         wrappedExecutor = new SerializingExecutor(executor);
       }


carl-mastrangelo:
@njhill `SerializingExecutor` is not a public API and no users should be using it directly.  It an make assumptions about runnables that others cannot make.

njhill:
Thanks @carl-mastrangelo, and apologies for the misunderstanding. I do think such a class has general utility - presumably your statement applies just to this implementation and in particular the case that `executor.execute()` throws (where there are assumptions about single-producer and Runnable object uniqueness)?

I was really hoping to avoid the wrapping for executors that already have the required single-threaded characteristics (`Executors.newSingleThreadExecutor()` would be another example). I couldn't think of another way that could be done "safely" though. The only thing I can think of is having another public "util" serializing executor which doesn't make those same assumptions and which callers can then use for this purpose. Any thoughts/suggestions on this would be most welcome!

As a side note - while I was just looking at the newer implementation of `SerializingExecutor` I noticed a couple of additional simple optimizations. It's very possible that I overlooked something again so I won't open a PR but [here is a commit](https://github.com/grpc/grpc-java/commit/f6705e75f43cc470c28994c2c0d67e4aeeb553ff) with them. If you think they make sense I can open another PR.

carl-mastrangelo:
The class does have general utility.   Guava has recently made a [Sequential Executor](https://github.com/google/guava/blob/v23.6/guava/src/com/google/common/util/concurrent/MoreExecutors.java#L498)   for this purpose.   Theirs is slightly more general than the SE that gRPC uses.


Concerning `Executors.newSingleThreadExecutor()` : it isn't actually serializing.  The problem is reentrancy.  Often times one Runnable to try to execute/schedule more.  SE is careful to finish the current thing it is working on before trying to work on something else.   For example, if your stub started a new RPC in the `onNext` callback of an existing RPC, you could have two runnables running at the same time; one inside the other.   There's only one thread, but two things.   SE avoids this problem.

Lastly: your optimization of extending CLQ is good, but it doesn't have a measureable impact, even under heavy load.  I have included it in previous (but rejected) PRs.    For gRPC pretty much every optimization needs to be measurable in order to be accepted.  If you look at many of my commits from 2016, I included microbenchmark numbers to backup the changes.  

This is an annoying bar to set, but it really needs to be there.  For example, your proposed optimization makes the `execute()` method much harder to inline, and does an extra atomic compare-and-swap in the inner loop.  Without benchmarks, how do we know if this is an acceptable cost?

In fact, your change is not even safe.  There is a race between the SE finishing all the work in the queue and items being added.  The SE is still scheduled, but in the process of exiting.  During that window of time, calls to `execute()` will fail to reschedule.  If you haven't already, try running the TransportBenchmark with JMH;  With near certainty I can guess that the code will hang due to lost callbacks.

carl-mastrangelo:
Also, for reference, I sent out https://github.com/grpc/grpc-java/pull/3778  a few weeks ago to explore optimizations of even dropping CLQ from SE.  CLQ is designed to be correct with an arbitrary number threads pushing and popping to it.   SE has prior knowledge that there will only be one consumer (itself), an one producer (the network thread).  I have yet to benchmark it.

njhill:
Thanks again @carl-mastrangelo, I hadn't noticed that new addition to guava. If/when grpc-java moves to that version, maybe a similar change to the one in this PR could be considered? `SequentialExecutor` itself is package-private though, so I guess some trick similar to what is currently done for `directExecutor()` would be needed.

> Concerning `Executors.newSingleThreadExecutor()` : it isn't actually serializing. The problem is reentrancy. 

Is reentrancy really a problem, since this is a single _separate_ thread backed by a queue, so Runnables submitted from the executor thread itself will be queued rather than run immediately (unlike `Executors.directExecutor()` which sounds more like what you're describing)?

Re the SE optimizations, sorry for including a bug. The race was introduced by changing the `isEmpty()` to an optimistic `poll()`, which of course can't be done if the running state isn't owned. [I fixed it](https://github.com/grpc/grpc-java/compare/master...njhill:ser-exec-opt) and the overall change is simpler. It actually didn't contain an additional compare-and-swap, that's just the one which was previously done inside the call to `schedule(null)`.

The observation was just that if new runnables are found in the queue after the loop finishes, the existing task/thread can continue to process the queue rather than having to re-schedule via the delegate executor (avoiding the associated sync and context-switch).

And thanks for pointing out your SPSC version, it looks neat. The same "inlining" simplification can also be applied to that. I have been playing with benchmarking all of these variations (including yours) but I think I need to do it on a system where same-run performance is a bit more consistent. If I discover any meaningful improvement then I'll report back.

Thanks again for your patience and I'll make sure to do the benchmarking in future before suggesting any optimization!

carl-mastrangelo:
> Is reentrancy really a problem, since this is a single separate thread backed by a queue

It is.  For example suppose you want to limit the queue size of your single threaded executor.   If you used the CallerRunsPolicy, it degrades back to a direct executor.

The fixed version of your code looks better now.  One other optimization you might consider making is swapping from compareAndSet to getAndSet.   CAS is slower and I have been meaning to get rid of it.

PR #3915, opened 2018-01-04
protobuf-lite: should not have dependency on core_internal

dapengzhang0:


core/src/main/java/io/grpc/Grpc.java
```diff
@@ -40,4 +40,11 @@ private Grpc() {
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1710")
   public static final Attributes.Key<SSLSession> TRANSPORT_ATTR_SSL_SESSION =
           Attributes.Key.of("ssl-session");
+
+  /**
+   * The default maximum uncompressed size (in bytes) for inbound messages. Defaults to 4 MiB.
+   * Not for public use.
+   */
+  @Internal
+  public static final int INTERNAL_DEFAULT_MAX_MESSAGE_SIZE = 4 * 1024 * 1024;
```
carl-mastrangelo: Can't put it here, it would be public.

dapengzhang0:
Changed. PTAL @carl-mastrangelo

PR #3917, opened 2018-01-05
services: introduce BinaryLogSinkProvider

zpencer:
The `BinaryLog` will write `GrpcLogEntry` objects to the sink.

### DIFF
--- a/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
+++ b/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
@@ -30,6 +30,7 @@ import io.grpc.ServerCallHandler;
 import io.grpc.ServerInterceptor;
 import io.grpc.ServerMethodDefinition;
 import java.io.Closeable;
+import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -159,6 +160,12 @@ public abstract class BinaryLogProvider implements Closeable {
   @Nullable
   public abstract ClientInterceptor getClientInterceptor(String fullMethodName);
 
+  @Override
+  public void close() throws IOException {
+    // default impl: noop
+    // TODO(zpencer): make BinaryLogProvider provide a BinaryLog, and this method belongs there
+  }
+
   /**
    * A priority, from 0 to 10 that this provider should be used, taking the current environment into
    * consideration. 5 should be considered the default, and then tweaked based on environment
@@ -186,11 +193,6 @@ public abstract class BinaryLogProvider implements Closeable {
     @Override
     protected int priority() {
       return 0;
-    }
-
-    @Override
-    public void close() {
-      // noop
     }
   }
 
--- a/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
+++ b/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
@@ -29,7 +29,9 @@ import io.grpc.MethodDescriptor.Marshaller;
 import io.grpc.ServerCallHandler;
 import io.grpc.ServerInterceptor;
 import io.grpc.ServerMethodDefinition;
+import java.io.Closeable;
 import java.io.InputStream;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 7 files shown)


services/src/main/java/io/grpc/services/BinaryLog.java
```diff
@@ -45,45 +48,120 @@
...
     return this.maxHeaderBytes == that.maxHeaderBytes
-        && this.maxMessageBytes == that.maxMessageBytes;
+        && this.maxMessageBytes == that.maxMessageBytes
+        && this.sink.equals(that.sink);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hashCode(maxHeaderBytes, maxMessageBytes);
+    return Objects.hashCode(maxHeaderBytes, maxMessageBytes, sink);
```
carl-mastrangelo: Since sink is an abstract class, and not immutable, the hash code is not well defined.  maybe just leave it out?

services/src/main/java/io/grpc/services/BinaryLogSinkProvider.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * Subclasses must be thread safe, and are responsible for writing the binary log message to
+ * the appropriate destination.
+ */
+@ThreadSafe
+public abstract class BinaryLogSinkProvider implements Closeable {
+
+  public static BinaryLogSinkProvider provider() {
```
carl-mastrangelo: what does this method do?

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -138,6 +140,11 @@ public ClientInterceptor getClientInterceptor(String fullMethodName) {
     protected int priority() {
       return 0;
     }
+
+    @Override
+    public void close() throws IOException {
```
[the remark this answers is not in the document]
zpencer: done

services/src/main/java/io/grpc/services/BinaryLog.java
zpencer: done

services/src/main/java/io/grpc/services/BinaryLogSinkProvider.java
zpencer: I have a PR to create a ServiceProvider util class. If it gets merged then we can avoid some duplicate code here.
https://github.com/grpc/grpc-java/pull/3886

Added a TODO placeholder.

zpencer:
@carl-mastrangelo ping

services/src/main/java/io/grpc/services/BinaryLogSinkProvider.java
carl-mastrangelo: This was a mistake we made with ServerBuilder.   Having the static factory of the class be on itself and be extensible made for confusing behavior.    The factory and the Sink should be in separate classes.

zpencer:
@carl-mastrangelo ping

zpencer:
@carl-mastrangelo  PTAL, rebased with some minor updates

services/src/main/java/io/grpc/services/BinaryLog.java
```diff
@@ -46,33 +49,124 @@
...
+    sink.write(entry);
+  }
+
+  /**
+   * Logs the outbound message. This method logs the appropriate number of bytes from
+   * {@code message}, and returns an {@link InputStream} that contains the original message.
+   * The number of bytes logged is determined by the binary logging configuration.
+   */
+  public void logOutboundMessage(
+      byte[] message, boolean compressed, boolean isServer, byte[] callId) {
```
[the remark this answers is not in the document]
zpencer: It can, although I don't think it's going to make the code much simpler at the moment. Is this to make the signatures match up with some proposed future changes on the Marshaller API?

I also see a UnsafeByteOperations.unsfeWrap() that allows protobuf to avoid making another copy of the contents. It seems appropriate for this use case, since the underlying byte[]/ByteBuffer is only shared between the binary logger and grpc marshaller, and neither will modify the bytes.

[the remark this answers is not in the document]
carl-mastrangelo: The change is for performance, not so much for API convenience.

zpencer:
@carl-mastrangelo PTAL

PR #3924, opened 2018-01-06
bazel: fix protobuf sha256

jyane:
Currently, grpc-java could not build by bazel because protobuf's sha256 is incorrect.
It might need some CI (travis or kokoro(?)) checking.

```
$ bazel build //protobuf:all
....................
ERROR: /Users/jyane/workspace/repos/github.com/jyane/grpc-java/protobuf/BUILD.bazel:1:1: no such package '@com_google_protobuf_java//': Error downloading [https://github.com/google/protobuf/archive/v3.5.1.zip] to /private/var/tmp/_bazel_jyane/6860dd76b063c3142c2bc69dd88c44cb/external/com_google_protobuf_java/v3.5.1.zip: Checksum was 1f8b9b202e9a4e467ff0b0f25facb1642727cdf5e69092038f15b37c75b99e45 but wanted cef7f1b5a7c5fba672bec2a319246e8feba471f04dcebfe362d55930ee7c1c30 and referenced by '//protobuf:protobuf'
ERROR: /Users/jyane/workspace/repos/github.com/jyane/grpc-java/protobuf/BUILD.bazel:1:1: no such package '@com_google_protobuf_java//': Error downloading [https://github.com/google/protobuf/archive/v3.5.1.zip] to /private/var/tmp/_bazel_jyane/6860dd76b063c3142c2bc69dd88c44cb/external/com_google_protobuf_java/v3.5.1.zip: Checksum was 1f8b9b202e9a4e467ff0b0f25facb1642727cdf5e69092038f15b37c75b99e45 but wanted cef7f1b5a7c5fba672bec2a319246e8feba471f04dcebfe362d55930ee7c1c30 and referenced by '//protobuf:protobuf'
ERROR: Analysis of target '//protobuf:protobuf' failed; build aborted: no such package '@com_google_protobuf_java//': Error downloading [https://github.com/google/protobuf/archive/v3.5.1.zip] to /private/var/tmp/_bazel_jyane/6860dd76b063c3142c2bc69dd88c44cb/external/com_google_protobuf_java/v3.5.1.zip: Checksum was 1f8b9b202e9a4e467ff0b0f25facb1642727cdf5e69092038f15b37c75b99e45 but wanted cef7f1b5a7c5fba672bec2a319246e8feba471f04dcebfe362d55930ee7c1c30
INFO: Elapsed time: 23.761s
FAILED: Build did NOT complete successfully (13 packages loaded)
```

Issue #3502, opened 2017-09-26
Add CI for Bazel


ejona86:
While users are good at noticing when Bazel breaks, I suspect they'd appreciate it to compile without their intervention.

This probably should be on Kokoro.

drigz:
FYI: https://github.com/bazelbuild/continuous-integration/blob/master/docs/owner.md is the documentation for adding a new project to https://ci.bazel.io.

PR #3925, opened 2018-01-06
bazel,travis: add bazel build checking

jyane:
add bazel build checking to `.travis.yml`.

This PR's CI would be failed because protobuf's sha256 is incorrect.
When it (https://github.com/grpc/grpc-java/pull/3924) be merged, I will rebase this branch to pass CI.

This PR is partial fix for this issue: https://github.com/grpc/grpc-java/issues/3502

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,14 @@ sudo: false
 
 language: java
 
+addons:
+  apt:
+    packages:
+      - wget
+
 env:
   global:
+    - BAZEL_VERSION=0.9.0
     - GRADLE_OPTS=-Xmx512m
     - PROTOBUF_VERSION=3.5.1
     - LDFLAGS=-L/tmp/protobuf/lib
@@ -23,6 +29,9 @@ before_install:
   - echo "checkstyle.ignoreFailures=false" >> $HOME/.gradle/gradle.properties
   - echo "failOnWarnings=true" >> $HOME/.gradle/gradle.properties
   - echo "errorProne=true" >> $HOME/.gradle/gradle.properties
+  - mkdir -p /tmp/bazel
+  - wget -nc "https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel_${BAZEL_VERSION}-linux-x86_64.deb" -P /tmp/bazel
+  - sudo dpkg -i "/tmp/bazel/bazel_${BAZEL_VERSION}-linux-x86_64.deb"
 
 install:
   - ./gradlew assemble generateTestProto install
@@ -34,6 +43,7 @@ before_script:
 
 script:
   - ./gradlew check :grpc-all:jacocoTestReport
+  - bazel build //compiler:all //context:all //core:all //netty:all //okhttp:all //protobuf:all //protobuf-lite:all //protobuf-nano:all //stub:all //testing:all
 
 after_success:
   - if \[ "$TRAVIS_OS_NAME" = linux \]; then ./gradlew :grpc-all:coveralls; fi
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -34,6 +43,7 @@ before_script:
 
 script:
   - ./gradlew check :grpc-all:jacocoTestReport
+  - bazel build //compiler:all //context:all //core:all //netty:all //okhttp:all //protobuf:all //protobuf-lite:all //protobuf-nano:all //stub:all //testing:all
```
carl-mastrangelo: This can be `//...`  right?

ejona86: +1

dapengzhang0: Slightly different and `...` will unintentionally include `//examples:all` which is a non-target.

ejona86: If examples is the only reason for the expansion, then we should subtract it out:

```
bazel build -- ... -examples/...
```

Also `/...` should be preferred to `:all` here, since there's no need to disable descending into subpackages (although they don't exist today).

Honestly, I would be okay with including the examples in this line. I know that users will see something slightly different since the WORKSPACE is handled differently, but building examples using its own WORKSPACE will rebuild everything a _third_ time.

ejona86:
I think it may be better to use a matrix build (or similar) here. The bazel build takes 6 minutes, which is a third of the total time. A matrix build could run it in parallel.

.travis.yml
jyane: I will change building each services to `...`.

jyane:
Thank you for your comments!
I will change to use the parallel computing :)

Caching bazel artifacts is also probably needed.
But I think splitting PRs would be better because the caching might break CI (voilate/corrupt).

jyane:
I have added matrix testing via my forked repository (https://github.com/jyane/grpc-java/pull/1), but it is so dirty...

.travis.yml
```diff
@@ -1,43 +1,40 @@
-sudo: false
+sudo: required
```
ejona86: Wait... how did it work before with `sudo: false`? I didn't look closely enough earlier to notice it was running `dpkg`. The need for sudo is why I didn't mess with this earlier; I'd rather not lose that. It looks like it is possible [to install into the home directory though](https://docs.bazel.build/versions/master/install-ubuntu.html).

.travis.yml
```diff
@@ -1,43 +1,40 @@
...
 
 language: java
 
+addons:
+  apt:
+    packages:
+      - wget
```
ejona86: Is this actually necessary? We've already been using wget in make_dependencies.sh and the [Travis documentation](https://docs.travis-ci.com/user/reference/trusty/) says wget is already installed.

.travis.yml
jyane: I don't know why...

.travis.yml
jyane: Sorry, it is an oversight...

jyane:
Thank you very much for reviews and comments, @ejona86, c.arl-mastrangelo and d.apengzhang0.

Close this PR due to https://github.com/grpc/grpc-java/pull/3933

ejona86:
@jyane, I wasn't trying to criticize your decisions in the PR. Yes, my initial issue said Kokoro, but that wasn't a strong requirement. It actually pains me quite a bit how "blind" Kokoro is to external users. And, yes, lack of CI for Bazel has been creating problems and so a PR to add Travis testing for it is quite reasonable.

My preference is actually to do it in Travis as well, but I had anticipated some of the problems discussed here and thought Kokoro would be simpler. My main problem with Travis here is that it is a pain to have a completely different build[1]. Having 6 highly-coupled files would be unfortunate. I think there can be some acceptable organizations of the commands, but it seems like Kokoro is simple enough here to warrant accepting its limitations.

1. This has been a problem for Android builds as well. We have to change the language to "android," and then the build wouldn't work on OS X.

jyane:
Thank you for your comment, and I understand. Please don't care about me. You can always close my PR.

PR #3933, opened 2018-01-09
buildscripts: Add Kokoro-based Bazel CI

ejona86:
The internal change is cl/181324357

So... this was really easy (granted, completely untested). But it does have the standard problem of Kokoro of being hard to see externally. I think that's likely "okay" right now, and eventually that limitation should go away. Given there aren't even any tests being run, I hope that just the "failure" signal would be enough for most people.

I'm not against adding bazel to Travis, but it seemed to become a bit of a can of worms, as could be seen in #3925. I think we could make it reasonably clean, but I'm not sure it's worth the effort given how simple this solution is.

CC @jyane

### DIFF
--- a/buildscripts/kokoro/bazel.cfg
+++ b/buildscripts/kokoro/bazel.cfg
@@ -1,0 +1,5 @@
+# Config file for internal CI
+
+# Location of the continuous shell script in repository.
+build_file: "grpc-java/buildscripts/kokoro/bazel.sh"
+timeout_mins: 45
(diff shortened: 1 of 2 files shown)


jyane:
Using kokoro is much simple I thought was. I have closed https://github.com/grpc/grpc-java/pull/3925.
I'm looking forward to opening the kokoro. (Kokoro(心) means heart/mind in Japanese)

ejona86:
@jyane, you're not able to see all the changes; there's a little bit of glue internally. But it's about the same complexity of what you see here. I think there are some ways to upload logs to GCS for public viewing, but it requires enough effort that it's worth waiting to see how much pain is being caused first.

ejona86:
The new CI ran and appears to be working.

ejona86:
@zpencer, friendly ping. Should be easy.

ejona86:
Fixed #3502

PR #3936, opened 2018-01-09
buildscripts: Avoid Java 9 for Windows CI

ejona86:
Our build isn't happy on Java 9 yet.


ejona86:
@dapengzhang0, ping. This fixes the broken Windows CI

PR #3937, opened 2018-01-09
core: install the binary logging client interceptor

zpencer:
Some changes required to make this work:
- Fix `MethodDescriptor.toBuilder` so that it copies all fields
- It is better to make `BinaryLogProvider.provider` return `null` rather than the NullBinaryLogProvider. This is because on client side, the pipeline of interceptors is hard coded when the ManagedChannelImpl is created. It is desirable to avoid inserting a binary logging interceptor shim if we know there is no binary log.


examples/src/main/proto/testservice.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
```
ejona86: Why was this file added? It seems unused.

zpencer: Oops removed

core/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -438,7 +438,8 @@ public static String extractFullServiceName(String fullMethodName) {
         .setFullMethodName(fullMethodName)
         .setIdempotent(idempotent)
         .setSafe(safe)
-        .setSampledToLocalTracing(sampledToLocalTracing);
+        .setSampledToLocalTracing(sampledToLocalTracing)
```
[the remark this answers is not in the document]
zpencer: Posted https://github.com/grpc/grpc-java/pull/3941, but I'll leave this part alone so the test can work.

examples/src/test/java/io/grpc/examples/header/HeaderServerInterceptorTest.java
```diff
@@ -78,17 +75,21 @@ public void sayHello(HelloRequest request, StreamObserver<HelloReply> responseOb
...
-      ClientCall.Listener<?> spyListener;
+      Metadata capturedMetadata;
 
       @Override
       public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
           MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
         return new SimpleForwardingClientCall<ReqT, RespT>(next.newCall(method, callOptions)) {
           @Override
           public void start(Listener<RespT> responseListener, Metadata headers) {
-            spyListener = responseListener =
```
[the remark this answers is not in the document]
zpencer: Restructuring of `ForwardingClientCallListener` and `PartialForwardingClientCallListener` broke the spy, probably because the method we want is now an inherited method. 

This works but is less obvious than the non spy():
```
spy(new SimpleForwardingClientCallListener(responseListener) {
  @Override
  public void onHeaders(Metadata headers) {
    super.onHeaders(headers);
  }
});
```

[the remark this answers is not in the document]
ejona86: It is broken by PartialForwardingClientCallListener being package-private. Making it public fixes it. But it's also fixed if I bump mockito to 2.13.0.

If we use AdditionalAnswers.delegatesTo instead of spy (which is what we should do anywhere we use spy, really), then it starts working:
```java
            spyListener = responseListener =
                mock(ClientCall.Listener.class,
                    delegatesTo(new SimpleForwardingClientCallListener(responseListener) {}));
```

Although then it can be simplified to:
```java
            spyListener = responseListener =
                mock(ClientCall.Listener.class, delegatesTo(responseListener));
```

zpencer:
Rebased against master, resolved conflicts. @ejona86 PTAL

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -94,21 +112,21 @@ public int compare(BinaryLogProvider f1, BinaryLogProvider f2) {
 
   /**
    * Returns a {@link ServerInterceptor} for binary logging. gRPC is free to cache the interceptor,
-   * so the interceptor must be reusable across server calls. At runtime, the request and response
-   * types passed into the interceptor is always {@link java.io.InputStream}.
+   * so the interceptor must be reusable across calls. At runtime, the request and response
+   * marshallers are always {@code Marshaller<Inputstream>}.
```
ejona86: s/stream/Stream/

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -154,4 +172,44 @@ protected static boolean isAndroid() {
       return false;
     }
   }
+
+  // Creating a named class makes debugging easier
+  private static final class IdentityMarshaller implements Marshaller<InputStream> {
+    @Override
+    public InputStream stream(InputStream value) {
+      return value;
```
ejona86: Add a TODO to deal with retry?

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
zpencer: fixed

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
zpencer: fixed

Issue #357, opened 2015-04-29
Try to allow building java without having to build codegen


ejona86:
It may be possible to use the last released codegen binary.

zhangkun83:
Some thoughts:

- Run a continuous build service (e.g., Jenkins) to build codegen and push it to snapshot repo for every commit. We will need three machines (Linux, Mac and Windows) for this.
- Disable ``grpc-compiler`` by default when building the root project. If ``grpc-compiler`` is disabled, other subproject will use the codegen with the same version number as the current code. So if the code is ``SNAPSHOT``, codegen will be from the snapshot repo. This should relieve a lot of pain for most contributors who don't touch the codegen.
- Contributors that do change the codegen must use an option to enable building compiler locally. Alternatively they could install their changed codegen locally.

ejona86:
I thought of another option that I think could work well and isn't hacky: we check in the generated code. We make code generation not part of "compile" and only run it when we are modifying the code generator.

zhangkun83:
This is actually a good idea. Where do you think we should put the generated code? I am thinking ``src/generated/java/``

ejona86:
SGTM.

PR #3939, opened 2018-01-10
bazel,grpclb: add a bazel build definition

jyane:
add a bazel build definition to grpclb.

I think `src/generated` was created for resolving this issue https://github.com/grpc/grpc-java/issues/357 then I did not add generated files. I just use `java_grpc_library`.

But Bazel requires a full path from WORKSPACE in a proto file (https://github.com/cgrushko/proto_library/blob/04369f0d2ade8c8566727e0b6f3a53f1ba8925c0/src/person.proto#L5) for importing other proto files.
In this PR, it does not relate, but services will require a full path (https://github.com/grpc/grpc-java/compare/master...jyane:add-services-bazel#diff-afc0c0ce1aad649c36404301917e883cR21)

"Do not add generated files" or "Use generated files in Bazel" which is better for now?

### DIFF
--- a/grpclb/BUILD.bazel
+++ b/grpclb/BUILD.bazel
@@ -21,9 +21,7 @@ java_library(
 
 proto_library(
     name = "load_balancer_proto",
-    srcs = [
-        "src/main/proto/load_balancer.proto",
-    ],
+    srcs = ["src/main/proto/load_balancer.proto"],
     deps = [
         "@com_google_protobuf//:timestamp_proto",
         "@com_google_protobuf//:duration_proto",
@@ -32,17 +30,11 @@ proto_library(
 
 java_proto_library(
     name = "load_balancer_java_proto",
-    deps = [
-        ":load_balancer_proto",
-    ],
+    deps = [":load_balancer_proto"],
 )
 
 java_grpc_library(
     name = "load_balancer_java_grpc",
-    srcs = [
-        ":load_balancer_proto",
-    ],
-    deps = [
-        ":load_balancer_java_proto",
-    ],
+    srcs = [":load_balancer_proto"],
+    deps = [":load_balancer_java_proto"],
 )
--- a/grpclb/BUILD.bazel
+++ b/grpclb/BUILD.bazel
@@ -1,0 +1,48 @@
+load("//:java_grpc_library.bzl", "java_grpc_library")
+
+java_library(
+    name = "grpclb",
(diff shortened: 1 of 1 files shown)


ejona86:
> I think src/generated was created for resolving this issue #357 then I did not add generated files. I just use java_grpc_library.

+1. That's what I was hoping to do.

> But Bazel requires a full path from WORKSPACE in a proto file for importing other proto files.

Yeah. That's going to be a pain. Just before the holidays @carl-mastrangelo and I were figuring out how we wanted to handle that, because we have that problem internally as well. I think the plan was to generally rely on https://github.com/grpc/grpc-proto for any protos shared between repositories. Those protos could have the proper imports and we'd have copies in the repo for only gradle to use.

We'd still have to figure out something for testing and java-specific protos, since they wouldn't live in grpc-proto. We could maybe just use genrule()+sed to rewrite them.

grpclb/BUILD.bazel
```diff
@@ -0,0 +1,48 @@
...
+    ],
+    deps = [
+        "@com_google_protobuf//:timestamp_proto",
+        "@com_google_protobuf//:duration_proto",
+    ],
+)
+
+java_proto_library(
+    name = "load_balancer_java_proto",
+    deps = [
```
ejona86: nit for in the future: I've been preferring to put the srcs and deps for *_{grpc,proto}_library on one line, since they will normally only have one element and there can be a ton of those rules, for the various languages.

This formatting is still acceptable, though.

ejona86:
@jyane, I'm going to try to merge #3933 before this PR, to use the shiny new CI.

grpclb/BUILD.bazel
jyane: I will fix it, please wait for the merging 🙇

jyane:
fixed, PTAL.

> Just before the holidays c.arl-mastrangelo and I were figuring out how we wanted to handle that, because we have that problem internally as well.

> We'd still have to figure out something for testing and java-specific protos, since they wouldn't live in grpc-proto. We could maybe just use genrule()+sed to rewrite them.

It is very interesting for me :) I have tried to add `import_prefix` to `proto_library` but I gave up(https://github.com/jyane/bazel/commit/fc96cc680a5120b9c2562efa8e560bd764ed0e96).

ejona86:
@zhangkun83, could you also review this?

PR #3943, opened 2018-01-11
netty: make GrpcHttp2ConnectionHandler able to indicate it will no longer be used

carl-mastrangelo:
This adds a method on GrpcHttp2ConnectionHandler which, when called, indicates that the channel associated with the handler is no longer needed.   

Notes:

* The handler may not be on the channel, but will either a. need to be added or will never be added.
* The channel will only be "unused" on the server side.
* It is expected that after calling `notifyUnused()`, the channel will be deregistered from the loop without being properly shut down.   This allows the channel to be handed off to a Non-netty API.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/GrpcHttp2ConnectionHandler.java
+++ b/netty/src/main/java/io/grpc/netty/GrpcHttp2ConnectionHandler.java
@@ -23,19 +23,25 @@ import io.netty.handler.codec.http2.Http2ConnectionDecoder;
 import io.netty.handler.codec.http2.Http2ConnectionEncoder;
 import io.netty.handler.codec.http2.Http2ConnectionHandler;
 import io.netty.handler.codec.http2.Http2Settings;
+import javax.annotation.Nullable;
 
 /**
  * gRPC wrapper for {@link Http2ConnectionHandler}.
  */
 @Internal
 public abstract class GrpcHttp2ConnectionHandler extends Http2ConnectionHandler {
-  public GrpcHttp2ConnectionHandler(Http2ConnectionDecoder decoder,
+
+  @Nullable
+  protected final ChannelPromise channelUnused;
+
+  public GrpcHttp2ConnectionHandler(
+      ChannelPromise channelUnused,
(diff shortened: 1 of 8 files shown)


netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -156,6 +157,24 @@ public void start(ServerListener serverListener) throws IOException {
...
+          @Override
+          public void operationComplete(ChannelFuture future) throws Exception {
+            if (!done) {
+              done = true;
+              eventLoopReferenceCounter.release();
+            }
+          }
+        }
+
+        ChannelFutureListener loopReleaser = new LoopReleaser();
```
ejona86: Move this to just before its usage?

netty/src/main/java/io/grpc/netty/GrpcHttp2ConnectionHandler.java
```diff
@@ -34,6 +35,8 @@ public GrpcHttp2ConnectionHandler(Http2ConnectionDecoder decoder,
     super(decoder, encoder, initialSettings);
   }
 
+  protected ChannelPromise channelUnused;
```
ejona86: Place before the constructor?

ejona86: Could this be initialized from the constructor? Having the super-class reach in is harder to follow. It still being @Nullable is fine. I'd be fine with either two constructors (one with and one without the promise) or having the client pass in a `null` instance.

carl-mastrangelo: Done.  When I was making this CL, I thought there was a cycle between creating the channel (for the promise) and the handler.  That only ended up in the tests, so Done.

netty/src/main/java/io/grpc/netty/NettyServer.java
carl-mastrangelo: Done.

PR #3957, opened 2018-01-11
Allow specifying a ProxyDetector per ManagedChannel

snowp:
Problem: In our current setup we use grpc to communicate between apps within our datacenters, and company policy is to route all external traffic through a web proxy (for security reasons we want to whitelist who we talk to externally). However, due to how grpc handles proxies (through the env variable or java proxy settings) there is no way for us to specify a different proxy behavior between external grpc clients (e.g. for calling out to GCP) and internal grpc clients (which we don't want going through our webproxy). 

This PR allows specifying what `ProxyDetector` you want to use when building a `ManagedChannel`. In order to do so I also had to update the `NameResolver` abstract class as currently the `DNSNameResolver` requires knowledge about whether we're using a proxy or not. 

As a follow up to this I can imagine us removing the special casing for App Engine in selecting a default `ProxyDetector` in favor of using this new API, but that can follow in a later PR. 

cc @lukaszx0 @ejona86 @carl-mastrangelo

### DIFF
--- a/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
@@ -16,6 +16,7 @@
 
 package io.grpc;
 
+import io.grpc.internal.ProxyDetector;
 import java.util.List;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
@@ -142,6 +143,11 @@ public abstract class ForwardingChannelBuilder<T extends ForwardingChannelBuilde
   @Override
   public T maxInboundMessageSize(int max) {
     delegate().maxInboundMessageSize(max);
+    return thisT();
+  }
+
+  @Override public T proxyDetector(ProxyDetector proxyDetector) {
+    delegate().proxyDetector(proxyDetector);
     return thisT();
   }
 
--- a/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
@@ -16,7 +16,7 @@
 
 package io.grpc;
 
-import io.grpc.internal.ProxyDetector;
+import java.net.Proxy;
 import java.util.List;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
@@ -146,8 +146,8 @@ public abstract class ForwardingChannelBuilder<T extends ForwardingChannelBuilde
     return thisT();
   }
 
-  @Override public T proxyDetector(ProxyDetector proxyDetector) {
-    delegate().proxyDetector(proxyDetector);
+  @Override public T proxy(Proxy proxy) {
+    delegate().proxy(proxy);
     return thisT();
   }
 
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -16,7 +16,7 @@
 
 package io.grpc;
 
-import io.grpc.internal.ProxyDetector;
+import java.net.Proxy;
 import java.util.List;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
@@ -281,13 +281,13 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   }
 
   /**
-   * Sets the {@link ProxyDetector} to use when resolving proxy parameters for outgoing
+   * Sets the {@link Proxy} to use when resolving proxy parameters for outgoing
    * connections.
    *
-   * @param proxyDetector the proxy detector to use
+   * @param proxy the proxy to use
    * @return this
    */
-  public T proxyDetector(ProxyDetector proxyDetector) {
+  public T proxy(Proxy proxy) {
     // intentional nop
     return thisT();
   }
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -16,6 +16,7 @@
 
 package io.grpc;
 
+import io.grpc.internal.ProxyDetector;
 import java.util.List;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
@@ -275,6 +276,18 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/2307")
   public T maxInboundMessageSize(int max) {
+    // intentional nop
+    return thisT();
+  }
+
+  /**
+   * Sets the {@link ProxyDetector} to use when resolving proxy parameters for outgoing
+   * connections.
+   *
+   * @param proxyDetector the proxy detector to use
+   * @return this
+   */
+  public T proxyDetector(ProxyDetector proxyDetector) {
     // intentional nop
     return thisT();
   }
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -81,7 +81,7 @@ final class DnsNameResolver extends NameResolver {
   private final int port;
   private final Resource<ScheduledExecutorService> timerServiceResource;
   private final Resource<ExecutorService> executorResource;
-  private final ProxyDetector proxyDetector;
+  private final boolean usingProxy;
   @GuardedBy("this")
   private boolean shutdown;
   @GuardedBy("this")
@@ -95,14 +95,20 @@ final class DnsNameResolver extends NameResolver {
   @GuardedBy("this")
   private Listener listener;
 
-  DnsNameResolver(@Nullable String nsAuthority, String name, Attributes params,
+  DnsNameResolver(@Nullable String nsAuthority, URI targetUri, Attributes params,
       Resource<ScheduledExecutorService> timerServiceResource,
       Resource<ExecutorService> executorResource,
-      ProxyDetector proxyDetector) {
+      boolean usingProxy) {
     // TODO: if a DNS server is provided as nsAuthority, use it.
     // https://www.captechconsulting.com/blogs/accessing-the-dusty-corners-of-dns-with-java
     this.timerServiceResource = timerServiceResource;
     this.executorResource = executorResource;
+    this.usingProxy = usingProxy;
+
+    String targetPath = Preconditions.checkNotNull(targetUri.getPath(), "targetPath");
+    Preconditions.checkArgument(targetPath.startsWith("/"),
+        "the path component (%s) of the target (%s) must start with '/'", targetPath, targetUri);
+    String name = targetPath.substring(1);
     // Must prepend a "//" to the name when constructing a URI, otherwise it will be treated as an
     // opaque URI, thus the authority and host of the resulted URI would be null.
     URI nameUri = URI.create("//" + name);
@@ -120,7 +126,6 @@ final class DnsNameResolver extends NameResolver {
     } else {
       port = nameUri.getPort();
     }
-    this.proxyDetector = proxyDetector;
   }
 
   @Override
@@ -160,9 +165,8 @@ final class DnsNameResolver extends NameResolver {
           resolving = true;
         }
         try {
-          InetSocketAddress destination = InetSocketAddress.createUnresolved(host, port);
-          ProxyParameters proxy = proxyDetector.proxyFor(destination);
-          if (proxy != null) {
+          if (usingProxy) {
+            InetSocketAddress destination = InetSocketAddress.createUnresolved(host, port);
             EquivalentAddressGroup server = new EquivalentAddressGroup(destination);
             savedListener.onAddresses(Collections.singletonList(server), Attributes.EMPTY);
             return;
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -298,7 +298,8 @@ public final class ManagedChannelImpl
       // did not cancel idleModeTimer, both of which are bugs.
       nameResolver.shutdown();
       nameResolverStarted = false;
-      nameResolver = getNameResolver(target, nameResolverFactory, nameResolverParams);
+      nameResolver =
+          getNameResolver(target, nameResolverFactory, nameResolverParams, proxyDetector);
       lbHelper.lb.shutdown();
       lbHelper = null;
       subchannelPicker = null;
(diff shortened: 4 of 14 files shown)


thelinuxfoundation:

* User @snowp isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

ejona86:
I'll have to look into this in more detail, but so I understand better, why can't the ProxySelector distinguish between internal and external hosts?

We were going to need some more plumbing in the NameResolver anyway. Are you using a custom NameResolver for internal hosts, such that maybe with the proper plumbing it could specify the proxy to use?

lukaszx0:
> I'll have to look into this in more detail, but so I understand better, why can't the ProxySelector distinguish between internal and external hosts?

How? Unless I'm missing something, there's no way to instruct grpc client that the given client is created for bigtable or your internal foobar service.

Basically what we need, and it's just an example because I know that such plumbing is not trivial in current state of things, is `setProxyHost(...)` (or `setProxyProvider` etc.) method on ManagedChannelBuilder so that you can set proxy for bigtable client and leave it unset for internal foobar service.

> We were going to need some more plumbing in the NameResolver anyway. Are you using a custom NameResolver for internal hosts, such that maybe with the proper plumbing it could specify the proxy to use?

Yeah, I think that'd work. I'm not sure off top of my head if the proxy is needed in other parts of networking stack in grpc or just NameResolver. If it's the latter, than improving `NameResolver` API to expose proxy setting would do, I think, and we wouldn't not need to make any changes on `ManagedChannel` level.

core/src/main/java/io/grpc/ForwardingChannelBuilder.java
```diff
@@ -16,6 +16,7 @@
 
 package io.grpc;
 
+import io.grpc.internal.ProxyDetector;
```
carl-mastrangelo: We can't depend on io.grpc.internal from io.grpc

snowp:
> The idea seems fine, but you''ll need to find another way to expose ProxyDetector.

How do you feel about making `ProxyDetector` part of the public API? We could rename it to something like `ProxyProvider` or something that's more generic, while keeping the same interface (SocketAddress -> @Nullable ProxyDetails). Another option would be to allow specifying the proxy details directly in the `ManagedChannelBuilder`, although it seems like a `Provider` interface would be more flexible. 

> We were going to need some more plumbing in the NameResolver anyway. Are you using a custom NameResolver for internal hosts, such that maybe with the proper plumbing it could specify the proxy to use?

Yeah we use our own `NameResolver` to resolve the internal addresses, and having it decide the proxy information would most likely work for our use case like @lukaszx0 said. It does seem a bit strange for a `NameResolver` to also resolve proxy information (seems like a slight breach of the single responsibility principle). The cpp implementation seems to first figure out proxy details then pass it to the resolver: https://github.com/grpc/grpc/blob/master/src/core/ext/filters/client_channel/client_channel.cc#L757

I don't really have a strong opinion on which option is best, so I'd love to hear what you all think

ejona86:
> How? Unless I'm missing something, there's no way to instruct grpc client that the given client is created for bigtable or your internal foobar service.

If the hostname passed to [ProxySelector.select](https://docs.oracle.com/javase/8/docs/api/java/net/ProxySelector.html#select-java.net.URI-)  is `bigtable.googleapis.com`, then it's external.

Normally I'd expect some way to detect internal hostnames (IP addresses matching a certain pattern and some "internal" domains) and have everything else go through the proxy. But other options are possible.

ProxySelector is similar to using a WPAD PAC (but without the central management).

lukaszx0:
> If the hostname passed to ProxySelector.select is bigtable.googleapis.com, then it's external.

Ah, I see. But that's true only if I can control my clients. The example I've used is a real world problem that we have: a `foo` app that talks to `bar` and `bigtable`. We use official google `bigtable` client so we don't have "low level" access to grpc client that's abstracted away in `bigtable` client so when we set a global proxy env var, it affects *all* grpc clients in a `foo` process.

Simply put, currently, proxy env setting is global for the process and we want to be able to specify different proxies (or none) for different clients in the same process.

ejona86:
> We use official google bigtable client so we don't have "low level" access to grpc client that's abstracted away in bigtable client so when we set a global proxy env var, it affects all grpc clients in a foo process.

Now it sounds like you're arguing that this PR doesn't work, because you couldn't configure the ManagedChannel anyway because it is hidden from the API surface.

> Simply put, currently, proxy env setting is global for the process and we want to be able to specify different proxies (or none) for different clients in the same process.

That's exactly what ProxySelector allows. ProxySelector itself is global, but given the intended destination it can choose appropriate proxy configuration. Since it is global that means it also works for client libraries that hide the ManagedChannel.

lukaszx0:
> Now it sounds like you're arguing that this PR doesn't work, because you couldn't configure the ManagedChannel anyway because it is hidden from the API surface.

Yes, even with this change, the 3rd party clients that we use will have to be updated to allow users to change this (so we'll need to send follow up PR for bigtable client etc.). With env vars currently it's impossible.

> That's exactly what ProxySelector allows. ProxySelector itself is global, but given the intended destination it can choose appropriate proxy configuration. Since it is global that means it also works for client libraries that hide the ManagedChannel.

I apologize, my bad - I've misread your comment and thought you're talking about Proxy _Detector_, not Selector. Yes, the Selector would work and I was actually not aware that you can use that. It's definitely a solution to the problem, however I'd still prefer to have a way to explicitly expose some sort of API that would let us set proxy for managed channel. Both proxy detector and proxy selector, aside from the fact that they're global, in my opinion are a little bit magical. What do you think?

carl-mastrangelo:
>  (seems like a slight breach of the single responsibility principle)

@snowp  I actually change my mind, I think the name resolver is the correct place now.  The single responsibility is "convert a string to socket addresses", and that feels like a name resolving thing.   

Can I ask if you have heterogeneous backends for a given target?  For example, if you try to resolve your service, there may be in-datacenter backends to talk to, but also out-of-datacenter backends that need to go through a proxy.   (for example, an encrypting proxy).  If you have mixed proxy results, the NameResolver would be a good place to do proxy resolution, because the results can be included along with the SocketAddresses.

snowp:
> Can I ask if you have heterogeneous backends for a given target?

 No, right now all the targets resolved by our `NameResolver` hit internal datacenters, but I can see the use case of having the proxy information come the resolver (e.g. it might be stored in some service discovery system). 

I imagine having the `NameResolver` resolve proxy information would be sufficient for our use case, as we'd set a global proxy that the `DNSNameResolver` could pick up on and have our own `NameResolver` use no proxy. That said, would this make it difficult to have names resolved by the `DNSNameResolver` use different proxy settings? From what I can tell there's no way to override the behavior of `DNSNameResolver` as it gets loaded by default, but I could be wrong. Maybe it's good enough to be able to use the `ProxySelector` to specify proxy details in this case?

How were you all thinking of having `NameResolver` return the proxy information? Would it make sense to have the eag object also contain proxy information? Or does this belong in the `Attributes` object?

snowp:
Oh, just realized you can specify the `NameResolver` explicitly on the `ManagedChannelBuilder` (and we're actually doing so), so ignore my comment above about the `DnsNameResolver`

core/src/main/java/io/grpc/ForwardingChannelBuilder.java
```diff
@@ -145,6 +146,11 @@ public T maxInboundMessageSize(int max) {
     return thisT();
   }
 
+  @Override public T proxyDetector(ProxyDetector proxyDetector) {
```
ejona86: I don't think we'd want to pass in the ProxyDetector. It should either be a ProxySelector or a Proxy (and I'm not yet sure which; it seems the "power" of ProxySelector doesn't add much value, but still unclear).

ejona86:
> Oh, just realized you can specify the NameResolver explicitly on the ManagedChannelBuilder (and we're actually doing so), so ignore my comment above about the DnsNameResolver

Note that I'd like to remove/de-emphasize that method of setting the NameResolver if possible. So if that's required for a common case, I'd be interested in alternatives.

ejona86:
> Both proxy detector and proxy selector, aside from the fact that they're global, in my opinion are a little bit magical. What do you think?

Yes and no. Proxy configuration tends to be done by the end-user instead of an application or library developer. That means the global approach tends to be about the only thing that would work. If you think ProxySelector is magical, WPAD PAC uses a javascript file that is autodiscovered by the client and executed for each request.

I'm not against being able to manually specify the Proxy, but I would have expected some other use-case. "I need to use a proxy, so I should configure ProxySelector" and then you don't actually care about grpc specifically (it could even be an implementation detail).

snowp:
> Proxy configuration tends to be done by the end-user instead of an application or library developer.

I agree in the general case, but in our case we've got internal library developers who are in charge of configuring our proxies to handle the outbound message sent by their library. There's really no end-user in our scenario, as we're talking about hundreds of services, all of which would have to use the same proxy settings. 

Here's another example of a use case we're considering: Suppose we wanted to add code that adds instrumentation to internal framework code that writes trace information directly to Stackdriver. Being the person writing this code I know what proxy this should go through, so I hard code this into the framework. By doing so, I can easily roll out this to hundreds of apps without the need for updating the proxy rules for the individual app. Having global proxy settings that's shared between all our apps would also solve this, but it seems like a larger effort and we already have apps defining their own `ProxySelector`, so I'd like to avoid the risk of conflicting with that, especially considering that consumers of this shared code is spread between multiple repos. 

Really I think it boils down to us a) not wanting to force everyone to configure their own proxy when it's the same for everyone and b) not wanting to trample on other's config by setting global proxy settings

> Now it sounds like you're arguing that this PR doesn't work, because you couldn't configure the ManagedChannel anyway because it is hidden from the API surface.

It's true that we'd be unable to specify this in the current API for whatever library we're using, but until that gets added we'd most likely just construct our own grpc client instead of using the one provided by the library.

> I don't think we'd want to pass in the ProxyDetector. It should either be a ProxySelector or a Proxy (and I'm not yet sure which; it seems the "power" of ProxySelector doesn't add much value, but still unclear).

Yeah I agree. I think your suggestion of specifying a `Proxy` is more than enough, so I'll push a commit to this PR to use that approach instead.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -95,14 +95,20 @@
...
       Resource<ExecutorService> executorResource,
-      ProxyDetector proxyDetector) {
+      boolean usingProxy) {
     // TODO: if a DNS server is provided as nsAuthority, use it.
     // https://www.captechconsulting.com/blogs/accessing-the-dusty-corners-of-dns-with-java
     this.timerServiceResource = timerServiceResource;
     this.executorResource = executorResource;
+    this.usingProxy = usingProxy;
+
+    String targetPath = Preconditions.checkNotNull(targetUri.getPath(), "targetPath");
```
snowp: the code in this constructor (moved some of it in from the `DnsNameResolverProvider` class) is what i extract into `GrpcUtil#unresolvedAddress` but I ran into some test failures while trying to read the host/authority back from the `InetSocketAddress` so right now it's just kinda duplicated in order to get the tests green. just wanted to get some eyes on this change as a whole before I spent the time to remove this duplication

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -279,6 +280,18 @@ public T maxInboundMessageSize(int max) {
...
   }
 
+  /**
+   * Sets the {@link Proxy} to use when resolving proxy parameters for outgoing
+   * connections.
+   *
+   * @param proxy the proxy to use
+   * @return this
+   */
+  public T proxy(Proxy proxy) {
```
ejona86: `@ExperimentalApi`.

This might be a reasonable case to throw. If someone specifies a SOCKS proxy for instance... Dunno.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -160,9 +165,8 @@ public void run() {
           resolving = true;
         }
         try {
-          InetSocketAddress destination = InetSocketAddress.createUnresolved(host, port);
-          ProxyParameters proxy = proxyDetector.proxyFor(destination);
```
ejona86: We want the name resolver doing the proxy detection. As part of #4029 we'll start allowing the name resolver to return proxy information.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -527,7 +531,9 @@ static NameResolver getNameResolver(String target, NameResolver.Factory nameReso
         // Should not be possible.
         throw new IllegalArgumentException(e);
       }
-      NameResolver resolver = nameResolverFactory.newNameResolver(targetUri, nameResolverParams);
+      NameResolver resolver = nameResolverFactory.newNameResolver(
+          targetUri, nameResolverParams,
+          proxyDetector.proxyFor(GrpcUtil.unresolvedAddress(targetUri, defaultPort)) != null);
```
ejona86: The URIs are NameResolver-specific. `ManagedChannelImpl` can't understand them; it relies entirely on the NameResolver. So this should go back the way it was.

I actually don't quite follow why it was necessary. I would have thought the ManagedChannelBuilder change would have been enough.

core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java
```diff
@@ -188,6 +198,22 @@ private static InetSocketAddress overrideProxy(String proxyHostPort) {
     return InetSocketAddress.createUnresolved(parts[0], port);
   }
 
+  private static Supplier<ProxySelector> proxySelectorSupplier(final Proxy proxy) {
+    return new Supplier<ProxySelector>() {
+      @Override public ProxySelector get() {
+        return new ProxySelector() {
+          @Override public List<Proxy> select(URI uri) {
+            return ImmutableList.of(proxy);
```
ejona86: We avoid immutablelist (and similar collections) to reduce our method count on Android. That may not be necessary any more, but we've not been hurting that badly without them.

ejona86:
> Being the person writing this code I know what proxy this should go through, so I hard code this into the framework. 

That's fair enough. You could also use this to hard-code it to be Proxy.DIRECT in other cases. I'm fine with that.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
ejona86: #4029 is closed, but it seems we don't yet provide the ProxyDetector to the NameResolver. I'm thinking that can be passed in in the `Attributes` passed to `newNameResolver()`.

ejona86:
So this has stalled a bit. @snowp, are there pending questions that we need to answer? Are you still interested in seeing this change happen?

I think there may be a few things to figure out, but it looks like overall this PR will become smaller/simpler.

snowp:
@ejona86 Unfortunately I'm not going to be able to see this one through anytime soon, though I'd still like the functionality. Maybe close the PR and if someone wants to pick it up they can reopen?

ejona86:
@snowp, thanks for the contribution all the same.

Closing at @snowp's suggestion. I feel this PR is actually pretty close and would be pretty small after some of our comments were addressed. If anyone else is interested in picking this up, please do!

Issue #1944, opened 2016-06-16
Implement per-stream compression


ejona86:
This is to replace our current per-message compression for greater compression ratios. I don't think there is a public spec yet.

ejona86:
Per-stream compression was neat, but has been removed from c core due to lack of customer before it ever really rolled out.

PR #3963, opened 2018-01-12
core: support full stream compression

ericgribkoff:
This adds support for compressing outbound streams, and also adds the API methods to enable decompressing inbound streams on the server side. The main changes are in `MessageFramer`; the bulk of the rest of the file modifications are just API glue to support enabling full stream (de)compression. This is a follow-up to https://github.com/grpc/grpc-java/pull/3403

Without enabling any of the new options, this PR does still add `content-encoding: identity` to the server's outbound headers. This is in line with current behavior with `grpc-encoding` (see `ServerCallImpl#sendHeaders`), but can be suppressed if this might pose an integration risk while this feature is still experimental.

edit: https://github.com/grpc/grpc-java/issues/1944 can be closed once this is in

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractClientStream.java
+++ b/core/src/main/java/io/grpc/internal/AbstractClientStream.java
@@ -127,11 +127,6 @@ public abstract class AbstractClientStream extends AbstractStream
   }
 
   @Override
-  public final void setFullStreamDecompression(boolean fullStreamDecompression) {
-    transportState().setFullStreamDecompression(fullStreamDecompression);
-  }
-
-  @Override
   public final void setDecompressorRegistry(DecompressorRegistry decompressorRegistry) {
     transportState().setDecompressorRegistry(decompressorRegistry);
   }
@@ -202,7 +197,6 @@ public abstract class AbstractClientStream extends AbstractStream
     private final StatsTraceContext statsTraceCtx;
     private boolean listenerClosed;
     private ClientStreamListener listener;
-    private boolean fullStreamDecompression;
     private DecompressorRegistry decompressorRegistry = DecompressorRegistry.getDefaultInstance();
 
     private boolean deframerClosed = false;
@@ -220,10 +214,6 @@ public abstract class AbstractClientStream extends AbstractStream
         TransportTracer transportTracer) {
       super(maxMessageSize, statsTraceCtx, transportTracer);
       this.statsTraceCtx = checkNotNull(statsTraceCtx, "statsTraceCtx");
-    }
-
-    private void setFullStreamDecompression(boolean fullStreamDecompression) {
-      this.fullStreamDecompression = fullStreamDecompression;
     }
 
     private void setDecompressorRegistry(DecompressorRegistry decompressorRegistry) {
@@ -262,18 +252,20 @@ public abstract class AbstractClientStream extends AbstractStream
       statsTraceCtx.clientInboundHeaders();
 
       boolean compressedStream = false;
-      String streamEncoding = headers.get(CONTENT_ENCODING_KEY);
-      if (fullStreamDecompression && streamEncoding != null) {
-        if (streamEncoding.equalsIgnoreCase("gzip")) {
-          setFullStreamDecompressor(new GzipInflatingBuffer());
-          compressedStream = true;
-        } else if (!streamEncoding.equalsIgnoreCase("identity")) {
-          deframeFailed(
-              Status.INTERNAL
-                  .withDescription(
-                      String.format("Can't find full stream decompressor for %s", streamEncoding))
-                  .asRuntimeException());
-          return;
+      if (fullStreamDecompression) {
+        String streamEncoding = headers.get(CONTENT_ENCODING_KEY);
+        if (streamEncoding != null) {
+          if (streamEncoding.equalsIgnoreCase("gzip")) {
+            enableFullStreamDecompressor();
+            compressedStream = true;
+          } else if (!streamEncoding.equalsIgnoreCase("identity")) {
+            deframeFailed(
+                Status.INTERNAL
+                    .withDescription(
+                        String.format("Can't find full stream decompressor for %s", streamEncoding))
+                    .asRuntimeException());
+            return;
+          }
         }
       }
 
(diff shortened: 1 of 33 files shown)


ejona86:
We should never send `identity`, even in the old code path, so let's not add it here. Sending `identity` enables new logic in the client even when unnecessary. This caused incompatibility in the past with a buggy version of the OkHttp transport whose compression handling was not properly flag-protected.

core/src/main/java/io/grpc/internal/MessageFramer.java
```diff
@@ -163,18 +170,27 @@ public void writePayload(InputStream message) {
...
   private int writeUncompressed(InputStream message, int messageLength) throws IOException {
     if (messageLength != -1) {
-      currentMessageWireSize = messageLength;
       return writeKnownLengthUncompressed(message, messageLength);
     }
     BufferChainOutputStream bufferChain = new BufferChainOutputStream();
-    int written = writeToOutputStream(message, bufferChain);
+    int written;
+    if (streamCompression) {
+      // "Full stream" compression with unknown length reduces to separate concatenated gzip blocks
```
ejona86: We want to reuse the gzip context, so if we need to "eat" the copy, so be it. It may be possible to avoid the copy by using the deflater directly, but let's not worry about it at this point. We will probably want to use a byte[]-backed buffer chain so that we can read from it, however.

Note that this code doesn't look "safe" as-is because `gzipOutputStream` hasn't been finished and retains its window, which will be wrong for the receiver.

core/src/main/java/io/grpc/internal/MessageFramer.java
```diff
@@ -225,32 +241,65 @@ private int writeKnownLengthUncompressed(InputStream message, int messageLength)
...
+                  .getConstructor(OutputStream.class, boolean.class)
+                  .newInstance(gzipOutputStreamAdapter, true);
+        } catch (Throwable t) {
+          // The two-arg GZIPOutputStream constructor is not available on JDK6
+          gzipOutputStream = new GZIPOutputStream(gzipOutputStreamAdapter);
+          // GZIPOutputStream#flush on JDK6 does not flush the compressor
+          gzipCloseOnFlush = true;
+        }
+      }
+      gzipOutputStream.write(headerScratch);
```
ejona86: Ditto about `messageCompression`.

core/src/main/java/io/grpc/internal/MessageFramer.java
```diff
@@ -225,32 +241,65 @@ private int writeKnownLengthUncompressed(InputStream message, int messageLength)
...
-    return writeToOutputStream(message, outputStreamAdapter);
+    OutputStream outputStreamToWrite;
+    if (streamCompression) {
+      if (gzipOutputStream == null) {
+        try {
+          gzipOutputStream =
+              GZIPOutputStream.class
+                  .getConstructor(OutputStream.class, boolean.class)
+                  .newInstance(gzipOutputStreamAdapter, true);
+        } catch (Throwable t) {
```
ejona86: I'd much prefer we caught more-precise exceptions, or at the least only Exception instead of Throwable.

core/src/main/java/io/grpc/internal/MessageFramer.java
```diff
@@ -296,6 +352,17 @@ private void writeRaw(byte[] b, int off, int len) {
    */
   @Override
   public void flush() {
+    if (gzipOutputStream != null) {
+      if (gzipCloseOnFlush) {
+        closeGzipOutputStream();
+      } else {
+        try {
+          gzipOutputStream.flush();
+        } catch (IOException e) {
```
ejona86: It's weird this exception is "handled" but thrown away. This seems like it should be logged at warning, right?

core/src/main/java/io/grpc/internal/MessageFramer.java
```diff
@@ -105,6 +107,12 @@ public MessageFramer setCompressor(Compressor compressor) {
     return this;
   }
 
+  @Override
+  public Framer setStreamCompression(boolean enable) {
+    streamCompression = enable;
```
ejona86: The stream compression code is tighly integrated and seems error-prone. Would it make more sense to make a "fake" Sink that does the gzipping and then delegates to the real one? We'd also need to fake the WritableBuffer to be byte[]-based.

That would, however, be pretty inefficient and require a lot more copying. If we created a new sink-like API (let's call it "better sink") but that also included a `write(byte[], int, int)`, then it seems we could clean up the code while being efficient. Basically we'd move `buffer`/writeRaw/commitToSink (anything WritableBuffer-related) into the "default" better-sink, so that the code acts more like a stream.

ejona86: Because it is non-obvious, I don't think we'd include anything concerning `messageCompression` in better-sink. Instead, when `messageCompression` changes, we'd call the gzip sink directly (knowing its class, not the interface) to inform it of the new setting. Although that could be included in the interface, it probably doesn't gain much by doing so.

core/src/main/java/io/grpc/internal/AbstractClientStream.java
```diff
@@ -262,18 +252,20 @@ protected void inboundHeadersReceived(Metadata headers) {
...
-                  .asRuntimeException());
-          return;
+      if (fullStreamDecompression) {
+        String streamEncoding = headers.get(CONTENT_ENCODING_KEY);
+        if (streamEncoding != null) {
+          if (streamEncoding.equalsIgnoreCase("gzip")) {
+            enableFullStreamDecompressor();
+            compressedStream = true;
+          } else if (!streamEncoding.equalsIgnoreCase("identity")) {
+            deframeFailed(
```
uweschaefer: i was expecting a lookup in the registry. Am i right to assume that 'gzip' the only supported Codec?

ejona86: Yes. The integration needed here is quite strong, such that we can't use the registry.

uweschaefer: Somewhat sad. Thank you, anyway!

Issue #3966, opened 2018-01-12
Support Conscrypt on Android


ericgribkoff:
This will allow OkHttp to use TLS1.2 even on Android API levels < 21. This was first brought up in https://github.com/grpc/grpc-java/issues/1906 and https://github.com/grpc/grpc-java/pull/3301, which exposed some issues with how gRPC OkHttp picks the security provider.

ericgribkoff:
Closed via #3971

PR #3971, opened 2018-01-15
okhttp: support Conscrypt security provider

ericgribkoff:
This addresses https://github.com/grpc/grpc-java/issues/3966 and incorporates the text changes to `SECURITY.md` from #3301. Tested with `org.conscrypt:conscrypt-android:1.0.0.RC14`, as well as Google Play's security provider.

This changes our detection of the Google Play security provider from using the class name to the provider name, `GmsCore_OpenSSL`. This would seem to be equally as stable, and recent versions of OkHttp do the [check similarly](https://github.com/square/okhttp/blob/1d9c158a96382facc3be672f6b41dc79fb5eb48b/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java#L196), although I realized in testing this that OkHttp uses `GMSCore_OpenSSL`, which did not work in my tests (checking elsewhere, it should indeed be `GmsCore_OpenSSL`, and the lookup is case-sensitive).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -10,11 +10,37 @@ On Android, use the [Play Services Provider](#tls-on-android). For non-Android s
 
 ## TLS on Android
 
-On Android we recommend the use of the [Play Services Dynamic Security Provider](http://appfoundry.be/blog/2014/11/18/Google-Play-Services-Dynamic-Security-Provider) to ensure your application has an up-to-date OpenSSL library with the necessary ciper-suites and a reliable ALPN implementation.
+On Android we recommend the use of the [Play Services Dynamic Security
+Provider](https://www.appfoundry.be/blog/2014/11/18/Google-Play-Services-Dynamic-Security-Provider/)
+to ensure your application has an up-to-date OpenSSL library with the necessary
+ciper-suites and a reliable ALPN implementation. This requires [updating the
+security provider at
+runtime](https://developer.android.com/training/articles/security-gms-provider.html).
 
-You may need to [update the security provider](https://developer.android.com/training/articles/security-gms-provider.html) to enable ALPN support, especially for Android versions < 5.0. If the provider fails to update, ALPN may not work.
+Although ALPN mostly works on newer Android releases (especially since 5.0),
+there are bugs and discovered security vulnerabilities that are only fixed by
+upgrading the security provider. Thus, we recommend using the Play Service
+Dynamic Security Provider for all Android versions.
 
 *Note: The Dynamic Security Provider must be installed **before** creating a gRPC OkHttp channel. gRPC's OkHttpProtocolNegotiator statically initializes the security protocol(s) available to gRPC, which means that changes to the security provider after the first channel is created will not be picked up by gRPC.*
+
+### Bundling Conscrypt
+
+If depending on Play Services is not an option for your app, then you may bundle
+[Conscrypt](https://conscrypt.org) with your application. Binaries are available
+on [Maven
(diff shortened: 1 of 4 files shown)


okhttp/src/test/java/io/grpc/okhttp/OkHttpProtocolNegotiatorTest.java
```diff
@@ -229,6 +230,31 @@ public void pickTlsExtensionType_android41() throws Exception {
...
+          }
+        };
+
+    AndroidNegotiator.TlsExtensionType tlsExtensionType =
+        AndroidNegotiator.pickTlsExtensionType(cl);
+
+    assertEquals(TlsExtensionType.ALPN_AND_NPN, tlsExtensionType);
+
+    // Clean up
+    Security.removeProvider(fakeConscrypt.getName());
```
ejona86: try-finally or similar (could maybe remove it unconditionally in the `@After`).

SECURITY.md
```diff
@@ -10,12 +10,38 @@ On Android, use the [Play Services Provider](#tls-on-android). For non-Android s
...
+
+Like the Play Services Dynamic Security Provider, you must still "install"
+Conscrypt before use.
+
+```java
+import org.conscrypt.Conscrypt;
+import java.security.Security;
+...
+
+Security.addProvider(Conscrypt.newProvider());
```
ejona86: We should prefer insertProviderAt so that Conscrypt can be prioritized.

okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/Platform.java
```diff
@@ -199,11 +210,18 @@ private static Provider getAppEngineProvider() {
...
 
   /**
-   * Select from the available security providers in preference order. If a preferred provider
-   * is not found then warn but continue.
+   * Select from the available security providers in preference order. If a preferred provider is
+   * not found then warn but continue.
    */
   private static Provider getAndroidSecurityProvider() {
-    for (String providerClassName : ANDROID_SECURITY_PROVIDERS) {
+    for (String providerName : ANDROID_SECURITY_PROVIDERS) {
```
ejona86: We should loop through the providers in order and choose the first we can use. That way the user has some way of configuring our behavior, and our behavior will better match expectations.

SECURITY.md
ericgribkoff: Done

okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/Platform.java
ericgribkoff: Done

ericgribkoff:
Moved the logic to decide what TLS extensions are available into `Platform`. This seems like where it belongs, since the `Platform` already decides on the security provider using similar logic. This means that tests of this logic are gone from `OkHttpProtocolNegotiatorTest` (but this maybe isn't so much worse than the status quo since we never had tests for picking the actual provider anyways). I've manually tested the changes on emulators, but this almost certainly deserves some test coverage. I didn't add tests yet, as to-date we don't have any (?) tests of the copied OkHttp classes (and simulating all the different platforms/API levels looks a little tricky), but I can add tests if the approach here looks good.

Worth noting: recent Android builds have AndroidNSSP (Android Network Security Policy) as the #1 security provider. I haven't looked into it, but it might be worthwhile adding this to the list of recognized security providers later.

ericgribkoff:
@ejona86 FYI this was heavily refactored/redone since the earlier approval, so I'm holding off on merging until you have a chance to (re)review

PR #3976, opened 2018-01-17
core: handle long dns txt records properly, parse service config, and add tests

carl-mastrangelo:


carl-mastrangelo:
@zhangkun83 friendly ping

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -67,9 +77,15 @@
   private static final String SERVICE_CONFIG_NAME_PREFIX = "_grpc_config.";
   // From https://github.com/grpc/proposal/blob/master/A5-grpclb-in-dns.md
   private static final String GRPCLB_NAME_PREFIX = "_grpclb._tcp.";
+  // From https://github.com/grpc/proposal/blob/master/A2-service-configs-in-dns.md
+  private static final String SERVICE_CONFIG_PREFIX = "_grpc_config=";
+  private static final Set<String> SERVICE_CONFIG_CHOICE_KEYS =
+      Collections.unmodifiableSet(
+          new HashSet<String>(
+              Arrays.asList("clientLanguage", "percentage", "clientHostname", "serviceConfig")));
```
jyane: Just a question: Why is it not `ImmutableSet`?

carl-mastrangelo: Guava collections adds a large number of methods to our library, which Android users would not like.   Android users have a limit of 64K methods per java program, and Guava takes from 5-10K.   Proguard gets rid of most of them, but not the collections classes.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -194,13 +213,29 @@ public void run() {
...
-                GrpcAttributes.NAME_RESOLVER_ATTR_SERVICE_CONFIG,
-                Collections.unmodifiableList(new ArrayList<String>(resolvedInetAddrs.txtRecords)));
-            */
+            JsonObject serviceConfig = null;
+            try {
+              JsonArray allServiceConfigChoices = parseTxtResults(resolvedInetAddrs.txtRecords);
+              for (JsonElement serviceConfigChoice : allServiceConfigChoices) {
+                try {
+                  serviceConfig = maybeChooseServiceConfig(
+                      serviceConfigChoice.getAsJsonObject(), random, null);
```
zhangkun83: null hostname? It doesn't look right.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
jyane: Thanks!

core/src/main/java/io/grpc/internal/DnsNameResolver.java
carl-mastrangelo: D'oh!  Done, and manually verified.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -95,6 +111,9 @@
   @GuardedBy("this")
   private Listener listener;
 
+  @VisibleForTesting
+  Random random = new Random();
```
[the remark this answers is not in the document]
carl-mastrangelo: Done.  Originally this was going to be a ThreadLocalRandom, but it's not available in Java 6.

carl-mastrangelo:
Failures seem unrelated.

PR #3978, opened 2018-01-17
README.md: include supported Android API levels

ericgribkoff:
We can make this more nuanced (I'd have to test it out, but as far as I remember we work just fine on earlier Android levels if you don't care about TLS) but API level 14 is the minimum required by Google Play Services. Are we okay with setting that as the minimum supported version?


ejona86:
Yes. We've been tracking Google Play Services.

PR #3981, opened 2018-01-17
core: retry part 4, retry attempt header

dapengzhang0:


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -437,12 +440,18 @@ void postCommit() {
         }
 
         @Override
-        ClientStream newStream() {
+        ClientStream newStream(int previousAttempts) {
+          Metadata newHeaders = headers;
+          if (previousAttempts > 0) {
```
ejona86: I'd much rather this logic be in RetriableStream. Could we instead pass method, callOptions, etc (or I guess maybe only headers) to RetriableStream and then have it pass them as arguments here?

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -603,5 +602,7 @@ State committed(Substream winningSubstream) {
 
     // GuardedBy RetriableStream.lock
     boolean closed;
+
+    int previousAttempts;
```
ejona86: Shouldn't this be incremented somewhere? It looks like it's always `0`?

dapengzhang0: It is incremented in `retry(substream.previousAttempts + 1)`. (For hedging, this number in each of the hedged substreams will be set incrementally when RetriableStream starts.) The field `previousAttempts` in `Substream` is effectively final through.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: @zhangkun83 suggested to rename this method to `"newSubstream()"` because we already have `ClientTransport.newStream()` and yet another `RetriableStream.newStream()`, which does not return the same stream, is confusing. If we renamed this method to `"newSubstream()"`, then it would make perfect sense to take `int previousAttempts` as its argument. Otherwise we need to pass the `headers` here to RetriableStream and then pass (modified) `headers` back here to this method.

dapengzhang0: @zpencer may affect/be affected by #3967

zpencer: For #3967: As long as the new code still asks the load balancer picker for a new `PickResult` and calls `GrpcUtil.getTransportFromPickResult` the channelz substream code will work.

core/src/main/java/io/grpc/internal/RetriableStream.java
ericgribkoff: FWIW, I don't think the intent was for hedged RPCs to set the `grpc-previous-rpc-attempts` header.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -99,25 +99,24 @@ private void commitAndRun(Substream winningSubstream) {
...
+    Substream substream = createSubstream(previousAttempts);
     drain(substream);
   }
 
-  private Substream createSubstream() {
+  private Substream createSubstream(int previousAttempts) {
     Substream sub = new Substream();
+    // TODO(zdapeng): transparent-retry-attempts
+    sub.previousAttempts = previousAttempts;
     // NOTICE: This set _must_ be done before stream.start() and it actually is.
```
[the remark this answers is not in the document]
dapengzhang0: It make sense to add constraint on the Javadoc for `RetriableStream#newStream`. Will add it.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: I agree that this logic should not be owned by `ManagedChannelImpl`. Moved it to `RetriableStream`, but I do not pass it as a field because it is highly mutable but should not be mutated and should only be used in one method.

zhangkun83: I would rather mutate the headers in `RetriableStream.retry()` and pass it down to `createSubstream()` then `newSubstream()`, which means `RetriableStream` will keep a reference to the headers, which should be fine. I am not super concerned about `newSubstream()` mutating the headers when it shouldn't, but the calling back and forth (spaghetti-like) between classes does make the code harder to follow and test.

ejona86: > Otherwise we need to pass the headers here to RetriableStream and then pass (modified) headers back here to this method.

This is exactly what I was suggesting. That seems much clearer than the back-and-forth going on now. Moving ownership of the Metadata to RetriableStream seems really easy.

dapengzhang0: PTALA

markdroth:
@ericgribkoff, why would we want to expose the number of attempts to the application?  I think we will want to expose this via the stats we export via census (and as per email discussion, we do still have some work to do on that front), but I don't think we should expose it to the application via our API.  The whole point of retries in gRPC is that it should be transparent to the application -- the application should not have to know or care whether or not the RPC was retried.

In any case, I don't believe that the retry design says anything about exposing this to the application, and I think we would need to add that to the design (with appropriate design review) if this was something that we wanted to do.

dapengzhang0:
Rebased.

ericgribkoff:
@markdroth 

From https://github.com/grpc/proposal/blob/master/A6-client-retries.md#exposed-retry-metadata:

> Likewise, the client will receive the number of retry attempts made when receiving the results of an RPC.

One use case: mobile clients (and possibly others) care about number of network requests and number of failures. Right now they can, and do, easily capture this information for gRPC at the application level and record it. If we don't expose the number of retry attempts to the client when returning a RPC, apps will have no way of monitoring these now-hidden RPC failures.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -146,15 +151,31 @@ public ClientStreamTracer newClientStreamTracer(CallOptions callOptions, Metadat
...
-    sub.stream = newStream(tracerFactory);
+    sub.stream = newSubstream(tracerFactory, newHeaders);
     return sub;
   }
 
   /**
-   * Creates a new physical ClientStream that represents a retry/hedging attempt.
+   * Creates a new physical ClientStream that represents a retry/hedging attempt. The returned
+   * Client stream must not be started already.
```
ejona86: s/must/will/

ejona86: Alternatively:
The returned client stream is not yet _(or already)_ started.

markdroth:
> From https://github.com/grpc/proposal/blob/master/A6-client-retries.md#exposed-retry-metadata:
> 
> > Likewise, the client will receive the number of retry attempts made when receiving the results of an RPC.
> 

Wow, that wording is really unclear and seems almost hidden in the text -- I completely missed that in my previous reads of the design.

I'll start an email thread where we can discuss this.

PR #3983, opened 2018-01-18
core: Remove unused ServiceConfiguration

ejona86:
It was going to be used, but then the design changed. It's now
unnecessary.


dapengzhang0:
@ejona86 any other reason than just being unused to remove this class? I may want to use this class for parsing retry policy, other wise I need to parse retry policy from service config json object step by step.

ejona86:
I had spoken to Carl and he said that it was no longer planned to be used. You shouldn't need to parse something for the retry policy; it'd be parsed before it gets to the retry pieces.

PR #3984, opened 2018-01-18
buildscripts: Add Kokoro-based CI for Android APK stats

ericgribkoff:
~~Internal change is cl/182328214.~~

~~The first commit just pulls in some prerequisite build updates from #3975 to allow the gradle dexcount and apksize plugins to work.~~

Edit: now incorporating the APK stat diffs into the existing cronet kokoro job, see comment below.


buildscripts/kokoro/android-apk-stats.sh
```diff
@@ -0,0 +1,50 @@
...
+fields_delta=$((new_fields-old_fields))
+
+# Kororo URLs are in the form "grpc/job/macos/job/master/job/grpc_build_artifacts"
+KOKORO_JOB_PATH=$(echo "${KOKORO_JOB_NAME}" | sed "s|/|/job/|g")
+
+cd ../../../
+mkdir -p reports
+echo '<html><head></head><body>' > reports/kokoro_index.html
+echo '<h1>'${KOKORO_JOB_NAME}', build '#${KOKORO_BUILD_NUMBER}'</h1>' >> reports/kokoro_index.html
+echo '<h2><a href="https://kokoro2.corp.google.com/job/'${KOKORO_JOB_PATH}'/'${KOKORO_BUILD_NUMBER}'/">Kokoro build dashboard (internal only)</a></h2>' >> reports/kokoro_index.html
```
matt-kwong: Just a heads up: we might be moved from kokoro2 in the future. Just saying so that if this link stops working, you'll know the reason this link broke the last time for us last time.

buildscripts/kokoro/android-apk-stats.sh
```diff
@@ -0,0 +1,50 @@
...
+echo '<h2><a href="https://sponge.corp.google.com/invocation?id='${KOKORO_BUILD_ID}'&searchFor=">Test result dashboard (internal only)</a></h2>' >> reports/kokoro_index.html
```
matt-kwong: This will likely also be switched to a result storage site that is publicly accessible.

ericgribkoff:
@ejona86 This change is now redone as an addition to our existing Cronet Kokoro build (which I'll rename to simply `android.{cfg,sh}` in this PR and internally once this is approved, as it now includes our Android examples as well). The APK size and dex count deltas are reported as Github statuses as part of the Android build process.

The second and third commits here can be ignored and will be dropped before merging. The second commit pulls in #4114 to fix the Android example builds, and the third is an artificial change to core that increases the APK size and dex count to demo the new status messages.

buildscripts/kokoro/cronet.sh
```diff
@@ -3,6 +3,119 @@
...
+../../../buildscripts/set_github_status.py \
+  --sha1 $KOKORO_GITHUB_PULL_REQUEST_COMMIT \
+  --state success \
+  --description "New DEX reference count: $new_dex_count (delta: $dex_count_delta)" \
+  --context android/dex_diff --oauth_file ~/oauth_token.txt
+
+../../../buildscripts/set_github_status.py \
+  --sha1 $KOKORO_GITHUB_PULL_REQUEST_COMMIT \
+  --state success \
+  --description "New APK size in bytes: $new_apk_size (delta: $apk_size_delta)" \
```
ejona86: Could we make these easier to read? Probably `$(printf "%'d" "$new_apk_size")` to add thousands separator would be enough.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -105,6 +106,8 @@
   static final Status SUBCHANNEL_SHUTDOWN_STATUS =
       Status.UNAVAILABLE.withDescription("Subchannel shutdown invoked");
 
+  private final Gson gson = new Gson();
```
ejona86: Revert?

buildscripts/kokoro/cronet.sh
```diff
@@ -3,6 +3,119 @@
 set -exu -o pipefail
 cat /VERSION
 
-cd ./github/grpc-java/cronet
+BASE_DIR=$(pwd)
+
+# Set up APK size and dex count statuses
+
+gsutil cp gs://grpc-testing-secrets/github_credentials/oauth_token.txt ~/
```
ejona86: It seems the code needs to handle KOKORO_GITHUB_PULL_REQUEST_COMMIT being unset. This script can be run on master/release branches.

buildscripts/kokoro/cronet.sh
```diff
@@ -3,6 +3,119 @@
 set -exu -o pipefail
 cat /VERSION
 
-cd ./github/grpc-java/cronet
+BASE_DIR=$(pwd)
```
ejona86: Quote all variables when used. Quote commands (like `"$(pwd)"`).

buildscripts/kokoro/cronet.sh
```diff
@@ -3,6 +3,119 @@
...
+
+# Set up APK size and dex count statuses
+
+gsutil cp gs://grpc-testing-secrets/github_credentials/oauth_token.txt ~/
+
+function set_status_to_fail_on_error {
+  cd $BASE_DIR/github/grpc-java
+
+  # TODO(ericgribkoff) Remove once merged
+  git checkout $KOKORO_GITHUB_PULL_REQUEST_COMMIT
```
ejona86: This seems unfortunate. I don't think it's something I'd want us to remove (rather, I want us to always use the most recent version of the script.) Maybe there's an alternative, like copying the script some place else...

buildscripts/kokoro/cronet.sh
```diff
@@ -3,6 +3,119 @@
...
+read -r ignored new_dex_count < \
+  <(${ANDROID_HOME}/tools/bin/apkanalyzer dex references app/build/outputs/apk/release/app-release-unsigned.apk)
+
+new_apk_size=$(stat --printf=%s app/build/outputs/apk/release/app-release-unsigned.apk)
+
+
+# Get the APK size and dex count stats using the target branch
+
+sudo apt-get install -y jq
+target_branch=$(curl -s https://api.github.com/repos/grpc/grpc-java/pulls/$KOKORO_GITHUB_PULL_REQUEST_NUMBER | jq -r .base.ref)
```
ejona86: Parsing JSON... that's unfortunate. I wonder if we could look at the merge commit parents. Also, it should probably be `.base.sha` since the ref may be different from when the clone was done.

ejona86: We want to compare against the base _this commit_ was built from. That is fixed at the start of the build. However, master can change over time, even during our build. If there is a merge commit, we could do `git rev-parse HEAD^` (or more unlikely `HEAD^2`, depending on how the merge was done). If there's no merge commit... maybe we could use `git merge-base`... But I sort of expect to see merge commits.

ericgribkoff: Done. I was assuming that the base branch (e.g., master) was already in the local repo at the start of the build. That is not always the case, so if the checkout ends up pulling from the remote repository, using the base branch name could indeed end up pulling in changes that didn't exist when the build began.

It turns out that Github already creates the merge commit for you, and that can be used to build and test the PR: it's not quite documentation, but it appears to match reality: https://stackoverflow.com/questions/42626068/how-do-i-list-all-remote-refs

As far I can tell there will never be more than two parents of the merge commit, so we don't need to worry about `HEAD^2`.

cc @matt-kwong The `curl -s https://api.github.com/repos/grpc/grpc-java/pulls/$KOKORO_GITHUB_PULL_REQUEST_NUMBER | jq -r .base.ref` approach was copied from what grpc/grpc was doing. It looks like there might be a simpler and more correct alternative in `git rev-parse`  (Also, it is possible to change the target branch for a PR mid-build via Github's web UI, so the value being parsed from the JSON here could be different than what was used for the merge commit)

PR #3986, opened 2018-01-20
core: retry part 5, add RetryPolicy data object

dapengzhang0:
The logic of using retry policy to compute retry plan will be in one of upcoming PRs.


dapengzhang0:
Rebased.

thelinuxfoundation:

_Regards,
CLA GitHub bot_

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -777,4 +785,58 @@ public long addAndGet(long newBytesUsed) {
...
+    RetryPolicy(
+        int maxAttempts, double initialBackoffInSeconds, double maxBackoffInSeconds,
+        double backoffMultiplier, Collection<Status.Code> retryableStatusCodes) {
+      checkArgument(maxAttempts >= 1, "maxAttempts");
+      this.maxAttempts = maxAttempts;
+      checkArgument(initialBackoffInSeconds >= 0D, "initialBackoffInSeconds");
+      this.initialBackoffInSeconds = initialBackoffInSeconds;
+      checkArgument(
+          maxBackoffInSeconds >= initialBackoffInSeconds,
+          "maxBackoofInSeconds should be at least initialBackoffInSeconds");
```
ericgribkoff: s/Backoof/Backoff

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -777,4 +785,58 @@ public long addAndGet(long newBytesUsed) {
...
+          && Double.compare(initialBackoffInSeconds, that.initialBackoffInSeconds) == 0
+          && Double.compare(maxBackoffInSeconds, that.maxBackoffInSeconds) == 0
+          && Objects.equal(retryableStatusCodes, that.retryableStatusCodes);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hashCode(
+          maxAttempts, initialBackoffInSeconds, maxBackoffInSeconds, backoffMultiplier,
+                              retryableStatusCodes);
```
ericgribkoff: Indentation seems off here?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1068,6 +1096,104 @@ public void run() {
...
+        }
+      }
+    }
+
+    return new RetryPolicies() {
+      @Override
+      public RetryPolicy get(MethodDescriptor<?, ?> method) {
+        RetryPolicy retryPolicy = fullMethodNameMap.get(method.getFullMethodName());
+        if (retryPolicy == null) {
+          retryPolicy = serviceNameMap
```
ericgribkoff: Add test coverage for this fallback logic?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1068,6 +1096,104 @@ public void run() {
     }
   }
 
+  private static RetryPolicies getRetryPolicies(Attributes config) {
+    JsonObject serviceConfig = config.get(GrpcAttributes.NAME_RESOLVER_ATTR_SERVICE_CONFIG);
+    return getRetryPolicies(serviceConfig);
+  }
+
+  // TODO(zdapeng): maybe move it to a more appropriate place
```
ericgribkoff: +1 (but optional, because I'm not sure where would be more appropriate, but it seems like it might be nice to keep the json parsing in a separate location)

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1068,6 +1096,104 @@ public void run() {
...
+
+      if (methodConfigs != null) {
+        for (JsonElement methodConfig : methodConfigs) {
+          JsonArray names = methodConfig.getAsJsonObject().getAsJsonArray("name");
+          JsonObject retryPolicy = methodConfig.getAsJsonObject().getAsJsonObject("retryPolicy");
+          if (retryPolicy != null) {
+            int maxAttempts = retryPolicy.getAsJsonPrimitive("maxAttempts").getAsInt();
+            String initialBackoffStr =
+                retryPolicy.getAsJsonPrimitive("initialBackoff").getAsString();
+            checkState(
```
ericgribkoff: These should be checkArgument(), shouldn't they?

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -777,4 +785,58 @@ public long addAndGet(long newBytesUsed) {
       return bufferUsed.addAndGet(newBytesUsed);
     }
   }
+
+  @Immutable
+  static final class RetryPolicy {
+    private final int maxAttempts;
+    private final double initialBackoffInSeconds;
+    private final double maxBackoffInSeconds;
+    private final double backoffMultiplier;
```
ericgribkoff: backoffMultiplier should be an int

dapengzhang0: That surprises me. Why I can not set backoffMultiplier = 1.5 if I consider 2 is too large?

ericgribkoff: The schema from the design (and quoted in this PR in the comment block in `ManagedChannelImpl#getRetryPolicies`) says: `"backoffMultiplier": number`. I'm not aware of any strong constraint that the design couldn't be changed to allow floating point numbers here, but neither did a strong (or any) need for this come up during the design phase, so it's currently specified as an integer.

ericgribkoff: Nevermind, I was incorrect - I hadn't realized json numbers can be either float or int. We're using the same json type for maxAttempts (which must be an integer) as for backoffMultiplier, but in the proto representation backoffMultiplier can indeed be a floating point number.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: My understanding is `checkArgument` (`/** Ensures the truth of an expression involving one or more parameters to the calling method */`) should check the argument (here is `serviceConfig`) directly in the expression, but here we check `initialBackoffStr` which is an intermediate object so `checkState` is more suitable. I'm not sure my understanding is correct through.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: I think this test code covers it

```java
assertEquals(
          new RetryPolicy(
              3, parseDouble("2.1"), parseDouble("2.2"), parseDouble("3"),
              Arrays.asList(Code.UNAVAILABLE, Code.RESOURCE_EXHAUSTED)),
          retryPolicies.get(builder.setFullMethodName("SimpleService1/not_exist").build()));
```

core/src/main/java/io/grpc/internal/RetriableStream.java
dapengzhang0: Done

core/src/main/java/io/grpc/internal/RetriableStream.java

dapengzhang0:
Moved out the gson logic in ManagedChannelImpl and leave it unimplemented for the moment. Moved the gson logic into test because in test it will still be helpful when I have a valid RetryPolicy.

core/src/test/java/io/grpc/internal/RetryPolicyTest.java
```diff
@@ -0,0 +1,188 @@
...
+                maxBackoffStr.charAt(maxBackoffStr.length() - 1) == 's',
+                "invalid value of maxBackoff");
+            double maxBackoff =
+                Double.parseDouble(maxBackoffStr.substring(0, maxBackoffStr.length() - 1));
+            double backoffMultiplier =
+                retryPolicy.getAsJsonPrimitive("backoffMultiplier").getAsDouble();
+            JsonArray retryableStatusCodes = retryPolicy.getAsJsonArray("retryableStatusCodes");
+            Set<Code> codeSet = new HashSet<Code>(retryableStatusCodes.size());
+            for (JsonElement retryableStatusCode : retryableStatusCodes) {
+              codeSet.add(Code.valueOf(retryableStatusCode.getAsString()));
```
ejona86: Should unknown statuses be ignored? I think so. That should probably be explicitly stated in the spec.

dapengzhang0: + @ericgribkoff if the spec need be updated

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -80,14 +86,16 @@
...
+      Executor callExecutor, ScheduledExecutorService scheduledExecutorService,
+      RetryPolicy retryPolicy) {
     this.method = method;
     this.channelBufferUsed = channelBufferUsed;
     this.perRpcBufferLimit = perRpcBufferLimit;
     this.channelBufferLimit = channelBufferLimit;
     this.callExecutor = callExecutor;
     this.scheduledExecutorService = scheduledExecutorService;
     this.headers = headers;
+    this.retryPolicy = checkNotNull(retryPolicy);
```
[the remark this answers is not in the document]

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -777,4 +785,57 @@ public long addAndGet(long newBytesUsed) {
...
+      this.maxBackoffInSeconds = maxBackoffInSeconds;
+      checkArgument(backoffMultiplier > 0D, "backoffMultiplier");
+      this.backoffMultiplier = backoffMultiplier;
+      this.retryableStatusCodes = Collections.unmodifiableSet(
+          new HashSet<Status.Code>(checkNotNull(retryableStatusCodes, "retryableStatusCodes")));
+    }
+
+    /** No retry. */
+    static final RetryPolicy DEFAULT =
+        new RetryPolicy(1, 0, 0, 1, Collections.<Code>emptyList());
```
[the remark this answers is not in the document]

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -777,4 +785,57 @@ public long addAndGet(long newBytesUsed) {
...
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
```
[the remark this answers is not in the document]

PR #3988, opened 2018-01-22
Gradle 4.4.1

fkorotkov:
Plus configured Gradle to run builds in parallel, use Build Cache and have lazy configuration for faster startup.

All of the above seems to speedup Travis CI builds from 13-15 mins to 8-10 minutes.


thelinuxfoundation:

* User @fkorotkov isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

fkorotkov:
@carl-mastrangelo thanks! updated examples too 👍

.gitignore
```diff
@@ -1,6 +1,5 @@
 # Gradle
 build
-gradle.properties
```
carl-mastrangelo: @ejona86 should this be deleted?

fkorotkov: I've also added `gradle.properties` with configs to enable Build Cache, parallel execution and on demand configuration (aka lazy configuration).

ejona86: That should normally be user-specific, not something that comes from the repository. So we'll want to do something different.

gradle.properties
```diff
@@ -0,0 +1,5 @@
+org.gradle.daemon=true
+org.gradle.caching=true
```
ejona86: This doesn't seem like it would help us at all. If it helps you, add it to your user-specific ~/.gradle/gradle.properties.

gradle.properties
```diff
@@ -0,0 +1,5 @@
...
+org.gradle.parallel=true
+org.gradle.configureondemand=true
+org.gradle.jvmargs=-Dfile.encoding=UTF-8
```
ejona86: I don't think we want to do this here. What is this trying to fix?

gradle.properties
```diff
@@ -0,0 +1,5 @@
+org.gradle.daemon=true
```
ejona86: I thought daemon was already the default. It seems to be the default on my machine.

gradle.properties
```diff
@@ -0,0 +1,5 @@
...
+org.gradle.caching=true
+org.gradle.parallel=true
```
ejona86: Since parallel is incubating, we let people opt-in to it still. If we want to enable it on Travis we can include it in the .travis.yml. Earlier it wasn't enabled because it was incubating and we weren't seeing much benefit. However, we've been using parallel without issue for quite a while, are now using the sudo: false infrastructure, and do different things in the build. So if it is helpful now, that's fine.

gradle.properties
```diff
@@ -0,0 +1,5 @@
...
+org.gradle.configureondemand=true
```
ejona86: This seems like another one that is fine for your own ~/.gradle/gradle.properties. It's incubating and it is unlikely to help much when daemon's already enabled. But if it does help Travis, we can enable it in .travis.yml.

fkorotkov:
@ejona86 thank you for your feedback! Reverted changes and tweaked `.travis.yml`.

fkorotkov:
Gradle 4.5 is out. I'll close this PR, do better performance testing to figure out effects of different properties and will open a new PR with all the details.

Issue #3991, opened 2018-01-23
When configuring TLS, getting vague error "Couldn't find the address for the requested channel" 


nddipiazza:
I've got a TLS enabled grpc server that won't let me connect to it and I get a bland error: `io.grpc.StatusRuntimeException: UNAVAILABLE: Channel closed while performing protocol negotiation`

https://github.com/nddipiazza/grpc-java/tree/Add-Tls-HelloWorld

### Tested from OS 

`Linux mate 4.13.0-26-generic #29~16.04.2-Ubuntu SMP Tue Jan 9 22:00:44 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`

### gRPC version

Issue happens on `master` branch.

### OpenSSL version

`OpenSSL 1.0.2g  1 Mar 2016`

### Reproducer

I created a reproducer on https://github.com/nddipiazza/grpc-java/tree/Add-Tls-HelloWorld 

I am using the server certs from `grpc-java/testing/src/main/resources/certs`

## Steps to reproduce

 1. `git clone https://github.com/nddipiazza/grpc-java $GRPC_JAVA_DIR`
 1. `cd $GRPC_JAVA_DIR; git checkout Add-Tls-HelloWorld`
 1. `cd $GRPC_JAVA_DIR/examples; ../gradlew -PskipCodegen=true installDist`
 1. In a new terminal open `$GRPC_JAVA_DIR/examples/build/install/examples/bin/hello-world-server-tls`
 1. In a new terminal open `$GRPC_JAVA_DIR/examples/build/install/examples/bin/hello-world-client-tls`

 4. Check the SSL configuration using: `openssl s_client -showcerts -connect localhost:50051`

```
openssl s_client -showcerts -connect localhost:50051
CONNECTED(00000003)
depth=0 CN = localhost
verify error:num=18:self signed certificate
verify return:1
depth=0 CN = localhost
verify return:1
---
Certificate chain
 0 s:/CN=localhost
   i:/CN=localhost
-----BEGIN CERTIFICATE-----
MIIEnDCCAoQCAQEwDQYJKoZIhvcNAQELBQAwFDESMBAGA1UEAwwJbG9jYWxob3N0
MB4XDTE4MDEyMzE2MjkzN1oXDTE5MDEyMzE2MjkzN1owFDESMBAGA1UEAwwJbG9j
[...]

PR #3997, opened 2018-01-25
 netty: only add gRPC negotiator once SSL is established

carl-mastrangelo:
Updates #3991


carl-mastrangelo:
I confirmed manually that this makes the correct error show up.

nddipiazza:
I confirmed as well. With the addition of this change, I can now see the real SSL exception.

PR #4003, opened 2018-01-25
core: disable JNDI, as service config is not ready

ejona86:


ejona86:
> I think this shold be a a flag to disable just the TXT records, because this will disable SRV records too.

That's fine, but it seems you should write it. This flag was just enabled and it isn't safe to enable. Instead of reverting the entire commit, I'm just reverting the one problematic change.

ejona86:
Aside: I'd want to talk with Kun to understand if the GRPCLB stuff is actually ready before enabling it. It seems there was stuff being fixed in it still recently.

ejona86:
This is obsolete, as the code was reverted due to JSON complications (and has since been added back, but it's disabled already).

Issue #3624, opened 2017-10-26
Make GrpcServerRule more customizable


zhangkun83:
Currently it's not possible to customize the in-process server before it's built, e.g., adding `ServerTransportFilter` or `ServerStreamTracer` factory.

PR #1953, opened 2016-06-20
POC: inprocess: Support anonymous servers

ejona86:
Anonymous servers are very convenient for testing. They still require
shutdown, but do not require a unique name.

Note that InProcessInternalServer is just a rename of InProcessServer.

The API presented in up for debate. Usage would look like:

```
InProcessServer server = InProcessServerBuilder.newAnonymous().build().start();
ManagedChannel channel = server.newChannelBuilder().build();
```

The main alternative I considered is to add a ServerImpl.getInternalServer() method for the channel builder to use:

```
Server server = InProcessServerBuilder.newAnonymous().build().start();
// Throws if Server is not in-process. Note that no "cloning" is going on
ManagedChannel channel = ManagedChannelBuilder.forServer(server).build();
```

cc @jhump

PR #4012, opened 2018-01-28
Add server and client builder customization to GrpcServerRule

rmichela:
Implements #3624

* Add `GrpcServerRule.withServerBuilderSetup()` method.
* Add `GrpcServerRule.withChannelBuilderSetup()` method.

### DIFF
--- a/testing/src/main/java/io/grpc/testing/GrpcServerRule.java
+++ b/testing/src/main/java/io/grpc/testing/GrpcServerRule.java
@@ -16,6 +16,7 @@
 
 package io.grpc.testing;
 
+import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 
 import io.grpc.BindableService;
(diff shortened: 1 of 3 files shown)


testing/src/main/java/io/grpc/testing/GrpcServerRule.java
```diff
@@ -59,6 +79,30 @@ public final GrpcServerRule directExecutor() {
...
+  }
+
+  /**
+   * Returns {@code this} configured to apply a custom {@link ChannelBuilderSetup} configuration
+   * each time a test is run. This can only be called at the rule instantiation.
+   */
+  public final GrpcServerRule withChannelBuilderSetup(ChannelBuilderSetup channelBuilderSetup) {
+    checkNotNull(channelBuilderSetup, "channelBuilderSetup");
+    checkState(serverName == null, "withChannelBuilderSetup() can only be called at the rule "
+        + "instantiation");
```
dapengzhang0: reformat like

```java
checkState(
    serverName == null,
    "withChannelBuilderSetup() can only be called at the rule instantiation");
```

And ditto.

rmichela: Fixed formatting.

rmichela:
Fixed formatting and commit message per CONTRIBUTING.md

rmichela:
Does anything else need to be done to merge this?

dapengzhang0:
@rmichela thank you for your PR. We want to be prudent designing the API, once APIs are added, it's impossible or hard to change them. @ejona86 and I were considering making the rule even more customizable, say server/channel builder can be configured inside each test method, like

```java
@Test
public void testFoo() throws Exception {
  // create a server that will be automatically shutdown by the rule.
  Server server = rule.newServerBuilder().someBuilderOptions(someArgs).build().start();
  ...
}
```

Making this work or work better might need implementation of anonymous server (#1953). We still need some time to decide which way will be better.

dapengzhang0:
@rmichela we introduced `GrpcCleanupRule` #4394 with the flexibility: allow configuring inside individual test methods after setup complete, allow multiple channels, server needn't start up before channel creation, support not only in-process transports. Closing this PR in favor of the other. Thanks anyway for your effort on the issue.

PR #4018, opened 2018-01-30
Binary log provider cleanup

zpencer:
The only behavioral change is that BinaryLogProvider can now return
a null value, rather than returning a noop implementation.
 
   - BinaryLogProvider now uses ServiceProviders instead of rolling its
      own service provider code
    - BinaryLogProvider only provides BinaryLog objs
    - BinaryLog is an abstract class for plugging in binlog functionality
    - BinaryLogImpl is our default implementation of binlog, it works by
      returning BinaryLogInterceptor objs
    - BinaryLogInterceptor is renamed from what was
      io.grpc.services.BinaryLog.

This PR depends on https://github.com/grpc/grpc-java/pull/3917


zpencer:
I will repost this as several smaller PRs, since I will need to resolve the conflicts here anyways.

PR #4020, opened 2018-01-30
core: use JsonReader to parse service config

ericgribkoff:
This PR addresses the increase in method count on Android caused by the Gson dependency introduced in https://github.com/grpc/grpc-java/commit/b01609572ab4f0dba6eae23fe5be05a302ecefc7 and https://github.com/grpc/grpc-java/commit/98aa69af7295df22fde75dbfbf05728f0562200e (pending reversion for both PRs in https://github.com/grpc/grpc-java/pull/4015)

Rather than pass the service config as a Gson `JsonObject`, this PR uses gson's `JsonReader` to parse the service config into a standard Java `Map`. 

This approach requires a 95% smaller increase in gRPC's method count to parse the service config json. After Proguard, gRPC's Hello World app currently pulls in just under 3000 methods from the `io.grpc` and `com.google.protobuf` dependencies combined. This PR will add 40 total methods from the `com.google.json` package to handle the parsing of service config json, for a ~1% increase in total method count. In contrast, the use of `JsonObject` would require the app to pull in an additional 760 methods from `com.google.gson`, representing a method count increase for users of ~25%.

There is a small savings in APK size as well, although method count was the dominant concern: the APK size increases by 8Kb from the Gson dependency introduced here, versus 42Kb if using `JsonObject`.

The approach in this PR will currently require (lots of) casts to use the resulting `Map`. Considering that right now there is only the one usage of the service config, which only needs to fetch a single string for the `loadBalancingPolicy`, this seems like an acceptable cost. We can avoid using casts in the future by parsing directly into the necessary service config objects rather than a generic Map, but this can wait until we have non-trivial uses of the service config. The current Map does not insure immutability of the underlying data, which is another item to consider.

This PR includes the test retry service config JSON from #3986 as a test case for the json parser.


core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -501,86 +502,110 @@ ResolutionResults resolve(String host) throws NamingException {
...
+        reader.beginObject();
+        while (reader.hasNext()) {
+          object.put(reader.nextName(), readJsonElement(reader));
         }
-      }
+        reader.endObject();
+        return object;
+      case NUMBER:
+        // nextString() will return the string representation of the number
+        return reader.nextString();
```
ejona86: Why not `nextDouble()`?

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -501,86 +502,110 @@ ResolutionResults resolve(String host) throws NamingException {
...
-        for (JsonElement clientLanguage : clientLanguages) {
-          String lang = clientLanguage.getAsString().toLowerCase(Locale.ROOT);
-          if ("java".equals(lang)) {
-            javaPresent = true;
-            break;
+  @SuppressWarnings("unchecked")
+  static Map<String, Object> maybeChooseServiceConfig(
+      JsonReader reader, Random random, String hostname) throws ClassCastException, IOException {
+    Map<String, Object> serviceConfig = null;
+    reader.beginObject();
```
ejona86: I'd really rather we didn't mix parsing with business logic. Can we just `readJsonElement()` and the process the resulting object? Then we can use things like `get("clientLanguage")`.

core/src/test/resources/io/grpc/internal/test_retry_service_config.json
```diff
@@ -0,0 +1,53 @@
+{
```
dapengzhang0: It's better to use different files (at least copies with different file names) for general service config test vs retry test, because future changes on the data for one of the tests won't affect the other.

ericgribkoff:
Closing, superseded by https://github.com/grpc/grpc-java/pull/4186

PR #4027, opened 2018-01-31
core: Resolve address of GRPC_PROXY_EXP hostname

snowp:
Since this address is never resolved gRPC fails to connect with an
UnresolvedAddressException when the env variable is specified. This
should resolve the address before we attempt to proxy to it.

I know this env variable is scheduled to be removed, so if that's happening in the next version I guess this fix probably isn't necessary.

Also not sure whether there's a good reason why we don't make a DNS lookup here (c.f. comment), but I couldn't find anywhere down the pipeline where this would otherwise get resolved.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java
+++ b/core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java
@@ -184,8 +184,7 @@ class ProxyDetectorImpl implements ProxyDetector {
             + "be removed in a future release. Use the JVM flags "
             + "\"-Dhttps.proxyHost=HOST -Dhttps.proxyPort=PORT\" to set the https proxy for "
             + "this JVM.");
-    // Return an unresolved InetSocketAddress to avoid DNS lookup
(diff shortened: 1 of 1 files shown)


thelinuxfoundation:

* User @snowp isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

snowp:
CLA signed

ejona86:
CC @ericgribkoff

ejona86:
As discussed elsewhere, we really don't want to do DNS lookups here, but we were before and without it things are broken. The real solution will need to plumb the proxy address through the name resolver. That is being tracked by #4029.

ejona86:
Oh, and to be clear why we don't want to do DNS lookups here: it is triggered during static initialization. The code before ProxySelector resolved DNS when the builder was built, which was also sub-optimal.

ejona86:
For posterity, this reverts part of #3789.

Issue #4016, opened 2018-01-30
1.9.0 has different error than 1.8.0 for revoked client cert


pgr0ss:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?

1.9.0

### What did you expect to see?

We have a test which asserts that we get an error when trying to use a revoked client cert. On 1.8.0, we get a `StatusRuntimeException` on the call, and this exception has a cause with a nice message. Our test looks like this:

```java
  @Test
  public void failsWithRevokedClient() throws SSLException {
    try {
      callWithClientCert("revoked.client.crt", "revoked.client.key.pkcs8");
      fail();
    } catch (StatusRuntimeException e) {
      assertEquals(Status.Code.UNAVAILABLE, e.getStatus().getCode());
      assertThat(e.getCause(), instanceOf(SSLHandshakeException.class));
      assertThat(e.getCause().getMessage(), new StringContains("CERTIFICATE_REVOKED"));
    }
  }
```

This works on 1.8.0, but now on 1.9.0, the `e.getCause()` returns `null`. So the request still fails, but we can't assert that it's failing for the right reasons.

Did something change around revoked certs or error handling in 1.9.0?

I didn't see any tests in grpc-java that seem to test an explicitly revoked client certificate. I did see this message for untrusted certs:

https://github.com/grpc/grpc-java/blob/ff0ad5fac3cf944a5ce5fb3ccef7df4a2d0bdf14/interop-testing/src/test/java/io/grpc/testing/integration/TlsTest.java#L187-L188

ejona86:
Can you paste the errors you get in each version?

ejona86:
Interesting. This may have been fixed by #3997. That'd be interesting because if so, this is likely a regression in Netty.

@pgr0ss, is it easy for you to test against master? I just created a fresh 1.10.0-SNAPSHOT in https://oss.sonatype.org/content/repositories/snapshots/ if that helps.

pgr0ss:
@ejona86 I was able to test against master and it does seem like the cause is back:

```
io.grpc.StatusRuntimeException: UNAVAILABLE: io exception                                                         
javax.net.ssl.SSLHandshakeException: error:10000414:SSL routines:OPENSSL_internal:SSLV3_ALERT_CERTIFICATE_REVOKED 
```

The `StatusRuntimeException` has a new piece of the message (`: io exception`) but that seems totally fine.

Any chance this fix can be backported to a 1.9.1? Thanks!

[...]

PR #4030, opened 2018-02-01
netty: Add test to verify error for TLS failure

ejona86:
This is to notice regressions like in #4016

### DIFF
--- a/netty/src/test/java/io/grpc/netty/NettyClientTransportTest.java
+++ b/netty/src/test/java/io/grpc/netty/NettyClientTransportTest.java
@@ -17,6 +17,7 @@
 package io.grpc.netty;
 
 import static com.google.common.base.Charsets.UTF_8;
+import static com.google.common.truth.Truth.assertThat;
 import static io.grpc.internal.GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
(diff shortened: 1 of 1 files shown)


netty/src/test/java/io/grpc/netty/NettyClientTransportTest.java
```diff
@@ -269,6 +272,42 @@ public void run() {
...
+
+    Rpc rpc = new Rpc(transport).halfClose();
+    try {
+      rpc.waitForClose();
+      fail("expected exception");
+    } catch (ExecutionException ex) {
+      StatusException sre = (StatusException) ex.getCause();
+      assertEquals(Status.Code.UNAVAILABLE, sre.getStatus().getCode());
+      assertThat(sre.getCause()).isInstanceOf(SSLHandshakeException.class);
+      assertThat(sre.getCause().getMessage()).contains("SSLV3_ALERT_HANDSHAKE_FAILURE");
```
carl-mastrangelo: Does netty epoll give the same error message?

ejona86: Yes, the test passes when the client is using epoll.

PR #4035, opened 2018-02-02
core: register channels, subchannels, oobchannels with channelz

zpencer:
- channelz is nullable and is null by default
- the *channels are registered at creation and unregistered at
  termination

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
@@ -133,7 +133,6 @@ public abstract class AbstractManagedChannelImplBuilder
   long perRpcBufferLimit = DEFAULT_PER_RPC_BUFFER_LIMIT_IN_BYTES;
   boolean retryDisabled = true; // TODO(zdapeng): default to false
 
-  @Nullable
   Channelz channelz = Channelz.instance();
 
   protected TransportTracer.Factory transportTracerFactory = TransportTracer.getDefaultFactory();
--- a/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
@@ -133,6 +133,9 @@ public abstract class AbstractManagedChannelImplBuilder
   long perRpcBufferLimit = DEFAULT_PER_RPC_BUFFER_LIMIT_IN_BYTES;
   boolean retryDisabled = true; // TODO(zdapeng): default to false
 
+  @Nullable
+  Channelz channelz = Channelz.instance();
+
   protected TransportTracer.Factory transportTracerFactory = TransportTracer.getDefaultFactory();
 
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -283,11 +285,30 @@ void handleNotInUse() {
...
+          children.add(internalSubchannel.getLogId());
+        }
+        for (OobChannel oobChannel : oobChannels) {
+          children.add(oobChannel.getLogId());
+        }
+        builder.setSubchannels(children);
+        ret.set(builder.build());
+      }
+    });
+    if (!ret.isDone()) {
```
carl-mastrangelo: This will always be true.  ChannelExecutor won't start until drain is called.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -130,6 +130,9 @@
   long retryBufferSize = DEFAULT_RETRY_BUFFER_SIZE_IN_BYTES;
   long perRpcBufferLimit = DEFAULT_PER_RPC_BUFFER_LIMIT_IN_BYTES;
 
+  @Nullable
+  Channelz channelz = Channelz.instance();
```
carl-mastrangelo: I think this should never be null.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
zpencer: I was perhaps microoptimizing for the case where we raced with a gRPC thread, then we can avoid calling drain. Since we are potentially a low pri application thread, we don't want to risk taking on important work from the queue.

core/src/main/java/io/grpc/internal/Channelz.java
```diff
@@ -89,9 +163,23 @@ public Builder setLastCallStartedMillis(long lastCallStartedMillis) {
...
       public ChannelStats build() {
         return new ChannelStats(
-            target, state, callsStarted,  callsSucceeded,  callsFailed,  lastCallStartedMillis);
+            target,
+            state,
+            callsStarted,
+            callsSucceeded,
+            callsFailed,
+            lastCallStartedMillis,
+            subchannels);
```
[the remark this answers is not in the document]
zpencer: wrapped

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
carl-mastrangelo: Don't optimize for that case.  In the case drain() isn't called very soon after, this may not execute.   It's like not calling flush and hoping someone else does.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
carl-mastrangelo: can you remove the `@Nullable` ?

zpencer:
@carl-mastrangelo care to take one last look at this before it goes in? `InstrumentedImpl` may be removed later if we move the orphan detection to a wrapper class.

PR #4036, opened 2018-02-02
core: retry part 5 continued, logic on retry policy

dapengzhang0:
Implement logic for making retry decision using RetryPolicy.

A lot of old TODOs in tests that depend on backoff implementation are fulfilled.

### DIFF
--- a/core/src/main/java/io/grpc/internal/RetriableStream.java
+++ b/core/src/main/java/io/grpc/internal/RetriableStream.java
@@ -85,7 +85,7 @@ abstract class RetriableStream<ReqT> implements ClientStream {
 
   private ClientStreamListener masterListener;
   private Future<?> scheduledRetry;
-  private double nextBackoffInSeconds;
+  private double nextBackoffIntervalInSeconds;
 
   RetriableStream(
       MethodDescriptor<ReqT, ?> method, Metadata headers,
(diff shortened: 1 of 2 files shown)


dapengzhang0:
@ericgribkoff had a chance to take a look?

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -81,6 +85,7 @@
 
   private ClientStreamListener masterListener;
   private Future<?> scheduledRetry;
+  private double nextBackoffInSeconds;
```
ericgribkoff: I think this name is slightly confusing. It's not the amount of time until the next scheduled attempt (which is `RetryPlan#backoffInMillis`), but rather this defines the interval `(0, nextBackoffInSeconds)` from which the subsequent backoff time can be chosen. Maybe `nextBackoffWindowInSeconds` or `nextBackoffIntervalInSeconds` would be better names?

core/src/test/java/io/grpc/internal/RetriableStreamTest.java
```diff
@@ -74,6 +77,31 @@
   private static final int MAX_OUTNBOUND_MESSAGE_SIZE = 5678;
   private static final long PER_RPC_BUFFER_LIMIT = 1000;
   private static final long CHANNEL_BUFFER_LIMIT = 2000;
+  private static final int MAX_ATTEMPTS = 6;
+  private static final double INITIAL_BACKOOF_IN_SECONDS = 100D;
```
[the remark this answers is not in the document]
dapengzhang0: ooof!

Issue #3819, opened 2017-12-01
ServerCallImpl.internalClose triggers onComplete, not onCancel


ejona86:
When we get an outbound error with ServerCallImpl, we call [`internalClose(Status)`](
https://github.com/grpc/grpc-java/blob/v1.8.0/core/src/main/java/io/grpc/internal/ServerCallImpl.java#L206). However, it calls stream.close() exactly as if the client closed the RPC.
```java
  private void internalClose(Status internalError) {
    stream.close(internalError, new Metadata());
  }
...
  public void close(Status status, Metadata trailers) {
    ...
    stream.close(status, trailers);
  }
```

So the transport thinks this was an application-requested failure and so will call `onComplete()`. It should call `onCancel()` instead.

Issue #3746, opened 2017-11-15
Make ServerImpl's internalClose thread-safe


ejona86:
So we have this in `ServerImpl.JumpToApplicationThreadServerStreamListener`:

```java
/**
 * Like {@link ServerCall#close(Status, Metadata)}, but thread-safe for internal use.
 */
private void internalClose() {
  // TODO(ejona86): this is not thread-safe :)
  stream.close(Status.UNKNOWN, new Metadata());
}
```

We should fix that. There's two ways to fix this:
1. Allow truncated messages in our stream and deliver trailers. This would be nicest and would be helpful in other cases (like if too many requests are sent in a server call), as we can deliver a message to the client
2. Treat this as a cancel and send RST_STREAM. This would be similar to client-side cancellation, but we'd still need to verify it'd be safe on server-side.

In either case it'd probably be good if we made this trigger onCancel instead of onCompleted. That'd probably happen naturally for (2), but (1) would need special plumbing, but it's probably not too onerous.

ejona86:
As seen in #4168, the RST_STREAM option doesn't work in the case the server hasn't sent anything yet, as users may be expecting UNKNOWN.

PR #4038, opened 2018-02-04
core: change serverimpl,servercallimpl's internalclose to cancel stream

ramaraochavali:
Signed-off-by: Rama <ramaraochavali@gmail.com>
Fixes https://github.com/grpc/grpc-java/issues/3819, https://github.com/grpc/grpc-java/issues/3746

### DIFF
--- a/core/src/main/java/io/grpc/internal/ServerCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerCallImpl.java
@@ -205,7 +205,7 @@ final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
    * on.
    */
   private void internalClose(Status internalError) {
-    stream.close(internalError, new Metadata());
+    stream.cancel(internalError);
   }
 
   /**
--- a/core/src/main/java/io/grpc/internal/ServerCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerCallImpl.java
@@ -22,7 +22,6 @@ import static com.google.common.base.Preconditions.checkState;
 import static io.grpc.internal.GrpcUtil.ACCEPT_ENCODING_SPLITTER;
 import static io.grpc.internal.GrpcUtil.MESSAGE_ACCEPT_ENCODING_KEY;
 import static io.grpc.internal.GrpcUtil.MESSAGE_ENCODING_KEY;
-import static java.lang.String.format;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.MoreExecutors;
@@ -207,7 +206,7 @@ final class ServerCallImpl<ReqT, RespT> extends ServerCall<ReqT, RespT> {
    * on.
    */
   private void internalClose(Status internalError) {
-    log.log(Level.WARNING,format("cancelling the stream with status %s",internalError.toString()));
+    log.log(Level.WARNING, "Cancelling the stream with status {0}", new Object[] {internalError});
     stream.cancel(internalError);
   }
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/grpc/internal/ServerCallImpl.java
```diff
@@ -205,7 +205,7 @@ public String getAuthority() {
    * on.
    */
   private void internalClose(Status internalError) {
-    stream.close(internalError, new Metadata());
+    stream.cancel(internalError);
```
ejona86: This status is stored internally in gRPC, but won't actually be exposed anywhere on server-side. We should probably go ahead and log it here.

interop-testing/src/test/java/io/grpc/testing/integration/MoreInProcessTest.java
```diff
@@ -247,7 +247,7 @@ public void onCompleted() {
         .onNext(StreamingInputCallRequest.getDefaultInstance());
 
     assertTrue(finishLatch.await(900, TimeUnit.MILLISECONDS));
-    assertEquals(Status.UNKNOWN, Status.fromThrowable(throwableRef.get()));
+    assertEquals(Status.CANCELLED.getCode(), Status.fromThrowable(throwableRef.get()).getCode());
```
ejona86: Right now calling `serverStream.cancel()` will propagate a [CANCEL HTTP/2 error](https://github.com/grpc/grpc-java/blob/35665af/netty/src/main/java/io/grpc/netty/NettyServerHandler.java#L658) to the remote. It'd probably be better if it was [INTERNAL_ERROR](http://httpwg.org/specs/rfc7540.html#ErrorCodes), which will get [mapped to a INTERNAL](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#errors) grpc status code. That way the client would get a signal that it shouldn't retry.

I think it'd be fair to map gRPC's DEADLINE_EXCEEDED and CANCELLED error codes to HTTP/2's CANCEL, and everything else map to HTTP/2's INTERNAL_ERROR. Right now we use [DEADLINE_EXCEEDED for when deadline expires](https://github.com/grpc/grpc-java/blob/5d5e8c6/core/src/main/java/io/grpc/internal/ServerImpl.java#L496). That mapping could be in `NettyServerHandler.cancelStream()`.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -610,8 +610,7 @@ void setListener(ServerStreamListener listener) {
      * Like {@link ServerCall#close(Status, Metadata)}, but thread-safe for internal use.
      */
     private void internalClose() {
-      // TODO(ejona86): this is not thread-safe :)
-      stream.close(Status.UNKNOWN, new Metadata());
+      stream.cancel(Status.INTERNAL);
```
jyane: Fix doc comments if you change to `cancel` ?

jyane: ~~Fix doc comments if you change?~~ edit: Please don't care.

core/src/main/java/io/grpc/internal/ServerCallImpl.java
ramaraochavali: ok. which level? Is SEVERE the correct level?

ramaraochavali:
@ejona86 addressed the comments, PTAL and let me know if any further changes are needed.

core/src/main/java/io/grpc/internal/ServerCallImpl.java
ejona86: WARNING seems good. SEVERE seems a bit too much, because we were able to recover.

core/src/main/java/io/grpc/internal/ServerCallImpl.java
```diff
@@ -205,7 +207,8 @@ public String getAuthority() {
    * on.
    */
   private void internalClose(Status internalError) {
-    stream.close(internalError, new Metadata());
+    log.log(Level.WARNING,format("cancelling the stream with status %s",internalError.toString()));
```
ejona86: Pass `internalError` as an argument and use MessageFormat-style syntax:
```java
log.log(Level.WARNING, "Cancelling the stream with status {0}", new Object[] {internalError});
```

We mainly use `String.format()` when we need to create a gRPC status description with an argument.

ejona86:
One small change, and then I can merge.

ramaraochavali:
@ejona86 changed. PTAL.

PR #4039, opened 2018-02-05
compiler: avoid invoking experimental method in generated code

ericgribkoff:
https://github.com/grpc/grpc-java-api-checker/pull/1 (cc @jyane) adds a tool for checking for unintended uses of `@ExperimentalApi` or `@Internal` code via an error prone plugin. Since `@ExperimentalApi` is itself `@Internal`, the `@ExperimentalApi` annotation in the generated code causes the annotation checker to report an error, whenever the generated gRPC service code is not in an `io.grpc` (sub)package.

One option is to `@SuppressWarnings("Internal")`, as done here. An alternative might be to configure the annotation checker to exclude generated code, but that seems more cumbersome - right now we're generating code for the user to compile that includes use of a gRPC `@Internal` API, so just suppressing the warnings related to this seems appropriate.

@carl-mastrangelo What do you think?

### DIFF
--- a/compiler/src/java_plugin/cpp/java_generator.cpp
+++ b/compiler/src/java_plugin/cpp/java_generator.cpp
@@ -78,6 +78,10 @@ static inline string MethodPropertiesGetterName(const MethodDescriptor* method) 
   return MixedLower("get_" + method->name() + "_method");
 }
 
+static inline string MethodPropertiesGetterHelperName(const MethodDescriptor* method) {
+  return MixedLower("get_" + method->name() + "_method_helper");
+}
+
 static inline string MethodIdFieldName(const MethodDescriptor* method) {
   return "METHODID_" + ToAllUpperCase(method->name());
 }
@@ -319,6 +323,7 @@ static void PrintMethodFields(
     (*vars)["method_field_name"] = MethodPropertiesFieldName(method);
     (*vars)["method_new_field_name"] = MethodPropertiesGetterName(method);
     (*vars)["method_method_name"] = MethodPropertiesGetterName(method);
+    (*vars)["method_method_name_helper"] = MethodPropertiesGetterHelperName(method);
     bool client_streaming = method->client_streaming();
     bool server_streaming = method->server_streaming();
     if (client_streaming) {
@@ -346,7 +351,7 @@ static void PrintMethodFields(
           "@$ExperimentalApi$(\"https://github.com/grpc/grpc-java/issues/1901\")\n"
           "@$Deprecated$ // Use {@link #$method_method_name$()} instead. \n"
           "public static final $MethodDescriptor$<$input_type$,\n"
-          "    $output_type$> $method_field_name$ = $method_method_name$();\n"
+          "    $output_type$> $method_field_name$ = $method_method_name_helper$();\n"
           "\n"
           "private static volatile $MethodDescriptor$<$input_type$,\n"
           "    $output_type$> $method_new_field_name$;\n"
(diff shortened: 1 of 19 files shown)


ejona86:
We could also just remove the `@Internal` annotation from `@ExperimentalApi` (or make it experimental).

ejona86:
Actually... because we are referencing `@ExperimentalApi` from generated code, that means it must be stable, since if we ever broke the annotation that would break compiling the old generated code.

carl-mastrangelo:
@ejona86 how about we just drop the Experimental tag?  I dont think there has been any issues since we moved to the MD getters.

ejona86:
We're going to want to use the annotation again in the future.

carl-mastrangelo:
This change makes me uncomfortable.    SuppressWarnings("Internal") is too generic and not readable in my opinion.  It's hard to see what exactly is internal, the method, or the annotation on the method.  

I added the ExperimentalApi annotation here, but I think in retrospect that was a bad idea.   Everything public in the stub should be api stable from the get go, because it cannot reasonably be unstable.  The bar is higher.

ejona86:
> Everything public in the stub should be api stable from the get go, because it cannot reasonably be unstable.

Why is that? We could still remove the old generated APIs, without causing any more trouble than normal.

jyane:
IMHO, another suggestion, we might could skip when a symbol owner is annotated with `@Generated`.
I think (It might wrong) generated service codes were annotated with `@Generated` so we can find `@Generated` via AST.

This approach is simply traverse AST for finding the `@Generated`, I've created the POC.
https://github.com/jyane/obsolate/compare/master...skip-generated (This is previous repository of grpc-java-api-checker)

jyane:
> An alternative might be to configure the annotation checker to exclude generated code, but that seems more cumbersome

Sorry, I overlooked...

ericgribkoff:
@jyane You're right, the generated code included an annotation stating it was produced by the gRPC compiler. If the checker itself can programmatically exclude generated code, that sounds like an ideal solution to me. (In my earlier comment about configuring the checker to avoid generated code, I was only thinking of configuration via the end user's build file manually excluding the generated files - doing this instead programmatically sounds very appropriate here)

ericgribkoff:
Updated the PR to depend on https://github.com/grpc/grpc-java/pull/4051 (removing the `@Internal` annotation on `Internal` and `ExperimentalApi`). We still have the deprecated `@ExperimentalApi METHOD_UNARY_CALL` type constants that invoke the `@ExperimentalApi getUnaryCallMethod()` methods, so a suppress warnings is still necessary for the api checker on these constants - we could also make a new private function to workaround the experimental call, but it seemed ok to just suppress the warnings on these (deprecated) calls instead.

@jyane I discussed this more with the grpc team, and in general our codegen should not be depending on `@Internal` or `@ExperimentalApi` code, so excluding our generated code from the API checker is not necessary - instead gRPC should correctly address the root cause of the problem rather than requiring a workaround in the api checker.

ericgribkoff:
...and there are a lot more invocations of the `getNAMEMethod()` experimental methods in the generated code, so may need to switch to new non-experimental private method after all to avoid suppress warnings all over the place, will update PR.

ericgribkoff:
PTAL - added a private `getNAMEMethodHelper()` method and use that throughout the generated code to avoid invoking the public `@ExperimentalApi` method. 

Will later need to include some compiler tests that verifying future changes don't re-break the annotation checker tool, but that will be a follow-up PR once the annotation checker is released

compiler/src/java_plugin/cpp/java_generator.cpp
```diff
@@ -78,6 +78,10 @@ static inline string MethodPropertiesGetterName(const MethodDescriptor* method)
   return MixedLower("get_" + method->name() + "_method");
 }
 
+static inline string MethodPropertiesGetterHelper(const MethodDescriptor* method) {
```
ejona86: s/Helper/HelperName/ to match the others.

ericgribkoff: Done

PR #4041, opened 2018-02-05
kokoro: Avoid getting cmake if it's installed

ejona86:
Kokoro already has cmake installed, so it's not necessary in that
environment. We still keep the old code around because it's helpful for
setting up new Windows environments in general.


ejona86:
I'm not using nested ifs because they break in weird ways.

PR #4042, opened 2018-02-05
core: extract provider class out of ServiceProvidersTest

zpencer:
The resource files name containing '$' is causing problems for our dev process.


core/src/test/java/io/grpc/ServiceProvidersTestAbstractProvider.java
```diff
@@ -0,0 +1,12 @@
+package io.grpc;
```
ericgribkoff: This needs a copyright. I think that's the cause of the complaint from checkstyle on Travis

ejona86:
Why was `ServiceProvidersTestAbstractProvider` moved out into a separate class?

PR #4044, opened 2018-02-06
core: retry part 6, client options - disable retry

dapengzhang0:


core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -360,6 +360,34 @@ public T perRpcBufferLimit(long bytes) {
     throw new UnsupportedOperationException();
   }
 
+
+  /**
+   * Disables the retry mechanism provided by the gRPC library. This is designed for the case when
+   * users have their own retry implementation and want to avoid their own retry taking place
+   * simultaneously with the gRPC library layer retry.
+   *
+   * @return {@code this}
```
ericgribkoff: optional: the javadoc on other methods here just have `@return this` without the `{@code }` tag (did a quick search and not using `@code` here seems to be more common)

PR #4045, opened 2018-02-06
core: retry part 6, client options - maxAttempts

dapengzhang0:


core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -327,6 +327,38 @@ public T keepAliveWithoutCalls(boolean enable) {
     throw new UnsupportedOperationException();
   }
 
+  /**
+   * Sets max number of retry attempts. The total number of retry attempts for each RPC will not
+   * exceed this number even service config may allow a higher number. Setting this number to zero
```
ericgribkoff: s/even if the service config/even service config/

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -327,6 +327,38 @@ public T keepAliveWithoutCalls(boolean enable) {
...
+   * @since 1.11.0
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/3982")
+  public T maxRetryAttempts(int maxRetryAttempts) {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Sets max number of hedged attempts. The total number of hedged attempts for each RPC will not
+   * exceed this number even service config may allow a higher number.
```
ericgribkoff: s/even if the service config/even service config/

PR #4052, opened 2018-02-08
netty: pass full authority to SSLEngine

carl-mastrangelo:
Previously, attempting to call overrideAuthority with a valid
authority would drop the beginning portion of the authority if it
contained a User.  For example `TestUser@testserver`.   This would
make the SSL handshake fail because the certificate name woudl be
`TestUser@testserver`, but the name to match would be `testserver`.

This usage is uncommon, but it may not be possible to modify the
certificate being used.  The only previous recourse was to disable
hostname verification, which is less than ideal.

Instead, pass the authority through without the port, rather than
the hostname.  This also solves the problem with invalid host names
like "test_cert".   This is the same behavior as the Go gRPC
implementation.

Note: the SSL engine has called this parameter "host", but I
believe this is a historical artifact.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
+++ b/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
@@ -256,41 +256,33 @@ public final class ProtocolNegotiators {
   public static ProtocolNegotiator tls(SslContext sslContext, String authority) {
     Preconditions.checkNotNull(sslContext, "sslContext");
     URI uri = GrpcUtil.authorityToUri(Preconditions.checkNotNull(authority, "authority"));
-    String host;
-    int port;
-    if (uri.getHost() != null) {
-      host = uri.getHost();
-      port = uri.getPort();
+    int port = uri.getPort();
+    String portlessAuthority;
+    if (port == -1) {
+      log.log(Level.FINE, "Couldn't find port in " + authority);
+      // if we can't find a port, just pass it through.
+      portlessAuthority = authority;
     } else {
-      /*
-       * Implementation note: We pick -1 as the port here rather than deriving it from the original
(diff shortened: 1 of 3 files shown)


ejona86:
I feel like the driving factor of this PR is to support hostnames with underscores. Unfortunately, it fails to do so, as the URI parser does not parse a port if the hostname contains an underscore. Thus, the code determines the authority of `bad_host:80` is `bad_host:80`, not `bad_host`.

> Previously, attempting to call overrideAuthority with a valid
authority would drop the beginning portion of the authority if it
contained a User. For example TestUser@testserver.

This is the proper behavior. Try going to https://ejona@www.google.com/ . It works, even though ejona@www.google.com isn't in the certificate. The username was intended for authentication (although now heavily deprecated because it was a bad idea). It is not used as part of [hostname verification](https://tools.ietf.org/html/rfc2818#section-3.1), as it is not a hostname.

While it is possible to receive a certificate for username@host, this is like an email address and used for client identification, not server. In no way should it validate successfully in this case.

> Note: the SSL engine has called this parameter "host", but I
believe this is a historical artifact.

No. It's correct. It is the host part of the URI.

carl-mastrangelo:
> I feel like the driving factor of this PR is to support hostnames with underscores.
It's not about underscores.

> This is the proper behavior. Try going to https://ejona@www.google.com/ . It works, even though ejona@www.google.com isn't in the certificate. The username was intended for authentication (although now heavily deprecated because it was a bad idea). It is not used as part of hostname verification, as it is not a hostname.

The problem this PR fixes is the inverse: the `ejona@www.google.com` is part of the certificate.

ejona86:
@carl-mastrangelo, do you want to close this, since we spoke offline about it and looked into the RFCs?

carl-mastrangelo:
Yes, sorry, I will close this.

PR #4060, opened 2018-02-09
core: retry part 5 continued, throttle policy

dapengzhang0:
This is the last feature of retry impl PRs (not including transparent retry and hedging).

### DIFF
--- a/core/src/main/java/io/grpc/internal/RetriableStream.java
+++ b/core/src/main/java/io/grpc/internal/RetriableStream.java
@@ -588,42 +588,40 @@ abstract class RetriableStream<ReqT> implements ClientStream {
     private RetryPlan makeRetryDecision(RetryPolicy retryPolicy, Status status, Metadata trailer) {
       boolean shouldRetry = false;
       long backoffInMillis = 0L;
+      boolean isRetryableStatusCode = retryPolicy.retryableStatusCodes.contains(status.getCode());
+
       String pushbackStr = trailer.get(GRPC_RETRY_PUSHBACK_MS);
-      if (pushbackStr == null) {
-        if (retryPolicy.retryableStatusCodes.contains(status.getCode())
-            && (throttle == null || throttle.onQualifiedFailureThenCheckIsAboveThreshold())
-            && retryPolicy.maxAttempts > substream.previousAttempts + 1) {
-          shouldRetry = true;
-          backoffInMillis = (long) (nextBackoffIntervalInSeconds * 1000D * random.nextDouble());
-          nextBackoffIntervalInSeconds = Math.min(
-              nextBackoffIntervalInSeconds * retryPolicy.backoffMultiplier,
-              retryPolicy.maxBackoffInSeconds);
-        } // else no retry
-      } else {
-        int pushback;
+      Integer pushback = null;
+      if (pushbackStr != null) {
         try {
-          pushback = Integer.parseInt(pushbackStr);
+          pushback = Integer.valueOf(pushbackStr);
         } catch (NumberFormatException e) {
           pushback = -1;
         }
-        if (pushback < 0) { // no retry
-          if (throttle != null) {
-            throttle.onQualifiedFailureThenCheckIsAboveThreshold();
-          }
-        } else if (
-            (
-                retryPolicy.retryableStatusCodes.contains(status.getCode())
-                && (throttle == null || throttle.onQualifiedFailureThenCheckIsAboveThreshold())
-                && retryPolicy.maxAttempts > substream.previousAttempts + 1)
-            || (
-                !retryPolicy.retryableStatusCodes.contains(status.getCode())
-                && (throttle == null || throttle.isAboveThreshold())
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -831,11 +847,70 @@ public void outboundWireSize(long bytes) {
...
+     * a not-to-retry pushback.
+     */
+    @VisibleForTesting
+    boolean onQualifiedFailureThenCheckIsAboveThreshold() {
+      while (true) {
+        int currentCount = tokenCount.get();
+        if (currentCount == 0) {
+          return false;
+        }
+        int decremented = currentCount - 1000;
```
ericgribkoff: Maybe replace 1000 with a constant, so it's not necessary to read the comment in the constructor to understand why it's not just subtracting 1 here?

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -831,11 +847,70 @@ public void outboundWireSize(long bytes) {
...
+    long addAndGet(long newBytesUsed) {
       return bufferUsed.addAndGet(newBytesUsed);
     }
   }
 
+  /**
+   * Used for retry throttling.
+   */
+  static final class Throttle {
+    final int maxTokens;
```
ericgribkoff: Optional: maybe helpful to include comments (text from the retry design) saying what these values represent

core/src/test/java/io/grpc/internal/RetriableStreamTest.java
```diff
@@ -1143,6 +1165,204 @@ public void pushback_noRetry() {
...
+    // mimic some other call in the channel triggers a throttle countdown
+    assertTrue(throttle.onQualifiedFailureThenCheckIsAboveThreshold()); // count = 3
+
+    sublistenerCaptor.getValue().closed(Status.fromCode(RETRIABLE_STATUS_CODE_1), new Metadata());
+    verify(retriableStreamRecorder).postCommit();
+    assertFalse(throttle.isAboveThreshold()); // count = 2
+  }
+
+  @Test
+  public void throttledStream_FailWithNonretriableStatusCode_WithoutPushback() {
```
ericgribkoff: Nit: s/Nonretriable/NonRetriable/

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -519,6 +522,9 @@ public void headersRead(Metadata headers) {
       commitAndRun(substream);
       if (state.winningSubstream == substream) {
         masterListener.headersRead(headers);
+        if (throttle != null) {
+          throttle.onSuccess();
```
ericgribkoff: Isn't this too early to say the RPC is a success? What if the winning substream later fails with a retryable status code? I think this would ends up with the same stream recording as a success (increment) and a failure (decrement) to the throttling policy's token count.

ericgribkoff: I missed that committing the substream here will prevent the `#makeRetryDecision()` code from running when the stream is closed, so it can't subsequently decrement the count.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -582,35 +588,45 @@ public void run() {
...
+        } catch (NumberFormatException e) {
+          pushback = -1;
         }
+        if (pushback < 0) { // no retry
+          if (throttle != null) {
+            throttle.onQualifiedFailureThenCheckIsAboveThreshold();
+          }
+        } else if (
+            (
+                retryPolicy.retryableStatusCodes.contains(status.getCode())
```
ericgribkoff: The conditionals here are pretty complicated, especially with the side effects of invoking `#onQualifiedFailureThenCheckIsAboveThreshold()`.  It would be nice if the flow here could be made a little more straightforward. No simplification immediately jumps out, except FWIW you can factor out the `&& retryPolicy.maxAttempts > substream.previousAttempts + 1` condition since it's shared by both sides of the OR condition...that doesn't help a whole lot though. Turning that check into a boolean variable along with maybe `boolean retryableStatusCode = retryPolicy.retryableStatusCodes.contains(status.getCode())` might help somewhat, as that's also used in all of the conditionals here.

core/src/main/java/io/grpc/internal/RetriableStream.java

core/src/main/java/io/grpc/internal/RetriableStream.java

core/src/test/java/io/grpc/internal/RetriableStreamTest.java
dapengzhang0: Done.

core/src/main/java/io/grpc/internal/RetriableStream.java
dapengzhang0: Thanks for the suggestion! Done.

dapengzhang0:
@ericgribkoff Thanks a lot for the review. PTAL

PR #4061, opened 2018-02-09
context: add basic context profiling

carl-mastrangelo:
This will allow us to measure how often contexts of particular sizes are used.

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -104,6 +104,9 @@ public class Context {
       new PersistentHashArrayMappedTrie<Key<?>, Object>();
 
   static final AtomicLongArray withValueCounts;
+  /**
+   * Counts how many times a value added to the context was already present.
+   */
   static final AtomicLongArray withValueDupeCounts;
   static final AtomicLongArray getCounts;
   private static final Method threadLocalRandomCurrent;
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -105,9 +105,9 @@ public class Context {
 
   static final AtomicLongArray withValueCounts;
   /**
-   * Counts how many times a value added to the context was already present.
+   * Counts how many times a unique value added to the context.
    */
-  static final AtomicLongArray withValueDupeCounts;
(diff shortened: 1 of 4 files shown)


context/src/main/java/io/grpc/PersistentHashArrayMappedTrie.java
```diff
@@ -221,13 +245,21 @@ public V get(K key, int hash, int bitsConsumed) {
             newValues,
             compressedIndex + 1,
             values.length - compressedIndex);
-        return new CompressedIndex<K,V>(newBitmap, newValues);
+        int newSize = size();
+        newSize += 1; // The new leaf
+        for (int i = compressedIndex; i < values.length; i++) {
```
zpencer: I think we just want to `size() + 1` here. The two arraycopy calls ensure all the original items are in the new array.

context/src/main/java/io/grpc/Context.java
```diff
@@ -316,23 +365,49 @@ public CancellableContext withDeadline(Deadline deadline,
    *
    */
   public <V> Context withValue(Key<V> k1, V v1) {
-    return new Context(this, keyValueEntries.put(k1, v1));
+    PersistentHashArrayMappedTrie<Key<?>, Object> newKeyValueEntries = keyValueEntries.put(k1, v1);
+    if (shouldSample()) {
+      withValueDupeCounts.addAndGet(
```
zpencer: It's not obvious what withValueDupeCounts is tracking, can we add a comment?

context/src/main/java/io/grpc/PersistentHashArrayMappedTrie.java
carl-mastrangelo: Done.

context/src/main/java/io/grpc/Context.java
carl-mastrangelo: Done.

zpencer: I think it should be `N - oldSize + newSize` for this stat to work, like if we added a key and the size increased from 3 to 4:

1 + 4 - 3 = 2 (right now we report 2 duplicates)
Instead if it's flipped we get:
1 + 3 - 4 = 0 (report 0 duplicates)

PR #4068, opened 2018-02-12
core: remove runDueTasks with filter

dapengzhang0:
In `FakeClock` we have `runDueTasks()` (without filter), `numPendingTasks(filter)`, and `getPendingTasks(filter)` which are useful. The `runDueTasks(filter)` method seems not a right way to test. 

This PR is low priority.


jyane:
Is there a typo in the commit message? `runDueTasks`?

dapengzhang0:
@jyane yes :smile:

PR #4071, opened 2018-02-12
netty: upgrade to 4.1.21

carl-mastrangelo:
In newer netty, bootstrap can return a non-null Channel (called a FailedChannel), even if the promise was also a failure.


netty/src/main/java/io/grpc/netty/NettyClientTransport.java
```diff
@@ -217,8 +217,8 @@ public Runnable start(Listener transportListener) {
      */
     b.handler(negotiationHandler);
     ChannelFuture regFuture = b.register();
-    channel = regFuture.channel();
-    if (channel == null) {
+    if (regFuture.isDone() && !regFuture.isSuccess()) {
```
ejona86: If channel can't be null, then it seems this if should be replaced with something that listens on `regFuture` like done for the writing of NOOP_MESSAGE.

PR #4072, opened 2018-02-12
gRPC ALTS

jiangtaoli2016:
gRPC Java ALTS code. The codes are the same as the internal version, except (1) changes in the import header, and (2) a few minor cosmetics changes such as adding comments on test files to avoid checkstyle errors.

Tested:
./gradlew build
In examples directory, manually tested on a real GCE VM.
./gradlew build
./gradlew installDist
./build/install/examples/bin/hello-world-alts-server
./build/install/examples/bin/hello-world-alts-client


alts/BUILD.bazel
```diff
@@ -0,0 +1,82 @@
...
+    name = "protobuf_imports",
+    srcs = glob(["src/main/proto/*.proto"]),
+    outs = [
+        "protobuf_out/altscontext.proto",
+        "protobuf_out/handshaker.proto",
+        "protobuf_out/transport_security_common.proto",
+    ],
+    cmd = "for fname in $(SRCS); do " +
+          "sed 's,import \",import \"alts/protobuf_out/,g' $$fname > " +
+          "$(@D)/protobuf_out/$$(basename $$fname); done",
```
jyane: Just a question: Is it works with Windows? Or grpc-java does not support Windows with Bazel?

Same problems will appear other projects (e.g. grpc-services).
Previous discussions are available at [here](https://github.com/grpc/grpc-java/pull/3939).

jiangtaoli2016: For windows, you could either use gradle or compile in linux then copy .jar to windows. 

Once bazel team has a fix on this problem https://github.com/bazelbuild/bazel/issues/4544, I will remove this genrule.

jiangtaoli2016:
It seems that all the tests failures are due to Deprecated method `valueOf()` called in generated code.
https://github.com/google/protobuf/issues/2054
https://github.com/google/protobuf/pull/4046
It seems that protobuf already suppress the warning in the last trunk. But kokoro still uses older version of protobuf compiler.

ejona86:
In alts/build.gradle, you can add after `configureProtoCompilation()`:
```
[compileJava, compileTestJava].each() {
        // protobuf calls valueof. Will be fixed in next release (google/protobuf#4046)
        it.options.compilerArgs += ["-Xlint:-deprecated"]
}
```

alts/BUILD.bazel
jyane: Thanks. Yeah, it probably works. (I thought that grpc-java does not support Windows with Bazel)

jiangtaoli2016:
The test failure comes from grpc-java/examples. It assumes grpc-alts.jar has been available. But it is not. 

@ejona86 it seems that we may need to move ALTS example client/server to grpc-java/alts directory. But that example also depends on helloworld.proto. Do we want to duplicate the proto in alts folder as well?

PR #4074, opened 2018-02-13
core: non mocking tests for forwarding {Server,Client}{Call, CallListener}

zpencer:
Instead of mocking ServerCall and ClientCall this mocks
NoopServerCall and NoopClientCall.


core/src/test/java/io/grpc/ForwardingClientCallListenerTest.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright 2015, gRPC Authors All rights reserved.
```
carl-mastrangelo: 2018

zpencer:
Ah ok, so much for this idea. I'll look into a different approach.

PR #4037, opened 2018-02-02
examples: add kotlin and kotlin android examples (low priority)

zpencer:
A few people asked about getting gRPC to work with kotlin, and the new protobuf plugin makes this work out of the box. Adding Kotlin versions of the JVM and Android examples:
* https://github.com/google/protobuf-gradle-plugin/issues/100
* https://github.com/google/protobuf-gradle-plugin/issues/176

PR #4075, opened 2018-02-13
examples: match applicationId to AndroidManifest.xml package

ericgribkoff:
`applicationId` from the gradle file replaces the package value declared in `AndroidManifest.xml` when the apk is built (see https://developer.android.com/studio/build/application-id.html#change_the_package_name), which makes for confusing behavior when the values are not aligned - the package name is important for authorization/credentials with google APIs.

### DIFF
--- a/examples/android/clientcache/app/build.gradle
+++ b/examples/android/clientcache/app/build.gradle
@@ -5,7 +5,7 @@ android {
     compileSdkVersion 27
 
     defaultConfig {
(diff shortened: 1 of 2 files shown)


ericgribkoff:
@zpencer We should probably make the same change for the Kotlin example in https://github.com/grpc/grpc-java/pull/4037/files#diff-a7539a6a51427a0750e9cfb98f6c922fR10

zpencer:
ack, updated the kotlin pr

PR #4081, opened 2018-02-14
build: run unit tests on osx, ensure proto output included

zpencer:
This adds back functionality that was accidentally dropped when
porting from travis:
- make sure generating protos will not lead to any uncomitted changes
- actually run the unit tests

### DIFF
--- a/buildscripts/kokoro/unix.sh
+++ b/buildscripts/kokoro/unix.sh
@@ -54,6 +54,7 @@ export CXXFLAGS="-I/tmp/protobuf/include"
 ./gradlew assemble generateTestProto install $GRADLE_FLAGS
 
 if [[ -z "${SKIP_CLEAN_CHECK:-}" && ! -z $(git status --porcelain) ]]; then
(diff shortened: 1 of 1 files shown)


buildscripts/kokoro/unix.sh
```diff
@@ -50,8 +50,16 @@ export LD_LIBRARY_PATH=/tmp/protobuf/lib
 export LDFLAGS=-L/tmp/protobuf/lib
 export CXXFLAGS="-I/tmp/protobuf/include"
 
-# Run tests
+# Ensure all *.proto changes include *.java generated code
 ./gradlew assemble generateTestProto install $GRADLE_FLAGS
+
+if [[ -z "${SKIP_CLEAN_CHECK:-}" && ! -z $(git status --porcelain) ]]; then
+  echo "Error Working directory is not clean. Forget to commit generated files?"
```
ejona86: Run `git status` as well, so they can see what changes there are.

PR #4087, opened 2018-02-15
Cleaner release docker

ejona86:
CC @zpencer @zhangkun83 

This removes pre-compiled protoc from the docker image, since there's nothing actually tying the version in the docker image to our source. Similarly it no longer checks out grpc-java as part of the docker image because that checkout is guaranteed to be old.

Then the question is how we want to build within the docker image. You could fire it up like before, and then do a git clone and then go on your way. run_in_docker.sh ~~is a POC that~~ drops you into the docker container with your current grpc-java checkout and as your current user so that you can develop there mostly as normal. I'd imagine the release process would eventually become `./buildscripts/run_in_docker.sh env MVN_ARTIFACTS=1 ./buildscripts/kokoro/unix.sh` which would drop results to mvn-release which is available outside the docker image. You'd then sign and upload using the scripts sign-local-repo.sh and sonatype-upload.sh.

~~This can't go in now because the new release flow involving sign-local-repo.sh and sonatype-upload.sh is still in development, and until then this would break the current release process (although that could be fixed by making a new docker script here instead of modifying our existing one).~~

### DIFF
--- a/buildscripts/kokoro/unix.sh
+++ b/buildscripts/kokoro/unix.sh
@@ -70,6 +70,7 @@ popd
 # TODO(zpencer): also build the GAE examples
 
 LOCAL_MVN_TEMP=$(mktemp -d)
+# Note that this disables parallel=true from GRADLE_FLAGS
 ./gradlew clean grpc-compiler:build grpc-compiler:uploadArchives $GRADLE_FLAGS -PtargetArch=x86_64 \
   -Dorg.gradle.parallel=false -PrepositoryDir=$LOCAL_MVN_TEMP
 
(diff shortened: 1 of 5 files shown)


buildscripts/kokoro/unix.sh
```diff
@@ -69,8 +70,8 @@ popd
 # TODO(zpencer): also build the GAE examples
 
 LOCAL_MVN_TEMP=$(mktemp -d)
-./gradlew clean grpc-compiler:build grpc-compiler:uploadArchives -PtargetArch=x86_64 \
-  -Dorg.gradle.parallel=false -PrepositoryDir=$LOCAL_MVN_TEMP $GRADLE_FLAGS
+./gradlew clean grpc-compiler:build grpc-compiler:uploadArchives $GRADLE_FLAGS -PtargetArch=x86_64 \
+  -Dorg.gradle.parallel=false -PrepositoryDir=$LOCAL_MVN_TEMP
```
zpencer: nit: add a comment about overriding `parallel` in `GRADLE_FLAGS`

ejona86:
@zpencer, PTAL. I'm now following the style guide a bit better and addressed your comments. I've also split the docker piece into a separate docker image to allow this to go in now.

PR #4088, opened 2018-02-15
alts: change visibility of `NettyServerBuilder`

jyane:
I think `AltsServerBuilder` wraps `NettyServerBuilder` and delegate could be private.
Changing delegate (intended / not intended) would break the builder pattern.


carl-mastrangelo:
cc @jiangtaoli2016

PR #4093, opened 2018-02-16
core: move ManagedChannelImpl orphan check to wrapper class

zpencer:
Spoke with ejona and determined that okhttp and netty libraries will
retain hard references to ManagedChannelImpl via its non static nested
classes. Moving the orphan check to a wrapper class solves this
problem.

- Added ForwardingManagedChannel + test
- To make testing easier: added OrphanWrappableManagedChannel,
  let refqueue and ref set be passed into the reference.
- Add integration tests for wrapper


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -197,7 +193,8 @@
   private volatile boolean terminated;
   private final CountDownLatch terminatedLatch = new CountDownLatch(1);
 
-  private final ManagedChannelReference phantom;
+  // If set, must be set immediately after instantiation
+  private Reference<?> phantom;
```
carl-mastrangelo: I think this should be volatile, since creation and shutdown may be on different threads

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -84,7 +79,8 @@
 
 /** A communication channel for making outgoing RPCs. */
 @ThreadSafe
-public final class ManagedChannelImpl extends ManagedChannel implements Instrumented<ChannelStats> {
+public final class ManagedChannelImpl extends OrphanWrappableManagedChannel
```
carl-mastrangelo: Does this actually need to be public?

core/src/main/java/io/grpc/internal/ManagedChannelOrphanWrapper.java
```diff
@@ -0,0 +1,163 @@
...
+import java.lang.ref.SoftReference;
+import java.lang.ref.WeakReference;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
+
+class ManagedChannelOrphanWrapper extends ForwardingManagedChannel {
+  private static final ReferenceQueue<ManagedChannelOrphanWrapper> REFQUEUE =
```
carl-mastrangelo: queues arent constant, so they should not get a constant name.

core/src/main/java/io/grpc/internal/ManagedChannelOrphanWrapper.java
```diff
@@ -0,0 +1,163 @@
...
+      new ReferenceQueue<ManagedChannelOrphanWrapper>();
+  // Retain the References so they don't get GC'd
+  private static final ConcurrentMap<ManagedChannelReference, ManagedChannelReference> REFS =
```
carl-mastrangelo: same with maps

core/src/main/java/io/grpc/internal/ManagedChannelOrphanWrapper.java
```diff
@@ -0,0 +1,163 @@
...
+      new ConcurrentHashMap<ManagedChannelReference, ManagedChannelReference>();
+  static final Logger logger = Logger.getLogger(ManagedChannelOrphanWrapper.class.getName());
```
carl-mastrangelo: private

core/src/main/java/io/grpc/internal/OrphanWrappableManagedChannel.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package io.grpc.internal;
+
+import io.grpc.ManagedChannel;
+import java.lang.ref.Reference;
+
+/**
+ * An ManagedChannel that can be wrapped by {@link ManagedChannelOrphanWrapper}.
```
carl-mastrangelo: a/An/A/

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -1624,71 +1611,14 @@ public void resetConnectBackoff_noOpWhenNameResolverNotStarted() {
...
-      assertThat(lr.getMessage()).contains("shutdown");
-      assertThat(lr.getParameters()).asList().containsExactly(logId, target).inOrder();
-      assertEquals(Level.SEVERE, lr.getLevel());
-    } finally {
-      channelLogger.setFilter(oldFilter);
-    }
+  public void orphanedRefClearedOnTermination() throws Exception {
+    FakeNameResolverFactory nameResolverFactory = new FakeNameResolverFactory(true);
+    createChannel(nameResolverFactory, NO_INTERCEPTOR, false /* requestConnection */,
+            ManagedChannelImpl.IDLE_TIMEOUT_MILLIS_DISABLE);
```
carl-mastrangelo: indent -4

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -1624,71 +1611,14 @@ public void resetConnectBackoff_noOpWhenNameResolverNotStarted() {
...
+    Reference<?> ref = mock(Reference.class);
```
carl-mastrangelo: im slightly afraid of mocking special jvm classes like references.  consider extending it like normal and making sure clear is called.

core/src/main/java/io/grpc/internal/ManagedChannelOrphanWrapper.java
```diff
@@ -0,0 +1,163 @@
...
+        RuntimeException maybeAllocationSite = ref.allocationSite.get();
+        ref.clearInternal(); // technically the reference is gone already.
+        if (!(ref.shutdown && ref.channel.isTerminated())) {
+          orphanedChannels++;
+          Level level = ref.shutdownNow ? Level.FINE : Level.SEVERE;
+          if (logger.isLoggable(level)) {
+            String fmt = new StringBuilder()
+                .append("*~*~*~ Channel {0} was not ")
+                // Prefer to complain about shutdown if neither has been called.
+                .append(!ref.shutdownNow ? "shutdown" : "terminated")
```
ejona86: Why was `ref.shutdown` changed to `ref.shutdownNow`?

core/src/main/java/io/grpc/internal/ManagedChannelOrphanWrapper.java
```diff
@@ -0,0 +1,163 @@
...
+  ManagedChannelOrphanWrapper(
+      OrphanWrappableManagedChannel delegate,
+      ReferenceQueue<ManagedChannelOrphanWrapper> refqueue,
+      ConcurrentMap<ManagedChannelReference, ManagedChannelReference> refs) {
+    super(delegate);
+    phantom = new ManagedChannelReference(this, delegate, refqueue, refs);
+    // zpencer: Can this be implemented with notifyWhenStateChanged and fire at SHUTDOWN instead?
+    // The reason OrphanableMC exists is because ManagedChannelImpl is both unmockable and
+    // difficult to construct.
+    delegate.setPhantom(phantom);
```
ejona86: Let's not do this. Let's either drop that part of the feature or use `awaitTermination` as a signal.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
ejona86: +1

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
ejona86: It is safe as-is. Although verifying whether it is set immediately after instantiation requires looking at multiple classes.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
zpencer: locked it down

core/src/main/java/io/grpc/internal/ManagedChannelOrphanWrapper.java
zpencer: renamed here and below

core/src/main/java/io/grpc/internal/ManagedChannelOrphanWrapper.java
zpencer: made private

core/src/main/java/io/grpc/internal/OrphanWrappableManagedChannel.java
zpencer: fixed

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
zpencer: Ack, but removed this test in the end

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
zpencer: Removed this test

core/src/main/java/io/grpc/internal/ManagedChannelOrphanWrapper.java
zpencer: Fixed, this was my mistake when fixing variables with the IDE

core/src/main/java/io/grpc/internal/ManagedChannelOrphanWrapper.java
zpencer: `awaitTermination` in the forwarding class is a nice way to do this, we can remove the interface as well.

core/src/main/java/io/grpc/internal/ManagedChannelOrphanWrapper.java
```diff
@@ -0,0 +1,170 @@
...
+    phantom.shutdownNow = true;
+    return super.shutdownNow();
+  }
+
+  @Override
+  public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
+    try {
+      return super.awaitTermination(timeout, unit);
+    } finally {
+      phantom.clear();
```
ejona86: Only clear if awaitTermination returns `true`.

PR #4094, opened 2018-02-16
core,netty,okhttp: Transparent retry

dapengzhang0:
Changes:

- `ClientStreamListener.onClose(Status status, RpcProgress rpcProgress, Metadata trailers)` added.

- `AbstractClientStream.transportReportStatus(Status status, RpcProgress rpcProgress, boolean stopDelivery, Metadata trailers)` added

- `ClientCallImpl.ClientStreamListenerImpl` will ignore the arg `rpcProgress` (non retry)

- `RetriableStream.SubListener` will handle `rpcProgress` and decide if transparent retry.

- `NettyClientHandler` and `OkHttpClientTransport` will pass `RpcProgress.REFUSED` to client stream listener for later stream ids when received GOAWAY, or for stream received a RST_STREAM frame with REFUSED code.

- All other files are just a result of refactoring.

### DIFF
--- a/core/src/main/java/io/grpc/internal/InternalSubchannel.java
+++ b/core/src/main/java/io/grpc/internal/InternalSubchannel.java
@@ -628,7 +628,7 @@ final class InternalSubchannel implements Instrumented<ChannelStats> {
 
     @Override
     public ClientStream newStream(
-        MethodDescriptor<?, ?> method, Metadata headers, CallOptions callOptions) {
+        MethodDescriptor<?, ?> method, Metadata headers, final CallOptions callOptions) {
       final ClientStream streamDelegate = super.newStream(method, headers, callOptions);
       return new ForwardingClientStream() {
         @Override
@@ -649,6 +649,13 @@ final class InternalSubchannel implements Instrumented<ChannelStats> {
             public void closed(Status status, Metadata trailers) {
               callTracer.reportCallEnded(status.isOk());
               super.closed(status, trailers);
+            }
+
+            @Override
+            public void closed(
+                Status status, RpcProgress rpcProgress, Metadata trailers) {
+              callTracer.reportCallEnded(status.isOk());
+              super.closed(status, rpcProgress, trailers);
             }
           });
         }
--- a/core/src/main/java/io/grpc/internal/InternalSubchannel.java
+++ b/core/src/main/java/io/grpc/internal/InternalSubchannel.java
@@ -628,7 +628,7 @@ final class InternalSubchannel implements Instrumented<ChannelStats> {
 
     @Override
     public ClientStream newStream(
-        MethodDescriptor<?, ?> method, Metadata headers, final CallOptions callOptions) {
+        MethodDescriptor<?, ?> method, Metadata headers, CallOptions callOptions) {
       final ClientStream streamDelegate = super.newStream(method, headers, callOptions);
       return new ForwardingClientStream() {
         @Override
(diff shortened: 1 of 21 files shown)


dapengzhang0:
@ejona86 Do you have any concerns for this transparent retry approach?

ejona86:
This approach seems fine. It would be nice to have a better name than InternalStatuses though. ResultSeverity? ResultType? CompletionType? Progress (with DROP (did try; won't ever), REFUSED (didn't try), PROCESSED (tried))?  Kill or strength (STRONG (drop), SOFT ("normal"), WEAK (refused))?

dapengzhang0:
Refactored the enum name. Test added.
Now ready for review.
PTAL.

netty/src/main/java/io/grpc/netty/NettyClientHandler.java
```diff
@@ -348,7 +350,11 @@ private void onRstStreamRead(int streamId, long errorCode) {
     if (stream != null) {
       Status status = GrpcUtil.Http2Error.statusForCode((int) errorCode)
           .augmentDescription("Received Rst Stream");
-      stream.transportReportStatus(status, false /*stop delivery*/, new Metadata());
+      stream.transportReportStatus(
+          status,
+          errorCode == Http2Error.REFUSED_STREAM.code() ? REFUSED : PROCESSED,
```
ejona86: nit: I'd prefer if the static import was avoided. This would be more clear knowing the type. Ditto elsewhere.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -626,7 +626,7 @@ protected ConnectionClientTransport delegate() {
 
     @Override
     public ClientStream newStream(
-        MethodDescriptor<?, ?> method, Metadata headers, CallOptions callOptions) {
+        MethodDescriptor<?, ?> method, Metadata headers, final CallOptions callOptions) {
```
ejona86: Why this change?

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -183,13 +184,21 @@ abstract ClientStream newSubstream(
...
       newHeaders.put(GRPC_PREVIOUS_RPC_ATTEMPTS, String.valueOf(previousAttempts));
     }
+    if (transparentRetryAttempts > 0) {
+      if (newHeaders == originalHeaders) {
+        newHeaders = new Metadata();
+        newHeaders.merge(originalHeaders);
+      }
+      newHeaders.put(GRPC_TRANSPARENT_RETRY_ATTEMPTS, String.valueOf(transparentRetryAttempts));
+    }
     return newHeaders;
```
ejona86: This must always be new headers. The transports can modify the metadata today (although I'm unsure if they actually do; it could be reasonable to institute a "no modify" policy for the stream).

core/src/main/java/io/grpc/internal/InternalSubchannel.java
dapengzhang0: Reversed.

core/src/main/java/io/grpc/internal/RetriableStream.java

netty/src/main/java/io/grpc/netty/NettyClientHandler.java
dapengzhang0: Done.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -545,6 +554,19 @@ public void closed(Status status, Metadata trailers) {
       }
 
       if (state.winningSubstream == null) {
+        if (rpcProgress == RpcProgress.REFUSED) {
```
ericgribkoff: Discussed offline, but summarizing here: this case should be retried at most once (since it does go out on the wire, we want to cap the max additional load created by the transparent retries here).

Since the transparent attempts is capped at 1, the value of sending the "grpc-transparent-retry-attempts" header in the metadata is also minimal. I'll send an update to the retry design doc dropping this requirement, as it seems like we should be able to get a reasonable approximation of these failures in purely client side stats.

dapengzhang0: Fixed. PTAL

PR #4101, opened 2018-02-20
build.gradle: bump protobuf plugin to 0.8.5

zpencer:
This update automatically adds generated sources and proto IDLs to the
`idea` plugin, so clean up the configs.


zpencer:
@zhangkun83 PTAL

PR #4104, opened 2018-02-20
core: BinaryLogProvider should only provide a BinaryLog

zpencer:
No behavioral changes in this PR
- Move the actual logic into BinaryLog.java, so that BinaryLogProvider
  only provides the BinaryLog
- the loading is now done via ServiceProvider util instead of rolling
  our own in BinaryLogProvider

### DIFF
--- a/core/src/main/java/io/grpc/internal/BinaryLog.java
+++ b/core/src/main/java/io/grpc/internal/BinaryLog.java
@@ -48,23 +48,25 @@ public abstract class BinaryLog implements Closeable {
 
   /**
    * Returns a {@link ServerInterceptor} for binary logging. gRPC is free to cache the interceptor,
-   * so the interceptor must be reusable across calls. At runtime, the request and response
-   * marshallers are always {@code Marshaller<InputStream>}.
+   * so the interceptor must be reusable across calls. At runtime, the interceptor's request and
+   * response marshallers are always {@code Marshaller<InputStream>} rather than those of the
+   * {@link MethodDescriptor} parameter.
    * Returns {@code null} if this method is not binary logged.
    */
   // TODO(zpencer): ensure the interceptor properly handles retries and hedging
   @Nullable
-  public abstract ServerInterceptor getServerInterceptor(String fullMethodName);
+  public abstract ServerInterceptor getServerInterceptor(MethodDescriptor<?, ?> method);
 
   /**
    * Returns a {@link ClientInterceptor} for binary logging. gRPC is free to cache the interceptor,
-   * so the interceptor must be reusable across calls. At runtime, the request and response
-   * marshallers are always {@code Marshaller<InputStream>}.
+   * so the interceptor must be reusable across calls. At runtime, the interceptor's request and
+   * response marshallers are always {@code Marshaller<InputStream>} rather than those of the
+   * {@link MethodDescriptor} parameter.
    * Returns {@code null} if this method is not binary logged.
    */
   // TODO(zpencer): ensure the interceptor properly handles retries and hedging
   @Nullable
-  public abstract ClientInterceptor getClientInterceptor(String fullMethodName);
+  public abstract ClientInterceptor getClientInterceptor(MethodDescriptor<?, ?> method);
 
   @Override
   public void close() throws IOException {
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/grpc/internal/BinaryLog.java
```diff
@@ -0,0 +1,153 @@
...
+
+  /**
+   * Returns a {@link ServerInterceptor} for binary logging. gRPC is free to cache the interceptor,
+   * so the interceptor must be reusable across calls. At runtime, the request and response
+   * marshallers are always {@code Marshaller<InputStream>}.
+   * Returns {@code null} if this method is not binary logged.
+   */
+  // TODO(zpencer): ensure the interceptor properly handles retries and hedging
+  @Nullable
+  public abstract ServerInterceptor getServerInterceptor(String fullMethodName);
```
carl-mastrangelo: Can this just take the MethodDescriptor as the input?

core/src/main/java/io/grpc/internal/BinaryLog.java
```diff
@@ -0,0 +1,153 @@
...
+  private static MethodDescriptor<InputStream, InputStream> toInputStreamMethod(
+      MethodDescriptor<?, ?> method) {
+    return method.toBuilder(IDENTITY_MARSHALLER, IDENTITY_MARSHALLER).build();
+  }
+
+  /**
+   * Wraps a {@link ServerMethodDefinition} such that it performs binary logging if needed.
+   */
+  final <ReqT, RespT> ServerMethodDefinition<?, ?> wrapMethodDefinition(
+      ServerMethodDefinition<ReqT, RespT> oMethodDef) {
```
carl-mastrangelo: nit: rename this `originalMethodDef`?

zpencer:
I think the BinaryLogSink being pluggable is an implementation detail to support both google3 and OSS. The BinaryLog itself being pluggable is not a part of the design nor strictly required, but is just a nice to have.

core/src/main/java/io/grpc/internal/BinaryLog.java
zpencer: This may reveal too much about the method, because the binary log design spec allows users to configure it using per gRPC method name options. Also, there is some magic happening that swaps out the Marshaller to be InputStream based, so the method descriptor can not match the service definition and the actual interceptor's Marshaller at the same time.

core/src/main/java/io/grpc/internal/BinaryLog.java
```diff
@@ -0,0 +1,153 @@
...
+import io.grpc.ServerCallHandler;
+import io.grpc.ServerInterceptor;
+import io.grpc.ServerMethodDefinition;
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
+
+public abstract class BinaryLog implements Closeable {
```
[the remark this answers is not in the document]
zpencer: added

core/src/main/java/io/grpc/internal/BinaryLog.java
zpencer: renamed

zpencer:
@carl-mastrangelo PTAL

core/src/main/java/io/grpc/internal/BinaryLog.java
carl-mastrangelo: The MD has a minor advantage from an API point of view, because the method name is unambiguous..   It also provides the opportunity for caching (like we do for the `:path:` header).  Lastly it allows additional info to be provided.  For example, we could add a MD attribute called "loggable" which controls if certain methods should be logged.  I'm not saying we will do these things, but it leaves the door open for it.

I won't keep pushing on this if you want to keep it string, but please consider it.

core/src/main/java/io/grpc/internal/BinaryLog.java
```diff
@@ -0,0 +1,157 @@
...
+  final <ReqT, RespT> ServerMethodDefinition<?, ?> wrapMethodDefinition(
+      ServerMethodDefinition<ReqT, RespT> originalMethodDef) {
+    ServerInterceptor binlogInterceptor =
+        getServerInterceptor(originalMethodDef.getMethodDescriptor().getFullMethodName());
+    if (binlogInterceptor == null) {
+      return originalMethodDef;
+    }
+    MethodDescriptor<InputStream, InputStream> binMethod =
+        toInputStreamMethod(originalMethodDef.getMethodDescriptor());
+    ServerMethodDefinition<InputStream, InputStream> binDef = InternalServerInterceptors
```
carl-mastrangelo: nit: wrap line at highest semantic level, so at the "="

core/src/main/java/io/grpc/internal/BinaryLog.java
zpencer: I think making the parameter a `MethodDescriptor<?, ?>` should make it obvious that its marshallers should not be used.

core/src/main/java/io/grpc/internal/BinaryLog.java
zpencer: done

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -16,237 +16,33 @@
...
-          Level.SEVERE, "caught exception loading BinaryLogProvider, will disable binary log", t);
-      return null;
-    }
-  }
-
-  private static BinaryLogProvider loadHelper(ClassLoader classLoader) {
-    if (isAndroid()) {
-      return null;
-    }
+public abstract class BinaryLogProvider {
```
ejona86: This class is weird. I don't know why you're deviating from the standard Provider pattern.

I'm okay with stuff moving out of this class, but the providers should extend this class and things like priority()/isAvailable() should be here.

zpencer: Perhaps I misunderstood [this comment](https://github.com/grpc/grpc-java/pull/3917#discussion_r164028522):
>This was a mistake we made with ServerBuilder. Having the static factory of the class be on itself and be extensible made for confusing behavior. The factory and the Sink should be in separate classes.

Maybe if I renamed `BinaryLog` to `BinaryLogProvider` to follow the pattern, and renamed `BinaryLogProvider` to `BinaryLogInstance` things would be cleared up.

zpencer:
Let's go for the more straight forward refactor to remove redundancy: https://github.com/grpc/grpc-java/pull/4161 , and move on to other PRs to complete the binlog functionality.

Issue #4028, opened 2018-01-31
DNS resolution failure on Android after connectivity changes


userar:
### What version of gRPC are you using?
1.9.0

### What did you expect to see?
For the grpc channel to be able to handle Android connectivity changes (eg. from wifi to mobile data, or from no data connection to wifi). 
For the `resetConnectBackoff()` call on a channel to successfully short-circuit the backoff timer and make it reconnect immediately when triggered from a connectivity change.

### What did you do
Built a grpc channel using the `OkHttpChannelBuilder`. Registered an Android `BroadcastReceiver` against connectivity changes which calls the channels `resetConnectBackoff()`  (as recommended in #4011).

### What did you see instead
The `resetConnectBackoff()`  being called from the broadcast receiver event (for `android.net.conn.CONNECTIVITY_CHANGE`) and failing to short-circuit the backoff timer. Had to wait approx 60 seconds before the channel became usable again. Reports a host name resolution failure until the 60 seconds passes.
A sleep (of a few seconds) between the connectivity change and the `resetConnectBackoff()` call seems to fix the issue.
Is there any way to decrease the default backoff time. It may be a useful feature in situations like this.

ericgribkoff:
Thanks for reporting this. I'm having trouble reproducing the issue, and if you could answer a few more questions I will be better able to help diagnose the problem.

Does this issue happen reliably or is it intermittent? Is it specific to transitioning from no connection to wifi or from no connection to cellular? The backoff timer shouldn't impact switching from wifi to mobile or vice versa; let me know if this appears not to be the case.

Are you encountering this issue on an emulator or physical device? Which Android API level(s)?

That a few seconds sleep between the connectivity change and invoking `resetConnectBackoff()` seems to help is a bit strange. Are you sure that [`NetworkInfo#isConnected`](https://developer.android.com/reference/android/net/NetworkInfo.html#isConnected()) returns true before you invoke `resetConnectBackoff()`? Can you share your `BroadcastReceiver` implementation or other code that reproduces the issue?

ericgribkoff:
Interesting. Thanks for the very detailed additional information! I was able to reproduce this issue on about 1 in 10 attempts (also on a Pixel XL @ API level 27) upon switching from wifi to mobile. 

[...]

Issue #3685, opened 2017-11-08
Use exponential backoff in DNS resolver


ejona86:
In places like #3268, people are experiencing a long delay for DNS to resolve addresses after a failure. Since we're using exponential backoff for the connection retry, we should maybe also use it for initial DNS resolution instead of a [hard-coded 60 seconds](https://github.com/grpc/grpc-java/blob/7df9ae9/core/src/main/java/io/grpc/internal/DnsNameResolver.java#L168-L170). This would reduce the need for users to plumb #2169. While #2169 is "a good thing," simply having better behavior by default is also "a good thing."

The problem we encounter today is if initial DNS resolution fails, it is retried every 60 seconds. (Note that exiting idle mode re-creates the DNS name resolver, so it should be considered "initial" resolution.) If later DNS resolutions fail, then we are fine because:

1. We will continue using the old addresses, and
2. Because [after each connection attempt](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java#L671-L675) we'll call `refresh()`.

But if the initial attempt fails then there won't be an InternalSubchannel and it won't fail periodically which won't trigger refresh().

So there's two ways to trigger the backoff:
1. implement the backoff in DnsNameResolver directly (so we don't block new connections), or
2. periodically trigger refresh() if the initial resolution fails (if the addresses are bad we will still do exponential backoff).

(1) is more straight-forward, but (2) would benefit all name resolvers. Since name resolvers notify the ChannelImpl on failure it should be possible to do (2) within ChannelImpl. It may also be possible to plumb things through the LB, such that failed name resolution still creates a subchannel but with zero addresses, and that subchannel could do backoff like normal and trigger refresh().

CC @ericgribkoff, @zhangkun83

zhangkun83:
Doing it in ChannelImpl makes the most sense to me. It can use the same backoff policy as the reconnections do. I don't see benefit of involving LB.

PR #4105, opened 2018-02-21
core: use exponential backoff for name resolution

ericgribkoff:
This addresses https://github.com/grpc/grpc-java/issues/3685. 

Drops the fixed 60-second backoff timer from `DnsNameResolver`. Instead, `ManagedChannelImpl` will use its backoff policy to invoke `NameResolver.refresh()` when `NameResolver.Listener.onError` signals that an error occurred during resolution.


ericgribkoff:
This will also resolve https://github.com/grpc/grpc-java/issues/4028

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -745,24 +786,27 @@ public void run() {
...
+    channelExecutor
+        .executeLater(
+            new Runnable() {
+              @Override
+              public void run() {
+                if (shutdown.get()) {
+                  return;
+                }
+                if (nameResolverBackoffFuture != null) {
+                  cancelNameResolverBackoff();
```
ejona86: For clarity, could you add an `assert nameResolverStarted;`? That also helps during auditing.

ejona86:
+@zhangkun83 for the NameResolver error+refresh semantic change

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -587,10 +595,50 @@ public void callOptionsExecutor() {
   @Test
   public void nameResolutionFailed() {
     Status error = Status.UNAVAILABLE.withCause(new Throwable("fake name resolution error"));
+    FakeNameResolverFactory nameResolverFactory = new FakeNameResolverFactory(error);
+    // Name resolution is started as soon as channel is created.
+    createChannel(nameResolverFactory, NO_INTERCEPTOR);
+    FakeNameResolverFactory.FakeNameResolver resolver = nameResolverFactory.resolvers.get(0);
+    verify(mockLoadBalancer).handleNameResolutionError(same(error));
 
+    timer.forwardNanos(RECONNECT_BACKOFF_INTERVAL_NANOS);
```
zhangkun83: To verify refresh() is not called sooner than expected, you will need to forward by `RECONNECT_BACKOFF_INTERVAL_NANOS - 1` and verify that refresh() is not called, then forward by 1.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -587,10 +595,50 @@ public void callOptionsExecutor() {
...
+
+    assertEquals(1, resolver.refreshCalled);
+    verify(mockLoadBalancer, times(2)).handleNameResolutionError(same(error));
+
+    // Allow the next refresh attempt to succeed
+    resolver.error = null;
+
+    // For the second attempt, the backoff should occur at RECONNECT_BACKOFF_INTERVAL_NANOS * 2
+    timer.forwardNanos(RECONNECT_BACKOFF_INTERVAL_NANOS);
```
zhangkun83: Ditto. The forward time here should be RECONNECT_BACKOFF_INTERVAL_NANOS * 2 - 1, and the next forward should be 1.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -587,10 +595,50 @@ public void callOptionsExecutor() {
...
+    assertEquals(1, resolver.refreshCalled);
+    timer.forwardNanos(RECONNECT_BACKOFF_INTERVAL_NANOS);
+    assertEquals(2, resolver.refreshCalled);
+    assertEquals(0, timer.numPendingTasks());
+
+    // Verify that the successful resolution reset the backoff policy
+    resolver.listener.onError(error);
+    timer.forwardNanos(RECONNECT_BACKOFF_INTERVAL_NANOS);
```
zhangkun83: Ditto.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -587,10 +595,50 @@ public void callOptionsExecutor() {
...
+    // Name resolution is started as soon as channel is created.
+    createChannel(nameResolverFactory, NO_INTERCEPTOR);
+    FakeNameResolverFactory.FakeNameResolver resolver = nameResolverFactory.resolvers.get(0);
+    verify(mockLoadBalancer).handleNameResolutionError(same(error));
 
+    timer.forwardNanos(RECONNECT_BACKOFF_INTERVAL_NANOS);
+
+    assertEquals(1, resolver.refreshCalled);
+    verify(mockLoadBalancer, times(2)).handleNameResolutionError(same(error));
+
```
zhangkun83: To verify that you will not schedule duplicate timers, maybe also call refresh() here, and verify `handleNameResolutionError()` is called twice, and there is still one timer scheduled.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -587,10 +595,50 @@ public void callOptionsExecutor() {
...
     // Name resolution is started as soon as channel is created.
-    createChannel(new FailingNameResolverFactory(error), NO_INTERCEPTOR);
+    createChannel(nameResolverFactory, NO_INTERCEPTOR);
     verify(mockLoadBalancer).handleNameResolutionError(same(error));
+
+    FakeClock.ScheduledTask nameResolverBackoff = getNameResolverBackoff();
+    assertNotNull(nameResolverBackoff);
+    assertFalse(nameResolverBackoff.isCancelled());
+
+    channel.shutdown();
```
zhangkun83: shutdown() doesn't cancel the timer. delayedTransport termination does. To verify this, you'd need to start a call that stops delayedTransport from terminating.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -2148,13 +2237,22 @@ void allResolved() {
...
+        this.error = error;
+      }
 
       @Override public String getServiceAuthority() {
         return expectedUri.getAuthority();
       }
 
       @Override public void start(final Listener listener) {
         this.listener = listener;
+        if (error != null) {
```
zhangkun83: Can we consolidate this with the error simulation in refresh() and put it in resolved()?

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -2211,4 +2286,17 @@ private static ChannelStats getStats(
       Instrumented<ChannelStats> instrumented) throws Exception {
     return instrumented.getStats().get();
   }
+
+  private FakeClock.ScheduledTask getNameResolverBackoff() {
+    FakeClock.ScheduledTask nameResolverBackoff = null;
+    for (FakeClock.ScheduledTask task : timer.getPendingTasks()) {
+      if (task.command.toString().contains("NameResolverBackoff")) {
```
zhangkun83: You may consider using `getPendingTasks(TaskFilter filter)` instead.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -402,6 +404,45 @@ public void run() {
         idleTimeoutMillis, TimeUnit.MILLISECONDS);
   }
 
+  // Run from channelExecutor
+  private class NameResolverBackoff implements Runnable {
```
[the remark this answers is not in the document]
ericgribkoff: Done (changed to `NameResolverRefresh`)

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
ericgribkoff: Done

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
ericgribkoff: Done

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
ericgribkoff: Done

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
ericgribkoff: Done

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
ericgribkoff: Done. Also made a larger scale refactor to switch to a builder for the FakeNameResolverFactory, as it was confusing sorting through the four different constructors and their different default values.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
ericgribkoff: Done

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
ericgribkoff: Done

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -2293,13 +2300,11 @@ private static ChannelStats getStats(
...
   private FakeClock.ScheduledTask getNameResolverRefresh() {
     FakeClock.ScheduledTask nameResolverRefresh = null;
-    for (FakeClock.ScheduledTask task : timer.getPendingTasks()) {
-      if (task.command.toString().contains("NameResolverRefresh")) {
-        assertNull(
-            "There shouldn't be more than one name resolver refresh task", nameResolverRefresh);
-        assertFalse(task.isDone());
-        nameResolverRefresh = task;
-      }
+    for (FakeClock.ScheduledTask task : timer.getPendingTasks(NAME_RESOLVER_REFRESH_TASK_FILTER)) {
```
zhangkun83: This can be simplified to:
```java
nameResolverRefresh = Iterables.getOnlyElement(timer.getPendingTasks(NAME_RESOLVER_REFRESH_TASK_FILTER));

PR #4110, opened 2018-02-21
core: add InProcessServerBuilder.generateName()

dapengzhang0:


dapengzhang0:
#1953

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
```diff
@@ -76,6 +78,13 @@ public static InProcessServerBuilder forPort(int port) {
     throw new UnsupportedOperationException("call forName() instead");
   }
 
+  /**
+   * Generates a random and unique server name.
```
ejona86: Don't include "a random". This should be an okay implementation: "generated-name-" + atomic.getAndIncrement()

Issue #1772, opened 2016-05-03
Tracking Issue for Plaintext being Experimental. 


carl-mastrangelo:
Specific usages:
- `ManagedChannelBuilder.usePlaintext`

dapengzhang0:
widely used, proposing to next

ejona86:
API review:
* Add new zero-argument overload. Equivalent of `true` today.
* Mark current version Deprecated, and begin migration.

igorbernstein2:
Is there a timeline for this stabilizing? I would like to use this for implementing a bigtable emulator

ejona86:
API review:
 * The zero-arg usePlaintext is good for stabilization.

dapengzhang0:
API is stabilized. The deprecated method `usePlaintext(boolean)` is still there and will be deleted in a year.

PR #4117, opened 2018-02-23
core,netty,okhttp,alts,inprocess: deprecate usePlaintext(boolean)

carl-mastrangelo:
Updates #1772

### DIFF
--- a/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
@@ -98,7 +98,6 @@ public abstract class ForwardingChannelBuilder<T extends ForwardingChannelBuilde
   @Override
   @Deprecated
   public T usePlaintext(boolean skipNegotiation) {
-    @SuppressWarnings("deprecation")
     ManagedChannelBuilder<?> o = delegate().usePlaintext(skipNegotiation);
     return thisT();
   }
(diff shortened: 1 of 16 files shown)


core/src/main/java/io/grpc/ForwardingChannelBuilder.java
```diff
@@ -91,9 +92,20 @@ public T overrideAuthority(String authority) {
...
   }
 
+  /**
+   * @deprecated use {@link #usePlaintext()} instead.
+   */
   @Override
+  @Deprecated
   public T usePlaintext(boolean skipNegotiation) {
-    delegate().usePlaintext(skipNegotiation);
+    @SuppressWarnings("deprecation")
```
ejona86: Is this necessary? I thought `@Deprecated` methods could call other `@Deprecated` methods.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -152,13 +152,37 @@
    *
    * @param skipNegotiation @{code true} if there is a priori knowledge that the endpoint supports
    *                        plaintext, {@code false} if plaintext use must be negotiated.
+   * @deprecated Use {@link #usePlaintext()} instead.
    *
    * @throws UnsupportedOperationException if plaintext mode is not supported.
    * @return this
    * @since 1.0.0
    */
+  @Deprecated
```
ejona86: Keep the ExperimentalApi annotation, so that users know it can actually be removed.

core/src/main/java/io/grpc/ForwardingChannelBuilder.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
carl-mastrangelo: Done.

PR #4119, opened 2018-02-23
buildscripts: check errorlevel of error-prone commands in .bat files

ejona86:
There's no 'set -e' in BAT, so we get to do it manually. Wee!

Note that we need to use 'exit /b' instead of 'exit' in the scripts, as
otherwise it exits more than just the current script.

CC @matt-kwong


matt-kwong:
FYI: https://githubengineering.com/crypto-removal-notice/ - GitHub is probably blocking older versions of TLS connections. This broke some of our builds yesterday and was a pain to figure out.

ejona86:
@matt-kwong, yeah, that's the source of the problem, thanks! But now I get to figure out how to fix powershell...

ejona86:
@zpencer ping. Should be easy.

PR #4133, opened 2018-02-23
alts: re-order dependencies in ALTS bazel build

jiangtaoli2016:


jiangtaoli2016:
renamed. Thanks for reminding.

PR #4136, opened 2018-02-23
 okhttp: support JDK9 ALPN

ericgribkoff:
Adds support for JDK9's built-in ALPN to the OkHttp transport.

Manually tested with JDK9. I removed the alpnagent from the interop-testing dependencies to test this locally, but did not remove them in this PR (I'm not sure if our test infrastructure is all running on JDK9 yet).

The first commit contains two test changes that are necessary to build on JDK9 (one copied from https://github.com/grpc/grpc-java/pull/3638).


okhttp/src/test/java/io/grpc/okhttp/OkHttpProtocolNegotiatorTest.java
```diff
@@ -110,7 +112,9 @@ public void negotiatorNotNull() {
 
   @Test
   public void negotiate_handshakeFails() throws IOException {
+    SSLParameters parameters = mock(SSLParameters.class);
```
ejona86: Why not `new SSLParameters()`? Please use real objects when easily possible.

okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/Platform.java
```diff
@@ -199,6 +209,23 @@ private static Platform findPlatform() {
       throw new RuntimeException(nsae);
     }
 
+    // Find JDK9+ ALPN support
+    try {
+      Method setApplicationProtocols =
+          SSLParameters.class.getMethod("setApplicationProtocols", String[].class);
+      Method getApplicationProtocol =
+          AccessController.doPrivileged(
+              new PrivilegedExceptionAction<Method>() {
```
ejona86: Why is this privileged and not the earlier getMethod?

okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/Platform.java
```diff
@@ -375,6 +402,56 @@ public TlsExtensionType getTlsExtensionType() {
...
+    }
+
+    /** Returns the negotiated protocol, or null if no protocol was negotiated. */
+    @Override
+    public String getSelectedProtocol(SSLSocket socket) {
+      try {
+        return (String) getApplicationProtocol.invoke(socket);
+      } catch (IllegalAccessException e) {
+        throw new RuntimeException(e);
+      } catch (InvocationTargetException e) {
```
ejona86: It is known that getApplicationProtocol can throw `UnsupportedOperationException`. This case should be handled more cleanly.

For one, we can do `e.getCause()`. From there we can either propagate it directly (if it is a RuntimeException) or wrap it and provide a message with additional information.

(Actually, if we verify the provider supports this method ahead-of-time, then I'm fine with this current code.)

okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/Platform.java
```diff
@@ -199,6 +209,23 @@ private static Platform findPlatform() {
...
+          SSLParameters.class.getMethod("setApplicationProtocols", String[].class);
+      Method getApplicationProtocol =
+          AccessController.doPrivileged(
+              new PrivilegedExceptionAction<Method>() {
+                @Override
+                public Method run() throws Exception {
+                  return SSLSocket.class.getMethod("getApplicationProtocol");
+                }
+              });
+      return new JdkAlpnPlatform(sslProvider, setApplicationProtocols, getApplicationProtocol);
```
ejona86: We should verify that `sslProvider` supports ALPN by calling `getApplicationProtocol()`.

okhttp/src/test/java/io/grpc/okhttp/OkHttpProtocolNegotiatorTest.java
ericgribkoff: Done

okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/Platform.java
ericgribkoff: An oversight. Fixed.

okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/Platform.java
ericgribkoff: Done. AFAICT this requires instantiating an SSLEngine to invoke `getApplicationProtocol()`, which is the approach now taken.

okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/Platform.java
ericgribkoff: Now verify at startup that an SSLEngine created using the provider does not throw when `getApplicationProtocol` is invoked.

PR #4137, opened 2018-02-24
core,okhttp,netty,alts,testing: Plumb proxy resolved addr to transports

### LINKED ISSUE #4029: Plumb proxy address through Name Resolver
Netty can't handle unresolved addresses. To avoid things like #4027 and any concerns of `ProxySelector`s returning unresolved addresses, we should resolve the address while the NameResolver is running. We then should plumb that result through to the transport.

We should probably have a check on addresses coming from the Name Resolver that verifies the addresses are resolved.

zpencer:
This PR has two commits, and the second commit has a lot of low
impact code churn due to an internal API cleanup.

First commit:
    ProxyDetector should resolve proxy address and plumb results
    
    ProxyDetector is now responsible for resolving the proxy's
    `InetSocketAddress`, and `ProxyParameters` asserts that the address is
    resolved. The results are plumbed through using a special subclass of
    `SocketAddress`.
    
    If a proxy should be used but the proxy can not be resolved, the
    `DnsNameResolver` will re-attempt the resolution later.
    
    Remove the unit test testing for unresolved proxy addresses, since
    it's no longer applicable.

Second commit:
    Plumb `ProxiedInetSocketAddress` down to transports
    
    Do not extract the `ProxyParameters` field and pass it into
    `ClientTransportFactory.newClientTransport`.

fixes #4029

### DIFF
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -72,13 +72,15 @@ final class DnsNameResolver extends NameResolver {
   @VisibleForTesting
   static boolean enableJndi = Boolean.parseBoolean(JNDI_PROPERTY);
 
+  @VisibleForTesting
+  final ProxyDetector proxyDetector;
+
   private DelegateResolver delegateResolver = pickDelegateResolver();
 
   private final String authority;
   private final String host;
   private final int port;
   private final Resource<ExecutorService> executorResource;
-  private final ProxyDetector proxyDetector;
   @GuardedBy("this")
   private boolean shutdown;
   @GuardedBy("this")
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -25,6 +25,7 @@ import io.grpc.Attributes;
 import io.grpc.EquivalentAddressGroup;
 import io.grpc.NameResolver;
 import io.grpc.Status;
+import io.grpc.internal.ProxyDetector.ProxiedInetSocketAddress;
 import io.grpc.internal.SharedResourceHolder.Resource;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
@@ -145,9 +146,17 @@ final class DnsNameResolver extends NameResolver {
         }
         try {
           InetSocketAddress destination = InetSocketAddress.createUnresolved(host, port);
-          ProxyParameters proxy = proxyDetector.proxyFor(destination);
+          ProxyParameters proxy;
+          try {
+            proxy = proxyDetector.proxyFor(destination);
+          } catch (Exception e) {
+            savedListener.onError(
+                Status.UNAVAILABLE.withDescription("Unable to resolve host " + host).withCause(e));
+            return;
+          }
           if (proxy != null) {
-            EquivalentAddressGroup server = new EquivalentAddressGroup(destination);
+            EquivalentAddressGroup server =
+                new EquivalentAddressGroup(new ProxiedInetSocketAddress(destination, proxy));
             savedListener.onAddresses(Collections.singletonList(server), Attributes.EMPTY);
             return;
           }
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -25,8 +25,8 @@ import io.grpc.Attributes;
 import io.grpc.EquivalentAddressGroup;
 import io.grpc.NameResolver;
 import io.grpc.Status;
-import io.grpc.internal.ProxyDetector.ProxiedInetSocketAddress;
 import io.grpc.internal.SharedResourceHolder.Resource;
+import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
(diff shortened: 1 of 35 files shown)


core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -161,28 +162,24 @@ public void run() {
         }
         try {
           InetSocketAddress destination = InetSocketAddress.createUnresolved(host, port);
-          ProxyParameters proxy = proxyDetector.proxyFor(destination);
+          final ProxyParameters proxy;
+          try {
+            proxy = proxyDetector.proxyFor(destination);
+          } catch (Exception e) {
```
ejona86: Just catch IOException?

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -161,28 +162,24 @@ public void run() {
...
+            rescheduleWithCause(savedListener, e);
+            return;
+          }
           if (proxy != null) {
-            EquivalentAddressGroup server = new EquivalentAddressGroup(destination);
+            EquivalentAddressGroup server =
+                new EquivalentAddressGroup(new ProxiedInetSocketAddress(destination, proxy));
```
ejona86: ProxiedInetSocketAddress is weird, and it can't be internal long-term. Any NR should be able to support proxies.

The discussion about it should probably include @zhangkun83. (IIRC, the problem here is how should flattening by pick-first impact the attributes of the `EquivalentAddressGroup`.)

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -205,6 +202,21 @@ public void run() {
           }
         }
       }
+
+    private void rescheduleWithCause(Listener savedListener, Exception cause) {
```
ejona86: Note that this will cause some git conflicts with #4105. (It looks like it'll be unnecessary after that other PR.)

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -147,8 +147,6 @@ void handleUncaughtThrowable(Throwable t) {
   // Only null after channel is terminated. Must be assigned from the channelExecutor.
   private NameResolver nameResolver;
 
-  private final ProxyDetector proxyDetector;
```
ejona86: Note that we should plumb the `proxyDetector` eventually through here to the Name Resolver. We shouldn't pass the `proxyDetector` is the constructor like we do now. We could pass it in as `Attributes params`.

(CC @zhangkun83, I wonder if we should revisit that `params` argument to the NameResolver and do something more similar to Params/Helper from LB. That can be later though.)

core/src/main/java/io/grpc/internal/DnsNameResolver.java
zhangkun83: ProxiedInetSocketAddress was my suggestion. I don't see a problem making it public if we need to. Can you elaborate your concerns?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
zhangkun83: Will application provide ProxyDetector? Do we want NameResolvers to decide whether to use proxy or not?

ejona86: We want proxies to make the final determination, but we should provide the detector. We shouldn't assume such information is global. There's #3957 for example.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
ejona86: @zhangkun83 and I talked at length about this offline. It seems we're going to be mucking with how NameResolvers and LBs communicate again. That will need some other requirements to be gathered. But in the short term we want some sort of `PairSocketAddress` (I'm saying that name because it is obviously bad; we need a different name) that would contain `(SocketAddress, Attr)`. It basically shoe-horns something similar to the deleted `ResolvedServerInfo` without a full redesign of the APIs.

zpencer: Does this imply `EquivalentAddressGroup::getAddresses()` would return a list of `PairSocketAddress`? Sounds like we will just commit to making this class public now, and reusing it later for NameResolvers+LBs.

ejona86: It would continue to return a list of `SocketAddress`. No, we're not ready to make this public. There needs to be a longer-term design and that needs some investigation into requirements.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
zpencer: done

core/src/main/java/io/grpc/internal/DnsNameResolver.java
zpencer: ack, changed it to `PairSocketAddress`

core/src/main/java/io/grpc/internal/DnsNameResolver.java
zpencer: Rebased and removed unnecessary rescheduling

jjpet123:
Hi!

Quick question, it is my understanding that this will fix the "UnresolvedAddressException" we hit in 1.90 when passing proxy hostnames to the system property "https.proxyHost" (we are NOT using GRPC_PROXY_EXP), is that right? Please confirm, thanks!

zpencer:
Yes, this PR fixes the unresolved address issue and allows you to use -DhttpsProxyHost and optionally with a password via `java.net.Authenticator`.

PR #4139, opened 2018-02-26
core: allow application to provide all threads - inprocess channel

dapengzhang0:
The is one of a sequence of PRs to allow application to provide all threads. 
This PR added `InProcess{Channel, Server}Builder.scheduledExecutorService()` API

### DIFF
--- a/core/src/main/java/io/grpc/inprocess/InProcessServer.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessServer.java
@@ -18,14 +18,11 @@ package io.grpc.inprocess;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.common.annotations.VisibleForTesting;
 import io.grpc.ServerStreamTracer;
 import io.grpc.internal.InternalServer;
 import io.grpc.internal.ObjectPool;
 import io.grpc.internal.ServerListener;
 import io.grpc.internal.ServerTransportListener;
-import io.grpc.internal.SharedResourceHolder.Resource;
-import io.grpc.internal.SharedResourcePool;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
```diff
@@ -98,10 +105,31 @@ private InProcessServerBuilder(String name) {
...
+    return this;
+  }
+
   @Override
   protected InProcessServer buildTransportServer(
       List<ServerStreamTracer.Factory> streamTracerFactories) {
-    return new InProcessServer(name, GrpcUtil.TIMER_SERVICE, streamTracerFactories);
+    ObjectPool<ScheduledExecutorService> schedulerPool =
+        scheduledExecutorService == null
+            ? SharedResourcePool.forResource(GrpcUtil.TIMER_SERVICE)
```
ejona86: Use a ObjectPool field instead of ScheduledExecutorService and make this the default value. Then in the `scheduledExecutorService()` method overwrite it with the FixedObjectPool.

You could do this pattern in InProcessChannelBuilder as well.

core/src/main/java/io/grpc/inprocess/InProcessServer.java
```diff
@@ -55,13 +52,6 @@ static InProcessServer findServer(String name) {
    */
```
ejona86: You need to update schedulerPool's documentation.

dapengzhang0: Fixed

core/src/main/java/io/grpc/inprocess/InProcessServerBuilder.java
dapengzhang0: Fixed. The pattern does not provide much gain for InProcessChannelBuilder, and currently it's in the same pattern as NettyChannelBuilder and OkHttpChannelBuilder.

PR #4143, opened 2018-02-26
alts: Remove Java 8 library usages

ejona86:
This doesn't seem that much worse than avoiding Java 8 language features. ~~Is this acceptable, or should I split up the interop-test Blaze target?~~ Discussed elsewhere; this seems good.


jiangtaoli2016:
LGTM

PR #4144, opened 2018-02-26
testing: lazily init default server of GrpcServerRule

dapengzhang0:
In the future, GrpcServerRule will provide custom Server/Channel Builders, and users may not need the default server and channel at all if they only use servers and channels built from the Builders. For backward compatibility, we still have to keep `GrpcServerRule.getServer()` and `GrpcServerRule.getChannel()` API, but we don't have to eagerly start the server all the time.

### DIFF
--- a/testing/src/main/java/io/grpc/testing/GrpcServerRule.java
+++ b/testing/src/main/java/io/grpc/testing/GrpcServerRule.java
@@ -25,20 +25,19 @@ import io.grpc.Server;
 import io.grpc.ServerServiceDefinition;
 import io.grpc.inprocess.InProcessChannelBuilder;
 import io.grpc.inprocess.InProcessServerBuilder;
-import io.grpc.stub.AbstractStub;
 import io.grpc.util.MutableHandlerRegistry;
-import java.util.UUID;
+import java.io.IOException;
 import java.util.concurrent.TimeUnit;
 import org.junit.rules.ExternalResource;
 import org.junit.rules.TestRule;
 
 /**
- * {@code GrpcServerRule} is a JUnit {@link TestRule} that starts an in-process gRPC service with
- * a {@link MutableHandlerRegistry} for adding services. It is particularly useful for mocking out
+ * {@code GrpcServerRule} is a JUnit {@link TestRule} that is particularly useful for mocking out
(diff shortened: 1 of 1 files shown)


ejona86:
> GrpcServerRule will provide custom Server/Channel Builders, and users may not need the default server and channel at all if they only use servers and channels built from the Builders

I'd really like to see some sort of sketch of that API before making changes here. I'm not yet convinced it should be part of GrpcServerRule.

dapengzhang0:
I'd like to add something like the following API to `GrpcServerRule`:

```java
  // registry used for teardown the channels and servers.
  final List<ManagedChannel> managedChannelsRegistry = new ArrayList<ManagedChannel>();
  final List<Server> managedServersRegistry = new ArrayList<Server>();
  
  public ManagedChannelBuilder<?> newChannelBuilder(final String serverName) {
    return new ForwardingChannelBuilder() {
      @Override
      protected ManagedChannelBuilder<?> delegate() {
        return InProcessChannelBuilder.forName(serverName);
      }
      
      @Override
      public ManagedChannel build() {
        ManagedChannel channel = delegate().build();
        managedChannelsRegistry.add(channel);
        return channel;
      }
    };
  }

  public ServerBuilder<?> newServerBuilder(final String serverName) {
    return new ForwardingServerBuilder() {
      @Override
      protected ServerBuilder<?> delegate() {
        return InProcessServerBuilder.forName(serverName);
      }

      @Override
      public Server build() {
        Server server = delegate().build();
        managedServersRegistry.add(server);
        return server;
      }
    };
  }
```

ejona86:
Calling `getChannel()` will then break any server created via `newServerBuilder()`. It seems like it should be a separate rule.

dapengzhang0:
> Calling getChannel() will then break any server created via newServerBuilder(). It seems like it should be a separate rule.

I'm definitely in favor to create a separate rule. Why will `getChannel()` break? It returns the default channel, not any of the channels in the registry.

ejona86:
> Why will getChannel() break? It returns the default channel, not any of the channels in the registry.

Getting the channel implicitly (lazily) creates a _server_, and that server will be listening on the same name as one created via `newServerBuilder()`.

PR #4145, opened 2018-02-26
alts: annotate ALTS public API

jiangtaoli2016:


alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
```diff
@@ -45,6 +46,7 @@
  * ALTS version of {@code ManagedChannelBuilder}. This class sets up a secure and authenticated
  * commmunication between two cloud VMs using ALTS.
  */
+@ExperimentalApi
```
carl-mastrangelo: We typically make a tracking issue for each usage of ExperimentalApi,    You can see it in other usages of the annotation.  It allows us to keep track of when to make it non experimental.

alts/src/main/java/io/grpc/alts/transportsecurity/AltsAuthContext.java
```diff
@@ -17,12 +17,14 @@
...
 import com.google.common.annotations.VisibleForTesting;
+import io.grpc.Internal;
 import io.grpc.alts.Altscontext.AltsContext;
 import io.grpc.alts.Handshaker.HandshakerResult;
 import io.grpc.alts.TransportSecurityCommon.RpcProtocolVersions;
 import io.grpc.alts.TransportSecurityCommon.SecurityLevel;
 
 /** AltsAuthContext contains security-related context information about an ALTs connection. */
+@Internal
 public final class AltsAuthContext {
```
carl-mastrangelo: If this is actually internal, you should make the class package private, and using an Internal accessor.   See some of the classes named Internal* in the core repo.

alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
jiangtaoli2016: Added a tracking issue.

alts/src/main/java/io/grpc/alts/transportsecurity/AltsAuthContext.java
jiangtaoli2016: AltsAuthContext and TsiPeer are not used today, but will be used in the future for authorization check, thus they are public. let me update them as ExperimentalApi

jiangtaoli2016:
@carl-mastrangelo I have revised the code. Not sure whether I need to rename the Internal Public Class. Let me send an email to discuss.

PR #4147, opened 2018-02-27
core: add ServerImpl to channelz

zpencer:


core/src/main/java/io/grpc/internal/ServerCallImpl.java
```diff
@@ -174,6 +178,7 @@ public void close(Status status, Metadata trailers) {
     }
 
     stream.close(status, trailers);
+    serverCallTracer.reportCallEnded(status.isOk());
```
carl-mastrangelo: I believe this should be in a finally, incase the stream closure fails.

PR #4166, opened 2018-03-02
Partially revert "core: Tracking NameResolverProvider being experimen…

ejona86:
…tal"

This partially reverts commit c07ad68cbd4fd836900c166693cdaa2c7613edc0.

The PARAMS_DEFAULT_PORT in NameResorverProvider is the preferred name.
The hope is still to kill NameResolver.Factory.

CC @zpencer


ejona86:
> Does that mean we should replace all usages of NameResolver.Factory.PARAMS_DEFAULT_PORT by NameResolverProvider.PARAMS_DEFAULT_PORT in other places?

It'd be fair to, but that's the easy part of removing the Factory.

PR #4167, opened 2018-03-02
core: Disable ManagedChannelImpl orphan check

ejona86:
Some internal tests are broken and noticing the warning logs. They need
to be fixed before this is enabled.

CC @zpencer, @carl-mastrangelo


zhangkun83:
Have you filed an issue to track it?

ejona86:
Yes. I'm including tracking bug references in the sync CL.

PR #4169, opened 2018-03-02
alts: refactoring into alts and alts.internal packages

jiangtaoli2016:
refactoring alts into two packages: io.grpc.alts and io.grpc.alts.internal. The internal package has limited visibility and is for internal use only.


jiangtaoli2016:
Thanks much for review. I will merge now.

PR #4172, opened 2018-03-05
android: add AndroidChannelBuilder

ericgribkoff:
Introduces a wrapper around an OkHttp channel that accepts an Android `Context`and registers as a network listener to automatically invoke the necessary methods on `ManagedChannel` (`prepareToLoseNetwork`/`enterIdle` and `resetConnectBackoff`) when the device connection state changes.

This will have to be a new artifact with a direct Android dependency: interacting with the Android network monitoring APIs in the existing OkHttp package via reflection could work, but we need to provide the connectivity manager with implementations of Android callback interfaces, which would require using the `java.lang.reflect.Proxy` API to construct these reflectively. It seems much cleaner to just have (yet one more) new artifact for Android-specific functionality.

Manually tested with a local app. Roboelectric tests to follow. This reduces the burden on users to take advantage of the gRPC channel API methods: this can all be accomplished with gRPC's public API, but doing so requires users to (a) keep track of every channel reference and (b) understand the Android network monitoring APIs, which have changed over time, and how to plumb these changes to the respective gRPC channel methods.

cc @srini100


ericgribkoff:
@ejona86 Friendly ping (branch cut for v1.12 is in five days, looks like this is definitely at risk for making the cut now?)

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
```diff
@@ -0,0 +1,255 @@
...
+        networkReceiver = new NetworkReceiver();
+        IntentFilter networkIntentFilter =
+            new IntentFilter(ConnectivityManager.CONNECTIVITY_ACTION);
+        context.registerReceiver(networkReceiver, networkIntentFilter);
+      }
+    }
+
+    private void unregisterNetworkListener() {
+      if (needToUnregisterListener) {
+        synchronized (lock) {
```
ejona86: Do you need to check needToUnregisterListener again within the lock?

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
```diff
@@ -0,0 +1,255 @@
...
+    private final Context context;
+    private final ConnectivityManager connectivityManager;
+
+    private DefaultNetworkCallback defaultNetworkCallback;
+    private NetworkReceiver networkReceiver;
+
+    private final Object lock = new Object();
+
+    // May only go from true to false, and lock must be held when assigning this
+    private volatile boolean needToUnregisterListener = true;
```
ejona86: I'd suggest making this just `@GuardedBy("lock")`. It's only going to be rarely used, because shutdown isn't called frequently. Little benefit in optimizing the avoidance of the lock.

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
```diff
@@ -0,0 +1,255 @@
...
+        IntentFilter networkIntentFilter =
+            new IntentFilter(ConnectivityManager.CONNECTIVITY_ACTION);
+        context.registerReceiver(networkReceiver, networkIntentFilter);
+      }
+    }
+
+    private void unregisterNetworkListener() {
+      if (needToUnregisterListener) {
+        synchronized (lock) {
+          if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N && connectivityManager != null) {
```
ejona86: Does repeating the `Build` check here help dead code elimination? If not, is it necessary?

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
```diff
@@ -0,0 +1,255 @@
...
+   * ManagedChannel#resetConnectBackoff}) when the device network state changes.
+   */
+  @VisibleForTesting
+  static final class AndroidChannel extends ManagedChannel {
+
+    private final ManagedChannel delegate;
+    private final Context context;
+    private final ConnectivityManager connectivityManager;
+
+    private DefaultNetworkCallback defaultNetworkCallback;
```
ejona86: Since these are only used for shut down, you could consider just storing a Runnable for clean-up. It's fine as-is though.

(I'm mainly considering it because of the `Build.VERSION.SDK_INT >= Build.VERSION_CODES.N` check during shut down)

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
```diff
@@ -0,0 +1,255 @@
...
+ */
+@ExperimentalApi("https://github.com/grpc/grpc-java/issues/4056")
+public final class AndroidChannelBuilder extends ForwardingChannelBuilder<AndroidChannelBuilder> {
+
+  private final ManagedChannelBuilder delegateBuilder;
+  private final Context context;
+
+  /** Always fails. Call {@link #forAddress(String, int, Context)} instead. */
+  public static AndroidChannelBuilder forTarget(String target) {
+    throw new UnsupportedOperationException("call forTarget(String, Context) instead");
```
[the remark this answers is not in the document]
ericgribkoff: Done. Context is now optional and set via the context() builder method.

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
```diff
@@ -0,0 +1,255 @@
...
+    @VisibleForTesting
+    AndroidChannel(final ManagedChannel delegate, Context context) {
+      this.delegate = delegate;
+      this.context = context;
+      connectivityManager =
+          (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+
+      // Android N added the registerDefaultNetworkCallback API to listen to changes in the device's
+      // default network. For earlier Android API levels, use the BroadcastReceiver API.
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N && connectivityManager != null) {
```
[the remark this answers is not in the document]
ericgribkoff: These work for libraries - they are statically available constants about the device, not compile-time build constants. Javadoc [here](https://developer.android.com/reference/android/os/Build.VERSION.html#SDK_INT), and these are used in Cronet and in`AndroidPlatform` in [okhttp3](https://github.com/square/okhttp/blob/19726969aabbd0791d315994384a0cc46bf41648/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java#L84).

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
ericgribkoff: Done

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
ericgribkoff: Done

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
ericgribkoff: Yes, good catch - fixed by switch to guardedby

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
ericgribkoff: Refactored to use runnable instead of this conditional

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
```diff
@@ -0,0 +1,291 @@
...
+@ExperimentalApi("https://github.com/grpc/grpc-java/issues/4056")
+public final class AndroidChannelBuilder extends ForwardingChannelBuilder<AndroidChannelBuilder> {
+
+  @Nullable
+  private static final Class<?> OKHTTP_CHANNEL_BUILDER_CLASS =
+      findClass("io.grpc.okhttp.OkHttpChannelBuilder");
+
+  private static final Class<?> findClass(String className) {
+    try {
+      return Class.forName(className);
```
ejona86: You need the literal `Class.forName("io.grpc.okhttp.OkHttpChannelBuilder")`, otherwise ProGuard won't detect the reference.

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
```diff
@@ -0,0 +1,291 @@
...
+        // The connection status may change before registration of the listener is complete, but
+        // this will at worst result in invoking resetConnectBackoff() instead of enterIdle() (or
+        // vice versa) on the first network change.
+        boolean isConnected = currentNetwork != null && currentNetwork.isConnected();
+
+        defaultNetworkCallback = new DefaultNetworkCallback(isConnected);
+        connectivityManager.registerDefaultNetworkCallback(defaultNetworkCallback);
+        unregisterRunnable =
+            new Runnable() {
+              @TargetApi(Build.VERSION_CODES.LOLLIPOP)
```
ejona86: I'd like to understand these annotations at some point.

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
```diff
@@ -0,0 +1,291 @@
...
+   */
+  @VisibleForTesting
+  static final class AndroidChannel extends ManagedChannel {
+
+    private final ManagedChannel delegate;
+
+    @Nullable private final Context context;
+    @Nullable private final ConnectivityManager connectivityManager;
+
+    @Nullable private DefaultNetworkCallback defaultNetworkCallback;
```
ejona86: I had sort of imagined these two fields would just be local variables, since the Runnables could just get a closure of them. This is fine too.

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
ericgribkoff: Done

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
ericgribkoff: Analogous to SuppressWarnings. Without it there are linter errors about unregisterNetworkCallback requiring Android L in the runnable.

android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
ericgribkoff: Done

ericgribkoff:
@ejona86 Made a few small changes to address comments and catch the security exception thrown for apps that don't have the ACCESS_NETWORK_STATE permission. This should be ready to merge now, PTAL

Issue #3585, opened 2017-10-18
Bazel java_grpc_library(flavor="lite") doesn't work out-of-the-box


fis:
### What version of gRPC are you using?
Git master (@8187f1e)

### What JVM are you using (`java -version`)?
1.8.0_144

### What did you do?
Configured a Bazel project with `@grpc_java` pointing at this codebase (and protobuf deps as per the Bazel native protobuf rules, pointing at latest protobuf code), added a `java_grpc_library(..., flavor = "lite")` rule.

### What did you expect to see?
Successfully generated & compiled code from `bazel build`.

### What did you see instead?
A number of different build problems, starting with `no such package '@com_google_protobuf_java_lite//'`. Details for working around them below.

---

I've managed to successfully build an Android client using a lite-proto java_grpc_library, but it involved doing the following changes:

#### Adjust deps added by the `java_grpc_library` macro

```diff
--- a/java_grpc_library.bzl
+++ b/java_grpc_library.bzl
@@ -109,14 +109,19 @@ def java_grpc_library(name, srcs, deps, flavor=None,
   added_deps = [
       "@grpc_java//core",
       "@grpc_java//stub",
-      "@grpc_java//protobuf",
       "@com_google_guava_guava//jar",
   ]
   if flavor == "normal":
-    added_deps += ["@com_google_protobuf_java//:protobuf_java"]
+    added_deps += [
+        "@com_google_protobuf_java//:protobuf_java",
+        "@grpc_java//protobuf",
+    ]
   elif flavor == "lite":
     # TODO: This is currently blocked on https://github.com/google/protobuf/issues/2762
-    added_deps += ["@com_google_protobuf_java_lite//:protobuf_java_lite"]
+    added_deps += [
+        "@com_google_protobuf_javalite//:protobuf_java_lite",
+        "@grpc_java//protobuf-lite:protobuf_lite",
+    ]
   else:
     fail("Unknown flavor type", "flavor")
```

This removes `@grpc_java//protobuf` from being added unconditionally, as it was pulling in the non-lite protobuf runtime into the resulting APK. It's replaced by `@grpc_java//protobuf-lite:protobuf_lite` instead. It also renames the javalite remote repository from `@com_google_protobuf_java_lite` to `@com_google_protobuf_javalite`, because the latter name seems more "canonical", in the sense that it's what Bazel native `java_lite_proto_library` rules implicitly depend on.

[...]

PR #4173, opened 2018-03-06
enable lite flavor bazel rules

codesuki:
resolves https://github.com/grpc/grpc-java/issues/3585
Adding the changes described in the above issue.
I read somewhere that protobuf-java is scrapping lite support, 
so if this PR is useless please just close it. 
For now we successfully compile a grpc android app with that.

### DIFF
--- a/java_grpc_library.bzl
+++ b/java_grpc_library.bzl
@@ -45,7 +45,7 @@ _gensource = rule(
         "flavor": attr.string(
             values = [
                 "normal",
-                "lite",  # Not currently supported
+                "lite",
             ],
             default = "normal",
         ),
@@ -109,14 +109,18 @@ def java_grpc_library(name, srcs, deps, flavor=None,
   added_deps = [
       "@grpc_java//core",
       "@grpc_java//stub",
-      "@grpc_java//protobuf",
       "@com_google_guava_guava//jar",
   ]
   if flavor == "normal":
-    added_deps += ["@com_google_protobuf//:protobuf_java"]
+    added_deps += [
+        "@com_google_protobuf//:protobuf_java",
+        "@grpc_java//protobuf",
+    ]
   elif flavor == "lite":
-    # TODO: This is currently blocked on https://github.com/google/protobuf/issues/2762
-    added_deps += ["@com_google_protobuf_java_lite//:protobuf_java_lite"]
+    added_deps += [
+        "@com_google_protobuf_javalite//:protobuf_java_lite",
+        "@grpc_java//protobuf-lite:protobuf_lite",
+    ]
   else:
     fail("Unknown flavor type", "flavor")
 
--- a/protobuf-lite/BUILD.bazel
+++ b/protobuf-lite/BUILD.bazel
@@ -3,12 +3,11 @@ java_library(
     srcs = glob([
         "src/main/java/**/*.java",
     ]),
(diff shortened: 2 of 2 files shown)


java_grpc_library.bzl
```diff
@@ -109,14 +109,18 @@ def java_grpc_library(name, srcs, deps, flavor=None,
...
       "@grpc_java//core",
       "@grpc_java//stub",
-      "@grpc_java//protobuf",
       "@com_google_guava_guava//jar",
   ]
   if flavor == "normal":
-    added_deps += ["@com_google_protobuf//:protobuf_java"]
+    added_deps += [
+        "@com_google_protobuf//:protobuf_java",
+        "@grpc_java//protobuf",
```
jyane: Here is already removed (should not depend on workspace, be like a `//protobuf`) and workspace was renamed to `io_grpc_grpc_java`, please check your branch is up to date.
https://github.com/grpc/grpc-java/blob/ae1fb9467cf65ddb8830f381c073ca1597a8a894/WORKSPACE#L1

codesuki: Thanks! I forgot to rebase on master. Now I have to fix my build first because using master broke it somehow. I'll send an update tomorrow.

protobuf-lite/BUILD.bazel
```diff
@@ -3,12 +3,11 @@ java_library(
...
     ]),
-    # TOOD(zdapeng): fix visibility and deps (https://github.com/google/protobuf/issues/2762)
-    visibility = ["//protobuf:__pkg__"],
+    visibility = ["//visibility:public"],
     deps = [
         "//core",
         "@com_google_code_findbugs_jsr305//jar",
         "@com_google_guava_guava//jar",
-        "@com_google_protobuf//:protobuf_java",
+        "@com_google_protobuf_javalite//:protobuf_java_lite",
```
ejona86: So the problem is that this breaks non-lite; you can't mix the two runtimes. We need to have a `select()` to auto-choose the correct dependency (lite on android, otherwise full protobuf) and I've not had a chance to get back and try some options.

ejona86:
I've not heard about scrapping 'lite'. 'nano' should be avoided, but 'lite' is still good to my knowledge.

codesuki:
That would be good! Maybe I mixed something up.

protobuf-lite/BUILD.bazel
codesuki: I didn't think it would use both runtimes, because it's the `protobuf-lite` folder. 
But there is a possible conflict here 
https://github.com/grpc/grpc-java/blob/master/java_grpc_library.bzl#L112 
https://github.com/grpc/grpc-java/blob/master/protobuf/BUILD.bazel#L9
But I have not much experience with this repository, so not sure, so I'll leave this to you.
We decided to move away from lite to have any support.

codesuki:
Got confused by this https://github.com/google/protobuf/issues/4314
But it just says that option is removed, not lite itself.

ejona86:
@codesuki, Hmm... So one approach that I know how to do now is to split any targets depending on protobuf into a "normal" and a `_android`-suffixed target. I was trying to avoid that though, because it causes a rift and "you can't cross the streams." But it would get functional support, and I guess the rift wouldn't be that bad today.

Would you mind updating this PR to do that (instead of modifying protobuf_lite create a new protobuf_lite_android target)? Or I can.

codesuki:
Sorry for the late reply, when I find the time I will add it.

PR #4177, opened 2018-03-06
POC: add PostBuildOption to keep reference of server/channel for test Rule

dapengzhang0:
The current GrpcServerRule does not provide Server/Channel Builder, so it is not possible to provider a lot of useful builder options such as setting ScheduledExecutorService. To provide Server/Channel Builder, if requiring that builder can be configured differently in each test method, the Rule need be able to keep references of server/channel being built in order to manage their lifecycle.


dapengzhang0:
Close in favor of #4330

Issue #4157, opened 2018-03-01
fix regression in v1.9.0 where using GRPC_PROXY_EXP with the Netty transport would cause an UnresolvedAddressException is not available in 1.10.0


arief-hidayat:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?
1.10.0

### What did you expect to see?
The following is listed in the release note (
https://github.com/grpc/grpc-java/releases/tag/v1.10.0)
netty: fix regression in v1.9.0 where using GRPC_PROXY_EXP with the Netty transport would cause an UnresolvedAddressException (#4027).

However, i still face the same issue and fix seems not included in this 1.10.0 release.

ejona86:
Does the fix in 1.9.1 work for you? The problem is that it missed the v1.10.x branch cut and failed to be backported, right? (I see the change on master.)

PR #4178, opened 2018-03-07
core: Resolve address of GRPC_PROXY_EXP hostname (1.10 backport)

zpencer:
Since this address is never resolved gRPC fails to connect with an
UnresolvedAddressException when the env variable is specified. This
should resolve the address before we attempt to proxy to it.


ejona86:
This is a backport of #4027

ejona86:
This will fix #4157

PR #4183, opened 2018-03-07
interop-testing: Improve failure messages for ping_ping

ejona86:
Timeouts and Status returns can be common, and previously those cases
weren't that clear. For example, if there was a StatusRuntimeException
it would just print the status code and message, but not the causal
chain.

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -703,12 +703,12 @@ public abstract class AbstractInteropTest {
     for (int i = 0; i < requests.size(); i++) {
       assertNull(queue.peek());
       requestObserver.onNext(requests.get(i));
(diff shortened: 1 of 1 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -703,8 +703,12 @@ public void onCompleted() {
     for (int i = 0; i < requests.size(); i++) {
       assertNull(queue.peek());
       requestObserver.onNext(requests.get(i));
-      assertEquals(goldenResponses.get(i),
-                   queue.poll(operationTimeoutMillis(), TimeUnit.MILLISECONDS));
+      Object o = queue.poll(operationTimeoutMillis(), TimeUnit.MILLISECONDS);
```
dapengzhang0: nit: `Object actualResponse`

ejona86: Done

PR #4185, opened 2018-03-07
SECURITY.md: Add reference to grpc-netty-shaded

ejona86:
Many people going to the troubleshooting section would be fixed by
swapping to grpc-netty-shaded, so we should call that out.

@saturnism, would you mind reviewing?


ejona86:
@saturnism, could you do the GitHub review and approve?

PR #4186, opened 2018-03-07
core: split JSON parsing into its own file

carl-mastrangelo:
This is part 1 of https://github.com/grpc/grpc-java/commit/b01609572ab4f0dba6eae23fe5be05a302ecefc7

Next will be re validating the service config, and part 3 will be enforcing it in MCI.

### DIFF
--- a/core/src/main/java/io/grpc/internal/JsonParser.java
+++ b/core/src/main/java/io/grpc/internal/JsonParser.java
@@ -54,25 +54,24 @@ final class JsonParser {
   }
 
   private static Object parseRecursive(JsonReader jr) throws IOException {
-    while (jr.hasNext()) {
-      switch (jr.peek()) {
-        case BEGIN_ARRAY:
-          return parseJsonArray(jr);
-        case BEGIN_OBJECT:
-          return parseJsonObject(jr);
-        case STRING:
-          return jr.nextString();
-        case NUMBER:
-          return jr.nextDouble();
-        case BOOLEAN:
-          return jr.nextBoolean();
-        case NULL:
-          return parseJsonNull(jr);
-        default:
-          throw new IllegalStateException("Bad token: " + jr.getPath());
-      }
+    checkState(jr.hasNext(), "unexpected end of JSON");
+    checkState(jr.hasNext(), "unexpected end of JSON");
+    switch (jr.peek()) {
+      case BEGIN_ARRAY:
+        return parseJsonArray(jr);
+      case BEGIN_OBJECT:
+        return parseJsonObject(jr);
+      case STRING:
+        return jr.nextString();
+      case NUMBER:
+        return jr.nextDouble();
+      case BOOLEAN:
+        return jr.nextBoolean();
+      case NULL:
+        return parseJsonNull(jr);
+      default:
+        throw new IllegalStateException("Bad token: " + jr.getPath());
     }
-    throw new IllegalStateException("unexpected end of JSON");
   }
 
   private static Map<String, Object> parseJsonObject(JsonReader jr) throws IOException {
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/grpc/internal/JsonParser.java
```diff
@@ -0,0 +1,147 @@
...
+      try {
+        jr.close();
+      } catch (IOException e) {
+        logger.log(Level.WARNING, "Failed to close", e);
+      }
+    }
+  }
+
+  private static Object parseRecursive(JsonReader jr) throws IOException {
+    while (jr.hasNext()) {
```
dapengzhang0: The while loop is actually not looping. So instead of the while loop and  `throw new IllegalStateException("unexpected end of JSON")` at the end, you can check argument instead:

```java
private static Object parseRecursive(JsonReader jr) throws IOException {
  checkArgument(jr.hasNext(), "unexpected end of JSON");
  switch (jr.peek()) {
    case BEGIN_ARRAY:
        return parseJsonArray(jr);
    case BEGIN_OBJECT:
        return parseJsonObject(jr);
    case STRING:
       return jr.nextString();
    case NUMBER:
        return jr.nextDouble();
    case BOOLEAN:
        return jr.nextBoolean();
    case NULL:
        return parseJsonNull(jr);
    default:
        throw new IllegalStateException("Bad token: " + jr.getPath());
  }
}
```

core/src/test/java/io/grpc/internal/JsonParserTest.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+/**
+ * Tests for {@link JsonParser}.
+ */
+public class JsonParserTest {
```
dapengzhang0: `@RunWith`

core/src/test/java/io/grpc/internal/JsonParserTest.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  @Test
+  public void emptyArray() throws IOException {
+    assertEquals(new ArrayList<Object>(), JsonParser.parse("[]"));
+  }
+
+  @Test
+  public void intNumber() throws IOException {
+    assertEquals(1.0, JsonParser.parse("1"));
```
dapengzhang0: `assertEquals(double, double)` is deprecated.

Maybe `assertEquals(Double.valueOf("1"), (Double) JsonParser.parse("1"))`?

Ditto.

core/src/test/java/io/grpc/internal/JsonParserTest.java
```diff
@@ -0,0 +1,105 @@
...
+  public void objectNoName() throws IOException {
+    thrown.expect(MalformedJsonException.class);
+
+    JsonParser.parse("{: 1");
+  }
+
+  @Test
+  public void objectStringName() throws IOException {
+    LinkedHashMap<String, Object> expected = new LinkedHashMap<String, Object>();
+    expected.put("hi", 2.0);
```
dapengzhang0: Double.valueOf("2")

core/src/test/java/io/grpc/internal/JsonParserTest.java
```diff
@@ -0,0 +1,105 @@
...
+  @Test
+  public void doubleNumber() throws IOException {
+    assertEquals(1.2, JsonParser.parse("1.2"));
+  }
+
+  @Test
+  public void longNumber() throws IOException {
+    assertEquals(9999999999.0, JsonParser.parse("9999999999"));
+  }
+
```
dapengzhang0: Also add tests for json boolean, null, NaN?

PR #4187, opened 2018-03-08
alts: minor update to ALTS handshaker proto

jiangtaoli2016:
We are synchronizing the proto change across 3 languages.


jyane:
Re-generating a java file is required?

jiangtaoli2016:
Good point. I thought comments in proto will be ignored in generated files.

PR #4195, opened 2018-03-09
Ignore proto-generated code for ErrorProne

ejona86:
Previously if protobuf-generated code triggered ErrorProne we'd have to disable
the failing check for all code in that task. With -XepExcludedPaths we can
disable the ErrorProne-checking just for protobuf. Note that we continue using
ErrorProne on our generated code. Also note this only applies to ErrorProne
checks; JDK checks still require task-level disabling.


dapengzhang0:
Do we have to exclude ServiceGrpc.java?

ejona86:
I'm purposefully not excluding our generated code. We tend to want to see ErrorProne warnings in it and fix them.

PR #4196, opened 2018-03-09
core: fix channel.getState(true) will not reconnect

### LINKED ISSUE #4170: channel.getState(true) will not reconnect
The javadoc for ManagedChannel.getState:
_Gets the current connectivity state. Note the result may soon become outdated.
Parameters:requestConnection if true, the channel will try to make a connection if it is currently IDLE_

However, it attempts to connect only if there has not been an open connection already. After disconnect, channel.getState(true) will not attempt to reconnect (state is IDLE).

1.9.0

channel.getState(true) should try to reconnect the channel if the state is idle.

dapengzhang0:
resolves #4170

### DIFF
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -821,6 +821,12 @@ final class ManagedChannelImpl extends ManagedChannel implements Instrumented<Ch
             @Override
             public void run() {
               exitIdleMode();
+              if (subchannelPicker != null) {
+                if (shutdown.get() || panicMode) {
+                  return;
+                }
+                subchannelPicker.requestConnection();
+              }
             }
           }).drain();
     }
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -822,9 +822,6 @@ final class ManagedChannelImpl extends ManagedChannel implements Instrumented<Ch
             public void run() {
               exitIdleMode();
               if (subchannelPicker != null) {
-                if (shutdown.get() || panicMode) {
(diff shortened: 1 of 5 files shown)


zhangkun83:
@ejona86, I proposed this solution. Does it sound good to you?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -821,6 +821,12 @@ public ConnectivityState getState(boolean requestConnection) {
             @Override
             public void run() {
               exitIdleMode();
+              if (subchannelPicker != null) {
+                if (shutdown.get() || panicMode) {
```
ejona86: This looks unnecessary. I would really hope this is unnecessary. We don't want to be checking these all over the place.

zhangkun83: We can probably instead check channelStateManager.getState() again here.

ejona86: If `shutdownNameResolverAndLoadBalancer()` has been called, which is done by both shutdown and panic, then `subchannelPicker` will have been cleared. If in `panicMode`, it will point to a `panicPickResult` which should do nothing. There's no risk in using the `subchannelPicker` incorrectly now, and that should be maintained in the future.

Code _setting_ subchannelPicker could reasonably need to look at those fields. But readers shouldn't need to, as a rule.

zhangkun83: `shutdownNameResolverAndLoadBalancer()` is called when the channel is terminating, not immediately after `shutdown()`. That said, even when state == SHUTDOWN and `subchannelPicker` is not null, there is no harm calling `requestConnection()`. Yes I agree we can remove this `if`.

PR #4200, opened 2018-03-09
core: keep root channels separate from subchannels in channelz

zpencer:
It is a channelz requirement that subchannels and channels are
separate, and we can not retrieve a root channel by asking for a
subchannel and vice versa.

### DIFF
--- a/core/src/main/java/io/grpc/internal/Channelz.java
+++ b/core/src/main/java/io/grpc/internal/Channelz.java
@@ -33,7 +33,7 @@ public final class Channelz {
       new ConcurrentHashMap<Long, Instrumented<ServerStats>>();
   private final ConcurrentMap<Long, Instrumented<ChannelStats>> rootChannels =
       new ConcurrentHashMap<Long, Instrumented<ChannelStats>>();
-  private final ConcurrentMap<Long, Instrumented<ChannelStats>> channels =
+  private final ConcurrentMap<Long, Instrumented<ChannelStats>> subchannels =
       new ConcurrentHashMap<Long, Instrumented<ChannelStats>>();
   private final ConcurrentMap<Long, Instrumented<SocketStats>> sockets =
       new ConcurrentHashMap<Long, Instrumented<SocketStats>>();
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/grpc/internal/Channelz.java
```diff
@@ -33,7 +33,7 @@
       new ConcurrentHashMap<Long, Instrumented<ServerStats>>();
   private final ConcurrentMap<Long, Instrumented<ChannelStats>> rootChannels =
       new ConcurrentHashMap<Long, Instrumented<ChannelStats>>();
-  private final ConcurrentMap<Long, Instrumented<ChannelStats>> channels =
+  private final ConcurrentMap<Long, Instrumented<ChannelStats>> subchannels =
```
carl-mastrangelo: why not keep them all in the same list?

core/src/main/java/io/grpc/internal/Channelz.java
```diff
@@ -33,7 +33,7 @@
       new ConcurrentHashMap<Long, Instrumented<ServerStats>>();
   private final ConcurrentMap<Long, Instrumented<ChannelStats>> rootChannels =
```
carl-mastrangelo: fyi, this will need to be scannable so CHM will not be a long term solution.

core/src/main/java/io/grpc/internal/Channelz.java
zpencer: I think we won't want a single navigable concurrent map for all channels and subchannels. We want to be fast when looking for all root channels and avoid having to scan through subchannels. Since we can't get rid of the special root channels navigable map, and the fact that getChannel is different from getSubchannel, I think mixing them won't give us much.

core/src/main/java/io/grpc/internal/Channelz.java
zpencer: I realized this when implementing the channelz service, but I wanted to keep this PR focused on one change.

PR #4205, opened 2018-03-10
core,services: add ChannelzService class

zpencer:
This implements the class to expose the stats as
protos. GetServerSockets is still unimplemented and will require a
follow up change.

This PR is large but more than half of the LOC are tests.

### DIFF
--- a/services/src/main/java/io/grpc/services/ChannelzProtoUtil.java
+++ b/services/src/main/java/io/grpc/services/ChannelzProtoUtil.java
@@ -31,6 +31,8 @@ import io.grpc.channelz.v1.Channel;
 import io.grpc.channelz.v1.ChannelData;
 import io.grpc.channelz.v1.ChannelData.State;
 import io.grpc.channelz.v1.ChannelRef;
+import io.grpc.channelz.v1.GetServersResponse;
+import io.grpc.channelz.v1.GetTopChannelsResponse;
 import io.grpc.channelz.v1.Server;
 import io.grpc.channelz.v1.ServerData;
 import io.grpc.channelz.v1.ServerRef;
@@ -41,6 +43,8 @@ import io.grpc.channelz.v1.Subchannel;
 import io.grpc.channelz.v1.SubchannelRef;
 import io.grpc.internal.Channelz;
 import io.grpc.internal.Channelz.ChannelStats;
+import io.grpc.internal.Channelz.RootChannelList;
+import io.grpc.internal.Channelz.ServerList;
 import io.grpc.internal.Channelz.ServerStats;
 import io.grpc.internal.Channelz.SocketStats;
 import io.grpc.internal.Channelz.TransportStats;
@@ -215,6 +219,26 @@ final class ChannelzProtoUtil {
       subchannelBuilder.addSubchannelRef(toSubchannelRef(childSubchannel));
     }
     return subchannelBuilder.build();
+  }
+
+  static GetTopChannelsResponse toGetTopChannelResponse(RootChannelList rootChannels) {
+    GetTopChannelsResponse.Builder responseBuilder = GetTopChannelsResponse
+        .newBuilder()
+        .setEnd(rootChannels.end);
+    for (Instrumented<ChannelStats> c : rootChannels.channels) {
+      responseBuilder.addChannel(ChannelzProtoUtil.toChannel(c));
+    }
+    return responseBuilder.build();
+  }
+
+  static GetServersResponse toGetServersResponse(ServerList servers) {
+    GetServersResponse.Builder responseBuilder = GetServersResponse
+        .newBuilder()
+        .setEnd(servers.end);
+    for (Instrumented<ServerStats> s : servers.servers) {
+      responseBuilder.addServer(ChannelzProtoUtil.toServer(s));
+    }
+    return responseBuilder.build();
   }
 
   private static <T> T getFuture(ListenableFuture<T> future) {
--- a/services/src/main/java/io/grpc/services/ChannelzProtoUtil.java
+++ b/services/src/main/java/io/grpc/services/ChannelzProtoUtil.java
@@ -17,10 +17,12 @@
 package io.grpc.services;
 
 import com.google.common.base.Preconditions;
+import com.google.common.util.concurrent.ListenableFuture;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Int64Value;
 import com.google.protobuf.util.Timestamps;
 import io.grpc.ConnectivityState;
+import io.grpc.Status;
 import io.grpc.channelz.v1.Address;
 import io.grpc.channelz.v1.Address.OtherAddress;
 import io.grpc.channelz.v1.Address.TcpIpAddress;
@@ -88,9 +90,8 @@ final class ChannelzProtoUtil {
         .build();
   }
 
-  static Server toServer(Instrumented<ServerStats> obj)
-      throws ExecutionException, InterruptedException {
-    ServerStats stats = obj.getStats().get();
+  static Server toServer(Instrumented<ServerStats> obj) {
+    ServerStats stats = getFuture(obj.getStats());
     return Server
         .newBuilder()
         .setRef(toServerRef(obj))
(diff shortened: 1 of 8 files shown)


services/src/main/java/io/grpc/services/ChannelzProtoUtil.java
```diff
@@ -0,0 +1,226 @@
...
+    Address.Builder builder = Address.newBuilder();
+    if (address instanceof InetSocketAddress) {
+      InetSocketAddress inetAddress = (InetSocketAddress) address;
+      builder.setTcpipAddress(
+          TcpIpAddress
+              .newBuilder()
+              .setIpAddress(
+                  ByteString.copyFrom(inetAddress.getAddress().getAddress()))
+              .build());
+    } else if (address.getClass().getName().equals("io.netty.channel.unix.DomainSocketAddress")) {
```
carl-mastrangelo: This should be an `endsWith` to account for shading.

services/src/main/java/io/grpc/services/ChannelzProtoUtil.java
```diff
@@ -0,0 +1,226 @@
...
+        .setState(toState(stats.state))
+        .setCallsStarted(stats.callsStarted)
+        .setCallsSucceeded(stats.callsSucceeded)
+        .setCallsFailed(stats.callsFailed)
+        .setLastCallStartedTimestamp(Timestamps.fromMillis(stats.lastCallStartedMillis))
+        .build();
+  }
+
+  static State toState(ConnectivityState state) {
+    if (state == ConnectivityState.CONNECTING) {
```
carl-mastrangelo: Enum.valueOf(State.class, state.name());

services/src/main/java/io/grpc/services/ChannelzService.java
```diff
@@ -0,0 +1,167 @@
...
+import io.grpc.channelz.v1.GetTopChannelsResponse;
+import io.grpc.internal.Channelz;
+import io.grpc.internal.Channelz.ChannelStats;
+import io.grpc.internal.Channelz.ServerList;
+import io.grpc.internal.Channelz.ServerStats;
+import io.grpc.internal.Channelz.SocketStats;
+import io.grpc.internal.Instrumented;
+import io.grpc.stub.StreamObserver;
+
+public class ChannelzService extends ChannelzGrpc.ChannelzImplBase {
```
carl-mastrangelo: final

services/src/main/java/io/grpc/services/ChannelzService.java
```diff
@@ -0,0 +1,167 @@
...
+  private static final int DEFAULT_MAX_PAGE_SIZE = 25;
+  private final Channelz channelz;
+  private final int maxPageSize;
+
+  public ChannelzService() {
```
carl-mastrangelo: since this is public api, lets make this a factory.

services/src/main/java/io/grpc/services/ChannelzService.java
```diff
@@ -0,0 +1,167 @@
...
+          = channelz.getRootChannels(request.getStartChannelId(), maxPageSize);
+      GetTopChannelsResponse.Builder responseBuilder = GetTopChannelsResponse
+          .newBuilder()
+          .setEnd(rootChannels.end);
+      for (Instrumented<ChannelStats> c : rootChannels.channels) {
+        responseBuilder.addChannel(ChannelzProtoUtil.toChannel(c));
+      }
+      responseObserver.onNext(responseBuilder.build());
+      responseObserver.onCompleted();
+    } catch (Throwable t) {
```
carl-mastrangelo: let this propagate

services/src/main/java/io/grpc/services/ChannelzService.java
```diff
@@ -0,0 +1,167 @@
...
+        return;
+      }
+
+      responseObserver.onNext(
+          GetChannelResponse
+              .newBuilder()
+              .setChannel(ChannelzProtoUtil.toChannel(s))
+              .build());
+      responseObserver.onCompleted();
+    } catch (Throwable t) {
```
carl-mastrangelo: same here.  It would be better to fail hard, rather than silently.

services/src/main/java/io/grpc/services/ChannelzService.java
```diff
@@ -0,0 +1,167 @@
...
+      responseObserver.onError(t);
+    }
+  }
+
+  @Override
+  public void getServerSockets(
+      GetServerSocketsRequest request, StreamObserver<GetServerSocketsResponse> responseObserver) {
+    // TODO(zpencer): fill this one out after refactoring channelz class
+    throw new UnsupportedOperationException();
```
carl-mastrangelo: Status.UNIMPLEMENTED.asRuntimeException() ?

services/src/test/java/io/grpc/services/ChannelzProtoUtilTest.java
```diff
@@ -0,0 +1,385 @@
...
+  }
+
+  @Test
+  public void toSocketRef() {
+    assertEquals(socketRef, ChannelzProtoUtil.toSocketRef(socket));
+  }
+
+  @Test
+  public void toState() {
+    assertEquals(State.CONNECTING, ChannelzProtoUtil.toState(ConnectivityState.CONNECTING));
```
carl-mastrangelo: this should be a for loop

services/src/test/java/io/grpc/services/ChannelzProtoUtilTest.java
```diff
@@ -0,0 +1,385 @@
...
+  @Test
+  public void toSocketData() {
+    assertEquals(
+        socketData,
+        ChannelzProtoUtil.toSocketData(socket.transportStats));
+  }
+
+  @Test
+  public void toAddress_inet() throws Exception {
+    InetSocketAddress inet4 = new InetSocketAddress(Inet4Address.getByName("10.0.0.1"), 1000);
```
carl-mastrangelo: does this work for inprocess?

services/src/main/java/io/grpc/services/ChannelzProtoUtil.java
zpencer: done

services/src/main/java/io/grpc/services/ChannelzProtoUtil.java
zpencer: done

services/src/main/java/io/grpc/services/ChannelzService.java
zpencer: done

services/src/main/java/io/grpc/services/ChannelzService.java
zpencer: good point about public api, I've also added `@ExperimentalApi` to the class

services/src/main/java/io/grpc/services/ChannelzService.java
zpencer: Since we are unwrapping `ListenableFuture`, I will need to catch the exceptions somewhere. I moved it into `ChannelzProtoUtil` to just throw a StatusRuntimeException if any exceptions are thrown.

services/src/main/java/io/grpc/services/ChannelzService.java
zpencer: done

services/src/main/java/io/grpc/services/ChannelzService.java
zpencer: done

services/src/test/java/io/grpc/services/ChannelzProtoUtilTest.java
zpencer: done

services/src/test/java/io/grpc/services/ChannelzProtoUtilTest.java
zpencer: `InProcessTransport` does not keep stats right now, but it will eventually. Specifically for sockets, `ChannelzProtoUtil.toAddress` will turn non UDS, non inet sockets into unknown sockets. So we just need to make sure `InProcessSocketAddress` has a properly working `toString`.

zpencer:
Moved some code out of `ChannelzService` so that all heavy lifting is done by either `ChannelzProtoUtil` or `Channelz` itself. This is needed to test `GetServerSockets` more easily, because that service method has some interesting edge cases that are easier to unit test at the low level.

PR #4213, opened 2018-03-13
services: fix binary logging empty metadata NPE

zpencer:
Also added test to cover this case, and renamed test variables
to make it more obvious in the future that the default metadata
is non empty.

### DIFF
--- a/services/src/main/java/io/grpc/services/BinaryLog.java
+++ b/services/src/main/java/io/grpc/services/BinaryLog.java
@@ -416,10 +416,10 @@ final class BinaryLog {
   static io.grpc.binarylog.Metadata metadataToProto(Metadata metadata, int maxHeaderBytes) {
     Preconditions.checkState(maxHeaderBytes >= 0);
     Builder builder = io.grpc.binarylog.Metadata.newBuilder();
-    if (maxHeaderBytes > 0) {
-      byte[][] serialized = InternalMetadata.serialize(metadata);
+    // This code is tightly coupled with Metadata's implementation
+    byte[][] serialized;
+    if (maxHeaderBytes > 0 && (serialized = InternalMetadata.serialize(metadata)) != null) {
       int written = 0;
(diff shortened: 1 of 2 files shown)


services/src/test/java/io/grpc/services/BinaryLogTest.java
```diff
@@ -371,6 +371,15 @@ public void socketToProto_unknown() throws Exception {
         BinaryLog.socketToProto(unknownSocket));
   }
 
+  @Test
+  public void metadataToProto_empty() throws Exception {
+    assertEquals(
+        io.grpc.binarylog.Metadata
```
carl-mastrangelo: `io.grpc.binarylog.Metadata.getDefaultInstance()`

services/src/main/java/io/grpc/services/BinaryLog.java
```diff
@@ -416,10 +416,10 @@ static Peer socketToProto(SocketAddress address) {
   static io.grpc.binarylog.Metadata metadataToProto(Metadata metadata, int maxHeaderBytes) {
     Preconditions.checkState(maxHeaderBytes >= 0);
     Builder builder = io.grpc.binarylog.Metadata.newBuilder();
-    if (maxHeaderBytes > 0) {
-      byte[][] serialized = InternalMetadata.serialize(metadata);
+    // This code is tightly coupled with Metadata's implementation
+    byte[][] serialized;
+    if (maxHeaderBytes > 0 && (serialized = InternalMetadata.serialize(metadata)) != null) {
```
sonyakc: suggestion: I'd add a null check for ```metadata``` before setting ```serialized ```

zpencer: Good idea

services/src/test/java/io/grpc/services/BinaryLogTest.java
zpencer: done

Issue #1788, opened 2016-05-03
Tracking Issue for (Call|ServerCall)StreamObserver being Experimental.


carl-mastrangelo:
Specific usages:
- `CallStreamObserver`
- `ServerCallStreamObserver`

dapengzhang0:
#1848

dapengzhang0:
#1848

dapengzhang0:
#1848 Tracking Health Service APIs being Experimental #1848

wu-sheng:
The `ClientCallStreamObserver` and `CallStreamObserver` have be annotated by `@ExperimentalApi`. Can you tell me why?

Is the api good enough to use on product env. ? This API is very useful for our project. But I need *grpc-team* to confirm.

Thanks.  @ejona86

ejona86:
The APIs are functional but we still reserve the right to change the API. Once there have been some users and agreement that the APIs are working well then we would remove the annotation.

I'm not quite settled with `disableAutoInboundFlowControl`, for example, because it still has an implicit `request()` when the RPC starts. The basic functionally it provides is likely to stay (since flow control is essential) but the precise form can change.

vkedia:
@ejona86 Do you have any further updates on making this non experimental? We use this API in the cloud spanner client library and it has worked well for us so far. But this being experimental prevents our client from going GA.

ejona86:
@vkedia, no I don't. This depends on stabilizing ClientCallStreamObserver and I think there are some issues that need to be resolved for client vs server APIs.

ejona86:
API discussion notes for disableAutoInboundFlowControl:
General agreement that a different name would make the API easier to understand and worth converting existing users. Although not important enough by itself, removing the special case on unary would be good. Since we're wanting a new name, also remove unary special case.

Names:
disableAutoInboundFlowControl
disableAutoRequest +1 +1
disableImplicitRequest
useExplicitRequest
setAutoRequest(false) +1 +1

I'm interested in eventually adding the auto/implicit request() calls to ClientCall/ServerCall (but it would be opt-in). That influenced people's preference for name (e.g., moved a vote from useExplicitRequest to setAutoRequest)
useImplicitRequest
enableAutoRequest
enableImplicitRequest
setAutoRequest(true)

No clear winner. Will loop back around to it.

carl-mastrangelo:
requestOnMessage?

ejona86:
API discussion notes:
[...]

PR #4219, opened 2018-03-14
stub: no initial request when auto inbound flow control is disabled

barnardb:
This commit prevents ClientCalls from making an initial request for
response messages when the user provides a ClientResponseObserver that
has disabled auto inbound flow control by calling
ClientCallStreamObserver.disableAutoInboundFlowControl() in beforeStart.

The previous behaviour was confusing and undersirable for observers
trying to implement their own flow control,
as values were produced before they were requested.
This behaviour was mentioned in
https://github.com/grpc/grpc-java/issues/1788#issuecomment-260898948

The initial requests are still made for ResponseObservers that haven't
disabled auto inbound flow control.

### DIFF
--- a/services/src/test/java/io/grpc/protobuf/services/ProtoReflectionServiceTest.java
+++ b/services/src/test/java/io/grpc/protobuf/services/ProtoReflectionServiceTest.java
@@ -532,8 +532,11 @@ public class ProtoReflectionServiceTest {
         (ClientCallStreamObserver<ServerReflectionRequest>)
             stub.serverReflectionInfo(clientResponseObserver);
 
-    // ClientCalls.startCall() calls request(1) initially, so we should get an immediate response.
+    // We should not get an initial response until we request it.
     requestObserver.onNext(flowControlRequest);
+    assertEquals(0, clientResponseObserver.getResponses().size());
+
+    requestObserver.request(1);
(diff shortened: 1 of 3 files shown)


ejona86:
We shouldn't change the behavior of `disableAutoInboundFlowControl()` as that will break existing users without giving them help upgrading. We do agree the current behavior is confusing. [The plan](https://github.com/grpc/grpc-java/issues/1788#issuecomment-341568779) was to create a new method that will disable all automatic `request()`s and deprecate+remove `disableAutoInboundFlowControl()`.

barnardb:
Ah, I see. I should have read that comment more clearly.

I could have a go at that approach. Have you had any further thoughts on an appropriate name?

ejona86:
We discussed it today. We're going to go with `disableAutoRequest(void)`. Please have a go with that approach!

zpencer:
Hi @barnardb friendly ping on this PR.

ejona86:
This was fixed by a combination of https://github.com/grpc/grpc-java/pull/6807 and https://github.com/grpc/grpc-java/pull/7071 . There's now `disableAutoRequest()` (server-side) and `disableAutoRequestWithInitial()` (client-side) methods. They were added in v1.30.

PR #4220, opened 2018-03-14
core,netty,services: add server listen sockets to channelz proto service

zpencer:
This PR exposes listen sockets to the channelz service.
These sockets are not in the set of sockets discoverable via
`GetServerSockets`.

Server listen sockets differ from normal sockets in that only have
a local address and socket options.

Fixed an issue of socket addresses not reporting the port.

### DIFF
--- a/core/src/main/java/io/grpc/internal/Channelz.java
+++ b/core/src/main/java/io/grpc/internal/Channelz.java
@@ -263,7 +263,7 @@ public final class Channelz {
     public final long callsSucceeded;
     public final long callsFailed;
     public final long lastCallStartedMillis;
-    // TODO(zpencer): add listen sockets
+    public final List<Instrumented<SocketStats>> listenSockets;
 
     /**
      * Creates an instance.
@@ -272,11 +272,13 @@ public final class Channelz {
         long callsStarted,
         long callsSucceeded,
         long callsFailed,
-        long lastCallStartedMillis) {
+        long lastCallStartedMillis,
+        List<Instrumented<SocketStats>> listenSockets) {
       this.callsStarted = callsStarted;
       this.callsSucceeded = callsSucceeded;
       this.callsFailed = callsFailed;
       this.lastCallStartedMillis = lastCallStartedMillis;
+      this.listenSockets = Preconditions.checkNotNull(listenSockets);
     }
 
     public static final class Builder {
@@ -284,6 +286,7 @@ public final class Channelz {
       private long callsSucceeded;
       private long callsFailed;
       private long lastCallStartedMillis;
+      public List<Instrumented<SocketStats>> listenSockets = Collections.emptyList();
 
       public Builder setCallsStarted(long callsStarted) {
         this.callsStarted = callsStarted;
@@ -305,6 +308,13 @@ public final class Channelz {
         return this;
       }
 
+      /** Sets the listen sockets. */
+      public Builder setListenSockets(List<Instrumented<SocketStats>> listenSockets) {
+        Preconditions.checkNotNull(listenSockets);
+        this.listenSockets = Collections.unmodifiableList(listenSockets);
+        return this;
+      }
+
       /**
        * Builds an instance.
        */
@@ -313,7 +323,8 @@ public final class Channelz {
             callsStarted,
             callsSucceeded,
             callsFailed,
-            lastCallStartedMillis);
+            lastCallStartedMillis,
+            listenSockets);
       }
     }
   }
@@ -434,11 +445,11 @@ public final class Channelz {
   }
 
   public static final class SocketStats {
-    public final TransportStats data;
+    @Nullable public final TransportStats data;
     public final SocketAddress local;
-    public final SocketAddress remote;
-    public final Security security;
+    @Nullable public final SocketAddress remote;
     public final SocketOptions socketOptions;
+    @Nullable public final Security security;
 
     /** Creates an instance. */
     public SocketStats(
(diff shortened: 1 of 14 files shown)


core/src/main/java/io/grpc/internal/Channelz.java
```diff
@@ -226,6 +229,13 @@ public Builder setLastCallStartedMillis(long lastCallStartedMillis) {
         return this;
       }
 
+      /** Sets the listen sockets. */
+      public Builder setListenSockets(List<Instrumented<SocketStats>> listenSockets) {
+        Preconditions.checkNotNull(listenSockets);
+        this.listenSockets = Collections.unmodifiableList(listenSockets);
```
carl-mastrangelo: nit: if the caller retained access to `listenSockets`, they could still mutate it.   Consider making a defensive copy.

core/src/main/java/io/grpc/internal/InternalServer.java
```diff
@@ -16,7 +16,10 @@
 
 package io.grpc.internal;
 
+import com.sun.istack.internal.NotNull;
```
carl-mastrangelo: Probably not the one you want

core/src/main/java/io/grpc/internal/InternalServer.java
```diff
@@ -47,4 +50,10 @@
    * available or does not make sense.
    */
   int getPort();
+
+  /**
+   * Returns the listen sockets of this server.
+   */
+  @NotNull
```
carl-mastrangelo: This probably needs to go into the java doc rather than annotation

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -574,7 +574,8 @@ public LogId getLogId() {
 
   @Override
   public ListenableFuture<ServerStats> getStats() {
-    ServerStats.Builder builder = new ServerStats.Builder();
+    ServerStats.Builder builder = new ServerStats.Builder()
```
carl-mastrangelo: break line at highest semantic level, i.e. the `=`

netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -116,6 +127,7 @@
     this.maxConnectionAgeGraceInNanos = maxConnectionAgeGraceInNanos;
     this.permitKeepAliveWithoutCalls = permitKeepAliveWithoutCalls;
     this.permitKeepAliveTimeInNanos = permitKeepAliveTimeInNanos;
+    this.channelz = channelz;
```
carl-mastrangelo: checkNotNull?

netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -245,6 +269,9 @@ public void operationComplete(ChannelFuture future) throws Exception {
         if (!future.isSuccess()) {
           log.log(Level.WARNING, "Error shutting down server", future.cause());
         }
+        for (Instrumented<SocketStats> listenSocket : listenSockets) {
+          channelz.removeSocket(listenSocket);
```
carl-mastrangelo: nit: you probably should use the iterator, and remove the sockets as you go along to avoid retaining memory.

This also means the list cannot be Immutable.

netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -291,4 +318,48 @@ public ReferenceCounted touch(Object hint) {
...
+        // This is necessary, otherwise we will block forever if we get the future from inside
+        // the event loop.
+        ret.set(new SocketStats(
+            /*data=*/ null,
+            ch.localAddress(),
+            /*remoteAddress=*/ null,
+            /*security=*/ null));
+        return ret;
+      }
+      ch.eventLoop().submit(
```
carl-mastrangelo: nit: on failure you need to fail `ret`.    In this case, probably adding a listener to the future returned by submit.

netty/src/main/java/io/grpc/netty/NettyServer.java
zpencer: I don't think there will be enough server listen sockets for this to be a serious issue. I can set the list to null with low effort, however.

netty/src/main/java/io/grpc/netty/NettyServer.java
zpencer: fixed here and other classes

netty/src/main/java/io/grpc/netty/NettyServer.java
zpencer: done

core/src/main/java/io/grpc/internal/ServerImpl.java
zpencer: done

core/src/main/java/io/grpc/internal/InternalServer.java
zpencer: fixed

core/src/main/java/io/grpc/internal/Channelz.java
zpencer: Made defensive copy

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -571,7 +571,8 @@ public LogId getLogId() {
 
   @Override
   public ListenableFuture<ServerStats> getStats() {
-    ServerStats.Builder builder = new ServerStats.Builder()
+    ServerStats.Builder builder
```
carl-mastrangelo: nit: please consider following the rectangle rule, and wrap lines such that the obey it.  In this case, `.setListenSockets(transportServer.getListenSockets());` should be indented.

netty/src/main/java/io/grpc/netty/NettyClientTransport.java
```diff
@@ -337,7 +339,16 @@ public void run() {
...
+        })
+        .addListener(
+            new GenericFutureListener<Future<Object>>() {
+              @Override
+              public void operationComplete(Future<Object> future) throws Exception {
+                if (!future.isSuccess()) {
+                  result.setException(future.cause());
+                }
+              }
+            });;
```
carl-mastrangelo: double semi colons

PR #4226, opened 2018-03-15
core,services: track server sockets in channelz

zpencer:
Rather than querying the ServerImpl for its sockets, we register them
into channelz as with all the other entities, for consistency.

Adjust the server/serversocket registration so that:
* server is added before the first transport
* server is removed after the last transport

Error conditions are checked with `assert` statements, so we throw in
tests. But in prod, an error in channelz should not prevent normal
gRPC operation.

I have tested a branch where we avoid using Long as map keys, and
instead use the LogId as the keys:
https://gist.github.com/zpencer/c04efaaa8b6772a8fa392539c45410ab

Long keys benchmark:
https://gist.github.com/zpencer/a31d470bb46c1205d3110e20fca0050e

LogId keys benchmark:
https://gist.github.com/zpencer/5a5c6c2cea9b12bfe480b7ee5c6ab0e4

We expect servers to have many sockets, and it appears using
WithLogId is worse. Let's stick with Long for now.

### DIFF
--- a/core/src/main/java/io/grpc/internal/Channelz.java
+++ b/core/src/main/java/io/grpc/internal/Channelz.java
@@ -144,7 +144,8 @@ public final class Channelz {
 
   /** Returns a server list. */
   public ServerList getServers(long fromId, int maxPageSize) {
-    List<Instrumented<ServerStats>> serverList = new ArrayList<Instrumented<ServerStats>>();
+    List<Instrumented<ServerStats>> serverList
+        = new ArrayList<Instrumented<ServerStats>>(maxPageSize);
     Iterator<Instrumented<ServerStats>> iterator = servers.tailMap(fromId).values().iterator();
 
     while (iterator.hasNext() && serverList.size() < maxPageSize) {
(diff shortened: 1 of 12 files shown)


zpencer:
Updated PR to reflect the benchmark class used for results.

core/src/main/java/io/grpc/internal/Channelz.java
```diff
@@ -119,10 +153,40 @@ public ServerList getServers(long fromId, int maxPageSize) {
...
   }
 
+  /** Returns socket refs for a server. */
+  @Nullable
+  public ServerSocketsList getServerSockets(long serverId, long fromId, int maxPageSize) {
+    ServerSocketMap serverSockets = perServerSockets.get(serverId);
+    if (serverSockets == null) {
+      return null;
+    }
+    List<WithLogId> socketList = new ArrayList<WithLogId>();
```
carl-mastrangelo: `new ArrayList<WithLogId>(maxPageSize);`

PR #4227, opened 2018-03-15
netty: http2 server transport graceful shutdown sends 2 GOAWAYs

### LINKED ISSUE #3442: Netty server doesn't send multiple GOAWAYs during graceful shutdown
For full graceful shutdown, we should send one GOAWAY with last-steam-id=MAX_INT and a PING. When the PING ACK is received then we send another GOAWAY with the actual last-stream-id. We can choose to set a timer for if the PING ACK doesn't arrive back promptly.

It looks like we try to do shutdown gracefully with maxConnectionAge, but it seems like it may not actually function.

dapengzhang0:
resolves #3442

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
@@ -771,7 +771,9 @@ class NettyServerHandler extends AbstractNettyHandler {
               decoder().flowController().initialWindowSize(connection().connectionStream())));
         }
       } else if (data == GRACEFUL_SHUTDOWN_PING) {
-        checkNotNull(gracefulShutdown, "gracefulShutdownRunner");
+        if (gracefulShutdown == null) {
+          throw new AssertionError("gracefulShutdown is null");
+        }
         gracefulShutdown.secondGoAwayAndClose(ctx);
       } else if (data != KEEPALIVE_PING) {
         logger.warning("Received unexpected ping ack. No ping outstanding");
--- a/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerHandler.java
@@ -99,7 +99,6 @@ class NettyServerHandler extends AbstractNettyHandler {
   private static final long KEEPALIVE_PING = 0xDEADL;
   private static final long GRACEFUL_SHUTDOWN_PING = 0x97ACEF001L;
   private static final long GRACEFUL_SHUTDOWN_PING_TIMEOUT_NANOS = TimeUnit.SECONDS.toNanos(10);
-  private static final SystemTicker SYSTEM_TICKER = new SystemTicker();
 
   private final Http2Connection.PropertyKey streamKey;
   private final ServerTransportListener transportListener;
@@ -127,8 +126,6 @@ class NettyServerHandler extends AbstractNettyHandler {
   private ScheduledFuture<?> maxConnectionAgeMonitor;
   @CheckForNull
   private GracefulShutdown gracefulShutdown;
-
-  private Ticker ticker = SYSTEM_TICKER;
 
   static NettyServerHandler newHandler(
       ServerTransportListener transportListener,
@@ -322,7 +319,7 @@ class NettyServerHandler extends AbstractNettyHandler {
           new LogExceptionRunnable(new Runnable() {
             @Override
             public void run() {
-              if (gracefulShutdown != null) {
+              if (gracefulShutdown == null) {
                 gracefulShutdown = new GracefulShutdown("max_age", maxConnectionAgeGraceInNanos);
                 gracefulShutdown.start(ctx);
                 ctx.flush();
@@ -840,10 +837,6 @@ class NettyServerHandler extends AbstractNettyHandler {
      */
     boolean pingAckedOrTimeout;
 
-    /**
-     * Deadline of the shutdown.
-     */
-    long deadline;
     Future<?> pingFuture;
 
     GracefulShutdown(String goAwayMessage,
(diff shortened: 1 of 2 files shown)


netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -837,6 +827,105 @@ public void onPingTimeout() {
...
+     * Sends out first GOAWAY and ping, and schedules second GOAWAY and close.
+     */
+    void run() {
+      goAway(
+          ctx,
+          Integer.MAX_VALUE,
+          Http2Error.NO_ERROR.code(),
+          ByteBufUtil.writeAscii(ctx.alloc(), goAwayMessage),
+          ctx.newPromise());
+      ctx.flush();
```
ejona86: This flush is unnecessary. Another is following it.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -837,6 +827,105 @@ public void onPingTimeout() {
...
+            @Override
+            public void run() {
+              secondGoAwayAndClose();
+            }
+          },
+          gracefulShutdownPingTimeout,
+          TimeUnit.NANOSECONDS);
+
+      encoder().writePing(ctx, false /* isAck */, payload, ctx.newPromise());
+      ctx.flush();
```
ejona86: I'd prefer the caller do the flush. The only time explicit flushes are necessary here is for writes due to timers. For writes caused by reads, we do flush in onReadComplete. For writes coming from the application, it ends up flushing.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -837,6 +827,105 @@ public void onPingTimeout() {
...
+          ctx,
+          Integer.MAX_VALUE,
+          Http2Error.NO_ERROR.code(),
+          ByteBufUtil.writeAscii(ctx.alloc(), goAwayMessage),
+          ctx.newPromise());
+      ctx.flush();
+
+      long gracefulShutdownPingTimeout = GRACEFUL_SHUTDOWN_PING_TIMEOUT_NANOS;
+      if (graceTimeInNanos != null) {
+        deadline = ticker.read() + graceTimeInNanos;
```
ejona86: Count grace time starting at `secondGoAwayAndClose()`, since we'll still be accepting new requests until that point.

With that change, it also seems the `ticker` would no longer be necessary.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -837,6 +827,105 @@ public void onPingTimeout() {
...
+      pingFuture.cancel(false);
+
+      // send the second GOAWAY with last stream id
+      goAway(
+          ctx,
+          connection().remote().lastStreamCreated(),
+          Http2Error.NO_ERROR.code(),
+          ByteBufUtil.writeAscii(ctx.alloc(), goAwayMessage),
+          ctx.newPromise());
+      ctx.flush();
```
ejona86: Ditto remove. You only need the flush in the gracefulShutdownPingTimeout handling (not for the ping receiving).

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -787,6 +772,12 @@ public void onPingAckRead(ChannelHandlerContext ctx, long data) throws Http2Exce
           logger.log(Level.FINE, String.format("Window: %d",
               decoder().flowController().initialWindowSize(connection().connectionStream())));
         }
+      } else if (data == MAX_CONNECTION_AGE_PING) {
+        checkNotNull(maxAgeShutdownRunner, "maxAgeShutdownRunner");
```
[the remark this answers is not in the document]
ejona86: Then make it an `assert` or manually `throw new AssertionError()`. `checkNotNull` is best used when verifying your input, and telling the caller they made a mistake.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -837,6 +825,96 @@ public void onPingTimeout() {
...
+          connection().remote().lastStreamCreated(),
+          Http2Error.NO_ERROR.code(),
+          ByteBufUtil.writeAscii(ctx.alloc(), goAwayMessage),
+          ctx.newPromise());
+
+      // gracefully shutdown with specified grace time
+      long savedGracefulShutdownTimeMillis = gracefulShutdownTimeoutMillis();
+      long gracefulShutdownTimeoutMillis = savedGracefulShutdownTimeMillis;
+      if (graceTimeInNanos != null) {
+        gracefulShutdownTimeoutMillis = TimeUnit.NANOSECONDS.toMillis(deadline - ticker.read());
```
ejona86: Why is this consulting `deadline`? I thought we agreed that `graceTimeInNanos` would be relative to the second GOAWAY.

dapengzhang0: This was a partial fix. Removing ticker is in the upcoming commit.

dapengzhang0: Done.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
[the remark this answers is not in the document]
dapengzhang0: Done.

dapengzhang0:
@ejona86 Thanks for the review. PTAL.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
dapengzhang0: Done.

netty/src/main/java/io/grpc/netty/NettyServerHandler.java
```diff
@@ -787,6 +770,11 @@ public void onPingAckRead(ChannelHandlerContext ctx, long data) throws Http2Exce
           logger.log(Level.FINE, String.format("Window: %d",
               decoder().flowController().initialWindowSize(connection().connectionStream())));
         }
+      } else if (data == GRACEFUL_SHUTDOWN_PING) {
+        if (gracefulShutdown == null) {
```
ejona86: Oh, I remember why I put that comment. This is doing a check based on what is _received_. In your earlier response you said:
> It will never throw a RuntimeException unless there's a bug in the code.

But the bug could be in the _remote_ code, not in this code. That's why I had suggested a warning.

PR #4228, opened 2018-03-15
core,netty,okhttp,services: expose socket options to channelz

zpencer:
For okhttp, expose the standard options from the Socket object.

For netty, expose all the `io.netty.channel.ChannelOption`s of the
`channel.config()`.

### DIFF
--- a/core/src/main/java/io/grpc/internal/Channelz.java
+++ b/core/src/main/java/io/grpc/internal/Channelz.java
@@ -466,9 +466,10 @@ public final class Channelz {
         Integer timeoutMillis,
         Integer lingerSeconds,
         Map<String, String> others) {
+      Preconditions.checkNotNull(others);
       this.soTimeoutMillis = timeoutMillis;
       this.lingerSeconds = lingerSeconds;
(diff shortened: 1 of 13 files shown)


core/src/main/java/io/grpc/internal/Channelz.java
```diff
@@ -359,20 +360,79 @@ public ChannelStats build() {
...
+    @Nullable public final Integer lingerSeconds;
+
+    /** Creates an instance. */
+    public SocketOptions(
+        Integer timeoutMillis,
+        Integer lingerSeconds,
+        Map<String, String> others) {
+      this.soTimeoutMillis = timeoutMillis;
+      this.lingerSeconds = lingerSeconds;
+      this.others = Collections.unmodifiableMap(Preconditions.checkNotNull(others));
```
carl-mastrangelo: This won't actually be immutable.  Adding things to the builder *after* building options will modify the SocketOptions.   
```
b = new Builder();
so = b.build();
b.addOption(a, b);
// so changed!
```

PR #4237, opened 2018-03-19
core: parse retry policy from service config

dapengzhang0:


core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -59,6 +66,139 @@ static String getLoadBalancingPolicy(Map<String, Object> serviceConfig) {
...
+          }
+        ]
+      }
+      */
+
+      if (serviceConfig.containsKey("methodConfig")) {
+        List<Object> methodConfigs = getList(serviceConfig, "methodConfig");
+        for (int i = 0; i < methodConfigs.size(); i++) {
+          Map<String, Object> methodConfig = getObject(methodConfigs, i);
+          List<Object> names = getList(methodConfig, "name");
```
ericgribkoff: nit: why get the names here instead of when they're used (line 135)?

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -59,6 +66,139 @@ static String getLoadBalancingPolicy(Map<String, Object> serviceConfig) {
     return null;
   }
 
+  static RetryPolicies getRetryPolicies(
```
ericgribkoff: Maybe add Javadoc (or just a comment) that this will throw an exception if the service config doesn't parse?

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -59,6 +66,139 @@ static String getLoadBalancingPolicy(Map<String, Object> serviceConfig) {
...
+      @Nullable Map<String, Object> serviceConfig, int customMaxAttempts) {
```
ericgribkoff: optional: s/customMaxAttempts/maxAttemptsLimit/ (or maxMaxAttempts...)

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -174,6 +314,20 @@ static String getLoadBalancingPolicy(Map<String, Object> serviceConfig) {
         String.format("value %s for key %s in %s is not object", value, key, obj));
   }
 
+  /**
+   * Gets an object from a list of objects for the given index.
+   */
+  @SuppressWarnings("unchecked")
+  private static Map<String, Object> getObject(List<Object> list, int i) {
+    assert i >= 0 && i < list.size();
```
ericgribkoff: question: why assert and not checkState?

Issue #3553, opened 2017-10-06
Make sure retry and hedging implementation handle balancer drop correctly


zhangkun83:
As mentioned in #3355. RPCs dropped by LoadBalancer should never by retried. We should also decide how hedging implementation should deal with dropped RPCs. We should have unit tests to cover them.

/cc @dapengzhang0

dapengzhang0:
Was fixed by #4246

PR #4246, opened 2018-03-20
core: no retry on dropped PickResult

dapengzhang0:
Address #3553 for normal retry. Not covering the unimplemented hedging case.

### DIFF
--- a/core/src/test/java/io/grpc/internal/FailingClientStreamTest.java
+++ b/core/src/test/java/io/grpc/internal/FailingClientStreamTest.java
@@ -35,15 +35,21 @@ import org.junit.runners.JUnit4;
 public class FailingClientStreamTest {
 
   @Test
-  public void rpcProgressPopulatedToListener() {
+  public void processedRpcProgressPopulatedToListener() {
     ClientStreamListener listener = mock(ClientStreamListener.class);
     Status status = Status.UNAVAILABLE;
 
     ClientStream stream = new FailingClientStream(status);
     stream.start(listener);
     verify(listener).closed(eq(status), eq(RpcProgress.PROCESSED), any(Metadata.class));
+  }
 
-    stream = new FailingClientStream(status, RpcProgress.DROPPED);
+  @Test
+  public void droppedRpcProgressPopulatedToListener() {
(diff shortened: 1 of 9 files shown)


core/src/test/java/io/grpc/internal/FailingClientStreamTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.junit.runners.JUnit4;
+
+/**
+ * Unit tests for {@link FailingClientStream}.
+ */
+@RunWith(JUnit4.class)
+public class FailingClientStreamTest {
+
+  @Test
+  public void rpcProgressPopulatedToListener() {
```
zhangkun83: Better to split it into two methods.

core/src/test/java/io/grpc/internal/RetriableStreamTest.java
```diff
@@ -1506,6 +1508,30 @@ public void normalRetry_thenNoTransparentRetry_andNoMoreRetry() {
...
+    doReturn(mockStream1).when(retriableStreamRecorder).newSubstream(0);
+    retriableStream.start(masterListener);
+
+    verify(retriableStreamRecorder).newSubstream(0);
+    ArgumentCaptor<ClientStreamListener> sublistenerCaptor1 =
+        ArgumentCaptor.forClass(ClientStreamListener.class);
+    verify(mockStream1).start(sublistenerCaptor1.capture());
+
+    // normal retry
+    doReturn(mockStream2).when(retriableStreamRecorder).newSubstream(1);
```
zhangkun83: All stubbing should happen at the beginning of the test. You don't know whether the stubbed method was called before this line.

core/src/test/java/io/grpc/internal/FailingClientStreamTest.java
dapengzhang0: Done.

core/src/test/java/io/grpc/internal/RetriableStreamTest.java
dapengzhang0: Done.

dapengzhang0:
@zhangkun83 Thanks for the review.

PR #4253, opened 2018-03-22
buildscripts: build windows artifacts in kokoro

zpencer:
We always build the artifacts, but only store them in placer via
kokoro if MVN_ARTIFACTS is set to a non empty value.
MVN_ARTIFACTS is not set by default right now.

Click on the windows invocation URL and click placer link to find the artifacts:
https://source.cloud.google.com/results/invocations/8cf374ec-52f8-4e21-ae81-3a957d60953d/details

### DIFF
--- a/buildscripts/kokoro/windows.bat
+++ b/buildscripts/kokoro/windows.bat
@@ -17,53 +17,8 @@ set ESCWORKSPACE=%WORKSPACE:\=\\%
 set JAVA_HOME=
 set PATH=C:\Program Files\java\jdk1.8.0_152\bin;%PATH%
 
-mkdir grpc-java-helper32
-cd grpc-java-helper32
-call "%VS120COMNTOOLS%\vsvars32.bat" || exit /b 1
-call "%WORKSPACE%\buildscripts\make_dependencies.bat" || exit /b 1
-
-cd "%WORKSPACE%"
-
-SET TARGET_ARCH=x86_32
-SET FAIL_ON_WARNINGS=true
-SET VC_PROTOBUF_LIBS=%ESCWORKSPACE%\\grpc-java-helper32\\protobuf-%PROTOBUF_VER%\\cmake\\build\\Release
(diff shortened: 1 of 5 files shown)


buildscripts/kokoro/windows.bat
```diff
@@ -40,4 +41,32 @@ for /r %%F in (TEST-*.xml) do (
...
+)
+
+@rem make sure no daemons have any files open
+cmd.exe /C "%WORKSPACE%\gradlew.bat --stop"
+
+cmd.exe /C "%WORKSPACE%\gradlew.bat  %GRADLE_FLAGS% -Dorg.gradle.parallel=false -PrepositoryDir=%WORKSPACE%\artifacts\x86_32\ clean grpc-compiler:build grpc-compiler:uploadArchives" || exit /b 1
+
+mkdir grpc-java-helper64
+cd grpc-java-helper64
+call "%VS120COMNTOOLS%\..\..\VC\bin\amd64\vcvars64.bat" || exit /b 1
```
ejona86: It scares me to run these again in the same environment. For example, this script will break if you do 64 bit first, because CMAKE_VSARCH would still be set from the first call. Let's at least change to using CMD to call make_dependencies.bat instead of CALL. To do better we'd have to split this into two extra files I think.

buildscripts/make_dependencies.bat
```diff
@@ -26,7 +26,14 @@ del protobuf.zip
...
 mkdir build
 cd build
-cmake -Dprotobuf_BUILD_TESTS=OFF -G "Visual Studio %VisualStudioVersion:~0,2%" .. || exit /b 1
+
+@rem cmake does not detect x86_64 from the vcvars64.bat variables.
+@rem If vcvars64.bat has set PLATFORM to X64, then inform cmake to use the Win64 version of VS
+if "%PLATFORM%" == "X64" (
+  @rem Note the space
+  SET CMAKE_VSARCH= Win64
+)
```
ejona86: Let's not assume the default value
```
) else (
  SET CMAKE_VSARCH=
)
```

PR #4254, opened 2018-03-22
routeguide: reimplement distance calculation

jtattermusch:
Same as https://github.com/grpc/grpc/pull/14794.

### DIFF
--- a/examples/src/main/java/io/grpc/examples/routeguide/RouteGuideServer.java
+++ b/examples/src/main/java/io/grpc/examples/routeguide/RouteGuideServer.java
@@ -268,25 +268,23 @@ public class RouteGuideServer {
 
     /**
      * Calculate the distance between two points using the "haversine" formula.
-     * This code was taken from http://www.movable-type.co.uk/scripts/latlong.html.
+     * The formula is based on http://mathforum.org/library/drmath/view/51879.html.
      *
      * @param start The starting point
      * @param end The end point
      * @return The distance between the points in meters
      */
     private static int calcDistance(Point start, Point end) {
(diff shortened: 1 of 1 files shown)


ejona86:
Making a comment to try to kick cla/linuxfoundation to do its thing.

examples/src/main/java/io/grpc/examples/routeguide/RouteGuideServer.java
```diff
@@ -268,25 +268,23 @@ private Feature checkFeature(Point location) {
...
+      int r = 6371000; // earth radius in meters
+      double lat1 = toRadians(RouteGuideUtil.getLatitude(start));
+      double lat2 = toRadians(RouteGuideUtil.getLatitude(end));
+      double lon1 = toRadians(RouteGuideUtil.getLongitude(start));
+      double lon2 = toRadians(RouteGuideUtil.getLongitude(end));
+      double deltaLat = lat2 - lat1;
+      double deltaLon = lon2 - lon1;
+
+      double a = sin(deltaLat / 2) * sin(deltaLat / 2)
+          + cos(lat1) * cos(lat2) * sin(deltaLon / 2) * sin(deltaLon / 2);
```
ejona86: Like the other implementation, this looks wrong, as we're passing non-radians to cos.

jtattermusch: Not sure what you mean:
`double lat1 = toRadians(RouteGuideUtil.getLatitude(start));`

ejona86: nm

PR #4260, opened 2018-03-23
core: apply service config 

carl-mastrangelo:
Several things are interlocked, thus a big CL.

* Most of service config parsing logic has been moved into ServiceConfigUtil.   Several methods are copied for duration parsing to match what protobuf will do. I do not intend to copy the tests
* ServiceConfigInterceptor will apply the service config.   It gets updates from the resolver, parses and applies it to the channel.   This was done internally, but I have moved it to open source.    The tests **are copied**
* Retry policy now uses the data type from Service Config, and has stricter parsing.   More testing for this is needed, but there are already TODOs for this, and this PR is getting unwieldy.  Let's get this in and add the tests in a followup.
* Retry throttling hasn't been consolidated.   This will also be a followup.


carl-mastrangelo:
@dapengzhang0  ping

carl-mastrangelo:
thanks @dapengzhang0

PR #4266, opened 2018-03-24
core,services: binary log should use real peer socket and call id

zpencer:
The peer socket is read from TRANSPORT_ATTR_REMOTE_ADDR from the
stream attributes. We only log the peer on receive initial metadata.

The call id assumes census is available. The call ID is read from
the context via SERVER_CALL_ID_CONTEXT_KEY on server side, and
read from CallOptions via CLIENT_CALL_ID_CALLOPTION_KEY on client
side. The value is copied from census's using its public APIs,
but nothing prevents other values from being used instead.

`getClientCallIdSetter` and `getServerCallIdSetter` will move outside of core
and into the external package along with the wrapper methods in a follow up.

### DIFF
--- a/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
+++ b/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
@@ -16,8 +16,6 @@
 
 package io.grpc.internal;
 
-import static io.opencensus.trace.unsafe.ContextUtils.CONTEXT_SPAN_KEY;
-
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import io.grpc.CallOptions;
@@ -39,6 +37,7 @@ import io.grpc.ServerMethodDefinition;
 import io.grpc.ServerStreamTracer;
 import io.opencensus.trace.Span;
 import io.opencensus.trace.TraceId;
+import io.opencensus.trace.Tracing;
 import java.io.ByteArrayInputStream;
 import java.io.Closeable;
 import java.io.IOException;
@@ -143,13 +142,17 @@ public abstract class BinaryLogProvider implements Closeable {
   private static final ServerStreamTracer SERVER_CALLID_SETTER = new ServerStreamTracer() {
     @Override
     public Context filterContext(Context context) {
-      Span span = CONTEXT_SPAN_KEY.get(context);
-      if (span == null) {
-        return context;
+      Context toRestore = context.attach();
+      try {
+        Span span = Tracing.getTracer().getCurrentSpan();
+        if (span == null) {
+          return context;
+        }
+        TraceId traceId = span.getContext().getTraceId();
+        return context.withValue(SERVER_CALL_ID_CONTEXT_KEY, new CallId(traceId.getBytes()));
+      } finally {
+        context.detach(toRestore);
       }
-
-      TraceId traceId = span.getContext().getTraceId();
-      return context.withValue(SERVER_CALL_ID_CONTEXT_KEY, new CallId(traceId.getBytes()));
     }
   };
 
@@ -173,7 +176,7 @@ public abstract class BinaryLogProvider implements Closeable {
     @Override
     public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
         MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
-      Span span = CONTEXT_SPAN_KEY.get();
+      Span span = Tracing.getTracer().getCurrentSpan();
       if (span == null) {
         return next.newCall(method, callOptions);
       }
--- a/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
+++ b/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
@@ -16,21 +16,29 @@
 
 package io.grpc.internal;
 
+import static io.opencensus.trace.unsafe.ContextUtils.CONTEXT_SPAN_KEY;
+
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import io.grpc.CallOptions;
 import io.grpc.Channel;
 import io.grpc.ClientCall;
 import io.grpc.ClientInterceptor;
 import io.grpc.ClientInterceptors;
+import io.grpc.Context;
 import io.grpc.InternalClientInterceptors;
 import io.grpc.InternalServerInterceptors;
 import io.grpc.InternalServiceProviders;
 import io.grpc.InternalServiceProviders.PriorityAccessor;
+import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
 import io.grpc.MethodDescriptor.Marshaller;
 import io.grpc.ServerCallHandler;
 import io.grpc.ServerInterceptor;
 import io.grpc.ServerMethodDefinition;
+import io.grpc.ServerStreamTracer;
+import io.opencensus.trace.Span;
+import io.opencensus.trace.TraceId;
 import java.io.ByteArrayInputStream;
 import java.io.Closeable;
 import java.io.IOException;
@@ -40,6 +48,10 @@ import java.util.logging.Logger;
 import javax.annotation.Nullable;
 
 public abstract class BinaryLogProvider implements Closeable {
+  public static final Context.Key<CallId> SERVER_CALL_ID_CONTEXT_KEY
+      = Context.key("binarylog-context-key");
+  public static final CallOptions.Key<CallId> CLIENT_CALL_ID_CALLOPTION_KEY
+      = CallOptions.Key.of("binarylog-calloptions-key", null);
   @VisibleForTesting
   public static final Marshaller<byte[]> BYTEARRAY_MARSHALLER = new ByteArrayMarshaller();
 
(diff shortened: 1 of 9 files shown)


services/src/main/java/io/grpc/services/BinaryLog.java
```diff
@@ -275,13 +282,38 @@ static BinaryLog getLog(String fullMethodName) {
...
+
   @Override
   public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
       final MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+    final byte[] callId = getCallIdForClient(callOptions);
     return new SimpleForwardingClientCall<ReqT, RespT>(next.newCall(method, callOptions)) {
       @Override
       public void start(Listener<RespT> responseListener, Metadata headers) {
-        writer.logSendInitialMetadata(headers, CLIENT, dumyCallId, DUMMY_SOCKET);
+        writer.logSendInitialMetadata(headers, CLIENT, callId);
```
zhangkun83: Since you already have a `CallId` type, why not change the writer interface to accept that instead of `byte[]`?

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
```diff
@@ -310,6 +321,53 @@ public void sendMessage(RespT message) {
...
+  @Test
+  public void serverCallIdSetter() {
+    ServerStreamTracer tracer = BinaryLogProvider.SERVER_CALLID_SETTER_FACTORY
+        .newServerStreamTracer(
+            "service/method", new Metadata());
+    MockableSpan mockableSpan = MockableSpan.generateRandomSpan(new Random(0));
+    Context context = Context.current().withValue(CONTEXT_SPAN_KEY, mockableSpan);
+    Context filtered = tracer.filterContext(context);
+    CallId callId = BinaryLogProvider.SERVER_CALL_ID_CONTEXT_KEY.get(filtered);
+    assertTrue(Arrays.equals(mockableSpan.getContext().getTraceId().getBytes(), callId.id));
```
zhangkun83: `assertThat(mockableSpan.getContext().getTraceId().getBytes()).isEqualTo(callId.id)`

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
```diff
@@ -310,6 +321,53 @@ public void sendMessage(RespT message) {
...
+        };
+        BinaryLogProvider.CLIENT_CALLID_SETTER.interceptCall(
+            TestMethodDescriptors.voidMethod(),
+            CallOptions.DEFAULT,
+            channel);
+        CallOptions callOptions = future.get();
+        CallId callId = callOptions
+            .getOption(BinaryLogProvider.CLIENT_CALL_ID_CALLOPTION_KEY);
+        assertTrue(
+            Arrays.equals(mockableSpan.getContext().getTraceId().getBytes(), callId.id));
```
zhangkun83: ditto

services/src/main/java/io/grpc/services/BinaryLog.java
```diff
@@ -77,11 +82,16 @@
   private static final boolean SERVER = true;
   private static final boolean CLIENT = false;
 
-  // TODO(zpencer): extract these fields from call and stop using dummy values
   @VisibleForTesting
-  static final byte[] dumyCallId = new byte[16];
+  static final byte[] emptyCallId = new byte[16];
```
zhangkun83: This can be defined as a `CallId` too.

services/src/main/java/io/grpc/services/BinaryLog.java
```diff
@@ -275,13 +282,38 @@ static BinaryLog getLog(String fullMethodName) {
     return DEFAULT_FACTORY.getLog(fullMethodName);
   }
 
+  static byte[] getCallIdForServer(Context context) {
```
zhangkun83: The return type can be `CallId`

services/src/main/java/io/grpc/services/BinaryLog.java
```diff
@@ -275,13 +282,38 @@ static BinaryLog getLog(String fullMethodName) {
...
+    CallId callId = BinaryLogProvider.SERVER_CALL_ID_CONTEXT_KEY.get(context);
+    if (callId == null) {
+      return emptyCallId;
+    }
+    return callId.id;
+  }
+
+  static byte[] getCallIdForClient(CallOptions callOptions) {
```
zhangkun83: ditto

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -143,13 +142,17 @@ public void close() throws IOException {
...
-      Span span = CONTEXT_SPAN_KEY.get(context);
-      if (span == null) {
-        return context;
+      Context toRestore = context.attach();
+      try {
+        Span span = Tracing.getTracer().getCurrentSpan();
+        if (span == null) {
+          return context;
+        }
+        TraceId traceId = span.getContext().getTraceId();
```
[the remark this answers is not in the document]
zpencer: Switched to span

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -128,6 +139,63 @@ public void close() throws IOException {
     // TODO(zpencer): make BinaryLogProvider provide a BinaryLog, and this method belongs there
   }
 
+  private static final ServerStreamTracer SERVER_CALLID_SETTER = new ServerStreamTracer() {
+    @Override
+    public Context filterContext(Context context) {
+      Context toRestore = context.attach();
```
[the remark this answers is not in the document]
zpencer: @ejona86 had some concerns about using `unsafe`, which is the reason for attaching it here

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -128,6 +139,63 @@ public void close() throws IOException {
...
+        }
+        TraceId traceId = span.getContext().getTraceId();
+        return context.withValue(SERVER_CALL_ID_CONTEXT_KEY, new CallId(traceId.getBytes()));
+      } finally {
+        context.detach(toRestore);
+      }
+    }
+  };
+
+  static final ServerStreamTracer.Factory SERVER_CALLID_SETTER_FACTORY
```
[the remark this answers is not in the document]
zpencer: Switched to using public APIs for testing

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -128,6 +139,63 @@ public void close() throws IOException {
...
+
+  /**
+   * Returns a {@link ServerStreamTracer.Factory} that copies the call ID to the {@link Context}
+   * as {@code SERVER_CALL_ID_CONTEXT_KEY}.
+   */
+  public ServerStreamTracer.Factory getServerCallIdSetter() {
+    return SERVER_CALLID_SETTER_FACTORY;
+  }
+
+  static final ClientInterceptor CLIENT_CALLID_SETTER = new ClientInterceptor() {
```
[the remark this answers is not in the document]
zpencer: Switched to public API

services/src/main/java/io/grpc/services/BinaryLog.java
zpencer: Done. It was a good suggestion because now we can encapsulate two byte[] of size 8.

core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
zpencer: done here and below

services/src/main/java/io/grpc/services/BinaryLog.java
zpencer: done here and below

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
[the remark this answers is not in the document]
zhangkun83: The `unsafe` is already being used in `io/grpc/internal/CensusTracingModule.java`. Using it here doesn't make it any worse. It's your call.

services/src/main/java/io/grpc/services/BinaryLog.java
zhangkun83: Maybe make `CallId` to keep two `long`s? Then the `CallId` can be completely immutable.

zpencer:
@ejona86 PTAL

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -40,6 +46,10 @@
 import javax.annotation.Nullable;
 
 public abstract class BinaryLogProvider implements Closeable {
+  public static final Context.Key<CallId> SERVER_CALL_ID_CONTEXT_KEY
```
ejona86: Package-private or `@Internal`? Ditto for next variable.

services/src/main/java/io/grpc/services/BinaryLog.java
```diff
@@ -77,9 +82,8 @@
   private static final boolean SERVER = true;
   private static final boolean CLIENT = false;
 
-  // TODO(zpencer): extract these fields from call and stop using dummy values
   @VisibleForTesting
-  static final byte[] dumyCallId = new byte[16];
+  static final CallId EMPTY_CALL_ID = new CallId(new byte[8], new byte[8]);
```
ejona86: s/EMPTY_CALL_ID/emptyCallId/. The value is mutable.

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -192,4 +258,29 @@ public InputStream stream(byte[] value) {
...
     }
   }
+
+  /**
+   * A CallId is two byte[] arrays both of size 8 that uniquely identifies the RPC. Users are
+   * free to use the byte arrays however they see fit.
+   */
+  public static final class CallId {
+    private static final byte[] emptyHi = new byte[8];
+    public final byte[] hi;
```
ejona86: If you're storing a high and low portion, why bother with the `byte[]`? It seems to incur the pain of `byte[]` without the benefits. Why not two `long`s or a single `byte[]`?

I guess it doesn't really matter.

zpencer:
Made changes for comments

PR #4267, opened 2018-03-24
okhttp: Convert to internal ConnectionSpec eagerly

ejona86:
This allows ProGuard to remove OkHttp's ConnectionSpec in most cases,
saving about 40 methods. The savings won't be realized until
DEFAULT_CONNECTION_SPEC is removed.


ejona86:
@ericgribkoff, so the DEFAULT_CONNECTION_SPEC is public. That makes it a bit unclear on what to do. It may be best to just give up on this particular detail.

ericgribkoff:
Not sure what to do here, unless we can deprecate `DEFAULT_CONNECTION_SPEC` and eventually delete it (internally, it seems no one is relying on this).  Otherwise the best I can see is having a parallel `INTERNAL_DEFAULT_CONNECTION_SPEC`, with no internal references to OkHttp's `DEFAULT_CONNECTION_SPEC`...this saves exactly 4 methods after proguard, so would not be worth the change.

ejona86:
@ericgribkoff, made the change as we discussed. PTAL

Issue #4198, opened 2018-03-09
ReadableBuffers.BufferInputStream does not release buffer on close()


srujann:
**gRPC Version used: 1.7.0
Netty Version used: 4.1.16.Final
Java version: 1.8.0_161**

Context:
Observed Netty leak detector complaining of bytebuf leaks on the gRPC server configured with Netty Epoll. This occurs more often, a few minutes after the service restarts on a prod machine. Here is the full trace of all 25 access for the bytebuf:
 [netty_buf_leak_trace.txt](https://github.com/grpc/grpc-java/files/1795622/netty_buf_leak_trace.txt)

The most recent access to the leaked bytebuf occurs at `io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:273)` but the bytebuf is actually released here.  Not sure what is causing the leak.

```
 ERROR [2018-03-09 02:25:49,034] io.netty.util.ResourceLeakDetector: LEAK: ByteBuf.release() was not called before it's garbage-collected. See http://netty.io/wiki/reference-counted-objects.html for more information.
 Recent access records: 25
 #25:
 	io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:273)
 	io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
 	io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
 	io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:340)
 	io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1359)
 	io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
 	io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
 	io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:935)
 	io.netty.channel.epoll.AbstractEpollStreamChannel$EpollStreamUnsafe.epollInReady(AbstractEpollStreamChannel.java:797)
 	io.netty.channel.epoll.EpollEventLoop.processReady(EpollEventLoop.java:404)
[...]

PR #4268, opened 2018-03-26
core: add tests for ReadableBuffers

jyane:
Address #4198, part 1. Just adding tests for `ReadableBuffers`.

### DIFF
--- a/core/src/test/java/io/grpc/internal/ReadableBuffersTest.java
+++ b/core/src/test/java/io/grpc/internal/ReadableBuffersTest.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2018, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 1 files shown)


jyane:
@ejona86 Could you please review this? Might reviewer be @ carl-mastrangelo?

core/src/test/java/io/grpc/internal/ReadableBuffersTest.java
```diff
@@ -0,0 +1,127 @@
...
+  @Test
+  public void empty_returnsEmptyBuffer() {
+    ReadableBuffer buffer = ReadableBuffers.empty();
+    assertArrayEquals(new byte[0], buffer.array());
+  }
+
+  @Test(expected = NullPointerException.class)
+  public void readArray_checksNotNull() {
+    ReadableBuffers.readArray(null);
+    fail("Exception expected");
```
ejona86: Fail is unnecessary. `expected` handles the case where no exception is thrown.

jyane: Done.

core/src/test/java/io/grpc/internal/ReadableBuffersTest.java
```diff
@@ -0,0 +1,123 @@
...
+      inputStream.read();
+    }
+    assertEquals(-1, inputStream.read());
+  }
+
+  @Test
+  public void bufferInputStream_read_returnsUnsignedByte() throws Exception {
+    ReadableBuffer buffer = ReadableBuffers.wrap(MSG_BYTES);
+    InputStream inputStream = ReadableBuffers.openStream(buffer, true);
+    // 104 represents 'h' in decimal ascii
```
dapengzhang0: 104 is not used in the code.

jyane: Removed.

jyane:
I think I've addressed all comments :)

ejona86:
Thanks!

PR #4269, opened 2018-03-26
buildscripts: add explicit dependency on appengine SDK

zpencer:
For some reason this is now required to avoid this exception:
java.lang.ClassNotFoundException: com.google.appengine.api.ThreadManager


zpencer:
Test comment

PR #4274, opened 2018-03-27
GAE: call channel.shutdown()

zpencer:
OkHttpClientInteropServlet and NettyClientInteropServlet both run the
`@After` method from AbstractInteropTest. Let's make sure we await
termination.

For the long lived channel test, do the cleanup in `destroy`.


gae-interop-testing/gae-jdk8/src/main/java/io/grpc/testing/integration/LongLivedChannel.java
```diff
@@ -41,6 +42,15 @@
   private final ManagedChannel channel =
       ManagedChannelBuilder.forTarget(INTEROP_TEST_ADDRESS).build();
 
+  @Override
+  public void destroy() {
+    try {
+      channel.shutdownNow().awaitTermination(1, TimeUnit.SECONDS);
+    } catch (Throwable t) {
```
ejona86: s/Throwable/Exception/

Issue #4261, opened 2018-03-23
ok http read header error


smartwjw:
### What version of gRPC are you using?
server: 1.10.0 / jdk8
client:  1.10.0/okhttp

Since nginx 1.13.10 support grpc proxy, we are considering replace envoy with nginx. It works very when client is `netty`, but got exception on Android.

```java
io.grpc.StatusRuntimeException: INTERNAL: Missing HTTP status code
trailers: Metadata(x-request-id=57743f9b-d196-4780-a689-8743e6fa7f9b,grpcjersey.headers-bin=,grpc-status=2,grpc-message=65)
```
After some dig, I found okhttp failed to parse `:status:` and `content-type` header.

![image](https://user-images.githubusercontent.com/1625837/37813085-8e714810-2e9e-11e8-9bfa-51a52a7ee5c8.png)

ericgribkoff:
Thanks for reporting this. It's interesting this is only occurring with OkHttp, as both Netty and OkHttp clients go through the same trailers validation method where that error message is being generated. I likely won't have time to set up nginx and try to reproduce/diagnose this bug until Monday, but I will update this issue once I have some more information.

ericgribkoff:
Update: I can reproduce this reliably. It's not specific to running on an Android device, and can be reproduced with the `--use_okhttp=true` option with our interop test client. 

nginx logs show that it is sending the HTTP status code, but OkHttp is failing to parse it from the response headers. Still digging in to identify the root cause of the bug; since gRPC with Netty is able to parse the HTTP status code, it's likely the bug is in our OkHttp transport rather than nginx.

smartwjw:
@ericgribkoff But, when I change from Nginx back to Envoy, okhttp works well.

ericgribkoff:
@smartwjw OkHttp was throwing out the previously processed headers when updating its dynamic table (used for decoding HPACK headers). This is improper behavior, but apparently "most" gRPC servers didn't set the max table size to zero as nginx is doing, so this bug wasn't encountered previously.

PR #4276 fixes the bug, and I am backporting it to get it in the 1.11.0  release.

PR #4276, opened 2018-03-27
okhttp: fix HPACK reader bug

ericgribkoff:
This fixes https://github.com/grpc/grpc-java/issues/4261.

Our OkHttp HPACK decoder was improperly tossing out the already-parsed header fields when it resized the dynamic table. From the [HPACK RFC section 3.1](https://tools.ietf.org/html/rfc7541#section-3.1):

> Once a header field is decoded and added to the reconstructed header list, the header field cannot be removed.  A header field added to the header list can be safely passed to the application.

This bug appears to have also been present on the [OkHttp 2.7 upstream branch](https://github.com/square/okhttp/blob/okhttp_27/okhttp/src/main/java/com/squareup/okhttp/internal/framed/Hpack.java#L162).

This was caught in #4261 by interop with nginx, which was setting the max dynamic table size to 0. When considering adding an item to the table and finding that item exceeds the max dynamic table size,
 OkHttp [clears](https://github.com/grpc/grpc-java/blob/14003c14ccddf2b30156b7a5a9321b5daad37f7e/okhttp/third_party/okhttp/java/io/grpc/okhttp/internal/framed/Hpack.java#L297) the dynamic table. This is the correct behavior, but it was also deleting the already processed header fields. The HTTP status code was among the fields that was erroneously thrown away, triggering the exception in #4261.

The only substantive change in this PR is the deletion of  line 167 (`headerList.clear()`) in the `clearDynamicTable()` method. The other changes are just two variable renames to make it clear that the values tracked in the dynamic table are *not* the same as the already-processed header values to return to the application in `headerList`.


ericgribkoff:
@ejona86 I locally ran the OkHttp 2.7 tests (https://github.com/square/okhttp/blob/okhttp_27/okhttp-tests/src/test/java/com/squareup/okhttp/internal/framed/HpackTest.java) on this change, and also added a new test case which verifies the fix for this bug. It seems like it might be time to add (at least) these tests to our forked codebase; what do you think?

ericgribkoff:
@ejona86 Split apart into 4 commits (let me know if that's a bit excessive). The first commit contains the actual fix and is backported in #4277

ejona86:
Could this be merged. I'd like it in master before doing the release with the backport.

Issue #4240, opened 2018-03-20
Flag to relax compression interop tests


raboof:
As discussed in #3195, since some implementations don't have access to the message compression flags, tests that rely on that cannot be exercised.

Would it be useful to add a flag that disables test steps that rely on this capability?

With such a flag, we could for example run `client_compressed_unary` against the Java server. The first call in that test would be skipped (as that requires the server to be able to inspect the message flags), but determining whether the subsequent calls succeed would still be useful.

carl-mastrangelo:
@raboof if this affects more than java, a change would be needed for all implementations (i.e. grpc and grpc-go)

Does this only affect the grpc java tests?

ejona86:
I think I'd prefer just creating a new test that doesn't do the initial server feature check. Say, `client_compressed_unary_novalidate`

ejona86:
@raboof, feel free to send a PR with the test addition. We can provide guidance for pieces that are unclear.

raboof:
Done in #4279

PR #4279, opened 2018-03-28
Allow client compressed tests without probing (#4240)

raboof:
I chose the `_noprobe` suffix instead of `_novalidate` since it more accurately
describes the difference: the validations are the same, but the new variations
skip the feature-probing call.


thelinuxfoundation:

* User @raboof isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

thelinuxfoundation:

* User @raboof isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

raboof:
(signed the CLA)

ericgribkoff:
Could you elaborate on the motivation for skipping the initial RPC? Generally speaking, the test cases here come from the cross-language specification in https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md. The compression tests are already a little out-of-place for Java, since we can't inspect the compression level as is done in C++, so these are only really useful when run against the C++ server. From the comments on the new methods in this PR, this restriction also applies when the initial probing RPC is skipped.

raboof:
@ericgribkoff sure! there's a little more background in #4240.

In short, servers that do not provide access to the message-level compression flag (like grpc-java, but I think this might go for other implementations?) cannot implement the server-side of the feature probe request. Because of that, running this test against such a server will fail at the feature probe step.

When skipping the feature probe step, running this test against such a server will be able to exercise the compression and check for interoperability by checking the messages arrive in good order - even though it is indeed not explicitly checked that they are, in fact, compressed.

ericgribkoff:
Ah, got it, thanks - I missed the discussion in #4240. Just kicked off the tests for this PR.

interop-testing/src/main/java/io/grpc/testing/integration/TestCases.java
```diff
@@ -26,9 +26,11 @@
   CACHEABLE_UNARY("cacheable unary rpc sent using GET"),
   LARGE_UNARY("single request and (large) response"),
   CLIENT_COMPRESSED_UNARY("client compressed unary request"),
+  CLIENT_COMPRESSED_UNARY_NOPROBE("client compressed unary request (skip initial feature-probing request)"),
```
ericgribkoff: This line (and `CLIENT_COMPRESSED_STREAMING_NOPROBE`) exceeds 100 characters. Can you change them to the format:

```
  CLIENT_COMPRESSED_UNARY_NOPROBE(
      "client compressed unary request (skip initial feature-probing request)"),
```

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -391,6 +391,18 @@ public void largeUnary() throws Exception {
...
    */
   public void clientCompressedUnary() throws Exception {
+    clientCompressedUnary(true);
+  }
+
+  /**
+   * Tests client per-message compression for unary calls. The Java API does not support inspecting
+   * a message's compression level, so this is primarily intended to run against a gRPC C++ server.
+   */
+  public void clientCompressedUnaryNoProbe() throws Exception {
```
ericgribkoff: Didn't catch this earlier - checkstyle fails here due to the separation between `clientCompressedUnary()` and `clientCompressedUnary(boolean)`. I'm ok with just dropping the no-arg methods (`clientCompressedUnary()` and `clientCompressedUnaryNoProbe()`) and directly invoking `clientCompressedUnary(true or false)` in `TestServiceClient`. Same for clientCompressedStreaming

PR #4282, opened 2018-03-29
interop-testing: fix empty proto package name

### LINKED ISSUE #4241: message Empty is defined twice in interop-testing sub-project
Message `Empty` is once defined in the `interop-testing` sub-project itself:

https://github.com/grpc/grpc-java/blob/95917b8f280115bdb946e455151034a2650f48a6/interop-testing/src/main/proto/io/grpc/testing/integration/empty.proto#L22-L30

And then again pulled in from `protobuf-java`

https://github.com/google/protobuf/blob/d6a17aadeb6fc306d3d8c3b5c07edc41f6bc551e/src/google/protobuf/empty.proto#L43-L52

via the dependency chain: `grpc-interop-testing` -> `grpc-protobuf` -> `protobuf-java-util` -> `protobuf-java`

This becomes a problem when generating grpc classes using ScalaPB as it tries to place both under the [same package and the same name](https://github.com/scalapb/ScalaPB/issues/243#issuecomment-279769902).

I would expect `interop-testing` to not define `Empty` message and use the one from  `protobuf-java` instead. Would that be possible to change?

jyane:
Intended to resolve #4241
grpc-java has already diverged this proto file, so changing the
this package name is reasonable for now.

To: @ejona86


jyane:
I've noticed that proto version is already different.
https://github.com/grpc/grpc/blob/fa7cbb50fe40962cf443f43cabe38e98f9f162a3/src/proto/grpc/testing/empty.proto#L16

Updating the proto version would be necessary. Other proto files' proto version might be different as well.

ejona86:
Sigh... I thought we fixed all the proto2 files to proto3 already...

Don't worry about it, unless you'd like to do the cleanup. The swap for Empty should be trivial. That should probably still be a separate PR.

carl-mastrangelo:
Thanks @jyane

PR #4288, opened 2018-03-31
java_grpc_library: re-add grpc-java repo to targets

ejona86:
They were removed in 137c74d1 since it was believed they were unnecessary.
However, since they are in a macro and not a rule, they are relative to the
caller, not their definition.

Added building the examples to the kokoro CI. Note that this means the examples
are built twice: once in grpc-java's build and once in their own (because it
has a WORKSPACE). Given that the Bazel build is our fastest build, this
slowdown won't probably be an issue.

CC @cgrushko

### DIFF
--- a/buildscripts/kokoro/bazel.sh
+++ b/buildscripts/kokoro/bazel.sh
@@ -5,3 +5,6 @@ cat /VERSION
 
 cd github/grpc-java
 bazel build ...
+
+cd examples
+bazel build ...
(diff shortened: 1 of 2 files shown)


cgrushko:
Mmm.. what you're saying makes sense, but I removed the repo name because it unbroke my build. But perhaps I mixed things up and it wasn't actually my problem.

ejona86:
@cgrushko, there was one instance I didn't revert, because it was in the rule() instead. That did behave like you were saying, so I didn't need to put the repo prefix.

I'm working on a version that uses java_common to fix the long-standing stict_dep issues, which will make all of this move into the rule. I still have to work out a problem with android, but I think I'll use java_import for that. Hopefully this change will become moot soon.

dapengzhang0:
@ejona86 is this going to backport to v1.11.x?

PR #4289, opened 2018-03-31
java_grpc_library: Add support for protobuf lite

ejona86:
gRPC's protobuf-lite auto-selects between full and lite protobuf based on the
value of crosstool_top. If the user is specifying their own
--android_crosstool_top, then it will not auto-detect correctly. One day,
platforms will fix problems like this, but for the moment it seems we get to
live with it.

This includes #4288 as the first commit.

Note: I would have added a BUILD for the helloworld Android app, but ran
into unrelated problems ("cannot access TaskStackBuilder").


ejona86:
CC @codesuki

codesuki:
Thanks! That looks good. You didn't even need to add a new rule. I will give it a spin when I am back from vacation. 
Sorry I couldn't help after all.

codesuki:
I am on mobile and didn't dig any deeper, let me ask a stupid question. Why does the 'protobuf-lite' rule even need the full protobuf in the default case. 
```
"//conditions:default": ["@com_google_protobuf//:protobuf_java"],
```
I am not saying it's not correct just thinking it's unintuitive :)

ejona86:
@codesuki, so... Lite is a _subset_ of full proto. Or rather, it was a subset _before the fork_. The fork should be temporary, but it will be with us a while longer.

grcp-protobuf-lite uses only the subset of full proto supported by lite. But it can work with both full and lite protobufs. grpc-protobuf actually depends on grpc-protobuf-lite. However, since Lite is a fork now, it's dangerous to depend on both full and lite protobuf as you'll get conflicts in the classpath. Thus we have to select which version we want.

ejona86:
Oh, and since it is dangerous to mix lite and full protobuf, I am sort of imposing the solution we have done internally: use lite if and only if on Android.

protobuf-lite/BUILD.bazel
```diff
@@ -1,14 +1,26 @@
...
-    ],
+    ] + select({
+        ":android": ["@com_google_protobuf_javalite//:protobuf_java_lite"],
+        "//conditions:default": ["@com_google_protobuf//:protobuf_java"],
+    }),
+)
+
+# This config is not fully-reliable. If it breaks, it is probably because you
+# are changing --android_crosstool_top. Instead of doing that, you can bind
+# your own toolchain on top of the default android/closstool, as mentioned at
```
ericgribkoff: s/closstool/crosstool

ejona86:
Coverage is blocked by getting a `android_library` in the build, yes.

> I manually built protobuf-lite with --crosstool_top=//external:android/crosstool which I think demonstrates that this works

It shows that grpc-protobuf-lite can be built against protobuf-lite. It doesn't show the rest of the stuff works. And it doesn't show that the `select()` is right. And such an argument command could break other targets. So such a test seems pretty limited use at the moment. I'd much rather introduce an android_library of some sort.

protobuf-lite/BUILD.bazel
ejona86: Done.

Issue #4278, opened 2018-03-27
OkHttpChannelBuilder#overrideAuthority not compatible with IPv6 hostname


OnlyInAmerica:
### What version of gRPC are you using?
1.10.0

### What did you expect to see?
Background: I'm using mutual TLS with a LAN client that presents an IPv6 link-local address as an IP SAN in its X509 certificate.

I expected to call `OkHttpChannelBuilder.overrideAuthority("FE80::FABB:BFFF:FE01:3A52")` to allow hostname verification to pass.

Instead, this call yields an `IllegalArgumentException("No host in authority")`. If I instead represent the authority as `"[FE80::FABB:BFFF:FE01:3A52]"`, `OkHostnameVerifier`'s `VERIFY_AS_IP_ADDRESS` regex does not match the address (correctly, I believe, as the square brackets are just a URL disambiguation detail and not part of the hostname), and so the verify mechanism ignores the IP SAN entries.

If you agree this behavior should be changed I'm happy to submit a PR. Perhaps we'd modify `GrpcUtil#checkAuthority` to strip square brackets from a validated hostname?

### Message Sequence Charts (Click to expand)
With `OkHttpChannelBuilder.overrideAuthority("FE80::FABB:BFFF:FE01:3A52")`:

![link-local-hostname-verification-okhostnameverify-fail-nobracket-msc](https://user-images.githubusercontent.com/458510/37923791-e9bf1240-30e4-11e8-8cfb-d05ddf3c5d3d.png)

With `OkHttpChannelBuilder.overrideAuthority("[FE80::FABB:BFFF:FE01:3A52]")`:

![link-local-hostname-verification-okhostnameverify-fail-msc](https://user-images.githubusercontent.com/458510/37923865-17711bde-30e5-11e8-9b05-08606d8f70df.png)

dapengzhang0:
Are you using JDK version earlier than jdk_8u60 or on Mac OS?

OnlyInAmerica:
I am on Mac OS, using Android Studio 3.1's bundled JDK which appears to be 1.8.0_152 (equivalent to jdk_8u152?)

ericgribkoff:
[...]

PR #4292, opened 2018-04-02
okhttp: properly verify IPv6 address hosts

OnlyInAmerica:
Address mismatch between IPv6 address hosts derived from URIs and X509 subjectAltName extensions as discussed in #4278

### DIFF
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpTlsUpgrader.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpTlsUpgrader.java
@@ -90,7 +90,7 @@ final class OkHttpTlsUpgrader {
   @VisibleForTesting
   static String canonicalizeHost(String host) {
     if (host.startsWith("[") && host.endsWith("]")) {
-      return host.substring(1, host.length()-1);
+      return host.substring(1, host.length() - 1);
     }
     return host;
   }
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpTlsUpgrader.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpTlsUpgrader.java
@@ -87,6 +87,7 @@ final class OkHttpTlsUpgrader {
    *
    * @return {@param host} in a form consistent with X509 certificates
    */
+  @VisibleForTesting
   static String canonicalizeHost(String host) {
     if (host.startsWith("[") && host.endsWith("]")) {
       return host.substring(1, host.length()-1);
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpTlsUpgrader.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpTlsUpgrader.java
@@ -70,9 +70,24 @@ final class OkHttpTlsUpgrader {
     if (hostnameVerifier == null) {
       hostnameVerifier = OkHostnameVerifier.INSTANCE;
     }
-    if (!hostnameVerifier.verify(host, sslSocket.getSession())) {
+    if (!hostnameVerifier.verify(canonicalizeHost(host), sslSocket.getSession())) {
       throw new SSLPeerUnverifiedException("Cannot verify hostname: " + host);
     }
     return sslSocket;
+  }
+
+  /**
+   * Converts a host from URI to X509 format.
+   *
+   * <p>IPv6 host addresses derived from URIs are enclosed in square brackets per RFC2732, but
+   * omit these brackets in X509 certificate subjectAltName extensions per RFC5280.
+   *
+   * @see <a href="https://www.ietf.org/rfc/rfc2732.txt">RFC2732</a>
+   * @see <a href="https://tools.ietf.org/html/rfc5280#section-4.2.1.6">RFC5280</a>
+   *
+   * @return {@param host} in a form consistent with X509 certificates
+   */
+  static String canonicalizeHost(String host) {
+    return host.replaceAll("[\\[\\]]", "");
   }
 }
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpTlsUpgrader.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpTlsUpgrader.java
@@ -88,6 +88,9 @@ final class OkHttpTlsUpgrader {
    * @return {@param host} in a form consistent with X509 certificates
    */
   static String canonicalizeHost(String host) {
-    return host.replaceAll("[\\[\\]]", "");
+    if (host.startsWith("[") && host.endsWith("]")) {
+      return host.substring(1, host.length()-1);
+    }
+    return host;
   }
 }
--- a/okhttp/src/test/java/io/grpc/okhttp/OkHttpTlsUpgraderTest.java
+++ b/okhttp/src/test/java/io/grpc/okhttp/OkHttpTlsUpgraderTest.java
@@ -16,6 +16,8 @@
 
 package io.grpc.okhttp;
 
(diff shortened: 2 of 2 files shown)


okhttp/src/main/java/io/grpc/okhttp/OkHttpTlsUpgrader.java
```diff
@@ -70,9 +70,24 @@ public static SSLSocket upgrade(SSLSocketFactory sslSocketFactory,
...
+   *
+   * <p>IPv6 host addresses derived from URIs are enclosed in square brackets per RFC2732, but
+   * omit these brackets in X509 certificate subjectAltName extensions per RFC5280.
+   *
+   * @see <a href="https://www.ietf.org/rfc/rfc2732.txt">RFC2732</a>
+   * @see <a href="https://tools.ietf.org/html/rfc5280#section-4.2.1.6">RFC5280</a>
+   *
+   * @return {@param host} in a form consistent with X509 certificates
+   */
+  static String canonicalizeHost(String host) {
```
ericgribkoff: This can be private

okhttp/src/main/java/io/grpc/okhttp/OkHttpTlsUpgrader.java
```diff
@@ -70,9 +70,24 @@ public static SSLSocket upgrade(SSLSocketFactory sslSocketFactory,
...
+    return host.replaceAll("[\\[\\]]", "");
```
ericgribkoff: This may be safe as-is, but we should be more conservative here: if the first and last characters are `[` and `]`, strip only those two characters out. Otherwise, just return the string.

This would mean that we can add a test like `assertEquals("[someIp]", canonicalizeHost("[[someIp]]"));`. This address would, as in #4278, result in `OkHostnameVerifier` failing to match the host - but that's good, because having two brackets is malformed anyways, and probably indicates a bug somewhere else that we don't want to mask here.

okhttp/src/test/java/io/grpc/okhttp/OkHttpTlsUpgraderTest.java
```diff
@@ -32,4 +34,44 @@
...
+    assertEquals(canonicalizeHost("1080:0:0:0:8:800:200C:417A"), "1080:0:0:0:8:800:200C:417A");
+    assertEquals(canonicalizeHost("[1080:0:0:0:8:800:200C:417A]"), "1080:0:0:0:8:800:200C:417A");
+    assertEquals(canonicalizeHost("1080::8:800:200C:417A"), "1080::8:800:200C:417A");
+    assertEquals(canonicalizeHost("[1080::8:800:200C:417A]"), "1080::8:800:200C:417A");
+    assertEquals(canonicalizeHost("FF01::101"), "FF01::101");
+    assertEquals(canonicalizeHost("[FF01::101]"), "FF01::101");
+    assertEquals(canonicalizeHost("0:0:0:0:0:0:13.1.68.3"), "0:0:0:0:0:0:13.1.68.3");
+    assertEquals(canonicalizeHost("[0:0:0:0:0:0:13.1.68.3]"), "0:0:0:0:0:0:13.1.68.3");
+    assertEquals(canonicalizeHost("0:0:0:0:0:FFFF:129.144.52.38"), "0:0:0:0:0:FFFF:129.144.52.38");
+    assertEquals(canonicalizeHost("[0:0:0:0:0:FFFF:129.144.52.38]"), "0:0:0:0:0:FFFF:129.144.52.38");
```
ericgribkoff: I think this fails checkstyle (100 char limit per line)

okhttp/src/test/java/io/grpc/okhttp/OkHttpTlsUpgraderTest.java
```diff
@@ -32,4 +34,44 @@
             || OkHttpTlsUpgrader.TLS_PROTOCOLS.indexOf(Protocol.GRPC_EXP)
                 < OkHttpTlsUpgrader.TLS_PROTOCOLS.indexOf(Protocol.HTTP_2));
   }
+
+  @Test public void upgrade_canonicalizeHost() {
+    // IPv4
+    assertEquals(canonicalizeHost("127.0.0.1"), "127.0.0.1");
```
ericgribkoff: The argument order for `assertEquals` is reversed. These should be in the form: `assertEquals("ip", canonicalizeHost("ip"))`. (this results in better output when a test fails, as error message will label the first argument as the expected value and the second argument as the actual value)

okhttp/src/test/java/io/grpc/okhttp/OkHttpTlsUpgraderTest.java
```diff
@@ -32,4 +34,44 @@
             || OkHttpTlsUpgrader.TLS_PROTOCOLS.indexOf(Protocol.GRPC_EXP)
                 < OkHttpTlsUpgrader.TLS_PROTOCOLS.indexOf(Protocol.HTTP_2));
   }
+
+  @Test public void upgrade_canonicalizeHost() {
```
ericgribkoff: Tests coverage look good to me (with the addition of a case to ensure we don't strip out additional brackets for malformed URIs). Out of curiosity, what methodology went into choosing the test hosts?

OnlyInAmerica: Will update to ensure we don't strip out additional brackets. 

I lifted these from okhttp's [HostnameVerifierTest](https://github.com/square/okhttp/blob/master/okhttp-tests/src/test/java/okhttp3/internal/tls/HostnameVerifierTest.java#L519), so I'll add some attribution.

okhttp/src/main/java/io/grpc/okhttp/OkHttpTlsUpgrader.java
OnlyInAmerica: Is package-private with `@VisibleForTesting` (ala `TLS_PROTOCOLS`) acceptable?

ericgribkoff: Oops, my mistake - yes, the visibility is fine, but the `@VisibleForTesting` annotation would be helpful.

okhttp/src/test/java/io/grpc/okhttp/OkHttpTlsUpgraderTest.java
ericgribkoff: Let's simplify the tests somewhat here, which will also remove the need for attribution - we actually don't care about all of the edge cases that the linked test case covers, because we don't need to distinguish between IP addresses and hostnames. Probably just the following are enough:

```
assertEquals("::1", canonicalizeHost("::1"));
assertEquals("::1", canonicalizeHost("[::1]"));
assertEquals("127.0.0.1", canonicalizeHost("127.0.0.1"));
assertEquals("some.long.url.com", canonicalizeHost("some.long.url.com"));

// Extra square brackets in a malformed URI are retained
assertEquals("[::1]", canonicalizeHost("[[::1]]"));
```

Also, does it make sense to rename the test method to just `canonicalizeHost`, as this isn't testing any other part of the upgrade logic?

OnlyInAmerica: 👍 Is `canonicalizeHosts` ok to avoid clash with the static import of `canonicalizeHost`?

ericgribkoff: sounds good to me

OnlyInAmerica:
fixed whitespace issue to appease travis. One more time from the top 🥁

ericgribkoff:
Thanks!

Issue #3469, opened 2017-09-15
Could not add interceptor(s) to `ProtoReflectionService`


jyane:
### What did you do?
Add interceptor(s) to `ProtoReflectionService`.

### What did you expect to see?
`ProtoReflectionService` works with interceptor(s).

### What did you see instead?
Got a NullPointerException or compilation error.

### Problem
I ask a question on Gitter.
https://gitter.im/grpc/grpc?at=59b9090bcfeed2eb65033bad
and I have got a reply from @e.jona86.
https://gitter.im/grpc/grpc?at=59b9a7f9c101bc4e3ac3404a
I think it is not easy to fix. So I report an issue.

`ProtoReflectionService` needs a reference to `Server` via `notifyOnBuild` but it could not, because an intercepted service is an instance of `ServerServiceDefinition`.

I think it is consisting some problems (may not a bug or may not a problem) as below.
- `ServerInterceptors#intercept` binds a service and returns `ServerServiceDefinition`.
- `ServerBuilder#addService` doesn't add `ServerServiceDefinition` to `notifyOnBuildList`.
- `ServerServiceDefinition` doesn't implement `InternalNotifyOnBuild`.

Therefore, I could not add interceptor(s) to `ProtoReflectionService`.

ejona86:
@ericgribkoff, were we aware of this limitation during the design?

ericgribkoff:
@ejona86 No, I don't think so. This was an oversight on my part. I'll look into getting the mechanism working with server interceptors.

ejona86:
As I mentioned in #3449, we could add a field of a method to run to `ServerServiceDefinition` as a replacement for `InternalNotifyOnBuild`.

Note: you may be able to use `ServerBuilder.intercept()` as a workaround. As long as you're okay with the interceptor applying to all services, that would work.

jyane:
Thank you for your reply, it works. But I want to add interceptor(s) to each services.
I'm waiting for resolving this issue because it seems to be hard to resolve for me...

Issue #3449, opened 2017-09-08
Feature: Lifecycle evens for service implementations


rmichela:
I'd like my service implementations to be notified of server lifecycle events for server start and shutdown. These operations could be used to warm caches, persist data, or close files.

The least intrusive implementation would be an interface a service implementer would opt into for receiving these events. Ex:
```java
public interface ServiceLifecycle {
  // Called when Server.start() is called
  void onStart(...);
 
  // Called when Server.shutdown() or Server.shutdownNow() is called
  void onShutdown(boolean hard, ...);
}
```

My initial thought is that `onStart()` would block `Server.start()`, so that the server wouldn't be ready until all services had successfully started. `onShutdown()` is trickier. It needs to be called only after the server has drained all requests.

ejona86:
During shutdown we already know when all requests have been drained, as this is when we transition to terminated. So that's feasible.

We used that sort of interface for Reflection, but someone noticed (I thought they were going to file an issue, but I don't see it...) that using an interceptor can break the interface-detection.

That interface would mean a service could only be used in one server. That's probably fine much of the time, but would maybe encourage people to want a server to be able to natively listen on multiple ports.

At this point we've expected people to do this sort of thing outside of gRPC. We'd need to think of implications doing it within, for example, error handling.

I would sort of expect people to find that interface convenient, as long as they aren't in some DI world that normally orchestrates startup/shutdown.

rmichela:
I looked at how service lifecycle events could be implemented _on top_ of gRPC using an interceptor, rather than from within gRPC as a code change. The advantage would be less code in gRPC and an opt-in feature for those who are interested in it.

What I discovered is that there is no way to get from the `MethodDescriptor` returned by `ServerCall.getMethodDescriptor()` available in `ServerInterceptor.interceptCall()` to the actual class and method implementing the requested operation. The generated `MethodDescriptor` constants only contain information about the proto, with nothing about the binding itself.

Is there some way we can link a `MethodDescriptor` to the actual `java.lang.reflect.Method` that the descriptor is bound to? If we did this, I could implement this use case entirely outside of gRPC-java.

ejona86:
#3469 is the issue with the existing `InternalNotifyOnBuild` used by reflection.

The [`ServerMethodDefinition`](https://grpc.io/grpc-java/javadoc/io/grpc/ServerMethodDefinition.html) is what specifies how to run the method. But the `ServerCallHandler` will be re-created when using interception.

[...]

PR #4293, opened 2018-04-03
Support interceptors for the reflection service

ericgribkoff:
This is intended to address https://github.com/grpc/grpc-java/issues/3469. The mechanism that replaces `InternalNotifyOnServerBuild` is `Server.LifecycleListener`, which half-addresses the request raised in https://github.com/grpc/grpc-java/issues/3449 (this PR only adds the `onStart` event, `onShutdown` would still be needed).

The two commits can be reviewed separately.


ejona86:
I remember we discussed this offline. I'm trying to remember our verdict. I remember we were considering two other alternatives, one being adding `ServerCall.getServer()`. I don't recall the other one. @ericgribkoff, do you remember?

PR #4295, opened 2018-04-03
upgrade protobuf-maven-plugin

jorgheymans:
due to https://github.com/xolstice/protobuf-maven-plugin/issues/30, out of the box user experience on windows for compiling protobufs was often suboptimal, needing to retry several times before the compilation succeeded. We have not seen the error since upgrading to the latest version.


jorgheymans:
signed

PR #4297, opened 2018-04-03
core: don't reschedule idle timer if it is already active

carl-mastrangelo:
@ejona86 @zhangkun83 

I haven't quite got the tests to pass yet but I think this is close enough for review to take a look.   The main idea is to avoid rescheduling the IdleModeTimer if it is already active.  This has pretty good results on my machine for the latency benchmarks.  With CPU freq scaling on, median latency goes from 197us to 185us.    (no TLS, but with census enabled, direct executor).

The changes to get the tests to work are going to be harder since this change depends on `System.nanoTime`.  The `FakeClock` we use in our tests doesn't make it easy to mock this out, so I'm working on ideas to fix this.  That said, it is starting to get invasive for tests to pass.

Raw numbers from my runs:
```
Before (w/freq scaling)
50.0%ile Latency (in nanos):		197071
90.0%ile Latency (in nanos):		253383
95.0%ile Latency (in nanos):		272479
99.0%ile Latency (in nanos):		318191
99.9%ile Latency (in nanos):		411519
100.0%ile Latency (in nanos):		16083967
QPS:                           5005


After:
50.0%ile Latency (in nanos):		185023
90.0%ile Latency (in nanos):		240671
95.0%ile Latency (in nanos):		259319
99.0%ile Latency (in nanos):		299887
99.9%ile Latency (in nanos):		379855
100.0%ile Latency (in nanos):		11492351
QPS:                           5320
```

```
Before (w/o freq scaling)

50.0%ile Latency (in nanos):		63907
90.0%ile Latency (in nanos):		73055
95.0%ile Latency (in nanos):		79443
99.0%ile Latency (in nanos):		93739
99.9%ile Latency (in nanos):		123583
100.0%ile Latency (in nanos):		14028287
QPS:                           14936

After:
After:
50.0%ile Latency (in nanos):		57355
90.0%ile Latency (in nanos):		69459
95.0%ile Latency (in nanos):		78495
99.0%ile Latency (in nanos):		93575
99.9%ile Latency (in nanos):		126819
100.0%ile Latency (in nanos):		12634111
QPS:                           16435
```

### DIFF
--- a/core/src/main/java/io/grpc/internal/KeepAliveManager.java
+++ b/core/src/main/java/io/grpc/internal/KeepAliveManager.java
@@ -293,7 +293,7 @@ public class KeepAliveManager {
     public abstract long read();
   }
 
-  static class SystemTicker extends Ticker {
+  private static class SystemTicker extends Ticker {
     @Override
     public long read() {
       return System.nanoTime();
--- a/core/src/main/java/io/grpc/internal/KeepAliveManager.java
+++ b/core/src/main/java/io/grpc/internal/KeepAliveManager.java
@@ -293,7 +293,7 @@ public class KeepAliveManager {
     public abstract long read();
   }
 
-  private static class SystemTicker extends Ticker {
+  static class SystemTicker extends Ticker {
     @Override
     public long read() {
       return System.nanoTime();
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -67,6 +67,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -307,26 +308,12 @@ final class ManagedChannelImpl extends ManagedChannel implements Instrumented<Ch
 
   // Run from channelExecutor
   private class IdleModeTimer implements Runnable {
-    // Only mutated from channelExecutor
-    boolean cancelled;
 
     @Override
     public void run() {
-      if (cancelled) {
-        // Race detected: this task was scheduled on channelExecutor before cancelIdleTimer()
-        // could cancel the timer.
-        return;
-      }
       enterIdleMode();
     }
   }
-
-  // Must be used from channelExecutor
-  @Nullable
-  private ScheduledFuture<?> idleModeTimerFuture;
-  // Must be used from channelExecutor
-  @Nullable
-  private IdleModeTimer idleModeTimer;
 
   // Must be called from channelExecutor
   private void shutdownNameResolverAndLoadBalancer(boolean verifyActive) {
@@ -360,7 +347,7 @@ final class ManagedChannelImpl extends ManagedChannel implements Instrumented<Ch
     if (inUseStateAggregator.isInUse()) {
       // Cancel the timer now, so that a racing due timer will not put Channel on idleness
       // when the caller of exitIdleMode() is about to use the returned loadBalancer.
-      cancelIdleTimer();
+      cancelIdleTimer(false);
     } else {
       // exitIdleMode() may be called outside of inUseStateAggregator.handleNotInUse() while
       // isInUse() == false, in which case we still need to schedule the timer.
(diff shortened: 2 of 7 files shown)


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -395,31 +382,100 @@ private void enterIdleMode() {
...
+        } else {
+          enabled = false;
+          wakeUp = null;
+          runnable.run();
+        }
+      }
+    }
+
+    /** Time source representing nanoseconds since fixed but arbitrary point in time. */
+    interface Ticker {
```
carl-mastrangelo: FYI: Guava's ticker is an abstract class, and marked `@Beta`.  I can't use it here, so I define an interface.

ejona86: I toyed with this some using Stopwatch. I think it ends up fine (no unnecessary now() calls in the normal paths), and it is a stable API.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -395,31 +382,100 @@ private void enterIdleMode() {
...
+    private final ChannelExecutor exec;
+    private final ScheduledExecutorService scheduler;
+
+    private ScheduledFuture<?> wakeUp;
+
+    // state
+    private long runAt;
+    private boolean enabled;
+
+    Rescheduler(Runnable r, ChannelExecutor exec, ScheduledExecutorService scheduler) {
```
zhangkun83: There is a way to decouple Rescheduler from ChannelExecutor.  Pass an `Executor` here and make it an requirement that it must serialize its runnables and `rechedule()`. In `ManagedChannelImpl`, you could make an `Executor` that delegates to `ChannelExecutor`. I am strong for splitting this class out of `ManagedChannelImpl`, which is good for test-ability and also for the fitness of `ManagedChannelImpl`.

carl-mastrangelo: Done.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
carl-mastrangelo: Switched to using stopwatch.

core/src/main/java/io/grpc/internal/KeepAliveManager.java
```diff
@@ -293,7 +293,7 @@ public void onPingTimeout() {
     public abstract long read();
   }
 
-  private static class SystemTicker extends Ticker {
+  static class SystemTicker extends Ticker {
```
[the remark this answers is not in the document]
carl-mastrangelo: Reverted, this was originally going to be reused.

core/src/test/java/io/grpc/internal/FakeClock.java
```diff
@@ -317,4 +317,4 @@ public long currentTimeMillis() {
      */
     boolean shouldAccept(Runnable runnable);
   }
-}
+}
```
ejona86: Revert changes to this file?

core/src/main/java/io/grpc/internal/Rescheduler.java
```diff
@@ -0,0 +1,119 @@
...
+      }
+    }
+  }
+
+  @VisibleForTesting
+  static boolean isEnabled(Runnable r) {
+    return ((FutureRunnable) r).rescheduler.enabled;
+  }
+
+  private long nanoTime() {
```
ejona86: I see what you did there. :smile:

core/src/main/java/io/grpc/internal/Rescheduler.java
```diff
@@ -0,0 +1,119 @@
...
+  // must be called from channel executor
+  void cancel(boolean permanent) {
+    enabled = false;
+    if (permanent && wakeUp != null) {
+      wakeUp.cancel(false);
+      wakeUp = null;
+    }
+  }
+
+  private static final class FutureRunnable implements Runnable {
```
ejona86: nit: Why is this static?

carl-mastrangelo: In `ManagedChannelImplIdlenessTest.java` there is a test that checks there are no more scheduled tasks in the scheduler.   It fails because the tasks are still left over, but now disabled.   I wanted to make it so the test scans the tasks and checks to see they are Rescheduler Runnables, and that they are disabled.

There needs to be a reference from the Runnable back to the Rescheduler, so this became static and took an explicit reference to the outer class.  This is so `isEnabled` below works.

core/src/test/java/io/grpc/internal/FakeClock.java
carl-mastrangelo: Done.

core/src/test/java/io/grpc/internal/ReschedulerTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    assertTrue(runner.ran);
+  }
+
+  @Test
+  public void cancels() {
+    assertFalse(runner.ran);
+    rescheduler.reschedule(1, TimeUnit.NANOSECONDS);
+    assertFalse(runner.ran);
+    rescheduler.cancel(/* permanent= */ false);
```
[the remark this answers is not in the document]
carl-mastrangelo: Done.

core/src/test/java/io/grpc/internal/ReschedulerTest.java
```diff
@@ -0,0 +1,105 @@
...
+    assertTrue(exec.executed);
+  }
+
+  @Test
+  public void reschedules() {
+    assertFalse(runner.ran);
+    rescheduler.reschedule(1, TimeUnit.NANOSECONDS);
+    assertFalse(runner.ran);
+    assertFalse(exec.executed);
+    rescheduler.reschedule(50, TimeUnit.NANOSECONDS);
```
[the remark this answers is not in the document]
carl-mastrangelo: Done.

core/src/main/java/io/grpc/internal/Rescheduler.java
ejona86: Ah, I see. You could have added another method to FutureRunnable to do something like `return Rescheduler.this`, but that's close to the same amount of effort. Makes sense.

PR #4300, opened 2018-04-03
core,netty,okhttp,services,testing: expose tls security info to channelz

zpencer:
Pull the TLS info from the SSLSession object. Run AbstractInteropTest
with TLS enabled for okhttp and netty flavors.


netty/src/main/java/io/grpc/netty/NettyServerTransport.java
```diff
@@ -59,6 +63,7 @@
   private final ChannelPromise channelUnused;
   private final ProtocolNegotiator protocolNegotiator;
   private final int maxStreams;
+  private NettyServerHandler grpcHandler;   // Only set once during start()
```
carl-mastrangelo: this will still be racy if there isnt a HB between assigning and accessing.   start() does not HB event loop access

core/src/main/java/io/grpc/internal/Channelz.java
```diff
@@ -450,21 +454,124 @@ public ChannelStats build() {
...
+      String localCert = null;
+      String remoteCert = null;
+      Certificate[] localCerts = session.getLocalCertificates();
+      if (localCerts != null) {
+        localCert = localCerts[0].toString();
+      }
+      try {
+        Certificate[] peerCerts = session.getPeerCertificates();
+        if (peerCerts != null) {
+          remoteCert = peerCerts[0].toString();
```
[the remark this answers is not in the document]
zpencer: That's correct, I can add a comment since this is non obvious. According to the javadoc for `getPeerCertificates`:
  an ordered array of peer certificates, with the peer's own certificate first followed by any certificate authorities.

core/src/main/java/io/grpc/internal/Channelz.java
```diff
@@ -450,21 +454,124 @@ public ChannelStats build() {
...
+        }
+      } catch (SSLPeerUnverifiedException e) {
+        // peer cert is not available
```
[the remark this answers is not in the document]
zpencer: added log

netty/src/main/java/io/grpc/netty/NettyServerTransport.java
zpencer: turned field into volatile

zpencer:
@carl-mastrangelo PTAL

zpencer:
@carl-mastrangelo ping

okhttp/build.gradle
```diff
@@ -8,6 +13,7 @@ dependencies {
     testCompile project(':grpc-core').sourceSets.test.output,
                 project(':grpc-testing'),
                 project(':grpc-netty')
+    tls libraries.conscrypt, libraries.netty_tcnative
```
carl-mastrangelo: @ejona86 FYI

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -489,9 +502,9 @@ public void close() {}
 
         FrameWriter rawFrameWriter;
         synchronized (lock) {
-          socket = sock;
+          socket = Preconditions.checkNotNull(sock);
```
carl-mastrangelo: fyi, we typically add a comment to checkNotNull to say which thing was null.  Its very helpful in user reported bugs since the line numbers may be off.

okhttp/build.gradle
ejona86: @ericgribkoff, FYI

core/src/main/java/io/grpc/internal/Channelz.java
```diff
@@ -450,21 +457,127 @@ public ChannelStats build() {
...
+      try {
+        Certificate[] peerCerts = session.getPeerCertificates();
+        if (peerCerts != null) {
+          // The javadoc of getPeerCertificate states that the peer's own certificate is the first
+          // element of the list.
+          remoteCert = peerCerts[0].toString();
+        }
+      } catch (SSLPeerUnverifiedException e) {
+        // peer cert is not available
+        log.log(Level.FINE, "Peer cert not available for peerHost={}", session.getPeerHost());
```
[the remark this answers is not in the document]
zpencer: Included

netty/src/main/java/io/grpc/netty/NettyClientTransport.java
```diff
@@ -354,6 +344,25 @@ public void operationComplete(Future<Object> future) throws Exception {
     return result;
   }
 
+  private SocketStats getStatsHelper(Channel ch) {
+    Preconditions.checkState(ch.eventLoop().inEventLoop());
```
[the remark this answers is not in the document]
zpencer: Changed to assert similar to other usages

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
zpencer: Added comment

zpencer:
@ejona86 PTAL

core/src/main/java/io/grpc/internal/Channelz.java
```diff
@@ -449,22 +455,125 @@ public ChannelStats build() {
...
+
   public static final class Security {
-    // TODO(zpencer): fill this in
+    @Nullable
+    public final Tls tls;
+    @Nullable
+    public final OtherSecurity other;
+
+    public Security(Tls tls) {
+      this.tls = tls;
```
ejona86: Is there an intention for this to be non-null? Ditto for `other`.

core/src/main/java/io/grpc/internal/Channelz.java
```diff
@@ -449,22 +455,125 @@ public ChannelStats build() {
     }
   }
 
+  public interface SecurityInfoProvider {
```
ejona86: So to make sure I understand, `SecurityInfoProvider` isn't strictly necessary here, right? It isn't used with Channelz directly. It's mainly for convenience in Netty's negotiation. Everywhere else (including `getSecurityInfoProvider()`) seems it could handle `Security` directly.

I sort of expect this was done because of a comment of mine concerned about placing it in `Attributes`, where it is something like "always required, even if not used." That was more about matching the producer and consumer and it being clear when setting the Attribute is necessary (vs setting it all the time just imagining it may be used by some code, somewhere). Using explicit plumbing makes it easier to comprehend the data flow and knowing "when it is necessary" and the constraints around the value. There is benefit in explicit plumbing making it easier to communicate an internally-convenient object while also avoiding polluting the API with things like Providers and being able to tweak each piece separately. `Attributes` is convenient, but I really don't want to use it except when necessary, like communicating transport-specific information through io.grpc's API.

core/src/main/java/io/grpc/internal/Channelz.java
zpencer: Yes, exactly one of the two should be non null. Added precondition.

core/src/main/java/io/grpc/internal/Channelz.java
zpencer: Yes it is possible to just pass the `Security` around, but the reason this exists is to lazy-create the Security object only if it is needed. The toString() of a certificate looks like this, which felt a bit expensive to me:
```
[
[
  Version: V3
  Subject: CN=testclient, O=Internet Widgits Pty Ltd, ST=Some-State, C=AU
  Signature Algorithm: SHA256withRSA, OID = 1.2.840.113549.1.1.11

  Key:  Sun RSA public key, 1024 bits
  modulus: 165956023044986254443485235100948203493326685384458545615234492728849631380108078637198717973188982046560886128784222562463584010868957932010637062095083604661206388470176657085122939473572534129256975030377378114543169035021846486089803529987006058984879439854491857390215123100891877094523673506332331716589
  public exponent: 65537
  Validity: [From: Mon Nov 09 17:09:58 PST 2015,
               To: Thu Nov 06 17:09:58 PST 2025]
  Issuer: CN=testca, O=Internet Widgits Pty Ltd, ST=Some-State, C=AU
  SerialNumber: [    0a]

Certificate Extensions: 4
[1]: ObjectId: 2.5.29.35 Criticality=false
AuthorityKeyIdentifier [
[CN=testca, O=Internet Widgits Pty Ltd, ST=Some-State, C=AU]
SerialNumber: [    91c6191e 1d1a2a07]
]

[2]: ObjectId: 2.5.29.19 Criticality=false
BasicConstraints:[
  CA:false
  PathLen: undefined
]

[3]: ObjectId: 2.5.29.15 Criticality=false
KeyUsage [
  DigitalSignature
  Non_repudiation
  Key_Encipherment
]

[4]: ObjectId: 2.5.29.14 Criticality=false
SubjectKeyIdentifier [
KeyIdentifier [
0000: 00 75 B5 B9 56 69 7F 0A   76 30 A9 D3 F7 98 E1 C3  .u..Vi..v0......
0010: 01 15 3E F3                                        ..>.
]
]

Unparseable certificate extensions: 2
[1]: ObjectId: 2.5.29.17 Criticality=false
Unparseable SubjectAlternativeName extension due to
java.io.IOException: No data available in passed DER encoded value.

0000: 30 00                                              0.

[2]: ObjectId: 2.5.29.18 Criticality=false
Unparseable IssuerAlternativeName extension due to
java.io.IOException: No data available in passed DER encoded value.

0000: 30 00                                              0.

]
  Algorithm: [SHA256withRSA]
  Signature:
0000: 7F 83 0C 93 EB 1D CD DE   F6 D0 37 EB 43 43 C5 DA  ..........7.CC..
0010: 00 DA B9 1D CC F6 E0 6E   6E 8C C3 B8 CB 85 EA E7  .......nn.......
0020: C0 B9 00 49 48 A1 90 10   68 9B 5D B5 1E 96 BC 07  ...IH...h.].....
0030: 9A 7A A5 52 34 F2 7E 89   55 73 11 01 55 A1 78 54  .z.R4...Us..U.xT
0040: 51 E6 9F BB BA BC FE D2   D4 D7 B1 20 37 4D 50 77  Q.......... 7MPw
0050: 14 2E AC 48 6F B4 F3 97   C2 58 AB C3 8E 1C 3D AB  ...Ho....X....=.
0060: 43 2E 07 B3 C0 57 69 52   33 C3 86 1C 68 21 80 31  C....WiR3...h!.1
0070: 2D 22 0D 7D 39 D9 2C BB   7B 62 61 3F C3 68 C6 FA  -"..9.,..ba?.h..

]```

File: netty/src/main/java/io/grpc/netty/GrpcHttp2ConnectionHandler.java [hunk 10]
```diff
@@ -43,15 +44,28 @@ public GrpcHttp2ConnectionHandler(
...
 
+  /**
+   * Same as {@link #handleProtocolNegotiationCompleted(Attributes, Channelz.SecurityInfoProvider)}
+   * but with no Channelz.SecurityInfoProvider
+   *
+   * @deprecated Use the two argument method instead.
+   */
+  @Deprecated
+  public void handleProtocolNegotiationCompleted(Attributes attrs) {
+    handleProtocolNegotiationCompleted(attrs, null);
```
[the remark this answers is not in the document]
zpencer: This is wrong, it should be NO_SECURITY_INFO

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -905,14 +917,17 @@ static Status toGrpcStatus(ErrorCode code) {
...
+      SocketAddress remote = socket == null ? null :  socket.getRemoteSocketAddress();
+      TransportStats stats = transportTracer.getStats();
       ret.set(new SocketStats(
-          transportTracer.getStats(),
-          socket.getLocalSocketAddress(),
-          socket.getRemoteSocketAddress(),
+          stats,
+          local,
+          remote,
           Utils.getSocketOptions(socket),
```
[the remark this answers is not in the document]
zpencer: Fixed

core/src/main/java/io/grpc/internal/Channelz.java
zpencer: Yes, exactly one should be non null. Added precondition.

services/src/main/java/io/grpc/services/ChannelzProtoUtil.java
```diff
@@ -127,11 +131,42 @@ static ServerData toServerData(ServerStats stats) {
...
+    Preconditions.checkNotNull(security);
+    Preconditions.checkState(
+        security.tls != null ^ security.other != null,
+        "one of tls or othersecurity must be non null");
+    if (security.tls != null) {
+      Tls.Builder tlsBuilder
+          = Tls.newBuilder().setStandardName(security.tls.cipherSuiteStandardName);
+      if (security.tls.localCert != null) {
+        tlsBuilder.setLocalCertificate(ByteString.copyFrom(
+            security.tls.localCert.toString(),
```
ejona86: I'm surprised that this is an arbitrary textual representation, especially since `local_certificate` is `bytes`. Is this not supposed to be the binary form (DER)? That is available via `getEncoded()`. I would agree the .proto isn't all that specific and should probably be clarified (and that the gRFC is silent on the issue).

alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
```diff
@@ -119,7 +122,8 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
                     .set(TSI_PEER_KEY, altsEvt.peer())
                     .set(ALTS_CONTEXT_KEY, altsContext)
                     .set(Grpc.TRANSPORT_ATTR_REMOTE_ADDR, ctx.channel().remoteAddress())
-                    .build());
+                    .build(),
+                new Security(new OtherSecurity("alts", Any.pack(altsContext.context))));
```
ejona86: FYI: `Any.pack()` eagerly serializes the content. I'm not too concerned; AltsContext is small and this is only done after we've done after a handshake.

services/src/main/java/io/grpc/services/ChannelzProtoUtil.java
zpencer: I'll just use the `getEncoded` for now which returns ~600 bytes. toString may be a useful representation but we can add that later if needed.

alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
zpencer: Right, AltsContext is smaller than the TLS info.

PR #4301, opened 2018-04-03
netty: fix getListenSockets race

zpencer:
Fix race by avoiding async handler. This was leading to a flakey test:
channelzListenSocket

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServer.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServer.java
@@ -92,8 +92,8 @@ class NettyServer implements InternalServer, WithLogId {
   private final List<ServerStreamTracer.Factory> streamTracerFactories;
   private final TransportTracer.Factory transportTracerFactory;
   private final Channelz channelz;
-  // Only set once during start(). This code assumes all listen sockets are created at startup
-  // and never changed. In the future we may have >1 listen socket. Unset at shutdown.
+  // Only modified in event loop but safe to read any time. Set at startup and unset at shutdown.
+  // In the future we may have >1 listen socket.
   private volatile ImmutableList<Instrumented<SocketStats>> listenSockets = ImmutableList.of();
 
   NettyServer(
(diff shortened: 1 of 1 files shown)


netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -240,16 +240,12 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
-      public void operationComplete(ChannelFuture f) throws Exception {
-        Instrumented<SocketStats> listenSocket = new ListenSocket(f.channel());
-        listenSockets = ImmutableList.of(listenSocket);
-        channelz.addListenSocket(listenSocket);
-      }
-    });
+    ChannelFuture future = b.bind(address);
     try {
       future.await();
+      Instrumented<SocketStats> listenSocket = new ListenSocket(future.channel());
```
carl-mastrangelo: you need to check for success first.  future.channel may be null.  See a few lines down.

zpencer: Updated

netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -258,6 +251,9 @@ public void operationComplete(ChannelFuture f) throws Exception {
       throw new IOException("Failed to bind", future.cause());
     }
     channel = future.channel();
+    Instrumented<SocketStats> listenSocket = new ListenSocket(channel);
+    listenSockets = ImmutableList.of(listenSocket);
```
carl-mastrangelo: This is still racy, because `listenSockets` is still accessed on event loop threads, but assigned from an application thread.   You would need to make this volatile to not trigger the race detector.   

Can you show an example of what race you are trying to solve?

zpencer:
@carl-mastrangelo PTAL

netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -258,6 +251,19 @@ public void operationComplete(ChannelFuture f) throws Exception {
...
+      public void run() {
+        Instrumented<SocketStats> listenSocket = new ListenSocket(channel);
+        listenSockets = ImmutableList.of(listenSocket);
+        channelz.addListenSocket(listenSocket);
+      }
+    });
+    try {
+      channelzFuture.await();
+    } catch (InterruptedException ex) {
+      throw new RuntimeException("Interrupted while registering listen socket to channelz");
```
carl-mastrangelo: please chain the exception

PR #4303, opened 2018-04-04
core,services: move binary log logic to services

zpencer:
- some fields previously exposed as public for testing are now
  package private
- remove tests from ManagedChannelImplTest and ServerImplTest, they
  are not really needed since BinaryLogProviderImplTest already do
  very similar things
- No more static instance of BinaryLog.Factory, it will be created by
  BinaryLogProviderImpl's constructor

### DIFF
--- a/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
+++ b/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
@@ -16,49 +16,18 @@
 
 package io.grpc.internal;
 
-import com.google.common.annotations.VisibleForTesting;
-import io.grpc.CallOptions;
 import io.grpc.Channel;
 import io.grpc.ClientCall;
 import io.grpc.ClientInterceptor;
-import io.grpc.ClientInterceptors;
-import io.grpc.Context;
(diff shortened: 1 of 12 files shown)


zpencer:
@ejona86 PTAL

zpencer:
More details on what changed:
- BinaryLogInterceptor: wrapMethodDefinition always was doing the
  heavy lifting, and getServerInterceptor was used as a way to provide
  an interceptor that gets plugged in. This is changed to just
  exposing wrapMethodDefinition and hiding the interceptor itself as
  an implementation detail. Analogous change for wrapChannel and
  getClientInterceptor
- BinaryLog.java this class had always implemented client and server
  interceptors. The factory interface is changed to return server and
  client interceptors independently to make unit tests easier to
  implement in this PR. Otherwise we would need to make BinaryLog an
  interface or change the unit test code a lot to mock a
  BinaryLog.class.
- BinaryLog had a static{} initializer that is removed. It is easier
  to do the initialization of BinaryLog when BinaryLogImpl is created,
  rather than in static code.

zpencer:
This will be split into other PRs

PR #4307, opened 2018-04-05
examples: upgrade protobuf-maven-plugin to 0.5.1

ejona86:
This was missed in 2094bb4d8

CC @jorgheymans, oops, we missed this in review.


jorgheymans:
Right my bad, my grep did not pick this one up for some reason. Thanks!

Op do 5 apr. 2018 17:37 schreef Eric Anderson <notifications@github.com>:

Issue #2195, opened 2016-08-24
gRPC code generator needs to handle reserved words in method names


carl-mastrangelo:
When defining an RPC like:

```
service Foo {
  rpc Import(ImportRequest) returns (ImportResponse) {};
}
```

The method Import gets translated to a java keyword.    When should adjust it by adding an underscore (prepended?)  to allow it to compile.

ejona86:
We should probably mimic protobuf (e.g., prepend if it prepends)

ejona86:
And methods for fields have `get` or `set` prefix, so they're also fine. Messages virtually always start with an upper-case character and keywords are all lowercase. And enums are ALL_UPPER_CASE.

PR #4309, opened 2018-04-05
compiler: allow capitalized java keywords as rpc names

zpencer:
previously `Import` fails because it gets turned into a method named
`import`. This PR makes the method be named `import_`, everything else
stays the same, such as the name in the method descriptor.

Note: `iMport` or `iMPORT` or any variation of capitalizations that is
not `Import` always worked. `Import` gets translated into `import`
because of our attempts to camel case the names.

fixes https://github.com/grpc/grpc-java/issues/2195

### DIFF
--- a/compiler/src/java_plugin/cpp/java_generator.cpp
+++ b/compiler/src/java_plugin/cpp/java_generator.cpp
@@ -4,6 +4,7 @@
 #include <iostream>
 #include <iterator>
 #include <map>
+#include <set>
 #include <vector>
 #include <google/protobuf/compiler/java/java_names.h>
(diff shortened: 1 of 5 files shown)


compiler/src/java_plugin/cpp/java_generator.cpp
```diff
@@ -48,7 +107,10 @@ static string MixedLower(const string& word) {
       after_underscore = false;
     }
   }
-  return w;
+  if (java_keywords.find(w) == java_keywords.end()) {
```
ejona86: Could you invert this condition so the final `return` is the common/successful case?

PR #4310, opened 2018-04-06
core: replumb RetryPolicy

dapengzhang0:
- replumbed `RetryPolicy` with `MethodInfo` without breaking the existing `RetryPolicyTest`.
- moved `ServiceConfigInterceptor.MethodInfo.RetryPolicy` out as a top level class so that `RetriableStream` does not import `ServiceConfigInterceptor`.


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -539,6 +533,8 @@ ClientStream newSubstream(ClientStreamTracer.Factory tracerFactory, Metadata new
     this.backoffPolicyProvider = backoffPolicyProvider;
     this.transportFactory =
         new CallCredentialsApplyingTransportFactory(clientTransportFactory, this.executor);
+    this.retryEnabled = !builder.retryDisabled;
```
ericgribkoff: Can we change the builder to store `retryEnabled` to avoid this negation?

core/src/main/java/io/grpc/internal/RetryPolicy.java
```diff
@@ -0,0 +1,107 @@
...
+        .add("maxAttempts", maxAttempts)
+        .add("initialBackoffNanos", initialBackoffNanos)
+        .add("maxBackoffNanos", maxBackoffNanos)
+        .add("backoffMultiplier", backoffMultiplier)
+        .add("retryableStatusCodes", retryableStatusCodes)
+        .toString();
+  }
+
+  /**
+   * Providers the most suitable retry policy for a call in the current situation.
```
ericgribkoff: Without looking at DelayedRetryPolicyProvider, this description is hard to understand. Might be helpful to say when this will have to provide a retry policy, as that defines what the "current situation" will be

core/src/main/java/io/grpc/internal/ServiceConfigInterceptor.java
```diff
@@ -104,6 +115,7 @@ void handleUpdate(Map<String, Object> serviceConfig) {
     // Okay, service config is good, swap it.
     serviceMethodMap.set(Collections.unmodifiableMap(newServiceMethodConfigs));
     serviceMap.set(Collections.unmodifiableMap(newServiceConfigs));
+    nameResolveComplete = true;
```
ericgribkoff: Is synchronization necessary for this?

core/src/main/java/io/grpc/internal/ServiceConfigInterceptor.java
```diff
@@ -316,10 +300,33 @@ public String toString() {
...
     }
-    if (info.retryPolicy != null) {
-      callOptions = callOptions.withOption(RETRY_POLICY_KEY, info.retryPolicy);
-    }
 
     return next.newCall(method, callOptions);
   }
+
+  @CheckForNull
+  private MethodInfo methodInfoOf(MethodDescriptor<?, ?> method) {
```
ericgribkoff: optional: would getMethodInfo be a better method name?

core/src/main/java/io/grpc/internal/ServiceConfigInterceptor.java
dapengzhang0: A note: If the stream listener is closed because of a status in a trailer from the server, `nameResolveComplete` will be observed to be true even not volatile.

Volatile is needed only for the case when the stream listener is closed for other reasons.

core/src/test/java/io/grpc/internal/ServiceConfigInterceptorTest.java
```diff
@@ -98,6 +100,16 @@ public void withWaitForReady() {
     assertThat(callOptionsCap.getValue().isWaitForReady()).isTrue();
   }
 
+  @Test
+  public void notYetHandleUpdate() {
+    interceptor.interceptCall(methodDescriptor, CallOptions.DEFAULT.withoutWaitForReady(), channel);
+
+    verify(channel).newCall(eq(methodDescriptor), callOptionsCap.capture());
+    assertThat(callOptionsCap.getValue().isWaitForReady()).isFalse();
```
ericgribkoff: Why verify wait for ready here?

core/src/test/java/io/grpc/internal/ServiceConfigInterceptorTest.java
```diff
@@ -98,6 +100,16 @@ public void withWaitForReady() {
     assertThat(callOptionsCap.getValue().isWaitForReady()).isTrue();
   }
 
+  @Test
+  public void notYetHandleUpdate() {
```
ericgribkoff: This test name isn't clear. I think it's intended to test that the default retry policy is returned if `interceptor.handleUpdate` hasn't been called, but there's no accompanying test for when `handleUpdate` has been called - this functionality seems covered in the `RetryPolicyTest` cases, but if it's going to be tested here as well (which makes sense) we should also cover the case when it is actually updated.

core/src/test/java/io/grpc/internal/ServiceConfigInterceptorTest.java
```diff
@@ -351,9 +363,9 @@ public void handleUpdate_matchNames() {
     assertThat(interceptor.serviceMethodMap.get())
         .containsExactly(
             methodDescriptor.getFullMethodName(),
-            new MethodInfo(methodConfig));
-    assertThat(interceptor.serviceMap.get())
-        .containsExactly("service2", new MethodInfo(methodConfig));
+            new MethodInfo(methodConfig, false, 1));
```
ericgribkoff: The `MethodInfo.retryPolicy` does not have `maxRetryAttempts` equal to 1. This makes these lines kind of confusing, especially since `MethodInfo.equals` ignores the `retryPolicy` altogether.

Should `MethodInfo.equals` compare retry policy as well?

Either way, it might be clearer to note that 1 is ignored since retryEnabled=false

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -399,7 +399,7 @@ public ManagedChannel build() {
     List<ClientInterceptor> effectiveInterceptors =
         new ArrayList<ClientInterceptor>(this.interceptors);
     if (statsEnabled) {
-      retryDisabled = true;
+      retryEnabled = false;
```
carl-mastrangelo: why does a method called get modify the builder?

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -641,21 +644,22 @@ private RetryPlan makeRetryDecision(RetryPolicy retryPolicy, Status status, Meta
...
+            nextBackoffIntervalNanos = Math.min(
+                (long) (nextBackoffIntervalNanos * retryPolicy.backoffMultiplier),
+                retryPolicy.maxBackoffNanos);
+
           } // else no retry
         } else if (pushback >= 0) {
           shouldRetry = true;
-          backoffInMillis = pushback;
-          nextBackoffIntervalInSeconds = retryPolicy.initialBackoffInSeconds;
+          backoffNanos = TimeUnit.MILLISECONDS.toNanos(pushback);
```
carl-mastrangelo: optional: rename pushback to pushbackMillis.

core/src/main/java/io/grpc/internal/RetryPolicy.java
```diff
@@ -0,0 +1,108 @@
...
+      int maxAttempts,
+      long initialBackoffNanos,
+      long maxBackoffNanos,
+      double backoffMultiplier,
+      @Nonnull Set<Code> retryableStatusCodes) {
+    this.maxAttempts = maxAttempts;
+    this.initialBackoffNanos = initialBackoffNanos;
+    this.maxBackoffNanos = maxBackoffNanos;
+    this.backoffMultiplier = backoffMultiplier;
+    this.retryableStatusCodes = retryableStatusCodes;
```
carl-mastrangelo: this needs to be a defensive copy

core/src/test/java/io/grpc/internal/ServiceConfigInterceptorTest.java
```diff
@@ -58,7 +59,8 @@ public void setUp() {
     MockitoAnnotations.initMocks(this);
   }
 
-  private final ServiceConfigInterceptor interceptor = new ServiceConfigInterceptor();
+  private final ServiceConfigInterceptor interceptor = new ServiceConfigInterceptor(
+      true /* retryEnabled */, 5 /* maxRetryAttemptsLimit */);
```
carl-mastrangelo: nit: these comments need to have a `=` after the variable name, or the linter will complain

PR #4311, opened 2018-04-06
netty: prepare NettySocketSupport for internal import

zpencer:
Fix lint warning. This lets the PR be imported cleanly
with an internal change.


netty/src/main/java/io/grpc/netty/NettySocketSupport.java
```diff
@@ -20,14 +20,15 @@
...
-import java.util.HashMap;
 import java.util.Map;
+import java.util.logging.Logger;
 import javax.annotation.Nullable;
 
 /**
  * An class for getting low level socket info.
  */
 final class NettySocketSupport {
+  private static final Logger log = Logger.getLogger(NettySocketSupport.class.getName());
```
carl-mastrangelo: ???

zpencer: This will be used in the internal version, I suppose it's ok for it not to be present at all in the common version. I was trying to minimize the number of text replacements in this file.

netty/src/main/java/io/grpc/netty/NettySocketSupport.java
```diff
@@ -20,30 +20,35 @@
...
 /**
  * An class for getting low level socket info.
  */
-final class NettySocketSupport {
+abstract class NettySocketSupport {
+  private static final Logger log = Logger.getLogger(NettySocketSupport.class.getName());
+  private static final NettySocketSupport INSTANCE;
+
+  static {
+    INSTANCE = create();
```
carl-mastrangelo: private static final NettySocketSupport INSTANCE = create();


?

netty/src/main/java/io/grpc/netty/NettySocketSupport.java
```diff
@@ -52,9 +57,27 @@ public NativeSocketOptions(
         Map<String, String> otherInfo) {
       Preconditions.checkNotNull(otherInfo);
       this.tcpInfo = tcpInfo;
-      this.otherInfo = ImmutableMap.copyOf(new HashMap<String, String>(otherInfo));
+      this.otherInfo = ImmutableMap.copyOf(otherInfo);
     }
   }
 
-  private NettySocketSupport() {}
+  public static NettySocketSupport instance() {
```
carl-mastrangelo: this will be wrong for subclasses right?

zpencer: Restructured to avoid ambiguity

zpencer:
@carl-mastrangelo any more issues for this one?

zpencer:
We're already doing this today with https://github.com/grpc/grpc-java/blob/master/context/src/main/java/io/grpc/Context.java#L135 and it doesn't seem too bad. I'm open to alternatives.

PR #4314, opened 2018-04-06
Run Android interop instrumentation tests on Firebase

ericgribkoff:
Adds a new Kokoro buildscript to build and test the android-interop-testing app.

Internal change is cl/201371950

### DIFF
--- a/buildscripts/kokoro/android-interop.sh
+++ b/buildscripts/kokoro/android-interop.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+set -exu -o pipefail
+if [[ -f /VERSION ]]; then
+  cat /VERSION
+fi
+
+# Install gRPC and codegen for the Android interop app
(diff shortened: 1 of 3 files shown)


ericgribkoff:
@zpencer ping

buildscripts/kokoro/android-interop.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+set -exu -o pipefail
+cat /VERSION
```
zpencer: https://github.com/ericgribkoff/grpc-java/blob/9b9c6fc5fbd998a6e4073902d5524a4a634491c0/buildscripts/kokoro/release_artifacts.sh#L4

Something like this will make the script easier to run on our workstations, because /VERSION only exists in kokoro and the flags on line 3 make the script exit on any command failure.

ericgribkoff: Done

PR #4316, opened 2018-04-06
core, netty: allow InputStream based certs

zpencer:
Allow ServerBuilder to read certs from InputStream, not just from a
File.

This PR revives https://github.com/grpc/grpc-java/issues/1881


ejona86:
/me runs over to compare to a branch I was working on https://github.com/ejona86/grpc-java/commits/security-inputstream, which I got pulled in via https://github.com/grpc/grpc-java/issues/1791#issuecomment-347358287 :smile:

This seems fine, and appears to be a subset of what I was working on, since I was more focused on cleaning up the current references. I want to look at the TlsTest a little bit.

ejona86:
BTW: I also tracked the InputStreams through SslContextBuilder.forServer() (and similar) and verified that the methods take on the responsibility of closing the InputStream.

netty/src/test/java/io/grpc/netty/TlsTest.java
```diff
@@ -176,6 +178,70 @@ public void basicClientServerIntegrationTest() throws Exception {
     client.unaryRpc(SimpleRequest.getDefaultInstance());
   }
 
+  /**
+   * Same as basicClientServerIntegrationTest except we test the InputStream based certs API.
+   */
+  @Test
+  public void basicClientServerIntegrationTestInputStreamCerts() throws Exception {
+    InputStream serverCertChain = null;
```
ejona86: None of these InputStreams are passed to our code. This test is not adding any value.

zpencer: These assigned to on lines 193/194 , 218/219 and used immediately after.

zpencer: I see, I will remove the test.

PR #4325, opened 2018-04-09
Bazel: Fix compilation in Java 9

davido:
Fixes: #3633.

Test Plan:

On most recent Bazel version run:

```
  $ bazel --host_javabase=/usr/lib64/jvm/java-9-openjdk \
    build --javacopt='--release 9' \
    --java_toolchain=@bazel_tools//tools/jdk:toolchain_jdk9 \
    examples:helloworld_java_grpc
```

### DIFF
--- a/core/BUILD.bazel
+++ b/core/BUILD.bazel
@@ -16,17 +16,6 @@ java_library(
     ],
 )
 
-# javax.annotation.Generated is not included in the default root modules in 9,
-# see: http://openjdk.java.net/jeps/320.
-java_import(
-    name = "javax_annotation",
-    jars = [
-        "@javax_annotation_api//file",
-    ],
-    neverlink = 1,  # @Generated is source-retention
-    visibility = ["//visibility:public"],
-)
-
 java_library(
     name = "inprocess",
     srcs = glob([
--- a/core/BUILD.bazel
+++ b/core/BUILD.bazel
@@ -16,6 +16,17 @@ java_library(
     ],
 )
 
+# javax.annotation.Generated is not included in the default root modules in 9,
+# see: http://openjdk.java.net/jeps/320.
+java_import(
+    name = "javax_annotation",
+    jars = [
+        "@javax_annotation_api//file",
+    ],
+    neverlink = 1,  # @Generated is source-retention
+    visibility = ["//visibility:public"],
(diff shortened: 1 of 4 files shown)


repositories.bzl
```diff
@@ -277,6 +280,15 @@ def io_opencensus_grpc_metrics():
       sha1 = "d57b877f1a28a613452d45e35c7faae5af585258",
   )
 
+def javax_annotation():
+  # TODO(davido): maven_jar does not support neverlink attribute.
+  # To circumvent use http_file following by java_import.
+  native.http_file(
+      name = "javax_annotation_api",
+      sha256 = "e04ba5195bcd555dc95650f7cc614d151e4bcd52d29a10b8aa2197f3ab89ab9b",
+      urls = ["http://central.maven.org/maven2/javax/annotation/javax.annotation-api/1.3.2/javax.annotation-api-1.3.2.jar"],
```
ejona86: We use version 1.2 in build.gradle and like to keep the versions in sync. 1.3.2 also doesn't support Java 7 (although I honestly don't know if Bazel supports Java 7). Let's change this to version 1.2 (unless there's some problem with doing so).

core/BUILD.bazel
```diff
@@ -16,6 +16,17 @@ java_library(
     ],
 )
 
+# javax.annotation.Generated is not included in the default root modules in 9,
+# see: http://openjdk.java.net/jeps/320.
+java_import(
+    name = "javax_annotation",
```
ejona86: Technically, having this in core seems a bad idea, but we don't have a third_party/ yet, so this seems good enough for now.

ejona86:
Thank you for this!

core/BUILD.bazel
davido: Ack.

repositories.bzl
davido: Done.

core/BUILD.bazel
dapengzhang0: Maybe in `stub` is more appropriate than `core`, although not as good as `third_party`.

davido: I moved the new javax_annotation rule to stub package instead of core. Done.

PR #4335, opened 2018-04-12
netty: fix visibility issues with InternalNettySocketSupport

zpencer:
Previous version is not actually extendable from other packages.


netty/src/main/java/io/grpc/netty/InternalNettySocketSupport.java
```diff
@@ -17,14 +17,25 @@
...
 @Internal
 public final class InternalNettySocketSupport {
 
-  public interface InternalHelper extends NettySocketSupport.Helper { }
+  public interface InternalHelper extends NettySocketSupport.Helper {
+    @Override
+    InternalNativeSocketOptions getNativeSocketOptions(io.netty.channel.Channel ch);
+  }
+
+  public static class InternalNativeSocketOptions extends NettySocketSupport.NativeSocketOptions {
```
carl-mastrangelo: final

PR #4337, opened 2018-04-12
core: forward toString for forwarding classes

dapengzhang0:
Forward `toString()` method for forwarding classes to improve debug information.

For example, `ForwardingManagedChannel.toString()` will return something like:

```
ForwardingManagedChannel{delegate=ManagedChannelImpl{logId=tag-13247, target=localhost:8080}}
```


dapengzhang0:
@zpencer thanks!

PR #4342, opened 2018-04-13
stub: add docs clarifying readiness

carl-mastrangelo:


core/src/main/java/io/grpc/ClientCall.java
```diff
@@ -223,6 +227,10 @@ public void onReady() {}
    *
    * <p>This abstract class's implementation always returns {@code true}. Implementations generally
    * override the method.
+   *
+   * <p>If the type of the call is either {@link MethodDescriptor.MethodType#UNARY} or
+   * {@link MethodDescriptor.MethodType#SERVER_STREAMING}, this method may return false. Calls that
```
ejona86: s/may return/may persistently return/

carl-mastrangelo: Done.

PR #4343, opened 2018-04-14
services: use opaque pagination tokens in channelz

zpencer:
It turns out, dealing with int64 in javascript is painful. It is
also nice not to depend on proto in the js, which means the stats
hook should convert everything into JSON.

I considered letting clients pass back in the last result from
the page as is, and let the server figure out whether to +1 to
it. But this leads edge cases that are tricky to deal with.


zpencer:
Result of offline discussion: it's not so bad include one result from the previous page, and if needed, the official google protobuf library for js has acceptable int64 support.

PR #4345, opened 2018-04-16
core: add sockets to subchannel stat

zpencer:
This is info not being reported at the moment and is missing from the
display.


core/src/test/java/io/grpc/internal/InternalSubchannelTest.java
```diff
@@ -931,6 +933,19 @@ public void channelzMembership() throws Exception {
     assertFalse(channelz.containsClientSocket(t0.transport.getLogId()));
   }
 
+  @Test
+  public void channelzStatContainsTransport() throws Exception {
+    SocketAddress addr = mock(SocketAddress.class);
```
carl-mastrangelo: `... = new SocketAddress(){};`

PR #4347, opened 2018-04-16
services: add human friendly representation of tcp ip address

zpencer:
The existing form is good for machines but not so much for a display
page GUI: `"ipAddress": "AAAAAAAAAAAAAAAAAAAAAA==",`


carl-mastrangelo:
I'm afraid I don't think this is a good idea.   The correct way to display this is up to the agent, not the server.  There are numerous libraries available for working with binary data in JS.   I don't see the JSON conversion semantics as being something that need to be worked around.

zpencer:
I guess some minimal data processing is unavoidable anyways.

PR #4349, opened 2018-04-16
core: refactor ManagedChannelImplTest with a configurable builder

dapengzhang0:
In current code of `ManagedChannelImplTest`, new tests will have to add more and more custom args in `createChannel(...)`, and each time when a custom arg is needed all other default args need be repeatedly provided in `createChannel(...)`. In most cases those args are just builder attributes. So this PR makes ManagedChannelImplTest easier to stub by providing a builder field, then new tests will just set necessary builder attributes before calling `createChannel()`.


dapengzhang0:
@carl-mastrangelo thanks!

PR #4354, opened 2018-04-18
all: check channelz diffs

carl-mastrangelo:
c


carl-mastrangelo:
The most likely thing to fail should be first on the list no?  I think channelz will not be modified so often.

dapengzhang0:
One remark: when we made a grpc/grpc-java release tag, the link to the remote proto must be changed from grpc/grpc-proto:master to a fixed commit or tag in grpc/grpc-proto. Otherwise the release tag will fail to build as grpc/grpc-proto:master evolves.

PR #4362, opened 2018-04-19
core: clarify binlog API

zpencer:
The interceptors are a part of the implementation and not meant for
callers.

### DIFF
--- a/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
+++ b/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
@@ -88,7 +88,7 @@ public abstract class BinaryLogProvider implements Closeable {
   /**
    * Wraps a channel to provide binary logging on {@link ClientCall}s as needed.
    */
-  public Channel wrapChannel(Channel channel) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -100,7 +100,7 @@ Channel wrapChannel(Channel channel) {
   /**
    * Wraps a {@link ServerMethodDefinition} such that it performs binary logging if needed.
    */
-  final <ReqT, RespT> ServerMethodDefinition<?, ?> wrapMethodDefinition(
+  public <ReqT, RespT> ServerMethodDefinition<?, ?> wrapMethodDefinition(
```
ejona86: Why remove the `final`? I'd be happy adding it to the client version. But I guess it doesn't matter at all since this will change soon anyway.

zpencer: The `final` is probably a good guard to prevent this from becoming a super interceptor API. I'll keep it until there's a good reason to remove it.

PR #4363, opened 2018-04-19
services: avoid static initialization for BinaryLog.java

zpencer:
The code that uses this will create an instance, namely the actual Impl of the BinaryLogProvider.


ejona86:
It's strange to see the usage of GRPC_BINARY_LOG_CONFIG just disappear.

PR #4364, opened 2018-04-20
core: also expose parser on JsonParser

carl-mastrangelo:
One more time...


core/src/main/java/io/grpc/internal/JsonParser.java
```diff
@@ -39,6 +39,10 @@
 
   private JsonParser() {}
 
+  /**
+   * Parses a json string, returning either a {@code Map<String, Object>}, {@code List<Object>},
+   * {@code String}, {@code Double}, {@code Boolean}, or {@code Void}.
```
ejona86: s/{@code Void}/the value {@code null}/ ?

carl-mastrangelo: Done.

PR #4365, opened 2018-04-20
services,core: simplify CallId generation

zpencer:
A lot of LOCs here are whitespace changes or removals.

Function changes after vs before:

After
=====

BinaryLog.java is the class that is responsible for intercepting
client and server calls. It now requires a CallId to be passed
in. The BinaryLogProvider is responsible for generating a
CallId.

CensusBinaryLogProvider will generate a CallId based on census
info. For the server CallId, my intention is to read the census
info from the Context. For the client CallId, the census span is
not accessible, so a pre-computed CallId is stored into the
CallOptions.

The default Impl is just an AtomicLong.

Before
======

The idea was to put a pre-computed CallId into the Context for
the server context, and a pre-computed CallId into the
CallOptions. But as we discussed there's some issues with how it
was trying to access the census info.

### DIFF
--- a/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
+++ b/core/src/main/java/io/grpc/internal/BinaryLogProvider.java
@@ -125,7 +125,8 @@ public abstract class BinaryLogProvider implements Closeable {
    */
   // TODO(zpencer): ensure the interceptor properly handles retries and hedging
   @Nullable
-  protected abstract ClientInterceptor getClientInterceptor(String fullMethodName);
+  protected abstract ClientInterceptor getClientInterceptor(
+      String fullMethodName, CallOptions callOptions);
 
   @Override
   public void close() throws IOException {
@@ -184,7 +185,8 @@ public abstract class BinaryLogProvider implements Closeable {
         MethodDescriptor<ReqT, RespT> method,
         CallOptions callOptions,
         Channel next) {
-      ClientInterceptor binlogInterceptor = getClientInterceptor(method.getFullMethodName());
+      ClientInterceptor binlogInterceptor = getClientInterceptor(
+          method.getFullMethodName(), callOptions);
       if (binlogInterceptor == null) {
         return next.newCall(method, callOptions);
       } else {
@@ -219,11 +221,11 @@ public abstract class BinaryLogProvider implements Closeable {
     }
   }
 
-  public CallId getServerCallId() {
+  protected CallId getServerCallId() {
     return new CallId(0, counter.getAndIncrement());
   }
 
-  public CallId getClientCallId(CallOptions options) {
+  protected CallId getClientCallId(CallOptions options) {
     return new CallId(0, counter.getAndIncrement());
   }
 }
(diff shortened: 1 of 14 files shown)


zpencer:
@ejona86 PTAL

zpencer:
@ejona86 friendly ping

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -280,8 +214,16 @@ public CallId(long hi, long lo) {
       this.lo = lo;
     }
 
-    static CallId fromCensusSpan(Span span) {
+    public static CallId fromCensusSpan(Span span) {
       return new CallId(0, ByteBuffer.wrap(span.getContext().getSpanId().getBytes()).getLong());
     }
   }
+
+  public CallId getServerCallId() {
```
ejona86: `protected`? Ditto below.

ejona86: Where are these methods called?

services/src/main/java/io/grpc/services/CensusBinaryLogProvider.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package io.grpc.services;
+
+import io.grpc.CallOptions;
+import io.grpc.internal.BinaryLogProvider;
+import io.opencensus.trace.Span;
+import io.opencensus.trace.Tracing;
+
+public class CensusBinaryLogProvider extends BinaryLogProviderImpl {
```
ejona86: `@Internal` or move it to internal package?

`final`?

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
zpencer: locked down visbility.

services/src/main/java/io/grpc/services/CensusBinaryLogProvider.java
zpencer: moved to internal and made final.

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
zpencer: I was probably overzealous in making this PR focused on only changing the generation. I plugged it into BinaryLogProviderImpl now.

zpencer:
@ejona86 PTAL at updated

core/src/main/java/io/grpc/internal/BinaryLogProvider.java
```diff
@@ -280,8 +216,16 @@ public CallId(long hi, long lo) {
       this.lo = lo;
     }
 
-    static CallId fromCensusSpan(Span span) {
+    public static CallId fromCensusSpan(Span span) {
       return new CallId(0, ByteBuffer.wrap(span.getContext().getSpanId().getBytes()).getLong());
     }
   }
+
+  protected CallId getServerCallId() {
```
ejona86: Why are these methods defined here and not `BinaryLogProviderImpl`?

zpencer:
Trash comment to try and unwedge cla bot.

Issue #4367, opened 2018-04-20
AbstractManagedChannelImplBuilder#executorPool set, but never used


jpitz:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?
1.11.0

### What did you expect to see?
I specified a custom executor, but noticed no changes on the client. After starting to investigated, I noticed that the executor that is set with executor() is never used in the AbstractManagedChannelImplBuilder#build method. Instead, it grabs the shared pool and uses that instead.

Snippet from source:
```
  @Override
  public ManagedChannel build() {
    return new ManagedChannelOrphanWrapper(new ManagedChannelImpl(
        this,
        buildTransportFactory(),
        // TODO(carl-mastrangelo): Allow clients to pass this in
        new ExponentialBackoffPolicy.Provider(),
        SharedResourcePool.forResource(GrpcUtil.SHARED_CHANNEL_EXECUTOR),
        GrpcUtil.STOPWATCH_SUPPLIER,
        getEffectiveInterceptors(),
        CallTracer.getDefaultFactory()));
  }
```

ejona86:
This was discussed some in #4368, but the short of it is the executor you pass to the builder is used: https://github.com/grpc/grpc-java/blob/v1.11.0/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java#L528

This usage of SHARED_CHANNEL_EXECUTOR is on purpose, although we have slowly been trying to figure out ways to avoid the hard-coded executors while also not exposing too much of our internals.

PR #4368, opened 2018-04-20
gh-4367: actually use of provided executor pool

jpitz:
Issue: https://github.com/grpc/grpc-java/issues/4367

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
@@ -390,7 +390,7 @@ public abstract class AbstractManagedChannelImplBuilder
         buildTransportFactory(),
         // TODO(carl-mastrangelo): Allow clients to pass this in
         new ExponentialBackoffPolicy.Provider(),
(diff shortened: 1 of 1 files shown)


thelinuxfoundation:

_Regards,
CLA GitHub bot_

carl-mastrangelo:
@jpitz The executor you modified is slightly different.  The one provided by the application to the ManagedChannelBuilder is for running RPC callbacks.  The channel executor is for running channel related operations that generally do not block for very long.  They have different purposes.

jpitz:
hmm... Can you point to where this is being used in code? I was trying to track it down, but couldn't find it.

jpitz:
I see so in the typical case with Netty, you don't necessarily need to worry about tuning the executor on the client, but rather only the workerEventLoopGroup.

ejona86:
@jpitz, I don't think that's the correct take-a-way. The executor you pass to the builder is used: https://github.com/grpc/grpc-java/blob/v1.11.0/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java#L528

There's a few other cases that need an executor but we didn't feel comfortable using the application's executor. That's what you found.

jpitz:
I see.

PR #4373, opened 2018-04-20
buildscripts: build linux artifacts in kokoro

zpencer:
Update unix.sh and make_dependencies.sh to support an ARCH env var.
Add linux.sh and linux.cfg to build both x86_64 and x86_32.


zpencer:
Just remembered that linux uses dockers. Doh!

PR #4374, opened 2018-04-20
core: try out flogger DO NOT SUBMIT

carl-mastrangelo:
Try out flogger.   I suspect this may need to be rolled back on import, but I wanted to see if it was feasible to use.    
If this PR looks good, I'll add the bazel deps.

@ronshapiro: This seems like a lot of dependencies to add to get this working.   Am I doing it right?


core/build.gradle
```diff
@@ -2,13 +2,14 @@ description = 'gRPC: Core'
 
 dependencies {
     compile project(':grpc-context'),
+            libraries.flogger,
+            libraries.flogger_backend,
```
carl-mastrangelo: This is a classpath dep needed to work, but we don't reference any classes from it.

ejona86:
:-/ "New DEX reference count: 13,413 (delta: 498)". I wonder what it's like without using the GoogleLogger or without a backend... 500 methods is surprisingly high. I also wonder if there's something small preventing dead code elimination.

carl-mastrangelo:
@ejona86 yeah, it looks like flogger is not ready for Android usage, though something is coming "eventually"  I don't expect it to be ready for at least another year, so I am closing this out.

carl-mastrangelo:
Reopening to see what methods are retained, since now the presubmit shows them.

carl-mastrangelo:
For posterity, here are all the extra methods brought in by flogger:
[flogger.diff.gz](https://github.com/grpc/grpc-java/files/1962565/flogger.diff.gz)

PR #4375, opened 2018-04-21
kokoro: Avoid --include-build for Android

ejona86:
Since 4369e8cd the --include-build just opens us up to trouble with
accidentally building the protoc plugin. Since we're going to do a
./gradlew install anyway, let's just wait until after that point for
building cronet.

This sort of problem was experienced while developing #4369.

CC @dapengzhang0

### DIFF
--- a/buildscripts/kokoro/android.sh
+++ b/buildscripts/kokoro/android.sh
@@ -5,13 +5,6 @@ cat /VERSION
 
 BASE_DIR="$(pwd)"
 
-# Build Cronet
-
-cd "$BASE_DIR/github/grpc-java/cronet"
-./cronet_deps.sh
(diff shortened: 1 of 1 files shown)


ericgribkoff:
This looks fine to me, but not sure what "accidentally" building protoc plugin means here or why the --include-build failed on #4369 but not e.g. https://github.com/grpc/grpc-java/pull/4376.

ejona86:
@ericgribkoff, it's because #4369 is adding a compiler dependency to core via the protobuf in tests. Even though that dependency should only be needed for the tests, the way we handle dependencies with the compiler it isn't smart enough to recognize it doesn't need the compiler dependency for the main compile.

PR #4376, opened 2018-04-21
compiler: unbreak docker file

zpencer:
The protoc-artifacts image does not provide the protoc code in
/protobuf anymore, we must download it ourselves:
https://github.com/google/protobuf/commit/5f052ae7ae8a757dd919e95e0f1481e08fef749b#diff-0bc9399720fa75bdc363a61a50cd2609

The wget workaround works to download the protobuf sources, but ./autogen.sh
still uses curl to fetch its depedencies:
https://github.com/google/protobuf/issues/4419

### DIFF
--- a/compiler/Dockerfile
+++ b/compiler/Dockerfile
@@ -1,5 +1,13 @@
 FROM protoc-artifacts:latest
 
+# Update curl so git cloning and dep fetching works
+#   https://github.com/google/protobuf/issues/4419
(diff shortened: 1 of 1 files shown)


ejona86:
No, we don't want to do it as part of the docker process. We want to do it after the docker image is built. (At the same time we build our gRPC code)

zpencer:
Closing this one. I will send out PRs that target the `buildscripts/` files instead.

PR #4379, opened 2018-04-23
alts: move alts protos to match proto package

carl-mastrangelo:


carl-mastrangelo:
fixed the bazel file too.

PR #4386, opened 2018-04-25
core: stickiness for RoundRobinLoadBalancer

dapengzhang0:


core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -245,6 +271,74 @@ private ConnectivityState getAggregatedState() {
...
+            protected boolean removeEldestEntry(Map.Entry<String,Ref<Subchannel>> eldest) {
+              return size() > MAX_ENTRIES;
+            }
+          };
+
+      final Map<Subchannel, Ref<Subchannel>> subchannelRefs =
+          new HashMap<Subchannel, Ref<Subchannel>>();
+
+      StickinessState(@Nonnull String stickinessKey) {
+        this.key = Key.of(stickinessKey, Metadata.ASCII_STRING_MARSHALLER);
```
ejona86: May want to protect ourselves from the key being `-bin`, since it could be in the future that support is added.

ejona86:
This looks good.

dapengzhang0:
@ejona86 Thanks for the review. Merged.

PR #4388, opened 2018-04-25
buildscripts: sonatype uploader ussing CONF before it is set

zpencer:
Set $CONF before we use it.
Also enable some defensive bash flags to catch errors.

### DIFF
--- a/buildscripts/sonatype-upload.sh
+++ b/buildscripts/sonatype-upload.sh
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -e
+set -eu -o pipefail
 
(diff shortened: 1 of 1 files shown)


buildscripts/sonatype-upload.sh
```diff
@@ -46,6 +46,7 @@ if [ -z "$DIR" ]; then
   exit 1
 fi
 
+CONF="$HOME/.config/sonatype-upload"
 [ -f "$CONF" ] && . "$CONF"
 
 if [ -z "$USERNAME" -o -z "$PASSWORD" ]; then
```
ejona86: These checks will break with the new set options.

PR #4389, opened 2018-04-25
buildscripts: do not echo secrets in sonatype upload

zpencer:
Making this a bit more secretive so we can use it with kokoro.


buildscripts/sonatype-upload.sh
```diff
@@ -87,7 +87,6 @@ REPOID="$(
   grep stagedRepositoryId |
   sed 's/.*<stagedRepositoryId>\(.*\)<\/stagedRepositoryId>.*/\1/'
   )"
-echo "Repository id: $REPOID"
```
ejona86: How is this a secret? It's "iogrpc-" with a predictable incrementing number on the end. Won't we want to know this anyway to issue the release?

PR #4392, opened 2018-04-25
buildscripts: add linux job that release_artifacts can dep on

zpencer:
Slight tweak so that linux.cfg can always run. release_artifacts.cfg will be a RELEASE
job that we only trigger if we want to release.


buildscripts/kokoro/linux_artifacts.cfg
```diff
@@ -0,0 +1,11 @@
+# Config file for internal CI
+
+# Location of the continuous shell script in repository.
+build_file: "grpc-java/buildscripts/kokoro/linux.sh"
```
matt-kwong: This needs to be updated to `linux_artifacts.sh`

PR #4394, opened 2018-04-26
 testing: GrpcCleanupRule

dapengzhang0:
Added `GrpcCleanupRule` based on offline discussions with @ejona86 

This will ease a lot of test scenarios that we want to automatically shut down servers and channels, with much more flexibility such as that mentioned by #3624 than `GrpcServerRule`.

**ManagedChannel/Server cleanup details:**
- If the test has already failed, call `shutdownNow()` for each of the resources registered.
- If no error happens up to now, call `shutdown()` for each of the resources registered. If any error occurs, immediately call `shutdownNow()` for each resource.
- If no error happens up to now, call `awaitTermination()` with `timeout = deadline - current time` and assert termination for each resource.  If any error occurs, immediately call `shutdownNow()` for the current resource and all the rest resources.
- once the rule evaluation is done, throw the first exception encountered if any .

### DIFF
--- a/testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
+++ b/testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
@@ -61,7 +61,7 @@ public final class GrpcCleanupRule implements TestRule {
    *
    * @return this
    */
-  public GrpcCleanupRule withTimeout(long timeout, TimeUnit timeUnit) {
+  public GrpcCleanupRule setTimeout(long timeout, TimeUnit timeUnit) {
     checkArgument(timeout > 0, "timeout should be positive");
     timeoutNanos = timeUnit.toNanos(timeout);
     return this;
@@ -74,7 +74,7 @@ public final class GrpcCleanupRule implements TestRule {
    */
   @SuppressWarnings("BetaApi") // Stopwatch.createUnstarted(Ticker ticker) is not Beta. Test only.
   @VisibleForTesting
-  GrpcCleanupRule withTicker(Ticker ticker) {
+  GrpcCleanupRule setTicker(Ticker ticker) {
     this.stopwatch = Stopwatch.createUnstarted(ticker);
     return this;
   }
--- a/testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
+++ b/testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
@@ -1,0 +1,268 @@
+/*
+ * Copyright 2018, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.testing;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Stopwatch;
+import com.google.common.base.Ticker;
+import io.grpc.ExperimentalApi;
+import io.grpc.ManagedChannel;
+import io.grpc.Server;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
+import javax.annotation.concurrent.NotThreadSafe;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+/**
+ * A JUnit {@link TestRule} that can register gRPC resources and manages its automatic release at
+ * the end of the test. If any of the resources registered to the rule can not be successfully
+ * released, the test will fail.
+ *
+ * @since 1.13.0
+ */
+@ExperimentalApi("https://github.com/grpc/grpc-java/issues/2488")
+@NotThreadSafe
+public final class GrpcCleanupRule implements TestRule {
+
+  private final List<Resource> resources = new ArrayList<Resource>();
+  private long timeoutNanos = TimeUnit.SECONDS.toNanos(10L);
+  private Stopwatch stopwatch = Stopwatch.createUnstarted();
+
+  private Throwable firstException;
+
+  /**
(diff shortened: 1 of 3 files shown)


testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
```diff
@@ -0,0 +1,268 @@
...
+   * Sets a positive total time limit for the automatic resource cleanup. If any of the resources
+   * registered to the rule fails to be released in time, the test will fail.
+   *
+   * <p>Note that the resource cleanup duration may or may not be counted as part of the JUnit
+   * {@link org.junit.rules.Timeout Timeout} rule's test duration, depending on which rule is
+   * applied first.
+   *
+   * @return this
+   */
+  public GrpcCleanupRule withTimeout(long timeout, TimeUnit timeUnit) {
```
ejona86: We tend to use `with` when a new instance will be returned. Let's use `set` or just `timeout` instead.

Looking at TemporaryFolder, it uses a builder instead. I'd be okay with that as well, if you'd prefer.

testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
```diff
@@ -0,0 +1,268 @@
...
+    timeoutNanos = timeUnit.toNanos(timeout);
+    return this;
+  }
+
+  /**
+   * Sets a specified time source for monitoring cleanup timeout.
+   *
+   * @return this
+   */
+  @SuppressWarnings("BetaApi") // Stopwatch.createUnstarted(Ticker ticker) is not Beta. Test only.
```
ejona86: What does "Stopwatch.createUnstarted(Ticker ticker) is not Beta." mean? It appears beta, as that's what you're suppressing...

testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
```diff
@@ -0,0 +1,268 @@
...
+  public Statement apply(final Statement base, Description description) {
+    return new Statement() {
+      @Override
+      public void evaluate() throws Throwable {
+        try {
+          base.evaluate();
+        } catch (Throwable t) {
+          firstException = t;
+        } finally {
+          teardown();
```
ejona86: Why is this in a finally?

testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
```diff
@@ -0,0 +1,268 @@
...
+    @Override
+    public void cleanUp() {
+      channel.shutdown();
+    }
+
+    @Override
+    public void forceCleanUp() {
+      try {
+        channel.shutdownNow();
+      } catch (Throwable ignorable) {
```
ejona86: shutdownNow should not throw. Let's not protect ourselves against that. If it happens, then we will just fix the bug.

testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
```diff
@@ -0,0 +1,268 @@
...
+    stopwatch.start();
+
+    for (int i = resources.size() - 1; i >= 0; i--) {
+      if (firstException != null) {
+        break;
+      }
+
+      try {
+        resources.get(i).cleanUp();
+      } catch (Throwable t) {
```
ejona86: shutdown should not throw. Let's not protect ourselves against that. If it happens, then we will just fix the bug.

testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
```diff
@@ -0,0 +1,268 @@
...
+        if (!released) {
+          firstException = new AssertionError(
+              "Resource " + resources.get(i) + " can not be released in time at the end of test");
+        }
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        if (firstException == null) {
+          firstException = e;
+        }
+      } catch (Throwable t) {
```
ejona86: awaitTerminated should not throw arbitrary exceptions. Let's not protect ourselves against that. If it happens, then we will just fix the bug.

testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
```diff
@@ -0,0 +1,268 @@
...
+        if (firstException == null) {
+          firstException = t;
+        }
+      } finally {
+        if (firstException != null) {
```
ejona86: This also seems unnecessary to be within a finally.

testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
dapengzhang0: The javadoc didn't say that. Even `ManagedChannelImpl` behaves so and javadoc says so, the user's test could be registering any `ManagedChannel` to the rule, say a mock `ManagedChannel` delegating to a `ManagedChannelImpl`, and their mock could happen to be really wild and evil. Same for `awaitTermination()` and `shutdown()`.

testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
dapengzhang0: Moved out of finally.

testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
dapengzhang0: Done.

testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
dapengzhang0: Done.

testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
ejona86: The methods do not define any exceptions will be thrown. Any Exception (vs Error) would be a bug (and as a general rule the amount of code that should avoid bugs in other code should be very limited) and any Error means "all hope is lost." If the user's test is doing something like that, I have no problem letting it fail. Those methods need to be reliable in general, and exposing such bugs seems fine to me. Garbage in, garbage out.

dapengzhang0: Done.

testing/src/main/java/io/grpc/testing/GrpcCleanupRule.java
dapengzhang0: I mean `Stopwatch` is not Beta, and it's API `Stopwatch.createUnstarted(Ticker ticker)` is also not marked with `@Beta`, even `Ticker` is marked with `@Beta` (and so I have to suppress).

ejona86: I still don't quite understand. Maybe "Ticker in Beta"?

dapengzhang0: Yes, Ticker is in Beta. I have to suprress.

Issue #2563, opened 2017-01-04
Tracking Issue for maxInboundMessageSize being Experimental


carl-mastrangelo:

ejona86:
This should include both ManagedChannelBuilder.setMaxInboundMessageSize and CallOptions.withMaxInboundMessageSize, since the two are closely related.

ejona86:
As mentioned in #2763:
> As per GoogleCloudPlatform/google-cloud-java#1499 , users of Pub/Sub see errors when their messages exceed 4MB, even though Pub/Sub's limit is 20 MB. Ideally we'd like to call `maxInboundMessageSize` to raise this maximum, but it turns out that `maxInboundMessageSize` is experimental.

ejona86:
I realized that all the other languages are using "max message size" on the channel. While Netty/OkhttpChannelBuilder had/have maxMessageSize, ManagedChannelBuilder has only had maxInboundMessageSize. I don't want to flip-flop users too much and the "inbound" portion of the name is a more precise, so I think it's fair to stabilize `ManagedChannelBuilder.maxInboundMessageSize(int)` as-is. That puts us reasonably close to other languages' support.

I think there are arguments to have the method on `CallOptions` as well (other than service config), but it doesn't seem to have any real use yet. It also has some awkwardness concerning `int` vs `Integer` with `getMaxOutboundMessageSize()`. So it doesn't seem quite time to `CallOptions.maxInboundMessageSize()`.

Unfortunately, `ServerBuilder` doesn't have any configuration option and `NettyServerBuilder` has `maxMessageSize()`.

tsl0922:
`4MB` is too low for our environment too, we have to set it on both server/client side manually after upgrade to 1.x version.

> Unfortunately, ServerBuilder doesn't have any configuration option and NettyServerBuilder has maxMessageSize().

Do you have any plan to make maxMessageSize() part of ServerBuilder?

ejona86:
> Do you have any plan to make maxMessageSize() part of ServerBuilder?

Yes. We'd add `maxInboundMessageSize()` to `ServerBuilder` to mirror `ManagedChannelBuilder`. I'd do that as part of removing the experimental annotation.

ejona86:
Server-side swaps to maxInboundMessageSize(). Trash maxMessageSize(). Then considered fine for being stable.

xin-au:
[...]

PR #4399, opened 2018-04-26
(low priority) core,netty,interop-testing: stabilize maxInboundMessageSize API

zpencer:
On server side, `maxMessageSize` is deprecated for
`maxInboundMessageSize` to match the channel builder.

Update usages to use new setter.

fixes https://github.com/grpc/grpc-java/issues/2563

### DIFF
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -293,12 +293,12 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
    * <p>This method is advisory, and implementations may decide to not enforce this.  Currently,
    * the only known transport to not enforce this is {@code InProcessTransport}.
    *
-   * @param max the maximum number of bytes a single message can be.
+   * @param bytes the maximum number of bytes a single message can be.
    * @return this
    * @throws IllegalArgumentException if max is negative.
    * @since 1.1.0
    */
-  public T maxInboundMessageSize(int max) {
+  public T maxInboundMessageSize(int bytes) {
     // intentional nop
     return thisT();
   }
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -294,8 +294,8 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
    * the only known transport to not enforce this is {@code InProcessTransport}.
    *
    * @param max the maximum number of bytes a single message can be.
-   * @throws IllegalArgumentException if max is negative.
    * @return this
+   * @throws IllegalArgumentException if max is negative.
    * @since 1.1.0
    */
   public T maxInboundMessageSize(int max) {
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -16,6 +16,7 @@
 
 package io.grpc;
 
+import com.google.common.base.Preconditions;
 import java.util.List;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -293,13 +293,12 @@ public T enableFullStreamDecompression() {
    * <p>This method is advisory, and implementations may decide to not enforce this.  Currently,
    * the only known transport to not enforce this is {@code InProcessTransport}.
    *
-   * @param max the maximum number of bytes a single message can be.
-   * @throws IllegalArgumentException if max is negative.
+   * @param bytes the maximum number of bytes a single message can be.
    * @return this
+   * @throws IllegalArgumentException if max is negative.
```
ejona86: s/max/bytes/

core/src/main/java/io/grpc/ServerBuilder.java
```diff
@@ -203,6 +203,25 @@ public T handshakeTimeout(long timeout, TimeUnit unit) {
...
+   * defaults to 4 MiB. The default provides protection to servers who haven't considered the
+   * possibility of receiving large messages while trying to be large enough to not be hit in normal
+   * usage.
+   *
+   * <p>This method is advisory, and implementations may decide to not enforce this.  Currently,
+   * the only known transport to not enforce this is {@code InProcessServer}.
+   *
+   * @param bytes the maximum number of bytes a single message can be.
+   * @return this
+   * @throws IllegalArgumentException if max is negative.
```
ejona86: s/max/bytes/

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -293,13 +293,12 @@ public T enableFullStreamDecompression() {
...
    * @since 1.1.0
    */
-  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/2307")
-  public T maxInboundMessageSize(int max) {
+  public T maxInboundMessageSize(int bytes) {
     // intentional nop
```
ejona86: Probably should add in a checkArgument for bytes >= 0.

core/src/main/java/io/grpc/ServerBuilder.java
```diff
@@ -203,6 +203,25 @@ public T handshakeTimeout(long timeout, TimeUnit unit) {
...
+   * @throws UnsupportedOperationException if unsupported.
+   * @since 1.13.0
+   */
+  public T maxInboundMessageSize(int bytes) {
+    throw new UnsupportedOperationException();
```
ejona86: Client-side doesn't throw. Since this is advisory, that seems the better approach?

zpencer: Changed and added comment.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
zpencer: Done

core/src/main/java/io/grpc/ServerBuilder.java
zpencer: Done

core/src/main/java/io/grpc/ManagedChannelBuilder.java
zpencer: Done

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -293,14 +294,14 @@ public T enableFullStreamDecompression() {
...
+   * @param bytes the maximum number of bytes a single message can be.
    * @return this
+   * @throws IllegalArgumentException if bytes is negative.
    * @since 1.1.0
    */
-  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/2307")
-  public T maxInboundMessageSize(int max) {
-    // intentional nop
+  public T maxInboundMessageSize(int bytes) {
+    // intentional noop rather than throw, this method is only advisory.
```
carl-mastrangelo: I think you need to have both methods for a short time.  We typically have one release with deprecation before removing it.

Also, this makes it easier to make internal users update before removal.

ejona86: No methods were removed in this PR.

zpencer: Yes, I just renamed `max` to `bytes` to be consistent.

Issue #3961, opened 2018-01-11
Drop support for Java 6


ejona86:
This would allow using (at least some) Java 7 language features and more of the Java 7 API. Android support for Java 7 has advanced. We need to figure out exactly what restrictions we'll need, but on the surface things seem far enough along to push forward with the drop of Java 6 support.

We'd still support the subset of Java 7 that is used by Android; we're not willing to reduce our API level support for this.

@ericgribkoff

ejona86:
For anyone seeing this and panicking, "it's okay." We plan to be careful with this and check that any solutions will work well for our users. This is an eventual goal.

adriancole:
java 7 has few notable language features, and I think most of these can be
rewritten with retrolambda.
java 6 allows old apps to be automatically instrumented with an agent,
something that @CodingFabian mentioned to me some time back.

AFAIK you there are some lowest common denominator bytecode level rules
(newer apis don't matter as long as they are not required to be called)

In brave, for example, we still routinely get requests for tracing spring
2.5 applications (over 10 years old), presumably the reason people ask is
indeed because they can't change them.

Reason I mention this is to just help spread awareness of the android
unrelated concern, not to suggest that concern must be addressed here

adriancole:
oops sorry I thought this was the census repo. Ignore me!

ejona86:
@adriancole, Census at least was partially following suit from gRPC, so it's still useful to know that even if gRPC drops support it may be good for Census to be more conservative.

We do actually agree that there were few Java 7 language features that are transformative. So we've not pushed too hard. But we do want Java 8 language features, and Java 7 is a stepping stone for that.

Android has come along and is even supporting Java 8 features. Java 7 is not receiving public updates and [Java 8/9 public updates will stop this year](http://www.oracle.com/technetwork/java/eol-135779.html). Combined with the ["one + three back" policy](http://openjdk.java.net/jeps/182) means we may find ourselves not easily able to support Java 6. Adding in Oracle's move to a 6 month release schedule forms a bit of a perfect storm. We could possibly post-process with retrolambda or similar, but EOL increases testing burden as well.

I'm sure we'll not be the only ones trying to figure things out, so some common approaches will probably surface. But the burden is increasing for unknown benefit. Even if we decide to continue supporting Java 6, we should have an idea of how much benefit it's providing and how to determine when to stop.

[...]

PR #4401, opened 2018-04-27
core,stub: Add toString() to ClientCallImpl and GrpcFuture

groakley:
Improves debugging experience with ClientCall instances.

Sent on behalf of @charlesmunger

### DIFF
--- a/core/src/main/java/io/grpc/internal/ClientCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ClientCallImpl.java
@@ -471,6 +471,11 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
     observer.onClose(status, trailers);
   }
 
+  @Override
+  public String toString() {
+    return super.toString() + "method=[" + method + "]";
+  }
+
   private class ClientStreamListenerImpl implements ClientStreamListener {
     private final Listener<RespT> observer;
     private boolean closed;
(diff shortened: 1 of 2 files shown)


groakley:
Hmm, looks like this project is being built against Guava version 20.0. pendingToString() was added in 23.0. Are there any plans to bump the Guava version?

thelinuxfoundation:

* User @groakley isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

groakley:
Updated corp CLA status.

ejona86:
@groakley, it seems the CLA is still a bit unhappy with you. I think #3961 is a blocker for us going as far as Guava 23. It'll happen, but take a little bit. `pendingToString()` is just an overridden method, not something you call, so it should be fine to have the method and just not put an `@Override`. If a user is using a newer Guava, they'll get the better toString.

charlesmunger:
@ejona86 #3961 shouldn't be a blocker for adopting newer versions of guava, since guava has two versions - one for java 8, one for android.

https://github.com/google/guava/wiki/Release22#new-android-version

Does grpc publish separate targets for android and not-android?

ejona86:
@charlesmunger, the Android version of Guava uses JDK 7 bytecode. The number of Java 6 users should be low, but it is a prerequisite.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -471,6 +471,11 @@ private void closeObserver(Listener<RespT> observer, Status status, Metadata tra
     observer.onClose(status, trailers);
   }
 
+  @Override
+  public String toString() {
+    return super.toString() + "method=[" + method + "]";
```
dapengzhang0: We are using Guava `MoreObjects.ToStringHelper` instead of harding coding `=[]`

Ditto.

groakley:
Commenting again with hopes that I've jumped through the CLA hoops properly.

ejona86:
Thank you!

PR #4402, opened 2018-04-28
android: avoid static init of robolectric shadows

ericgribkoff:
Some robolectric test runners do not have access to the Android runtime during static initialization, causing these variable declarations to fail.


android/src/test/java/io/grpc/android/AndroidChannelBuilderTest.java
```diff
@@ -48,10 +48,10 @@
 @RunWith(RobolectricTestRunner.class)
 @Config(shadows = {AndroidChannelBuilderTest.ShadowDefaultNetworkListenerConnectivityManager.class})
 public final class AndroidChannelBuilderTest {
-  private static final NetworkInfo WIFI_CONNECTED =
+  private final NetworkInfo WIFI_CONNECTED =
```
dapengzhang0: Seems should use camel case. But it's merge already...
https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names
Constants are static final fields whose contents are deeply immutable and whose methods have no detectable side effects.
Non-constant field names (static or otherwise) are written in lowerCamelCase.

PR #4404, opened 2018-04-28
core,services: no service providers for BinaryLogSink, add TempFileSink

zpencer:
No need to use service provider for BinaryLogSink, it can just be an
interface that is passed into BinaryLogProviderImpl.

Add a default TempFileSink that uses the protobuf object's
writeDelimited method to write to the output stream.


services/src/main/java/io/grpc/services/BinaryLogProviderImpl.java
```diff
@@ -34,12 +39,25 @@
...
-    this(BinaryLogSinkProvider.provider(), System.getenv("GRPC_BINARY_LOG_CONFIG"));
+    this(new TempFileSink(), System.getenv("GRPC_BINARY_LOG_CONFIG"));
   }
 
-  BinaryLogProviderImpl(BinaryLogSink sink, String configStr) {
+  BinaryLogProviderImpl(final BinaryLogSink sink, String configStr) {
     BinaryLog.Factory factory = null;
     try {
+      Preconditions.checkNotNull(sink);
+      Runtime.getRuntime().addShutdownHook(new Thread() {
```
ejona86: We can't be doing this in a library.

We can't even assume our library should be loaded when the JVM exits. For example, in servlet environments we could be unloaded very frequently, but this would retain our class loader.

services/src/main/java/io/grpc/services/TempFileSink.java
```diff
@@ -0,0 +1,102 @@
...
+    return this.outPath;
+  }
+
+  @Override
+  public synchronized void write(MessageLite message) {
+    if (closed) {
+      return;
+    }
+    try {
+      message.writeDelimitedTo(out);
```
ejona86: This scares me. Doing any sort of blocking has typically resulted in "bad things" of various sorts. It could maybe be acceptable here because there are "flow control" concerns, but we should discuss it.

zpencer:
Will come back to this one after https://github.com/grpc/grpc-java/pull/4431, which should make both of the concerns easier to address:
* users will be responsible for calling `close()` when appropriate
* explicit construction makes it easier to specify an executor for non blocking writes

zpencer:
@ejona86 this now contains the changes, PTAL

PR #4405, opened 2018-04-28
services: add null check to binlog provider

zpencer:
There needs to be a check for when no logging should happen and a null value was returned from the factory.

### DIFF
--- a/services/src/main/java/io/grpc/services/BinaryLogProviderImpl.java
+++ b/services/src/main/java/io/grpc/services/BinaryLogProviderImpl.java
@@ -52,14 +52,22 @@ public class BinaryLogProviderImpl extends BinaryLogProvider {
   @Nullable
   @Override
   public ServerInterceptor getServerInterceptor(String fullMethodName) {
-    return factory.getLog(fullMethodName).getServerInterceptor(getServerCallId());
+    BinaryLog log = factory.getLog(fullMethodName);
+    if (log == null) {
+      return null;
+    }
+    return log.getServerInterceptor(getServerCallId());
   }
 
   @Nullable
   @Override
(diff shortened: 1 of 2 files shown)


services/src/main/java/io/grpc/services/BinaryLogProviderImpl.java
```diff
@@ -52,14 +52,22 @@ public BinaryLogProviderImpl() {
   @Nullable
   @Override
   public ServerInterceptor getServerInterceptor(String fullMethodName) {
-    return factory.getLog(fullMethodName).getServerInterceptor(getServerCallId());
+    BinaryLog log = factory.getLog(fullMethodName);
```
sonyakc: lines 55-58 and 66-69 are the same and could effectively be pulled out to a helper method so the code is shared.

services/src/main/java/io/grpc/services/BinaryLogProviderImpl.java
```diff
@@ -52,14 +52,22 @@ public BinaryLogProviderImpl() {
...
+    }
+    return log.getServerInterceptor(getServerCallId());
   }
 
   @Nullable
   @Override
   public ClientInterceptor getClientInterceptor(
       String fullMethodName, CallOptions callOptions) {
-    return factory.getLog(fullMethodName).getClientInterceptor(getClientCallId(callOptions));
+    BinaryLog log = factory.getLog(fullMethodName);
```
sonyakc: What happens if ```fullMethodName``` is null? Would a null check help here?

zpencer: It would be a good defensive check. I'll add one.

PR #4406, opened 2018-04-28
buildscripts: automate the release process

zpencer:
This introduces the scripts for the auto release process.

In kokoro, there will be a job `release_artifacts` that the team
can trigger manually with an RPC, it does not run automatically.
The job will kick off 3 other jobs:
- macos: builds protoc-gen-grpc-java-1.13.0-osx-x86_64.exe
- windows: builds protoc-gen-grpc-java-1.13.0-windows-x86_32.exe and
  protoc-gen-grpc-java-1.13.0-windows-x86_64.exe
- linux_artifacts: builds protoc-gen-grpc-java-1.13.0-linux-x86_32.exe,
  protoc-gen-grpc-java-1.13.0-linux-x86_64.exe, and the jars

If all 3 jobs are successful then a child job `upload_artifacts` is
triggered. This job downloads all the artifacts, signs the files, then
uploads them to oss.sonatype.org in a staging repo.

### DIFF
--- a/buildscripts/grpc-java-releasing/Dockerfile
+++ b/buildscripts/grpc-java-releasing/Dockerfile
@@ -1,5 +1,14 @@
 FROM protoc-artifacts:latest
 
+# Work around the docker cache. Not really needed
+# because Kokoro instances start off with a blank
+# slate. But useful to have this here because we may
+# want to run these files locally.
+RUN echo "cache buster 4/24/2018"
+
+# protoc-artifacts base image is ancient and curl does
+# not work properly.
+RUN yum -y update; yum clean all || true
 COPY scl-enable-devtoolset.sh /
 
 # Start in devtoolset environment that uses GCC 4.7
--- a/buildscripts/grpc-java-releasing/Dockerfile
+++ b/buildscripts/grpc-java-releasing/Dockerfile
@@ -1,14 +1,9 @@
 FROM protoc-artifacts:latest
 
-# Work around the docker cache. Not really needed
-# because Kokoro instances start off with a blank
-# slate. But useful to have this here because we may
-# want to run these files locally.
-RUN echo "cache buster 4/24/2018"
-
 # protoc-artifacts base image is ancient and curl does
 # not work properly.
-RUN yum -y update; yum clean all || true
+RUN yum -y update || true # Update to 4/24/2018 or newer, ignore rpmdb checksum errors
+RUN yum clean all
 COPY scl-enable-devtoolset.sh /
 
 # Start in devtoolset environment that uses GCC 4.7
--- a/buildscripts/grpc-java-releasing/Dockerfile
+++ b/buildscripts/grpc-java-releasing/Dockerfile
@@ -1,9 +1,8 @@
 FROM protoc-artifacts:latest
 
 # protoc-artifacts base image is ancient and curl does
-# not work properly.
-RUN yum -y update || true # Update to 4/24/2018 or newer, ignore rpmdb checksum errors
-RUN yum clean all
+# not work properly. Ignore rpmdb checksum errors from update.
+RUN yum -y update; yum clean all # Update to >= 4/24/2018
 COPY scl-enable-devtoolset.sh /
 
 # Start in devtoolset environment that uses GCC 4.7
--- a/buildscripts/kokoro/linux_artifacts.sh
+++ b/buildscripts/kokoro/linux_artifacts.sh
@@ -5,6 +5,19 @@ if [[ -f /VERSION ]]; then
   cat /VERSION
 fi
 
-readonly GRPC_JAVA_DIR=$(cd $(dirname $0)/../.. && pwd)
+readonly GRPC_JAVA_DIR="$(cd "$(dirname "$0")"/../.. && pwd)"
 
-# A place holder at the moment
+rm -rf /tmp/source_head/protobuf
+mkdir -p /tmp/source_head/protobuf
+git clone https://github.com/google/protobuf.git /tmp/source_head/protobuf
+
+cd /tmp/source_head/protobuf
+git fetch origin
+git checkout origin/master
+docker build -t protoc-artifacts protoc-artifacts
+
+cd "$GRPC_JAVA_DIR"/buildscripts/
+
+docker build -t grpc-java-releasing grpc-java-releasing
+
+"$GRPC_JAVA_DIR"/buildscripts/run_in_docker.sh /grpc-java/buildscripts/build_artifacts_in_docker.sh
--- a/buildscripts/kokoro/linux_artifacts.sh
+++ b/buildscripts/kokoro/linux_artifacts.sh
@@ -7,8 +7,6 @@ fi
 
 readonly GRPC_JAVA_DIR="$(cd "$(dirname "$0")"/../.. && pwd)"
 
-PROTOBUF_VERSION=3.5.1
-
 rm -rf /tmp/source_head/protobuf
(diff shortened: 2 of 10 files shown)


buildscripts/kokoro/linux_artifacts.sh
```diff
@@ -5,6 +5,19 @@ if [[ -f /VERSION ]]; then
   cat /VERSION
 fi
 
-readonly GRPC_JAVA_DIR=$(cd $(dirname $0)/../.. && pwd)
+readonly GRPC_JAVA_DIR="$(cd "$(dirname "$0")"/../.. && pwd)"
 
-# A place holder at the moment
+rm -rf /tmp/source_head/protobuf
+mkdir -p /tmp/source_head/protobuf
```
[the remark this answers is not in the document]
zpencer: The "set -e" on line 2 makes it so that if any commands fail, the entire script exits with a failure. The other options are also good defensive features that we turn on.

zpencer:
@ejona86 PTAL I plan to use these scripts for the upcoming release.

buildscripts/grpc-java-releasing/Dockerfile
```diff
@@ -1,5 +1,14 @@
 FROM protoc-artifacts:latest
 
+# Work around the docker cache. Not really needed
+# because Kokoro instances start off with a blank
+# slate. But useful to have this here because we may
+# want to run these files locally.
+RUN echo "cache buster 4/24/2018"
```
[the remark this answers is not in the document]
zpencer: The "echo" is  just a garbage command that we can change next time we need to run yum update again. The reason we need to update is that the autogen step of protoc fails inside the container. It appears to be using curl to download gmock.

buildscripts/kokoro/linux_artifacts.sh
```diff
@@ -5,6 +5,19 @@ if [[ -f /VERSION ]]; then
...
+git clone https://github.com/google/protobuf.git /tmp/source_head/protobuf
```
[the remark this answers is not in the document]
zpencer: The latest release still does not have this patch:
https://github.com/google/protobuf/commit/5f052ae7ae8a757dd919e95e0f1481e08fef749b#diff-0bc9399720fa75bdc363a61a50cd2609

When the docker image is being built using the latest release, it fails at this step:
```
Step 13/14 : RUN git clone --depth 1 https://github.com/google/protobuf.git
 ---> Running in a302e905a09c
Initialized empty Git repository in /protobuf/.git/
error:  while accessing https://github.com/google/protobuf.git/info/refs

fatal: HTTP request failed
The command '/bin/sh -c git clone --depth 1 https://github.com/google/protobuf.git' returned a non-zero code: 128

```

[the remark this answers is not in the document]
ejona86: We shouldn't use master. Master changes. If you want to use current HEAD, that's fine. Check out `92898e9e`. You can also download it at https://github.com/google/protobuf/archive/92898e9e.zip

buildscripts/kokoro/linux_artifacts.sh
```diff
@@ -18,6 +22,6 @@ docker build -t protoc-artifacts protoc-artifacts
 
 cd "$GRPC_JAVA_DIR"/buildscripts/
```
ejona86: Delete?

buildscripts/kokoro/upload_artifacts.sh
```diff
@@ -0,0 +1,66 @@
...
+
+mkdir -p ~/java_signing/
+gsutil cp -r gs://grpc-testing-secrets/java_signing/ ~/
+gpg --batch  --import ~/java_signing/grpc-java-team-sonatype.asc
+
+# gpg commands changed between v1 and v2 are different.
+gpg --version
+
+# This is the version found on kokoro.
+if [[ $(gpg --version | grep 'gpg (GnuPG) 1.') ]]; then
```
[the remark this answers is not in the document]
ejona86: I don't think `$()` would be right even then. I think a subshell `()` is what you'd want.

buildscripts/grpc-java-releasing/Dockerfile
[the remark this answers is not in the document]
ejona86: Are you saying that doing a yum update now fixes curl? That previously wasn't the case. We should add that to protoc-artifacts instead of here.

buildscripts/grpc-java-releasing/Dockerfile
```diff
@@ -1,5 +1,14 @@
...
+
+# protoc-artifacts base image is ancient and curl does
+# not work properly.
+RUN yum -y update; yum clean all || true
```
ejona86: Why isn't this `yum -y update && yum clean all`?

buildscripts/run_in_docker.sh
```diff
@@ -10,31 +10,10 @@ quote() {
...
-  # be able to remove the 'quote' function.
-  exec su theuser -c "$(quote "$@")"
-fi
+readonly grpc_java_dir="$(dirname "$(readlink -f "$0")")/.."
+# We do not use -it because the input device in kokoro is not a TTY.
+# Use a trap function to fix file permissions upon exit, without affecting
+# the original exit code.
+exec docker run --rm=true -v "${grpc_java_dir}":/grpc-java -w /grpc-java \
+  grpc-java-releasing \
+  bash -c "function fixFiles() { chown -R "$(id -u)":"$(id -g)" /grpc-java; }; trap fixFiles EXIT; $(quote "$@")"
```
ejona86: The quotes around the `$(id blah)` seem either unnecessary or broken. Either remove them or make them `'`.

buildscripts/run_in_docker.sh
```diff
@@ -10,31 +10,10 @@ quote() {
...
-useradd theuser -u "$swap_uid" -g "$swap_gid" -m
-if [[ "$#" -eq 0 ]]; then
-  exec su theuser
-else
-  # runuser is too old in the container to support the -u flag; if it did, we'd
-  # be able to remove the 'quote' function.
-  exec su theuser -c "$(quote "$@")"
-fi
+readonly grpc_java_dir="$(dirname "$(readlink -f "$0")")/.."
+# We do not use -it because the input device in kokoro is not a TTY.
```
ejona86: Hmm... removing -it makes it quite annoying to play with. How about we conditionalize it based on whether stdin exists:
```
if [[ -t 0 ]]; then
  DOCKER_ARGS="-it"
else
  DOCKER_ARGS=
fi
```

Oh, so annoying.

buildscripts/grpc-java-releasing/Dockerfile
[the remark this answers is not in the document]
zpencer: done

buildscripts/run_in_docker.sh
zpencer: done

buildscripts/run_in_docker.sh
zpencer: Removed

buildscripts/grpc-java-releasing/Dockerfile
ejona86: No, `RUN yum -y update; yum clean all` would have been the way. Each RUN creates a new layer. Doing the clean in a separate layer is useless.

zpencer: done

buildscripts/kokoro/linux_artifacts.sh
zpencer: Removed

PR #4413, opened 2018-04-30
 grpclb: move load balancer proto to package-matching directory

carl-mastrangelo:


carl-mastrangelo:
fyi had to update the BUILD.bazel file

PR #4419, opened 2018-05-01
compiler: make testing service used by compiler not conflict with interop-testing proto

carl-mastrangelo:
This change puts the compiler test.proto file in a directory that matches it's path.  The current proto package has a namespace collision with the grpc.testing proto in interop-testing.  Thus, this PR:

* change the package of test.proto to be grpc.testing.compiler
* moves the test.proto into the matching grpc/testing/compiler/test.proto path
* updates the generated code.


ejona86:
Nudging the CLA bot.

Issue #4412, opened 2018-04-30
Retried RPCs have stale grpc-timeout metadata


ejona86:
Today, grpc-timeout is set by ClientCallImpl based on the deadline. But RetriableStream doesn't re-set the grpc-timeout when issuing more RPCs.

It'd probably be good to move grpc-timeout handling down into the transport. We already have the problem that any RPCs queued by DelayedTransport will have more stale timeouts in their headers. It'd be better to calculate the timeout as late as possible.

dapengzhang0:
Was there an issue number for the `DelayClientTransport`?

ejona86:
@dapengzhang0, no, I don't think so.

PR #4421, opened 2018-05-01
core: populate effective deadline to ClientStream

dapengzhang0:
Added `ClientStream.setDeadline(Deadline deadline)` method, which will set the timeout header.

Resolving #4412

### DIFF
--- a/core/src/main/java/io/grpc/inprocess/InProcessTransport.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessTransport.java
@@ -17,12 +17,15 @@
 package io.grpc.inprocess;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static io.grpc.internal.GrpcUtil.TIMEOUT_KEY;
+import static java.lang.Math.max;
 
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
 import io.grpc.Attributes;
 import io.grpc.CallOptions;
 import io.grpc.Compressor;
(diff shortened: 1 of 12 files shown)


core/src/main/java/io/grpc/inprocess/InProcessTransport.java
```diff
@@ -664,6 +668,13 @@ public void setMaxInboundMessageSize(int maxSize) {}
 
       @Override
       public void setMaxOutboundMessageSize(int maxSize) {}
+
+      @Override
+      public void setDeadline(Deadline deadline) {
+        headers.discardAll(TIMEOUT_KEY);
+        long effectiveTimeout = max(0, deadline.timeRemaining(TimeUnit.NANOSECONDS));
```
ejona86: This `max()` check can be put in `TIMEOUT_KEY`'s marshaller.

dapengzhang0: Then the marshaller will not look as nice as before because serialization and deserializeation is no longer inverse of each other.

ejona86: Inverse in what way? Like put()ing a `-1` doesn't get() a `-1`, or the method implementations no longer appear symmetric?

ejona86: We spoke offline. It was more the -1 case. Since `TIMEOUT_KEY` has a `< 0` check already, I was okay with leaving this code as-is.

PR #4422, opened 2018-05-01
core: Document Metadata ownership passes to the Call{,Listener}

ejona86:
This was already clearly documented in ServerCall.Listener and
ClientCall.

### DIFF
--- a/core/src/main/java/io/grpc/ClientCall.java
+++ b/core/src/main/java/io/grpc/ClientCall.java
@@ -108,8 +108,10 @@ public abstract class ClientCall<ReqT, RespT> {
   public abstract static class Listener<T> {
 
     /**
-     * The response headers have been received. Headers always precede messages. Since {@link
-     * Metadata} is not thread-safe, the caller must not access {@code headers} after this point.
+     * The response headers have been received. Headers always precede messages.
+     *
+     * <p>Since {@link Metadata} is not thread-safe, the caller must not access (read or write)
+     * {@code headers} after this point.
      *
      * @param headers containing metadata sent by the server at the start of the response.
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/grpc/ClientCall.java
```diff
@@ -108,7 +108,8 @@
   public abstract static class Listener<T> {
 
     /**
-     * The response headers have been received. Headers always precede messages.
+     * The response headers have been received. Headers always precede messages. Since {@link
+     * Metadata} is not thread-safe, the caller must not access {@code headers} after this point.
```
carl-mastrangelo: This is awkward, since move user will be receiving this callback (rather than forwarding it).   It currently reads as if the headers cannot be accessed upon getting this callback.

carl-mastrangelo: I missed that, maybe split that sentence out?  Also, regardless of threading, its still not safe for callers to mutate the MD afterwards.

ejona86: You mean make a new paragraph? That'd be easy.

Part of the point is they can't even _read_ from the object. There's been some logging interceptors internally that were getting that wrong, which prompted fixing up these missed spots.

carl-mastrangelo: new paragraph is fine.    I would say read or write, rather than access to be more explicit.  Those are the keywords I would look for.  Optional, but my opinion anyways.

ejona86:
@carl-mastrangelo, PTAL. I hope this is a bit more clear. Tell me what you think.

PR #4427, opened 2018-05-02
core: Adds @CheckReturnValue to CallOptions

### LINKED ISSUE #3671: Add @CheckReturnValue to CallOptions.withX() methods
This would help prevent errors like the following toy example

This is incorrect
```java
CallOptions setExecutor(CallOptions callOptions, Executor executor) {
  // @CheckReturnValue would have caused this line to be flagged
  callOptions.withExecutor(executor);
  return callOptions;
}
```

It should be
```java
CallOptions setExecutor(CallOptions callOptions, Executor executor) {
  return callOptions.withExecutor(executor);
}
```

groakley:
The handful of Google internal targets that violated this check have already been cleaned up. It looks like since @carl-mastrangelo first attempted this change, the errorprone tooling has improved such that this is no longer flagging false-positives, like unit tests.

Fixes #3671

### DIFF
--- a/core/src/main/java/io/grpc/CallOptions.java
+++ b/core/src/main/java/io/grpc/CallOptions.java
@@ -26,6 +26,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
+import javax.annotation.CheckReturnValue;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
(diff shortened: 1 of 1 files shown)


ejona86:
Woo!

dapengzhang0:
@groakley Do you happen to know which commit or CL in errorprone tooling improved such that this is no longer flagging false-positives, like unit tests? If open source users have a lot false-positives (although not likely), and they have to bump errorprone versions, that might break grpc-java build because `@DoNotMock` was removed from errorprone lately in open source world.

groakley:
@dapengzhang0 I don't have a comprehensive list, but here are a couple of representative issues that were addressed (both early 2016):

- https://github.com/google/error-prone/issues/226
- https://github.com/google/error-prone/issues/385

These sound like the false positives that @carl-mastrangelo was describing.

I would venture to guess v2.0.9 would be the first release to contain those changes

PR #4430, opened 2018-05-03
context/core/netty: Add @CheckReturnValue to Context

dapengzhang0:
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc-java/4430)

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -1065,6 +1065,8 @@ public class Context {
     }
   }
 
+  // Not using the standard com.google.errorprone.annotations.CheckReturnValue because that will
+  // introduce dependencies that some io.grpc.Context API consumers may not want.
   @interface CheckReturnValue {}
 
   @interface CanIgnoreReturnValue {}
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -16,6 +16,7 @@
 
 package io.grpc;
 
+import io.grpc.Context.CheckReturnValue;
 import java.io.Closeable;
 import java.util.ArrayList;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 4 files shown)


carl-mastrangelo:
This needs to be check with an internal presubmit first.

dapengzhang0:
> This needs to be check with an internal presubmit first.
Yes, this change is scary.

dapengzhang0:
The internal presubmit (CL/195275931) passed.

carl-mastrangelo:
One other concern: this package has traditionally not had any dependencies (including DoNotMock).   @adriancole I recall you being involved in the discussion around deps of Context.  Would taking a compile time dependency on ErrorProne annotations be feasible?

dapengzhang0:
> One other concern: this package has traditionally not had any dependencies (including DoNotMock). @adriancole I recall you being involved in the discussion around deps of Context. Would taking a compile time dependency on ErrorProne annotations be feasible?

One of Context users, Zipkin instrumentation libs, has no deps, not even annotation jars. So this PR is deliberately not adding any dependencies. So `com.google.errorprone.annotations.CheckReturnValue` can not be used in `io.grpc.Context`. Instead, it uses `io.grpc.Context.CheckReturnValue`.  It is up to the compiler, whether or not to apply the errorprone plug-in when compiling (in static analysis phase). Internally, the JavaBuilder will apply errorprone by default for all projects.

context/src/main/java/io/grpc/Context.java
```diff
@@ -1059,4 +1064,8 @@ private static void validateGeneration(int generation) {
           new Exception());
     }
   }
+
```
carl-mastrangelo: This probably deserves a comment why it isn't the error prone version.

dapengzhang0:
Thanks for the review @carl-mastrangelo .

PR #4431, opened 2018-05-03
core, services: make BinaryLog an explicit object that is passed into channels/servers

zpencer:
Previously: BinaryLogProvider is a service provider that channels
and servers consult with to find the binary log. Knowing when to
close the output stream is a problem with this approach.

Now io.grpc.BinaryLog is an opaque object that the user can
create at program start up. The only public method is
close(). This is the object to pass into ServerBuilder and
ManagedChannelBuilder.

`io.grpc.BinaryLogProvider` will be renamed to something
that no longer suggests an SPI. It is an `@Internal` class
and its visibility will be further locked down later.

Each commit represents one logical change.

### DIFF
--- a/core/src/main/java/io/grpc/BinaryLogProvider.java
+++ b/core/src/main/java/io/grpc/BinaryLogProvider.java
@@ -28,6 +28,7 @@ import java.nio.ByteBuffer;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
 
+// TODO(zpencer): rename class to AbstractBinaryLog
 @Internal
 public abstract class BinaryLogProvider implements Closeable {
   public static final CallOptions.Key<CallId> CLIENT_CALL_ID_CALLOPTION_KEY
--- a/core/src/main/java/io/grpc/BinaryLogProvider.java
+++ b/core/src/main/java/io/grpc/BinaryLogProvider.java
@@ -17,13 +17,15 @@
 package io.grpc;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import io.grpc.MethodDescriptor.Marshaller;
-import io.grpc.internal.IoUtils;
 import io.opencensus.trace.Span;
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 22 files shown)


core/src/test/java/io/grpc/internal/BinaryLogProviderTest.java
```diff
@@ -288,48 +288,6 @@ public void onMessage(String message) {
         (int) method.parseResponse(new ByteArrayInputStream((byte[]) serializedResp.get(0))));
   }
 
-  @Test
```
zpencer: This is needed because moving the class from `io.grpc.internal` to `io.grpc` means we lose package private access to census.

core/src/main/java/io/grpc/BinaryLogProvider.java
```diff
@@ -14,66 +14,34 @@
...
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.nio.ByteBuffer;
-import java.util.Collections;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
 
+// TODO(zpencer): rename class to AbstractBinaryLog
+@Internal
 public abstract class BinaryLogProvider implements Closeable {
```
ejona86: It doesn't seem this needs to be in io.grpc. Only the two public methods are necessary. Those can be abstract in a separate class from BinaryLog (like is done now) or they can be on BinaryLog itself. Long-term I'm not sure it makes sense for this to be locked down.

zpencer: I wanted something that can be passed into ManagedChannelBuilder and ServerBuilder, and I remember io.grpc can not depend on io.grpc.internal. Or are you suggesting to nest it inside of io.grpc.BinaryLog ?

ejona86: I'm suggesting to strip the body of BinaryLogProvider out. The two-method interface is all that's necessary in io.grpc. I'd understand a bit more if the fact that those two methods are `final` limited the scope of the API, but they don't really limit the power of the interface.

PR #4433, opened 2018-05-03
android: add artifact build

ericgribkoff:
This enables building the `grpc-android` .aar artifact. Conscrypt's conscrypt-android [build.gradle](https://github.com/google/conscrypt/blob/0b798b404f9072af38ca6268b2a3d0cf304481cd/android/build.gradle) was a useful reference.

Conscrypt includes conscrypt-android as a subproject of their root build, but this PR leaves the grpc-android artifact as a separate build instead. There are two reasons for this: (1) the plugins to build and deploy Android are different, so Conscrypt duplicated the deploy config anyways and (2) with the latest versions of gradle and the android gradle plugin, mixing java and android builds introduces classpath guava conflicts that break gradle (not the build, breaks the gradle tool itself). Resolving these issues and integrating grpc-android as a subproject of the top-level build would be helpful in the future.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,5 +1,7 @@
 apply plugin: 'com.android.library'
 
+group = "io.grpc"
+version = "1.13.0-SNAPSHOT" // CURRENT_GRPC_VERSION
 description = 'gRPC: Android'
 
 buildscript {
@@ -14,10 +16,12 @@ buildscript {
     dependencies {
         classpath 'com.android.tools.build:gradle:3.0.1'
         classpath "net.ltgt.gradle:gradle-errorprone-plugin:0.0.13"
+        classpath 'com.github.dcendents:android-maven-gradle-plugin:2.0'
     }
 }
 
 apply plugin: "net.ltgt.errorprone"
+apply plugin: 'com.github.dcendents.android-maven'
 
 android {
     compileSdkVersion 27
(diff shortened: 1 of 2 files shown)


android/build.gradle
```diff
@@ -45,9 +44,99 @@ repositories {
...
+    testImplementation 'io.grpc:grpc-okhttp:1.13.0-SNAPSHOT' // CURRENT_GRPC_VERSION
     testImplementation 'junit:junit:4.12'
     testImplementation 'org.robolectric:robolectric:3.7.1'
     testImplementation 'com.google.truth:truth:0.39'
 }
+
+task javadocs(type: Javadoc) {
+    source = android.sourceSets.main.java.srcDirs
+    // TODO(ericgribkoff) Fix javadoc errors
+    failOnError false
```
zpencer: What sorts of errors are we hitting?

ericgribkoff: Primarily lint errors from `{@link }` annotations that I haven't looked into fixing yet.

ericgribkoff:
@ejona86 I spent some time trying to get this working with the java plugin etc conditional, but when you actually go to build the subproject with the android plugin, it seems just having them both on the buildscript.dependencies.classpath breaks Gradle, due to incompatible Guava versions being pulled in from...somewhere. I will try to figure out a solution for this (in another PR) so we can avoid the duplication.

PR #4434, opened 2018-05-03
core,netty,okhttp: make toString more consistent for channelz

zpencer:
Use MoreObjects.toStringHelper and use only the log id's long value,
because the class name is already present in the toStringHelper.

### DIFF
--- a/core/src/main/java/io/grpc/internal/InternalSubchannel.java
+++ b/core/src/main/java/io/grpc/internal/InternalSubchannel.java
@@ -23,6 +23,7 @@ import static io.grpc.ConnectivityState.SHUTDOWN;
 import static io.grpc.ConnectivityState.TRANSIENT_FAILURE;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Stopwatch;
 import com.google.common.base.Supplier;
@@ -387,6 +388,16 @@ final class InternalSubchannel implements Instrumented<ChannelStats> {
     }
     if (savedPendingTransport != null) {
       savedPendingTransport.shutdown(reason);
+    }
+  }
+
+  @Override
+  public String toString() {
+    synchronized (lock) {
+      return MoreObjects.toStringHelper(this)
+          .add("logId", logId.getId())
+          .add("addressGroup", addressGroup.toString())
+          .toString();
     }
   }
 
--- a/core/src/main/java/io/grpc/internal/InternalSubchannel.java
+++ b/core/src/main/java/io/grpc/internal/InternalSubchannel.java
@@ -393,9 +393,11 @@ final class InternalSubchannel implements Instrumented<ChannelStats> {
 
   @Override
   public String toString() {
+    // addressGroupCopy being a little stale is fine, just avoid calling toString with the lock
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -390,6 +391,16 @@ public void shutdown(Status reason) {
...
 
+  @Override
+  public String toString() {
+    synchronized (lock) {
+      return MoreObjects.toStringHelper(this)
+          .add("logId", logId.getId())
+          .add("addressGroup", addressGroup.toString())
+          .toString();
+    }
+  }
```
dapengzhang0: nit: 

```java
Object addressGroupCopy;
synchronized (lock) {
  addressGroupCopy = addressGroup;
}
return MoreObjects.toStringHelper(this)
    .add("logId", logId.getId())
    ..add("addressGroup", addressGroupCopy)
    .toString();
```

netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -308,7 +308,7 @@ public LogId getLogId() {
   @Override
   public String toString() {
     return MoreObjects.toStringHelper(this)
-        .add("logId", logId)
+        .add("logId", logId.toString())
```
dapengzhang0: why need `toString()`?

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -390,6 +391,18 @@ public void shutdown(Status reason) {
     }
   }
 
+  @Override
+  public String toString() {
+    Object addressGroupCopy;
+    synchronized (lock) {
+      addressGroupCopy = addressGroup.toString();
```
dapengzhang0: I think you don't need call `toString()` here, EAG is immutable and pretty thread safe to call toString out of lock.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -390,6 +391,18 @@ public void shutdown(Status reason) {
...
+    }
+    return MoreObjects.toStringHelper(this)
+          .add("logId", logId.getId())
+          .add("addressGroup", addressGroupCopy.toString())
```
dapengzhang0: nit: no need to call `toString()` explicitly either, toStringHelper will do that for you.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
zpencer: The field may still be set in one thread but read in another thread, such as the server thread for Channelz.

dapengzhang0: >The field may still be set in one thread but read in another thread, such as the server thread for Channelz.

I don't see any problem with that. When you return the string, the returned value soon becomes stale even if you enclose everything of the toString method with the lock. toString() needn't be in the lock as long as the copying is in the lock
```java
synchronized (lock) {
   addressGroupCopy = addressGroup; 
   // no need to call toString() for a long list of addresses
}
```

zpencer: I see what you mean. Done.

PR #4435, opened 2018-05-03
android: add artifact build

ericgribkoff:
Backport of https://github.com/grpc/grpc-java/pull/4433/


android/build.gradle
```diff
@@ -1,5 +1,7 @@
 apply plugin: 'com.android.library'
 
+group = "io.grpc"
+version = "1.13.0-SNAPSHOT" // CURRENT_GRPC_VERSION
```
zpencer: I think the versions should be changed to 1.12.0-SNAPSHOT  here and below

ericgribkoff: Done, thanks

PR #4437, opened 2018-05-04
stub: Add @CheckReturnValue to AbstractStub

### LINKED ISSUE #3985: Most of the functions in AbstractStub should be annotated with @CheckReturnValue
These functions create new instances rather than changing the internal state, e.g.
```stub.withCallCredentials( ... )```
is wrong. One should always use
```stub = stub.withCallCredentials( ... )```

https://github.com/grpc/grpc-java/blob/c90f27f454f59f15fcd1030be5af8c69b0aad42c/stub/src/main/java/io/grpc/stub/AbstractStub.java#L48

groakley:
No internal google targets were broken by this change.

Fixes #3985


groakley:
Linking this relevant comment about errorprone versions from a similar change here. https://github.com/grpc/grpc-java/pull/4427#issuecomment-386694075

carl-mastrangelo:
@groakley I will try bringing this in, but in case it fails to import we may have to roll back.

PR #4449, opened 2018-05-08
services: encode the Status in the binarylog metadata object

zpencer:
One of the design goals of bin log is to do what a user application
can already do, which implies we stick to public APIs and avoid low
level internals. This does not exactly match the Java API which tries
really hard to hide the h2 implementation details. If we ask the C
folks to denormalize it out to a separate proto field then it won't
quite match the C API. Let's just make Java log something first and
revisit later.


ejona86:
I'm concerned we won't revisit later. What's the advantage of doing this now before the approach is agreed on?

Note that Go also doesn't include `grpc-*` headers in metadata: https://github.com/grpc/grpc-go/blob/v1.11.0/transport/http_util.go#L343-L351 (which is actually stronger than how Java does it).

zpencer:
superseded by https://github.com/grpc/grpc-proto/pull/20

PR #4454, opened 2018-05-09
core: do not use InternalServerInterceptors in BinaryLogProvider

zpencer:
Now that they are in `io.grpc` the accessor is not needed.


ericgribkoff:
Looks like there's still an InternalClientInterceptors reference too

PR #4457, opened 2018-05-11
core: stabilize custom CallOptions API

### LINKED ISSUE #1869: Tracking Issue for Custom CallOptions being Experimental.
Specific usages:
- ServerCalls
- AbstractStub

zpencer:
Deprecate `of()` in favor of `create()` and
`createWithDefault()`. Emphasize that the name string is only for
debug purposes.
    
Fixes #1869

### DIFF
--- a/core/src/main/java/io/grpc/CallOptions.java
+++ b/core/src/main/java/io/grpc/CallOptions.java
@@ -231,13 +231,12 @@ public final class CallOptions {
   /**
    * Key for a key-value pair. Uses reference equality.
    */
-  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1869")
   public static final class Key<T> {
-    private final String name;
+    private final String debugString;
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/grpc/CallOptions.java
```diff
@@ -250,20 +249,50 @@ public T getDefault() {
...
+    /**
+     * Factory method for creating instances of {@link Key}.
+     *
+     * @param debugString a string used to describe this key, used for debugging.
+     * @param defaultValue default value to return when value for key not set
+     * @param <T> Key type
+     * @return Key object
+     * @deprecated Use {@link #create} or {@link #createWithDefault} instead.
+     */
+    @Deprecated
```
ejona86: Without the experimental API, we can't delete this. It needs to be clear this API is going to be deleted.

Issue #1764, opened 2016-05-03
Tracking Issue for Attributes being Experimental.


carl-mastrangelo:
In #1700 it was decided that `Attributes` is still experimental.  This issue is acts the the reference from the source code.

dapengzhang0:
should we un-public `Attributes#keys()`?

friddle:
hello.I meet the problem.I can't get right attributes of by self made attributes.I think it is not appropriate。
I try to get Attributes like code down
`this.attributes.get(Attributes.Key.of("remote-addr"));` and it return null;but to `attributes.toString()` it return "{remote-addr=/127.0.0.1:41064}"

I check the code.find  problem.I think the Key should overrie quals and hashCode two method.
If convinece I can push a simple request.

Sorry for my poor english.

herbyderby:
Attributes.get(Key<T> key) current has @SuppressWarnings("unchecked"). It is actually not safe--Key.of will return a key for any <T> and it is never enforced. (It is generally a bad practice for a method signature to use a type parameter a single time).

I suggest:

- adding Class<T> argument to Key.of
- changing Attributes.get to return key.type.cast(data.get(key); then no suppression will be necessary

(This may be insufficient if there are any keys that use generic types)

ejona86:
@herbyderby, is there a code snippet that would show how it breaks? `Attributes.Key` is the key used for setting/lookup and it uses reference equality. So there's really no way I see to break the Generics.

herbyderby:
For example:

    Key<Integer> ikey = Key.of("mykey");
    Key<String> skey = (Key) ikey;

    Attributes attr = Attributes.newBuilder()
        .set(skey, "foo")
        .build();

    System.err.println("result: " + attr.get(ikey));

Prints "result: foo", even though get here is supposed to return an Integer.

It would be better for attr.get to fail. Currently it is delayed until you actually do something which causes a cast, if at all--methods on Object like equals and toString won't complain.

Admittedly this example requires a raw cast of the key. But given that it is not much work to make Attributes & Key actually typesafe, I think it is worth being a bit more defensive.

ejona86:
This seems like an obvious, "you're going to have a bad time":
```
Key<String> skey = (Key) ikey;
```

That seems no different than, say:
```
List<Integer> ints = ...;
List<String> l = (List) ints;
```

herbyderby:
List is a good example, because Collections.checkedList does exist, for reasons that also apply here:

[...]

PR #4458, opened 2018-05-11
core: stabilize Attributes API

zpencer:
Deprecate static builder method, Keys.of(), add a notice of plans to
deprecate keys(), emphasize that the name is only a debug label.

Related: https://github.com/grpc/grpc-java/issues/1764

### DIFF
--- a/core/src/main/java/io/grpc/Attributes.java
+++ b/core/src/main/java/io/grpc/Attributes.java
@@ -150,6 +150,8 @@ public final class Attributes {
    * equal at one point in time and not equal at another (due to concurrent mutation of attribute
    * values).
    *
+   * <p>This method is not implemented efficiently and is meant for testing.
+   *
    * @param o an object.
    * @return true if the given object is a {@link Attributes} equal attributes.
    */
--- a/core/src/main/java/io/grpc/Attributes.java
+++ b/core/src/main/java/io/grpc/Attributes.java
@@ -29,8 +29,8 @@ import javax.annotation.concurrent.Immutable;
 
 /**
  * An immutable type-safe container of attributes.
+ * @since 1.13.0
  */
-@ExperimentalApi("https://github.com/grpc/grpc-java/issues/1764")
 @Immutable
 public final class Attributes {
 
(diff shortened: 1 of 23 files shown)


core/src/main/java/io/grpc/Attributes.java
```diff
@@ -56,14 +56,24 @@ private Attributes(Map<Key<?>, Object> data) {
    * Returns set of keys stored in container.
    *
    * @return Set of Key objects.
+   * @deprecated This method is being considered for deprecation, if you feel this method is needed
```
ejona86: s/considered for deprecation/considered for removal/ ? Considering this PR _is_ deprecating it.

core/src/main/java/io/grpc/Attributes.java
```diff
@@ -29,8 +29,8 @@
 
 /**
  * An immutable type-safe container of attributes.
+ * @since 1.13.0
  */
-@ExperimentalApi("https://github.com/grpc/grpc-java/issues/1764")
```
ejona86: Since this is being removed, we need some way to denote that the `@Deprecated` things will actually be removed. You could copy this attribute to those deprecated methods, or maybe something similar.

core/src/main/java/io/grpc/Attributes.java
```diff
@@ -56,14 +56,25 @@ private Attributes(Map<Key<?>, Object> data) {
...
 
+  Set<Key<?>> keysForTest() {
+    return Collections.unmodifiableSet(data.keySet());
+  }
+
   /**
    * Create a new builder that is pre-populated with the content from a given container.
+   * @deprecated Use {@link Attributes#toBuilder()} on the {@link Attributes} instance instead.
    */
+  @Deprecated
```
ejona86: This one was also deprecated and needs ExperimentalApi

core/src/main/java/io/grpc/Attributes.java
```diff
@@ -76,32 +87,53 @@ public static Builder newBuilder() {
...
+
+    /**
+     * Factory method for creating instances of {@link Key}.
+     *
+     * @param debugString a string used to describe the key, used for debugging.
+     * @param <T> Key type
+     * @return Key object
+     * @deprecated use {@link #create} instead.
+     */
+    @Deprecated
```
ejona86: And this one.

core/src/main/java/io/grpc/Attributes.java
ejona86: Shoot. I just saw noticed that hashCode does not agree with equals. Let's leave the annotation on this class a bit longer until we get that sorted out.

PR #4461, opened 2018-05-14
core: remove io.grpc core dependency on census

zpencer:
io.grpc (core) does not strictly require census. Move the usages and
and remove census from BUILD.bazel .


ejona86:
CC @ericgribkoff

PR #4462, opened 2018-05-14
core: initialize round robin load balancer picker to random index

tleach:
RoundRobinLoadBalancerFactory creates a new Picker instance every time the set of provided address groups changes or the connection state of subchannels associated with existing address groups changes. In certain scenarios, such as deployment/replacement of the target service cluster, this can lead to high churn of Picker objects. Given that each new Picker's subchannel index is initialized to zero, in these scenarios requests can end up getting disproportionately routed through subchannels (and hence server nodes) which are earlier in the list of address groups.

At Netflix we have measured that some service nodes end up taking 3-4x the load that of other nodes during deployment.

This commit randomizes the start index of the RoundRobinLoadBalancerFactory.Picker which eliminates this behavior.

### DIFF
--- a/core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
+++ b/core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
@@ -49,6 +49,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Queue;
+import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
@@ -110,12 +111,14 @@ public final class RoundRobinLoadBalancerFactory extends LoadBalancer.Factory {
     private final Helper helper;
     private final Map<EquivalentAddressGroup, Subchannel> subchannels =
         new HashMap<EquivalentAddressGroup, Subchannel>();
+    private final Random random;
 
     @Nullable
     private StickinessState stickinessState;
 
     RoundRobinLoadBalancer(Helper helper) {
       this.helper = checkNotNull(helper, "helper");
+      this.random = new Random();
     }
 
     @Override
@@ -211,7 +214,12 @@ public final class RoundRobinLoadBalancerFactory extends LoadBalancer.Factory {
      */
     private void updateBalancingState(ConnectivityState state, Status error) {
       List<Subchannel> activeList = filterNonFailingSubchannels(getSubchannels());
-      helper.updateBalancingState(state, new Picker(activeList, error, stickinessState));
+      // initialize the Picker to a random start index to ensure that a high frequency of Picker
+      // churn does not skew subchannel selection.
+      int startIndex = activeList.isEmpty() ? 0 : random.nextInt(activeList.size());
+      helper.updateBalancingState(
+          state,
+          new Picker(activeList, error, startIndex, stickinessState));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


thelinuxfoundation:

* User @tleach isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

carl-mastrangelo:
@tleach One alternative is to shuffle the addresses in the name resolver before passing it to the LB.  This avoids the problem of all clients overloading the backends in the same order.   Picking a random start index  means all backends would still end up synchronizing (like metronomes).

tleach:
@carl-mastrangelo great point regarding clients synchronizing their access to backends. 

I considered shuffling the addresses but was concerned about introducing an O(n) operation every time the address list is updated if that list changes frequently. Perhaps that concern is overblown.

tleach:
Another thought that occurred to me: shuffling the addresses is likely a redundant operation for other load balancer implementations (e.g. random load balancer). Therefore doing this in the name resolver is probably the wrong place to do it (and arguably leaks information about the choice of load balancing algorithm). It still might make sense to shuffle the address list inside `RoundRobinLoadBalancer.handleResolvedAddressGroups`.

carl-mastrangelo:
@tleach I wouldn't worry about the time complexity here.    The number of backends is usually less than 1000, and list being shuffled very likely fits in the cache. I have an internal benchmark that uses 50,000 backends and didn't see any significant issues.   (Also, see https://github.com/grpc/grpc-java/pull/3579)

As for random load balancer:  probably not a good idea, since some of the backends get way more load than the others. For an example of how this happens see Genius's [blog post](https://genius.com/James-somers-herokus-ugly-secret-annotated)).

For gRPC we try to keep the LB implementations the same across languages, so they all behave roughly equally.  @zhangkun83 and @ejona86 might be able to shed light on potentially applying your change across implementations.

ejona86:
This is probably a good idea to do cross-language. But I'd like to discuss it with Kun, who will be back next week. I do agree that shuffling every update is a bit weird across _all_ NameResolvers (for some it would be a good idea, for others not so much). But I can see the discussion going in a few different directions.

Can we sit on this for a week-ish and get back to you?

tleach:
@ejona86 sure - no urgency on my end. We've shifted to a different (in-house) LB implementation for now. Thanks.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -368,14 +369,17 @@ synchronized Subchannel getSubchannel(String stickinessValue) {
...
-        List<Subchannel> list, @Nullable Status status,
-        @Nullable RoundRobinLoadBalancer.StickinessState stickinessState) {
+    Picker(List<Subchannel> list, @Nullable Status status,
+           @Nullable RoundRobinLoadBalancer.StickinessState stickinessState) {
       this.list = list;
       this.status = status;
       this.stickinessState = stickinessState;
+      // start off with a random address to ensure significant Picker churn does not skew
+      // subchannel selection lower-indexed addresses in the list
+      Random random = new Random();
```
ejona86: Don't re-create the Random instance so frequently. Instead, have it a field of RoundRobinLoadBalancer and probably pass the value (or chosen starting index) into the Picker constructor. Since the Pickers are only created from one thread at a time, there's no concurrency issues from using Random at that point.

tleach:
@zhangkun83 I've updated the pickerRoundRobin() test which now simply passes in a start index and verifies specific subchannels are returned in a specific order as before. This should address your concern about weakening the test.

Given that, in other tests I've preserved the approach of using nextSubchannel() to dynamically determine the next expected subchannel. This approach seems preferable to wiring through a random provider for the start index which introduces a fair amount of boilerplate.

PR #4464, opened 2018-05-15
core,services: Add ChannelTracing data object

dapengzhang0:
Added `ChannelTrace` as an inner class of `Channelz`.

This is in preparation for the implementation of [Channel Tracing](https://github.com/grpc/proposal/blob/master/A3-channel-tracing.md)

### DIFF
--- a/services/src/main/java/io/grpc/services/ChannelzProtoUtil.java
+++ b/services/src/main/java/io/grpc/services/ChannelzProtoUtil.java
@@ -34,6 +34,9 @@ import io.grpc.channelz.v1.ChannelConnectivityState;
 import io.grpc.channelz.v1.ChannelConnectivityState.State;
 import io.grpc.channelz.v1.ChannelData;
 import io.grpc.channelz.v1.ChannelRef;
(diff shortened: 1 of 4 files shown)


services/src/main/java/io/grpc/services/ChannelzProtoUtil.java
```diff
@@ -348,21 +353,49 @@ static Channel toChannel(Instrumented<ChannelStats> channel) {
...
+      builder.setTrace(toChannelTrace(stats.channelTrace));
+    }
+    return builder.build();
   }
 
   static ChannelConnectivityState toChannelConnectivityState(ConnectivityState s) {
     return ChannelConnectivityState.newBuilder().setState(toState(s)).build();
   }
 
+  private static ChannelTrace toChannelTrace(io.grpc.internal.Channelz.ChannelTrace channelTrace) {
```
zpencer: Can we say `Channelz.ChannelTrace` here?

dapengzhang0: Done.

PR #4468, opened 2018-05-16
core: Add ChannelTracer

dapengzhang0:
This is a class similar to CallTracer, to be used for Channel Tracing.

The constructor arg `maxEvents` is a param supposed to be provided by `ManagedChannelBuilder`.

### DIFF
--- a/core/src/test/java/io/grpc/internal/ChannelTracerTest.java
+++ b/core/src/test/java/io/grpc/internal/ChannelTracerTest.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2018 The gRPC Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/grpc/internal/ChannelTracerTest.java
```diff
@@ -0,0 +1,84 @@
...
+public class ChannelTracerTest {
+  @Rule
+  public final ExpectedException thrown = ExpectedException.none();
+
+  @Test
+  public void channelTracerWithZeroMaxEventsShouldThrow() {
+    thrown.expect(IllegalArgumentException.class);
+    thrown.expectMessage("maxEvents must be greater than zero");
+
+    new ChannelTracer(0 /* maxEvents*/, 3L /* channelCreationTimeNanos */);
```
zpencer: nit: use `/*argName=*/` because the linter will check when the label is correct.

dapengzhang0: Done.

PR #4470, opened 2018-05-16
services: differentiate between tcpi_retransmits and tcpi_retrans

zpencer:
These are subtley differently named but distinct fields.


dapengzhang0:
Oh, actually there is one test, and fails :)

PR #4473, opened 2018-05-16
all: TimeProvider to use nanos rather than millis more consistently

dapengzhang0:
This is the same practice as #2833
More PRs switching to nanos may come after.


dapengzhang0:
Thanks @zpencer for the review!

PR #4479, opened 2018-05-17
stub: stablize AbstractStub.withExecutor API

### LINKED ISSUE #3605: Tracking AbstractStub.withExecutor being experimental

dapengzhang0:
Resolves #3605


ejona86:
I see two uses internally: one that uses it to set directExecutor on a blocking stub for testing which does little, and another that creates a Channel in the same method and so could have specified the executor there instead. Those don't really support this API.

dapengzhang0:
As per the API Stabilization Notes 2017-12-14, "Stabilize, even without users, since the API is a copy of what’s on CallOptions"

PR #4483, opened 2018-05-18
core: add unit test for channel shutdown during retry backoff

dapengzhang0:


dapengzhang0:
Thanks for review. Rebased imports and merging.

PR #4485, opened 2018-05-21
core: fix client does not detect truncated message

### LINKED ISSUE #3264: Client does not detect a truncated message at the end of stream
As discovered during the course of #3145. The server does detect and fail in the case of truncated message frame ("Encountered end-of-stream mid-frame"), but client doesn't have such logic.

dapengzhang0:
Resolves #3264


dapengzhang0:
Thanks @ejona86

PR #4486, opened 2018-05-22
 auth: Require PRIVACY_AND_INTEGRITY for GoogleCredentials

ejona86:
This keeps them more secure. Other types of creds are left as-is, since
we don't quite know if it makes sense to have a similar restriction. (It
likely does make sense, but this is a more precise change for our
needs.)

I will keep the two separate commits when merging. Each commit has a commit message, although the two commits are pretty disjoint, so it's not a big deal to review it all at once.

### DIFF
--- a/auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
+++ b/auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
@@ -26,6 +26,7 @@ import io.grpc.Attributes;
 import io.grpc.CallCredentials;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
+import io.grpc.SecurityLevel;
 import io.grpc.Status;
 import io.grpc.StatusException;
 import java.io.IOException;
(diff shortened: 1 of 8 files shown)


auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
```diff
@@ -77,6 +90,14 @@ public void thisUsesUnstableApi() {}
   @Override
   public void applyRequestMetadata(MethodDescriptor<?, ?> method, Attributes attrs,
       Executor appExecutor, final MetadataApplier applier) {
+    SecurityLevel security = checkNotNull(attrs.get(ATTR_SECURITY_LEVEL), "securityLevel");
+    if (requirePrivacy && security != SecurityLevel.PRIVACY_AND_INTEGRITY) {
+      applier.fail(Status.UNAVAILABLE
```
zhangkun83: Should it be "FAILED_PRECONDITION"? This failure is not retry-able, while "UNAVAILABLE" implies retry-ability.

dapengzhang0: "FAILED_PRECONDITION" does not imply non-retry-ability, and "UNAVAILABLE" does not always imply retry-ability, all depends on retry policy in service config. To disallow retry explicitly, send "grpc-retry-pushback-ms" trailer metadata with "-1".

ejona86: FAILED_PRECONDITION is currently black-listed for our use in https://github.com/grpc/grpc/blob/master/doc/statuscodes.md .

I don't disagree that UNAVAILABLE is not perfect, but I don't know what's better. If you're in this situation though, you're pretty hosed independent of the status code.

PR #4487, opened 2018-05-22
protobuf: add import dropped on merge

carl-mastrangelo:


carl-mastrangelo:
not entirely sure why this is failing the Android / Linux artifacts build.

ejona86:
The Linux one looks like a flake. The Android failure doesn't make any sense.

PR #4494, opened 2018-05-22
core: Update OpenCensus version. 

songy23:
Related to https://github.com/census-instrumentation/opencensus-java/issues/1207.

OpenCensus has updated the RPC constants in [Specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md) recently, and in OpenCensus-Java these new constants are added since v0.13.0. This PR updates gRPC to use the latest OpenCensus version, ~~and record stats against the new RPC constants.~~

~~Note that gRPC also needs to record stats against the deprecated constants, to keep backwards compatibility.~~

/cc @bogdandrutu @dinooliva @sebright

### DIFF
--- a/core/src/main/java/io/grpc/internal/CensusStatsModule.java
+++ b/core/src/main/java/io/grpc/internal/CensusStatsModule.java
@@ -65,7 +65,6 @@ import javax.annotation.Nullable;
  * starts earlier than the ServerCall.  Therefore, only one tracer is created per stream/call and
  * it's the tracer that reports the summary to Census.
  */
-@SuppressWarnings("deprecation")
 public final class CensusStatsModule {
   private static final Logger logger = Logger.getLogger(CensusStatsModule.class.getName());
   private static final double NANOS_PER_MILLI = TimeUnit.MILLISECONDS.toNanos(1);
@@ -329,14 +328,14 @@ public final class CensusStatsModule {
       TagValue methodTag = TagValue.create(fullMethodName);
       this.startCtx =
           module.tagger.toBuilder(parentCtx)
-          .put(RpcMeasureConstants.RPC_METHOD, methodTag)
+          .put(DeprecatedCensusConstants.RPC_METHOD, methodTag)
           .put(RpcMeasureConstants.GRPC_CLIENT_METHOD, methodTag)
           .build();
       this.stopwatch = module.stopwatchSupplier.get().start();
       this.recordFinishedRpcs = recordFinishedRpcs;
       if (recordStartedRpcs) {
         module.statsRecorder.newMeasureMap()
-            .put(RpcMeasureConstants.RPC_CLIENT_STARTED_COUNT, 1)
+            .put(DeprecatedCensusConstants.RPC_CLIENT_STARTED_COUNT, 1)
             .put(RpcMeasureConstants.GRPC_CLIENT_STARTED_RPCS, 1)
             .record(startCtx);
       }
@@ -394,18 +393,20 @@ public final class CensusStatsModule {
       }
       MeasureMap measureMap = module.statsRecorder.newMeasureMap()
           // TODO(songya): remove the deprecated measure constants once they are completed removed.
-          .put(RpcMeasureConstants.RPC_CLIENT_FINISHED_COUNT, 1)
+          .put(DeprecatedCensusConstants.RPC_CLIENT_FINISHED_COUNT, 1)
           // The latency is double value
-          .put(RpcMeasureConstants.RPC_CLIENT_ROUNDTRIP_LATENCY, roundtripNanos / NANOS_PER_MILLI)
-          .put(RpcMeasureConstants.RPC_CLIENT_REQUEST_COUNT, tracer.outboundMessageCount)
-          .put(RpcMeasureConstants.RPC_CLIENT_RESPONSE_COUNT, tracer.inboundMessageCount)
-          .put(RpcMeasureConstants.RPC_CLIENT_REQUEST_BYTES, tracer.outboundWireSize)
-          .put(RpcMeasureConstants.RPC_CLIENT_RESPONSE_BYTES, tracer.inboundWireSize)
           .put(
-              RpcMeasureConstants.RPC_CLIENT_UNCOMPRESSED_REQUEST_BYTES,
+              DeprecatedCensusConstants.RPC_CLIENT_ROUNDTRIP_LATENCY,
+              roundtripNanos / NANOS_PER_MILLI)
+          .put(DeprecatedCensusConstants.RPC_CLIENT_REQUEST_COUNT, tracer.outboundMessageCount)
+          .put(DeprecatedCensusConstants.RPC_CLIENT_RESPONSE_COUNT, tracer.inboundMessageCount)
+          .put(DeprecatedCensusConstants.RPC_CLIENT_REQUEST_BYTES, tracer.outboundWireSize)
+          .put(DeprecatedCensusConstants.RPC_CLIENT_RESPONSE_BYTES, tracer.inboundWireSize)
+          .put(
+              DeprecatedCensusConstants.RPC_CLIENT_UNCOMPRESSED_REQUEST_BYTES,
               tracer.outboundUncompressedSize)
           .put(
-              RpcMeasureConstants.RPC_CLIENT_UNCOMPRESSED_RESPONSE_BYTES,
+              DeprecatedCensusConstants.RPC_CLIENT_UNCOMPRESSED_RESPONSE_BYTES,
               tracer.inboundUncompressedSize)
 
           // New RPC measure constants.
@@ -418,14 +419,14 @@ public final class CensusStatsModule {
           .put(RpcMeasureConstants.GRPC_CLIENT_SENT_BYTES_PER_RPC, tracer.outboundWireSize)
           .put(RpcMeasureConstants.GRPC_CLIENT_RECEIVED_BYTES_PER_RPC, tracer.inboundWireSize);
       if (!status.isOk()) {
-        measureMap.put(RpcMeasureConstants.RPC_CLIENT_ERROR_COUNT, 1);
+        measureMap.put(DeprecatedCensusConstants.RPC_CLIENT_ERROR_COUNT, 1);
       }
       TagValue statusTag = TagValue.create(status.getCode().toString());
       measureMap.record(
           module
               .tagger
               .toBuilder(startCtx)
-              .put(RpcMeasureConstants.RPC_STATUS, statusTag)
+              .put(DeprecatedCensusConstants.RPC_STATUS, statusTag)
               .put(RpcMeasureConstants.GRPC_CLIENT_STATUS, statusTag)
               .build());
     }
@@ -518,7 +519,7 @@ public final class CensusStatsModule {
       this.recordFinishedRpcs = recordFinishedRpcs;
       if (recordStartedRpcs) {
         module.statsRecorder.newMeasureMap()
-            .put(RpcMeasureConstants.RPC_SERVER_STARTED_COUNT, 1)
+            .put(DeprecatedCensusConstants.RPC_SERVER_STARTED_COUNT, 1)
             .put(RpcMeasureConstants.GRPC_SERVER_STARTED_RPCS, 1)
             .record(parentCtx);
       }
@@ -609,15 +610,21 @@ public final class CensusStatsModule {
       long elapsedTimeNanos = stopwatch.elapsed(TimeUnit.NANOSECONDS);
       MeasureMap measureMap = module.statsRecorder.newMeasureMap()
           // TODO(songya): remove the deprecated measure constants once they are completed removed.
-          .put(RpcMeasureConstants.RPC_SERVER_FINISHED_COUNT, 1)
+          .put(DeprecatedCensusConstants.RPC_SERVER_FINISHED_COUNT, 1)
           // The latency is double value
-          .put(RpcMeasureConstants.RPC_SERVER_SERVER_LATENCY, elapsedTimeNanos / NANOS_PER_MILLI)
-          .put(RpcMeasureConstants.RPC_SERVER_RESPONSE_COUNT, outboundMessageCount)
-          .put(RpcMeasureConstants.RPC_SERVER_REQUEST_COUNT, inboundMessageCount)
-          .put(RpcMeasureConstants.RPC_SERVER_RESPONSE_BYTES, outboundWireSize)
-          .put(RpcMeasureConstants.RPC_SERVER_REQUEST_BYTES, inboundWireSize)
-          .put(RpcMeasureConstants.RPC_SERVER_UNCOMPRESSED_RESPONSE_BYTES, outboundUncompressedSize)
-          .put(RpcMeasureConstants.RPC_SERVER_UNCOMPRESSED_REQUEST_BYTES, inboundUncompressedSize)
+          .put(
+              DeprecatedCensusConstants.RPC_SERVER_SERVER_LATENCY,
+              elapsedTimeNanos / NANOS_PER_MILLI)
+          .put(DeprecatedCensusConstants.RPC_SERVER_RESPONSE_COUNT, outboundMessageCount)
+          .put(DeprecatedCensusConstants.RPC_SERVER_REQUEST_COUNT, inboundMessageCount)
+          .put(DeprecatedCensusConstants.RPC_SERVER_RESPONSE_BYTES, outboundWireSize)
+          .put(DeprecatedCensusConstants.RPC_SERVER_REQUEST_BYTES, inboundWireSize)
+          .put(
+              DeprecatedCensusConstants.RPC_SERVER_UNCOMPRESSED_RESPONSE_BYTES,
+              outboundUncompressedSize)
+          .put(
+              DeprecatedCensusConstants.RPC_SERVER_UNCOMPRESSED_REQUEST_BYTES,
+              inboundUncompressedSize)
 
           // New RPC measure constants.
           .put(RpcMeasureConstants.GRPC_SERVER_SERVER_LATENCY, elapsedTimeNanos / NANOS_PER_MILLI)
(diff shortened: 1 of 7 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1891,22 +1897,36 @@ private void assertServerStatsTrace(String method, Status.Code code,
     }
   }
 
-  private static void checkStartTags(MetricsRecord record, String methodName) {
+  private static void checkStartTags(MetricsRecord record, String methodName, boolean isClient) {
```
sebright: Suggestion: Since there are many calls to checkStartTags and checkEndTags that are far from the definition, it might help to use an enum with CLIENT and SERVER values instead of a boolean.

songy23: Make sense to me, though there's [other method](https://github.com/grpc/grpc-java/blob/master/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java#L1948) in this class that already used a similar boolean check...

core/src/main/java/io/grpc/internal/CensusStatsModule.java
```diff
@@ -329,7 +330,9 @@ public void outboundMessage(int seqNo) {
       this.parentCtx = checkNotNull(parentCtx);
       this.startCtx =
           module.tagger.toBuilder(parentCtx)
-          .put(RpcMeasureConstants.RPC_METHOD, TagValue.create(fullMethodName)).build();
+          .put(RpcMeasureConstants.RPC_METHOD, TagValue.create(fullMethodName))
+          .put(RpcMeasureConstants.GRPC_CLIENT_METHOD, TagValue.create(fullMethodName))
```
zhangkun83: You probably want to create the TagValue once and re-use it.

core/src/main/java/io/grpc/internal/CensusStatsModule.java
```diff
@@ -410,6 +424,9 @@ void callEnded(Status status) {
               .tagger
               .toBuilder(startCtx)
               .put(RpcMeasureConstants.RPC_STATUS, TagValue.create(status.getCode().toString()))
+              .put(
+                  RpcMeasureConstants.GRPC_CLIENT_STATUS,
+                  TagValue.create(status.getCode().toString()))
```
zhangkun83: ditto

core/src/main/java/io/grpc/internal/CensusStatsModule.java
```diff
@@ -609,6 +634,9 @@ public void streamClosed(Status status) {
               .tagger
               .toBuilder(parentCtx)
               .put(RpcMeasureConstants.RPC_STATUS, TagValue.create(status.getCode().toString()))
+              .put(
+                  RpcMeasureConstants.GRPC_SERVER_STATUS,
+                  TagValue.create(status.getCode().toString()))
```
zhangkun83: ditto

core/src/main/java/io/grpc/internal/CensusTracingModule.java
```diff
@@ -211,9 +211,10 @@ private static EndSpanOptions createEndSpanOptions(
   }
 
   private static void recordNetworkEvent(
-      Span span, NetworkEvent.Type type,
+      Span span, io.opencensus.trace.NetworkEvent.Type type,
```
zhangkun83: Is it not the time yet to switch to the new class MessageEvent?

core/src/main/java/io/grpc/internal/CensusStatsModule.java
```diff
@@ -389,6 +392,7 @@ void callEnded(Status status) {
         tracer = BLANK_CLIENT_TRACER;
       }
       MeasureMap measureMap = module.statsRecorder.newMeasureMap()
+          // TODO(songya): remove the deprecated measure constants once they are completed removed.
```
zhangkun83: How is the migration handled on the consumer side in the Census implementation? For not to break the interoperability between gRPC and Census inside Google, it seems necessary for Census to read both the old and the new measurements, giving the new ones precedence. If that's the case, we could just remove the old measurements here.

core/src/main/java/io/grpc/internal/CensusStatsModule.java
songy23: Done.

core/src/main/java/io/grpc/internal/CensusStatsModule.java
songy23: Unfortunately, inside Google the new measurements are not completely consistent with Census Java, and only the old measurements will be read at this moment. Even for open source, users could still export and read the old measurements so I think it's important to continue recording stats for the old ones.

There are some ongoing discussion about making the new measurements consistent with Census Java. /cc @dinooliva

core/src/main/java/io/grpc/internal/CensusStatsModule.java
songy23: Done.

core/src/main/java/io/grpc/internal/CensusStatsModule.java
songy23: Done.

core/src/main/java/io/grpc/internal/CensusTracingModule.java
songy23: I would prefer to do that in a separate PR since it's unrelated to RPC constants. Also, I assume we will still need to record for both `NetworkEvent` and `MessageEvent` to be backwards compatible? @bogdandrutu @hailongwen

core/src/main/java/io/grpc/internal/CensusStatsModule.java
sebright: The artifact providing the gRPC views has [separate methods for registering the old and new views](https://github.com/census-instrumentation/opencensus-java/blob/07ad48d7c3031efcdd6ac3c8932cabe6ac3fff5b/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViews.java#L130-L152), so users can choose when to upgrade them.  It might be easier for users to upgrade if they could get data for both sets of views at the same time.

songy23:
Just found I forgot to update `repositories.bzl`. @ericgribkoff could you please run the kokoro build again?

core/src/main/java/io/grpc/internal/CensusTracingModule.java
zhangkun83: @sebright, do you mean old versions of opencensus-java implementations? So presumably, the new Census impl will treat NetworkEvent and MessageEvent as equivalent, then gRPC can switch to MessageEvent after all your customers has upgraded to the new impl.

core/src/main/java/io/grpc/internal/CensusStatsModule.java
zhangkun83: Is it possible to first change Census impls to convert the old measurements to the new ones, then switch gRPC to record the new measurements after all your customers have switched to the new impl and the new measurements? Then the migration of the measurements can go in line with the migration from NetworkEvent to MessageEvent (see my comment on that). Plus the gRPC code (esp. the tests) would be much cleaner if it ever has to deal with one API.

zhangkun83:
> Since we're adding more tags to the context, @sebright proposed maybe we need filtering on serialized tags when serializing contexts. We didn't do this before and wondered if the additional tags could be a problem.

Oh, I thought the method name and status tags were *not* propagated on the via in the first place. Is it not the case?

core/src/main/java/io/grpc/internal/CensusTracingModule.java
sebright: > @sebright, do you mean old versions of opencensus-java implementations?

I meant old versions of opencensus-api.  I don't think that the version of the OpenCensus implementation should matter to gRPC, as long as it is also compatible with the version of opencensus-api that gRPC uses.

The opencensus-java implementations still only reference NetworkEvent, and the API does the conversion.

core/src/main/java/io/grpc/internal/CensusStatsModule.java
sebright: I'm not sure, but I thought that we wanted users to explicitly switch to the new views, since the change isn't transparent like the switch to MessageEvent.  For example, the stats may be exported to a service that uses the view or measure names.

/cc @bogdandrutu

songy23: Yes, like @sebright said, OpenCensus users could be exporting the old RPC constants to other backends and using them for alerting rules. Since the new and old RPC constants have different names, they will be different metrics on other monitoring systems (Stackdriver, Prometheus). We would like to allow users to explicitly choose old or new RPC constants and update their monitoring accordingly, as in our APIs ([registerAllGrpcViews](https://github.com/census-instrumentation/opencensus-java/blob/master/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViews.java#L135) vs [registerAllViews](https://github.com/census-instrumentation/opencensus-java/blob/master/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViews.java#L195)). So I think it's import to have stats for both of them.

songy23:
Besides, it looks like the failed Kokoro Macos build is because the testing machine cannot connect to gradle host:

```
./gradlew clean -PtargetArch=x86_64 -Pcheckstyle.ignoreFailures=false -PfailOnWarnings=true -PerrorProne=true -Dorg.gradle.parallel=true

Downloading https://services.gradle.org/distributions/gradle-4.7-bin.zip

Exception in thread "main" java.net.NoRouteToHostException: No route to host (Host unreachable)
...
```

core/src/main/java/io/grpc/internal/CensusStatsModule.java
songy23: On top of that, as I mentioned in the previous comments, currently it's particularly difficult to map the old RPC constants to the new ones inside Google. For example, there are no *_count measures in the new constants.

zhangkun83:
> Server-method, server-status and client-status are non-propagating. The only thing we're not sure is client-method. From our interop testing it looks like client-method will be propagated to server.

What's the benefit of propagating client-method? Presumably it would be the same as server-method.

core/src/main/java/io/grpc/internal/CensusStatsModule.java
zhangkun83: I see you don't have the "started count" measurements in the new constants, but it is a required metric inside Google.

songy23: We removed started_count, finished_count and error_count in the new [specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md). There are some rationale behind why [error_count](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md#why-no-error-counts) and [finished_count](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md#why-are-completed_rpcs-views-defined-over-latency-measures) gets removed, but I cannot remember why we removed started_count.

@bogdandrutu @ramonza should we consider putting back the started_count measure since it's required?

songy23:
> What's the benefit of propagating client-method? Presumably it would be the same as server-method.

Oh sorry I got this wrong at first. Client-method is actually non-propagating, but server would have method in the context. From Dino's comments in https://github.com/census-ecosystem/opencensus-experiments/issues/18#issuecomment-376732523:

> The grpc client doesn't need to propagate the method tag but the grpc server does need to add it to the context where the rpc is being processed so there should be a method tag in scope on the server side.

> Also, if the grpc server subsequently makes a call as a grpc client, then the method tag will be propagated since it is in scope (I don't believe gRPC has any logic for explicitly removing the method tag).

And we're wondering if [ServerTracerFactory.filterContext()](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/CensusStatsModule.java#L616) needs to be updated after using the new `TagKey`s.

zhangkun83:
> Oh sorry I got this wrong at first. Client-method is actually non-propagating, but server would have method in the context. From Dino's comments in census-ecosystem/opencensus-experiments#18 (comment):
>> The grpc client doesn't need to propagate the method tag but the grpc server does need to add it to the context where the rpc is being processed so there should be a method tag in scope on the server side.

>> Also, if the grpc server subsequently makes a call as a grpc client, then the method tag will be propagated since it is in scope (I don't believe gRPC has any logic for explicitly removing the method tag).

> And we're wondering if ServerTracerFactory.filterContext() needs to be updated after using the new TagKeys.

gRPC puts `RPC_METHOD` (also `GRPC_SERVER_METHOD` as added in this PR) in the `parentCtx` in `CensusStatsModule.ServerTracerFactory.newServerStreamTracer`. This is already doing what @dinooliva was describing. `filterContext()` doesn't need to be changed.

songy23:
https://github.com/grpc/grpc-java/pull/4530 has updated gRPC to use `MessageEvent` instead of `NetworkEvent`, and has been submitted. Rebased again that PR (upgrade from `0.12.3` to `0.13.2`).

Please hold off this PR until I add started_rpcs to OC Java (https://github.com/census-instrumentation/opencensus-java/pull/1230) and make a `0.14.0` release. I'll update this PR after that.

core/src/main/java/io/grpc/internal/CensusStatsModule.java
songy23: I added *_started_rpcs constants to OpenCensus `v0.14.0`, and upgraded the version here. Now the started_rpcs are recorded. PTAL at the two latest commits.

sebright:
LGTM

songy23:
There was some concern about the performance regression caused by this PR so I held off merging it. By the last time I ran the benchmarks, there's still ~3% performance regression introduced by this PR.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -140,6 +140,7 @@
  *
  * <p> New tests should avoid using Mockito to support running on AppEngine.</p>
  */
+@SuppressWarnings("deprecation")
```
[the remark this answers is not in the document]
zhangkun83: The `RpcMeasureConstants.RPC_*` constants are deprecated.
Alternative is to create a holder class, e.g., `DeprecatedCensusMeasureConstants` in gRPC that define the same constants that points to the ones from Census. We can then have a suppression on the holder class. `CensusStatsModule` and `CensusModulesTest` will use the constants from that holder class, and won't need the suppression.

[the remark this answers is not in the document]
ejona86: Could we copy them to class-local constants and just put the @SuppressWarnings on those constant copies?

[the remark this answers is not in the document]
ejona86: Holder class would also be fine.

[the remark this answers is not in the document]
songy23: SGTM, I'll update this PR to use the holder approach as well as running the performance benchmarks again.

core/src/main/java/io/grpc/internal/DeprecatedCensusConstants.java
```diff
@@ -0,0 +1,80 @@
...
+  public static final MeasureLong RPC_SERVER_STARTED_COUNT =
+      RpcMeasureConstants.RPC_SERVER_STARTED_COUNT;
+  public static final MeasureLong RPC_SERVER_FINISHED_COUNT =
+      RpcMeasureConstants.RPC_SERVER_FINISHED_COUNT;
+  public static final MeasureLong RPC_SERVER_REQUEST_COUNT =
+      RpcMeasureConstants.RPC_SERVER_REQUEST_COUNT;
+  public static final MeasureLong RPC_SERVER_RESPONSE_COUNT =
+      RpcMeasureConstants.RPC_SERVER_RESPONSE_COUNT;
+
+  DeprecatedCensusConstants() {}
```
zhangkun83: Make it private.

songy23: Updated.

songy23:
So sorry I didn't update this thread for a long time. The most recent benchmarks still showed performance regression if we were to record both the new and old constants. For now I remove the new constants from this PR and just update the OpenCensus version (and fix deprecation warnings). Please help me submit this PR first to avoid blocking https://github.com/census-instrumentation/opencensus-java/pull/1563. @zhangkun83 

Moving forward about the new constants, I think at some point we should just switch to the new ones and completely get rid of the old constants, both internally and in open source. That will be a breaking change, but even with our initial plan (which is to record both new and old constants for some time and then remove the old constants) there will be a point when we need to make the same breaking change.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1918,22 +1926,27 @@ private void assertServerStatsTrace(String method, Status.Code code,
     }
   }
 
-  private static void checkStartTags(MetricsRecord record, String methodName) {
+  private static void checkStartTags(MetricsRecord record, String methodName, Type type) {
```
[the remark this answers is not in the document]
songy23: Correct, removed.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1918,22 +1926,27 @@ private void assertServerStatsTrace(String method, Status.Code code,
...
-    assertNotNull("method name tagged", methodNameTag);
-    assertEquals("method names match", methodName, methodNameTag.asString());
+    TagValue methodNameTagOld = record.tags.get(DeprecatedCensusConstants.RPC_METHOD);
+    assertNotNull("method name tagged", methodNameTagOld);
+    assertEquals("method names match", methodName, methodNameTagOld.asString());
   }
 
   private static void checkEndTags(
-      MetricsRecord record, String methodName, Status.Code status) {
+      MetricsRecord record, String methodName, Status.Code status, Type type) {
```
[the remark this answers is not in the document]
songy23: Done.

songy23:
Thanks!

PR #4497, opened 2018-05-23
core: fix Attributes hashCode

### LINKED ISSUE #4459: Attributes.equals disagrees with hashCode
```diff
diff --git a/core/src/test/java/io/grpc/AttributesTest.java b/core/src/test/java/io/grpc/AttributesTest.java
index 45319c8f0..adb94a458 100644
--- a/core/src/test/java/io/grpc/AttributesTest.java
+++ b/core/src/test/java/io/grpc/AttributesTest.java
@@ -75,5 +75,6 @@ public class AttributesTest {
     Attributes attr2 = Attributes.newBuilder().set(key, v2).build();
 
     assertEquals(attr1, attr2);
+    assertEquals(attr1.hashCode(), attr2.hashCode());
   }
 }
```

```
io.grpc.AttributesTest > valueEquality FAILED
    java.lang.AssertionError: expected:<244492244> but was:<1056354025>
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at io.grpc.AttributesTest.valueEquality(AttributesTest.java:78)
```

We need to calculate hashCode manually.

dapengzhang0:
Resolves #4459

### DIFF
--- a/core/src/main/java/io/grpc/Attributes.java
+++ b/core/src/main/java/io/grpc/Attributes.java
@@ -189,7 +189,11 @@ public final class Attributes {
    */
   @Override
   public int hashCode() {
-    return data.hashCode();
+    int hashCode = 0;
+    for (Object o : data.values()) {
+      hashCode += o.hashCode();
+    }
+    return hashCode;
   }
 
   /**
--- a/core/src/main/java/io/grpc/Attributes.java
+++ b/core/src/main/java/io/grpc/Attributes.java
@@ -190,8 +190,8 @@ public final class Attributes {
   @Override
   public int hashCode() {
     int hashCode = 0;
-    for (Object o : data.values()) {
-      hashCode += o.hashCode();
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/Attributes.java
```diff
@@ -189,7 +189,11 @@ public boolean equals(Object o) {
    */
   @Override
   public int hashCode() {
-    return data.hashCode();
+    int hashCode = 0;
+    for (Object o : data.values()) {
+      hashCode += o.hashCode();
```
ejona86: It seems you should include the key hash code as well.

dapengzhang0: Done.

PR #4506, opened 2018-05-24
stub: remove static Method descriptors and stabilize method accessors

### LINKED ISSUE #1901: Stabilize generated MethodDescriptor API
Today, the generated method descriptors are available as fields. This should probably be changed to be behind a method of some sort so that they can be generated lazily (to reduce static class loading time for Android).

carl-mastrangelo:
Fixes #1901

cc: @ericgribkoff


carl-mastrangelo:
@ejona86 should I wait to submit this till the next API review?

ejona86:
It'd be nice. I had originally thought you'd just delete the Deprecated API in a PR, and then after the meeting stabilize. But with this PR as-is, waiting until after the meeting is probably nicest. I'd not be _too_ bothered if it got submitted as-is, but I don't see much reason to avoid the current process.

PR #4507, opened 2018-05-24
netty,okhttp: make rpc followed by racy transport shutdown transparent-retry-able

dapengzhang0:
A new RPC starts with the following steps:

1. Pick a READY transport
2. the READY transport calls `transport.newStream()`
3. the new stream calls `stream.start()`
4. `stream.start()` invokes or enqueus `writeHeaders()` (or for GET request, noop)

A racy transport shutdown could happen between 3 and 4, and by the retry spec, the RPC should be transparent-retry-able in this case. For Netty and OkHttp transport implementation, before step 4, (even if step 1, 2, and 3 excluding 4 are made atomic,) the http2-stream for the RPC is not created, so the current transparent retry logic does not apply and need fix.

Of course, if step 1, 2, and 3 including 4 are made atomic, and not with GET, there will be no such problem.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
@@ -460,6 +460,8 @@ class NettyClientHandler extends AbstractNettyHandler {
           throws Exception {
     if (lifecycleManager.getShutdownThrowable() != null) {
       // The connection is going away, just terminate the stream now.
+      command.stream().transportReportStatus(
+          lifecycleManager.getShutdownStatus(), RpcProgress.REFUSED, true, new Metadata());
       promise.setFailure(lifecycleManager.getShutdownThrowable());
       return;
     }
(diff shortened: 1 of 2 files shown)


dapengzhang0:
cc @zhangkun83

netty/src/main/java/io/grpc/netty/NettyClientHandler.java
```diff
@@ -460,6 +460,8 @@ private void createStream(CreateStreamCommand command, final ChannelPromise prom
           throws Exception {
     if (lifecycleManager.getShutdownThrowable() != null) {
       // The connection is going away, just terminate the stream now.
+      command.stream().transportReportStatus(
```
ejona86: Add a comment that this is really the GOAWAY case.

ejona86:
For posterity: this really only fixes the case when the transport receives a GOAWAY, racing with a new RPC. Existing RPCs would be REFUSED, but that new RPC previously would not have been (it would have just failed).

This _could_ address racy `transport.shutdown()` via the LB, however we are currently employing a hack that will delaying (via timer) the `shutdown()` until it is highly unlikely any new RPCs will try to use the soon-to-be-shutdown transport. As long as that hack is in place, the `transport.shutdown()` race doesn't happen at the transport level and so this change won't impact it.

PR #4510, opened 2018-05-25
netty: add listener prior to enqueuing CreateStreamCommand

dapengzhang0:
This may let transport report status a little more efficiently if CreateSreamCommand fails.


dapengzhang0:
Offline discussed with @ejona86 and this is not the right behavior. Closing.

PR #4517, opened 2018-05-30
services: Use RE2/J

Capstan:
RE2/J ensures linear time matching, and as such is preferred over `java.util.regex` for safety in the mono repo. While currently safe, this conversion future-proofs any `Pattern`s in BinlogHelper, and removes the need to maintain an exception for gRPC wrt. use of regexes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -218,6 +218,7 @@ subprojects {
                 netty_tcnative: 'io.netty:netty-tcnative-boringssl-static:2.0.8.Final',
 
                 conscrypt: 'org.conscrypt:conscrypt-openjdk-uber:1.0.1',
+                re2j: 'com.google.re2j:re2j:1.2',
 
                 // Test dependencies.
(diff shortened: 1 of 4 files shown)


carl-mastrangelo:
cc: @zpencer & @ericgribkoff 

I like Re2, though it comes as an extra dependency, which we try to avoid.  If there is a chance binlogging is used on Android, it would not be okay to depend on another library I think.

ericgribkoff:
No issues for Android, as the services package isn't pulled in by default. Not sure if binary logging itself is even intended to work on Android, but if an app added a dependency on services, Re2j supports Android as well, so this change wouldn't alter the existing functionality.

dapengzhang0:
link to internal issue b/80455946

zpencer:
Android support is not something called out in the binlog design, but it sounds like this change won't make it any harder to add support in the future. Thanks for this PR @Capstan !

PR #4519, opened 2018-05-31
netty: Remove option to pass promise to WriteQueue

ejona86:
Passing a promise to WriteQueue was only misused to add a listener on
the promise before issuing the write. Although in this case the listener
ordering will be "random" because listeners are being added from two
different threads, in general we always want to add a listener after the
write returns to let any lower-level listeners be registered first.

Future work can resolve the "random" listener order by passing the
listener to the WriteQueue and adding the listener from the event loop.

### DIFF
--- a/netty/src/test/java/io/grpc/netty/NettyClientStreamTest.java
+++ b/netty/src/test/java/io/grpc/netty/NettyClientStreamTest.java
@@ -30,7 +30,6 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.anyBoolean;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Matchers.isA;
 import static org.mockito.Matchers.same;
@@ -58,6 +57,7 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelPromise;
+import io.netty.channel.DefaultChannelPromise;
 import io.netty.handler.codec.http2.DefaultHttp2Headers;
 import io.netty.handler.codec.http2.Http2Headers;
 import io.netty.util.AsciiString;
@@ -180,7 +180,6 @@ public class NettyClientStreamTest extends NettyStreamTestBase<NettyClientStream
     stream.flush();
     verify(writeQueue).enqueue(
         eq(new SendGrpcFrameCommand(stream.transportState(), messageFrame(MESSAGE), false)),
-        any(ChannelPromise.class),
         eq(true));
   }
 
@@ -196,12 +195,10 @@ public class NettyClientStreamTest extends NettyStreamTestBase<NettyClientStream
     verify(writeQueue).enqueue(
             eq(new SendGrpcFrameCommand(
                 stream.transportState(), messageFrame(MESSAGE).slice(0, 5), false)),
-            any(ChannelPromise.class),
             eq(false));
     verify(writeQueue).enqueue(
         eq(new SendGrpcFrameCommand(
             stream.transportState(), messageFrame(MESSAGE).slice(5, 11), false)),
-        any(ChannelPromise.class),
         eq(true));
   }
 
@@ -436,7 +433,10 @@ public class NettyClientStreamTest extends NettyStreamTestBase<NettyClientStream
     metadata.put(GrpcUtil.USER_AGENT_KEY, "bad agent");
     listener = mock(ClientStreamListener.class);
     Mockito.reset(writeQueue);
-    when(writeQueue.enqueue(any(QueuedCommand.class), any(boolean.class))).thenReturn(future);
+    ChannelPromise completedPromise = new DefaultChannelPromise(channel)
+        .setSuccess();
+    when(writeQueue.enqueue(any(QueuedCommand.class), any(boolean.class)))
+        .thenReturn(completedPromise);
 
     stream = new NettyClientStream(
         new TransportStateImpl(handler, DEFAULT_MAX_MESSAGE_SIZE),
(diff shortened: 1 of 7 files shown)


carl-mastrangelo:
I would need to be convinced that this was "misused" and the order "random".  I'm pretty sure there is a HB  between the promise having the first addListener() call, and the subsequent listeners being added (via. push and pop on the CLQ , and  MPSC queue).

From my reading, your change does the exact oppose of what the desc says?

ejona86:
Misused is simply what what happening before: adding a listener before calling the write. I was saying the _new_ order was random. It seems we're reading "in this case" differently; I can clarify.

netty/src/test/java/io/grpc/netty/NettyStreamTestBase.java
```diff
@@ -95,14 +88,16 @@
...
     when(channel.alloc()).thenReturn(UnpooledByteBufAllocator.DEFAULT);
     when(channel.pipeline()).thenReturn(pipeline);
     when(channel.eventLoop()).thenReturn(eventLoop);
     when(channel.newPromise()).thenReturn(new DefaultChannelPromise(channel));
     when(channel.voidPromise()).thenReturn(new DefaultChannelPromise(channel));
+    ChannelPromise completedPromise = new DefaultChannelPromise(channel)
+        .setSuccess();
+    when(channel.write(any())).thenReturn(completedPromise);
+    when(channel.writeAndFlush(any())).thenReturn(completedPromise);
+    when(writeQueue.enqueue(any(QueuedCommand.class), anyBoolean())).thenReturn(completedPromise);
```
dapengzhang0: Instead of these stubbings, I would return a ChannelPromise instance without setting success. And set success only in test method that needs it (i.e. `notifiedOnReadyAfterWriteCompletes()`) 

```java
when(channel.write(any())).thenReturn(new DefaultChannelPromise(channel));
when(channel.writeAndFlush(any())).thenReturn(new DefaultChannelPromise(channel));
when(writeQueue.enqueue(any(QueuedCommand.class), anyBoolean()))
    .thenReturn(new DefaultChannelPromise(channel));
```

Here `writeQueue` is a mock, so returning a non-null object is only to avoid NPE. Whether it's success or done or failure is irrelevant in most of the tests. Some tests may require a definite state of the returned future and it should be set explicitly in those tests.

netty/src/test/java/io/grpc/netty/NettyStreamTestBase.java
```diff
@@ -155,7 +150,8 @@ public void notifiedOnReadyAfterWriteCompletes() throws IOException {
     sendHeadersIfServer();
     assertTrue(stream.isReady());
     byte[] msg = largeMessage();
-    // The future is set up to automatically complete, indicating that the write is done.
+    // The channel.write future is set up to automatically complete, indicating that the write is
+    // done.
```
dapengzhang0: Set success explicitly here
```java
ChannelPromise completedPromise = new DefaultChannelPromise(channel)
    .setSuccess();
when(writeQueue.enqueue(any(QueuedCommand.class), anyBoolean()))
    .thenReturn(completedPromise);
```

netty/src/test/java/io/grpc/netty/NettyStreamTestBase.java
```diff
@@ -166,7 +162,8 @@ public void notifiedOnReadyAfterWriteCompletes() throws IOException {
   public void shouldBeReadyForDataAfterWritingSmallMessage() throws IOException {
     sendHeadersIfServer();
     // Make sure the writes don't complete so we "back up"
-    reset(future);
+    ChannelPromise uncompletedPromise = new DefaultChannelPromise(channel);
+    when(writeQueue.enqueue(any(QueuedCommand.class), anyBoolean())).thenReturn(uncompletedPromise);
```
dapengzhang0: This is not necessary if the suggestion in one of the other comments applied, but okay to keep it.

netty/src/test/java/io/grpc/netty/NettyStreamTestBase.java
```diff
@@ -180,7 +177,8 @@ public void shouldBeReadyForDataAfterWritingSmallMessage() throws IOException {
   public void shouldNotBeReadyForDataAfterWritingLargeMessage() throws IOException {
     sendHeadersIfServer();
     // Make sure the writes don't complete so we "back up"
-    reset(future);
+    ChannelPromise uncompletedPromise = new DefaultChannelPromise(channel);
+    when(writeQueue.enqueue(any(QueuedCommand.class), anyBoolean())).thenReturn(uncompletedPromise);
```
dapengzhang0: This is not necessary if the suggestion in one of the other comments applied, but okay to keep it.

netty/src/test/java/io/grpc/netty/NettyClientStreamTest.java
```diff
@@ -436,7 +433,10 @@ public void removeUserAgentFromApplicationHeaders() {
     metadata.put(GrpcUtil.USER_AGENT_KEY, "bad agent");
     listener = mock(ClientStreamListener.class);
     Mockito.reset(writeQueue);
-    when(writeQueue.enqueue(any(QueuedCommand.class), any(boolean.class))).thenReturn(future);
+    ChannelPromise completedPromise = new DefaultChannelPromise(channel)
+        .setSuccess();
+    when(writeQueue.enqueue(any(QueuedCommand.class), any(boolean.class)))
+        .thenReturn(completedPromise);
```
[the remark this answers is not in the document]
ejona86: I want to return a future that will complete. The mocking is horrible, and I don't want tests to just hang in the future because the code changes slightly.

netty/src/test/java/io/grpc/netty/NettyStreamTestBase.java
ejona86: I don't want tests that require or implicitly depend on the state of every mock in the system. Only the tests that need the future to be delayed should be aware of the future mocking. Having never-completed futures is simply a bad idea.

carl-mastrangelo:
I guess Im okay with the change, but I don't see how it's better.   It doesn't seem to fix any bugs or add any features.

ejona86:
@carl-mastrangelo, this came out of a discussion with #4510 where it was mentioned that other places are adding the listener before writing.

carl-mastrangelo:
Again, I dont see how this makes the library better, but meh.  LGTM

PR #4522, opened 2018-05-31
buildscripts: correct Android apk size and dex diff output

ericgribkoff:
#4511 only fixed the dex count comparison. This updates the full dex diff and apk size comparison


ericgribkoff:
It doesn't need to as-is, since the old_* stats are all collected after checking out the base commit and rebuilding the hello world app. We shouldn't be building other apps after checking out the base commit, so it would seemingly remain safe to use relative paths at that point

ericgribkoff:
@carl-mastrangelo Do you want changes to how the old_* stats are collected, or is this ok as-is?

PR #4523, opened 2018-05-31
services: allow binlog to blacklist methods

zpencer:
The spec says users can specify a blacklist a method from binlogs by
saying "-package.service/method".


services/src/test/java/io/grpc/services/BinlogHelperTest.java
```diff
@@ -264,6 +264,14 @@ public void configBinLog_multiConfig_noGlobal() throws Exception {
     assertNull(makeLog(configStr, "package.service2/absent"));
   }
 
+  @Test
+  public void configBinLog_blacklist() {
+    assertNull(makeLog("*,-p.s/method", "p.s/method"));
+    assertNull(makeLog("-p.s/method,*", "p.s/method"));
+    assertNull(makeLog("p.s/*,-p.s/method", "p.s/method"));
```
dapengzhang0: can we also check `makeLog("p.s/*,-p.s/method", "p.s/otherMethod")` not null?

zpencer: Done

PR #4524, opened 2018-06-01
auth: Require PRIVACY_AND_INTEGRITY for GoogleCredentials

zhangkun83:
This keeps them more secure. Other types of creds are left as-is, since
we don't quite know if it makes sense to have a similar restriction. (It
likely does make sense, but this is a more precise change for our
needs.)

This is a rollforward of 8e9d4cbe5cb455315e55846e91706cccecb4b577
which was rolled back in de9515269635d2e8e84540fcd6744cc1a69ce81b

Reference: #4528

### DIFF
--- a/auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
+++ b/auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
@@ -26,6 +26,7 @@ import io.grpc.Attributes;
 import io.grpc.CallCredentials;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
+import io.grpc.SecurityLevel;
 import io.grpc.Status;
 import io.grpc.StatusException;
 import java.io.IOException;
@@ -51,7 +52,10 @@ final class GoogleAuthLibraryCallCredentials implements CallCredentials {
       = Logger.getLogger(GoogleAuthLibraryCallCredentials.class.getName());
   private static final JwtHelper jwtHelper
       = createJwtHelperOrNull(GoogleAuthLibraryCallCredentials.class.getClassLoader());
+  private static final Class<? extends Credentials> googleCredentialsClass
+      = loadGoogleCredentialsClass();
 
+  private final boolean requirePrivacy;
   @VisibleForTesting
   final Credentials creds;
 
@@ -65,9 +69,18 @@ final class GoogleAuthLibraryCallCredentials implements CallCredentials {
   @VisibleForTesting
   GoogleAuthLibraryCallCredentials(Credentials creds, JwtHelper jwtHelper) {
     checkNotNull(creds, "creds");
+    boolean requirePrivacy = false;
+    if (googleCredentialsClass != null) {
+      // All GoogleCredentials instances are bearer tokens and should only be used on private
+      // channels. This catches all return values from GoogleCredentials.getApplicationDefault().
+      // This should be checked before upgrading the Service Account to JWT, as JWT is also a bearer
+      // token.
+      requirePrivacy = googleCredentialsClass.isInstance(creds);
+    }
     if (jwtHelper != null) {
       creds = jwtHelper.tryServiceAccountToJwt(creds);
     }
+    this.requirePrivacy = requirePrivacy;
     this.creds = creds;
   }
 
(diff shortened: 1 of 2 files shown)


auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
```diff
@@ -77,6 +90,14 @@ public void thisUsesUnstableApi() {}
   @Override
   public void applyRequestMetadata(MethodDescriptor<?, ?> method, Attributes attrs,
       Executor appExecutor, final MetadataApplier applier) {
+    SecurityLevel security = checkNotNull(attrs.get(ATTR_SECURITY_LEVEL), "securityLevel");
```
carl-mastrangelo: Is this guaranteed to be set?

auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
```diff
@@ -77,6 +90,14 @@ public void thisUsesUnstableApi() {}
...
+    if (requirePrivacy && security != SecurityLevel.PRIVACY_AND_INTEGRITY) {
+      applier.fail(Status.UNAUTHENTICATED
+          .withDescription("Credentials require channel with PRIVACY_AND_INTEGRITY security level. "
+            + "Observed security level: " + security));
```
carl-mastrangelo: indent +2

auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
```diff
@@ -214,6 +235,19 @@ static JwtHelper createJwtHelperOrNull(ClassLoader loader) {
...
+  private static Class<? extends Credentials> loadGoogleCredentialsClass() {
+    Class<?> rawGoogleCredentialsClass;
+    try {
+      // Can't use a loader as it disables ProGuard's reference detection and would fail to rename
+      // this reference. Unfortunately this will initialize the class.
+      rawGoogleCredentialsClass = Class.forName("com.google.auth.oauth2.GoogleCredentials");
+    } catch (ClassNotFoundException ex) {
+      return null;
+    }
+    return rawGoogleCredentialsClass.asSubclass(Credentials.class);
```
carl-mastrangelo: Credentials isn't a subclass is it?

auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
zhangkun83: API says so, although it depends on transport and channel implementation, and we already had a case internally where a custom channel is not setting it. Probably safer to tolerate null here.

auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
zhangkun83: Done.

auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
zhangkun83: I don't understand your question :)

carl-mastrangelo: You are calling asSubclass, but infact Credentials is a super class

zhangkun83: Javadoc says "Casts this Class object to represent a subclass of the class represented by the specified class object." So it returns "subclass of Credentials".

auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
```diff
@@ -214,6 +241,19 @@ static JwtHelper createJwtHelperOrNull(ClassLoader loader) {
...
 
+  @Nullable
+  private static Class<? extends Credentials> loadGoogleCredentialsClass() {
+    Class<?> rawGoogleCredentialsClass;
+    try {
+      // Can't use a loader as it disables ProGuard's reference detection and would fail to rename
+      // this reference. Unfortunately this will initialize the class.
+      rawGoogleCredentialsClass = Class.forName("com.google.auth.oauth2.GoogleCredentials");
+    } catch (ClassNotFoundException ex) {
+      return null;
```
carl-mastrangelo: This should be a fine level log.

zhangkun83: Done

PR #4525, opened 2018-06-01
grpclb: use exponential back-off for retries of balancer RPCs

zhangkun83:
If initial response was received during the previous RPC, the retry will be immediate, and the back-off will be reset. Otherwise, will use the current back-off sequence to schedule the next retry.

`FallbackPolicy` classes are made public visibility for grpclb to use. They are still internal API though.

@markdroth, please verify this is the desired behavior and consistent with C++.

Reviewers, please ignore the individual commits.


markdroth:
We actually trigger this when we've seen the initial response from the balancer, not the first serverlist.

zhangkun83:
> We actually trigger this when we've seen the initial response from the balancer, not the first serverlist.

Did you mean "reset" instead of "trigger"?

markdroth:
Yes.  When the call that just failed received the initial response from the balancer, then we reset the backoff and try to create a new call immediately; otherwise, we try to create a new call after the backoff delay.

zhangkun83:
@markdroth, I have updated the PR according to your description. PTAL.

grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancerTest.java
```diff
@@ -1583,6 +1600,109 @@ public void grpclbMultipleAuthorities() throws Exception {
...
+    inOrder.verify(backoffPolicy2).nextBackoffNanos();
+    assertEquals(1, fakeClock.numPendingTasks(LB_RPC_RETRY_TASK_FILTER));
+
+    // Fast-forward to a moment before the retry, the time spent in the last try is deducted.
+    fakeClock.forwardNanos(10 - 4 - 1);
+    verifyNoMoreInteractions(mockLbService);
+    // Then time for retry
+    fakeClock.forwardNanos(1);
+    inOrder.verify(mockLbService).balanceLoad(lbResponseObserverCaptor.capture());
+    lbResponseObserver = lbResponseObserverCaptor.getValue();
```
dapengzhang0: not used? why captor?

markdroth:
From a quick glance at the test code, the behavior looks right here.  Thanks!

grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancerTest.java
zhangkun83: Removed the assignment. I left the captor there because it's easier than any() which doesn't handle generics well.

PR #4526, opened 2018-06-01
core: channel tracing to log events for channel state change

dapengzhang0:


core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -314,6 +318,14 @@ private void gotoState(final ConnectivityStateInfo newState) {
       Preconditions.checkState(state.getState() != SHUTDOWN,
           "Cannot transition out of SHUTDOWN to " + newState);
       state = newState;
+      if (channelTracer != null) {
+        channelTracer.reportEvent(
+            new ChannelTrace.Event.Builder()
+                .setDescription("Entering " + state + " state")
+                .setSeverity(ChannelTrace.Event.Severity.CT_INFO)
+                .setTimestampNanos(timeProvider.currentTimeNanos())
```
zpencer: Do we ever set a timestamp that is different from when we call `reportEvent`? Maybe the `channelTracer` can handle the time stamping for us.

zpencer: That's a good reason, may be worth a comment somewhere.

PR #4532, opened 2018-06-04
cronet: fill in transport attributes as required by CallCredentials

zhangkun83:
This should unblock #4524


cronet/src/test/java/io/grpc/cronet/CronetClientTransportTest.java
```diff
@@ -61,7 +65,7 @@ public void setUp() {
         new CronetClientTransport(
             streamFactory,
             new InetSocketAddress("localhost", 443),
-            "",
+            "test.example.com",
```
ericgribkoff: nit/optional: make this string a static variable?

zhangkun83: Done.

zhangkun83:
Our internal user has verified that this fix works (b/80529703#comment5).

PR #4533, opened 2018-06-05
okhttp: deprecate OkHttpChannelBuilder#negotiationType

ericgribkoff:
The functionality provided by `#negotiationType` is identical to just directly invoking `#usePlaintext`/`#useTransportSecurity`. 

Removing the `#negotiationType` method (which isn't part of `ManagedChannelBuilder`) will make it cleaner to use `OkHttpChannelBuilder` as the delegate builder for a `ForwardingChannelBuilder` (specifically, this helps  AndroidChannelBuilder, which loads OkHttp via reflection and so doesn't have easy access to the `#negotiationType` method's `io.grpc.okhttp.NegotiationType` parameter)

### DIFF
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
@@ -209,7 +209,7 @@ public class OkHttpChannelBuilder extends
         negotiationType = NegotiationType.PLAINTEXT;
         break;
       default:
-        throw new RuntimeException("Unknown negotiation type: " + type);
+        throw new AssertionError("Unknown negotiation type: " + type);
     }
     return this;
   }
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -183,9 +195,22 @@ public final OkHttpChannelBuilder transportExecutor(@Nullable Executor transport
...
+    Preconditions.checkNotNull(type, "type");
+    switch (type) {
+      case TLS:
+        negotiationType = NegotiationType.TLS;
+        break;
+      case PLAINTEXT:
+        negotiationType = NegotiationType.PLAINTEXT;
+        break;
+      default:
+        throw new RuntimeException("Unknown negotiation type: " + type);
```
carl-mastrangelo: AssertionError maybe?

ericgribkoff: Done

PR #4539, opened 2018-06-06
Add clarification to ClientCall javadoc 

groakley:
- Clarify that some methods of ClientCall may be invoked from any thread. 
- Adds @throws clauses to request(). 
- Adds threading note to cancel() to match request().

### DIFF
--- a/core/src/main/java/io/grpc/ClientCall.java
+++ b/core/src/main/java/io/grpc/ClientCall.java
@@ -39,7 +39,7 @@ import javax.annotation.Nullable;
  * naturally acknowledges its request.
  *
  * <p>Methods are guaranteed to be non-blocking. Implementations are not required to be thread-safe
- * except for {@link #request} and {@link #cancel}, which may be called from any thread.
+ * except for {@link #request}, which may be called from any thread.
  *
  * <p>There is no interaction between the states on the {@link Listener Listener} and {@link
  * ClientCall}, i.e., if {@link Listener#onClose Listener.onClose()} is called, it has no bearing on
@@ -207,8 +207,6 @@ public abstract class ClientCall<ReqT, RespT> {
    * <p>It is recommended that at least one of the arguments to be non-{@code null}, to provide
    * useful debug information. Both argument being null may log warnings and result in suboptimal
    * performance. Also note that the provided information will not be sent to the server.
-   *
-   * <p>This method is safe to call from multiple threads without external synchronization.
    *
    * @param message if not {@code null}, will appear as the description of the CANCELLED status
    * @param cause if not {@code null}, will appear as the cause of the CANCELLED status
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/grpc/ClientCall.java
```diff
@@ -205,6 +208,8 @@ public void onReady() {}
    * useful debug information. Both argument being null may log warnings and result in suboptimal
    * performance. Also note that the provided information will not be sent to the server.
    *
+   * <p>This method is safe to call from multiple threads without external synchronization.
```
carl-mastrangelo: I'm not sure this is threadsafe.  request causes events to be fired, but doesn't change the state of the object, where this does.  The only known implementation of ClientCall is not threadsafe for cancel (i.e. `ClientCallImpl`, which has unsafe access to `cancelCalled`).   @ejona86  can comment on whether this method _should_ be thread safe.

groakley: iiuc, cancel() is called directly when someone with a handle to a ListenableFuture (e.g. from a FutureStub) calls ListenableFuture.cancel(true). This could be on any thread. 

https://github.com/grpc/grpc-java/blob/master/stub/src/main/java/io/grpc/stub/ClientCalls.java#L483

If the current implementation's cancel() is not thread-safe, then I think it needs to be fixed.

carl-mastrangelo: That sounds plausible.    We probably didn't notice due to hardly ever using the ListenableFuture stubs.   In any case, the docs for `cancel()` cannot be updated before the implementation.

groakley: Makes sense. I will confirm this and then create a tracking issue if correct. In the meantime, removed comments about cancel().

core/src/main/java/io/grpc/ClientCall.java
```diff
@@ -38,7 +38,8 @@
  * Applications are expected to utilize normal payload messages for such signals, as a response
  * naturally acknowledges its request.
  *
- * <p>Methods are guaranteed to be non-blocking. Implementations are not required to be thread-safe.
+ * <p>Methods are guaranteed to be non-blocking. Implementations are not required to be thread-safe
+ * except for {@link #request}, which may be called from any thread.
```
dapengzhang0: IMO Javadoc is facing the API users not the implementors, so `Implementations are not required ...` should not be mentioned here in the javadoc at all. It can be still be added in form of an ordinary comment `//...` or `/* ... */` out side the javadoc, or use the javadoc tag `@implNote` (require java 8).

dapengzhang0: Or rephrase it to be user-facing like `Not thread-safe except for  {@link #request}, which may be called from any thread.`

groakley: Reprhased. I did not clean up other examples of this kind of phrasing in this file, though.

core/src/main/java/io/grpc/ClientCall.java
groakley: Created #4547 to track.

PR #4540, opened 2018-06-06
all: add gradle format checker

dapengzhang0:
This PR adds an automatic gradle format checker and reformats all the *.gradle files. After this, new changes to *.gradle files will fail to build if not in good format, just like checkStyle failure.

commit 1 adds the [spotless formatter plugin](https://github.com/diffplug/spotless/tree/master/plugin-gradle).
commit 2 applies the formatter task.

### DIFF
--- a/all/build.gradle
+++ b/all/build.gradle
@@ -4,27 +4,23 @@ description = "gRPC: All"
 
 buildscript {
     repositories {
-        maven {
-            // The google mirror is less flaky than mavenCentral()
-            url "https://maven-central.storage-download.googleapis.com/repos/central/data/"
-        }
+        maven { // The google mirror is less flaky than mavenCentral()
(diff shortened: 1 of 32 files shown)


carl-mastrangelo:
Does this get applied automatically on `./gradlew build`?   If it's really easy for these files to drift out of being formatted right, I don't think this is a good change.  If not, then either ./gradlew build should automatically apply it, or there should be a test that checks that it has been applied.

dapengzhang0:
Just like checkStyle that we already have. Actually spotless can do
```
spotless {
  java {
    googleJavaFormat()
  }
```
that is similar to checkStyle.

dapengzhang0:
Thanks @carl-mastrangelo 
Rebasing

PR #996, opened 2015-09-10
Basic implementation of corking

louiscryan:
Implements #994

Issue #994, opened 2015-09-10
Implement cork/uncork in client and server calls


louiscryan:
By adding a cork/uncork to defer flushing messages we can significantly improve throughput for streaming operations when messages sizes are significantly smaller than the flow-control window

PR #4545, opened 2018-06-07
 Add (Client|Server)Call#cork method. 

dmvk:
Hi, this is a follow up for abandoned #994 and #996. This should allow user to trade-off latency for throughput.

We ended up doing manual batching on client side, which adds complexity to the client app.

### DIFF
--- a/core/src/main/java/io/grpc/ClientCall.java
+++ b/core/src/main/java/io/grpc/ClientCall.java
@@ -17,7 +17,6 @@
 package io.grpc;
 
 import com.google.errorprone.annotations.DoNotMock;
-import io.grpc.internal.ClientStream;
 import javax.annotation.Nullable;
 
 /**
@@ -256,15 +255,16 @@ public abstract class ClientCall<ReqT, RespT> {
   }
 
   /**
-   * Defers flushing the messages written using {@link #sendMessage(Object)} to the
-   * transport until the write buffer is full. The amount of data buffered while corked is
-   * dependent on the transport implementation so users of this API should not make strong
-   * assumptions about how long data is buffered while corked.
+   * Defers flushing the messages written using {@link #sendMessage(Object)} until the cork is
+   * closed or as necessary to prevent excessive buffering. The amount of data buffered while
+   * corked is dependent on the transport implementation so users of this API should not make
+   * strong assumptions about how long data is buffered while corked.
    *
    * <p>Corking can provide throughput gains by allowing a transport to perform fewer writes to
    * the underlying network.
    *
-   * <p>The {@link ClientStream#flush()} is being deferred until all opened corks are closed.
+   * <p>In case of multiple {@link Cork}s being open, all of them need to be closed in order to
+   * trigger flushing.
    *
    * @since 1.14.0
    */
--- a/core/src/main/java/io/grpc/ClientCall.java
+++ b/core/src/main/java/io/grpc/ClientCall.java
@@ -255,6 +255,20 @@ public abstract class ClientCall<ReqT, RespT> {
   }
 
   /**
+   * Defers flushing the messages written using {@link #sendMessage(Object)} to the
+   * transport until the write buffer is full. The amount of data buffered while corked is
+   * dependent on the transport implementation so users of this API should not make strong
+   * assumptions about how long data is buffered while corked.
+   *
+   * <p>Corking can provide throughput gains by allowing a transport to perform fewer writes to
+   * the underlying network.
+   */
+  @ExperimentalApi
+  public Cork cork() {
+    return new Cork.NoopCork();
+  }
+
+  /**
    * Returns additional properties of the call. May only be called after {@link Listener#onHeaders}
    * or {@link Listener#onClose}. If called prematurely, the implementation may throw {@code
    * IllegalStateException} or return arbitrary {@code Attributes}.
@@ -268,4 +282,5 @@ public abstract class ClientCall<ReqT, RespT> {
   public Attributes getAttributes() {
     return Attributes.EMPTY;
   }
+
 }
--- a/core/src/main/java/io/grpc/ClientCall.java
+++ b/core/src/main/java/io/grpc/ClientCall.java
@@ -17,6 +17,7 @@
 package io.grpc;
 
 import com.google.errorprone.annotations.DoNotMock;
+import io.grpc.internal.ClientStream;
 import javax.annotation.Nullable;
 
 /**
@@ -262,6 +263,10 @@ public abstract class ClientCall<ReqT, RespT> {
    *
    * <p>Corking can provide throughput gains by allowing a transport to perform fewer writes to
    * the underlying network.
+   *
+   * <p>The {@link ClientStream#flush()} is being deferred until all opened corks are closed.
+   *
+   * @since 1.14.0
    */
   @ExperimentalApi
   public Cork cork() {
--- a/core/src/main/java/io/grpc/Cork.java
+++ b/core/src/main/java/io/grpc/Cork.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2018 The gRPC Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc;
+
+import java.io.Closeable;
+
+/**
+ * Defers flushing the messages written using {@link ClientCall#sendMessage(Object)} and
+ * {@link ServerCall#sendMessage(Object)} transport until the write buffer is full.
+ * The amount of data buffered while corked is dependent on the transport implementation
+ * so users of this API should not make strong assumptions about how long data is buffered while
+ * corked.
+ *
+ * <p>Corking can provide throughput gains by allowing a transport to perform fewer writes to
+ * the underlying network.
+ */
+public interface Cork extends Closeable {
+
+  /**
+   * No-Op implementation of {@link Cork} interface.
+   */
+  class NoopCork implements Cork {
+
+    @Override
+    public void close() {
+      // noop
+    }
+  }
+}
--- a/core/src/main/java/io/grpc/Cork.java
+++ b/core/src/main/java/io/grpc/Cork.java
@@ -28,6 +28,7 @@ import java.io.Closeable;
  * <p>Corking can provide throughput gains by allowing a transport to perform fewer writes to
  * the underlying network.
  */
+@ExperimentalApi
 public interface Cork extends Closeable {
 
   /**
@@ -40,4 +41,7 @@ public interface Cork extends Closeable {
       // noop
     }
   }
+
+  @Override
+  void close();
 }
--- a/core/src/main/java/io/grpc/ServerCall.java
+++ b/core/src/main/java/io/grpc/ServerCall.java
@@ -189,6 +189,20 @@ public abstract class ServerCall<ReqT, RespT> {
   }
 
   /**
+   * Defers flushing the messages written using {@link #sendMessage(Object)} to the
+   * transport until the write buffer is full. The amount of data buffered while corked is
+   * dependent on the transport implementation so users of this API should not make strong
+   * assumptions about how long data is buffered while corked.
+   *
+   * <p>Corking can provide throughput gains by allowing a transport to perform fewer writes to
+   * the underlying network.
+   */
+  @ExperimentalApi
+  public Cork cork() {
+    return new Cork.NoopCork();
+  }
+
+  /**
    * Returns properties of a single call.
    *
    * <p>Attributes originate from the transport and can be altered by {@link ServerTransportFilter}.
--- a/core/src/main/java/io/grpc/ServerCall.java
+++ b/core/src/main/java/io/grpc/ServerCall.java
@@ -17,7 +17,6 @@
 package io.grpc;
 
 import com.google.errorprone.annotations.DoNotMock;
-import io.grpc.internal.ServerStream;
 import javax.annotation.Nullable;
 
 /**
@@ -190,15 +189,16 @@ public abstract class ServerCall<ReqT, RespT> {
   }
 
   /**
-   * Defers flushing the messages written using {@link #sendMessage(Object)} to the
-   * transport until the write buffer is full. The amount of data buffered while corked is
-   * dependent on the transport implementation so users of this API should not make strong
-   * assumptions about how long data is buffered while corked.
+   * Defers flushing the messages written using {@link #sendMessage(Object)} until the cork is
+   * closed or as necessary to prevent excessive buffering. The amount of data buffered while
+   * corked is dependent on the transport implementation so users of this API should not make
+   * strong assumptions about how long data is buffered while corked.
    *
    * <p>Corking can provide throughput gains by allowing a transport to perform fewer writes to
    * the underlying network.
    *
-   * <p>The {@link ServerStream#flush()} is being deferred until all opened corks are closed.
+   * <p>In case of multiple {@link Cork}s being open, all of them need to be closed in order to
+   * trigger flushing.
    *
    * @since 1.14.0
    */
--- a/core/src/main/java/io/grpc/ServerCall.java
+++ b/core/src/main/java/io/grpc/ServerCall.java
@@ -17,6 +17,7 @@
 package io.grpc;
 
 import com.google.errorprone.annotations.DoNotMock;
+import io.grpc.internal.ServerStream;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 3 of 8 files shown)


thelinuxfoundation:

* User @dmvk isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

dmvk:
signed

core/src/main/java/io/grpc/Cork.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Defers flushing the messages written using {@link ClientCall#sendMessage(Object)} and
+ * {@link ServerCall#sendMessage(Object)} transport until the write buffer is full.
+ * The amount of data buffered while corked is dependent on the transport implementation
+ * so users of this API should not make strong assumptions about how long data is buffered while
+ * corked.
+ *
+ * <p>Corking can provide throughput gains by allowing a transport to perform fewer writes to
+ * the underlying network.
+ */
```
dapengzhang0: Add `@ExperimentalApi`

core/src/main/java/io/grpc/ClientCall.java
```diff
@@ -254,6 +254,20 @@ public void setMessageCompression(boolean enabled) {
...
   }
 
+  /**
+   * Defers flushing the messages written using {@link #sendMessage(Object)} to the
+   * transport until the write buffer is full. The amount of data buffered while corked is
+   * dependent on the transport implementation so users of this API should not make strong
+   * assumptions about how long data is buffered while corked.
+   *
+   * <p>Corking can provide throughput gains by allowing a transport to perform fewer writes to
+   * the underlying network.
```
dapengzhang0: Add `@since 1.14.0`

core/src/main/java/io/grpc/ServerCall.java
```diff
@@ -188,6 +188,20 @@ public void setCompression(String compressor) {
...
 
+  /**
+   * Defers flushing the messages written using {@link #sendMessage(Object)} to the
+   * transport until the write buffer is full. The amount of data buffered while corked is
+   * dependent on the transport implementation so users of this API should not make strong
+   * assumptions about how long data is buffered while corked.
+   *
+   * <p>Corking can provide throughput gains by allowing a transport to perform fewer writes to
+   * the underlying network.
+   */
```
dapengzhang0: Reading the javadoc I'm uncertain the following:
Is this allowed to call this method multiple times without closing the previous one before calling next one?
If allowed, will it return a different or the same instance? If returning a different instance, what is expected behavior?

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -431,7 +435,7 @@ public void sendMessage(ReqT message) {
     // For unary requests, we don't flush since we know that halfClose should be coming soon. This
     // allows us to piggy-back the END_STREAM=true on the last message frame without opening the
     // possibility of broken applications forgetting to call halfClose without noticing.
-    if (!unaryRequest) {
+    if (!unaryRequest && corks.size() == 0) {
```
dapengzhang0: favor `corks.isEmpty()`

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -442,6 +446,22 @@ public void setMessageCompression(boolean enabled) {
...
+
+      @Override
+      public void close() {
+        checkState(corks.remove(this), "cork already closed");
+        if (corks.size() == 0) {
+          stream.flush();
+        }
+      }
+    };
+    checkState(corks.add(cork), "unable to add cork");
```
dapengzhang0: Should not `checkState`. `Set.add()` returns false only if the object to be added is already in the set, which is impossible here because the `cork` is `new`ed.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -442,6 +446,22 @@ public void setMessageCompression(boolean enabled) {
     stream.setMessageCompression(enabled);
   }
 
+  @Override
+  public Cork cork() {
+    final Cork cork = new Cork() {
+
+      @Override
+      public void close() {
+        checkState(corks.remove(this), "cork already closed");
```
dapengzhang0: Should not `checkState` here. Instead,

```java
if (!corks.remove(this)) {
  return;
}
```

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -84,6 +87,7 @@
   private boolean fullStreamDecompression;
   private DecompressorRegistry decompressorRegistry = DecompressorRegistry.getDefaultInstance();
   private CompressorRegistry compressorRegistry = CompressorRegistry.getDefaultInstance();
+  private final Set<Cork> corks = Sets.newConcurrentHashSet();
```
dapengzhang0: Is it possible to use an `AtomicInteger` instead?

dapengzhang0:
Thank you @dmvk for your PR!

core/src/main/java/io/grpc/Cork.java
```diff
@@ -0,0 +1,43 @@
...
+public interface Cork extends Closeable {
```
carl-mastrangelo: You should probably override `close` to not throw IOException

core/src/main/java/io/grpc/ClientCall.java
```diff
@@ -254,6 +254,20 @@ public void setMessageCompression(boolean enabled) {
...
+   */
+  @ExperimentalApi
+  public Cork cork() {
```
carl-mastrangelo: why not just have a method that toggles between corking and uncorking?   i.e. why have object separate from the ClientCall?

core/src/main/java/io/grpc/internal/ClientCallImpl.java
carl-mastrangelo: We cannot take a dependency on Guava collections in the client side code because it bloats Android jars too much.

ericgribkoff: Yes, please avoid pulling in Guava collections - it's overhead is significant on Android (unfortunately the Android increase in method counts doesn't show up much from our android/dex_diff CI check right now, because of a transitive dependency already bringing in Guava collections, but that dependency is being fixed in https://github.com/grpc/grpc-java/pull/4494)

core/src/main/java/io/grpc/internal/ClientCallImpl.java
dmvk: done

core/src/main/java/io/grpc/ClientCall.java
dmvk: done

core/src/main/java/io/grpc/Cork.java
dmvk: done

core/src/main/java/io/grpc/Cork.java
dmvk: done

dapengzhang0:
A general question: What is the benefit of defining the API signature as
```java
public Cork cork();

public interface extends Closable;
```

rather than the one in #996, which looks more straightforward and simple
```java
public void cork();
public void unCork();
```

dapengzhang0:
A general question: What is the benefit of defining the API signature as
```java
public Cork cork();

public interface Cork extends Closable;
```

rather than the one in #996, which looks more straightforward and simple
```java
public void cork();
public void unCork();
```

core/src/main/java/io/grpc/internal/ClientCallImpl.java
dmvk: Yes, i exchange of having `AtomicBoolean closed` flag inside each cork. Changed.

I think this is closely related to the discussion bellow (if we should have cork/uncork methods).

core/src/main/java/io/grpc/ServerCall.java
dmvk: added a paragraph that describes the behavior

core/src/main/java/io/grpc/internal/ClientCallImpl.java
dmvk: fixed

core/src/main/java/io/grpc/ClientCall.java
dmvk: From my point of view this is far more flexible for the user to implement custom interceptor logic, where you can call `cork` multiple times. Only concern may be allocation of new object.

I also think this API is easier for user to understand. Having both cork / uncork methods may be confusing in terms of what should happen if they're called multiple times (user would have to read javadocs carefully). 

What do you think?

carl-mastrangelo: On the pro side:
* corking can be stacked
* interceptors don't need to know about other callers of cork

On the con side:
* doesn't match TCP_CORK usages, from which the name is derived
* callers of uncork (`close`) may expect flushes to happen immediately.
* Cork object has different lifetime than Call, and needs to have threading semantics defined


Do you agree with the pros and cons?

dmvk: Agreed on pro side.

Few points for the cons side:
- I think we don't want match TCP_CORK contract, this would involve adding timers inside `Call`. We should leave this to the user to decide it should behave. [1]
- different lifetime: `Cork` is thread safe, but this can indeed result into bad implementation decision and memory leaks if the Cork reference would not be released after the `Call` finishes.

1) `man tcp`
```
If set, don't send out partial frames. All queued partial frames are sent when the option is cleared again. This is useful for prepending headers before calling sendfile(2), or for throughput optimization. As currently implemented, there is a 200 millisecond ceiling on the time for which output is corked by TCP_CORK. If this ceiling is reached, then queued data is automatically transmitted. This option can be combined with TCP_NODELAY only since Linux 2.5.71. This option should not be used in code intended to be portable. 
```

dmvk: @carl-mastrangelo Hi, did you make any decision how the API should look like? I would like to move this forward.

carl-mastrangelo:
@dmvk I spent some time trying to come up with an interceptor that does the same thing.  The main issue I ran into was that the interceptor can't be forced to only apply to a single call (maybe thats okay?).    

What I think needs to be balanced is how much more this complicates ClientCall, for how much improved performance it offers.  I mention the ClientInterceptor because it does almost the same thing, but doesn't modify the ClientCall API.  

Do you have a rough estimate for how much faster corking is with your PR?  

cc  @ejona86

dmvk:
In a very naive benchmark, with very small messages, no fine tuning of parameters, I was able to observe double of an original throughput.

https://github.com/dmvk/grpc-benchmark/blob/master/src/main/java/cz/dmvk/grpc/Benchmark.java

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -442,6 +446,22 @@ public void setMessageCompression(boolean enabled) {
     stream.setMessageCompression(enabled);
   }
 
+  @Override
+  public Cork cork() {
```
ejona86: I don't think we should have any part of this implementation be thread-safe. By necessity the caller has to coordinate with senders and sending is not thread-safe.

core/src/main/java/io/grpc/ClientCall.java
```diff
@@ -254,6 +255,24 @@ public void setMessageCompression(boolean enabled) {
...
+  /**
+   * Defers flushing the messages written using {@link #sendMessage(Object)} to the
+   * transport until the write buffer is full. The amount of data buffered while corked is
+   * dependent on the transport implementation so users of this API should not make strong
+   * assumptions about how long data is buffered while corked.
+   *
+   * <p>Corking can provide throughput gains by allowing a transport to perform fewer writes to
+   * the underlying network.
+   *
+   * <p>The {@link ClientStream#flush()} is being deferred until all opened corks are closed.
```
ejona86: This is an internal detail. I don't think it should be part of the API.

carl-mastrangelo:
@dmvk I see.    Would you consider making the cork methods on ClientCall simple incrementCork / decrementCork counts, rather than introduce a cork object?  I am fairly concerned about new types.

core/src/main/java/io/grpc/ClientCall.java
```diff
@@ -254,6 +255,24 @@ public void setMessageCompression(boolean enabled) {
     // noop
   }
 
+  /**
+   * Defers flushing the messages written using {@link #sendMessage(Object)} to the
```
ejona86: I think this should mention "or the cork is closed" to make it more obvious.

How about:
> Defers flushing the messages written using sendMessage() until the cork is closed or as necessary to prevent excessive buffering.

ejona86:
I don't think this feature can truly be added as interceptor. There can be really big gains for small messages.

I do wonder if we want to allow multiple concurrent corks, but it doesn't add much complexity so I can live with it either way. I'm more concerned about how hard it would be for a user to debug such a setup.

ejona86:
> ClientCall simple incrementCork / decrementCork counts, rather than introduce a cork object

I'm going to NAK that; that sounds like a really ugly API. If it was a boolean: cork(), disableCork() it might be fine. But honestly, it seems the allocation would be virtually nothing when the corking is helping. Since corking is implicit for unary, I don't see much issue.

I am interested to investigate whether the API (whatever it is) may make interceptors confused/confusing. I'll also note that this API does not support corking the request/response headers. That was important in C. It basically lets you make a streaming call as efficient as we've made unary calls.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
dmvk: I'm not sure about this one. I'm pretty sure many users would close cork after certain timeout, which would be called from different thread, so they would have to do synchronization on their own, but that may be ok since whole ClientCallImpl is not thread-safe.

core/src/main/java/io/grpc/ClientCall.java
dmvk: Ok, how about:  `In case of multiple {@link Cork}s being open, all of them need to be closed in order to continue flushing.`?

dmvk: changed

core/src/main/java/io/grpc/internal/ClientCallImpl.java
dmvk: Agreed with the batch processing, but this is also useful for stream processing, where user does not now if another write is coming, but still wants to trade of throughput for latency.

Anyway, I can remove the atomic properties and user can provide own synchronization if necessary.

What do you think?

ejona86: Yeah, let's do that. The `stream.flush();` here isn't actually thread-safe and the complexity involved to change it would be high and limit other optimizations.

devinrsmith:
I'm very interested in a feature that allows messages to be batched by the respective server or client in java. I don't have the full context or history behind this PR, but is there any reason why exposing `Stream#flush()` semantics to the appropriate interfaces can't accomplish the same thing? (And obviously not calling `Stream#flush()` on every `sendMessage` when the caller wants to batch.)

ejona86:
> but is there any reason why exposing Stream#flush() semantics to the appropriate interfaces can't accomplish the same thing

As new API, existing users wouldn't be calling flush(), so they would be broken. It would also make the simple use case more complicated. So instead the style is to have some API to inhibit the flush. This is the same approach taken by C core, although not in the "cork" style.

devinrsmith:
Thanks for the response. I was still thinking that existing users would work as advertised (flush() automatically by default), with new users having to opt into to batching / manually calling of flush(). But I do see some of the merit in having a completely new inhibit based interface.

ejona86:
> with new users having to opt into to batching / manually calling of flush()

@devinrsmith, ah, yes, that would work. I think we prefer something like corking so that the producing code doesn't need to be aware it is part of a batch. The batch code can simply create the cork before calling other code. I think that's a similar reason why TCP_CORK exists as it does, as you can use pre-existing read/write stdlib code.

zpencer:
Hi @dmvk , friendly ping for an update on this PR

bubenheimer:
This kind of feature is really important for my mobile use case, where I have streams of atomic updates that need to be batched up for power and bandwidth savings. Latency is largely irrelevant, maximum batch size is critical. I ran a test sending 10,000 miniscule messages to the server and back, and ended up with about 15,000 TCP packets on the wire, give or take, where a few hundred would have seemed appropriate. There was a small benefit from grpc's low-level write combining
, but the vast majority of packets contained very little actual data.

zhangkun83:
@ejona86 ping

rubensayshi:
wondering why this has gone stale, seems like a good feature to be added?

dsethi:
So my use case is sending a large number of small protobufs over rpc -- I need to batch them for throughput. Sample python client code -- which does not work -- 

    channel = grpc.insecure_channel('localhost:2020', options=[('grpc.max_concurrent_streams', 1),
                                                               ('grpc.max_send_message_length', 1 * 1024 * 1024),
                                                               ('grpc.max_receive_message_length', 1 * 1024 * 1024),
                                                               ('grpc.http2.max_frame_size', 16777215),
                                                               ('grpc.http2.write_buffer_size', 10 * 1024 * 1024),
                                                               ('WriteFlag.buffer_hint', True)

does this use corking underneath?

PR #4551, opened 2018-06-08
services: binlog status code, msg, details

zpencer:
These are first class concepts and must exist as top level fields.

### DIFF
--- a/services/src/main/java/io/grpc/services/BinlogHelper.java
+++ b/services/src/main/java/io/grpc/services/BinlogHelper.java
@@ -83,6 +83,13 @@ final class BinlogHelper {
   private static final Logger logger = Logger.getLogger(BinlogHelper.class.getName());
   private static final boolean SERVER = true;
   private static final boolean CLIENT = false;
+  // This is a special key that is set from above the application API,
+  // despite having a 'grpc-' prefix. Normally 'grpc-' metadata keys are set from within
+  // gRPC.
+  static final Metadata.Key<byte[]> STATUS_DETAILS_KEY =
+      Metadata.Key.of(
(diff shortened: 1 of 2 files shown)


services/src/main/java/io/grpc/services/BinlogHelper.java
```diff
@@ -83,6 +83,13 @@
   private static final Logger logger = Logger.getLogger(BinlogHelper.class.getName());
   private static final boolean SERVER = true;
   private static final boolean CLIENT = false;
+  // This is a special key that is set from above the application API,
+  // despite having a 'grpc-' prefix. Normally 'grpc-' metadata keys are set from within
+  // gRPC.
+  static final Metadata.Key<byte[]> STATUS_DETAILS_KEY =
```
carl-mastrangelo: Shouldn't this be snuck from `StatusProto`?

zpencer: The StatusProto version of this turns it into an actual proto object, but we only care about the bytes. Also, that Key is private which is inconvenient to deal with.

carl-mastrangelo: Can you comment saying the name must be identical to the actual key location?

zpencer: You mean the fact that "grpc-status-details-bin" is not some arbitrary name, but is instead a special well known key?

carl-mastrangelo: yes.  It needs to be clear that grpc- is a special

zpencer:
@carl-mastrangelo PTAL

Issue #4515, opened 2018-05-29
examples/helloworldtls: SSLHandshakeException when using ClientAuth.REQUIRE with two hosts


DmPanov:
I am trying to configure TLS with required mutual authentication between two hosts in LAN without DNS-server using https://github.com/grpc/grpc-java/tree/master/examples/src/main/java/io/grpc/examples/helloworldtls.
In my configuration server is accessed by IP so I've added subjectAltName to certificate generation script, see https://github.com/DmPanov/grpc-java/commit/0db1778c450d30d4470e92cec6ae3493d8eeb979

### What version of gRPC are you using?
1.12.0

### What did you expect to see?
I expect that this is a valid configuration

### What I get?

```
WARNING: RPC failed: Status{code=UNAVAILABLE, description=io exception, cause=javax.net.ssl.SSLHandshakeException: error:10000410:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.shutdownWithError(ReferenceCountedOpenSslEngine.java:862)
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.sslReadErrorResult(ReferenceCountedOpenSslEngine.java:1110)
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.unwrap(ReferenceCountedOpenSslEngine.java:1066)
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.unwrap(ReferenceCountedOpenSslEngine.java:1132)
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.unwrap(ReferenceCountedOpenSslEngine.java:1175)
	at io.netty.handler.ssl.SslHandler$SslEngineType$1.unwrap(SslHandler.java:216)
	at io.netty.handler.ssl.SslHandler.unwrap(SslHandler.java:1257)
	at io.netty.handler.ssl.SslHandler.decodeJdkCompatible(SslHandler.java:1159)
	at io.netty.handler.ssl.SslHandler.decode(SslHandler.java:1203)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:489)
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:428)
[...]

PR #4553, opened 2018-06-09
Use same CN on ca and client

bootstraponline:
This breaks the sample code.
SERVER_CN=example.com
CLIENT_CN=grpc_auth

Using the same CN in both CA and CLIENT fixes the error when using mutual TLS.

See https://github.com/grpc/grpc-swift/pull/244 for context.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -64,7 +64,7 @@ echo Generate CA key:
 openssl genrsa -passout pass:1111 -des3 -out ca.key 4096
 echo Generate CA certificate:
 # Generates ca.crt which is the trustCertCollectionFile
-openssl req -passin pass:1111 -new -x509 -days 365 -key ca.key -out ca.crt -subj "/CN=${SERVER_CN}"
+openssl req -passin pass:1111 -new -x509 -days 365 -key ca.key -out ca.crt -subj "/CN=${CLIENT_CN}"
 echo Generate server key:
 openssl genrsa -passout pass:1111 -des3 -out server.key 4096
 echo Generate server signing request:


thelinuxfoundation:

* User @bootstraponline isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

bootstraponline:
I signed it!

ejona86:
This doesn't look right. It seems like the CA CN could be anything. Maybe this would actually be good: change it to doesnot.matter.example.com.

I'm coming back from OOO, so I can't dig deeper to see why the CA's CN matters. But I'll try to take a glance later.

bootstraponline:
> This doesn't look right. It seems like the CA CN could be anything.

It can't. I have setup a demo project:

- https://github.com/bootstraponline/grpc_java_auth
- [openssl.sh key generation](https://github.com/bootstraponline/grpc_java_auth/blob/master/soseedygrpc/openssl/openssl.sh)

If the values don't match then the mutual auth fails.

> openssl req -new -x509 -days 365 -key ca.key -out ca.crt -subj "/CN=doesnot.matter.example.com"
>
> Caused by: javax.net.ssl.SSLHandshakeException: error:10000410:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE

ejona86:
I think the discussion from #4515 points to the better approach here. Basically, use ca.crt as the trust root on server-side instead of client.crt.

bootstraponline:
Thanks. I'll try that. If it works then I'll update the example.

bootstraponline:
That worked! I'll update the code.

- https://github.com/bootstraponline/grpc_java_auth/commit/ee196a9024a2b2cc1b8d1e5e78089e7f1c8254e0

bootstraponline:
https://github.com/grpc/grpc-java/pull/4561 has the correct fix.

PR #4555, opened 2018-06-11
all: fix lint warnings

dapengzhang0:


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1868,7 +1868,7 @@ private void assertServerStatsTrace(String method, Status.Code code,
       assertNotNull(serverEndRecord);
       checkStartTags(serverStartRecord, method);
       checkEndTags(serverEndRecord, method, code);
-      if (server != null && requests != null && responses != null) {
+      if (requests != null && responses != null) {
```
ericgribkoff: This is a lint fix? `server` is not final, so this is not redundant with the check on line 1851.

dapengzhang0: `server` is not volatile, so a second check does not help. I don't think it is a valid usecase that `server` can be changed back to `null` from `nonnull`.

ericgribkoff: Re: volatile: that's assuming none of the intervening methods have side-effects (e.g., this same thread could set `server` to null)... but I agree that this doesn't really matter here, and seems fine to change. I was more curious if an automatic lint check was actually complaining about this?

zpencer: Looking at the code, it doesn't look like `server` should change here.

dapengzhang0: It was not the internal lint check. It's from my local spotbug plug-in.

PR #4568, opened 2018-06-15
protobuf{,lite,nano}: make more classes final

carl-mastrangelo:
- Split anonymous classes in named and final classes
- Fix some Javadocs and annotate when things were added.


protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoLiteUtils.java
```diff
@@ -66,149 +67,31 @@
...
-        try {
-          stream.checkLastTagWas(0);
-          return message;
-        } catch (InvalidProtocolBufferException e) {
-          e.setUnfinishedMessage(message);
-          throw e;
-        }
-      }
-    };
+    return new ProtobufLiteMessageMarshaller<T>(defaultInstance);
```
dapengzhang0: nit: The class name can be shorter like `LiteMessageMarshaller` or even `MessageMarshaller`, as this is a private inner class, ProtobufLite is presumed. The same for `MetadataMarshaller`.

carl-mastrangelo: Done.

PR #4582, opened 2018-06-21
notes about musl: add hint to grpc-java alpine package

jstriebel:
Alpine has a `grpc-java` package with musl-compatible binaries (currently in edge). A note pointing to that might be helpful for alpine users.

It can be installed via `apk --no-cache add --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing grpc-java`.

Thanks for considering this!

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -331,6 +331,7 @@ If you are on a 32-bit operating system, or not on a [Transport Security support
 If you are using `musl` libc (e.g., with Alpine Linux), then
 `netty-tcnative-boringssl-static` won't work. There are several alternatives:
  - Use [netty-tcnative-alpine](https://github.com/pires/netty-tcnative-alpine)
+ - On Alpine Linux: Install the [grpc-java](https://pkgs.alpinelinux.org/package/edge/testing/x86_64/grpc-java) package from edge
  - Use a distribution with `glibc`
 
 If you are running inside of an embedded Tomcat runtime (e.g., Spring Boot),


carl-mastrangelo:
@jstriebel Hey, thanks for the PR.  I looked at that URL and it seems they are on an older version (1.11 vs. 1.12),   Do you know how frequently they publish artifacts?   I would prefer to recommend the latest and greatest if possible.

ejona86:
It looks like the package only contains protoc-gen-grpc-java. While that's useful, it's not directly helpful for tcnative concerns.
https://pkgs.alpinelinux.org/contents?branch=edge&name=grpc-java&arch=x86_64&repo=testing

jstriebel:
@carl-mastrangelo: For the versions, I dont't know how regularly this is updated.
@ejona86: True, that might not be as useful as I thought first, did not check it exactly. Feel free to close it, if it's not a concern.

dapengzhang0:
I think if we mention in the doc it's only a hint and may not be up-to-date, it's still useful.
> that might not be as useful as I thought first,

zpencer:
Since our artifacts do not work on musl there's still value in documenting how to use this alpine package. Let's move the note to `COMPILING.md` to tell alpine users that a pre-compiled package can be used like so:

```
protobuf {
  ...
  plugins {
    // Define a plugin with name 'grpc'
    grpc {
      path = '/usr/bin/protoc-gen-grpc-java'
    }
    ...
  }
  ...
}
```

ejona86:
@zpencer, COMPILING is for comping grpc, not for users of grpc. It seems like this would be better served in a different place.

zpencer:
In that case a `README_ALPINE.md` then?

ejona86:
Alpine-specific readme is fine. We may want to refer to it from the main README

ejona86:
Closing in favor of #4793

PR #4586, opened 2018-06-22
core: Add a toString to InProcessSocketAddress

kay:
The motivation here is that we want to write an interceptor to log the "remote-addr" attribute of a gRPC call. The attribute is a `SocketAddress` type which usually have reasonable `toString` methods but for the inprocess case we special case it. This PR is designed to provide a sensible `toString` for this case.


thelinuxfoundation:

* User @kay isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

kay:
I have now signed the CLA

carl-mastrangelo:
@kay Thanks!  This should be present in the 1.14 release.

kay:
Appreciated!

PR #4587, opened 2018-06-22
Update gson dependency to latest 2.8.5

njhill:


carl-mastrangelo:
Hi @njhill It looks like internally Google is using the old version.  We prefer to keep versions in sync, so let me see if I can do the internal upgrade before accepting this PR.

carl-mastrangelo:
@njhill I am still working on this.  A number of internal projects failed when trying to update (namely due to the reflection code in GSON).  I haven't forgotten about this though.

njhill:
Thanks @carl-mastrangelo, there's no rush from my side! I'd also just been tracking grpc's version until someone suggested upgrading, at which point I checked to see if there was any particular reason that it was back-level.

carl-mastrangelo:
I really hate to do this, but I'm going to close this PR as infeasible.   Trying to upgrade our internal gson dependency ran into some severe issues with some users expecting reflection to work a certain way.   I don't know if our internal users are a harbinger for external users, but it may be.    

I would *really* prefer to keep the github version the same as our internal version, so I don't know if this PR can be accepted (even though I strongly am in favor of upgrade).

njhill:
@carl-mastrangelo no problem, I'm sure it's not a big deal for others to override with a newer version if needed. It sounds like maybe the problem is on the gson side but I assume that grpc isn't going to remain depending on this version _forever_ while gson continues to progress?

carl-mastrangelo:
@njhill Yeah, that's right.   Gson reflection was pretty heavily modified for Java 9 support, which changed how the reflector (and class loading IIRC) worked.  Several internal teams depended on the exact behavior.   I filed bugs against them, so if they all fix I can reopen this.

We are still very happy to have you contributing @njhill !

PR #4588, opened 2018-06-22
services: binlog method name should include leading / char

zpencer:
Java is unlike the other implementations that our code strips the leading `/`.
When we log the method name, we should be consistent with the
other implementations and put back the `/`.

### DIFF
--- a/services/src/main/java/io/grpc/services/BinlogHelper.java
+++ b/services/src/main/java/io/grpc/services/BinlogHelper.java
@@ -119,7 +119,7 @@ final class BinlogHelper {
         CallId callId) {
       Preconditions.checkArgument(methodName == null || !isServer);
       Preconditions.checkArgument(timeout == null || !isServer);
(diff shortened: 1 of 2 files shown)


services/src/main/java/io/grpc/services/BinlogHelper.java
```diff
@@ -119,14 +119,17 @@ void logSendInitialMetadata(
         CallId callId) {
       Preconditions.checkArgument(methodName == null || !isServer);
       Preconditions.checkArgument(timeout == null || !isServer);
+      // In java, we strip the leading '/'. To be consistent with the rest of gRPC we must
```
ejona86: Java does not "strip" the leading slash. It just doesn't include it.

zpencer: I'll change the wording to be more clear. I was thinking of this code path https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/NettyServerHandler.java#L392

PR #4589, opened 2018-06-24
Make available the address the server is bound to

benmccann:
I'd like the address that the server was bound to. Right now there's no way to know if it's only accepting requests from localhost or if it's also accepting requests from other machines.

I'd like to add support for hot-reloading of GRPC servers. This feature means that if you've made a code change, when a request comes in the server will be recompiled and then continue with processing the request. This is a really nifty feature that Play Framework pioneered that requires integration with the build system (since the code needs to be recompiled). Gradle already has support for this feature built-in, but it currently expects the server to make available its address as an `InetSocketAddress`

P.S. hello @carl-mastrangelo! my first real GRPC PR :-) hope you've been well!

### DIFF
--- a/core/src/main/java/io/grpc/Server.java
+++ b/core/src/main/java/io/grpc/Server.java
@@ -17,6 +17,7 @@
 package io.grpc;
 
 import java.io.IOException;
+import java.net.InetSocketAddress;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
@@ -37,6 +38,16 @@ public abstract class Server {
    * @since 1.0.0
    */
   public abstract Server start() throws IOException;
+
+  /**
+   * Returns the address the server is listening on.  This can return null if there is no actual
+   * address or the result otherwise does not make sense.  Result is undefined after the server is
+   * terminated.
+   *
+   * @throws IllegalStateException if the server has not yet been started.
+   * @since 1.0.0
+   */
+  public abstract InetSocketAddress mainAddress();
 
   /**
    * Returns the port number the server is listening on.  This can return -1 if there is no actual
--- a/alts/src/main/java/io/grpc/alts/AltsServerBuilder.java
+++ b/alts/src/main/java/io/grpc/alts/AltsServerBuilder.java
@@ -203,6 +203,11 @@ public final class AltsServerBuilder extends ServerBuilder<AltsServerBuilder> {
     }
 
     @Override
+    public InetSocketAddress mainAddress() {
+      return delegate.mainAddress();
+    }
+
+    @Override
     public int getPort() {
       return delegate.getPort();
     }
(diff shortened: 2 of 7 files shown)


thelinuxfoundation:

* User @benmccann isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

benmccann:
I've signed the CLA

carl-mastrangelo:
@benmccann Hiya!  Some quick things on your PR:

* You can use NettyServerBuilder to pass a more specific listen address (like InetSocketAddress), and then you'll know for sure which address you are listening to
* `forPort` is currently part of our stable API, so we can't change the signature.  We can add new methods, however

benmccann:
Thanks for taking a quick look Carl! And on the weekend :-)

>You can use NettyServerBuilder to pass a more specific listen address (like InetSocketAddress), and then you'll know for sure which address you are listening to

Yeah. One thing that's a little tricky about that though is then I need to hold onto the address from creation time. I have the server other places in the code, but I don't necessarily have an easy way to pass data from creation time to all the other places the server is available

>forPort is currently part of our stable API, so we can't change the signature. We can add new methods, however

Good call. I'll leave that one alone

core/src/main/java/io/grpc/Server.java
```diff
@@ -38,6 +39,16 @@
...
 
+  /**
+   * Returns the address the server is listening on.  This can return null if there is no actual
+   * address or the result otherwise does not make sense.  Result is undefined after the server is
+   * terminated.
+   *
+   * @throws IllegalStateException if the server has not yet been started.
+   * @since 1.0.0
+   */
+  public abstract InetSocketAddress mainAddress();
```
carl-mastrangelo: unfortunately this is not stable as well.  Anyone extending Server would see compilation errors when they update to the latest gRPC.  We typically make new methods like this non-abstract and throw `UnsupportedOperationException`

carl-mastrangelo: gRPC also supports non TCP sockets (such as unix domain sockets), which would not be instances of `InetSocketAddress`.  Additionally, gRPC (will) support multiple listen addresses so there may not be one address.  You might consider making this:

```java
public List<SocketAddress> listeningAddresses() {
}
```

Then subclasses could override it appropriately.  Additionally, this would nicely fit with InprocessServerBuilder, as it could return an empty list.

benmccann:
Thanks for the great feedback Carl! I'm going to close this PR for now and see if I can support hot reloading without this change. Depending on how that goes I may come back later this week and reopen the PR addressing the comments you left

core/src/main/java/io/grpc/Server.java
ejona86: Carl's suggested API looks good to me as well. InProcessServerBuilder could actually return the InProcessSocketAddress.

Issue #4558, opened 2018-06-12
StreamObservers.copyWithFlowControl calls StreamObserver.onComplete twice


biran0079:
Client stream observer on ready handler is invoked after stream is half closed.
```java
      StreamObserver<HelloReply> resp = new ClientResponseObserver<HelloRequest, HelloReply>() {
        @Override
        public void onNext(HelloReply value) {

        }

        @Override
        public void onError(Throwable t) {
          t.printStackTrace();
        }

        @Override
        public void onCompleted() {
        }

        @Override
        public void beforeStart(ClientCallStreamObserver<HelloRequest> requestStream) {
          StreamObservers.copyWithFlowControl(
              ImmutableList.of(HelloRequest.newBuilder().setName(payload).build()), 
              requestStream);
        }
      };
     stub.sayHello(resp);
```
Following error is found:

```
> io.grpc.StatusRuntimeException: CANCELLED: Failed to call onReady.
	at io.grpc.Status.asRuntimeException(Status.java:526)
	at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:419)
	at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:37)
	at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
	at io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
	at io.grpc.internal.CensusStatsModule$StatsClientInterceptor$1$1.onClose(CensusStatsModule.java:684)
	at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:37)
	at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
	at io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
	at io.grpc.internal.CensusTracingModule$TracingClientInterceptor$1$1.onClose(CensusTracingModule.java:391)
	at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:471)
	at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:63)
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl.close(ClientCallImpl.java:553)
[...]

PR #4592, opened 2018-06-26
stub: clarify StreamObservers interaction

carl-mastrangelo:
Updates #4558

It appears in the linked bug that onError is called from outside.  I don't believe this is safe because thew onReadyCallback is going to call onCompleted.  In the case of an error, it seems like the iterator should be stopped, but there isn't a great way to phrase this.  Ideas welcome.


ejona86:
If we're worried about exceptions, it should call `target.onError(ex)` or similar.

carl-mastrangelo:
@ejona86 ack.  I don't know if that is the case,  @biran0079 can say more definitively

PR #4597, opened 2018-06-28
core: Avoid unnecessary unchecked cast in DnsNameResolver

ejona86:


core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -515,15 +515,13 @@ private ResourceResolver getResourceResolver() {
...
   @VisibleForTesting
   static ResourceResolverFactory getResourceResolverFactory(ClassLoader loader) {
     Class<? extends ResourceResolverFactory> jndiClazz;
     try {
       jndiClazz =
-          (Class<? extends ResourceResolverFactory>)
-              Class.forName("io.grpc.internal.JndiResourceResolverFactory", true, loader);
-      assert ResourceResolverFactory.class.isAssignableFrom(jndiClazz);
+          Class.forName("io.grpc.internal.JndiResourceResolverFactory", true, loader)
+              .asSubclass(ResourceResolverFactory.class);
```
dapengzhang0: `asSubclass()` throws `ClassCastException`, should it be catch-and-logged?

carl-mastrangelo: There was some reason why this wasn't feasible when I wrote this and ended up back with the cast.  I guess whatever reason is gone now.

PR #4601, opened 2018-06-29
core: make InternalHandlerRegistry a HandlerRegistry

carl-mastrangelo:
Also some minor formatting changes.


carl-mastrangelo:
@dapengzhang0 main perk is discoverability.  Also it gives credence that our APIs are good.

dapengzhang0:
Actually I might need this change in my servlet implementation, because I need access to this registry from a different package like `io.grpc.servlet.*`. Please ship it.

Issue #4302, opened 2018-04-04
Determine final NameResolver to LoadBalancer communication


ejona86:
Alluded to in https://github.com/grpc/grpc-java/pull/4137#discussion_r171640085 . The biggest issue is probably how Java allows a list of lists of addresses. This is different than the other implementations. We need to come to a cross-language agreement about whether list of lists is necessary.

But we also need to design the final solution for how proxy information should be plumbed from the NameResolver through the LoadBalancer into Subchannel. That will remove `PairSocketAddress` or stabilize it (with a better name!).

CC @zhangkun83

ejona86:
This issue is no longer impacted by the design of proxy information.

@zhangkun83 and I discussed how proxy will be handled. We feel good about having a `ProxySocketAddress` that wraps a `SocketAddress` (typically an `InetSocketAddress`). We're also comfortable doing this strategy in other cases where the extra information is required to properly use the address.

The case that isn't handled well is when this additional information is _optional_, because that forces all downstream consumers of the `FooSocketAddress` to understand it, which isn't really acceptable. Right now, we believe "this case is unlikely" because each address in an EAG is supposed to be pretty equivalent (like two addresses to the same host) and it seems really weird to have optional information at this level. If that turns out to be incorrect, we could introduce an `OptionalInfoSocketAddress` interface that has a `getNestedSocketAddress()` which would allow downstream consumers to unwrap the address independent of whatever new optional information is deemed necessary. That would be sub-optimal, but quite good as a fall-back when our design assumptions fail.

The "each address in an EAG is supposed to be pretty equivalent" is a false assumption today, as `PickFirstLoadBalancer` merges multiple EAGs into a single EAG to pass to `LoadBalancer.Helper.createSubchannel(EquivalentAddressGroup addrs, Attributes attrs)`. This will be solved by avoiding the merge with a new method for the purpose: `LoadBalancer.Helper.createSubchannel(List<EquivalentAddressGroup> addrs, Attributes attrs)`.

PR #4603, opened 2018-06-29
core: Add support for List<EAG> in Subchannels

ejona86:
This avoids the needs to flatten to EAGs for cases like PickFirst,
making the Attributes in EAGs able to be used in communication with
core. See #4302 for some discussion on the topic.

-----

This alone will not be enough to rid ourselves of all flattening. We also
need to support List\<EAG> with OobChannels. But I'll do that as a later
PR.

### DIFF
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -19,6 +19,7 @@ package io.grpc;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
+import java.util.Collections;
 import java.util.List;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -703,4 +711,68 @@ public void closed(
...
+    }
+
+    /** Update to new groups, resetting the current index. */
+    public void updateGroups(List<EquivalentAddressGroup> newGroups) {
+      addressGroups = newGroups;
+      reset();
+    }
+
+    /** Returns false if the needle was not found and the current index was left unchanged. */
+    public boolean seekTo(SocketAddress needle) {
```
zhangkun83: Interesting argument naming :)

carl-mastrangelo: Possibly inspired by http://php.net/manual/en/function.strpos.php ?

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -455,21 +456,61 @@ public boolean equals(Object other) {
...
+    /**
+     * Creates a Subchannel, which is a logical connection to the given group of addresses which are
+     * considered equivalent.  The {@code attrs} are custom attributes associated with this
+     * Subchannel, and can be accessed later through {@link Subchannel#getAttributes
+     * Subchannel.getAttributes()}.
+     *
+     * <p>The LoadBalancer is responsible for closing unused Subchannels, and closing all
+     * Subchannels within {@link #shutdown}.
+     *
+     * @throws IllegalArgumentException if {@code attrs} is empty
```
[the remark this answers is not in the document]
ejona86: Done. Very good catch. Amazingly good catch.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
ejona86: Quit trolling @carl-mastrangelo! :smile: Choose a more respectable source, like http://man7.org/linux/man-pages/man3/strstr.3.html !

PR #4607, opened 2018-07-01
Remove redundant SubchannelPicker refreshes in RoundRobinLoadBalancer

njhill:
The "active" list and corresponding circular index used by `RoundRobinLoadBalancer`'s picker is currently refreshed via a call to `helper.updateBalancingState()` after _any_ change to a Subchannel's state or the resolved address groups, regardless of whether that change actually results in a change to the contents of the list.

In fact it appears that the _majority_ of Picker refreshes may be no-ops - for example any time a new EAG is discovered while there are already active Subchannels, its new Subchannel is added in `IDLE` state followed by an immediate Picker refresh which will be unnecessary because the list only contains `READY` Subchannels. A similar pattern occurs during Subchannel removal/shutdown.

These avoidable perturbations to the load distribution may be exacerbating the behaviour described in https://github.com/grpc/grpc-java/pull/4462.

This PR makes changes to ensure this doesn't happen, and includes:
- Making it so that a Subchannel exists in the `subchannels` map if and only if its state is not `SHUTDOWN`. This also facilitates a simplification to the stickiness picker logic where list containment no longer needs to be checked when (re)registering a stickiness value.
- Adding an `EmptyPicker` class, since the picker parameters and logic are disjoint from the non-empty case


njhill:
@zhangkun83 any interest in this change?

njhill:
Thanks @carl-mastrangelo. I noticed your PR #4643 has some bits in common with this - might it make sense for `EmptyPicker` and `FailedPicker` to be protected static classes in the `LoadBalancer` superclass?

njhill:
@carl-mastrangelo @dapengzhang0 sorry to bug, wondering if you might have a chance to look at this?

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -173,10 +178,13 @@ public void handleResolvedAddressGroups(
       // Shutdown subchannels for removed addresses.
       for (EquivalentAddressGroup addressGroup : removedAddrs) {
         Subchannel subchannel = subchannels.remove(addressGroup);
-        subchannel.shutdown();
+        update = update || isReady(subchannel); // no need to update if channel was already excluded
```
carl-mastrangelo: Isn't this missing for the `addedAddrs` ?  i.e. why only update on removing addresses?

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -186,23 +194,40 @@ public void handleNameResolutionError(Status error) {
 
     @Override
     public void handleSubchannelState(Subchannel subchannel, ConnectivityStateInfo stateInfo) {
-      if (stateInfo.getState() == SHUTDOWN && stickinessState != null) {
+      Ref<ConnectivityStateInfo> stateInfoRef = getSubchannelStateInfoRef(subchannel);
+      ConnectivityState stateBefore = stateInfoRef.value.getState();
+      if (stateBefore == SHUTDOWN) {
```
carl-mastrangelo: This should be an error, as there should be no more events after the SC has shut down.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -186,23 +194,40 @@ public void handleNameResolutionError(Status error) {
...
+        return;
+      }
+      ConnectivityState newState = stateInfo.getState();
+      if (newState == SHUTDOWN && stickinessState != null) {
         stickinessState.remove(subchannel);
       }
       if (subchannels.get(subchannel.getAddresses()) != subchannel) {
```
carl-mastrangelo: @zhangkun83 How can this case happen?

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -186,23 +194,40 @@ public void handleNameResolutionError(Status error) {
...
         return;
       }
-      if (stateInfo.getState() == IDLE) {
+      if (newState == IDLE) {
         subchannel.requestConnection();
```
carl-mastrangelo: @zhangkun83 isn't this also redundant?  When a new channel is created the connection is already requested.  What will calling it twice do?  Also, won't this prevent channels from ever going idle?

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -186,23 +194,40 @@ public void handleNameResolutionError(Status error) {
...
       }
-      getSubchannelStateInfoRef(subchannel).value = stateInfo;
-      updateBalancingState(getAggregatedState(), getAggregatedError());
+      stateInfoRef.value = stateInfo;
+      if (!ready || (newState == READY ^ stateBefore == READY)) {
+        updateBalancingState(null, null);
```
carl-mastrangelo: nit: we generally avoid using literals that are not obvious.  IN this case, consider naming the literal nulls like:

`updateBalancingState(/*connectivityState= */ null, /* status= */ null);`

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -211,21 +236,30 @@ public void shutdown() {
      */
     private void updateBalancingState(ConnectivityState state, Status error) {
       List<Subchannel> activeList = filterNonFailingSubchannels(getSubchannels());
-      helper.updateBalancingState(state, new Picker(activeList, error, stickinessState));
+      ready = !activeList.isEmpty();
```
carl-mastrangelo: The logic here is starting to get complicated, and it is no longer "obviously" correct.  Can you please document the params?

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -211,21 +236,30 @@ public void shutdown() {
...
+      if (state == null) {
+        state = ready ? READY : getAggregatedState();
+      }
+      SubchannelPicker picker = ready ? new Picker(activeList, stickinessState) :
+          new EmptyPicker(error != null ? error : getAggregatedError());
```
carl-mastrangelo: there is a lot of conditional inline logic here which is hard to visually follow.  For example, why does `error` take precedence over aggregated error?  If the params are meant to be optional, it would be better expressed as an overload.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -211,21 +236,30 @@ public void shutdown() {
...
     /**
      * Filters out non-ready subchannels.
      */
     private static List<Subchannel> filterNonFailingSubchannels(
         Collection<Subchannel> subchannels) {
-      List<Subchannel> readySubchannels = new ArrayList<Subchannel>(subchannels.size());
+      List<Subchannel> readySubchannels = null;
       for (Subchannel subchannel : subchannels) {
-        if (getSubchannelStateInfoRef(subchannel).value.getState() == READY) {
+        if (isReady(subchannel)) {
```
carl-mastrangelo: @zhangkun83 isn't this going to be an expensive linear time operation?  Why is the list scanned each time, rather than Subchannel state change?

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
njhill: The main intention is to avoid calling `updateBalancingState()` in cases that won't result in any change to the Picker state, particularly when it already has active endpoints.

Any new `addedAddrs` are added in `IDLE` state and so will have no effect on the Picker state if already has at least one `READY` subchannel (and if it doesn't then `ready` will be false and so we update anyhow).

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
njhill: The state in the subchannel attribute is now set to `SHUTDOWN` when the EAG is removed in the `handleResolvedAddressGroups` callback (at the same time as `subchannel.shutdown()` is called and that it's removed from this LB's `subchannels` map - see line 220).

This is because from the load balancer's pov it makes more sense to consider it in a non-ready state as soon as the `NameResolver` reports it removed, rather than when a subsequent event is received that the shutdown is complete (i.e. here).

So this is really catching the "normal" `SHUTDOWN` state change following the subchannel's prior removal.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
njhill: Agree, I should restructure this to be more explicit/readable.

Some of it is to avoid unnecessary computation - for example `getAggregatedState()` is only applicable when the filtered `activeList` is empty, and not in the `handleNameResolutionError()` callback case. Similar thing for `error` -  it only applies in the empty active list case, where it's determine by `getAggregatedError()` _unless_ this is triggered by `handleNameResolutionError()` in which case the name resolver error is relayed.

njhill:
Thanks @carl-mastrangelo @zhangkun83 for looking at this.

> have you considered just changing updateBalancingState() so that it saves the current picker, compares the new picker with the current one, and only calls helper.updateBalancingState() if they differ?

Yes this was an initial thought, but it seemed kind of wrong/wasteful for me to regenerate and compare the new picker in all these cases where it's easily known easily at the point of the change that it will have no impact on the picker state, i.e. in the majority of `handleResolvedAddressGroups()` cases.

I will merge latest from master in, and have a go at making the logic simpler / easier to read. Part of the problem may actually be that I was trying to minimize the invasiveness by tweaking the existing structure. If that still isn't palatable then I agree just doing the picker comparison would be a much smaller change and will still achieve the primary goal.

ejona86:
Ping @zhangkun83

zhangkun83:
Thanks for the patience @njhill. I am not excited about the added complexity. I would prefer doing the picker comparison.

zhangkun83:
@njhill, are you still working on this?

njhill:
@zhangkun83 yes, apologies I was out travelling for a couple of days and bit behind with things, I will get back to it soon hopefully (if not this week then next week).

zhangkun83:
No worries. Just checking the status on your side :-)

njhill:
@zhangkun83 I've now updated this to save the current picker but instead of doing a full re-generation and comparison of the READY subchannel list every time, it uses a readily-accessible determination of whether the list has changed or not. I've added some comments and restructured a little to make this clearer.

It also more cleanly separates the different cases of a "ready" picker, an "empty" picker (which may or may not have an associated error), and a name resolution error. For example there is no need to touch the picker on name resolution errors as is currently done.

Please let me know what you think!

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -172,54 +175,90 @@ public void handleResolvedAddressGroups(
...
       for (EquivalentAddressGroup addressGroup : removedAddrs) {
         Subchannel subchannel = subchannels.remove(addressGroup);
-        subchannel.shutdown();
+        // Active list only changes if we remove a subchannel in READY state
+        activeListChanged = activeListChanged || isReady(subchannel);
+        shutdownSubchannel(subchannel);
       }
 
-      updateBalancingState(getAggregatedState(), getAggregatedError());
+      updateBalancingState(activeListChanged || stickinessStateChanged);
```
zhangkun83: While the logic is clear enough as it is, I still don't like the little addition of complexity. The list change handling is not in the RPC critical path, and the overhead of creating and discarding a new picker doesn't seem to be significant enough to justify the added code complexity, e.g., the boolean argument here passes ad-hoc information across function boundaries just because your logic has to be divided into multiple functions instead of being one -- not saying it's wrong, but I'd like to avoid such situation if possible.

njhill: Sure, I will rework to always re-generate and compare the pickers.

njhill:
@zhangkun83 now updated to do picker comparison, PTAL at your convenience!

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -198,28 +211,68 @@ public void handleSubchannelState(Subchannel subchannel, ConnectivityStateInfo s
...
+      stateInfoRef.value = stateInfo;
+      updateBalancingState();
+    }
+
+    private void shutdownSubchannel(Subchannel subchannel) {
+      subchannel.shutdown();
+      getSubchannelStateInfoRef(subchannel).value =
+          ConnectivityStateInfo.forNonError(SHUTDOWN);
+      if (stickinessState != null) {
+        stickinessState.remove(subchannel);
```
zhangkun83: This branch is [not covered](https://coveralls.io/builds/18764189/source?filename=core%2Fsrc%2Fmain%2Fjava%2Fio%2Fgrpc%2Futil%2FRoundRobinLoadBalancerFactory.java#L223).

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -198,28 +211,68 @@ public void handleSubchannelState(Subchannel subchannel, ConnectivityStateInfo s
...
+
+    private static boolean areEquivalentPickers(SubchannelPicker p1, SubchannelPicker p2) {
+      if (p1 == p2) {
+        return true;
+      }
+      if (p1.getClass() != p2.getClass()) {
+        return false;
+      }
+      if (p1 instanceof EmptyPicker) {
+        return Objects.equal(((EmptyPicker)p1).status, ((EmptyPicker)p2).status);
```
zhangkun83: Please add a space following the closing parenthesis of class casting.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -198,28 +211,68 @@ public void handleSubchannelState(Subchannel subchannel, ConnectivityStateInfo s
...
-          state,
-          new Picker(activeList, error, startIndex, stickinessState));
+      if (activeList.isEmpty()) {
+        // empty picker returns error or no result
+        updateBalancingState(getAggregatedState(), new EmptyPicker(getAggregatedError()));
+      } else {
+        // initialize the Picker to a random start index to ensure that a high frequency of Picker
+        // churn does not skew subchannel selection.
+        int startIndex = random.nextInt(activeList.size());
+        updateBalancingState(READY, new ReadyPicker(activeList, startIndex, stickinessState));
```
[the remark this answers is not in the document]
njhill: I've now combined/collapsed both the `getAggregatedState()` and `getAggregateError()` methods into `updateBalancingState()`.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -198,28 +211,68 @@ public void handleSubchannelState(Subchannel subchannel, ConnectivityStateInfo s
...
+
+    private void updateBalancingState(ConnectivityState state, SubchannelPicker picker) {
+      if (state != currentState || !areEquivalentPickers(picker, currentPicker)) {
+        helper.updateBalancingState(state, picker);
+        currentState = state;
+        currentPicker = picker;
+      }
+    }
+
+    private static boolean areEquivalentPickers(SubchannelPicker p1, SubchannelPicker p2) {
```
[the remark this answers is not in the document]
njhill: I refactored this in the latest commit, moving the logic to a new private abstract class method `RoundRobinPicker.isEquivalentTo()` (to help make it clearer that it only applies to the private RRLB pickers and not `SubchannelPicker`s in general).

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
njhill: I've extended a unit test to cover this case - name resolver removal of a sticky subchannel.

njhill:
@zhangkun83 thanks for the detailed review. Have made some updates which hopefully address your comments. One of the builds failed but I don't think it was code related.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -451,9 +451,36 @@ private Subchannel nextSubchannel() {
...
+    }
+  }
+
+  @VisibleForTesting
+  static final class EmptyPicker extends RoundRobinPicker {
+    @Nullable
+    private final Status status;
+
+    EmptyPicker(@Nullable Status status) {
+      this.status = status;
```
zhangkun83: Making it non-Nullable, and using OK to indicate `PickResult.withNoResult()` may simplify the code a little bit.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -384,59 +394,49 @@ Subchannel getSubchannel(String stickinessValue) {
...
     @Nullable
     private final RoundRobinLoadBalancer.StickinessState stickinessState;
     @SuppressWarnings("unused")
     private volatile int index;
 
-    Picker(
-        List<Subchannel> list, @Nullable Status status, int startIndex,
+    ReadyPicker(List<Subchannel> list, int startIndex,
         @Nullable RoundRobinLoadBalancer.StickinessState stickinessState) {
+      assert !list.isEmpty();
```
zhangkun83: This should be a checkArgument(). If this did happen in production and it was let pass this point, it would cause an out-of-bound exception at a later point.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancerFactory.java
```diff
@@ -451,9 +451,36 @@ private Subchannel nextSubchannel() {
       return list;
     }
 
-    @VisibleForTesting
-    Status getStatus() {
-      return status;
+    @Override
+    boolean isEquivalentTo(RoundRobinPicker picker) {
+      if (!(picker instanceof ReadyPicker)) {
+        return false;
```
zhangkun83: This is not covered.

njhill:
@zhangkun83 I've addressed your latest comments, PTAL. In particular there are now no new uncovered lines. Thanks!

njhill:
Thanks @zhangkun83, have now addressed your remaining minor comment.

zhangkun83:
@carl-mastrangelo do you want to take another look?

carl-mastrangelo:
@njhill Merged, thanks!

Sorry for the excessively long review process!  Hopefully this hasn't scared you off; we do appreciate your contributions.   This will be released in 1.16 in about 6 weeks.

PR #4608, opened 2018-07-02
core: add jndi timeout and dirContext.close

dapengzhang0:


dapengzhang0:
\+ @ejona86 as @carl-mastrangelo is OOO

Issue #3173, opened 2017-07-05
Create @GrpcClass annotation processor hook for generated source files.


daicoden:
To write an annotation processor that further augments the grpc protoc plugin is a little difficult.

Currently, I process `@Generated` annotations and search for `value ~= /gRpc/`.

As an example, looking at the new grpc-dagger in 2.11 (google/dagger#647), the additional dagger code generation is triggered off of an annotation applied to the implementing class. It would be nice to have all the code generation based off of the gprc classes.  Right now it feels like a two step process.  Add the proto, generate the grpc stub, implement the stub, add the processor annotation, compile again for the additional code.

WDYT?

carl-mastrangelo:
I'm not familiar with annotation processors, what would adding a special annotation do that would ease this process?

Also, as an alternative, why not generate your own stubs (as a protoc plugin) if you want to extend them?

daicoden:
I'm trying to use the grpc generated output so I can use other libraries which are based off of that, dagger-grpc is the only external use case I have right now... but I've written some additional generation that might be useful to others. Do you think most users will write their own protoc plugin for stub generation?

For the annotation processor, I'm not an expert either, but you basically create an annotation which tags elements, the processor can retrieve those elements, and the processor generates additional code. I can find the classes, for my own code generation, by scanning and name matching. However, if you envision annotation processing as a way to extend grpc functionality, it would be nice to have some hook for annotation processors to avoid the two cycle compile dance described above, and have one consistent way to find the grpc stub classes.

Other ideas:
* Can you chain protoc plugins together? If so, that could solve the problem, and could be the recommended way for third parties to extend.
* Can you pass arguments into the protoc plugin? If so, we could make an option like `-DannotateStub=example.custom.Annotation`.
* If writing protoc plugins is the recommended way for generated extension, can we somehow share the c++ which maintains the API to grpc, i.e. if someone renames io.grpc.stub.ServerCalls.UnaryCall, or changes the template arguments, or maybe collapses `ServerStreamingMethod` and `UnaryMethod` into `UnaryRequestMethod` etc, those changes will propagate. Maybe share the `method struct`, and common interpolation values in `vars`...

carl-mastrangelo:
>  two cycle compile dance described above.

I don't understand,  why are there two compilation steps?  

>  Add the proto, generate the grpc stub, implement the stub, add the processor annotation, compile again for the additional code.

Why do you have to implement the stub before processing annotations?

[...]

PR #4609, opened 2018-07-02
compiler, stub: Add @RpcMethod annotation (#3173)

jbingham-google:
This annotation will enable Java APT to generate code.

### DIFF
--- a/alts/src/generated/main/grpc/io/grpc/alts/internal/HandshakerServiceGrpc.java
+++ b/alts/src/generated/main/grpc/io/grpc/alts/internal/HandshakerServiceGrpc.java
@@ -30,7 +30,9 @@ public final class HandshakerServiceGrpc {
   private static volatile io.grpc.MethodDescriptor<io.grpc.alts.internal.Handshaker.HandshakerReq,
       io.grpc.alts.internal.Handshaker.HandshakerResp> getDoHandshakeMethod;
 
-  @io.grpc.stub.RpcMethod(
+  @io.grpc.stub.annotations.RpcMethod(
+      fullServiceName = "grpc.gcp.HandshakerService",
+      methodName = "DoHandshake",
       inputType = io.grpc.alts.internal.Handshaker.HandshakerReq.class,
       outputType = io.grpc.alts.internal.Handshaker.HandshakerResp.class,
       methodType = io.grpc.MethodDescriptor.MethodType.BIDI_STREAMING)
--- a/alts/src/generated/main/grpc/io/grpc/alts/internal/HandshakerServiceGrpc.java
+++ b/alts/src/generated/main/grpc/io/grpc/alts/internal/HandshakerServiceGrpc.java
@@ -30,6 +30,10 @@ public final class HandshakerServiceGrpc {
   private static volatile io.grpc.MethodDescriptor<io.grpc.alts.internal.Handshaker.HandshakerReq,
       io.grpc.alts.internal.Handshaker.HandshakerResp> getDoHandshakeMethod;
 
+  @io.grpc.stub.RpcMethod(
+      inputType = io.grpc.alts.internal.Handshaker.HandshakerReq.class,
+      outputType = io.grpc.alts.internal.Handshaker.HandshakerResp.class,
+      methodType = io.grpc.MethodDescriptor.MethodType.BIDI_STREAMING)
   public static io.grpc.MethodDescriptor<io.grpc.alts.internal.Handshaker.HandshakerReq,
       io.grpc.alts.internal.Handshaker.HandshakerResp> getDoHandshakeMethod() {
     io.grpc.MethodDescriptor<io.grpc.alts.internal.Handshaker.HandshakerReq, io.grpc.alts.internal.Handshaker.HandshakerResp> getDoHandshakeMethod;
--- a/compiler/src/java_plugin/cpp/java_generator.cpp
+++ b/compiler/src/java_plugin/cpp/java_generator.cpp
@@ -478,6 +478,8 @@ static void PrintMethodFields(
           "    $output_type$> $method_new_field_name$;\n"
           "\n"
           "@$RpcMethod$(\n"
+          "    fullServiceName = \"$Package$$service_name$\",\n"
+          "    methodName = \"$method_name$\",\n"
           "    inputType = $input_type$.class,\n"
           "    outputType = $output_type$.class,\n"
           "    methodType = $MethodType$.$method_type$)\n"
@@ -1279,7 +1281,7 @@ void GenerateService(const ServiceDescriptor* service,
   vars["ProtoMethodDescriptorSupplier"] =
       "io.grpc.protobuf.ProtoMethodDescriptorSupplier";
   vars["AbstractStub"] = "io.grpc.stub.AbstractStub";
-  vars["RpcMethod"] = "io.grpc.stub.RpcMethod";
+  vars["RpcMethod"] = "io.grpc.stub.annotations.RpcMethod";
   vars["MethodDescriptor"] = "io.grpc.MethodDescriptor";
   vars["NanoUtils"] = "io.grpc.protobuf.nano.NanoUtils";
   vars["StreamObserver"] = "io.grpc.stub.StreamObserver";
--- a/compiler/src/java_plugin/cpp/java_generator.cpp
+++ b/compiler/src/java_plugin/cpp/java_generator.cpp
@@ -477,6 +477,10 @@ static void PrintMethodFields(
           "private static volatile $MethodDescriptor$<$input_type$,\n"
           "    $output_type$> $method_new_field_name$;\n"
           "\n"
+          "@$RpcMethod$(\n"
+          "    inputType = $input_type$.class,\n"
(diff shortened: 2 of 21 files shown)


thelinuxfoundation:

_Regards,
CLA GitHub bot_

thelinuxfoundation:

* User @jbingham-google isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

thelinuxfoundation:

* User @jbingham-google isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

compiler/src/java_plugin/cpp/java_generator.cpp
```diff
@@ -477,6 +477,10 @@ static void PrintMethodFields(
           "private static volatile $MethodDescriptor$<$input_type$,\n"
           "    $output_type$> $method_new_field_name$;\n"
           "\n"
+          "@$RpcMethod$(\n"
```
ejona86: I think we need to include the method name as well, because the method name → Java method name conversion is lossy. We should either include the "full method name" or the "full service name"+"method name".

stub/src/main/java/io/grpc/stub/RpcMethod.java
```diff
@@ -0,0 +1,46 @@
...
+import io.grpc.MethodDescriptor.MethodType;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotates a method descriptor method to provide metadata for annotation processing.
+ */
+@Retention(RetentionPolicy.CLASS)
```
ejona86: We would normally annotate this initially with `@ExperimentalApi`. That might cause some problem with a experimental api checker. @ericgribkoff can help figure out what to do here.

stub/src/main/java/io/grpc/stub/RpcMethod.java
```diff
@@ -0,0 +1,46 @@
...
+@Target(ElementType.METHOD)
+public @interface RpcMethod {
```
ejona86: I'm not wild about the "RpcMethod" name, since we don't have anything "Rpc"-prefixed already. ... but I also can't think of anything better. @carl-mastrangelo, @zhangkun83, @dapengzhang0, @ericgribkoff, @zpencer, any thoughts on another name? We could throw it in a annotations package, but we'd still want the name to avoid colliding.

zhangkun83: +1 for annotations package, then I am fine with `RpcMethod`.

jbingham-google: Moved to stub.annotations package.

stub/src/main/java/io/grpc/stub/RpcMethod.java
jbingham-google: Added annotations, we can see if it needs to be removed.

compiler/src/java_plugin/cpp/java_generator.cpp
jbingham-google: Done.

stub/src/main/java/io/grpc/stub/RpcMethod.java
ericgribkoff: Yes, as-is this will trigger the grpc-java-api-checker tool to warn that user code (the generated code) is invoking `@ExperimentalApi`.

@ejona86 If other libraries are shipping gRPC's generated code, how does annotating part of that code as "experimental" work? If we remove/rename this class in the future, the previously generated code will break.

ejona86: Yeah, I agree. It can't be experimental because the generated code won't compile without it. I realized that yesterday. The earlier issue we had with the MethodDescriptors was a special case.

jbingham-google: Removed the annotation.

stub/src/main/java/io/grpc/stub/RpcMethod.java
ejona86: I guess I'm fine with the name. If we want another one on the service it would be RpcService. The only main competitor would probably be "GrpcMethod" and "GrpcService", although I would also sort of expect that to be more likely to collide with others (users of gRPC that make their own annotations for injection-based systems). So I think the "Rpc" is good.

alts/src/generated/main/grpc/io/grpc/alts/internal/HandshakerServiceGrpc.java
```diff
@@ -30,6 +30,12 @@ private HandshakerServiceGrpc() {}
   private static volatile io.grpc.MethodDescriptor<io.grpc.alts.internal.Handshaker.HandshakerReq,
       io.grpc.alts.internal.Handshaker.HandshakerResp> getDoHandshakeMethod;
 
+  @io.grpc.stub.annotations.RpcMethod(
+      fullServiceName = "grpc.gcp.HandshakerService",
+      methodName = "DoHandshake",
+      inputType = io.grpc.alts.internal.Handshaker.HandshakerReq.class,
```
carl-mastrangelo: personal preference: I like `requestType` and `responseType`.  Input and output don't match any of our other verbiage.

stub/src/main/java/io/grpc/stub/annotations/RpcMethod.java
```diff
@@ -0,0 +1,56 @@
...
+
+import io.grpc.MethodDescriptor.MethodType;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotates a method descriptor method to provide metadata for annotation processing.
+ */
```
carl-mastrangelo: this needs a `@since  1.14.0`

stub/src/main/java/io/grpc/stub/annotations/RpcMethod.java
```diff
@@ -0,0 +1,56 @@
...
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.METHOD)
+public @interface RpcMethod {
```
carl-mastrangelo: The usage of this is not super clear to me.  For example, there are several other fields in MethodDescriptor that weren't included (like schema descriptor).  It needs to be clear that this is a limitted subset, and the source of truth is the getter itself.

jbingham-google: When using a compile time annotation processor you only have access to limited metadata.  Method names, class names, class variable names, parameter types, return types, and annotations but the class isn't loadable so you can't execute the code.  This annotation won't be available at runtime (Retention = CLASS) but it is available for annotation processing.  They are both sources of truth but each is provided for a different (and in fact, mutually exclusive) use case

alts/src/generated/main/grpc/io/grpc/alts/internal/HandshakerServiceGrpc.java
ejona86: Good point. Yeah, we should fix that.

stub/src/main/java/io/grpc/stub/annotations/RpcMethod.java
carl-mastrangelo: What I am worried about is misinterpretted intents.  There is very little documentation in the class about how it *should* be used.   Can anyone outside of grpc use this?  Could their annotation processor generate their own MethodDescriptors from this?  

If this really is only for annotation processors, why is MethodType included?  Where is there an example of how this will be used?

jbingham-google: Yes, anyone outside of grpc could use this (but only at compile time, not runtime).

Method type is the only field that is truly necessary (the others can be derived, albeit very crudely) for the annotation processor since there is no way today of distinguishing between client streaming and bidi streaming methods using the metadata available to an annotation processor.  Unary and server streaming can be distinguished by looking at the signatures in the BlockingStub class.

An example of where this could be useful is if you generated the implementation of the ImplBase to invoke some piece of common code you had with separate methods for each method type.

ejona86: Maybe something like: ?

> Annotates a method descriptor method to provide metadata for annotation processing. This can be used by annotation processors to discover gRPC information, as the MethodDescriptors are not available when the annotation processor runs.

ejona86:
API Review:
 * inputType/outputType is better as requestType/responseType (already changed)
 * want to use "fullMethodName" instead of the split strings, because that matches MethodDescriptor
 * Want to default request/responseType to null. Mention that it relates to ReflectableMarshaller
 * Declare in JavaDoc methodType,fullMethodName is not null (double-checking whether it is possible to use null already).

With those tweaks, it LGTU.

Also, we'll be adding a MethodDescriptor.extractFullMethodName

Issue #3528, opened 2017-10-01
error usage of clearTextUpgrade in Netty


xujianhai:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?
v1.7.x

### What JVM are you using (`java -version`)?
java version "1.8.0_91"
Java(TM) SE Runtime Environment (build 1.8.0_91-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.91-b14, mixed mode)

### What did you do?
If possible, provide a recipe for reproducing the error.
that is error usage in CleartextUpgrade iof Netty n client side,

xujianhai:
see commit  : https://github.com/grpc/grpc-java/pull/3529/commits

ejona86:
It seems like this is trying to say client-side Upgrade support is broken. And the fix is in `PlaintextUpgradeNegotiator`:
```patch
-      return new BufferingHttp2UpgradeHandler(upgrader);
+      return new BufferingHttp2UpgradeHandler(httpClientCodec, upgrader);
```

PR #4612, opened 2018-07-03
Client-side support for h2c via Upgrade #4518

### LINKED ISSUE #4518: Client-side h2c (insecure HTTP/1.1-to-h2) upgrade does not work
When specifying `NegotiationType.PLAINTEXT_UPGRADE`, the client should connect over HTTP/1.1 with headers indicating that this request should be upgraded to (still insecure) h2.

There seems to be quite a bit of machinery in place to make this work, but currently the client does not start sending the request after performing the TCP handshake. A possible fix for this has been previously proposed in https://github.com/grpc/grpc-java/pull/3529/files#diff-e805c3ddf2a8c848220b85c467f5f408R379 (and when I tried this locally it did appear to help), but the code has evolved since.

Our use case for this feature is that while in production you'd use TLS and ALPN to serve both HTTP/1.1 and h2 requests on the same port, during development it can be convenient to run without TLS. `NegotiationType.PLAINTEXT` is available for non-TLS connections, but does not (really) support HTTP/1.1 and h2 coexisting on the same port.

Checked on 1.11.0, 1.12.0 and master.

I expected the grpc-java Netty client to successfully connect to a h2c endpoint when configured with `NegotiationType.PLAINTEXT_UPGRADE`.

raboof:
Upgrading from HTTP/1.1 to h2c (HTTP/2 in cleartext).

This is best avoided if possible (since it not widely implemented and comes at
the cost of a roundtrip), but a nice option to have.

Based on work by @xujianhai in #3529

Related to #3528 and fixes #4518

Not sure how to properly test this...

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
@@ -79,6 +79,7 @@ public class TestServiceClient {
   private String testCase = "empty_unary";
   private boolean useTls = true;
   private boolean useAlts = false;
+  private boolean useH2cUpgrade = false;
   private String customCredentialsType;
   private boolean useTestCa;
   private boolean useOkHttp;
@@ -120,6 +121,8 @@ public class TestServiceClient {
         testCase = value;
       } else if ("use_tls".equals(key)) {
         useTls = Boolean.parseBoolean(value);
+      } else if ("use_upgrade".equals(key)) {
+        useH2cUpgrade = Boolean.parseBoolean(value);
       } else if ("use_alts".equals(key)) {
         useAlts = Boolean.parseBoolean(value);
       } else if ("custom_credentials_type".equals(key)) {
@@ -148,7 +151,7 @@ public class TestServiceClient {
         break;
       }
     }
-    if (useAlts) {
+    if (useAlts || useH2cUpgrade) {
       useTls = false;
     }
     if (usage) {
@@ -166,6 +169,9 @@ public class TestServiceClient {
           + "\n  --use_tls=true|false        Whether to use TLS. Default " + c.useTls
           + "\n  --use_alts=true|false       Whether to use ALTS. Enable ALTS will disable TLS."
           + "\n                              Default " + c.useAlts
+          + "\n  --use_upgrade=true|false    Whether to use the h2c Upgrade mechanism."
+          + "\n                              Enabling h2c Upgrade will disable TLS."
+          + "\n                              Default " + c.useH2cUpgrade
           + "\n  --custom_credentials_type   Custom credentials type to use. Default "
             + c.customCredentialsType
           + "\n  --use_test_ca=true|false    Whether to trust our fake CA. Requires --use_tls=true "
@@ -370,7 +376,8 @@ public class TestServiceClient {
         NettyChannelBuilder nettyBuilder =
             NettyChannelBuilder.forAddress(serverHost, serverPort)
                 .flowControlWindow(65 * 1024)
-                .negotiationType(useTls ? NegotiationType.TLS : NegotiationType.PLAINTEXT)
+                .negotiationType(useTls ? NegotiationType.TLS :
+                  (useH2cUpgrade ? NegotiationType.PLAINTEXT_UPGRADE : NegotiationType.PLAINTEXT))
                 .sslContext(sslContext);
         if (serverHostOverride != null) {
           nettyBuilder.overrideAuthority(serverHostOverride);
--- a/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
@@ -50,6 +50,7 @@ import io.netty.handler.codec.http2.DefaultHttp2FrameReader;
 import io.netty.handler.codec.http2.DefaultHttp2FrameWriter;
 import io.netty.handler.codec.http2.DefaultHttp2LocalFlowController;
 import io.netty.handler.codec.http2.DefaultHttp2RemoteFlowController;
+import io.netty.handler.codec.http2.Http2CodecUtil;
 import io.netty.handler.codec.http2.Http2Connection;
 import io.netty.handler.codec.http2.Http2ConnectionAdapter;
 import io.netty.handler.codec.http2.Http2ConnectionDecoder;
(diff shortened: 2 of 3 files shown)


netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -726,6 +728,7 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
       // which causes the upgrade headers to be added
       DefaultHttpRequest upgradeTrigger =
           new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.GET, "/");
+      upgradeTrigger.headers().add(HttpHeaderNames.HOST, "");
```
carl-mastrangelo: This should be /something/ right?

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -710,8 +711,9 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
 
     private final GrpcHttp2ConnectionHandler grpcHandler;
 
-    BufferingHttp2UpgradeHandler(ChannelHandler handler, GrpcHttp2ConnectionHandler grpcHandler) {
-      super(handler);
+    BufferingHttp2UpgradeHandler(ChannelHandler handler, ChannelHandler upgradeHandler,
+                                 GrpcHttp2ConnectionHandler grpcHandler) {
```
carl-mastrangelo: We typically wrap lines +4 spaces.

netty/src/main/java/io/grpc/netty/NettyClientHandler.java
```diff
@@ -323,10 +323,13 @@ void returnProcessedBytes(Http2Stream stream, int bytes) {
   }
 
   private void onHeadersRead(int streamId, Http2Headers headers, boolean endStream) {
-    NettyClientStream.TransportState stream = clientStream(requireHttp2Stream(streamId));
-    stream.transportHeadersReceived(headers, endStream);
-    if (keepAliveManager != null) {
-      keepAliveManager.onDataReceived();
+    // Stream 1 is reserved for the Upgrade response, so we should ignore its headers here:
+    if (streamId != 1) {
```
carl-mastrangelo: It seems like this should check the negotiation type before ignoring the headers.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
ejona86: Seems like it should be the authority. See how `TlsNegotiator`'s host+port is plumbed, partly in this file.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
ejona86: Note that what is here does adhere to the style guide, but yes, we do tend to wrap with 4 spaces.

netty/src/main/java/io/grpc/netty/NettyClientHandler.java
ejona86: @carl-mastrangelo, how do you suggest checking the negotiation type? At present, Netty reserves streamId 1 only for upgrades.

netty/src/main/java/io/grpc/netty/NettyClientHandler.java
```diff
@@ -323,10 +323,13 @@ void returnProcessedBytes(Http2Stream stream, int bytes) {
...
+      NettyClientStream.TransportState stream = clientStream(requireHttp2Stream(streamId));
+      stream.transportHeadersReceived(headers, endStream);
+      if (keepAliveManager != null) {
```
ejona86: This should be outside of the if. It should always run.

interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
```diff
@@ -162,6 +165,9 @@ void parseArgs(String[] args) {
           + "\n  --use_tls=true|false        Whether to use TLS. Default " + c.useTls
           + "\n  --use_alts=true|false       Whether to use ALTS. Enable ALTS will disable TLS."
           + "\n                              Default " + c.useTls
+          + "\n  --use_upgrade=true|false    Whether to use the h2c Upgrade mechanism. "
+          + "Enabling h2c Upgrade will disable TLS."
```
ejona86: We would want this to be on a new line and indented, to make it look fine even on non-super-wide terminals.

netty/src/main/java/io/grpc/netty/NettyClientHandler.java
ejona86: Let's reference https://github.com/netty/netty/blob/netty-4.1.25.Final/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2CodecUtil.java#L47 here instead of just plain `1`.

ejona86:
> If possible, could you add a test that shows upgrade works?

The description mentions that it's unclear how to test this. I'd agree it is unclear, because today we mainly test it against a server... and our server doesn't support h2c Upgrade. Because the response is in h2... yeah, it's hard to avoid having a "real" server here (although the gRPC part could be hacked out; it could be based on the "http2 example" in Netty...) May be a pain...

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
carl-mastrangelo: authority was my guess too, but I don't know if it maps exactly to HTTP/1.1 Host.  Also, what happens if both are present? o_O

netty/src/main/java/io/grpc/netty/NettyClientHandler.java
carl-mastrangelo: The comment made it seem like it was more general than Netty.  Maybe the comment should be updated?

zpencer:
Hi @raboof friendly ping for an update on this PR.

zhangkun83:
@ejona86 @carl-mastrangelo ping

ejona86:
As I mentioned in my review, this is waiting on a test, or being okay with it breaking again because it doesn't have a test.

raboof:
Thanks for the ping. I would like to add a test and gave it a shot a while ago, but had trouble digesting `ProtocolNegotiatorsTest.httpProxy_completes` - I'm not too familiar with the infrastructure used there. I'm OK with merging as-is (and fully aware that means it could break again at any time :) )

ejona86:
@raboof, thanks for the fix! I might poke later to form up a test for it. I wrote `httpProxy_completes` so it is probably easier for me :)

ejona86:
I created #5102 to track the test creation.

PR #4619, opened 2018-07-06
grpclb: Plumb attributes for OOB and backend channels

ejona86:
These attributes can be used by ALTS-specific code to determine whether
ALTS or TLS should be used.

----

Updating the tests was a "find a failing line, figure out what attribute is appropriate" loop. I tried to make sure I put the correct attributes at each point, such that the tests are verifying the correct behavior instead of "whatever the behavior is now." But it was a pretty error-prone process.


ejona86:
CC @jiangtaoli2016. After this goes in, you should be able to look up the attributes in the protocol negotiator.

PR #4625, opened 2018-07-09
netty: Fix regression in Java 9 ALPN support

### LINKED ISSUE #4620: upgrading vision past 1.33.0 gives IllegalArgumentException: SunJSSE selected, but Jetty NPN/ALPN unavailable
I'm running a discord bot on a raspberry pi and using vision to perform OCR on mobile screenshots. It works fine with 1.33 and lower, but 1.34 and 1.35 throw `IllegalArgumentException: SunJSSE selected, but Jetty NPN/ALPN unavailable`.

```
uname -a
Linux raspberrypi 4.14.50-v7+ #1122 SMP Tue Jun 19 12:26:26 BST 2018 armv7l GNU/Linux
```

I'm running a build of Java 10 from here (since the official openjdk binaries are 64 bit):

https://www.bell-sw.com/java-for-raspberry-pi.html

```
java --version
openjdk 10.0.1-BellSoft 2018-04-17
OpenJDK Runtime Environment (build 10.0.1-BellSoft+0)
OpenJDK Server VM (build 10.0.1-BellSoft+0, mixed mode)
```

Relvant stack trace from 1.35:
```
java.lang.IllegalArgumentException: SunJSSE selected, but Jetty NPN/ALPN unavailable
	at io.grpc.netty.shaded.io.grpc.netty.GrpcSslContexts.configure(GrpcSslContexts.java:223)
	at io.grpc.netty.shaded.io.grpc.netty.GrpcSslContexts.configure(GrpcSslContexts.java:189)
	at io.grpc.netty.shaded.io.grpc.netty.GrpcSslContexts.configure(GrpcSslContexts.java:171)
	at io.grpc.netty.shaded.io.grpc.netty.GrpcSslContexts.forClient(GrpcSslContexts.java:120)
	at io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder$NettyTransportFactory$DefaultNettyTransportCreationParamsFilterFactory.<init>(NettyChannelBuilder.java:558)
	at io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder$NettyTransportFactory$DefaultNettyTransportCreationParamsFilterFactory.<init>(NettyChannelBuilder.java:551)
	at io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder$NettyTransportFactory.<init>(NettyChannelBuilder.java:489)
	at io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder.buildTransportFactory(NettyChannelBuilder.java:337)
	at io.grpc.internal.AbstractManagedChannelImplBuilder.build(AbstractManagedChannelImplBuilder.java:390)
	at com.google.api.gax.grpc.InstantiatingGrpcChannelProvider.createSingleChannel(InstantiatingGrpcChannelProvider.java:206)
	at com.google.api.gax.grpc.InstantiatingGrpcChannelProvider.createChannel(InstantiatingGrpcChannelProvider.java:157)
	at com.google.api.gax.grpc.InstantiatingGrpcChannelProvider.getTransportChannel(InstantiatingGrpcChannelProvider.java:149)
	at com.google.api.gax.rpc.ClientContext.create(ClientContext.java:151)
	at com.google.cloud.vision.v1.stub.GrpcImageAnnotatorStub.create(GrpcImageAnnotatorStub.java:84)
	at com.google.cloud.vision.v1.stub.ImageAnnotatorStubSettings.createStub(ImageAnnotatorStubSettings.java:120)
	at com.google.cloud.vision.v1.ImageAnnotatorClient.<init>(ImageAnnotatorClient.java:136)
	at com.google.cloud.vision.v1.ImageAnnotatorClient.create(ImageAnnotatorClient.java:117)
	at com.google.cloud.vision.v1.ImageAnnotatorClient.create(ImageAnnotatorClient.java:108)

```

Is this a bug, or intentional? Can I enable or disable something via config to update the client?

ejona86:
Enable testing on Java 9+ in TlsTest, to prevent future regressions.

Fixes #4620

### DIFF
--- a/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
+++ b/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
@@ -219,6 +219,8 @@ public class GrpcSslContexts {
         apc = ALPN;
       } else if (JettyTlsUtil.isJettyNpnConfigured()) {
         apc = NPN;
+      } else if (JettyTlsUtil.isJava9AlpnAvailable()) {
+        apc = ALPN;
       } else {
         throw new IllegalArgumentException(
             SUN_PROVIDER_NAME + " selected, but Jetty NPN/ALPN unavailable");
--- a/netty/src/test/java/io/grpc/netty/TlsTest.java
+++ b/netty/src/test/java/io/grpc/netty/TlsTest.java
@@ -65,7 +65,7 @@ import org.junit.runners.Parameterized.Parameters;
 public class TlsTest {
 
   public static enum TlsImpl {
-    TCNATIVE, JETTY, CONSCRYPT;
+    TCNATIVE, JDK, CONSCRYPT;
   }
 
   /**
@@ -74,7 +74,7 @@ public class TlsTest {
   @Parameters(name = "{0}")
   public static Iterable<Object[]> data() {
     return Arrays.asList(new Object[][] {
-      {TlsImpl.TCNATIVE}, {TlsImpl.JETTY}, {TlsImpl.CONSCRYPT},
+      {TlsImpl.TCNATIVE}, {TlsImpl.JDK}, {TlsImpl.CONSCRYPT},
     });
   }
 
@@ -101,7 +101,7 @@ public class TlsTest {
         Assume.assumeTrue(OpenSsl.isAvailable());
         sslProvider = SslProvider.OPENSSL;
         break;
-      case JETTY:
+      case JDK:
         Assume.assumeTrue(Arrays.asList(
             SSLContext.getDefault().getSupportedSSLParameters().getCipherSuites())
             .contains("TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"));
@@ -109,9 +109,16 @@ public class TlsTest {
         jdkProvider = Security.getProvider("SunJSSE");
         Assume.assumeNotNull(jdkProvider);
         try {
-          GrpcSslContexts.configure(SslContextBuilder.forClient(), jdkProvider);
-        } catch (IllegalArgumentException ex) {
-          Assume.assumeNoException("Jetty ALPN does not seem available", ex);
+          // Check for presence of an (ironic) class added in Java 9
+          Class.forName("java.lang.Runtime$Version");
+          // Java 9+
+        } catch (ClassNotFoundException ignored) {
+          // Before Java 9
+          try {
+            GrpcSslContexts.configure(SslContextBuilder.forClient(), jdkProvider);
+          } catch (IllegalArgumentException ex) {
+            Assume.assumeNoException("Not Java 9+ and Jetty ALPN does not seem available", ex);
+          }
         }
         break;
       case CONSCRYPT:


netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
```diff
@@ -219,6 +219,8 @@ public static SslContextBuilder configure(SslContextBuilder builder, Provider jd
         apc = ALPN;
       } else if (JettyTlsUtil.isJettyNpnConfigured()) {
         apc = NPN;
+      } else if (JettyTlsUtil.isJava9AlpnAvailable()) {
```
ericgribkoff: Is it necessary to restrict the use of Java 9's ALPN to when the provider is `SunJSSE`? (I assume that it may always  be SunJSSE now, but couldn't find docs on this)

netty/src/test/java/io/grpc/netty/TlsTest.java
```diff
@@ -101,17 +101,24 @@ public void setUp() throws NoSuchAlgorithmException {
...
             .contains("TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"));
         sslProvider = SslProvider.JDK;
         jdkProvider = Security.getProvider("SunJSSE");
         Assume.assumeNotNull(jdkProvider);
         try {
-          GrpcSslContexts.configure(SslContextBuilder.forClient(), jdkProvider);
-        } catch (IllegalArgumentException ex) {
-          Assume.assumeNoException("Jetty ALPN does not seem available", ex);
+          // Check for presence of an (ironic) class added in Java 9
+          Class.forName("java.lang.Runtime$Version");
```
ericgribkoff: With grouping the Jetty and JDK test cases here, it's not immediately clear that Java 9's presence means we should skip the call to `GrpcSslContexts.configure`. It seems like the configuration code path would actually prefer Jetty (if set up) over Java 9 ALPN.

What breaks if the test just invokes `GrpcSslContexts.configure` here and tries to let that "do the right thing" on Java 9?

netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
ejona86: Yes and no. With Jetty, yes, it was necessary. With Java 9, we can do better, by checking whether the Provider supports ALPN (this is done by creating an SSLEngine and trying to use the ALPN API; if it throws, or something like that, it isn't supported).

I want to make the selection logic better for Java 9, but haven't gotten to it. When we're on Java 9+, we can actually us the same selection logic for Conscrypt as SunJSSE, as both support the new API.

But that's for later. For now this hard-coding to SunJSSE is consistent with the approach in `findJdkProvider()`.

netty/src/test/java/io/grpc/netty/TlsTest.java
ejona86: So here's what's going on:
1) If on Java 9+, we want to run the test
2) When on Java 7/8, we want to run the test if Jetty is available. We are abusing `configure()` to find out if Jetty is available, which is a hack but "good enough." The return value of `configure()` is ignored
3) If the two options above aren't available, `Assume` will skip the test.

The problem is that configure() was broken for Java 9, so the assume skipped the test! My bug fix makes it so configure() no longer throws for Java 9+, so the test would begin running again. But avoiding the configure() for Java 9 guarantees that we'll be able to notice problems with configure() throwing, again.

netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
dapengzhang0: >For now this hard-coding to SunJSSE is consistent with the approach in findJdkProvider().

Is `findJdkProvider()` or equivalently `GrpcSslContexts.configure(clientContextBuilder, SslProvider.JDK)` not tested in `TlsTest.java`?

ejona86: `findJdkProvider()` is not tested in TlsTest. `configure()` is not tested when using Jetty.

`findJdkProvider()` is hard to test, because it is based on the environment.

dapengzhang0: Agreed.

PR #4628, opened 2018-07-09
services: double submit cookie interceptor for channelz

zpencer:
Add this form of XSRF protection to channelz. Non web browser clients
calling channelz must also include the cookie header and metadata
key.

### DIFF
--- a/services/src/main/java/io/grpc/services/RequireDoubleSubmitCookieInterceptor.java
+++ b/services/src/main/java/io/grpc/services/RequireDoubleSubmitCookieInterceptor.java
@@ -80,7 +80,7 @@ final class RequireDoubleSubmitCookieInterceptor implements ServerInterceptor {
     if (cookieHeaders == null) {
       return failCall(call);
     }
-    for (String cookieHeader: cookieHeaders) {
+    for (String cookieHeader : cookieHeaders) {
       try {
         for (HttpCookie cookie : HttpCookie.parse(cookieHeader)) {
           if (cookie.getName().equals(tokenName)) {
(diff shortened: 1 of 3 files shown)


services/src/main/java/io/grpc/services/RequireDoubleSubmitCookieInterceptor.java
```diff
@@ -0,0 +1,106 @@
...
+
+  @Override
+  public <ReqT, RespT> Listener<ReqT> interceptCall(ServerCall<ReqT, RespT> call, Metadata headers,
+      ServerCallHandler<ReqT, RespT> next) {
+    String xsrfCookie = null;
+    Iterable<String> cookieHeaders = headers.getAll(COOKIE);
+    if (cookieHeaders == null) {
+      return failCall(call);
+    }
+    for (String cookieHeader: cookieHeaders) {
```
carl-mastrangelo: nit: space after `:`

services/src/main/java/io/grpc/services/RequireDoubleSubmitCookieInterceptor.java
```diff
@@ -0,0 +1,106 @@
...
+            }
+          }
+        }
+      } catch (IllegalArgumentException e) {
+        log.log(Level.FINE, "Failed to parse cookie header", e);
+        return failCall(call);
+      }
+    }
+    String xsrfHeader = headers.get(headerKey);
+    if (xsrfCookie == null || !Objects.equal(xsrfCookie, xsrfHeader)) {
```
carl-mastrangelo: `MessageDigest.isEqual` for coolness.

services/src/main/java/io/grpc/services/RequireDoubleSubmitCookieInterceptor.java
zpencer: Done

services/src/main/java/io/grpc/services/RequireDoubleSubmitCookieInterceptor.java
zpencer: Getting the header as `bytes[]` is easy but a `bytes[]` out of a cookie string looks like will require reencoding.

PR #4629, opened 2018-07-09
compiler, stub: Rename inputType and outputType in @RpcMethod

jbingham-google:


carl-mastrangelo:
I also started working on some doc changes, consider weaving these in in the class level doc:

```
 * {@link RpcMethod} contains a limited subset of information about the RPC to assist
 * <a href="https://docs.oracle.com/javase/6/docs/api/javax/annotation/processing/Processor.html">
 * Java Annotation Processors.</a>
 *
 * <p>
 *   This annotation is used by the gRPC stub compiler to annotate {@link MethodDescriptor}
 *   getters.  Users should not annotate their own classes with annotation.  Not all stubs may have
 *   this annotation, so consumers should not assume that it is present.
 * </p>
```

jbingham-google:
Added the suggested docs.

stub/src/main/java/io/grpc/stub/annotations/RpcMethod.java
```diff
@@ -23,7 +23,17 @@
 import java.lang.annotation.Target;
 
 /**
- * Annotates a method descriptor method to provide metadata for annotation processing.
+ * {@link RpcMethod} contains a limited subset of information about the RPC to assist
```
carl-mastrangelo: fyi, the import of `import io.grpc.MethodDescriptor.MethodType;` should become `import io.grpc.MethodDescriptor;`

and on line 65: 

`MethodType` -> `MethodDescriptor.MethodType`

jbingham-google: Done

PR #4630, opened 2018-07-10
core: enterIdleMode() exits idle if still in use

ericgribkoff:
This fixes a bug with `ManagedChannel#enterIdle` where new calls could hang (remain buffered indefinitely, until the next RPC on the channel) if `#enterIdle` was invoked concurrently with the calls being started. 

The failure cases revolved around `#enterIdle` shutting down the channel's current resolver and LB policy, meaning that asynchronous name resolution or pick selection could (a) never start if the shutdown signal from `#enterIdle` was received before the actual resolution/selection began or (b) have their results ignored since the lb policy that generated them no longer matched the channel's lb policy.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -395,6 +395,9 @@ final class ManagedChannelImpl extends ManagedChannel implements Instrumented<Ch
               .build());
     }
     channelStateManager.gotoState(IDLE);
+    if (inUseStateAggregator.isInUse()) {
+      exitIdleMode();
+    }
   }
 
   // Must be run from channelExecutor
--- a/core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
+++ b/core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
@@ -1928,6 +1928,50 @@ public class ManagedChannelImplTest {
   }
 
   @Test
+  public void enterIdle_exitsIdleIfDelayedStreamPending() {
+    FakeNameResolverFactory nameResolverFactory =
+        new FakeNameResolverFactory.Builder(expectedUri)
+            .setServers(Collections.singletonList(new EquivalentAddressGroup(socketAddress)))
+            .build();
+    channelBuilder.nameResolverFactory(nameResolverFactory);
+    createChannel();
+
+    // Start a call that will be buffered in delayedTransport
+    ClientCall<String, Integer> call = channel.newCall(method, CallOptions.DEFAULT);
+    call.start(mockCallListener, new Metadata());
+
+    // enterIdle() will shut down the name resolver and lb policy used to get a pick for the delayed
+    // call
+    channel.enterIdle();
+    assertEquals(IDLE, channel.getState(false));
+
+    // enterIdle() will restart the delayed call by exiting idle. This creates a new helper.
+    ArgumentCaptor<Helper> helperCaptor = ArgumentCaptor.forClass(Helper.class);
+    verify(mockLoadBalancerFactory, times(2)).newLoadBalancer(helperCaptor.capture());
+    Helper helper2 = helperCaptor.getValue();
+
+    // Establish a connection
+    Subchannel subchannel = helper2.createSubchannel(addressGroup, Attributes.EMPTY);
+    subchannel.requestConnection();
+    ClientStream mockStream = mock(ClientStream.class);
+    MockClientTransportInfo transportInfo = transports.poll();
+    ConnectionClientTransport mockTransport = transportInfo.transport;
+    ManagedClientTransport.Listener transportListener = transportInfo.listener;
+    when(mockTransport.newStream(same(method), any(Metadata.class), any(CallOptions.class)))
+        .thenReturn(mockStream);
+    transportListener.transportReady();
+    when(mockPicker.pickSubchannel(any(PickSubchannelArgs.class)))
+        .thenReturn(PickResult.withSubchannel(subchannel));
+    helper2.updateBalancingState(READY, mockPicker);
+    assertEquals(READY, channel.getState(false));
+
+    // Verify the original call was drained
+    executor.runDueTasks();
+    verify(mockTransport).newStream(same(method), any(Metadata.class), any(CallOptions.class));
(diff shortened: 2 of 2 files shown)


ejona86:
It isn't clear from the PR what you believe the race to be. But I don't think this fixes the race I see.

The race I see:
1. `transportProvider.get(PSA)` grabs a reference to `subchannelPicker`, which is currently non-`null`
2. `enterIdleMode()` is called on the `channelExecutor`. Note that it calls `shutdownNameResolverAndLoadBalancer()` which sets `subchannelPicker = null`
3. `transportProvider` continues, asking the picker for a subchannel. The picker (for its own reasons, not this race) selects `delayedTransport`. It will then be used by `ClientCallImpl` and the call will be hung

The problem is that there is no time bound between (2) and (3). So while the change in this PR fixes the test, it isn't enough to fix the problem in real life. Note that we have this same transport race elsewhere, but we "solve" it via `SUBCHANNEL_SHUTDOWN_DELAY_SECONDS`. I don't think we have this race elsewhere for _delayedTransport_ itself.

ericgribkoff:
Here is the unabridged version of the commit message:

Modify the `greet` method in `HelloWorldClient` to:

```
Runnable task = () -> {
 channel.enterIdle();
};

new Thread(task).start();
response = blockingStub.sayHello(request);
```

This will now reliably (within ~10s of runs) result in a hanging RPC. The log outputs are always some variation of the following:

```
Jul 09, 2018 4:13:22 PM io.grpc.internal.ManagedChannelImpl <init>
FINE: [io.grpc.internal.ManagedChannelImpl-405] Created with target localhost:50051
Jul 09, 2018 4:13:22 PM io.grpc.examples.helloworld.HelloWorldClient greet
INFO: Will try to greet world ...
Jul 09, 2018 4:13:22 PM io.grpc.internal.ManagedChannelImpl exitIdleMode
FINE: [io.grpc.internal.ManagedChannelImpl-405] Exiting idle mode
Jul 09, 2018 4:13:22 PM io.grpc.internal.ManagedChannelImpl enterIdleMode
FINE: [io.grpc.internal.ManagedChannelImpl-405] Entering idle mode
Jul 09, 2018 4:13:22 PM io.grpc.internal.DnsNameResolver$1 run
FINE: Not resolving since shutdown (debug message added to https://github.com/grpc/grpc-java/blob/ffcb3b964bd908dda5793daf82cd0e2556b54592/core/src/main/java/io/grpc/internal/DnsNameResolver.java#L169)
```

There are other, more verbose possibilities, but all of the ones I have repeatedly and reliably reproduced are variations of this sequence of events:

1. Create a channel
1. Start an RPC *and* call enterIdle()
1. The new RPC causes the channel to start name resolution, then the enterIdle() call (in another thread) shutdowns the name resolver
1. The name resolver either sees the shutdown before resolution actually occurs (the above log output) or it resolves the address, but when it notifies `ManagedChannelImpl`, the result is ignored since the name resolver listener's helper does not match the channel's lb helper anymore. See https://github.com/grpc/grpc-java/blob/ffcb3b964bd908dda5793daf82cd0e2556b54592/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java#L1308

So the RPC never actually started (it's still in the delayed transport) when enterIdle() moves the channel into, well, idle, and the channel and the delayed transport's buffered RPC just sit there. But that's not at all the intended behavior: depending on who "wins" the race, the new call should either go out on the old connection (if any), or it should trigger a new connection (leaving idle). 

As is, somewhere close to 10% of the time the modified HelloWorldClient above instead *never* sends the RPC and hangs indefinitely.

With the change in this PR, the reproduction case does not hang after many 100,000s of runs. Therefore I believe that this fix, in a "real life" use case, solves *a* commonly occurring race condition.

The separate race you describe does indeed seem like it could be an issue (although I have not encountered it or reproduced it), but it's root doesn't seem to be *caused* by `enterIdle()`. The same sequence could occur if, somehow, the pick takes long enough that the idle timer period has elapsed and the channel has entered idle once the picker finally chooses the delayed transport. I'm sure this is unlikely, and the `enterIdle()` knob certainly makes this race much easier to trigger - but it seems a separate issue from the bug I'm addressing here. 

Note that, in the presumably more likely case where the picker selects a subchannel, the `SUBCHANNEL_SHUTDOWN_DELAY_SECONDS` "solution" also applies and should give the RPC enough time to start on the transport before `enterIdle()` causes the `InternalSubchannel` to shutdown.

ejona86:
Okay, so the race you are seeing is due to I/O in the NameResolver. While your change addresses the race, I can't help but feel like it is a workaround. I'd like to (or try to) form a better mental model of how this should work before we put in the workaround.

I'm also questioning whether this change should happen in enterIdleMode() directly.

ejona86:
I've considered it some more. The general approach here is fine, but it should be done in `enterIdleMode()`.

The invariant that was broken was, "if delayedTransport has streams, we aren't in idle mode." `enterIdleMode()` obviously can break that invariant, so it double-checking whether `delayedTransport` has streams makes sense. We still have the problem for races with new streams. SUBCHANNEL_SHUTDOWN_DELAY_SECONDS "solves" one class of races. The remaining race I see is the one I mentioned earlier when 1) in IDLE mode, 2) a new RPC is issued, 3) enterIdleMode is called. That could be solved by moving the `exitIdleMode()` call from `transportProvider.get()` to `delayedTransportListener.transportInUse(true)`. I think I may do that as a follow-up.

ejona86:
Oh. Interesting. We already have `delayedTransportListener.transportInUse()` triggering `exitIdleMode()`, via `inUseStateAggregator.handleInUse()`. I guess that means we don't need the `exitIdleMode()` in `transportProvider.get()`.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -902,6 +902,11 @@ public void run() {
         }
         cancelIdleTimer(/* permanent= */ false);
         enterIdleMode();
+        // New streams may have come in while shutting down resolver and LB policy. If so, restart
+        // processing by exiting idle mode.
+        if (delayedTransport.hasPendingStreams()) {
```
ejona86: Use the inUseAggregator instead, which will fix all the races we were talking about. I'll have a follow-up to change our idleness semantics to ignore in-progress RPCs (ones on real transports). I'll probably discuss it offline.

PR #4633, opened 2018-07-11
buildscripts: Remove set -x from run_in_docker.sh

ejona86:
Printing the commands is pretty ugly output and not particularly
helpful, unless debugging the logic. If needing to debug, just add the
-x locally.


ejona86:
@zpencer, ping

PR #4639, opened 2018-07-12
compiler,stub: update RpcMethod docs and usage

carl-mastrangelo:
Notes:

* `fullMethodName` is defined for clarity, rather than directly generating the string.  
* `@code` is used to avoid unused imports 

Follow up items:

* we should really test our annotations if we are going to commit to having them
* Update MethodDescriptor to be able to extract method name.


carl-mastrangelo:
One other note: requestType and responseType could be `default Object.class` and not make them required.  I don't know if this is useful.  The javadoc is such that the MD generics could be `?, ?`   and the annotation still be correct.

ejona86:
FYI, this is how JUnit gets "null" for classes: https://github.com/junit-team/junit4/blob/r4.12/src/main/java/org/junit/Test.java#L68-L84

carl-mastrangelo:
@ejona86 for junit, I considered something like that (possibly even making it private!)  but it would violate the constraint that it matches the response type.   Again, not an argument for or against doing that way, but one thing I considered.

PR #4646, opened 2018-07-13
core,grpclb,netty,okhttp,alts,testing: ProxySocketAddress

zpencer:
First commit:
```
Functional changes (see bullets)

- Add ProxySocketAddress (PSA) class
- EAG: add new PSA based constructors, the old constructors implicitly
  create PSAs with no proxy, the old getter returns the PSA's address
  without proxy.
- Remove Attr key PROXY_PARAMS_KEY
- Make netty and okhttp clases understand PSA
```

Second commit: mostly updating usages to use PSA


core/src/main/java/io/grpc/EquivalentAddressGroup.java
```diff
@@ -32,7 +34,7 @@
 @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1770")
 public final class EquivalentAddressGroup {
 
-  private final List<SocketAddress> addrs;
+  private final List<ProxySocketAddress> addrs;
```
ejona86: No, no, no. This should be something off-to-the-side, not explicit at every layer. We only want two pieces of code to be aware of this: the resolver and the transport. Everything else should be ignorant of it.

It's important that is isn't "baked" into the API like this, because as it is, "there can only be one" such "special" SocketAddress. We want to allow other such addresses as there is need, but we only want those pieces of code that need to care to care.

ejona86:
Also, the component prefix is out-of-hand here ("core,grpclb,netty,okhttp,alts,testing"). I suggest just dropping it, but if you're like @carl-mastrangelo and "always want a component" then use "all".

zpencer:
Closed in favor of https://github.com/grpc/grpc-java/pull/4649

PR #4649, opened 2018-07-17
core: turn PairSocketAddress into ProxySocketAddress

zpencer:
This internal class will only ever be used to plumb proxy info.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ProxySocketAddress.java
+++ b/core/src/main/java/io/grpc/internal/ProxySocketAddress.java
@@ -18,26 +18,25 @@ package io.grpc.internal;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
-import io.grpc.Attributes;
 import java.net.SocketAddress;
 
 /**
- * A data structure to associate a {@link SocketAddress} with {@link Attributes}.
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/grpc/internal/ProxySocketAddress.java
```diff
@@ -18,26 +18,25 @@
...
 import com.google.common.base.Preconditions;
-import io.grpc.Attributes;
 import java.net.SocketAddress;
 
 /**
- * A data structure to associate a {@link SocketAddress} with {@link Attributes}.
+ * A data structure to associate a {@link SocketAddress} with {@link ProxyParameters}.
  */
-final class PairSocketAddress extends SocketAddress {
+final class ProxySocketAddress extends SocketAddress {
```
ejona86: We are going to want/need this to be public. That's fine not to include in this PR.

zpencer: Are you thinking public in `io.grpc.internal` or public in `io.grpc`?

ejona86: Public in `io.grpc`. External `NameResolver`s need to be able to configure proxy usage.

PR #2010, opened 2016-07-01
compiler: Add option to disable version output

ejona86:
If the option becomes popular, we can just remove the version.

I didn't update the golden files in compiler/. They would be convenient to exclude the version,
but it may be better to include the version in them. Opinions welcome.

CC @zhangkun83

PR #4650, opened 2018-07-17
Start 1.15.0 development cycle

zpencer:


carl-mastrangelo:
I had made this PR but noticed that none of the generated code was updated.  Did something not run?

zpencer:
So you're saying on your PR, the git porcelain check failed because generated code was not updated?

ejona86:
Do you have skipCodegen=true in your gradle.properties?

carl-mastrangelo:
I don't but I guess it seems dropping the version was intended behavior, added in #2010.  WAI.

ejona86:
Oh, I see what you're saying. Yeah, our generated code doesn't have the version in it any more. Only the golden files still have the version.

PR #4653, opened 2018-07-18
Fix jdk10 support in travis

zpencer:
Fix class major version warning when using jdk10

fixes https://github.com/grpc/grpc-java/issues/4654


ejona86:
There were failures on Travis and Linux artifacts because Maven Central was returning Forbidden. I triggered rebuilds and they are looking better.

Issue #4452, opened 2018-05-09
Update erorr-prone to 2.3.1


jyane:
https://github.com/google/error-prone/releases

jyane:
@ejona86 Are there any concerns?

carl-mastrangelo:
I believe this drops support for DoNotMock, which we use.

jyane:
I didn't noticed that drops support for DoNotMock...
https://github.com/google/error-prone/pull/989
I don't know why it was removed, it seems to be internal problems.

ejona86:
#4659 removed the `@DoNotMock`s, so this is now possible.

ejona86:
ErrorProne was upgraded to 2.3.2 in 183e1f67

vinodhabib:
> 183e1f6

Yes Eric, This issue is related to an error-prone upgrade to 2.3.1 and as per the release documentation (https://github.com/google/error-prone/releases?q=&expanded=true ) it's already upgraded to 2.36.0, so no action required

PR #4655, opened 2018-07-18
context: Avoid deprecated Truth API

ejona86:
This helps an internal cleanup removing the old failWithRawMessage API.

CC @carl-mastrangelo


ejona86:
This is running into problems due to https://github.com/grpc/grpc-java/issues/4452

Truth is bumping the version of Guava and Guava is bumping the version of error_prone_annotations. Excluding dependencies isn't seeming to work for some reason.

ejona86:
Uh, no GitHub, one PR can not "fix" another PR, such that it should be closed...

Issue #4658, opened 2018-07-19
StreamTracer should be mockable


ejona86:
#3305 put DoNotMock on StreamTracer saying that the mocks are not thread-safe. But [mockito claims they are](https://github.com/mockito/mockito/wiki/FAQ#is-mockito-thread-safe), except for stubbing and verify methods. We need to figure out where the discrepancy is and either file a bug against Mockito or let the StreamTracer be mocked.

I discovered this as part of removing DoNotMock for #4452. For my changes, I left things as-is, but I found the documentation I wrote to be questionable.

CC @carl-mastrangelo

vinodhabib:
@shivaspeaks @ejona86 After analysis I understand like this issue does not have any action as i can see @DoNotMock annotation was removed as part of PR [#4659](https://github.com/grpc/grpc-java/pull/4659) from the StreamStracer class which was added/introduced as part of PR [#3305](https://github.com/grpc/grpc-java/pull/3305) earlier. 
Also @DoNotMock was removed from error_prone_annotations in 2.1.3 version upgrade as per PR [#4659](https://github.com/grpc/grpc-java/pull/4659) details later it was added as gradle dependency as part of grpc-core.

For the below highlighted description point :

**We need to figure out where the discrepancy is and either file a bug against Mockito or let the StreamTracer be mocked**  as per the below comment https://github.com/grpc/grpc-java/pull/4659/files#r203902174 from Eric can see as 
See https://github.com/grpc/grpc-java/issues/4658. I'm purposefully not trying to "change" anything; I'm trying to keep the status-quo. Yes, I see the tester is in internal. There's relatively few users of this, so I'm not too worried about them using something internal (especially inside google where it is easy to track who's using it). But this is also the state the class was already in.

![image](https://github.com/user-attachments/assets/ab0dc049-dd80-4222-b10e-b7e834603395)


By referring to the above comments and other PR details, I hope we don't have anything pending changes here. Kindly Suggest.

ejona86:
"We need to figure out where the discrepancy is and either file a bug against Mockito or let the StreamTracer be mocked" seems unresolved. I don't see how that comment says anything; that is me just saying, "I'm keeping the status quo" and was written the day after I made this issue.

I'm willing to resolve the last part of this by saying #3305 is mistaken and we remove the "DO NOT MOCK" from StreamTracer. At this point, the problem could have been long-resolved bugs in mockito.

vinodhabib:
[...]

PR #4659, opened 2018-07-19
Remove DoNotMock annotation in favor of JavaDoc

ejona86:
DoNotMock was removed from error_prone_annotations in 2.1.3, because
there was no enforcement mechanism (which is in google/error-prone#572).
Guava and Trust also depend on error_prone_annotations and are beginning
to use newer versions, so our usage of DoNotMock is causing diamond
dependency problems. This allows us to update to 2.2.0.

The annotations were useful internally; we're solving that in cl/205294089.

------

This came out of trying to fix #4655.

### DIFF
--- a/core/src/main/java/io/grpc/StreamTracer.java
+++ b/core/src/main/java/io/grpc/StreamTracer.java
@@ -16,15 +16,15 @@
 
 package io.grpc;
 
-import com.google.errorprone.annotations.DoNotMock;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
  * Listens to events on a stream to collect metrics.
+ *
+ * <p>DO NOT MOCK: Use TestStreamTracer. Mocks are not thread-safe
  */
 @ExperimentalApi("https://github.com/grpc/grpc-java/issues/2861")
 @ThreadSafe
-@DoNotMock
 public abstract class StreamTracer {
   /**
    * Stream is closed.  This will be called exactly once.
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/grpc/StreamTracer.java
```diff
@@ -16,15 +16,15 @@
 
 package io.grpc;
 
-import com.google.errorprone.annotations.DoNotMock;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
  * Listens to events on a stream to collect metrics.
+ *
+ * <p>DO NOT MOCK: Use TestStreamTracer. Mocks are not thread-safe
```
dapengzhang0: `TestStreamTracer` is an internal interface. Is thread-safety the only reason?  Can I mock it when testing with direct executor?

ejona86: See https://github.com/grpc/grpc-java/issues/4658. I'm purposefully not trying to "change" anything; I'm trying to keep the status-quo. Yes, I see the tester is in internal. There's relatively few users of this, so I'm not too worried about them using something internal (especially inside google where it is easy to track who's using it). But this is also the state the class was already in.

dapengzhang0: If "not to change anything" , "Mocks are not thread-safe" should not be added.
Otherwise LGTM

ejona86: See #3305. The reason for the removal was the thought that mocks aren't thread-safe. Otherwise there is no reasoning why to avoid a mock with this class.

PR #4660, opened 2018-07-20
interop-testing: Fix clientStreaming interop test does not fail on missing trailers

dapengzhang0:
If the server sends all response message but does not send the trailers, client throws
StatusRuntimeException: INTERNAL: Received unexpected EOS on DATA frame from server

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -559,8 +559,19 @@ public abstract class AbstractInteropTest {
       requestObserver.onNext(request);
     }
     requestObserver.onCompleted();
-    assertEquals(goldenResponse, responseObserver.firstValue().get());
     responseObserver.awaitCompletion();
+    assertEquals(goldenResponse, responseObserver.firstValue().get());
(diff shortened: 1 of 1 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -559,8 +559,19 @@ public void clientStreaming() throws Exception {
       requestObserver.onNext(request);
     }
     requestObserver.onCompleted();
-    assertEquals(goldenResponse, responseObserver.firstValue().get());
     responseObserver.awaitCompletion();
+    assertEquals(goldenResponse, responseObserver.firstValue().get());
+    assertThat(responseObserver.getValues()).hasSize(1);
+    final Throwable t = responseObserver.getError();
+    if (t != null) {
+      // AssertionError(String, Throwable) only present in Android API 19+ & Java 7+
```
ejona86: Can we just use `new RuntimeException(t)` instead? Or maybe even `assertNull(t)`.

dapengzhang0: Now I use `new AssertionError(t)` instead. 
`assertNull(t)` seems not to show details of `t` clear enough.

PR #4664, opened 2018-07-21
Downgrade to Truth 0.41 for Java 7 support

ejona86:
gae-jdk7 in particular is failing with the 0.42 Truth because it depends
on org.checkerframework:checker-qual 2.5.3 instead of 2.0.0.

------

This is a workaround. We're going to need to figure out a long-term strategy. We have to be able to update to newer versions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -229,7 +229,7 @@ subprojects {
             // Test dependencies.
             junit: 'junit:junit:4.12',
             mockito: 'org.mockito:mockito-core:1.9.5',
-            truth: 'com.google.truth:truth:0.42',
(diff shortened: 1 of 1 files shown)


ejona86:
This is fixing this error:
```
Unable to stage app: Class file is Java 8 but max supported is Java 7: org/checkerframework/dataflow/qual/Pure$Kind.class in /tmpfs/src/github/grpc-java/gae-interop-testing/gae-jdk7/build/exploded-grpc-gae-interop-testing-jdk7/WEB-INF/lib/checker-qual-2.5.3.jar
```

PR #4666, opened 2018-07-23
core: channel tracing log only when number of backends changed between zeor and nonzero

dapengzhang0:
This is to implement one of changes made in the spec
https://github.com/grpc/proposal/pull/89/files

### DIFF
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -211,7 +211,7 @@ final class ManagedChannelImpl extends ManagedChannel implements Instrumented<Ch
   @CheckForNull
   private final ChannelTracer channelTracer;
   private final Channelz channelz;
-  private Boolean nonzeroBackends; // a flag for doing channel tracing when flipped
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -211,6 +211,7 @@ void handleUncaughtThrowable(Throwable t) {
   @CheckForNull
   private final ChannelTracer channelTracer;
   private final Channelz channelz;
+  private Boolean nonzeroBackends; // a flag for doing channel tracing when flipped
```
zpencer: I think this part would be more clear if we created a separate `enum` for resolved and error, since the goal is to avoid consecutive reports for the same event type.

dapengzhang0: Per the new spec, error events in general are not logged unless it results in # of backends down to zero from nonzeor. (Both "zero backends resolved" and "general error" are considered # of backends <- zero). A separate `enum` for error seems overly specific. Maybe the flag can be refactored as "error"?

PR #4667, opened 2018-07-23
android: include proguard config to keep method names

ericgribkoff:
Redo of https://github.com/grpc/grpc-java/pull/4662. 

Also deletes some unnecessary forwarding methods from AndroidChannelBuilder

### DIFF
--- a/android/proguard-rules.txt
+++ b/android/proguard-rules.txt
@@ -1,5 +1,6 @@
 -keepclassmembers class io.grpc.okhttp.OkHttpChannelBuilder {
   io.grpc.okhttp.OkHttpChannelBuilder forTarget(java.lang.String);
+  io.grpc.okhttp.OkHttpChannelBuilder scheduledExecutorService(java.util.concurrent.ScheduledExecutorService);
   io.grpc.okhttp.OkHttpChannelBuilder sslSocketFactory(javax.net.ssl.SSLSocketFactory);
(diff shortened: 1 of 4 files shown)


android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
```diff
@@ -138,20 +126,6 @@ public AndroidChannelBuilder sslSocketFactory(SSLSocketFactory factory) {
     }
   }
 
-  /** Set the delegate channel builder's scheduledExecutorService. */
```
ejona86: I expect more people need this. The only reason they may not now is because it was unavailable for so long.

android/proguard-rules.txt
```diff
@@ -0,0 +1,5 @@
+-keepclassmembers class io.grpc.okhttp.OkHttpChannelBuilder {
```
ejona86: add scheduledExecutorService.

ejona86:
Do we want to back-port this?

android/proguard-rules.txt
ericgribkoff: Done, thanks

ericgribkoff:
Yes, I think this is worth backporting to 1.14.x. Users can directly configure these proguard rules in their app (so this could be viewed as addressing a documentation short-coming rather than fixing a bug), but this PR will make it much easier for users.

PR #4675, opened 2018-07-24
core: let channel tracing log service config changes

dapengzhang0:
This is the implementation of the second one of the changes to the channel tracing spec
https://github.com/grpc/proposal/pull/89

### DIFF
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -211,7 +211,10 @@ final class ManagedChannelImpl extends ManagedChannel implements Instrumented<Ch
   @CheckForNull
   private final ChannelTracer channelTracer;
   private final Channelz channelz;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -211,7 +211,10 @@ void handleUncaughtThrowable(Throwable t) {
   @CheckForNull
   private final ChannelTracer channelTracer;
   private final Channelz channelz;
+  @CheckForNull
   private Boolean zeroBackends; // a flag for doing channel tracing when flipped
```
zpencer: nit: sorry to bring this up again, but I just realized it might be even more clear if we made this `haveBackends` since the `zero` part is still effectively a negation.

dapengzhang0: Done. Thanks @zpencer

PR #4679, opened 2018-07-26
Upgrade Gradle to 4.9

ejona86:
This opens up the ability of dependency locking and the now-stable Maven
Publish Plugin. Also failOnVersionConflict no longer needs to be
commented out for the dependency insight report.


ejona86:
Created #4680 and #4681 for swapping to the new features.

ejona86:
I added a second commit to upgrade some plugins. I will leave it as a separate commit when merging, since if anything goes wrong with it it could be reverted separately. Initially I thought it applied to gradle 4.7 as well, but I could only reproduce the warnings with 4.9.

CC @zhangkun83, I guess the protoc plugin doesn't cause any warnings with our project.

PR #4682, opened 2018-07-26
interop-testing: Add special_status_message test

ejona86:


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1532,6 +1532,27 @@ public void statusCodeAndMessage() throws Exception {
     assertStatsTrace("grpc.testing.TestService/FullDuplexCall", Status.Code.UNKNOWN);
   }
 
+  @Test
+  public void specialStatusMessage() throws Exception {
+    int errorCode = 2;
+    String errorMessage = "\t\ntest with whitespace\r\nand Unicode BMP ☺ and non-BMP 😈\t\n";
+    SimpleRequest simpleRequest = SimpleRequest.newBuilder()
```
dapengzhang0: can you use something like `\u00ab` instead?

ejona86: To what end?

The first one could be \u263A. But the second one you have to "know" that U+1F608 will be split into surrogate pairs and use \uD83D\uDE08. I can do that, but it is a bit harder to verify the test is correct. (For example, if you get the UTF-16 surrogate pairs wrong, then it becomes invalid.)

dapengzhang0: 😈

Issue #4683, opened 2018-07-26
Tracking issue for StatusRuntimeException


carl-mastrangelo:

carl-mastrangelo:
Notes from discussion:

* Using Status.asException does the same thing, but not objectively better
* Several users of both constructor and static factory
* Considered making the ctor protected, so that the default construction is sans- trailers, but extenders can still call it.

Decision: Seems okay as is.

ejona86:
API review:
 * We should stabilize Status.asException(Metadata). Yes, it is two ways of doing the same thing, but has been around a long time and doesn't hurt much.

temawi:
Fixed in #10322.

PR #4686, opened 2018-07-27
core: stabilize Status Exceptions API, and warn on null trailers in S…

carl-mastrangelo:
…tatus


carl-mastrangelo:
Updates #4683

PR #4697, opened 2018-07-28
all:  add tracking issues for all experimental APIs and make it required

carl-mastrangelo:
Additionally, make Status*Exception.getTrailers() non experimental


core/src/main/java/io/grpc/StatusException.java
```diff
@@ -52,7 +52,6 @@ public final Status getStatus() {
   /**
    * Returns the received trailers.
    */
```
dapengzhang0: `@since ?`

carl-mastrangelo: Done.

PR #4702, opened 2018-07-30
bazel: Remove http_file trick for neverlink

ejona86:
The trick wasn't necessary, as any java_library can mark itself as
neverlink which applies to its dependencies.


dapengzhang0:
cc @davido

PR #4704, opened 2018-07-30
services: Fix binarylog's java_package to include version

ejona86:
Since this is a breaking change, it will need to be coordinated with the proto sync.

grpc/grpc-proto#31


ejona86:
@zpencer, ping

PR #4708, opened 2018-07-30
core: Avoid implicit requestConnection in PickFirst

ejona86:
This makes the behavior more clear.

----

Should I also remove the body of Picker.requestConnection?


zhangkun83:
Yes you can remove the body of Picker.requestConnection?

ejona86:
@zhangkun83, PTAL. Note that there was a bug because of how long ago these changes originated.

ejona86:
Actually. Hold off a sec. In looking to add a test I noticed a test is failing.

ejona86:
PTAL

PR #4709, opened 2018-07-30
travis.yml: Actually trigger coveralls

ejona86:
Previously it was just erroring:
[[: command not found


ejona86:
I could, but we probably don't care? It broke in #3638, so everything back to 1.12 would need the change.

PR #4711, opened 2018-07-30
core: channel tracing to log lb policy changes

dapengzhang0:


core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -43,11 +47,19 @@
   static final String GRPCLB_LOAD_BALANCER_FACTORY_NAME =
       "io.grpc.grpclb.GrpclbLoadBalancerFactory";
 
-  AutoConfiguredLoadBalancerFactory() {}
+  @Nullable
+  private final ChannelTracer channelTracer;
+  @Nullable
+  private final TimeProvider timeProvider;
+
+  AutoConfiguredLoadBalancerFactory(ChannelTracer channelTracer, TimeProvider timeProvider) {
```
zpencer: nit: use `@Nullable` here as well to be consistent

core/src/main/java/io/grpc/PickFirstBalancerFactory.java
```diff
@@ -120,6 +120,11 @@ public void shutdown() {
         subchannel.shutdown();
       }
     }
+
+    @Override
+    public String toString() {
+      return "PickFirstBalancer";
```
zpencer: If we don't define this explicitly, then we can differentiate between different instances. Would that be something we want?

If we really only want to show only the class name, can we get it via something like `getClass().getSimpleName()`?

dapengzhang0: Done.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
dapengzhang0: Done.

dapengzhang0:
PTALA

dapengzhang0:
@zpencer Friendly ping

PR #4723, opened 2018-08-01
android: add AndroidChannelBuilder#fromBuilder

ericgribkoff:


android/src/main/java/io/grpc/android/AndroidChannelBuilder.java
```diff
@@ -96,13 +99,23 @@ private AndroidChannelBuilder(String target) {
     }
   }
 
+  private AndroidChannelBuilder(ManagedChannelBuilder delegateBuilder) {
+    this.delegateBuilder = delegate;
```
carl-mastrangelo: checkNotNull ?

ericgribkoff: done, thanks

PR #4728, opened 2018-08-02
inprocess: prevent null names, and define socket address equality

carl-mastrangelo:
This is a minor API change, since names can no longer be null.  gRPC never constructed these addresses with null, but other users could.


carl-mastrangelo:
Done

PR #4735, opened 2018-08-03
testing-proto: Add test for @RpcMethod

jbingham-google:


testing-proto/src/test/java/io/grpc/testing/protobuf/SimpleServiceTest.java
```diff
@@ -57,4 +72,90 @@ public void serviceMethodDescriotrs() {
...
+
+  @Test
+  public void testRpcMethodAnnotations() throws Exception {
+    JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+    StandardJavaFileManager fileManager = compiler.getStandardFileManager(null, null, null);
+    AnnotationProcessor processor = new AnnotationProcessor();
+
+    Iterable<? extends JavaFileObject> obs =
+        fileManager.getJavaFileObjects(
+            "./src/generated/main/grpc/io/grpc/testing/protobuf/SimpleServiceGrpc.java");
```
ejona86: This will probably be annoying to maintain in different compilation environments, but I don't see much alternative.

ejona86:
Looks like this is failing on JDK 9+. `assertTrue(task.call());` is failing.

testing-proto/src/test/java/io/grpc/testing/protobuf/SimpleServiceTest.java
```diff
@@ -57,4 +72,90 @@ public void serviceMethodDescriotrs() {
...
+    }
+  }
+
+  @Test
+  public void testRpcMethodAnnotations() throws Exception {
+    JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+    StandardJavaFileManager fileManager = compiler.getStandardFileManager(null, null, null);
+    AnnotationProcessor processor = new AnnotationProcessor();
+
+    Iterable<? extends JavaFileObject> obs =
```
carl-mastrangelo: ideally this would be stored as a resource for the test.  I think our windows build may not like this test.

As a work around, maybe you can add an `Assume` call at the beginning to assert the file exists?

jbingham-google: Added an Assume.

jbingham-google:
JDK9+ needed the javax annotations as a runtime dependency.

Issue #1621, opened 2016-04-01
Add an ability to run protobuf server as a Servlet 4


trknz:
Pls add an ability to run a protobuf v3 server as a Servlet 4 on any server with the Servlet 4 support.

ejona86:
Dump from what I wrote on SO to have the information in one place:
> It does seem feasible to use the async servlet APIs to implement a gRPC server (as an alternative to the Netty server), except possibly for trailers. It's not 100% clear how to send trailers when the server is HTTP/2, since a common technique with HTTP/1 was for the servlet to manually perform chunked encoding which does not exist in HTTP/2.

cs224:
Could someone please give a status update on this topic? Thanks!

ejona86:
No progress. I still don't know how to send trailers in this environment.

jnehlmeier:
JavaDoc for Java EE 8 / Servlet 4.0 describes an API to set trailers on a response.

https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpServletResponse.html#setTrailerFields-java.util.function.Supplier-

ejona86:
@jnehlmeier, thank you!

I'll note one part of the docs that could be an annoyance, but I think they are talking about the remote, not the servlet (in which case we're fine to ignore it):
> The RFC requires the name of every key that is to be in the supplied Map is included in the comma separated list that is the value of the "Trailer" response header. The application is responsible for ensuring this requirement is met. Failure to do so may lead to interoperability failures.

If it is for servlet interoperability we can still live with it, but it requires the application to help us out. Go already has to deal with something similar when using the Go-provided http/2 server.

hofmanndavid:
Okay, Will try to get back when I have something. give me 2 weeks.

ejona86:
CC @dapengzhang0

dapengzhang0:
The "Trailer" in the headers is required by [RFC7230 (HTTP/1.1)](https://tools.ietf.org/html/rfc7230#section-4.4) . So hopefully the "interoperability failure" will only occur for HTTP/1.1 clients, and HTTP/2 clients can still interoperate.

> > The RFC requires the name of every key that is to be in the supplied Map is included in the comma separated list that is the value of the "Trailer" response header. The application is responsible for ensuring this requirement is met. Failure to do so may lead to interoperability failures.
> 
[...]

PR #4738, opened 2018-08-04
servlet: Implement gRPC server as a Servlet

dapengzhang0:
The goal is as described in #1621 to add the ability to run a gRPC server as a Servlet on any web container with the Servlet 4.0 support and HTTP/2 enabled. 

cc @vlsinitsyn @cs224 @jnehlmeier @hofmanndavid

cc @meltsufin for high level design

# API

This PR provides the following API:

An adapter
```java
/**
 * An adapter that transforms {@link HttpServletRequest} into gRPC request and lets a gRPC server
 * process it, and transforms the gRPC response into {@link HttpServletResponse}. An adapter can be
 * instantiated by {@link ServletServerBuilder#buildServletAdapter()}.
 *
 * <p>In a servlet, calling {@link #doPost(HttpServletRequest, HttpServletResponse)} inside {@link
 * javax.servlet.http.HttpServlet#doPost(HttpServletRequest, HttpServletResponse)} makes the servlet
 * backed by the gRPC server associated with the adapter. The servlet must support Asynchronous
 * Processing and must be deployed to a container that supports servlet 4.0 and enables HTTP/2.
 */
public final class ServletAdapter {
   /**
   * Call this method inside {@link javax.servlet.http.HttpServlet#doGet(HttpServletRequest,
   * HttpServletResponse)} to serve gRPC POST request.
   */
  public doPost(HttpServletRequest, HttpServletResponse);
```

and a grpc server builder
```java
/** Builder to build a gRPC server that can run as a servlet. */
public final class io.grpc.servlet.ServletServerBuilder extends ServerBuilder

  /**
   * <p>Users should not call this method directly. Instead users should call
   * {@link #buildServletAdapter()} which internally will call {@code build()} and {@code start()}
   * appropriately.
   * @throws IllegalStateException if this method is called by users directly
   */
  @Override
  public Server build();

  /** Creates a {@link ServletAdapter}.*/
  public ServletAdapter buildServletAdapter();

```

and a servlet impl
```
public class GrpcServlet extends HttpServlet {
  public GrpcServlet(List<BindableService> grpcServices);
}
```

---

A `ServletAdapter` instance will be backing one or more gRPC services with a `ServletServerBuilder`
```java
ServletAdapter servletAdapter = 
    new ServletServerBuilder().addService(new HelloServiceImpl()).buildServletAdapter();
```

---

A servlet powering the gRPC services will be as simple as either
```java
@WebServlet(asyncSupported = true)
public class HelloServlet extends HttpServlet {

  @Override
  protected void doPost(HttpServletRequest req, HttpServletResponse resp) {
    servletAdapter.doPost(req, resp);  
  }
}
```
or 
```java
@WebServlet(asyncSupported = true)
public class HelloServlet extends GrpcServlet {
}
```

## Alternative API

# Hello World example
See examples/example-servlet/src/main/java/io/grpc/servlet/examples/helloworld/

# Implementation
Aside from trivial API pluming, what the impl actually does is (in `doPost()` method)

- start an `AsyncContext`
- set `WriteListener` and `ReadListener` for the `ServletOutputStream` and `ServletInputStream`
- create and pass a new instance of `io.grpc.servlet.ServletServerStream` to  `ServerTransportListener.streamCreated(serverStream, method, headers)`
- This `ServletServerStream` calls `stream.transportState().inboundDataReceived()` in the `ReadListener.onDataAvailable()` callback
- The `ServletServerStream` holds a WritableBufferAllocator for the outbound data that will write to the  `ServletOutputStream`, and uses an atomic ref of `ServletServerStream.WriteState` to coordinate with the `WriteListener.onWritePossible()` callback.
- The `ServletServerStream.TransportState.runOnTransportThread()` method uses a `SerializingExecutor(directExecutor())`

# Test result
We tested it with a servlet backed by the [InteropTest](/grpc/grpc/blob/master/doc/interop-test-descriptions.md) gRPC service [`TestServiceImp`](blob/master/interop-testing/src/main/java/io/grpc/testing/integration/TestServiceImpl.java), and an ordinary gRPC InteropTest client, for the test cases `EMPTY_UNARY`, `LARGE_UNARY`, `CLIENT_STREAMING`, `SERVER_STREAMING`, and `PING_PONG`. The following are the test results of some of the web container vendors

### Jetty 10.0-SNAPSHOT
All tests passed! 
Jetty looks fully compliant with the Servlet 4.0 spec and HTTP/2 protocol. :100: 
@sbordet

---

### GlassFish 5.0 - Web Profile
All tests passed with minor workaround.
An issue of GlassFish is filed, cc @mattgill98 for help:
-  [Async servlet failed to send large data frames completely when HTTP/2 trailers are present](https://github.com/eclipse-ee4j/grizzly/issues/2016)

---

### Tomcat 9.0.10
Non-blocking servlet I/O (ReadListener/WriteListener) practically can not work in full-duplex case. Only `EMPTY_UNARY` test passed. 
(Update:
Filed multiple bugs to Tomcat: cc @markt-asf for help)
- [Async servlet over HTTP/2 setReadListener does not work if post request data arrives much later than headers](https://bz.apache.org/bugzilla/show_bug.cgi?id=62605)  - fixed 9.0.11 onwards
- [Async servlet over HTTP/2 WriteListener does not work because onWritePossible is never called back again](https://bz.apache.org/bugzilla/show_bug.cgi?id=62614)  - fixed 9.0.11 onwards
- [Async servlet over HTTP/2 large write data corrupted](https://bz.apache.org/bugzilla/show_bug.cgi?id=62620)  - fixed 9.0.12 onwards
- [Async servlet over HTTP/2 non-blocking write does not work if client sets custom SETTINGS_INITIAL_WINDOW_SIZE](https://bz.apache.org/bugzilla/show_bug.cgi?id=62628) - fixed 9.0.12 onwards
- [Async servlet over HTTP/2 response.flushBuffer() intermittently fails](https://bz.apache.org/bugzilla/show_bug.cgi?id=62635) - Not a bug, the async flush feature is not specified by the Servlet 4.0 spec; but a patch to support it is still added by Tomcat. Really appreciate it.
- [Async servlet over HTTP/2 on embedded server sporadically timeout forever when reading near end of request](https://bz.apache.org/bugzilla/show_bug.cgi?id=62899) - fixed 9.0.14 onwards

### (Update: Tomcat 9.0.x trunk
All tests passed!  :100:)

---

### Undertow 2.0.11.Final/WildFly-14.0.0.Beta1
Not able to test. Simply can not make the servlet container send out trailers for plain servlet. cc @ctomc @stuartwdouglas for help
(Update: Identified as undertow bug
- [HTTP/2 does not honour the trailer supplier attachment](https://issues.jboss.org/projects/UNDERTOW/issues/UNDERTOW-1389) - fixed

)

### (Update: Undertow 2.0.12.Final
All tests passed!  :100:)


stuartwdouglas:
Trailers should be supported. Is there an easy way for me to test this so I can see what is wrong?

dapengzhang0:
@stuartwdouglas The [StackOverflow](https://stackoverflow.com/questions/51524769/undertow-never-sends-http-2-trailers-for-plain-servlet) question has provided a simple servlet, I couldn't even make that work. I might have missed some configuration. If you can successfully run that servlet and _verify that client side receives the trailers_, just let me know the steps I might have missed. Really appreciate it. On my machine, I can verify Glassfish, Tomcat and Jetty all send trailers, but Undertow/Wildfly does not.

stuartwdouglas:
This should be fixed in Undertow 2.0.12.Final, which will be released shortly.

dapengzhang0:
Thanks a lot @stuartwdouglas , I will test on it once it's released. Right now it is not straight forward to run the InteropTest for this PR, I'm trying to make the steps easier.

sbordet:
@dapengzhang0 can you try Jetty 9.4.11? It should have similar support for 4.0, trailers, etc. that Jetty 10 has.

sbordet:
@dapengzhang0 thanks for trying. Yes, with 9.4.x you need to downcast to Jetty specific classes to get the trailer methods. Never mind 9.4.x, and happy that Jetty 10.0.x works well for this case.

servlet/src/main/java/io/grpc/servlet/ServerBuilder.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ScheduledExecutorService;
+import javax.annotation.concurrent.NotThreadSafe;
+
+/**
+ * Builder to build a gRPC server that can run as a servlet.
+ */
+@NotThreadSafe
+public final class ServerBuilder extends AbstractServerImplBuilder<ServerBuilder> {
```
ejona86: ServletServerBuilder? Let's not have it the same name as its ancestor.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -114,14 +114,16 @@
...
+
   /**
    * Construct a server.
    *
    * @param builder builder with configuration for server
    * @param transportServer transport server that will create new incoming transports
    * @param rootContext context that callbacks for new RPCs should be derived from
    */
-  ServerImpl(
+  public ServerImpl(
```
ejona86: Why this change?

servlet/src/main/java/io/grpc/servlet/ServerBuilder.java
```diff
@@ -0,0 +1,198 @@
...
+    if (!usingCustomScheduler) {
+      scheduledExecutorService = SharedResourceHolder.get(GrpcUtil.TIMER_SERVICE);
+    }
+
+    // Create only one "transport" for all requests because it has no knowledge of which request is
+    // associated with which client socket. This "transport" does not do socket connection, the
+    // container does.
+    ServerTransportImpl serverTransport =
+        new ServerTransportImpl(scheduledExecutorService, usingCustomScheduler);
+    return server.getServerListener().transportCreated(serverTransport);
```
ejona86: I don't think we should have a `getServerListener()`. Is the only problem that `InternalServerImpl` can't call the listener during start?

servlet/src/main/java/io/grpc/servlet/ServerStream.java
```diff
@@ -0,0 +1,398 @@
...
+
+    @Override
+    public void release() {}
+  }
+
+  /**
+   * A queue of WritableBuffers. Not safe for multiple concurrent polls, or multiple concurrent
+   * enqueues, but safe for concurrently calling one poll() and one enqueue().
+   */
+  static final class WritableBufferChain {
```
ejona86: Why not use LinkedBlockingQueue?

settings.gradle
```diff
@@ -19,6 +19,8 @@ include ":grpc-all"
 include ":grpc-alts"
 include ":grpc-benchmarks"
 include ":grpc-services"
+include ":grpc-servlet"
+include ":grpc-servlet-testing"
```
ejona86: What is this? I don't see it.

servlet/src/main/java/io/grpc/servlet/ServletAdapter.java
```diff
@@ -0,0 +1,95 @@
...
+ * process it, and transforms the gRPC response into {@link HttpServletResponse}. An adapter can be
+ * instantiated by {@link Factory#create}. The gRPC server is built from the ServerBuilder provided
+ * in {@link Factory#create}.
+ *
+ * <p>In a servlet, calling {@link #doPost(HttpServletRequest, HttpServletResponse)} inside {@link
+ * javax.servlet.http.HttpServlet#doPost(HttpServletRequest, HttpServletResponse)} makes the servlet
+ * backed by the gRPC server associated with the adapter. The servlet must support Asynchronous
+ * Processing and must be deployed to a container that supports servlet 4.0 and enables HTTP/2.
+ */
+public interface ServletAdapter {
```
ejona86: What's the point of this interface? Why don't we rename ServletAdaptorImpl to ServletAdaptor and use it instead?

servlet/src/main/java/io/grpc/servlet/ServletAdapter.java
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * @return true if the request comes from a gRPC client
+   */
+  static boolean isGrpc(HttpServletRequest request) {
+    return request.getContentType() != null
+        && request.getContentType().contains(GrpcUtil.CONTENT_TYPE_GRPC);
+  }
+
+  /** Factory of ServletAdapter. */
+  final class Factory {
```
ejona86: What's the point of this factory? Why don't we just add a new public method to the Builder?

core/src/main/java/io/grpc/internal/ServerImpl.java
dapengzhang0: Extend package level access to public access so that `io.grpc.servlet.(Servlet)ServerBuilder` can reuse the code. Any better idea to get around it? The only need is to get the `ServerListener` that the `serverImp` instance starts with.

servlet/src/main/java/io/grpc/servlet/ServerBuilder.java
dapengzhang0: The only problem is to get access the `ServerListener` that the `ServerImpl` instance starts with, so that I can  reuse the `ServerImpl`'s code.

settings.gradle
dapengzhang0: see line 46
```
+ project(':grpc-servlet-testing').projectDir = "$rootDir/servlet/interop-testing" as File
```

and `servlet/interop-testing/build.gradle`

core/src/main/java/io/grpc/internal/ServerImpl.java
ejona86: I don't see any usage of this in `ServerBuilder`. Can you point me to a line?

servlet/src/main/java/io/grpc/servlet/ServerBuilder.java
ejona86: So let's just have `InternalServerImpl` create the `ServerTransportImpl`.

settings.gradle
ejona86: Ah, I see. Let's name that `grpc-servlet-interop-testing`.

core/src/main/java/io/grpc/internal/ServerImpl.java
dapengzhang0: It was used in `ServerBuilder`, as show in the diff of the commit `get rid of public ServerImpl`, but never mind, I managed to git rid of it in the new commit.

servlet/src/main/java/io/grpc/servlet/ServerBuilder.java
dapengzhang0: I populated the `serverListener` in `InternalServerImpl` back to the builder and got rid of it.

dapengzhang0:
> There's a lot here, and much of it has obvious things that need to be cleaned up before a real review. What did you want me to look at for the initial review?

@ejona86 aside from impl details, for the initial review, is there anything fundamentally in a wrong approach, or anything important is missing (such as `runOnTransportThread()` I had missed), or anything I must fix before adding new stuff?

servlet/src/main/java/io/grpc/servlet/ServerStream.java
dapengzhang0: Because I just need a simple linked list, only requirement is next pointer be up-to-date, no compare-and-swap, no loop.

ejona86: Those are all things you don't need. What do you need? You wanted it to be faster?

You need a thread-safe queue. `LinkedBlockingQueue` is the canonical one, is fewer lines of code to use, and less likely to have a bug (other than maybe it having a more complex API).

servlet/src/main/java/io/grpc/servlet/ServerBuilder.java
dapengzhang0: Done.

settings.gradle
dapengzhang0: Done.

servlet/src/main/java/io/grpc/servlet/ServerStream.java
dapengzhang0: Why do you suggest `LinkedBlockingQueue`? Even `ConcurrentLinkedQueue` would be enough. I don't need the thread-safety it provides. The thread-safety is handled at the its usage outside, the thread-safety by the queue itself does not help and is still not enough.

ejona86: `ConcurrentLinkedQueue` is fine. I'm mainly arguing to not write your own unless you get some benefit from it.

ejona86: If you don't need thread-safety, then ArrayDeque or LinkedList would be appropriate. Both implement `java.util.Queue`.

dapengzhang0: The concurrency is handled by the `writeState` state machine. Unfortunately it's only capable of handling the concurrency for singly-linked queue. Deque or doubly-linked queue would messed up the state. `ArrayDeque` would work if there were no `doubleCapacity()`. So from `java.util.*`  only `ConcurrentLinkedQueue` can be used.

servlet/src/main/java/io/grpc/servlet/ServletAdapter.java
dapengzhang0: My thought was to be able to use CDI, i.e. `@Inject ServletAdapter servletAdapter`, in the servlet, and the servlet code needn't change when the producer is change from a grpc lib impl to a test object or to a container-vendor-provided impl. But it turned out that the `@PostConstruct` method to shutdown the server is not getting invoked if the producer uses `@Produces` method as in the example. So there is no advantage of using CDI for servletAdapter.

Changed the interface to a final class.

servlet/src/main/java/io/grpc/servlet/ServletAdapter.java
dapengzhang0: I don't mind to add a static method `ServletAdapter.create(Builder)` instead of `ServletAdapter.Factory.create(Builder)`, but add a new method to `Builder` might confuse the user because it already has a `build()` method.

dapengzhang0:
Added InteropTest and TransportTest with embedded Undertow servlet container. Rebased.
@ejona86 PTAL

servlet/src/main/java/io/grpc/servlet/ServletServerStream.java
```diff
@@ -0,0 +1,425 @@
...
+
+      if (flush) {
+        writeFrame(FLUSH);
+      }
+    }
+
+    private void writeFrame(ByteArrayWritableBuffer byteBuffer) {
+      int numBytes = byteBuffer.readableBytes();
+      WriteState curState = writeState.get();
+      if (curState.stillWritePossible) {
```
ejona86: This `stillWritePossible` logic seems complicated. It seems a lock would make this much more straight-forward.

Here it would be:
```java
synchronized (lock) {
  if (waitingOnWritePossible || !outputStream.isReady()) {
    waitingOnWritePossible = true;
    queue.offer(byteBuffer);
    return;
  }
}
// do the write
// No checking for isReady
```

In `onWritePossible`:
```java
while (true) {
  Object item;
  synchronized (lock) {
    if (!waitingOnWritePossible || queue.isEmpty()) {
      waitingOnWritePossible = false;
      return;
    }
  }
  item = queue.poll();
  // write
  if (!outputStream.isReady()) {
    return;
  }
}
```

I think the `isReady()` check within the lock is likely required, as otherwise a callback on onWritePossible could be missed. And it would be very complex to try to avoid that (which I think is the case in this code).

To check at the end:
```java
synchronized (lock) {
  if (waitingOnWritePossible) {
    queue.offer(byteBuffer);
    return;
  }
}
// do the write
synchronized (lock) {
  if (!outputStream.isReady()) {
    waitingOnWritePossible = true;
  }
}
```

servlet/src/main/java/io/grpc/servlet/ServletServerStream.java
```diff
@@ -0,0 +1,425 @@
...
+        try {
+          ServletOutputStream outputStream = resp.getOutputStream();
+          if (byteBuffer == FLUSH) {
+            resp.flushBuffer();
+          } else {
+            outputStream.write(byteBuffer.bytes, 0, byteBuffer.readableBytes());
```
ejona86: The writes should not be happening in two different files. The interactions with the writes is very important. I'd suggest having the listener call the stream when onWritePossible is called.

servlet/src/main/java/io/grpc/servlet/ServletServerStream.java
```diff
@@ -0,0 +1,425 @@
...
+          String key = new String(serializedHeaders[i], StandardCharsets.US_ASCII);
+          String newValue = new String(serializedHeaders[i + 1], StandardCharsets.US_ASCII);
+          String value = trailerSupplier.get().putIfAbsent(key, newValue);
+          if (value != null) {
+            trailerSupplier.get().put(key, value + "," + newValue);
+          }
+        }
+      }
+
+      while (true) {
```
ejona86: How does the servlet know when we are done writing? I see we have placed the trailers in place ready to be read, but what is the trigger? I sort of expected to see an `outputStream.close()` or similar.

ehallander9591:
It seems like this is moving closer to being included. I will keep monitoring it.

dapengzhang0:
@ehallander9591 Thank you for following the progress. I'll still working on it. Will send out updated PR in a week or so.

dapengzhang0:
@creamsoup Would you like to be a second reviewer, on AbstractStream part at least?

creamsoup:
@dapengzhang0, sure.

examples/example-servlet/src/main/proto/helloworld/helloworld.proto
```diff
@@ -0,0 +1,37 @@
+// Copyright 2015 The gRPC Authors
```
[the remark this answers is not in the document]
dapengzhang0: I'm not sure we should update the copyright year of an unmodified/moved copy of an existing file.

dapengzhang0:
@creamsoup

* `WriteListener#onWritePossible` just calls the `doWrite` (similar to `SynchronizedContext#drain`).

* any writes in `Sink` (header, data frame, trailers etc) need to be queued (`SynchronizedContext#executeLater`).
after queue the write, it must call `doWrite` immediately. Those two steps are similar to `SynchronizedContext#execute`.

* `doWrite` checks Reentrancy to make sure there is only one write just like `SynchronizedContext#drain`. the later `doWrite` returns immediately.

This is a great idea. However, on a second thought of the above approach, I found that it may cause the container to be in an illegal state in the following scenario:

(1) After some `doWrite()`, `outputStream.isReady()` returns false. An `onWritePossible()` callback will be called by the container later.
(2) Then the sink tries to write some data.There is a race between Sink's doWirte and `onWritePossible()`. Suppose Sink's write wins the race. The sink will check `outputStream.isReady()` before the already scheduled/promised `onWritePossible()` is called. This violate the contract, the behavior is undefined, and may cause problem for some containers.

To solve this problem, we may still need something like `writeState.stillWirtePossible`, which adds complexity.

Also, we can not completely reuse `SynchronizedContext` or `SerializingExecutor` because we can not `drain()` all the buffered things. Instead, we need `drainWhileReady()`. This also adds complexity, and hard to prove the implementation's correctness.

ejona86:
I remember vaguely looking at the write flow before. I agree it is complicated. Since the writes are async, I think we should just use a lock for the write path, including writing to the async output stream. It seems that would greatly simplify the code.

servlet/src/main/java/io/grpc/servlet/ServletServerStream.java
```diff
@@ -0,0 +1,526 @@
...
+      if (!headersSent) {
+        // Discard any application supplied duplicates of the reserved headers
+        trailers.discardAll(CONTENT_TYPE_KEY);
+        trailers.discardAll(GrpcUtil.TE_HEADER);
+        trailers.discardAll(GrpcUtil.USER_AGENT_KEY);
+
+        resp.setStatus(HttpServletResponse.SC_OK);
+        resp.setContentType(CONTENT_TYPE_GRPC);
+
+        byte[][] serializedHeaders = TransportFrameUtil.toHttp2Headers(trailers);
```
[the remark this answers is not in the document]
dapengzhang0: Done.

dapengzhang0:
@ejona86 Current write path implementation is so difficult to see the correctness. Will test out using a lock for the write path.  Will update the PR if it works for all the three major containers. If any of containers has an issue, I will update the PR once the issue is fixed on the container side (Must guarantee the impl is well tested on three containers).

dapengzhang0:
Inspired by @creamsoup , I found a simple and clean solution, reusing `SerializingExecutor` or `SynchronizationContext`, without lock, without other AtomicReference:

```java
Executor wirteExecutor = new SerializingExecutor(directExecutor());// SyncCtx also works
boolean isWriteReady;
Queue< WritableBuffer> wirteQueue = new ArrayDeque();

void onWritePossible() {
  writeExecutor.execute(
      () -> {
        assert !isWriteReady;
        isWriteReady = outputStream.isReady(); 
        // the above is the ONLY place isWriteReady could flip from false to true

        while (isWriteReady && !writQueue.isEmpty()) {
          WritableBuffer buffer = writeQueue.poll();   
          outputStream.wirte(writeQueue.poll().readableBytes());
          isWriteReady = outputStream.isReady();
          // if the above returns false, another onWritePossible() will be called
          // by the container *later*,
        }
      });
}

// The difference from @creamsoup's approach is that writeFrame here never calls drain()
void writeFrame(WritableBuffer buffer) {
  writeExecutor.execute(
    () -> {
      if (isWriteReady) {
        assert writeQueue.isEmpty();
        outputStream.wirte(buffer.readableBytes());
        isWriteReady = outputStream.isReady();
        // if the above returns false, onWritePossible() will be called
        // by the container *later*
      } else {
        writeQueue.offer(buffer);
      }
    })
}
```

creamsoup:
@dapengzhang0, that looks very concise and easy to follow. does it work on all 3 servlets?

servlet/src/main/java/io/grpc/servlet/ServletAdapter.java
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  /**
+   * Call this method when the adapter is no longer needed. The gRPC server will be terminated.
+   */
+  public void destroy() {
+    transportListener.transportTerminated();
+  }
+
+  private static final class GrpcAsycListener implements AsyncListener {
```
fleurer: typo? Asyc -> Async

dapengzhang0: Yeah. Thanks. I also noticed that just had not a chance to fix yet.

servlet/src/test/java/io/grpc/servlet/UndertowInteropTest.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
+  @Override
+  protected AbstractServerImplBuilder<?> getServerBuilder() {
+    return new ServletServerBuilder().maxInboundMessageSize(AbstractInteropTest.MAX_MESSAGE_SIZE);
+  }
+
+  @Override
+  protected void startServer(ServerBuilder<?> builer) {
```
loriadi: protected void startServer(ServerBuilder<?> builder)

change builer to builder on this line and the next?

hypnoce:
Hi all,
is there a plan to get this reviewed and/or merged ? I'm considering forking grpc locally to build one supporting servlet unless this PR make it through.
Thanks

servlet/src/main/java/io/grpc/servlet/ServletServerStream.java
```diff
@@ -0,0 +1,342 @@
...
+      InternalLogId logId) throws IOException {
+    super(ByteArrayWritableBuffer::new, statsTraceCtx);
+    transportState =
+        new ServletTransportState(maxInboundMessageSize, statsTraceCtx, new TransportTracer());
+    this.attributes = attributes;
+    this.authority = authority;
+    this.logId = logId;
+    this.asyncCtx = asyncCtx;
+    this.resp = (HttpServletResponse) asyncCtx.getResponse();
+    resp.getOutputStream().setWriteListener(new GrpcWriteListener());
```
MattGill98, line 83: When testing against Payara server, this class seemingly fails to completely initialise unless the `writer` field is initialised before the write listener on this line. When swapped I get NPEs in the OutputBuffer, but I'm willing to assume this could just be an expected Grizzly bug. I do, however, at least get the expected communication between the client and server.

MattGill98 (changes requested):
Firstly, really nice work on this feature - I think it's a brilliantly useful idea. I've been doing some testing and one bug came up for me that I wanted to mention!

lprimak:
Is this still maintained?

sanjaypujare:
> Is this still maintained?

@dapengzhang0 are you planning to drive this to completion?

lprimak:
Trying to use this in Payara, and getting all sorts of intermittent errors. Doesn't look like this works correctly currently
Looks to me like this is abandoned. What a shame.

dapengzhang0:
> Trying to use this in Payara, and getting all sorts of intermittent errors. 

@lprimak, I tested GlassFish 5.0/Grizzly, it didn't pass all the interop tests like Jetty/UnderTow/Tomcat. Seems there are some HTTP/2 & aysnc-servlet issues in Grizzly. Payara is also based on Grizzly kernel so there could be the same problems. I didn't have time to identify the Grizzly bugs. Sorry for being inactive for a long time.

lprimak:
@dapengzhang0 Thanks, but are you maintaining this branch? Will it ever be merged?
It's not compatible with the latest gRPC core even.

lprimak:
>  I tested GlassFish 5.0/Grizzly, it didn't pass all the interop tests like Jetty/UnderTow/Tomcat. Seems there are some HTTP/2 & aysnc-servlet 

Looks like these issues have been fixed in Payara in 2018 @dapengzhang0

lprimak:
Can you at least update this PR to fix https://github.com/grpc/grpc-java/pull/4738#pullrequestreview-617677307 and have it working on the lastes gRPC-Java please?
Thank you!

servlet/src/main/java/io/grpc/servlet/ServletServerStream.java
dapengzhang0: In https://github.com/grpc/grpc-java/pull/8596/commits/c903d75c1aca0dd29406282e20e1a0b37759b00b#diff-61714cd00cc4b70a569fd5ac7ac6919bcc9e2e76b6c0384ffbbeda9587c5ff93 of #8596, I swapped the assignment of the `writer` field. Thanks for catching this.

rufreakde:
@dapengzhang0 is this worked on? Looking forward to this! <3

dapengzhang0:
A newer version of this work is #8596, although there is no update on it for a long time too.  Close this PR.

hutchig:
https://github.com/grpc/grpc-java/pull/8596 is now merged in https://github.com/grpc/grpc-java/releases/tag/v1.53.0

Issue #2783, opened 2017-03-03
Test historic generated code


ejona86:
We should check in generated code for each minor version of grpc, and at least makes sure it compiles. If it compiles, it is highly likely to work.

Ideally we'd run some simple sanity tests, although it is hard to force all the code to run. Maybe we create a noop-ImplBase service and call bindService(). That would at least load all the MethodDescriptors... I don't think a literal test is necessary before closing this. The compilation itself can count as a "test." But it should at least be briefly investigated before closing.

PR #4739, opened 2018-08-06
compiler: Test-compile old codegen

ejona86:
This ensures we don't accidentally break compilation of old generated
code.

It also contains a cleanup commit:
compiler: Rename golden file to match output

The two commits will remain separate when merging.


ejona86:
#2783

PR #4743, opened 2018-08-08
core: hide Runnable from SerializingExecutor

dapengzhang0:
`run()` was a public method of `SerializingExecutor` but should never be called outside of this class.


carl-mastrangelo:
This was done on purpose to avoid the extra allocation of the runnable.   Really, this class should also extended ConcurrentLinkedQueue to avoid that allocation as well.

dapengzhang0:
I don't understand this extra allocation could cost what. This extra Runnable is allocated at most once per call and once per transport, given that creating a new call or a new transport object will also allocate tons of other objects than a single Runnable with the same lifespan.

dapengzhang0:
So beside final interface, we may also need interface with protected method.

PR #4753, opened 2018-08-13
README.md: Reorganize and trim

### LINKED ISSUE #4615: Basic setup doesn't work. make better documentation
I used the autogenerated code from protoc. I tried to use the client to make a simple call and the dependencies are all messed up. This might not be the right location for this issue, but I'm not sure where else to post it. Also, seeing that you have over 300 issues, this will probably never be resolved. That being said, after having to figure out the correct gradle magic spell to pull in the correct library and version of tcnative, I finally got it to work (apparently it's different depending on the OS). Why doesn't grpc-java just pull in the correct version(s) of tcnative? or why don't the docs say anything about how/what/where to get the basic hello world example to work.

1.13.1

something that builds without scouring the internet for days while wanting to blow my brains out

ejona86:
Reorganize the README to flow better for new users, pointing them in
good directions. Also try to reduce the size so they actually read the
thing. It's still to long, but is a step in the right direction.

Fixes #4615

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,8 +30,8 @@ For a guided tour, take a look at the [quick start
 guide](https://grpc.io/docs/quickstart/java.html) or the more explanatory [gRPC
 basics](https://grpc.io/docs/tutorials/basic/java.html).
 
-The [examples](https://github.com/grpc/grpc-java/tree/master/examples) and the
-[Android example](https://github.com/grpc/grpc-java/tree/master/examples/android)
+The [examples](https://github.com/grpc/grpc-java/tree/v1.14.0/examples) and the
+[Android example](https://github.com/grpc/grpc-java/tree/v1.14.0/examples/android)
 are standalone projects that showcase the usage of gRPC.
 
 Download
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,6 +23,17 @@ Provider on Android. Please see the [Security Readme](SECURITY.md).
...
 [![Coverage Status](https://coveralls.io/repos/grpc/grpc-java/badge.svg?branch=master&service=github)](https://coveralls.io/github/grpc/grpc-java?branch=master)
 
+Getting Started
+---------------
+
+For a guided tour, take a look at the [quick start
+guide](https://grpc.io/docs/quickstart/java.html) or the more explanatory [gRPC
+basics](https://grpc.io/docs/tutorials/basic/java.html).
+
+The [examples](https://github.com/grpc/grpc-java/tree/master/examples) and the
```
dapengzhang0: point to the release branch?

ejona86: Done.

PR #4754, opened 2018-08-13
alts: Use grpc-netty-shaded instead of grpc-netty

ejona86:
There's no good way to provide users of ALTS a choice between grpc-netty
and grpc-netty-shaded. Since Netty is not exposed through the ALTS API
surface, we opt for the shaded version as it has fewer deployment
issues. However, this also means that we _can't_ expose any Netty API,
like EventLoopGroup.

--------------

The core problem is that we need to be able to use ALTS with grpc-netty-shaded, as it greatly reduces the diamond dependency issues users experience. With grpc-netty it may not be possible to produce a working executable, depending on what your other dependencies are. ALTS is the first dependency we've had on grpc-netty.

It seems we are painting ourselves into a corner with ALTS, as I feel like I'm threading a needle amongst all the constraints involved. But I don't see any path that avoids doing so. So while I'm not happy with this approach, it seems better than all the other approaches.

CC @jiangtaoli2016

### DIFF
--- a/alts/build.gradle
+++ b/alts/build.gradle
@@ -1,3 +1,10 @@
+buildscript {
+    repositories { jcenter() }
+    dependencies { classpath 'com.github.jengelman.gradle.plugins:shadow:2.0.4' }
+}
+
+apply plugin: 'com.github.johnrengelman.shadow'
+
 description = "gRPC: ALTS"
 
 sourceCompatibility = 1.7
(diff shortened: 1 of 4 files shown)


jiangtaoli2016:
So ALTS will not be able to use eventLoopGroup in the future with this change? It does not use eventLoopGroup now though.

ejona86:
EventLoop is a resource, and we've seen users need to configure their resources. We simply won't be able to. This means that users can't configure the number of threads being used, and simply letting them specify the number of threads wouldn't quite work because users would want the threads to be shared across multiple Channels.

jiangtaoli2016:
Thanks Eric for clarification and fix! It seems that internal ProdChannelBuilder cannot config EventLoop as well. If in the future power-user of ALTS needs to configure EventLoop, we can revisit this.

ejona86:
> If in the future power-user of ALTS needs to configure EventLoop, we can revisit this.

Not really, as in the future we can't change the API. Unless the HTTP/2 API in Netty becomes stable, I don't think we'd be able to change the decision. I don't really see alternatives once the public API is stable.

ejona86:
Ugh. So the non-shaded and shaded jars are fighting with each other. I checked, and this is happening with grpc-netty-shaded as well, but we aren't depending on it within gradle so we didn't notice. I'll need to figure out something, but it is hard because `artifacts {}` implicitly includes `jar` via some magical lazy-initialization class and I'm having trouble figuring out how to disable it.

ejona86:
PTAL

PR #4758, opened 2018-08-14
examples: kotlin example test sources should be in  test sourceset

zpencer:
Putting them in the main sourceset is incorrect.


ejona86:
Looks like I did this #4794 without realizing this PR existed.

PR #4763, opened 2018-08-15
Implement flush coalescing in OkHttp.

### LINKED ISSUE #3258: Implement flush coalescing in OkHttp
`OkHttpClientTransport` currently forces the connection to use `TCP_NODELAY`: https://github.com/grpc/grpc-java/blob/master/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java#L426

For my use case (and, I would argue, other Android "next billion users" use cases) prioritizing low resource usage over performance seems to make more sense, so I would like to see this option to be configurable. I agree that enabling TCP_NODELAY seems the better default in general: https://github.com/grpc/grpc-java/pull/256

In my case I have a good number of low-throughput streaming RPCs that will stay open until the connection is closed. Speed and latency are secondary, low resource usage and reliability are important. Most of the streaming RPCs are primarily intended to let the server stream any changes in system state. For my Android "next billion users" use case I expect the connection to break fairly frequently, at which point the client will re-establish the streaming RPCs, so that it can catch up with the server, and the server can resume streaming any changes in system state. Consequently the first thing that will happen upon reconnect is all the streaming RPCs piping their header information down the new connection. With TCP_NODELAY enabled I expect this to mean several packets PER STREAMING RPC. Without that flag I would hope to see no more than a few packets TOTAL. Because the overall throughput of actual state data is low for this system and connections could break frequently, I think this could have a notable effect on total mobile data used.

I am still in the design phase for the system and identifying the potential bottlenecks from the perspective of a new gRPC & HTTP2 user, so please excuse and correct any false assumptions. I realize I could probably use just one streaming RPC to pipe data down to the client, but that kind of defeats the purpose of a protobuf-based RPC design, as I have different types of data to send. Being able to squeeze the connection initialization data into a few packets would make the overhead more negligible.

creamsoup:
Implementation is skipping flush task if another flush is queued.

Resolves #3258


okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
```diff
@@ -0,0 +1,138 @@
...
+  @Mock(answer = Answers.CALLS_REAL_METHODS)
+  private PartiallyTrackingFrameWriter frameWriter;
+
+  private QueueingExecutor queueingExecutor;
+  private AsyncFrameWriter asyncFrameWriter;
+
+  @Before
+  public void setUp() throws Exception {
+    queueingExecutor = new QueueingExecutor();
+    asyncFrameWriter = spy(new AsyncFrameWriter(transport, new SerializingExecutor(queueingExecutor)));
```
ejona86: Please, please no spies. They are so very magical, I don't trust tests with them.

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  @Test
+  public void noCoalesceRequired() throws IOException {
+    asyncFrameWriter.ping(true, 0, 1);
+    asyncFrameWriter.flush();
+    queueingExecutor.runAll();
+
+    verify(frameWriter, times(1)).ping(anyBoolean(), anyInt(), anyInt());
+    verify(asyncFrameWriter, times(1)).flush();
```
ejona86: What is the point of this verification, here and elsewhere? Isn't this test the thing that called this method, on line 56?

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
```diff
@@ -0,0 +1,138 @@
...
+      Runnable r;
+      while ((r = runnables.poll()) != null) {
+        r.run();
+      }
+    }
+  }
+
+  private static abstract class PartiallyTrackingFrameWriter implements FrameWriter {
+
+    private static final List<Method> methods = new ArrayList<Method>();
```
ejona86: This is just asking for trouble, as one test can influence another. Actually... how is flushCoalescing_shouldMergeTwoQueuedFlushes working? Nothing is clearing out these methods.

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
```diff
@@ -0,0 +1,151 @@
+/*
+ * Copyright 2015 The gRPC Authors
```
carl-mastrangelo: 2018

creamsoup: thanks!

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
creamsoup: right. removed unnecessary verify.

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
creamsoup: spy is gone!

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
```diff
@@ -0,0 +1,160 @@
...
+    asyncFrameWriter.ping(true, 0, 1);
+    asyncFrameWriter.flush();
+    asyncFrameWriter.ping(true, 0, 2);
+    asyncFrameWriter.flush();
+
+    queueingExecutor.runAll();
+
+    verify(frameWriter, times(2)).ping(anyBoolean(), anyInt(), anyInt());
+    verify(frameWriter, times(1)).flush();
+    assertThat(Iterables.getLast(PartiallyTrackingFrameWriter.getAllInvokedMethodsHistory()))
```
ejona86: Why not use `inOrder`?

creamsoup: didn't know that exists. thanks!

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+    queueingExecutor.runAll();
+
+    InOrder inOrder = inOrder(frameWriter);
+    inOrder.verify(frameWriter, times(2)).ping(anyBoolean(), anyInt(), anyInt());
+    inOrder.verify(frameWriter).flush();
+  }
+
+  /**
+   * Executor queues incoming runnables instead of running it. Runnables can be invoked via {@code
```
carl-mastrangelo: this should be `#link`

creamsoup: done.

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
```diff
@@ -0,0 +1,120 @@
...
+  @Mock private OkHttpClientTransport transport;
+  @Mock private Socket socket;
+  @Mock private FrameWriter frameWriter;
+
+  private QueueingExecutor queueingExecutor;
+  private AsyncFrameWriter asyncFrameWriter;
+
+  @Before
+  public void setUp() throws Exception {
+    queueingExecutor = new QueueingExecutor();
```
ejona86: nit: Initialize at the declaration above.

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
```diff
@@ -0,0 +1,120 @@
...
+    asyncFrameWriter = new AsyncFrameWriter(transport, new SerializingExecutor(queueingExecutor));
+    asyncFrameWriter.becomeConnected(frameWriter, socket);
+  }
+
+  @After
+  public void tearDown() throws Exception {
+    asyncFrameWriter.close();
```
ejona86: What's the point of this? It is guaranteed to do nothing. If we wanted to close the (mock) socket there would need to be a `runAll()` here.

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
```diff
@@ -0,0 +1,120 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.InOrder;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class AsyncFrameWriterTest {
+
+  @Mock private OkHttpClientTransport transport;
```
ejona86: Mocking a functional class like the transport is worrisome to me. Only a single method is used, could we make an interface for it and have the transport implement that interface?

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
creamsoup: done.

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
creamsoup: yes, created an interface TransportExceptionHandler, can you review this part?

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
creamsoup: done.

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
```diff
@@ -0,0 +1,137 @@
...
+  @Before
+  public void setUp() throws Exception {
+    asyncFrameWriter =
+        new AsyncFrameWriter(transportExceptionHandler, new SerializingExecutor(queueingExecutor));
+    asyncFrameWriter.becomeConnected(frameWriter, socket);
+  }
+
+  @After
+  public void tearDown() throws Exception {
+    queueingExecutor.clear();
```
ejona86: This does nothing, as this test class instance will be thrown away. So you're clearing something before it is cleared by the GC.

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
```diff
@@ -0,0 +1,137 @@
...
+      runnables.clear();
+    }
+  }
+
+  /** Rethrows as Internal error. */
+  private static class EscalatingTransportErrorHandler implements TransportExceptionHandler {
+
+    @Override
+    public void onException(Throwable throwable) {
+      throw Status.INTERNAL.asRuntimeException();
```
ejona86: This throws away the Throwable, and we don't use StatusRuntimeException too much like this. How about just `new AssertionError(throwable)`?

okhttp/src/test/java/io/grpc/okhttp/AsyncFrameWriterTest.java
```diff
@@ -0,0 +1,137 @@
...
+  private QueueingExecutor queueingExecutor = new QueueingExecutor();
+  private TransportExceptionHandler transportExceptionHandler =
+      new EscalatingTransportErrorHandler();
+
+  private AsyncFrameWriter asyncFrameWriter;
+
+  @Before
+  public void setUp() throws Exception {
+    asyncFrameWriter =
+        new AsyncFrameWriter(transportExceptionHandler, new SerializingExecutor(queueingExecutor));
```
ejona86: nit: this could also be moved up to the declaration, now that the transport isn't a mock.

PR #4765, opened 2018-08-15
gae-interop-testing: increase response verbosity

zpencer:
for jdk7 include the passing test names in the order they were run
for both, include the JUL log output

### DIFF
--- a/gae-interop-testing/gae-jdk7/src/main/java/io/grpc/testing/integration/OkHttpClientInteropServlet.java
+++ b/gae-interop-testing/gae-jdk7/src/main/java/io/grpc/testing/integration/OkHttpClientInteropServlet.java
@@ -30,6 +30,10 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.List;
+import java.util.logging.Handler;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
+import java.util.logging.SimpleFormatter;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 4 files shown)


gae-interop-testing/gae-jdk7/src/main/java/io/grpc/testing/integration/OkHttpClientInteropServlet.java
```diff
@@ -49,6 +53,42 @@
...
+      }
+      return sb.toString();
+    }
+  }
+
+  private final LogEntryRecorder handler = new LogEntryRecorder();
+
+  @Override
+  public void init() {
+    Logger.getLogger("").addHandler(handler);
```
ejona86: This is init for the servlet. Surely a new servlet isn't created for each request, right? Keeping all logs all the time is worrisome. Maybe just do this within the request?

zpencer: I was thinking this is fine for the narrow use case of this servlet, but now I realize the HTTP request may be retried and we don't care about previous entries. I'll do it naively here instead of using the fancier servlet APIs.

gae-interop-testing/gae-jdk7/src/main/java/io/grpc/testing/integration/OkHttpClientInteropServlet.java
```diff
@@ -49,8 +55,45 @@
...
+      }
+      return sb.toString();
+    }
+  }
+
   @Override
   public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+    LogEntryRecorder handler = new LogEntryRecorder();
+    try {
+      Logger.getLogger("").addHandler(handler);
```
ejona86: Do this before the try.

PR #4770, opened 2018-08-17
core: minor cleanup of NameResolverProvider

carl-mastrangelo:
* Make the list of providers an immutable List
* Make obvious that the list is statically initialized
* Add documentation for when methods were added.
* Log on missing classes
* make more classes final, hide constructors.
* Use proper UnsupportedOperationException, rather than IllegalStateException.

### DIFF
--- a/core/src/main/java/io/grpc/NameResolverProvider.java
+++ b/core/src/main/java/io/grpc/NameResolverProvider.java
@@ -127,7 +127,7 @@ public abstract class NameResolverProvider extends NameResolver.Factory {
       if (providers.isEmpty()) {
         String msg = "No NameResolverProviders found via ServiceLoader, including for DNS. "
             + "This is probably due to a broken build. If using ProGuard, check your configuration";
-        throw new UnsupportedOperationException(msg);
+        throw new RuntimeException(msg);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/NameResolverProvider.java
```diff
@@ -115,26 +124,41 @@ public String getDefaultScheme() {
     }
 
     private void checkForProviders() {
-      Preconditions.checkState(!providers.isEmpty(),
-          "No NameResolverProviders found via ServiceLoader, including for DNS. "
-          + "This is probably due to a broken build. If using ProGuard, check your configuration");
+      if (providers.isEmpty()) {
+        String msg = "No NameResolverProviders found via ServiceLoader, including for DNS. "
+            + "This is probably due to a broken build. If using ProGuard, check your configuration";
+        throw new UnsupportedOperationException(msg);
```
zpencer: I would prefer to leave the exception as is, since UnsupportedOperationException implies to me that the class as defined does not support this operation at all.

carl-mastrangelo: ISE has a specific meaning: methods were called out of order.   This is for calling read() after close()  and patterns like that.  I'm open to another Exception, maybe just plain RuntimeException?

zpencer: RuntimeException would be ok. This change probably won't cause problems for users but worth documenting in the release notes.

carl-mastrangelo: Done.

PR #4773, opened 2018-08-17
gae-interop-testing/jdk7: Cleanup after Assert/Assume

ejona86:


gae-interop-testing/gae-jdk7/src/main/java/io/grpc/testing/integration/OkHttpClientInteropServlet.java
```diff
@@ -178,6 +177,17 @@ private void doGetHelper(HttpServletRequest req, HttpServletResponse resp) throw
     writer.println(sb);
   }
 
+  private static AutoCloseable toCloseable(final Object o, final List<Method> methods) {
+    return new AutoCloseable() {
+      @Override
+      public void close() throws Exception {
+        for (Method method : methods) {
```
zpencer: If one `@After` throws, should the rest still run?

ejona86:
@zpencer, PTAL

PR #4774, opened 2018-08-17
interop-testing: Clean up even if interrupted

ejona86:
Most environments won't trigger an interruption, but AppEngine will if
the request takes too long.

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -296,9 +296,15 @@ public abstract class AbstractInteropTest {
 
   /** Clean up. */
   @After
-  public void tearDown() throws Exception {
+  public void tearDown() {
     if (channel != null) {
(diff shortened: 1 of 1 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -296,9 +296,15 @@ public void setUp() {
...
   @After
-  public void tearDown() throws Exception {
+  public void tearDown() {
     if (channel != null) {
-      channel.shutdownNow().awaitTermination(1, TimeUnit.SECONDS);
+      channel.shutdownNow();
+      try {
+        channel.awaitTermination(1, TimeUnit.SECONDS);
+      } catch (InterruptedException unused) {
+        // Best effort. If there is an interruption, we want to continue cleaning up, but quickly
```
carl-mastrangelo: this should be logged.

ejona86: Done, at FINE level. Since we do `shutdownNow()` I had considered just removing the awaitTermination completely.

PR #4780, opened 2018-08-17
core: add stackless variants for asException and use it throughout

carl-mastrangelo:
I tried my best to chase each of these exceptions to see where they were consumed.   The only *serious* change in API is at the stub level, where stackless Runtime exceptions are now used.  This makes most RPC Exceptions purely expose the message, code, and cause.

Key changes:

* Add `asStacklessException` and `asStacklessRuntimeException`  
* Swap nearly all in gRPC usages to these methods.  Some places were not swapped (like in StatusProto) as the API contract is not so clear here.
* Tests updated to unwrap the exception, rather than cast.  
* Avoid using StatusException in favor of StatusRuntimeException.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
@@ -26,7 +26,7 @@ import com.google.common.base.Supplier;
 import io.grpc.Attributes;
 import io.grpc.Metadata;
 import io.grpc.Status;
-import io.grpc.StatusException;
+import io.grpc.StatusRuntimeException;
 import io.grpc.internal.Channelz;
 import io.grpc.internal.ClientStreamListener.RpcProgress;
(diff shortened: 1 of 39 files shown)


netty/src/main/java/io/grpc/netty/NettyClientHandler.java
```diff
@@ -484,7 +484,7 @@ private void createStream(CreateStreamCommand command, final ChannelPromise prom
     final int streamId;
     try {
       streamId = incrementAndGetNextStreamId();
-    } catch (StatusException e) {
+    } catch (StatusRuntimeException e) {
```
carl-mastrangelo: It's unfortunate that Exceptions are being used a flow control for a normal event.

larry-safran:
This could help performance, but is far out of sync with the current code base and would need to be redone.  Therefore closing.

PR #4785, opened 2018-08-20
stub: disableAutoInboundFlowControl javadoc correction

### LINKED ISSUE #4772: Doc seems to be self-contradicting on CallStreamObserver.disableAutoInboundFlowControl

looking at master HEAD

See https://github.com/grpc/grpc-java/blob/master/stub/src/main/java/io/grpc/stub/CallStreamObserver.java#L80-L93

Given that last paragraph I'd expect to see SERVER_STREAMING (1 request to * responses) to be affected on the server and CLIENT_STREAMING (* requests to 1 response) on the client, since these are the ones that have one inbound message.

SERVER_STREAMING has multiple inbound messages from the client's perspective and the same is true for CLIENT_STREAMING from the server's point of view, so listing those in the `<ul>` contradicts the part above.

zpencer:
The docs refer to the wrong places where this feature is ignored.

Fixes #4772


tferi:
Thank you for addressing the issue.

PR #4786, opened 2018-08-20
core: narrow SharedResourceHolder types, and make the scheduler unconfigurable  

carl-mastrangelo:


carl-mastrangelo:
The type cast is safe because SharedResourceHolder asserts instance equality before calling close.

carl-mastrangelo:
ping

PR #4790, opened 2018-08-22
core: plumb hedging policy

dapengzhang0:
This is only API plumbing for hedging, following exactly the same way as its retry counterpart, so it is almost trivial.

### DIFF
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -424,9 +424,9 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
 
 
   /**
-   * Disables the retry mechanism provided by the gRPC library. This is designed for the case when
-   * users have their own retry implementation and want to avoid their own retry taking place
-   * simultaneously with the gRPC library layer retry.
+   * Disables the retry and hedging mechanism provided by the gRPC library. This is designed for the
+   * case when users have their own retry implementation and want to avoid their own retry taking
+   * place simultaneously with the gRPC library layer retry.
    *
    * @return this
    * @since 1.11.0
@@ -437,7 +437,7 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   }
 
   /**
-   * Enables the retry mechanism provided by the gRPC library.
+   * Enables the retry or hedging mechanism provided by the gRPC library.
    *
    * <p>This method may not work as expected for the current release because retry is not fully
    * implemented yet.
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -437,7 +437,7 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   }
 
   /**
-   * Enables the retry or hedging mechanism provided by the gRPC library.
+   * Enables the retry and hedging mechanism provided by the gRPC library.
    *
    * <p>This method may not work as expected for the current release because retry is not fully
    * implemented yet.
--- a/core/src/main/java/io/grpc/internal/HedgingPolicy.java
+++ b/core/src/main/java/io/grpc/internal/HedgingPolicy.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2018 The gRPC Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.internal;
+
+import com.google.common.base.MoreObjects;
+import com.google.common.base.Objects;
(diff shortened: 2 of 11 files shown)


core/src/main/java/io/grpc/internal/HedgingPolicy.java
```diff
@@ -0,0 +1,92 @@
...
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("maxAttempts", maxAttempts)
+        .add("hedgingDelayNanos", hedgingDelayNanos)
+        .add("nonFatalStatusCodes", nonFatalStatusCodes)
+        .toString();
+  }
+
+  /**
+   * Providers the most suitable hedging policy for a call when this will have to provide one.
```
ericgribkoff: s/Providers/Provides/

ericgribkoff: What does "when this will have to provide one" mean here?

core/src/main/java/io/grpc/internal/HedgingPolicy.java
```diff
@@ -0,0 +1,92 @@
...
+   *
+   * <p>Never returns null.
```
ericgribkoff: I think it makes more sense to talk about return value on the method's comments, rather than the interface itself

core/src/main/java/io/grpc/internal/ServiceConfigInterceptor.java
```diff
@@ -190,6 +201,7 @@ public String toString() {
           .toString();
     }
 
+    @SuppressWarnings("BetaApi") // Verify isn't all that beta
```
ericgribkoff: extra nitpicky nit: "isn't all that beta" is a little...informal. The same wording is already used in several other places in the repo (thus proving it's part of guava's stable api ;)), so it looks fine here too

core/src/main/java/io/grpc/internal/ServiceConfigInterceptor.java
```diff
@@ -236,9 +249,43 @@ private static RetryPolicy retryPolicy(Map<String, Object> retryPolicy, int maxA
...
+  @SuppressWarnings("BetaApi") // Verify isn't all that beta
+  private static HedgingPolicy hedgingPolicy(
+      Map<String, Object> hedgingPolicy, int maxAttemptsLimit) {
+    int maxAttempts = checkNotNull(
+        ServiceConfigUtil.getMaxAttemptsFromHedgingPolicy(hedgingPolicy),
+        "maxAttempts cannot be empty");
+    checkArgument(maxAttempts >= 2, "maxAttempts must be greater than 1: %s", maxAttempts);
+    maxAttempts = Math.min(maxAttempts, maxAttemptsLimit);
+
+
```
ericgribkoff: extra newline

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -437,7 +437,7 @@ public T disableRetry() {
   }
 
   /**
-   * Enables the retry mechanism provided by the gRPC library.
+   * Enables the retry or hedging mechanism provided by the gRPC library.
```
ericgribkoff: I think this should be "and hedging mechanisms" instead of "or"

core/src/main/java/io/grpc/internal/HedgingPolicy.java
```diff
@@ -0,0 +1,92 @@
...
+
+package io.grpc.internal;
+
+import com.google.common.base.MoreObjects;
+import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableSet;
+import io.grpc.Status.Code;
+import java.util.Collections;
+import java.util.Set;
+import javax.annotation.ParametersAreNonnullByDefault;
```
ericgribkoff: FYI I don't think this annotation is used anywhere else in the repository. Not sure how useful it will be if its only used in isolation in this one class?

core/src/main/java/io/grpc/internal/HedgingPolicy.java
ericgribkoff: I see we already had this [discussion](https://github.com/grpc/grpc-java/pull/4310#discussion_r179897836) :)

Happy to defer to your judgement here, but my suggestion would be that just leaving this as "Provides the most suitable hedging policy for a call." would be a sufficient comment. Otherwise, adding something more specific like "Provides the most suitable hedging policy for a call, taking into account the service config and factors like number of attempts sent already" might be more informative?

core/src/main/java/io/grpc/ManagedChannelBuilder.java
dapengzhang0: Done.

core/src/main/java/io/grpc/internal/HedgingPolicy.java
dapengzhang0: I don't have strong opinion on it. Just try to stop using `@Nonnull` annotations. Removed `@ParametersAreNonnullByDefault`

core/src/main/java/io/grpc/internal/HedgingPolicy.java
dapengzhang0: Done.

core/src/main/java/io/grpc/internal/HedgingPolicy.java
dapengzhang0: Done. Just chose the simplest.

core/src/main/java/io/grpc/internal/ServiceConfigInterceptor.java
dapengzhang0: Modified with formal comment.

core/src/main/java/io/grpc/internal/ServiceConfigInterceptor.java
dapengzhang0: oops

PR #4791, opened 2018-08-24
core: implement hedging

dapengzhang0:
Because we built up a good foundation when implementing RETRY, the HEDGING implementation turns out not so invasive, mostly filling up code at the existing comments in `RetriableStream.java` like `//TODO: for hedging do this and that ... `.

By the way, I replace the field `RetriableStream.scheduledRetry` with the field `RetriableStream.Substream.scheduledRetry`, because

```java
scheduledRetry = scheduledExecutorService.schedule(someRunnableThatCanChangeTheField);
```
is racy, whereas

```java
substream.scheduledRetry = scheduledExecutorService.schedule(someRunnableThatCreatesANewSubstream)
```
is safe.

Tests will be added shortly.


core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -77,21 +83,25 @@
   @Nullable
   private final Throttle throttle;
 
+  private Set<Substream> activeHedges; // not null and unchanged once isHedging = true
```
ejona86: I don't see any comprehensive thread-safety story. How is this class safe?

The synchronization involved between these new variables is really hard to follow. There are _4_ (actually 5 if you consider the timer) new independently synchronized variables.

Maybe it's also that it isn't clear the behavior of these variables. Like `activeHedges` I see things like checking `size() == 1` and iteration and it isn't clear to me how those are safe, logically. I understand they would be synchronized by the `Set`, but I don't understand how you can safely combine together logical conclusions if other threads could be performing concurrent modification of the set.

There is one new variable that _isn't_ synchronized: `latestSubstream`. However `hasPotentialHedging()` reads `latestSubstream` from the closed() callback, so it seems that it _should_ be synchronized because a new stream can start at any time.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -77,21 +83,25 @@
...
+  private ConcurrentRescheduler rescheduler; // not null and unchanged once isHedging = true
+
   private volatile State state = new State(
       new ArrayList<BufferEntry>(8), Collections.<Substream>emptyList(), null, false, false);
 
   /**
    * Either transparent retry happened or reached server's application logic.
    */
-  private boolean noMoreTransparentRetry;
+  private final AtomicBoolean noMoreTransparentRetry = new AtomicBoolean();
```
ejona86: I think this should be a property of the substream. It would 1) be simpler as there would not be any synchronization and 2) sort of seems more appropriate to allow one transparent retry _per attempt_. The spec isn't really clear how transparent retries mixes with normal retries and hedging. We should probably get that clarified.

core/src/main/java/io/grpc/internal/RetriableStream.java
dapengzhang0: `activeHedges` and `rescheduler` can be regarded as final fields. (The `RetriableStream.cancel()` thread may read stale values of them but it's not big of a deal because things will still be cleaned up soon)

`activeHedges.size() == 1` is only checked after `hasPotentialHedging() == false`, which means absolutely no new hedge will start. (The only one left active hedge can be closed in a race, but the RPC can still get committed whichever thread wins)

The value `latestSubstream` may be stale when calling `hasPotentialHedging()`, and thus `hasPotentialHedging()` returns `true` (which could have been `false` if strongly synchronized); but it doesn't matter (you can regard that `hasPotentialHedging()` were called before the real `latestSubstream` was born), at least one other hedge will observe `hasPotentialHedging()` return `false` when they are `closed()`.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -269,7 +291,35 @@ public final void start(ClientStreamListener listener) {
     class StartEntry implements BufferEntry {
       @Override
       public void runWith(Substream substream) {
+        if (hedgingPolicy == null) {
```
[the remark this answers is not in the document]
ejona86: But `drain()` runs now, as well. So I don't think this is doing anything.

core/src/main/java/io/grpc/internal/RetriableStream.java
ejona86: > activeHedges and rescheduler can be regarded as final fields

I wasn't speaking as much to their reference, but to their contents. They both maintain mutating state.

> activeHedges.size() == 1 is only checked after hasPotentialHedging() == false, which means absolutely no new hedge will start.

Let's say there's three hedges and they all fail at the same time. `state.winningSubstream == null` could be `true` for all three hedges. It's a bit hard to tell, but it seems `hasPotentialHedging()` could end up being true for two of those hedges.

I see another race with `hedgingFrozen`. It seems a hedge could be starting _right now_ within `HedgingRunnable`, it could see `hedgingFrozen == false`, then hedging is frozen because of `RpcProgress.DROPPED`, and then `HedgingRunnable` calls `createSubstream()` which will do `activeHedges.add()`.

Logic-ing out this sort of thing is _really hard_ and _error prone_. Doing that for one or two variables is many times necessary, but here we are combining `state`, a _different_ `state` within `hasPotentialHedging()`, unsynchronized `latestSubstream`, `hedgingFrozen`, and `activeHedges`.

> The value latestSubstream may be stale when calling hasPotentialHedging()

This is something _really_ important to call out in the code. But also, I know Carl won't be happy about that because the race detector will complain about it.

core/src/main/java/io/grpc/internal/RetriableStream.java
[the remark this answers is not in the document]
dapengzhang0: Yes, for delayed stream is really difficult to schedule the hedging, in this case need let `drain()` happen, and delay scheduling the hedging once name resolution completes. Otherwise there is no hedging policy, so no hedging at all (Unless there is a local service config). It's too difficult, so I put a comment there.

core/src/main/java/io/grpc/internal/RetriableStream.java
ejona86: Let's get confirmation on that, because I was in the meeting where it was discussed and I'd expect it to be one per attempt. I expect most people don't care too much, but if it is an opportunity to remove synchronized state, that would be a win.

core/src/main/java/io/grpc/internal/RetriableStream.java
ejona86: * `activeHedges`: trash. Let's not bother with early commitment. Because of transparent retries, most RPCs won't be able to early commit, so it's no use optimizing.
* `hedgingFrozen`: Merge into `state`. Could actually be handled by making previousAttempts really large. But I don't know if that is most clear in the code.
* `latestSubstream.previousAttempts`: Merge into `state`
* `latestSubstream.scheduledRetry`: since this is a pre-existing race and not related to hedging, let's handle this separately. So undo from this PR (TODO)
* `noMoreTransparentRetry`: check with spec authors whether we can do per-attempt transparent retry. If so, move this to `Substream`.

TODO (after first pass of simplification): interactions between `rescheduler` and `hedgingFrozen`/`previousAttempts`.

dapengzhang0:
Now I use the same lock as that of `RetriableStream` for `ConcurrentRescheduler`. I added a `precondition` object to `ConcurrentRescheduler` add the precondition is guarded by the same lock. This avoids a lot of race conditions that would occur if not using the precondition object inside `ConcurrentRescheduler` but checking the precondition before and after scheduling.

I make all the newly introduced state field guarded by the lock. I didn't merge them in the `state` object, and I think it is equivalent to merging them into `state`, which I don't mind to do if you do think it improves  readability but needs some boilerplate builders like

```java
@CheckReturnValue
// GuardedBy RetriableStream.lock
State withLatestAttempt(int latestAttempt) {...}

state = stateCopy.withLatestAttempt(latestAttempt);
```

In `Sublistener.closed()`, I couldn't make a single snapshot of state at the beginning and then call `freezeHeding()` later and call `hasPotentialHedging(stateCopy)` at the end because they are not in the same synchronized block.

For `activeHedges`, I found that even without early commit, I still need `numOfActiveHedges`. I didn't remove early commit yet.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -173,9 +194,29 @@ public ClientStreamTracer newClientStreamTracer(CallOptions callOptions, Metadat
...
     sub.stream = newSubstream(tracerFactory, newHeaders);
+    if (isHedging()) {
+      synchronized (lock) {
+        activeHedges.add(sub);
+      }
+    }
     return sub;
   }
 
+  private boolean isHedging() {
```
ejona86: This is a pretty scary method. 1) It has the same name as a variable that it initializes, but it isn't clear when to use the variable vs this method. 2) If called twice, it runs everything again, potentially blindly overwriting activeHedges. Since this is called from createSubstream, I think it is broken. 3) It's non-obvious that asking for the hedging policy sets the retry policy.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -613,30 +721,53 @@ public void run() {
...
-      if (!hasHedging()) {
-        commitAndRun(substream);
-        if (state.winningSubstream == substream) {
-          masterListener.closed(status, trailers);
+        if (isHedging) {
+          Substream sub = null;
+          synchronized (lock) {
+            activeHedges.remove(substream);
+            if (!isFatal) {
+              if (hasPotentialHedging()) {
```
ejona86: I'm bothered by the complexity involved with `activeHedges` and the synchronization here and the next several lines. I had mentioned previously that I wanted `state` to be passed into `hasPotentialHedging`, to make it obvious it was being used. But there's also interactions between state and createSubstream() and HedgingRunnable. I think it is all "correct", but it relies on non-obvious behaviors and takes too much effort to verify.

Most of the trouble seems to be the "grab the lock to modify activeHedges" in random places and trying to figure out what the guarantees are at those points. For example, things would be more obvious if HedgingRunnable changed activeHedges and latestAttempt without releasing the lock between. But I've not even gotten to observing interactions with `hedgingFrozen` and the rescheduler.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -278,9 +325,38 @@ public void runWith(Substream substream) {
...
-    // TODO(zdapeng): schedule hedging if needed
+  private final class HedgingRunnable implements Runnable {
+
+    @Override
+    public void run() {
+      callExecutor.execute(new Runnable() {
+        @Override
+        public void run() {
+          int previousAttempts;
+          synchronized (lock) {
```
ejona86: I'm really surprised this doesn't check `hasPotentialHedging` or the precondition.

core/src/main/java/io/grpc/internal/ConcurrentRescheduler.java
```diff
@@ -0,0 +1,182 @@
...
+    Future<?> existingTask = null;
+    FutureRunnable newWakeUp;
+
+    synchronized (lock) {
+      FutureRunnable existingWakeUp = wakeUp;
+      if (existingWakeUp != null) {
+        // cancel the existing one and schedule new one
+        existingWakeUp.cancelled = true;
+        existingTask = existingWakeUp.task;
+      } else if (precondition != null && !precondition.met()) {
```
ejona86: Is the point of the precondition just to avoid holding the lock when calling `schedule()`?

core/src/main/java/io/grpc/internal/RetriableStream.java
dapengzhang0: Now transparent retry does not change the size of `activeHedges`, does not affect `hasPotentialHedging`.

dapengzhang0:
@ejona86 PTALA

dapengzhang0:
Done. @ejona86 Thanks a lot for the review, and all the helpful discussions offline!

PR #4802, opened 2018-08-27
alts: convert handshaker service channel to SharedResourceHolder

jiangtaoli2016:


jiangtaoli2016:
@ejona86 This PR converts handshaker service channel to SharedResourceHolder. Once we have plumbing on shutdown at ProtocolNegotiator, we can release the resource.

alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
```diff
@@ -104,7 +105,8 @@ public AltsChannelBuilder enableUntrustedAltsForTesting() {
 
   /** Sets a new handshaker service address for testing. */
   public AltsChannelBuilder setHandshakerAddressForTesting(String handshakerAddress) {
-    HandshakerServiceChannel.setHandshakerAddressForTesting(handshakerAddress);
+    HandshakerServiceChannel.SHARED_HANDSHAKER_CHANNEL.setHandshakerAddressForTesting(
```
ejona86: This is really awkward right now, in that it is a builder-level method that sets a `static`. That's _really_ surprising and seems like a bad idea. I guess that's how it was before, but I just didn't realize it.

alts/src/main/java/io/grpc/alts/AltsServerBuilder.java
```diff
@@ -177,8 +179,12 @@ public Server build() {
               @Override
               public TsiHandshaker newHandshaker() {
                 // Used the shared grpc channel to connecting to the ALTS handshaker service.
+                // Release the channel if it is not used.
```
ejona86: Is this supposed to be a TODO?

alts/src/main/java/io/grpc/alts/GoogleDefaultChannelBuilder.java
```diff
@@ -116,7 +117,9 @@ TransportCreationParamsFilterFactory getTcpfFactoryForTest() {
           @Override
           public TsiHandshaker newHandshaker() {
             // Used the shared grpc channel to connecting to the ALTS handshaker service.
-            ManagedChannel channel = HandshakerServiceChannel.get();
+            // Release the channel if it is not used. https://github.com/grpc/grpc-java/issues/4755.
```
ejona86: TODO

alts/src/main/java/io/grpc/alts/HandshakerServiceChannel.java
```diff
@@ -16,57 +16,51 @@
...
+    public void close(ManagedChannel instance) {
+      instance.shutdown();
+    }
 
-  public static synchronized ManagedChannel get() {
-    if (channel == null) {
-      channel = createChannel();
+    @Override
+    public String toString() {
+      return NAME;
```
ejona86: Feel free to have the literal inline. `GrpcUtil.SHARED_CHANNEL_EXECUTOR` uses a constant `NAME` because it uses the string twice in the class.

alts/src/main/java/io/grpc/alts/HandshakerServiceChannel.java
```diff
@@ -16,57 +16,51 @@
...
+      ThreadFactory clientThreadFactory = new DefaultThreadFactory("handshaker pool", true);
+      return NettyChannelBuilder.forTarget(handshakerAddress)
+          .directExecutor()
+          .eventLoopGroup(new NioEventLoopGroup(1, clientThreadFactory))
+          .usePlaintext()
+          .build();
+    }
+
+    @Override
+    public void close(ManagedChannel instance) {
```
ejona86: You also need to shut down the `NioEventLoopGroup`. Although you can't actually do that until the channel is terminated. I'd suggest making it `instance.shutdownNow()` below to make it reasonably safe.

alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
jiangtaoli2016: I don't like this as well. I guess this is one reason we did not use SharedResourceHolder in the first place. Basically, ALTS client needs to connect to the handshaker service (metadata.google.internal:8080). In some testing environment, we override this address to another address, since we may not launch metadata server during testing. We use to use gFlags to control the testing address. But we cannot use gFlags here. 

Let me know if you have better idea. I am not sure I understand your proposal.

alts/src/main/java/io/grpc/alts/HandshakerServiceChannel.java
jiangtaoli2016: good catch. Revised.

jiangtaoli2016:
Code revised. PTAL

alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
ejona86: I provided a better idea: pass in an `ObjectPool` during testing.

PR #4807, opened 2018-08-28
alts: if ALTS is not running on GCP, fails call

jiangtaoli2016:
In ALTS, rather than throws RuntimeException if the application is not running on GCP, we fail at interceptCall.

### DIFF
--- a/alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
+++ b/alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
@@ -20,13 +20,15 @@ import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.common.annotations.VisibleForTesting;
 import io.grpc.CallOptions;
+import io.grpc.Channel;
 import io.grpc.ClientCall;
-import io.grpc.ConnectivityState;
+import io.grpc.ClientInterceptor;
 import io.grpc.ExperimentalApi;
 import io.grpc.ForwardingChannelBuilder;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import io.grpc.MethodDescriptor;
+import io.grpc.Status;
 import io.grpc.alts.internal.AltsClientOptions;
 import io.grpc.alts.internal.AltsProtocolNegotiator;
 import io.grpc.alts.internal.AltsTsiHandshaker;
@@ -43,6 +45,8 @@ import io.grpc.netty.NettyChannelBuilder;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.annotation.Nullable;
 
 /**
@@ -52,6 +56,7 @@ import javax.annotation.Nullable;
 @ExperimentalApi("https://github.com/grpc/grpc-java/issues/4151")
 public final class AltsChannelBuilder extends ForwardingChannelBuilder<AltsChannelBuilder> {
 
+  private static final Logger logger = Logger.getLogger(AltsChannelBuilder.class.getName());
   private final NettyChannelBuilder delegate;
   private final AltsClientOptions.Builder handshakerOptionsBuilder =
       new AltsClientOptions.Builder();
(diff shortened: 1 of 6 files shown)


alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
```diff
@@ -115,13 +120,25 @@ protected NettyChannelBuilder delegate() {
...
+    Status status = Status.OK;
+    if (!CheckGcpEnvironment.isOnGcp()) {
+      if (enableUntrustedAlts) {
+        logger.log(
+            Level.WARNING,
+            "Untrusted ALTS mode is enabled and we cannot guarantee the trustworthiness of the "
+                + "ALTS handshaker service");
+      } else {
+        status =
+            Status.FAILED_PRECONDITION.withDescription(
```
ejona86: The gRPC library can't use FAILED_PRECONDITION: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
```diff
@@ -115,13 +120,25 @@ protected NettyChannelBuilder delegate() {
...
+      }
+    }
+
     TcpfFactory tcpfFactory = new TcpfFactory();
     InternalNettyChannelBuilder.setDynamicTransportParamsFactory(delegate(), tcpfFactory);
-
     tcpfFactoryForTest = tcpfFactory;
 
-    return new AltsChannel(delegate().build());
+    return delegate().intercept(new AltsClientInterceptor(status)).build();
```
ejona86: Only add the interceptor when it will be helpful (like in the else case above).

You should add a note that you can't call `build()` multiple times.

alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
jiangtaoli2016: How about INTERNAL? I feel FAILED_PRECONDITION is the right description though.
 "Operation was rejected because the system is not in a state required for the operation's execution."

ejona86: INTERNAL is okay. We can change it later if need be.

alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
```diff
@@ -115,13 +120,25 @@ protected NettyChannelBuilder delegate() {
...
+      if (enableUntrustedAlts) {
+        logger.log(
+            Level.WARNING,
+            "Untrusted ALTS mode is enabled and we cannot guarantee the trustworthiness of the "
+                + "ALTS handshaker service");
+      } else {
+        status =
+            Status.INTERNAL.withDescription("ALTS is only allowed to run on Google Cloud Platform");
+      }
+      delegate().intercept(new AltsClientInterceptor(status));
```
ejona86: Move this into the else.

alts/src/test/java/io/grpc/alts/AltsChannelBuilderTest.java
```diff
@@ -50,7 +44,7 @@ public void buildsNettyChannel() throws Exception {
     assertThat(altsClientOptions).isNull();
 
     ManagedChannel channel = builder.build();
-    assertThat(channel).isInstanceOf(AltsChannel.class);
+    assertThat(channel).isInstanceOf(ManagedChannel.class);
```
ejona86: This check dose nothing, since it's a `ManagedChannel` that is returned.

jiangtaoli2016:
@ejona86 code revised. PTAL

alts/src/main/java/io/grpc/alts/GoogleDefaultChannelBuilder.java
```diff
@@ -91,9 +90,7 @@ public ManagedChannel build() {
       credentials = MoreCallCredentials.from(GoogleCredentials.getApplicationDefault());
     } catch (IOException e) {
       status =
-          Status.FAILED_PRECONDITION
-              .withDescription("Failed to get Google default credentials")
-              .withCause(e);
+          Status.INTERNAL.withDescription("Failed to get Google default credentials").withCause(e);
```
ejona86: Probably UNAUTHENTICATED

alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
```diff
@@ -189,69 +206,22 @@ public AltsProtocolNegotiator getProtocolNegotiator() {
...
+    public AltsClientInterceptor(Status status) {
+      this.status = status;
     }
 
     @Override
-    public void enterIdle() {
-      delegate.enterIdle();
+    public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+        MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+      if (!status.isOk()) {
```
ejona86: No more need for this check.

alts/src/main/java/io/grpc/alts/AltsServerBuilder.java
```diff
@@ -182,77 +200,27 @@ public TsiHandshaker newHandshaker() {
...
 
     @Override
-    public String toString() {
-      return MoreObjects.toStringHelper(this).add("delegate", delegate).toString();
+    public <ReqT, RespT> Listener<ReqT> interceptCall(
+        ServerCall<ReqT, RespT> serverCall,
+        Metadata metadata,
+        ServerCallHandler<ReqT, RespT> nextHandler) {
+      if (!status.isOk()) {
+        serverCall.close(status, new Metadata());
```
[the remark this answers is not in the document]
ejona86: This would be appropriate, although, isn't it impossible to actually get here? Won't the server be unable to accept new connections? And without connections it won't get any RPCs.

Server-side does have `start()` that can throw an `IOException`...

jiangtaoli2016:
@ejona86 Revised. The reason we want to check GCP platform is to prevent network attack on handshaker service. If attacker re-directs metadata server, it may fool grpc client.

In the normal case, if app is not running on GCP, it would fail on connecting to handshaker server.
In the case of being attack (by redirect to a fake handshaker server) and not running on GCP, grpc server or client would fail the call during interceptCall -- although connecting to attacker's handshaker server succeeded.

This GCP residency check is defense in depth.

PR #4812, opened 2018-08-29
core: DnsNameResolver caches refresh

### LINKED ISSUE #4745: DnsNameResolver should delay refresh if one has happened recently
RR and other LBs may cause `refresh()` to be called very frequently. There is a need to avoid constantly issuing DNS requests, so the cross-language design (can't find it right now; it's probably on a random C issue/PR) was to _delay_ (not ignore/squelch) queries if a query had recently been completed. This allows time to combine the refresh() with other refresh()es and reduce the total number of queries.

grpc-java today relies on Java caching DNS results. This works fine. However, the cache does not apply to queries performed with JNDI. As we begin querying SRV and TXT records, we will need to add the delay logic.

This could be implemented in ManagedChannelImpl.

This is a blocker for enabling gRPC-LB. Putting this in 1.15 milestone since it impacts automatic ALTS detection.

CC @zhangkun83, @carl-mastrangelo

creamsoup:
DnsNameResolver caches refresh using java system property networkaddress.cache.ttl.

Resolves #4745

### DIFF
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -130,7 +130,7 @@ final class DnsNameResolver extends NameResolver {
   private final String host;
   private final int port;
   private final Resource<ExecutorService> executorResource;
-  private final long networkAddressCacheTtlNanoSeconds;
+  private final long networkAddressCacheTtlNanos;
   private final Stopwatch stopwatch;
   @GuardedBy("this")
   private boolean shutdown;
@@ -168,7 +168,7 @@ final class DnsNameResolver extends NameResolver {
     }
     this.proxyDetector = proxyDetector;
     this.stopwatch = Preconditions.checkNotNull(stopwatch, "stopwatch");
-    this.networkAddressCacheTtlNanoSeconds = getNetworkAddressCacheTtlNanoSeconds();
+    this.networkAddressCacheTtlNanos = getNetworkAddressCacheTtlNanos();
   }
 
   @Override
@@ -199,9 +199,9 @@ final class DnsNameResolver extends NameResolver {
             return;
           }
           boolean resourceRefreshRequired = cachedResolutionResults == null
-              || networkAddressCacheTtlNanoSeconds == 0
-              || (networkAddressCacheTtlNanoSeconds > 0
-                  && stopwatch.elapsed(TimeUnit.NANOSECONDS) >= networkAddressCacheTtlNanoSeconds);
+              || networkAddressCacheTtlNanos == 0
+              || (networkAddressCacheTtlNanos > 0
+                  && stopwatch.elapsed(TimeUnit.NANOSECONDS) > networkAddressCacheTtlNanos);
           if (!resourceRefreshRequired) {
             return;
           }
@@ -235,7 +235,7 @@ final class DnsNameResolver extends NameResolver {
             resolutionResults =
                 resolveAll(addressResolver, resourceResolver, enableSrv, enableTxt, host);
             cachedResolutionResults = resolutionResults;
-            if (networkAddressCacheTtlNanoSeconds > 0) {
+            if (networkAddressCacheTtlNanos > 0) {
               stopwatch.reset().start();
             }
           } catch (Exception e) {
@@ -285,7 +285,7 @@ final class DnsNameResolver extends NameResolver {
     };
 
   /** Returns value of network address cache ttl property. */
-  private static long getNetworkAddressCacheTtlNanoSeconds() {
+  private static long getNetworkAddressCacheTtlNanos() {
     String cacheTtlPropertyValue = System.getProperty(NETWORKADDRESS_CACHE_TTL_PROPERTY);
     long cacheTtl = DEFAULT_NETWORK_CACHE_TTL_SECONDS;
     if (cacheTtlPropertyValue != null) {
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -21,7 +21,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Stopwatch;
-import com.google.common.base.Ticker;
 import com.google.common.base.Verify;
 import io.grpc.Attributes;
 import io.grpc.EquivalentAddressGroup;
@@ -34,7 +33,6 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.URI;
 import java.net.UnknownHostException;
-import java.security.Security;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -200,17 +217,33 @@ public void run() {
           }
 
           ResolutionResults resolutionResults;
-          try {
-            ResourceResolver resourceResolver = null;
-            if (enableJndi) {
-              resourceResolver = getResourceResolver();
+          long networkAddressCacheTtl = getNetworkAddressCacheTtl();
```
ejona86: Let's just do this once for the class. We're not concerned about the value changing over time (because Java won't respect the change anyway).

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -200,17 +217,33 @@ public void run() {
...
+              return;
             }
-            resolutionResults =
-                resolveAll(addressResolver, resourceResolver, enableSrv, enableTxt, host);
-          } catch (Exception e) {
-            savedListener.onError(
-                Status.UNAVAILABLE.withDescription("Unable to resolve host " + host).withCause(e));
-            return;
+          } else {
+            resolutionResults = cachedResolutionResults;
```
ejona86: It seems like in this case we should just return now. Since the values we pass to `onAddresses()` will be the same, there's no need to call it.

core/src/test/java/io/grpc/internal/DnsNameResolverTest.java
```diff
@@ -177,6 +180,8 @@ public void invalidDnsName_containsUnderscore() {
 
   @Test
   public void resolve() throws Exception {
+    // disable network address cache
+    Security.setProperty(DnsNameResolver.NETWORKADDRESS_CACHE_TTL_SECURITY_PROPERTY, "-1");
```
ejona86: Any global changes should be restored by the end of the test. Either a try-catch or a `@Before` to save the value before the test runs and a `@After` to restore it.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -88,6 +92,17 @@
...
+  /**
+   * Java networking security properties name for caching DNS result.
+   *
+   * <p>Default value is -1 (cache forever) if security manager is installed. If security manager is
+   * not installed, the ttl value is {@code null} which falls back to {@link
+   * #DEFAULT_NETWORK_CACHE_TTL gRPC default value}.
+   */
+  static final String NETWORKADDRESS_CACHE_TTL_SECURITY_PROPERTY = "networkaddress.cache.ttl";
+  /** Default DNS cache duration if network cache ttl value is not specified ({@code null}). */
+  private static final long DEFAULT_NETWORK_CACHE_TTL = 5 * 60;
```
carl-mastrangelo: TimeUnit.MINUTES.toSeconds(5);

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -123,6 +138,8 @@
   private boolean resolving;
   @GuardedBy("this")
   private Listener listener;
+  private ResolutionResults cachedResolutionResults;
+  private Stopwatch stopwatch = Stopwatch.createUnstarted();
```
carl-mastrangelo: I think this needs to be volatile

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -200,17 +217,33 @@ public void run() {
           }
 
           ResolutionResults resolutionResults;
-          try {
-            ResourceResolver resourceResolver = null;
-            if (enableJndi) {
-              resourceResolver = getResourceResolver();
+          long networkAddressCacheTtl = getNetworkAddressCacheTtl();
+          boolean refreshRequired = networkAddressCacheTtl <= 0
```
carl-mastrangelo: This should be named resourceRefreshRequired, to distinguish it from the address results needing refresh.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -253,6 +286,27 @@ public void run() {
       }
     };
 
+  /** Returns value of security property of network address cache ttl. */
+  private long getNetworkAddressCacheTtl() {
+    String cacheTtlPropertyValue = Security.getProperty(NETWORKADDRESS_CACHE_TTL_SECURITY_PROPERTY);
+    long cacheTtl = DEFAULT_NETWORK_CACHE_TTL;
```
carl-mastrangelo: DEFAULT_NETWORK_CACHE_TTL_SECONDS

carl-mastrangelo: Optional: but I strong recommend converting everything to nanoseconds on API boundaries.  It makes the code base consistent, and avoids converting in the internal code paths.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -253,6 +286,27 @@ public void run() {
...
+    if (cacheTtlPropertyValue != null) {
+      try {
+        cacheTtl = Long.parseLong(cacheTtlPropertyValue);
+      } catch (NumberFormatException e) {
+        // do nothing (use default value).
```
carl-mastrangelo: Always log, no exception should be lost.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -253,6 +286,27 @@ public void run() {
       }
     };
 
+  /** Returns value of security property of network address cache ttl. */
+  private long getNetworkAddressCacheTtl() {
```
carl-mastrangelo: this should be static

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -253,6 +286,27 @@ public void run() {
...
+    String cacheTtlPropertyValue = Security.getProperty(NETWORKADDRESS_CACHE_TTL_SECURITY_PROPERTY);
```
carl-mastrangelo: I'm a little uncomfortable parsing this each time.  Why not parse it once at class init?

core/src/test/java/io/grpc/internal/DnsNameResolverTest.java
```diff
@@ -203,6 +208,76 @@ public void resolve() throws Exception {
     verify(mockResolver, times(2)).resolveAddress(Matchers.anyString());
   }
 
+  @Test
```
carl-mastrangelo: No test for an invalid system property?

core/src/main/java/io/grpc/internal/DnsNameResolver.java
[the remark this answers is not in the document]
creamsoup: it was some arbitrary value.  
https://docs.oracle.com/javase/7/docs/technotes/guides/net/properties.html says default is -1.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -166,6 +167,8 @@
       port = nameUri.getPort();
     }
     this.proxyDetector = proxyDetector;
+    this.stopwatch = Preconditions.checkNotNull(stopwatch, "stopwatch");
+    this.networkAddressCacheTtlNanoSeconds = getNetworkAddressCacheTtlNanoSeconds();
```
carl-mastrangelo: optional: we typically call these 'nanos'

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -195,6 +198,13 @@ public void run() {
           if (shutdown) {
             return;
           }
+          boolean resourceRefreshRequired = cachedResolutionResults == null
+              || networkAddressCacheTtlNanoSeconds == 0
+              || (networkAddressCacheTtlNanoSeconds > 0
+                  && stopwatch.elapsed(TimeUnit.NANOSECONDS) >= networkAddressCacheTtlNanoSeconds);
```
carl-mastrangelo: micronit: I think this should be `>` rather than `>=`

Issue #4809, opened 2018-08-28
OkHttp: INTERNAL: Flow control error Rst Stream


ejona86:
Our android-interop test failed: https://source.cloud.google.com/results/invocations/ed2d6385-7608-437e-84ed-f8e34cdb35f7/log

```
junit.framework.ComparisonFailure: very_large_request failed expected:<[Failed... : INTERNAL: Flow control error
Rst Stream
io.grpc.StatusRuntimeException: INTERNAL: Flow control error
Rst Stream
  at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:233)
  at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:214)
  at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:139)
  at io.grpc.testing.integration.TestServiceGrpc$TestServiceBlockingStub.unaryCall(TestServiceGrpc.java:624)
  at io.grpc.testing.integration.AbstractInteropTest.veryLargeRequest(AbstractInteropTest.java:959)
  at io.grpc.android.integrationtest.InteropTask.runTest(InteropTask.java:113)
  at io.grpc.android.integrationtest.InteropTask.doInBackground(InteropTask.java:62)
  at io.grpc.android.integrationtest.InteropTask.doInBackground(InteropTask.java:31)
  at android.os.AsyncTask$2.call(AsyncTask.java:305)
  at java.util.concurrent.FutureTask.run(FutureTask.java:266)
  at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:243)
  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
  at java.lang.Thread.run(Thread.java:764)
]> but was:<[Success!]>
  at junit.framework.Assert.assertEquals(Assert.java:85)
  at io.grpc.android.integrationtest.InteropInstrumentationTest.runTest(InteropInstrumentationTest.java:134)
  at io.grpc.android.integrationtest.InteropInstrumentationTest.interopTests(InteropInstrumentationTest.java:99)
  at java.lang.reflect.Method.invoke(Native Method)
  at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
[...]

PR #4816, opened 2018-08-30
interop-testing: add very_large_request case to test client

ericgribkoff:
This was shown by https://github.com/grpc/grpc-java/issues/4809 to be a useful test. This change makes it easier to reproduce issues by running the test case manually with the interop client.


creamsoup:
missing a value ("very_large_request") in interop-testing/src/test/java/io/grpc/testing/integration/TestCasesTest.java ?

ericgribkoff:
Thanks, updated the test case test.

PR #4813, opened 2018-08-30
examples: Add google mirrored maven central to examples pom.xml to deflake kokoro.

creamsoup:
error message:

cmd: mvn verify
Downloading from central: https://repo.maven.apache.org/maven2/kr/motd/maven/os-maven-plugin/1.5.0.Final/os-maven-plugin-1.5.0.Final.pom
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] Unresolveable build extension: Plugin kr.motd.maven:os-maven-plugin:1.5.0.Final or one of its dependencies could not be resolved: Failed to read artifact descriptor for kr.motd.maven:os-maven-plugin:jar:1.5.0.Final @ 
 @ 
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project io.grpc:examples:1.16.0-SNAPSHOT (/home/travis/build/grpc/grpc-java/examples/pom.xml) has 1 error
[ERROR]     Unresolveable build extension: Plugin kr.motd.maven:os-maven-plugin:1.5.0.Final or one of its dependencies could not be resolved: Failed to read artifact descriptor for kr.motd.maven:os-maven-plugin:jar:1.5.0.Final: Could not transfer artifact kr.motd.maven:os-maven-plugin:pom:1.5.0.Final from/to central (https://repo.maven.apache.org/maven2): Access denied to: https://repo.maven.apache.org/maven2/kr/motd/maven/os-maven-plugin/1.5.0.Final/os-maven-plugin-1.5.0.Final.pom , ReasonPhrase:Forbidden. -> [Help 2]

PR #4820, opened 2018-09-04
move google maven central mirror to settings.xml in travis

creamsoup:


.travis.settings.xml
```diff
@@ -0,0 +1,10 @@
+<settings>
```
ejona86: Go ahead and put this in buildscripts/ or similar. I'd rather not pollute the root folder.

.travis.yml
```diff
@@ -10,7 +10,7 @@ env:
     - LD_LIBRARY_PATH=/tmp/protobuf/lib
 
 before_install:
-  - rm ~/.m2/settings.xml || true # Avoid repository.apache.org, which has QPS limits and is useless
+  - cp .travis.settings.xml ~/.m2/settings.xml
```
ejona86: This still deserves some comments. It is important, for example, that this overwrites the default config which references repository.apache.org.

.travis.settings.xml
```diff
@@ -0,0 +1,10 @@
...
+  <mirrors>
+    <mirror>
+      <id>google-maven-central</id>
+      <name>Google Maven Central</name>
+      <url>https://maven-central.storage.googleapis.com/repos/central/data/</url>
```
ejona86: We use this mirror no Kororo because we're on GCE with Kokoro. The container-based Trusty is [currently on EC2 or Packet.net](https://docs.travis-ci.com/user/reference/overview/), so it isn't as good of an option. So I'd suggest we [copy the Travis configuration](https://github.com/grpc/grpc-java/pull/4813#issuecomment-418514808) but only with central + sonatype repos. If we want, we can have the second or third be the Google mirror.  But I'd want to keep the default/first mirror still repo.maven.apache.org.

creamsoup:
we suspect that the downloading error in maven central issue is temporary issue.
close for now.

PR #4825, opened 2018-09-05
okhttp: settings acks back after apply settings before sending any data

### LINKED ISSUE #4809: OkHttp: INTERNAL: Flow control error Rst Stream

creamsoup:
settings acks back after apply settings before sending any data as a result of the change.
Resolves #4809 also, make #4816 the not flaky.

### DIFF
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
@@ -1068,12 +1068,6 @@ class OkHttpClientTransport implements ConnectionClientTransport, TransportExcep
           stopDelivery, null, null);
     }
 
-    /**
-     * Handles inbound {@code SETTINGS} frame. The changed settings are not finalized until {@code
-     * SETTINGS} acknowledgment frame is sent. Any writes due to update in settings must be sent
-     * after {@code SETTINGS} acknowledgment frame, otherwise it will cause a stream error ({@code
-     * RST_STREAM}).
-     */
     @Override
     public void settings(boolean clearPrevious, Settings settings) {
       boolean outboundWindowSizeIncreased = false;
@@ -1094,9 +1088,12 @@ class OkHttpClientTransport implements ConnectionClientTransport, TransportExcep
           firstSettings = false;
         }
 
+        // The changed settings are not finalized until SETTINGS acknowledgment frame is sent. Any
+        // writes due to update in settings must be sent after SETTINGS acknowledgment frame,
+        // otherwise it will cause a stream error (RST_STREAM).
         frameWriter.ackSettings(settings);
 
-        // any pending bytes / streams frames must sent after ack settings.
+        // send any pending bytes / streams
         if (outboundWindowSizeIncreased) {
           outboundFlow.writeStreams();
         }
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
@@ -1070,6 +1070,7 @@ class OkHttpClientTransport implements ConnectionClientTransport, TransportExcep
 
     @Override
     public void settings(boolean clearPrevious, Settings settings) {
+      boolean outboundWindowSizeIncreased = false;
       synchronized (lock) {
         if (OkHttpSettingsUtil.isSet(settings, OkHttpSettingsUtil.MAX_CONCURRENT_STREAMS)) {
           int receivedMaxConcurrentStreams = OkHttpSettingsUtil.get(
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -1080,16 +1081,23 @@ public void settings(boolean clearPrevious, Settings settings) {
...
         if (firstSettings) {
           listener.transportReady();
           firstSettings = false;
         }
-        startPendingStreams();
       }
 
       frameWriter.ackSettings(settings);
+
+      // Write pending bytes / streams after ack settings.
```
ejona86: I think there needs to be a stronger comment somewhere saying that we _must_ not write any frames until after the `ackSettings()`.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -1080,16 +1081,23 @@ public void settings(boolean clearPrevious, Settings settings) {
         if (OkHttpSettingsUtil.isSet(settings, OkHttpSettingsUtil.INITIAL_WINDOW_SIZE)) {
           int initialWindowSize = OkHttpSettingsUtil.get(
               settings, OkHttpSettingsUtil.INITIAL_WINDOW_SIZE);
-          outboundFlow.initialOutboundWindowSize(initialWindowSize);
+          outboundWindowSizeIncreased = outboundFlow.initialOutboundWindowSize(initialWindowSize);
```
ejona86: Hmm... We're updating the window size here. It seems like it would be possible [for a stream to write more DATA](https://github.com/grpc/grpc-java/blob/v1.14.0/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java#L373) between when this lock is released and the `ackSettings()`.

It looks like [elsewhere](https://github.com/grpc/grpc-java/blob/v1.14.0/okhttp/src/main/java/io/grpc/okhttp/OutboundFlowController.java#L122) we're calling `frameWriter` while the lock is held. It seems we should probably do the same here.

okhttp/src/main/java/io/grpc/okhttp/OutboundFlowController.java
```diff
@@ -46,9 +46,15 @@
   }
 
   /**
-   * Must be called with holding transport lock.
+   * Adjusts outbound window size requested by peer. When window size is increased, it does not send
+   * any pending frames. Caller should send any pending frames after notify the peer about
```
ejona86: s/Caller should send any pending frames/If this method returns `true`, the caller should call `writeStreams()`/

okhttp/src/main/java/io/grpc/okhttp/OutboundFlowController.java
```diff
@@ -46,9 +46,15 @@
...
+   * successful window size update.
```
ejona86: "window size update" here makes it seem like WINDOW_UPDATEs are involved. Since this method is only used for SETTINGS, let's just say "SETTINGS ACK" here.

PR #4830, opened 2018-09-06
build.gradle: remove intellij workaround for gradle/issues/2315

zpencer:
This is no longer needed on recent versions of intellij.

Thank you @elharo for pointing this out.


zpencer:
I see there's already an outstanding PR that I missed

PR #4835, opened 2018-09-07
alts: Initialize ProtocolNegotiators eagerly

ejona86:
This simplifies the construction process, as we have fewer "interesting"
phases to weed through.

----

~~The first commit is from #4834, which this PR depends on since TLS needs to
be eagerly initialized as well. I'll wait until #4834 is approved and then manually
rebase+merge.~~ Now rebased on master which includes that PR.

CC @jiangtaoli2016


ejona86:
@carl-mastrangelo, friendly ping

PR #4839, opened 2018-09-07
core: DnsNameResolver caches refresh (1.15.x backport)

### LINKED ISSUE #4745: DnsNameResolver should delay refresh if one has happened recently

ejona86:
DnsNameResolver caches refresh using java security property networkaddress.cache.ttl.

Resolves #4745 

-----

This is a backport of #4812

This will wait until the the v1.15.0 build is made, for inclusion in v1.15.1. We'd like it to "sit" longer before being in a release to avoid reducing stability.

### DIFF
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -20,6 +20,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Stopwatch;
 import com.google.common.base.Verify;
 import io.grpc.Attributes;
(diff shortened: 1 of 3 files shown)


ejona86:
CC @zhangkun83

PR #4841, opened 2018-09-08
Enable TransportTest.flowControlPushBack and fix a flake

ejona86:
This attempts to fix a flake seen exactly once with the
currently-disabled OkHttpTransportTest.flowControlPushBack:
```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at io.grpc.internal.testing.AbstractTransportTest.flowControlPushBack(AbstractTransportTest.java:1300)
```

That was a failure for assertTrue(serverStream.isReady()), because the
awaitOnReady was finding the previous invocation of onReady. We now
track how many times it has been called. This was a bug introduced in
a8db154 but wouldn't have been noticed since the in-process transport
is deterministic.

----

It appears to be stable now. Ran for 1000s of times. I do see some
general flakiness in TransportTest, but it applies to the tests in
general and isn't specific to this one test. It is:

```
org.mockito.exceptions.verification.WantedButNotInvoked:
Wanted but not invoked:
listener.transportReady();
-> at io.grpc.internal.testing.AbstractTransportTest.startTransport(AbstractTransportTest.java:1815)
Actually, there were zero interactions with this mock.
```

This flake is not seen often because it occurs less frequently when
running all the tests (~.1% vs 1%). One of the early tests must warm
something up to make it less likely.

### DIFF
--- a/testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
+++ b/testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
@@ -709,7 +709,7 @@ public abstract class AbstractTransportTest {
     assertNotNull(serverStream.getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR));
 
     serverStream.request(1);
-    assertTrue(clientStreamListener.awaitOnReady(TIMEOUT_MS, TimeUnit.MILLISECONDS));
+    assertTrue(clientStreamListener.awaitOnReadyAndDrain(TIMEOUT_MS, TimeUnit.MILLISECONDS));
     assertTrue(clientStream.isReady());
     clientStream.writeMessage(methodDescriptor.streamRequest("Hello!"));
     assertThat(clientStreamTracer1.nextOutboundEvent()).isEqualTo("outboundMessage(0)");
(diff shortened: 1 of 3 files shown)


ejona86:
I have a fix for the build, but while working on it I got one flake that was new and real! It's amazing the other runs didn't trigger it.

ejona86:
@dapengzhang0, PTAL. The two commits will be remain separate.

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
```diff
@@ -1293,10 +1293,9 @@ public void flowControlPushBack() throws Exception {
     serverReceived +=
         verifyMessageCountAndClose(serverStreamListener.messageQueue, clientSent - serverReceived);
 
-    assertTrue(clientStreamListener.awaitOnReady(TIMEOUT_MS, TimeUnit.MILLISECONDS));
-    assertTrue(clientStreamListener.awaitOnReady(TIMEOUT_MS, TimeUnit.MILLISECONDS));
```
dapengzhang0: I don't understand the previous code, was it duplicate lines and intended to be only one line?

ejona86: If you look back in a8db154 it was originally a mock verification with times(2). However, when it was changed it became a noop, since onReady was already previously called, but it _looked_ like it was checking twice.

Today I'm seeing onReady only called once, but the behavior isn't actually that precise since there aren't any other calls to notice if `isReady() == false`. So now I just require it to be called once, and if it happens twice that's "okay" as long it isn't racy, which could be detected via the isReady() call on the next line.

ejona86:
The Travis-CI status check is hosed. Rebuilding travis didn't fix it. I'm going to ignore the status check because travis shows green.

PR #4846, opened 2018-09-11
core,services: v1 binlog

zpencer:
- Log using new proto definition, this is effectively a rewrite
- Remove io.grpc.BinaryLog.CallId because a call ID is now an AtomicLong
- Add the concept of "always included" and "never included" metadata
  keys. This is needed because grpc-status-details-bin is already
  logged in the binlog msg, and we will log grpc-trace-bin for the
  census info.


services/src/main/java/io/grpc/services/BinaryLogProviderImpl.java
```diff
@@ -28,9 +28,11 @@
  * The default implementation of a {@link BinaryLogProvider}.
  */
 class BinaryLogProviderImpl extends BinaryLogProvider {
+  // avoid using 0 because proto3 long fields default to 0 when unset
+  private static final AtomicLong counter = new AtomicLong(1);
```
ejona86: Why did this become `static`?

services/src/main/java/io/grpc/services/BinlogHelper.java
```diff
@@ -109,134 +101,193 @@
...
       this.sink = sink;
+      this.timeProvider = timeProvider;
       this.maxHeaderBytes = maxHeaderBytes;
       this.maxMessageBytes = maxMessageBytes;
     }
 
+    GrpcLogEntry.Builder newTimestampedBuilder() {
+      long epochNanos = timeProvider.currentTimeNanos();
+      return GrpcLogEntry.newBuilder()
+          .setTimestamp(Timestamp.newBuilder()
```
ejona86: You didn't/couldn't use `Timestamps.fromNanos()` from protobuf?

services/src/main/java/io/grpc/services/BinlogHelper.java
```diff
@@ -332,56 +387,86 @@ private static Deadline min(@Nullable Deadline deadline0, @Nullable Deadline dea
...
   }
 
-  public ClientInterceptor getClientInterceptor(final CallId callId) {
+  interface TimeProvider {
+    /** Returns the current nano time. */
+    long currentTimeNanos();
+
+    TimeProvider SYSTEM_TIME_PROVIDER = new TimeProvider() {
+      final long offsetNanos =
+          TimeUnit.MILLISECONDS.toNanos(System.currentTimeMillis()) - System.nanoTime();
```
ejona86: If the clock changes, then the timestamps will be incorrect. Is that what is supposed to happen?

services/src/main/java/io/grpc/services/BinlogHelper.java
```diff
@@ -109,134 +101,193 @@
...
-      GrpcLogEntry.Builder entryBuilder = GrpcLogEntry.newBuilder()
-          .setSequenceIdWithinCall(seq)
-          .setType(Type.SEND_INITIAL_METADATA)
-          .setLogger(isServer ? GrpcLogEntry.Logger.SERVER : GrpcLogEntry.Logger.CLIENT)
-          .setCallId(callIdToProto(callId));
-      addMetadataToProto(entryBuilder, metadata, maxHeaderBytes);
-      if (methodName != null) {
-        entryBuilder.setMethodName("/" + methodName);
-      }
+      Preconditions.checkArgument(!methodName.startsWith("/"));
```
ejona86: Include an explanation string?

services/src/main/proto/grpc/binlog/v1/binarylog.proto
```diff
@@ -0,0 +1,177 @@
+// Copyright 2018 The gRPC Authors
```
ejona86: This file is different from the one in grpc-proto. Why is that?

PR #4855, opened 2018-09-13
netty: Allow specifying a ProtocolNegotatior directly to Channels

ejona86:
This will be the replacement for TransportCreationParamsFilterFactory
and matches what used to be done and what is done on server-side.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
+++ b/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
@@ -81,7 +81,7 @@ public final class NettyChannelBuilder
   private long keepAliveTimeoutNanos = DEFAULT_KEEPALIVE_TIMEOUT_NANOS;
   private boolean keepAliveWithoutCalls;
   private TransportCreationParamsFilterFactory dynamicParamsFactory;
-  private ProtocolNegotiator protocolNegotiator;
+  private ProtocolNegotiatorFactory protocolNegotiatorFactory;
 
   /**
    * Creates a new builder with the given server address. This factory method is primarily intended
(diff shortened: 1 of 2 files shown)


netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -334,16 +335,18 @@ protected ClientTransportFactory buildTransportFactory() {
...
+      if (negotiator == null) {
+        SslContext localSslContext = sslContext;
+        if (negotiationType == NegotiationType.TLS && localSslContext == null) {
+          try {
+            localSslContext = GrpcSslContexts.forClient().build();
+          } catch (SSLException ex) {
+            throw new RuntimeException(ex);
+          }
         }
+        negotiator = createProtocolNegotiatorByType(negotiationType, localSslContext);
```
carl-mastrangelo: If `this.protocolNegotiator` is not null won't this overwrite `negotiator` ?

ejona86: It can't be non-null, as this is run only `if (negotiator == null)`.

carl-mastrangelo: oh, one indentation level out.  my mystake.

carl-mastrangelo: One other thing;  This is a little wierd.  If no PN is set, each build channel will get a unique PN.  If one has been set, each built channel will get the exact same one.  That's a slightly surprising behavior change.

ejona86: Hmm... true, and that would apply to ALTS as well. Do you think the builder should allow setting a factory for the PN?

carl-mastrangelo: The thought had occurred; it does make things more complicated since construction would be deferred.   What would you expect?   A new one each time, or it to be reused?

ejona86: So most of the time it looks like this will be per-Channel or could be made per-Channel easily. In order to not complicate things much, when I add the close/shutdown I'll also add a method called when a new Channel has been built. So I guess that becomes a ref/unref sort of style. I expect most implementations to create a new instance per-Channel and just ignore the ref method.

ejona86:
@carl-mastrangelo, PTAL, swapped to factory. I also updated the ALTS PR to match.

PR #4856, opened 2018-09-13
alts: Remove usage of TransportCreationParamsFilterFactory

ejona86:
Provide a ProtocolNegotiator directly instead.
TransportCreationParamsFilterFactory is being removed.

-----

~The first two commits are just dependencies. This depends on #4835 and #4855~ Those are merged. Rebased.

CC @jiangtaoli2016

### DIFF
--- a/alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
+++ b/alts/src/main/java/io/grpc/alts/AltsChannelBuilder.java
@@ -16,8 +16,6 @@
 
 package io.grpc.alts;
 
-import static com.google.common.base.Preconditions.checkArgument;
-
 import com.google.common.annotations.VisibleForTesting;
 import io.grpc.CallOptions;
 import io.grpc.Channel;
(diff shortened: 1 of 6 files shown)


netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -334,16 +335,18 @@ protected ClientTransportFactory buildTransportFactory() {
...
+      if (negotiator == null) {
+        SslContext localSslContext = sslContext;
+        if (negotiationType == NegotiationType.TLS && localSslContext == null) {
+          try {
+            localSslContext = GrpcSslContexts.forClient().build();
+          } catch (SSLException ex) {
+            throw new RuntimeException(ex);
+          }
         }
+        negotiator = createProtocolNegotiatorByType(negotiationType, localSslContext);
```
carl-mastrangelo: I would feel a lot more comfortable with this if the lifetime of protocol negotiators was documented.  Up until now, it's been ambiguous if there is one per channel, channel builder, transport, etc.   I'm not so keen on which one it is, as long as it's easily understandable.

I'm uncomfortable with creating a new PN each time, except if it has been previously set by outside.  I would have expected `this.protocolNegotiator` to be set here, and cleared out by changing the SSL settings.  The reason is that it's unclear what will happen if SSL related settings are changed on the builder, *after* `protocolNegotiator()` is called.

ejona86:
@carl-mastrangelo, PTAL. Rebased since the dependencies are in. The second commit "Swap to factory" are the only actual changes since last review.

PR #4862, opened 2018-09-19
github/lock: Override some of the defaults before enabling

ejona86:
These are settings for the Lock app: https://github.com/apps/lock

We'll probably want to reduce the daysUntilLock to something lower, but
want to try it out first and a larger value will impact fewer issues.
365 is the default, but making it clear what setting we are using.

I disable the comment because I don't want notification noise. I disable
the resolved feature because I'm not quite sure what it is and I'm not
really looking for anything fancy.

The defaults can be found at
https://github.com/dessant/lock-threads/blob/master/src/schema.js

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,0 +1,3 @@
+daysUntilLock: 365
+lockComment: false
+setLockReason: false


ejona86:
After this is merged I'll enable the App and see how it goes.

nicolasnoble:
Oh, by the way... the "set lock reason: resolved", this is a typical github feature. When you hit the "Lock conversation" button, you get the following screen:

image

PR #4874, opened 2018-09-25
netty: Remove TransportCreationParamsFilterFactory and add ProtocolNegotiator.close

ejona86:
There's two commits here that are easiest to review separately. Removing the FilterFactory was necessary to have a clear lifecycle of the ProtocolNegotiator.

@jiangtaoli2016, after this is merged you can plumb cleanup login in ALTS.

After this is approved but before merging I'll add the necessary close() internally, to make the sync easy.


jiangtaoli2016:
Thank you @ejona86 for the plumbing!

ejona86:
Rebased past 8e72351a , which split the ALTS protocol negotiator into two.

PR #4883, opened 2018-09-26
core: TimeProvider should not assume that the clock never changes

zpencer:
We should reflect changes in the system clock.

### DIFF
--- a/core/src/main/java/io/grpc/internal/TimeProvider.java
+++ b/core/src/main/java/io/grpc/internal/TimeProvider.java
@@ -27,12 +27,9 @@ public interface TimeProvider {
   long currentTimeNanos();
 
   TimeProvider SYSTEM_TIME_PROVIDER = new TimeProvider() {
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/grpc/internal/TimeProvider.java
```diff
@@ -27,12 +27,9 @@
...
 
   TimeProvider SYSTEM_TIME_PROVIDER = new TimeProvider() {
-    final long offsetNanos =
-        TimeUnit.MILLISECONDS.toNanos(System.currentTimeMillis()) - System.nanoTime();
-
     @Override
     public long currentTimeNanos() {
-      return System.nanoTime() + offsetNanos;
+      return TimeUnit.MILLISECONDS.toNanos(System.currentTimeMillis());
     }
```
dapengzhang0: After your change, it will return the same nano value for any invocation that happens within one millisecond.

PR #4886, opened 2018-09-27
core: permanently store authority at channel creation

zpencer:
Getting the authority must not rely on the name resolver being
non-null, because that can trivially happen if the channel is shut
down.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -159,6 +159,10 @@ final class ManagedChannelImpl extends ManagedChannel implements
   // Only null after channel is terminated. Must be assigned from the channelExecutor.
   private NameResolver nameResolver;
 
+  // Set when the NameResolver is initially created. When we create a new NameResolver for the same
+  // target, the new instance must have the same value.
+  private final String cachedAuthority;
+
   // Must be accessed from the channelExecutor.
   private boolean nameResolverStarted;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -798,7 +803,7 @@ public boolean isTerminated() {
 
   @Override
   public String authority() {
-    return interceptorChannel.authority();
+    return checkNotNull(cachedAuthority, "cachedAuthority");
```
ejona86: Why was this changed?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -828,8 +833,7 @@ private Executor getCallExecutor(CallOptions callOptions) {
 
     @Override
     public String authority() {
-      String authority = nameResolver.getServiceAuthority();
-      return checkNotNull(authority, "authority");
+      return checkNotNull(cachedAuthority, "cachedAuthority");
```
ejona86: Remove the checkNotNull.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -2801,6 +2801,14 @@ protected ClientTransportFactory buildTransportFactory() {
     mychannel.shutdownNow();
   }
 
+  @Test
+  public void getAuthorityAfterShutdown() {
+    createChannel();
+    assertEquals(SERVICE_NAME, channel.authority());
+    channel.shutdownNow();
```
ejona86: Also do awaitTerminated().

PR #4888, opened 2018-09-28
POC: bounded Attribute keys

zhangkun83:
Introduced `AttributeMap` to replace `Attributes`. The Key has an additional type parameter, which `AttributeMap` has too, so that each Key is bound with a certain `AttributeMap` passed on certain interfaces. This makes attribute keys more manageable and prevent misusing.

`ClientCall.getAttributes()` will be replaced by `ClientCall.getTransportAttrs()`. If there is a need for passing per-Call attributes, we can easily add `ClientCall.getCallAttrs()`.

`Attributes` is changed to be compatible with `AttributeMap.Key`. `Attributes` and `AttributeMap` can be converted back and forth. This allows non-breaking migration. Once `AttributeMap` is used and `Attributes` have been deprecated on all APIs, we can mark `Attributes` as deprecated.

### DIFF
--- a/alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
+++ b/alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
@@ -19,6 +19,7 @@ package io.grpc.alts.internal;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.protobuf.Any;
+import io.grpc.AttributeMap;
 import io.grpc.Attributes;
 import io.grpc.CallCredentials;
 import io.grpc.Grpc;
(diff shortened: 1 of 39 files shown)


zhangkun83:
Forgot to mention: I only changed `ClientCall.getAttributes()` so far. I will update the rest of the interfaces if the general idea is approved.

ejona86:
I'm not wild how verbose this becomes for the user, especially for Keys. We could sort of improve that by defining `TransportAttributeMap extends AttributeMap<Grpc.TransportAttr>` and doing the same for the key. Although Key construction would need boilerplate for each type and now people can extend the classes...

Also, the way things are going a key present in EquivalentAddressGroup could be plumbed all the way through to ClientCall.getAttributes(). That is to say that we may not have too many different types of Attributes.

We already decided that NameResolver.newNameResolver(URI, Attributes) should swap to a custom type. So what other "types" of Attributes do we have?

zhangkun83:
The other type of attributes is the one passed to `NameResolver.Listener.onAddresses()`. After offline discussion with @ejona86 , I decided to drop this because we are not sure if it will really meet our needs in the future, and once we introduced this type system and it doesn't work, it will be very hard to change it.

I will look into less intrusive options, either organizing the attribute keys in interface holders, or use annotations on the keys.

PR #4894, opened 2018-10-01
 core: make security level and authority first-class in CallCredentials

zhangkun83:
Security level and authority are parameters required to be passed to `applyRequestMetadata()`.  This change wraps them, along with `MethodDescriptor` and the transport attributes to `RequestInfo`, which is
more clear to the implementers.

`ATTR_SECURITY_LEVEL` is moved to the internal `GrpcAttributes` and annotated as `TransportAttr`, because transports are required to set it, but no user is actually reading them from `{Client,Server}Call.getAttributes()`.

`ATTR_AUTHORITY` is removed, because no transport is overriding it.

Note this breaks `CallCredentials` API, which is experimental. All involved interfaces are changed to abstract classes, as this will make further API changes smoother.

### DIFF
--- a/alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
+++ b/alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
@@ -41,18 +41,12 @@ import io.netty.util.AsciiString;
  */
 public abstract class AltsProtocolNegotiator implements ProtocolNegotiator {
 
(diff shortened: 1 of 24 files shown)


zhangkun83:
Bummer, [FirestoreCallCredentials](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreCallCredentials.java) is opensourced.

zhangkun83:
We are going to proceed with a multi-step introduction of the new API outlined by #4901 to allow for graceful migration of implementations. Closing.

Issue #4567, opened 2018-06-15
OkHttp does not have MAX_HEADER_LIST_SIZE


ejona86:
It seems it isn't imposing any limit now; I don't see any reference to DEFAULT_MAX_HEADER_LIST_SIZE within it. But it could also be hard-coded to something.

That makes it a bit weird to add, as it could impact existing users, but the limit is also there to protect the client and help encourage limited metadata for the benefit of proxies. This is also a behavior difference compared to the other implementations.

Related: #4050

CC @ericgribkoff

ejona86:
To add this would require changing `okhttp3.internal.http2.Hpack`. It'd be as easy as replacing `headerList.add(` with a call to some helper that counts the cumulative size so far. And then something in `getAndResetHeaderList()` that returns null or something when the size was exceeded. But it's unclear to me whether we actually want to modify that code. We've tried to have our changes to okhttp be minimal.

This is what I had (HEAD at 146b6006b3) before I decided to stop.
```diff
diff --git a/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java b/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
index 6a3c0881e..f505c9bb9 100644
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
@@ -153,6 +153,8 @@ class OkHttpClientTransport implements ConnectionClientTransport {
   // Wrap on executor, to guarantee some operations be executed serially.
   private final SerializingExecutor serializingExecutor;
   private final int maxMessageSize;
+  /** Only enforced when positive. */
+  private final int maxHeaderListSize;
   private int connectionUnacknowledgedBytesRead;
   private ClientFrameHandler clientFrameHandler;
   // Caution: Not synchronized, new value can only be safely read after the connection is complete.
@@ -516,6 +518,10 @@ class OkHttpClientTransport implements ConnectionClientTransport {
           // and before any possible pending stream operations.
           rawFrameWriter.connectionPreface();
           Settings settings = new Settings();
+          if (maxHeaderListSize > 0) {
+            OkHttpSettingsUtil.set(
+                settings, OkHttpSettingsUtil.MAX_HEADER_LIST_SIZE, maxHeaderListSize);
+          }
           rawFrameWriter.settings(settings);
         } catch (Exception e) {
           onException(e);
[...]

PR #4897, opened 2018-10-02
Add ManagedChannelBuilder/ServerBuilder.maxInboundMetadataSize

### LINKED ISSUE #4050: Generalize NettyChannelBuilder.maxHeaderListSize to ManagedChannelBuilder
This is necessary for Spanner in google-cloud-java.

https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java#L301

ejona86:
Support was added to InProcess and OkHttp transports. As mentioned
in #4567 OkHttp will not actually limit the memory used.

This API was a minor rename to the pre-existing Netty API,
so has already undergone API review and thus is not ExperimentalApi.

In adding transport tests I discovered Netty's behavior for trailers is
broken causing the stream to be orphaned and the client to hang. I have
verified this is fixed by https://github.com/netty/netty/pull/8332 .

The commits will remain separate when merging. They may help during
review, but probably aren't too necessary.

Fixes #4050 and #4567.

CC @garrettjonesgoogle

### DIFF
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -306,6 +306,28 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   }
 
   /**
+   * Sets the maximum size of metadata allowed to be received. {@code Integer.MAX_VALUE} disables
+   * the enforcement. The default is implementation-dependent, but is not generally less than 8 KiB
+   * and may be unlimited.
+   *
+   * <p>This is cumulative size of the metadata. The precise calculation is
+   * implementation-dependent, but implementations are encouraged to follow the calculation used for
+   * <a href="http://httpwg.org/specs/rfc7540.html#rfc.section.6.5.2">
+   * HTTP/2's SETTINGS_MAX_HEADER_LIST_SIZE</a>. It sums the bytes from each entry's key and value,
+   * plus 32 bytes of overhead per entry.
+   *
+   * @param bytes the maximum size of received metadata
+   * @return this
+   * @throws IllegalArgumentException if bytes is non-positive
+   * @since 1.17.0
+   */
+  public T maxInboundMetadataSize(int bytes) {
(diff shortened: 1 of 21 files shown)


core/src/main/java/io/grpc/inprocess/InProcessTransport.java
```diff
@@ -266,6 +295,19 @@ private synchronized void notifyTerminated() {
...
   }
 
+  private static int metadataSize(Metadata metadata) {
+    byte[][] serialized = InternalMetadata.serialize(metadata);
+    if (serialized == null) {
+      return 0;
+    }
+    long size = 0;
+    for (int i = 0; i < serialized.length; i += 2) {
+      size += 32 + serialized[i].length + serialized[i + 1].length;
```
zpencer: nit here and below: add a comment about the 32 coming from rfc7540 6.5.2

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -305,6 +305,23 @@ public T maxInboundMessageSize(int bytes) {
...
+   * the enforcement. The default is implementation-dependent, but is not generally less than 8 KiB
+   * and may be unlimited.
+   *
+   * <p>This is cumulative size of the headers. The precise calculation is implementation-dependent,
+   * but implementations are encouraged to follow the calculation used for
+   * <a href="http://httpwg.org/specs/rfc7540.html#rfc.section.6.5.2">
+   * HTTP/2's SETTINGS_MAX_HEADER_LIST_SIZE</a>. It sums the bytes from each entry's key and value,
+   * plus 32 bytes of overhead per entry.
+   */
+  public NettyChannelBuilder maxInboundMetadataSize(int maxInboundMetadataSize) {
```
zpencer: nit: was the parameter supposed to be named `bytes` for this set of functions? For InProcess it is currently named  `bytes`.

ejona86:
@zpencer PTAL

ejona86:
Rebased to get past conflicts. Waiting on CI and then will merge.

Issue #4901, opened 2018-10-03
CallCredentials interface, 2nd edition


zhangkun83:
Security level and authority are parameters required to be passed to applyRequestMetadata(). We will wrap them, along with MethodDescriptor and the transport attributes, to a first-class object `RequestInfo`, which is more clear to the implementers than via `Attributes`.

`ATTR_SECURITY_LEVEL` will be moved to the internal `GrpcAttributes` and annotated as `TransportAttr`, because transports are required to set it, but no user is actually reading them from
`{Client,Server}Call.getAttributes()`.

`ATTR_AUTHORITY` should be removed, because no transport is overriding it.

This is an API-breaking change because `CallCredentials` is `interface`. It will be changed to abstract class, so that further API changes can be done in multiple steps without breaking implementations.

The CallCredentials name is stabilized, thus it must be the name of the new interface. We will take the following steps to finish the change.

1. Introduce `CallCredentials2`, which is the new API, which implements the old `CallCredentials`. Change all stock implementations to it. Deprecate the old `CallCredentials`.
2. Make a release, and ask all implementations (including [FirestoreCallCredentials](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreCallCredentials.java)) to implement `CallCredentials2` instead.
3. Replace `CallCredentials` with `CallCredentials2`, keeping `CallCredentials2` as an alias.
4. Make a release, and ask all implementations (including FirestoreCallCredentials) to switch back to the name `CallCredentials`.
5. Delete the alias `CallCredentials2`

During the whole process, the name `CallCredentials` stays usable, and we require no change in code that passes `CallCredentials` objects along.

/cc @ejona86

ejona86:
As discussed some in https://github.com/grpc/grpc-java/issues/5002#issuecomment-433531077, it is fine if CallCredentials implementations _don't_ migrate to CallCredentials2 and just accept the breakage when it comes with 1.17.0.

carl-mastrangelo:
@zhangkun83 ping on this.  Can it be marked complete?

zhangkun83:
The only remaining item is "5. Delete the alias CallCredentials2"

carl-mastrangelo:
@zhangkun83 I think this should get some attention, as it has been on the todo list for a while.

ejona86:
We should have a few releases before removing CallCredentials2, since #5216 was only recent.

[...]

PR #4902, opened 2018-10-03
core: add CallCredentials2 and deprecate CallCredentials' old interface

zhangkun83:
This is the first step of introducing the 2nd edition of CallCredentials API (#4901)

Security level and authority are parameters required to be passed to
applyRequestMetadata(). This change wraps them, along with
MethodDescriptor and the transport attributes to RequestInfo, which is
more clear to the implementers.

ATTR_SECURITY_LEVEL is moved to the internal GrpcAttributes and
annotated as TransportAttr, because transports are required to set it,
but no user is actually reading them from
{Client,Server}Call.getAttributes().

ATTR_AUTHORITY is removed, because no transport is overriding it.

All involved interfaces are changed to abstract classes, as this will
make further API changes smoother.

### DIFF
--- a/core/src/main/java/io/grpc/CallCredentials.java
+++ b/core/src/main/java/io/grpc/CallCredentials.java
@@ -40,11 +40,15 @@ public interface CallCredentials {
    * The security level of the transport. It is guaranteed to be present in the {@code attrs} passed
    * to {@link #applyRequestMetadata}. It is by default {@link SecurityLevel#NONE} but can be
    * overridden by the transport.
+   *
+   * @deprecated transport implementations should use {@code
+   * io.grpc.internal.GrpcAttributes.ATTR_SECURITY_LEVEL} instead.
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1914")
   @Grpc.TransportAttr
+  @Deprecated
   public static final Key<SecurityLevel> ATTR_SECURITY_LEVEL =
-      Key.create("io.grpc.CallCredentials.securityLevel");
+      Key.create("io.grpc.internal.GrpcAttributes.securityLevel");
 
   /**
    * The authority string used to authenticate the server. Usually it's the server's host name. It
@@ -54,6 +58,7 @@ public interface CallCredentials {
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1914")
   @Grpc.TransportAttr
+  @Deprecated
   public static final Key<String> ATTR_AUTHORITY = Key.create("io.grpc.CallCredentials.authority");
 
   /**
@@ -75,6 +80,7 @@ public interface CallCredentials {
    *        method returns.
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1914")
+  @Deprecated
   void applyRequestMetadata(
       MethodDescriptor<?, ?> method, Attributes attrs,
       Executor appExecutor, MetadataApplier applier);
@@ -92,16 +98,16 @@ public interface CallCredentials {
    * <p>Exactly one of its methods must be called to make the RPC proceed.
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1914")
-  public interface MetadataApplier {
+  public abstract class MetadataApplier {
     /**
      * Called when headers are successfully generated. They will be merged into the original
      * headers.
      */
-    void apply(Metadata headers);
+    public abstract void apply(Metadata headers);
 
     /**
      * Called when there has been an error when preparing the headers. This will fail the RPC.
      */
-    void fail(Status status);
+    public abstract void fail(Status status);
   }
 }
--- a/core/src/main/java/io/grpc/CallCredentials.java
+++ b/core/src/main/java/io/grpc/CallCredentials.java
@@ -112,4 +112,31 @@ public interface CallCredentials {
      */
     public abstract void fail(Status status);
   }
+
+  /**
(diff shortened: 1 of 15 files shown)


core/src/main/java/io/grpc/CallCredentials2.java
```diff
@@ -0,0 +1,109 @@
...
+
+import java.util.concurrent.Executor;
+
+/**
+ * The new interface for {@link CallCredentials}.
+ *
+ * <p>THIS CLASS IS MEANT TO BE REFERENCED BY IMPLEMENTIONS ONLY.  All users should reference {@link
+ * CallCredentials}.
+ */
+public abstract class CallCredentials2 implements CallCredentials {
```
ejona86: This entire class should be marked experimental. It should deserve a note that it is part of a migration and is necessary to be used in the short-term but will be removed.

zhangkun83: I wanted to use the deprecation warnings on the old interface to push implementations to switch to the new interface. If we deprecate it now, implementations would have to suppress the deprecation warnings and may accidentally mask off implementations that have not been migrated, and would also cause confusion when we want implementations to switch from `CallCredentials2` back to `CallCredentials`. I think it's better to deprecate it after we replace `CallCredentials` with the new interface.

I have added the note, and an `@Internal` annotation to dissuade usage.

ejona86: I wasn't suggesting marking it deprecated, just experimental. I guess internal works as well, although it seems experimental is more appropriate since we do expect some "normal" users to use it. I'd normally prefer to keep internal a "if you are using this, you are probably doing it wrong."

zhangkun83: Sorry, read too fast :). Switched to experimental.

igorbernstein2:
I'm not sure if this is the right place to ask. But since you are creating a new interface for CallCredentials, would it be possible to make provisions to allow Credentials to take & respect the call deadlines?

carl-mastrangelo:
@igorbernstein2 The io.grpc.Context is available at the time, so you can get it from there.

igorbernstein2:
@carl-mastrangelo I apologize for being dense, but can you clarify what you mean by “the grpc Context is available”? If I’m not mistaken,  the deadline needs to be available inside applyRequestMetadata so that it can be passed to the credentials.

Thanks

igorbernstein2:
Ah, I see, via a threadlocal. Would it not be cleaner to make it part of the api? All of the grpc seems to pass it around as parameter, it feels odd that CallCredentials have to rely on a threadlocal.

igorbernstein2:
Actually never mind, I’m getting confused with gax. Sorry for the noise

carl-mastrangelo:
_Technically_ Context does not have to be backed by a thread local.   I have met users who use different backing storage.  It is implemented by default as a thread local because that was the paradigm internally at Google.   Legacy code has a strong influence, unfortunately.

igorbernstein2:
I took a closer look and it seems like `Context.current().getDeadline()` might not be enough.  A deadline might be set via `CallOptions.withDeadline` which will not be stamped on the current `Context` and not be accessible by the `CallCredentials`

zhangkun83:
I can easily add `CallOptions` to the `RequestInfo`

igorbernstein2:
Should CallCredentials subclasses be responsible for figuring out the effective deadline between grpc.Context and CallOptions? I don't have a good answer, I was just requesting for this be a consideration in the api change

zhangkun83:
@igorbernstein2, can you file an issue for accessing deadline from CallCredentials?
It's not as trivial as I thought initially. Let's discuss it on its own issue.

igorbernstein2:
done: #4929

carl-mastrangelo:
@zhangkun83 I don't feel comfortable expecting CallCredentials implementations having to recalculate the deadline as done in ClientCallImpl.  Can the effective deadline be passed in instead?

zhangkun83:
@carl-mastrangelo, let's discuss this in #4929

core/src/main/java/io/grpc/CallCredentials2.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+      };
+    applyRequestMetadata(requestInfo, appExecutor, applier);
+  }
+
+  /**
+   * The request-related information passed to {@link #applyRequestMetadata}.
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1914")
+  public abstract static class RequestInfo {
```
ejona86: Should this be in CallCredentials? I'm thinking that moving the class to CallCredentials later may not be ABI-safe.

core/src/main/java/io/grpc/CallCredentials.java
```diff
@@ -92,16 +100,16 @@ void applyRequestMetadata(
    * <p>Exactly one of its methods must be called to make the RPC proceed.
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1914")
-  public interface MetadataApplier {
+  public abstract class MetadataApplier {
```
carl-mastrangelo: static?

zhangkun83: An class under an interface is implicitly static. But it's no harm to add it (and it will be needed once CallCredentials becomes an abstract class).

core/src/main/java/io/grpc/CallCredentials2.java
zhangkun83: Makes sense. Moved to CallCredentials.

PR #4907, opened 2018-10-04
netty: expose setting a local socket address

### LINKED ISSUE #4900: Need a way to specify source address when creating ManagedChannel as a client
Is it possible to provide an option to specify source address to use when creating the ManagedChannel as a client. At the moment, the choice of source address appears to have been left to the operating system?

Current:
`ManagedChannel channel = NettyChannelBuilder.forAddress(SERVER_ADDRESS, SERVER_PORT).usePlaintext().build();
`
Proposed:
`ManagedChannel channel = NettyChannelBuilder.forAddress(SERVER_ADDRESS, SERVER_PORT).sourceAddress(SPECIFIC_LOCAL_ADDRESS).usePlaintext().build();`

1.14.0

An option to specify a source address to use on client side.

carl-mastrangelo:
Fixes #4900

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
+++ b/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
@@ -80,6 +80,7 @@ public final class NettyChannelBuilder
   private long keepAliveTimeoutNanos = DEFAULT_KEEPALIVE_TIMEOUT_NANOS;
   private boolean keepAliveWithoutCalls;
   private ProtocolNegotiatorFactory protocolNegotiatorFactory;
+  private LocalSocketPicker localSocketPicker;
 
   /**
    * Creates a new builder with the given server address. This factory method is primarily intended
@@ -326,6 +327,28 @@ public final class NettyChannelBuilder
     return this;
   }
 
+
+  /**
+   * If non-{@code null}, attempts to create connections bound to a local port.
+   */
+  public NettyChannelBuilder localSocketPicker(@Nullable LocalSocketPicker localSocketPicker) {
+    this.localSocketPicker = localSocketPicker;
+    return this;
+  }
+
+  @ExperimentalApi("FIXME")
+  public interface LocalSocketPicker {
+
+    /**
+     * Called by gRPC to pick local socket to bind to.  This may be called multiple times.
+     *
+     * @param remoteAddress the remote address to connect to.
+     * @param attrs the Attributes present on the {@link io.grpc.EquivalentAddressGroup} associated
+     *        with the address.
+     */
+    @Nullable SocketAddress createSocketAddress(SocketAddress remoteAddress, Attributes attrs);
+  }
+
   @Override
   @CheckReturnValue
   @Internal
@@ -348,7 +371,7 @@ public final class NettyChannelBuilder
         negotiator, channelType, channelOptions,
         eventLoopGroup, flowControlWindow, maxInboundMessageSize(),
         maxHeaderListSize, keepAliveTimeNanos, keepAliveTimeoutNanos, keepAliveWithoutCalls,
-        transportTracerFactory.create());
+        transportTracerFactory.create(), localSocketPicker);
   }
 
   @Override
@@ -457,6 +480,7 @@ public final class NettyChannelBuilder
     private final long keepAliveTimeoutNanos;
     private final boolean keepAliveWithoutCalls;
     private final TransportTracer transportTracer;
+    private final LocalSocketPicker localSocketPicker;
 
     private boolean closed;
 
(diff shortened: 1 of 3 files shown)


netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -326,6 +327,28 @@ public NettyChannelBuilder keepAliveWithoutCalls(boolean enable) {
...
+  /**
+   * If non-{@code null}, attempts to create connections bound to a local port.
+   */
+  public NettyChannelBuilder localSocketPicker(@Nullable LocalSocketPicker localSocketPicker) {
+    this.localSocketPicker = localSocketPicker;
+    return this;
+  }
+
+  @ExperimentalApi("FIXME")
+  public interface LocalSocketPicker {
```
zhangkun83: Probably better to be an abstract class to make it easier to change the interface in the future.

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -326,6 +327,28 @@ public NettyChannelBuilder keepAliveWithoutCalls(boolean enable) {
...
+
+    /**
+     * Called by gRPC to pick local socket to bind to.  This may be called multiple times.
+     *
+     * @param remoteAddress the remote address to connect to.
+     * @param attrs the Attributes present on the {@link io.grpc.EquivalentAddressGroup} associated
+     *        with the address.
+     */
+    @Nullable SocketAddress createSocketAddress(SocketAddress remoteAddress, Attributes attrs);
```
zhangkun83: Annotate attrs with `@EquivalentAddressGroup.Attr`.

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
zhangkun83: I am not sure if this should be a transport-agnostic interface or netty-only, but I guess if we ever need to support it in other transports, we could just promote the setter to `ManagedChannelBuilder`.

carl-mastrangelo: Done.

carl-mastrangelo: yeah, this is experimental.  Just trying it out.

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
carl-mastrangelo: Done.

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -326,6 +328,41 @@ public NettyChannelBuilder keepAliveWithoutCalls(boolean enable) {
...
+    return this;
+  }
+
+  /**
+   * This class is meant to be overriden with a custom implementation of
+   * {@link #createSocketAddress}.  The default implementation is a no-op.
+   *
+   * @since 1.16.0
+   */
+  @ExperimentalApi("FIXME")
```
zhangkun83: Please file an issue to track it.

carl-mastrangelo: Done.

netty/src/main/java/io/grpc/netty/NettyClientTransport.java
```diff
@@ -263,7 +272,7 @@ public void operationComplete(ChannelFuture future) throws Exception {
       }
     });
     // Start the connection operation to the server.
-    channel.connect(address);
+    channel.connect(remoteAddress);
```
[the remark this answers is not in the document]
carl-mastrangelo: This is surprising.  From what I can tell, the socket is created at channel creation time, not at connect, so I assumed the local address already had to be present.  

I didn't test this myself, since our CI doesn't have a stable address to test this with.  @normanmaurer can comment if this is expected behavior.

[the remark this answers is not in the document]
nikhilvgs: Yes - it is surprising, but it appears that localAddress only takes effect on a bind. Not sure we want to be doing a bind on the channel (though internally, that is what is probably getting called on the socket). 

I tried this locally by building the library with this diff and it works for me:

ngshetty-Mac:grpc-java ngshetty$ git diff
diff --git a/netty/src/main/java/io/grpc/netty/NettyClientTransport.java b/netty/src/main/java/io/grpc/netty/NettyClientTransport.java
index 2be71c20d..7cf964b58 100644
--- a/netty/src/main/java/io/grpc/netty/NettyClientTransport.java
+++ b/netty/src/main/java/io/grpc/netty/NettyClientTransport.java
@@ -272,7 +272,11 @@ class NettyClientTransport implements ConnectionClientTransport {
       }
     });
     // Start the connection operation to the server.
-    channel.connect(remoteAddress);
+    if (localAddress == null) {
+      channel.connect(remoteAddress);
+    } else {
+      channel.connect(remoteAddress, localAddress);
+    }

     if (keepAliveManager != null) {

PR #4908, opened 2018-10-04
Fix copy pasta error in method comment

IcyMidnight:
"Constructs an exception with both a status." -> "Constructs an exception with a status."


thelinuxfoundation:

* User @IcyMidnight isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

PR #4910, opened 2018-10-04
config stats and tracing at InProcessChannelBuilder

yuzheng21:
make `setStatsEnabled` and `setTracingEnabled` public to support test retry feature

### DIFF
--- a/core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
@@ -128,6 +128,16 @@ public final class InProcessChannelBuilder extends
     return this;
   }
 
+  @Override
+  public void setStatsEnabled(boolean value) {
+    super.setStatsEnabled(value);
+  }
+
+  @Override
+  public void setTracingEnabled(boolean value) {
+    super.setTracingEnabled(value);
+  }
+
   /**
    * Provides a custom scheduled executor service.
    *


thelinuxfoundation:

* User @yuzheng21 isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

dapengzhang0:
\+ @zhangkun83 @ejona86

dapengzhang0:
@yuzheng21 Why InProcessChannel? For testing?

yuzheng21:
@dapengzhang0 yes. for testing. These two lines makes writing tests and retry examples much easier. I can use InProcess server and channel to confirm the retry behavior now. Without it, I need to use `InternalNettryChannelBuilder` to disable stats and tracking to enable retry.

zhangkun83:
I don't feel comfortable exposing any public API for stats and tracing, because they largely remain implementation details of the core and subject to change. If testing our retry is your purpose, we could make some changes in the retry implementation to make it work when stats and tracing enabled, although the stats and tracing behavior may not be ideal, which you probably don't care. @dapengzhang0

yuzheng21:
Hi @zhangkun83 , thanks for your reply. This is only to expose stats and tracing at `InProcessChannelBuilder.java`, which is designed for test use (if I understand correctly). I see `InternalNettryChannelBuilder` does the similar way to disable stats and tracing. But yeah, it is also great to disable this inside retry implementation, as long as retry can be enabled via exposed public API.

zhangkun83:
@yuzheng21, while InProcess channel/server are extremely useful in tests, they can also be used in production. We intent to keep the options internal because they are not the final form that we expect thus they are subject to change.

yuzheng21:
@zhangkun83 ic. so, instead of exposing these two methods public, how about setting stats and tracing to be disabled when retry is enabled?

zhangkun83:
Given current code, stats and tracing will throw with retry enabled. It sounds good to me if we just automatically disable stats and tracing if retry is enabled.

dapengzhang0:
@ejona86 @zhangkun83  To disable stats and tracing when retry is enable for the moment seems good to me. I will work on it.

dapengzhang0:
Now `enableRetry()` will disable census in master branch/next release for the moment. See #4941.
Thanks @yuzheng21 for raising up this issue.

PR #4916, opened 2018-10-05
okhttp: move async mechanism from FrameWriter to sink (AsyncSink)

### LINKED ISSUE #4860: Android OOM upon network disconnect
With grpc-java 1.15.0 on Android I am seeing out of memory issues when network connectivity goes away. `isReady()` indicates unlimited readiness on the outbound stream for a bidi call until Java runs out of memory.

Here is the relevant code, the complete example is at https://github.com/bubenheimer/grpcsinglestreamingcallflush

    StreamDataGrpc.newStub(channel).streamData(
            new ClientResponseObserver<Item, Item>() {
                private int outCounter = 0;
                private int inCounter = 0;
    
                @Override
                public void beforeStart(
                        final ClientCallStreamObserver<Item> requestStream) {
                    requestStream.setOnReadyHandler(() -> {
                        while (requestStream.isReady()) {
                            final Item item = Item.newBuilder()
                                    .setValue(Integer.toString(outCounter++)).build();
                            requestStream.onNext(item);
                            if (outCounter == 50_000) {
                                requestStream.onCompleted();
                                break;
                            }
                        }
                    });
                }
                @Override
                public void onNext(final Item value) {
                    ++inCounter;
                }
    
                @Override
                public void onError(final Throwable t) {
                    Log.w(TAG, t);
                }
    
                @Override
                public void onCompleted() {
                    Log.i(TAG, "Call completed after " + inCounter + " Items");
                }
            });

where `channel` is more persistently defined as

    channel = AndroidChannelBuilder
                // localhost server from Android emulator
                .forAddress("10.0.2.2", 8082)
                .usePlaintext()
                .context(getApplicationContext())
                .build();

To trigger the issue I do one complete run of the RPC (50,000 messages upstream & 50,000 messages downstream) with the mobile data network on, then I turn mobile data off and do a second run on the same channel. The second run quickly runs out of memory as grpc-java buffers messages it can't send. The channel does not see a break in the connection when mobile data switches off.

I expected that isReady() would take care of this problem, but it did not. I have tried different variations, including one where writing to the stream is done from a separate, dedicated thread, with 1ms delays between each write, to not block anything and avoid races, but the outcome was the same.

creamsoup:
OkHttp transport's memory useage by removing queuing in 3 locations (AsyncFrameWriter, OutboundFlowController and OkHttpClientStream).
Also, removing AsyncFrameWriter and push the async operation down to Sink to keep the flush coalescing. This new approach has similar performance characteristic (throughput and each latency is
within 1% of original performance).

Resolves #4860 

NOTE: if AsyncSink is not used, it is actually faster (qps, latency) than original implementation. But, implementing flush coalescing can be challenging.


okhttp/src/main/java/io/grpc/okhttp/AsyncSink.java
```diff
@@ -0,0 +1,131 @@
...
+import javax.annotation.concurrent.GuardedBy;
+import okio.Buffer;
+import okio.Sink;
+import okio.Timeout;
+
+/**
+ * A sink that asynchronously write / flushes a buffer internally. AsyncSink provides flush
+ * coalescing to minimize network packing transmit.
+ */
+class AsyncSink implements Sink {
```
carl-mastrangelo: final

okhttp/src/main/java/io/grpc/okhttp/AsyncSink.java
```diff
@@ -0,0 +1,131 @@
...
+
+  static AsyncSink sink(Sink sink, SerializingExecutor executor) {
+    return new AsyncSink(sink, executor);
+  }
+
+  @Override
+  public void write(final Buffer source, final long byteCount) throws IOException {
+    checkNotNull(source, "source");
+    if (closed) {
+      throw new IOException("closed");
```
carl-mastrangelo: Maybe make the exception the cause?

okhttp/src/main/java/io/grpc/okhttp/AsyncSink.java
```diff
@@ -0,0 +1,131 @@
...
+      @Override
+      public void run() {
+        try {
+          synchronized (lock) {
+            if (buffer.size() > 0) {
+              sink.write(buffer, buffer.size());
+            }
+          }
+        } catch (IOException e) {
+          exception = e;
```
carl-mastrangelo: doesn't the exception need to be volatile?

okhttp/src/main/java/io/grpc/okhttp/AsyncSink.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+      }
+    });
+  }
+
+  @Override
+  public void flush() throws IOException {
+    if (closed) {
+      throw new IOException("closed");
```
carl-mastrangelo: chain the exception?

okhttp/src/main/java/io/grpc/okhttp/DelegatingFrameWriter.java
```diff
@@ -0,0 +1,204 @@
...
+import io.grpc.okhttp.internal.framed.Header;
+import io.grpc.okhttp.internal.framed.Settings;
+import java.io.IOException;
+import java.net.Socket;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import okio.Buffer;
+
+class DelegatingFrameWriter implements FrameWriter {
```
carl-mastrangelo: final

okhttp/src/main/java/io/grpc/okhttp/DelegatingFrameWriter.java
```diff
@@ -0,0 +1,204 @@
...
+  public void flush() {
+    try {
+      frameWriter.flush();
+    } catch (Exception e) {
+      transportExceptionHandler.onException(e);
+    }
+  }
+
+  @Override
+  public void synStream(boolean outFinished, boolean inFinished, int streamId,
```
carl-mastrangelo: This line wrapping is kinda funky.  I prefer to put all the args on their own line if it would wrap.  This also follows the rectangle rule.

okhttp/src/main/java/io/grpc/okhttp/DelegatingFrameWriter.java
```diff
@@ -0,0 +1,204 @@
...
+      transportExceptionHandler.onException(e);
+    }
+  }
+
+  @Override
+  public void close() {
+    if (frameWriter != null) {
+      try {
+        frameWriter.close();
+        socket.close();
```
carl-mastrangelo: Maybe null out socket and frameWriter ?

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -389,13 +391,13 @@ Object getOutboundFlowState() {
     return outboundFlowState;
   }
 
-  private static class PendingData {
-    Buffer buffer;
-    boolean endOfStream;
-    boolean flush;
+  private static class PendingFrameInfo {
```
carl-mastrangelo: final

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -421,115 +412,99 @@ public Runnable start(Listener listener) {
...
+    outboundFlow = new OutboundFlowController(this, frameWriter);
+
+    try {
+      // Do these with the raw FrameWriter, so that they will be done in this thread,
+      // and before any possible pending stream operations.
+      rawFrameWriter.connectionPreface();
+      Settings settings = new Settings();
+      rawFrameWriter.settings(settings);
+    } catch (Exception e) {
+      onException(e);
```
carl-mastrangelo: should this return?

okhttp/src/main/java/io/grpc/okhttp/AsyncSink.java
```diff
@@ -0,0 +1,133 @@
...
+    synchronized (lock) {
+      buffer.write(source, byteCount);
+    }
+    serializingExecutor.execute(new Runnable() {
+      @Override
+      public void run() {
+        try {
+          synchronized (lock) {
+            if (buffer.size() > 0) {
+              sink.write(buffer, buffer.size());
```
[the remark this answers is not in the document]
ejona86: Doh, I forgot to update this. Even if `sink` is buffered, it may still do I/O if a lot of data is already buffered.

okhttp/src/main/java/io/grpc/okhttp/AsyncSink.java
```diff
@@ -0,0 +1,149 @@
...
+        return;
+      }
+    }
+    serializingExecutor.execute(new Runnable() {
+      @Override
+      public void run() {
+        Buffer buf = new Buffer();
+        synchronized (lock) {
+          while (buffer.size() >= COMPLETE_SEGMENT_SIZE) {
+            buf.write(buffer, COMPLETE_SEGMENT_SIZE);
```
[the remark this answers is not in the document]
creamsoup: discussed offline, using okio.Buffer's completeSegmentByteCount to determine actual write to sink.

creamsoup:
actually, during the offline discuss i mentioned that flush can clear writeEnqueued to false. This is actually sub-optimal and lie in case write is really enqueued to the executor. Thus, it could create a 2nd write task which doesn't need to be there since there's already a write task.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -144,15 +144,14 @@
   private final int initialWindowSize;
   private Listener listener;
   private FrameReader testFrameReader;
-  private AsyncFrameWriter frameWriter;
+  private ExceptionHandlingFrameWriter frameWriter;
```
ejona86: Does this now need to be GuardedBy("lock")? AsyncFrameWriter was thread-safe whereas the new implementation is not.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -432,125 +422,111 @@ void removePendingStream(OkHttpClientStream pendingStream) {
...
-          executor.execute(clientFrameHandler);
-          synchronized (lock) {
-            maxConcurrentStreams = Integer.MAX_VALUE;
-            startPendingStreams();
-          }
-          frameWriter.becomeConnected(testFrameWriter, socket);
-          connectedFuture.set(null);
-          return;
-        }
+    synchronized (lock) {
```
ejona86: This lock does I/O while held. That would cause `shutdownNow()` to block.

I'm thinking we can take a page from the previous code: let's go ahead and create `AsyncSink` and delay creating `Okio.sink(sock)`. We can schedule work on the `serializingExecutor` to create the connection and then afterward use a simple setter to set it on the `AsyncSink`.

ejona86:
> actually, during the offline discuss i mentioned that flush can clear writeEnqueued to false. This is actually sub-optimal and lie in case write is really enqueued to the executor. Thus, it could create a 2nd write task which doesn't need to be there since there's already a write task.

There won't be a second write because `flushEnqueued == true`.

I'd argue that clearing `writeEnqueued` alone does nothing. To make it useful you also would add a check in the write's Runnable to see whether `writeEnqueued == true`. If it is false there must be a flush coming and so you can return early.

Everything said, this is a minor case, and so may not be worth the effort to consider optimizing it.

okhttp/src/main/java/io/grpc/okhttp/ExceptionHandlingFrameWriter.java
```diff
@@ -0,0 +1,224 @@
...
+      transportExceptionHandler.onException(e);
+    }
+  }
+
+  @Override
+  public void close() {
+    try {
+      frameWriter.close();
+      if (socket != null) {
+        socket.close();
```
ejona86: This close shouldn't be here. It should be in AsyncSink instead. Being here would prevent writes from going on when shutting down.

It looks like you may have done this for testing. We may need to tweak how testing is plumbed.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -447,29 +446,48 @@ public Runnable start(Listener listener) {
           keepAliveWithoutCalls);
       keepAliveManager.onTransportStarted();
     }
+    final OkHttpClientTransport okHttpClientTransport = this;
```
ejona86: This doesn't look necessary. FYI: If you are within an anonymous class, you can do `OkHttpClientTransport.this`

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -885,7 +899,17 @@ private void stopIfNecessary() {
 
     // We will close the underlying socket in the writing thread to break out the reader
     // thread, which will close the frameReader and notify the listener.
-    frameWriter.close();
+    // When the transport is middle of start and received stop before the frameWriter init
```
ejona86: That would be an invalid API call; start() must have returned before calling shutdown(). The only way it looks like that could happen is via KeepAliveManager (although it would be a _very_ rare race), because it is initialized before the lock (which could be fixed)

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -447,29 +446,48 @@ public Runnable start(Listener listener) {
...
+    if (isForTest()) {
+      synchronized (lock) {
+        frameWriter = new ExceptionHandlingFrameWriter(okHttpClientTransport, testFrameWriter);
+        outboundFlow =
+            new OutboundFlowController(okHttpClientTransport, frameWriter, initialWindowSize);
+      }
+      serializingExecutor.execute(new Runnable() {
+        @Override
+        public void run() {
+          if (isForTest()) {
```
ejona86: Remove this check?

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -527,19 +544,16 @@ public void close() {}
...
           if (sslSession != null) {
             securityInfo = new InternalChannelz.Security(new InternalChannelz.Tls(sslSession));
           }
         }
 
-        rawFrameWriter = variant.newWriter(sink, true);
-        frameWriter.becomeConnected(rawFrameWriter, socket);
-
         try {
           // Do these with the raw FrameWriter, so that they will be done in this thread,
```
ejona86: This doesn't look like it will work any more. It looks like they may need to be done directly within start().

okhttp/src/main/java/io/grpc/okhttp/ExceptionHandlingFrameWriter.java
```diff
@@ -0,0 +1,214 @@
...
+final class ExceptionHandlingFrameWriter implements FrameWriter {
+
+  private static final Logger log = Logger.getLogger(OkHttpClientTransport.class.getName());
+  // Some exceptions are not very useful and add too much noise to the log
+  private static final Set<String> QUIET_ERRORS =
+      Collections.unmodifiableSet(new HashSet<>(Arrays.asList("Socket closed")));
+
+  private final TransportExceptionHandler transportExceptionHandler;
+
+  private FrameWriter frameWriter;
```
ejona86: final

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -465,98 +464,104 @@ public void run() {
...
-        try {
-          if (proxy == null) {
-            sock = new Socket(address.getAddress(), address.getPort());
-          } else {
-            sock = createHttpProxySocket(
-                address, proxy.proxyAddress, proxy.username, proxy.password);
-          }
+      // Connecting in the serializingExecutor, so that some stream operations like synStream
+      // will be executed after connected.
+      serializingExecutor.execute(new Runnable() {
```
ejona86: Including this execute within the lock seems unnecessary and indents the code making it very difficult to review. Could it be removed from the lock to be unindented?

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -465,98 +464,104 @@ public void run() {
...
-          rawFrameWriter.connectionPreface();
-          Settings settings = new Settings();
-          rawFrameWriter.settings(settings);
-        } catch (Exception e) {
-          onException(e);
-          return;
-        }
-      }
-    });
+      // Do these with the transport lock, so that they will be done right after the start and
```
ejona86: This seems a bit weird of a comment, since any usage of `frameWriter` must be done with the transport lock. Maybe:

> Enqueue these immediately, so that they will be the first writes on the connection.

creamsoup:
just discovered that test being flaky is actually from the race between application write and the initial settings (connectionPreface and settings).

To avoid the race, need to perform the preface and settings with the transport lock to make sure no such race can happen. Since this make the synchronized block to be very big, refactored the Runnable to a inner class.

@ejona86 can you review this part again? also, there was some change in OkHttpClientTransportTest (using custom capture).

ejona86:
@creamsoup, can you point me to the test that was seeing the race? start() should have returned before anything else was done with the transport.

creamsoup:
most failure is from travis, i tested on 5 different environment (2 laptops, workstation, kokoro, travis). without the synchronized block change, i only see the issue on travis. Before other fixes/modification, I observed same behavior on my chromebook and travis. 

okhttp test timed out
https://travis-ci.org/grpc/grpc-java/jobs/451485104 

interops test timeout
https://travis-ci.org/grpc/grpc-java/jobs/450359238

PR #4918, opened 2018-10-06
services: fix callbacks not being forwarded in binlog interceptors

zpencer:
The callback must be forwarded otherwise the RPC hangs.


zpencer:
@carl-mastrangelo

PR #4920, opened 2018-10-08
services: fix channelz javadoc lint for reference not found

zpencer:
Referring to the public abstract class seems to be OK, but referring
to the package private impl trips the linter.

### DIFF
--- a/services/src/main/java/io/grpc/services/ChannelzService.java
+++ b/services/src/main/java/io/grpc/services/ChannelzService.java
@@ -61,7 +61,7 @@ public final class ChannelzService extends ChannelzGrpc.ChannelzImplBase {
     this.maxPageSize = maxPageSize;
   }
 
-  /** Returns top level channel aka {@link io.grpc.internal.ManagedChannelImpl}. */
+  /** Returns top level channel aka {@link io.grpc.ManagedChannel}. */
   @Override
   public void getTopChannels(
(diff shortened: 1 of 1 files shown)


services/src/main/java/io/grpc/services/ChannelzService.java
```diff
@@ -61,7 +61,7 @@ public static ChannelzService newInstance(int maxPageSize) {
     this.maxPageSize = maxPageSize;
   }
 
-  /** Returns top level channel aka {@link io.grpc.internal.ManagedChannelImpl}. */
+  /** Returns top level channel aka {@link io.grpc.ManagedChannel}. */
   @Override
   public void getTopChannels(
```
dapengzhang0: An ordinary `//` comment here seems more suitable.
```
/** @inheritDoc */
// Returns top level channel aka {@link io.grpc.internal.ManagedChannelImpl}.
```

After all, the javadoc of its super class is available.

zpencer: I get the error below. It think it may be because the protobuf generated javadoc are inside `<pre>` tags, so the linter does not treat it as a sentence:
` First sentence of Javadoc is incomplete (period is missing) or not present. [SummaryJavadoc]`

dapengzhang0: Hmm, interesting.

PR #4930, opened 2018-10-10
services: implement Health.Watch

zhangkun83:
Part of #4932

### DIFF
--- a/services/src/main/java/io/grpc/services/HealthServiceImpl.java
+++ b/services/src/main/java/io/grpc/services/HealthServiceImpl.java
@@ -17,8 +17,6 @@
 package io.grpc.services;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.HashMultimap;
-import com.google.common.collect.Multimap;
 import com.google.common.util.concurrent.MoreExecutors;
 import io.grpc.Context.CancellationListener;
 import io.grpc.Context;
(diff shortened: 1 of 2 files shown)


services/src/main/java/io/grpc/services/HealthServiceImpl.java
```diff
@@ -16,30 +16,40 @@
...
-  private final Map<String, ServingStatus> statusMap
-      = new ConcurrentHashMap<String, ServingStatus>();
+   * However, need use ConcurrentHashMap to allow concurrent reading by check() */
+  private final ConcurrentHashMap<String, ServingStatus> statusMap =
+      new ConcurrentHashMap<String, ServingStatus>();
+
+  private final Object watchLock = new Object();
+
+  @GuardedBy("watchLock")
+  private final Multimap<String, StreamObserver<HealthCheckResponse>> watchers =
```
carl-mastrangelo: If possible I think it would be good if this was a Map<String, List<StreamObserver>>, to avoid guava deps

services/src/main/java/io/grpc/services/HealthServiceImpl.java
```diff
@@ -50,16 +60,63 @@ public void check(HealthCheckRequest request,
...
+      responseObserver.onNext(getResponseForWatch(status));
+      watchers.put(service, responseObserver);
+    }
+    Context.current().addListener(
+        new CancellationListener() {
+          @Override
+          // Called when the client has closed the stream
+          public void cancelled(Context context) {
+            synchronized (watchLock) {
+              watchers.remove(service, responseObserver);
```
carl-mastrangelo: I don't think there is promise about the equality of observers.   An interceptor may pass in a StreamObserver that has equality based on call state.   

I think the multimap _values_  should be an identity hash set.  That would make removal O(1), and avoid equality.

services/src/main/java/io/grpc/services/HealthServiceImpl.java
zhangkun83: Although Guava collections dependency is permissible for server-side code, making Multimap use identity hash set would require extra code and lose the benefits of being shorter. I changed to the plain Java collections.

services/src/main/java/io/grpc/services/HealthServiceImpl.java
zhangkun83: `HashMultimap` is O(1) removal, but the equality is a valid concern. Switch to identity hash set.

PR #4949, opened 2018-10-14
Update net.ltgt.errorprone plugin and Error Prone, enable Error Prone on JDK 10, and JDK 11

tbroyer:
Note that the update of Error Prone past 2.3.1 adds a lot of warnings, and emits an error related to `GuardedBy` (I changed the code to please Error Prone, and added a FIXME because I think Error Prone is right, and the change is not a real fix).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,14 +41,8 @@ os:
   - linux
 
 jdk:
-  # net.ltgt.errorprone supports jdk8 and jdk9, but has problems with jdk10
-  # For jdk10, we need to switch over to using net.ltgt.errorprone-javacplugin,
-  # and likely update to the latest com.google.errorprone:error_prone_core.
-  # We have decided not to make our build.gradle support both plugins, so when
-  # we finally move off of jdk8 and jdk9 we will need use the javac annotation
-  # processor based plugin.
-  - oraclejdk8  # if both jdk 8 and 9 are removed, migrate to net.ltgt.errorprone-javacplugin (see above comment)
-  - oraclejdk9  # if both jdk 8 and 9 are removed, migrate to net.ltgt.errorprone-javacplugin (see above comment)
+  - oraclejdk8
+  - oraclejdk9
   - oraclejdk10
 
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,7 @@ jdk:
   - oraclejdk8
   - oraclejdk9
   - oraclejdk10
+  - oraclejdk11
 
 notifications:
   email: false
--- a/alts/build.gradle
+++ b/alts/build.gradle
@@ -32,6 +32,7 @@ dependencies {
         // prefer 20.0 from libraries instead of 19.0
         exclude group: 'com.google.guava', module: 'guava'
     }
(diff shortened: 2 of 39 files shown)


thelinuxfoundation:

* User @tbroyer isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

tbroyer:
If you look at the first commit (https://github.com/grpc/grpc-java/pull/4949/commits/8a6cbef1c90672a246a73d7e927132b46088bf2a), it only adds `ByteBufferBackingArray` and `CatchAndPrintStackTrace` warnings (that I suppressed in the code).

The second commit (https://github.com/grpc/grpc-java/pull/4949/commits/8b4764db5e6b6c23f96c4be345713efbc96f31c0, bumping to 2.3.2, so that Error Prone could be enabled on JDK 10+) required a few changes to avoid [`UnnecessaryParentheses`](https://errorprone.info/bugpattern/UnnecessaryParentheses), [`EqualsGetClass`](https://errorprone.info/bugpattern/EqualsGetClass), and [`UndefinedEquals`](https://errorprone.info/bugpattern/UndefinedEquals) mostly, and a couple [`BadImport`](https://errorprone.info/bugpattern/BadImport).

Wrt `UndefinedEquals` in tests, using Truth might be a good idea for `Iterable`s (using `containsExactlyElementsIn`), and because `AsciiString`s are `Comparable`, Truth would give you a `ComparableSubject` so `comparesEqualTo` could be used (or alternatively using `toString()` as suggested by Error Prone)

But I think updating Error Prone would be a good idea: it found a real (IIUC, as I'm actually really new to the codebase) issue in `OkHttpClientTransport`. I think `TransportTracer` should ideally be made threadsafe rather than guarded by a lock (but then being shared between objects, each using its own lock, so incorrectly guarding access)

Do you mind if I continue trying to make the project build with JDK 11? Or would you prefer that I split this PR?

ejona86:
Let's keep JDK 11 separate. I'd also be happy to separate upgrading of the errorprone plugin from error prone itself. I hadn't realized that net.ltgt.errorprone was merged with errorprone-javacplugin. That's great news.

Note that for new ErrorProne checks that are a bit too invasive for us initially, we can either disable it for all of our projects (like done [for tests here](https://github.com/grpc/grpc-java/blob/v1.15.0/build.gradle#L89), I think because of LinkedList, but maybe something else also) or for a particular project (like done for [netty and futures](https://github.com/grpc/grpc-java/blob/v1.15.0/netty/build.gradle#L19)). We can have an issue tracking cleaning up the failure.

ejona86:
I'm working on Java 11 in https://github.com/grpc/grpc-java/pull/4958 (there's a failure in the examples).

tbroyer:
OK, so I've created #4965 that bumps Error Prone to 2.3.2 (as bumping to 2.3.x is a prerequisite to updating net.ltgt.errorprone when using custom checks like BetaApi).

I kept the same changes as in this PR for now, let me know if you'd prefer that I make fewer changes to the code and instead disable the checks.

Issue #4932, opened 2018-10-10
Implement client-side health checking


zhangkun83:
Spec: https://github.com/grpc/proposal/blob/master/A17-client-side-health-checking.md

PR #4950, opened 2018-10-15
core: add internal Subchannel.asChannel()

zhangkun83:
Returns a Channel that allows a LoadBalancer to make auxiliary RPCs on already-established application connections.  We need this to implement client-side health-checking (#4932)

See comments on the API for its semantics.

Notable changes:
- Transports are modified to use InUseStateAggregator so that they can exclude RPCs made on Subchannel.asChannel() when reporting in-use state for idle mode.
- OobChannel shares the same Executor as Subchannel.asChannel(). Because the latter is not a ManagedChannel and doesn't have life-cycle, thus can't determine when to return the Executor to a pool, the Executor is now returned only when ManagedChannelImpl is terminated.

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractClientStream.java
+++ b/core/src/main/java/io/grpc/internal/AbstractClientStream.java
@@ -96,7 +96,7 @@ public abstract class AbstractClientStream extends AbstractStream
 
   private final TransportTracer transportTracer;
   private final Framer framer;
-  private final CallOptions callOptions;
+  private boolean shouldBeCountedForInUse;
   private boolean useGet;
   private Metadata headers;
   /**
@@ -115,7 +115,7 @@ public abstract class AbstractClientStream extends AbstractStream
       boolean useGet) {
     checkNotNull(headers, "headers");
     this.transportTracer = checkNotNull(transportTracer, "transportTracer");
-    this.callOptions = checkNotNull(callOptions, "callOptions");
+    this.shouldBeCountedForInUse = GrpcUtil.shouldBeCountedForInUse(callOptions);
     this.useGet = useGet;
     if (!useGet) {
       framer = new MessageFramer(this, bufferAllocator, statsTraceCtx);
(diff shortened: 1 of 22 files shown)


zhangkun83:
@ejona86, please review the PR as a whole.

netty/src/main/java/io/grpc/netty/CreateStreamCommand.java
```diff
@@ -27,16 +28,21 @@
...
   private final boolean get;
+  private final CallOptions callOptions;
 
-  CreateStreamCommand(Http2Headers headers,
-                      NettyClientStream.TransportState stream) {
-    this(headers, stream, false);
+  CreateStreamCommand(
+      Http2Headers headers,
+      NettyClientStream.TransportState stream) {
+    this(headers, stream, CallOptions.DEFAULT, false);
```
ejona86: Lying about the CallOptions worries me. Marking this a Deprecated or VisibleForTesting would make me feel better.

netty/src/main/java/io/grpc/netty/NettyClientHandler.java
```diff
@@ -530,6 +538,10 @@ public void operationComplete(ChannelFuture future) throws Exception {
                     stream.getStatsTraceContext().clientOutboundHeaders();
                     http2Stream.setProperty(streamKey, stream);
 
+                    if (GrpcUtil.shouldBeCountedForInUse(command.callOptions())) {
+                      inUseState.updateObjectInUse(http2Stream, true);
```
ejona86: Let's put a comment here that this delays the inUse state until the I/O completes, which technically may be later than we would like.

core/src/main/java/io/grpc/internal/AbstractClientStream.java
```diff
@@ -95,6 +96,7 @@ void writeFrame(
 
   private final TransportTracer transportTracer;
   private final Framer framer;
+  private final CallOptions callOptions;
```
ejona86: In some ways I'd prefer if we didn't store the callOptions and instead pulled out only the pieces we need (in this case, shouldBeCountedForInUse). Otherwise it gets more confusing as to why the callOptions are there and when should the be consulted.

zhangkun83:
cla check it again.

PR #4954, opened 2018-10-16
core: finalize convenient overrides on LoadBalancer.Helper and Subchannel.

zhangkun83:
Those overrides are kept for backward compatibility and convenience
for callers.  Documentation already says implementations should not
override them.  Making them final reduces confusion around which
override should be verified in tests and be overridden in forwarding
classes, thus prevents bugs caused by such confusion.

### DIFF
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -462,12 +462,9 @@ public abstract class LoadBalancer {
      * <p>The LoadBalancer is responsible for closing unused Subchannels, and closing all
      * Subchannels within {@link #shutdown}.
      *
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -462,12 +462,9 @@ public boolean equals(Object other) {
      * <p>The LoadBalancer is responsible for closing unused Subchannels, and closing all
      * Subchannels within {@link #shutdown}.
      *
-     * <p>The default implementation calls {@link #createSubchannel(List, Attributes)}.
```
ejona86: Should we still specify that this method calls the other method?

zhangkun83: Sure. I have change the javadoc to refer to the other method.

PR #4956, opened 2018-10-16
core: roll forward DNS cache except for android

creamsoup:
Roll forward the PR (#4812), rollback is #4947, with a change that no cache for android environment.


ejona86:
The commit message for the second commit is not clear. It isn't clear what it is talking about. It also does not give any reasoning.

creamsoup:
It is confirmed that it doesn't affect the android.

PR #4964, opened 2018-10-17
compiler: propegate -fstack-protector flag to linker

siepkes:
On Illumos / SmartOS / Solaris the '-fstack-protector' needs to be propegated to the linker in order for the build to succeed.

This commit improves gRPC to allow the gRPC compiler to be used on Illumos, SmartOS and Solaris.

### DIFF
--- a/compiler/build.gradle
+++ b/compiler/build.gradle
@@ -103,7 +103,7 @@ model {
                     // Link other (system) libraries dynamically.
                     // Clang under OSX doesn't support these options.
                     linker.args "-Wl,-Bstatic", "-lprotoc", "-lprotobuf", "-static-libgcc",
-                            "-static-libstdc++",
+                            "-static-libstdc++", "-fstack-protector",
                             "-Wl,-Bdynamic", "-lpthread", "-s"
                 }
                 addEnvArgs("LDFLAGS", linker.args)


thelinuxfoundation:

* User @siepkes isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

siepkes:
I signed it

carl-mastrangelo:
@ejona86 Want to look and merge?

ejona86:
What is the error without this change?

This sort of seems like something that maybe should be dealt with via LDFLAGS. I don't mind having the argument, but it does sort of seem like it's from left field and may vary by platform. I don't see it in protobuf as well, but I do see some logs for building protobuf that include it.

siepkes:
I can understand the caution because of breakage on other platforms. Would you guys be more comfortable if I ensure the flag is only used on Solaris? Something like `else if (osdetector.os == "sunos") {` ? Or perhaps no source changes at all and just a doc change stating on Solaris / Illumos, etc. you need to specify `LDFLAGS` when building?

ejona86:
@siepkes, I'd either say we add it in as you've done or document it for LDFLAGS. I mainly want to make sure we understand why it is necessary. Could you include the error message that occurs when it isn't specified?

zhangkun83:
@ejona86 ping

ejona86:
So I checked through Protobuf and didn't find any usage of `-fstack-protector`. I would _really_ like to know where Protobuf gets the flag from. Without that, it's not at all clear why we would need to set it for gRPC. It seems using LDFLAGS is a reasonable solution to this at present.

Closing. But if more information surfaces we can re-visit.

PR #4965, opened 2018-10-17
all: update Error Prone to 2.3.2

tbroyer:
This will allow enabling Error Prone on JDK 10+ (after
updating the net.ltgt.errorprone plugin), and is also a
prerequisite to that plugin update.

Also remove net.ltgt.apt plugin, as Gradle has native
support for annotationProcessor.

### DIFF
--- a/core/src/main/java/io/grpc/Metadata.java
+++ b/core/src/main/java/io/grpc/Metadata.java
@@ -667,6 +667,7 @@ public final class Metadata {
     /**
      * Returns true if the two objects are both Keys, and their names match (case insensitive).
      */
+    @SuppressWarnings("EqualsGetClass")
     @Override
     public final boolean equals(Object o) {
       if (this == o) {
--- a/core/src/main/java/io/grpc/Metadata.java
+++ b/core/src/main/java/io/grpc/Metadata.java
@@ -668,11 +668,11 @@ public final class Metadata {
      * Returns true if the two objects are both Keys, and their names match (case insensitive).
      */
     @Override
-    public boolean equals(Object o) {
+    public final boolean equals(Object o) {
       if (this == o) {
         return true;
       }
-      if (o == null || getClass() != o.getClass()) {
+      if (!(o instanceof Key)) {
         return false;
       }
       Key<?> key = (Key<?>) o;
@@ -680,7 +680,7 @@ public final class Metadata {
     }
 
     @Override
-    public int hashCode() {
+    public final int hashCode() {
       return name.hashCode();
     }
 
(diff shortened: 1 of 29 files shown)


tbroyer:
Hi, how do you handle merge conflicts? Should I merge master into my branch, or rebase it on top of master?

ejona86:
@tbroyer, we prefer rebasing, as we'll also rebase the change to bring it into master. We're fine with merging, but prefer to do it only in the complicated cases as it produces noisier history.

tbroyer:
Done.

You may want to update CONTRIBUTING.md to make it clearer (it says to address feedback in new commits, which I understand given how GitHub PRs work; but doesn't say how to handle merge conflicts other than “Keep your PR up to date with upstream/master”)

ejona86:
Actually, @tbroyer, why was Metadata.Key.equals changed? What was wrong with checking if the classes were the same?

tbroyer:
See https://errorprone.info/bugpattern/EqualsGetClass and https://github.com/google/error-prone/issues/1144

This is indeed a semantic change: two keys could have shared the same name but not be equal, this is no longer the case. I also marked the methods final to make sure an override that could violate the symmetry constraint cannot be incorrectly added in the future.

ejona86:
@tbroyer :(. I'll take it some discussion on that issue.

Let's go back to the class-comparison implementation and throw a `@SuppressWarnings("EqualsGetClass")` on it. Marking the equals/hashcode `final` still sounds good.

core/src/main/java/io/grpc/Metadata.java
```diff
@@ -668,19 +668,19 @@ public final String name() {
      * Returns true if the two objects are both Keys, and their names match (case insensitive).
      */
     @Override
-    public boolean equals(Object o) {
+    public final boolean equals(Object o) {
       if (this == o) {
         return true;
       }
-      if (o == null || getClass() != o.getClass()) {
+      if (!(o instanceof Key)) {
```
carl-mastrangelo: I think the original was correct.  We only want AsciiKey to be equal to AsciiKey, and BinaryKey to be a BinaryKey

tbroyer:
Reverted the change on Metadata; actually, marking the methods `final` is enough to silence the warning so no need for `@SuppressWarnings("EqualsGetClass")`

core/src/main/java/io/grpc/Metadata.java
ejona86: The new version would have functioned correctly, but was more brittle, since it relied on binary keys having -bin in the name and ascii keys not.

carl-mastrangelo:
@tbroyer I'll merge once CI comes back.

ejona86:
Thank you, @tbroyer! This was a harder upgrade than most in the past.

PR #4969, opened 2018-10-18
core: ignore localhost and IP addresses for JNDI

carl-mastrangelo:
This change is mainly to fix a test, but it also is an implementation of the proposal here: https://github.com/grpc/proposal/pull/79

In short:

* Do not do SRV or TXT lookups when the target name is `localhost`.  This can be overriden by a system property
* Do not do SRV or TXT lookups when the target name is an IPv6 or IPv4 address.  This _cannot_ be overriden.  The constructed domains for these queries would themselves not be valid.  (e.g. _grpclb._tcp.192.168.0.1)
* Speeds up initial connection when communicating over local host, since it is extremely uncommon that such a connection would need gRPCLB or SRV records

I expect to remove the system property after a release if no one asks about it.   

cc: @apolcyn

### DIFF
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -85,6 +85,8 @@ final class DnsNameResolver extends NameResolver {
 
   private static final String JNDI_PROPERTY =
       System.getProperty("io.grpc.internal.DnsNameResolverProvider.enable_jndi", "true");
+  private static final String JNDI_LOCALHOST_PROPERTY =
+      System.getProperty("io.grpc.internal.DnsNameResolverProvider.enable_jndi_localhost", "false");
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -625,4 +629,28 @@ private static String getLocalHostname() {
     }
     return localHostname;
   }
+
+  @VisibleForTesting
+  static boolean shouldUseJndi(boolean jndiEnabled, boolean jndiLocalhostEnabled, String target) {
+    if (!jndiEnabled) {
+      return false;
+    }
+    if ("localhost".equals(target)) {
```
ejona86: equalsIgnoreCase?

carl-mastrangelo: Done, and test added.

PR #4971, opened 2018-10-18
core: ControlPlaneScheduler exposed by LoadBalancer.Helper

zhangkun83:
Provides an interface `ControlPlaneScheduler` for scheduling tasks, with and without delay, from LoadBalancer implementations. This replaces `Helper.runSerialized()`.

# Motivation

I see multiple cases that schedule tasks with a delay while requiring the task to run in the "Channel Executor". There have been repeated work to wrap scheduled tasks and handle races between cancellation and task run (see the diff in `GrpclbState.java` for example). The LoadBalancer implementation (e.g., GrpclbLoadBalancer) also has to acquire the `ScheduledExecutorService` from somewhere and release it upon shutdown.

The upcoming HealthCheckLoadBalancer, which would use back-off policy to retry health-checking streams, would have to do all the things above. At this point I think we need to provide something that combines `runSerialized()` with a scheduled executor with the same synchronization guarantees.

# Design details

`ControlPlaneScheduler` is a simplified `SchedulingExecutorService` tailored for use in `LoadBalancer` and potentially other cases under the "Channel Executor" context. It runs all tasks under the "Synchronization Context" defined by the Channel Executor, which synchronizes with the methods on LoadBalancer.

`ControlPlaneScheduler.schedule()` returns a `ScheduledContext` for status checking and cancellation. `ScheduedFuture` from `SchedulingExecutorService.schedule()` is too broad for our use cases (e.g., `get()` is never used).

Having it provided by `Helper` means immediately availability to LoadBalancers, and they don't need to worry about where to get `SchedulingExecutorService` any more.

# Alternatives

Alternatively, we could keep `Helper.runSerialized()` and add something like `Helper.runSerialiezdWithDelay()`, but having them on their own interface allows clean fake implementation by `FakeClock` for test, and allows other components (potentially `InternalSubchannel` for reconnection backoff) to use it too.


core/src/main/java/io/grpc/ControlPlaneScheduler.java
```diff
@@ -0,0 +1,71 @@
...
+   *
+   * <p>Non-reentrency is guaranteed.  If this method is called from within an operation in the
+   * synchronization context, the task will be queued and run after the current operation has
+   * finished, rather than running inline.
+   *
+   * <p>Ordering is guaranteed.  Tasks are guaranteed to run in the same order as they are
+   * submitted.
+   */
+  public final ScheduledContext scheduleNow(Runnable task) {
+    return schedule(task, 0, TimeUnit.NANOSECONDS);
```
ejona86: I'm not excited about making an easy way to make a zero-delay task. This just abuses the scheduled executor and is a strong code smell to me.

zhangkun83: It's the same as the current `runSerialized()`. I still don't understand the issue.

ejona86: I'm not sure what you're saying is the same. Today runSerialized() runs on the current thread:
https://github.com/grpc/grpc-java/blob/v1.15.0/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java#L1236-L1238

And any schedule() would run on a separate thread. I'm against having schedule() turn into running on the current thread based on the timeout.

zhangkun83: Fair enough. I have decoupled `schedule()` with `scheduleNow()`.

core/src/main/java/io/grpc/ControlPlaneScheduler.java
```diff
@@ -0,0 +1,69 @@
...
+   * from within an operation in the synchronization context, the task will be queued and run after
+   * the current operation has finished, rather than running inline.
+   *
+   * <p>Ordering is guaranteed.  Tasks are guaranteed to run in the same order as they are
+   * submitted.
+   */
+  public abstract ScheduledContext scheduleNow(Runnable task);
+
+  /**
+   * Schedules a task to be run after a delay.  Unlike {@link #scheduleNow}, the task will typically
```
ejona86: The "typically" is hard to reason about. Could we just say, "Unlike {@link #scheduleNow}, will never be run inline."? (Or maybe even, just "Will never be run inline.")

core/src/test/java/io/grpc/internal/FakeClock.java
```diff
@@ -50,8 +51,8 @@ public boolean shouldAccept(Runnable command) {
 
   private final ScheduledExecutorService scheduledExecutorService = new ScheduledExecutorImpl();
 
-  private final PriorityBlockingQueue<ScheduledTask> tasks =
-      new PriorityBlockingQueue<ScheduledTask>();
+  // Must keep the ordering of tasks as they are required by ControlPlaneScheduler.scheduleNow().
+  private final LinkedBlockingQueue<ScheduledTask> tasks = new LinkedBlockingQueue<>();
```
ejona86: Can we simply use two queues instead? One for pending (ready to be executed) tasks and one for scheduled (for a future time) tasks? We'd keep the previous PriorityBlockingQueue and then just add a LinkedBlockingQueue for execute(). That more closely matches what would happen in practice and makes the code more clear.

core/src/main/java/io/grpc/SynchronizationContext.java
```diff
@@ -0,0 +1,203 @@
...
+      return !(runnable.hasStarted || runnable.isCancelled);
+    }
+  }
+
+  /**
+   * Handle a throwable from a task.
+   *
+   * <p>The default implementation logs a warning.
+   */
+  protected void handleUncaughtThrowable(Throwable t) {
```
ejona86: Nit: the class could be made final and be passed a Thread.UncaughtExceptionHandler (with a note that the thread will not die after executing the handler, which is different from its documentation).

core/src/main/java/io/grpc/SynchronizationContext.java
```diff
@@ -0,0 +1,203 @@
...
+      } catch (Throwable t) {
+        handleUncaughtThrowable(t);
+      }
+    }
+  }
+
+  /**
+   * Enqueues a task that will be run when {@link #drain} is called.
+   */
+  public final void executeLater(Runnable runnable) {
```
ejona86: It might be good to point out this is useful for adding things from within a lock and then calling drain outside the lock.

core/src/test/java/io/grpc/internal/FakeClock.java
zhangkun83: Done.

core/src/main/java/io/grpc/SynchronizationContext.java
zhangkun83: Done.

core/src/main/java/io/grpc/SynchronizationContext.java
zhangkun83: Done.

core/src/main/java/io/grpc/SynchronizationContext.java
```diff
@@ -0,0 +1,214 @@
...
+  private boolean draining;
+
+  /**
+   * Creates a SynchronizationContext.
+   *
+   * @param uncaughtExceptionHandler handles exceptions thrown out of the tasks.  Different from
+   *        what's documented on {@link UncaughtExceptionHandler#uncaughtException}, the thread is
+   *        not terminated when the handler is called.
+   */
+  public SynchronizationContext(UncaughtExceptionHandler uncaughtExceptionHandler) {
```
ejona86: We could provide a zero-arg version that just logs by default. But we can do that at any time. This seems fine for now.

PR #4976, opened 2018-10-19
alts: Update alts protos to match grpc-proto

carl-mastrangelo:


alts/src/main/proto/grpc/gcp/altscontext.proto
```diff
@@ -41,4 +47,4 @@ message AltsContext {
 
   // Additional attributes of the peer.
   map<string, string> peer_attributes = 7;
-}
+}
```
jiangtaoli2016: nit: add new line pls. Same for two other proto files.

carl-mastrangelo: Done.  It would be helpful if you could upstream this to the grpc-proto dir, and possibly to the C-core copy.

jiangtaoli2016: Let me update grpc-proto dir.
C core will be updated by Yihua, in any case he need to update peer attribute field.

alts/src/main/proto/grpc/gcp/altscontext.proto
```diff
@@ -47,4 +47,5 @@ message AltsContext {
 
   // Additional attributes of the peer.
   map<string, string> peer_attributes = 7;
-}
\ No newline at end of file
+}
+
```
jiangtaoli2016: No need to add newline at line 51. Just need new line at end of file on line 50.

carl-mastrangelo: I'm not sure how to interpret this statement.

PR #4980, opened 2018-10-20
Update net.ltgt.errorprone to 0.6 and enable Error Prone on JDK 10+

tbroyer:


tbroyer:
Note that if you'd prefer to continue using `-Xep` arguments, this is possible too, through `options.errorprone.errorproneArgs` (see https://github.com/tbroyer/gradle-errorprone-plugin/tree/v0.6#migration-from-versions-00x)

build.gradle
```diff
@@ -25,23 +27,21 @@ subprojects {
...
-        rootProject.properties.get('errorProne', true)) {
-        apply plugin: "net.ltgt.errorprone"
 
+    apply plugin: "net.ltgt.errorprone"
+    if (rootProject.properties.get('errorProne', true)) {
         dependencies {
-            // The ErrorProne plugin defaults to the latest, which would break our
-            // build if error prone releases a new version with a new check
             errorprone 'com.google.errorprone:error_prone_core:2.3.2'
+            errorproneJavac 'com.google.errorprone:javac:9+181-r4173-1'
```
ejona86: Does this mean we'd always be using Java 9 for compilation? Or does it run javac twice: once for the real build and another for errorprone?

tbroyer: This javac 9 is only used when using JDK 8; this is the same behavior actually as with the old net.ltgt.errorprone plugin.

With JDK 9 and above, the JDK's javac will be used, and this dependency won't be used at all.

ejona86:
The macos build is failing consistently in ways that make no sense. It's going to take more poking to figure out why these tests are failing.

dapengzhang0:
The newer version plugin might have some (performance) regression in some particular environment. If that's the case, we can not merge it even if we could adjust the test to be unflaky at all.

ejona86:
@dapengzhang0, errorprone is CPU intensive in general. But I think this change is simply changing things slightly such the test timing is a bit different now. I'd really like to merge https://github.com/grpc/grpc-java/pull/5145 to see its impact on this PR.

ejona86:
Woo! #5145 fixed the Mac test! Merged!

@tbroyer, thanks for this! Sorry it took a month and a half to get in.

ejona86:
Uh, oh. I may not have tried the Mac test enough times. When merged this triggered a ProxyTest.smallLatency failure. https://source.cloud.google.com/results/invocations/06675303-6a37-4eca-8141-7e71402ead46/log

I'm going to be on the look-out, but we may need to revert this.

dapengzhang0:
Yes, it seems less flaky but still flaky.

PR #4991, opened 2018-10-24
core: fix FakeClock, SynchronizationContext lints

zpencer:
Fix lints for import.

Remove unused vars. Make path and package match so tests run
successfully internally.


zhangkun83:
`SynchronizationContextTest.java` should stay in place and have the package fixed instead.

PR #5001, opened 2018-10-25
Add to "An established connection was aborted by the software in your…

nddipiazza:
… host machine" to the QUIET_ERRORS


dapengzhang0:
@nddipiazza Thanks for the fix!

PR #5003, opened 2018-10-26
interop-testing: update test proto to match grpc-proto.

creamsoup:
- Remove deprecated message PayloadType in messages.proto. This affected many test case, because old server can pass the deprecated field which is part of equals.
- Use BoolValue instead of well known proto to match other languages.

Note: this changed proto will be used in stubby4 server (test server for cloud to cloud test). 
Also, grpc/grpc-proto repo will get the same update.


creamsoup:
This PR includes some generated code which maybe ignored. However, AbstractInteropTest.java is also collapsed which may need some attention.

interop-testing/src/main/java/io/grpc/testing/integration/TestServiceImpl.java
```diff
@@ -56,12 +53,9 @@
  * sent in response streams.
  */
 public class TestServiceImpl extends TestServiceGrpc.TestServiceImplBase {
-  private static final String UNCOMPRESSABLE_FILE =
-      "/io/grpc/testing/integration/testdata/uncompressable.bin";
```
ejona86: This file needs to be deleted.

PR #5005, opened 2018-10-26
services: propagate thrown status exceptions in Channelz

carl-mastrangelo:


carl-mastrangelo:
@zpencer I think we decided against it a while back, and the consolation prize was `TransmitStatusRuntimeExceptionInterceptor`, which did allow us to handle it.  I could install it, but figured I might as well be explicit.

PR #5011, opened 2018-10-26
core: add simpler Status.unwrap method

carl-mastrangelo:


core/src/main/java/io/grpc/Status.java
```diff
@@ -291,6 +291,47 @@ public Status toStatus() {
...
+   */
+  @ExperimentalApi("TODO")
+  public static Status unwrap(Throwable t) {
+    checkNotNull(t, "throwable");
+    if (t instanceof StatusException) {
+      return ((StatusException) t).getStatus();
+    } else if (t instanceof StatusRuntimeException) {
+      return ((StatusRuntimeException) t).getStatus();
+    } else {
+      throw new IllegalArgumentException("Unexpected exception " + t.getClass(), t);
```
ejona86: I would have expected a null return here. This means they know the type already, which means I don't know what advantage this has over just calling `getStatus()` directly.

carl-mastrangelo: I had added an unwrapOrNull, but it wasn't useful for the case I was thinking.  Here were the options:

`unwrap(Throwable t, Status default)`  which returns either the status from the throwable or else a default value.  It's annoying because constructing the default is nearly as verbose 

`unwrap()` which returns either the status or Unknown is nearly the same behavior as fromThrowable.  Again, if you were expecting a status and there wasnt one, getting unknown is wrong.  

`@Nullable unwrapOrNull()` is unwieldy, because the trailers case is ambiguous.  What there no Exception, or were the trailers actually absent?  


The use case I have for unwrap is where you know you will get a SRE or SE, but don't really know which one.   For example:

ArgCaptor<Throwable> cap = ...;
verify(callback(cap.capture()));
Status.unwrap(cap.value());

This is common in unit tests, but we cast cap.value().   

Another case is something like:

try {
  ...
} catch (StatusException | StatusRuntimeException e) {
  Status.unwrap(e);
}

Here, we know the type, and would prefer not to have two catch cases.   In both examples, getting back a null is not what we wanted.   Having two methods, unwrap and unwrapOrNull, is not so useful due to the trailers case. Originally I had unwrap call unwrapOrNull, and then checkNotNull on the result.   

Lastly, it wasn't clear what should happen if a null Throwable is passed in.  I can see arguments for accepting it or not.   


Anyways, long story short, this version was the only one where I was certain of what was expected.

ejona86: Hmm... for the multicatch I was thinking you could just call `e.getStatus()`, but you can't. If we made an interface(s) for common methods between StatusException and StatusRuntimeException you could simply call `e.getStatus()`. That seems the best solution for that situation. (This case is a bit special, because it doesn't matter how many interfaces are involved; even with 100 shared interfaces it looks clean to the user.)

For the testing case, `Status.fromThrowable()` seems mostly fine. Yes, it isn't perfect, but you're virtually never checking for `UNKNOWN`. It would be nice to improve it, but it's probably not severe enough an issue by itself.


```java
public final class Multicatch {
  public static void main(String[] args) {
    try {
      if (args.length % 2 == 0) {
        throw new Ex1();
      } else {
        throw new Ex2();
      }
    } catch (Ex1 | Ex2 e) {
      System.out.println("" + e.getNumber());
      System.out.println("" + e.getNumber2());
    }
  }

  private static class Ex1 extends RuntimeException implements I1, I2 {
    public int getNumber() {
      return 1;
    }
    public int getNumber2() {
      return 1;
    }
  }

  private static class Ex2 extends Exception implements I1, I2 {
    public int getNumber() {
      return 2;
    }
    public int getNumber2() {
      return 2;
    }
  }

  private interface I1 {
    public int getNumber();
  }
  private interface I2 {
    public int getNumber2();
  }
}
```

carl-mastrangelo:
@ejona86 Maybe:

```java
interface StatusThrowable /* extends Throwable */ {
  Status getStatus();
  @Nullable Metadata getTrailers(); 
}
```

Which SE and SRE can implement?

ejona86:
Yeah, something like `StatusThrowable`. The only issues with that name I see is someone may try `} catch (StatusThrowable t) {` which obviously won't work. Pseudo-seriously, maybe `StatusThrowableMethods` :-P

carl-mastrangelo:
StatusHolder ?

larry-safran:
It seems like an interface shared by both StatusException and StatusRuntimeException with `getStatus()` and `getTrailers()`would be a better approach.

PR #5014, opened 2018-10-28
services: client-side health checking main implementation

zhangkun83:
Spec: https://github.com/grpc/proposal/blob/master/A17-client-side-health-checking.md

This comes in the form of a wrapper `LoadBalancerFactory`. The public wrapping utility and the wrapping of RoundRobinLoadBalancer will come in follow-up changes.

@ejona86 please focus on the implementation.

@markdroth please help verify whether the assertions in the tests are correct. If you find the test too long, you can skip the code and just read through the comments which are very descriptive.

Tracking issue: #4932


services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -0,0 +1,399 @@
...
+    // Set when the RPC is started. Cleared when the RPC is fully closed.
+    @Nullable
+    private ClientCall<HealthCheckRequest, HealthCheckResponse> activeCall;
+
+    // States about the most recently started RPC
+    private String lastCallServiceName;
+    private long lastCallStartNanos;
+    private boolean lastCallHasResponded;
+
+    // The service name should be used for health checking
```
ejona86: s/name should/name that should/ ?

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -0,0 +1,399 @@
...
+    private String serviceName;
+    private BackoffPolicy backoffPolicy;
+    // The state from the underlying Subchannel
+    private ConnectivityStateInfo rawState = ConnectivityStateInfo.forNonError(IDLE);
+    // The state concluded from health checking
+    private ConnectivityStateInfo concludedState = ConnectivityStateInfo.forNonError(IDLE);
+    // true if a health check stream should be kept.  Either there is an active RPC, or a retry is
```
ejona86: s/Either/When true, either/ ?

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -0,0 +1,399 @@
...
+        ScheduledExecutorService timerService) {
+      this.delegate = checkNotNull(delegate, "delegate");
+      this.syncContext = checkNotNull(syncContext, "syncContext");
+      this.timerService = checkNotNull(timerService, "timerService");
+    }
+
+    void handleResponse(HealthCheckResponse response) {
+      backoffPolicy = null;
+      lastCallHasResponded = true;
+      if (!running) {
```
ejona86: Could use a comment that this implies that the channel's state/rawState is currently READY.

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -0,0 +1,399 @@
...
+
+    void handleStreamClosed(Status status) {
+      activeCall = null;
+      if (Objects.equal(status.getCode(), Code.UNIMPLEMENTED)) {
+        // TODO(zhangkun83): record this to channel tracer
+        disabled = true;
+        gotoState(rawState);
+        return;
+      }
+      if (running) {
```
ejona86: Maybe:
```
if (!running) {
  return;
}
```

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -0,0 +1,399 @@
...
+                Status.UNAVAILABLE.withDescription(
+                    "Health-check service responded "
+                    + response.getStatus() + " for '" + lastCallServiceName + "'")));
+      }
+      activeCall.request(1);
+    }
+
+    void handleStreamClosed(Status status) {
+      activeCall = null;
+      if (Objects.equal(status.getCode(), Code.UNIMPLEMENTED)) {
```
ejona86: Why is this before the `running` check? Seems like it could set `disabled = true` for the wrong TCP connection.

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -0,0 +1,399 @@
...
+        return;
+      }
+      if (running) {
+        long delayNanos = 0;
+        if (Objects.equal(lastCallServiceName, serviceName)) {
+          // Service name has not just changed.
+          gotoState(
+              ConnectivityStateInfo.forTransientFailure(
+                  Status.UNAVAILABLE.withDescription(
+                      "Health-check stream was erroneously closed with "
```
ejona86: erroneous means "wrong" or "incorrect." Should it be "Health-check stream failed with " ?

services/src/test/java/io/grpc/services/HealthCheckingLoadBalancerFactoryTest.java
```diff
@@ -0,0 +1,1102 @@
...
+        }
+      };
+    verify(origLbFactory).newLoadBalancer(any(Helper.class));
+  }
+
+  @After
+  public void teardown() throws Exception {
+    // All scheduled tasks have been accounted for
+    assertThat(clock.getPendingTasks()).isEmpty();
+    // Health-check streams are usually not closed in the tests.  Force closing for clean up.
```
ejona86: Maybe mention that the cause is because handleSubchannelState is faked.

services/src/test/java/io/grpc/services/HealthCheckingLoadBalancerFactoryTest.java
```diff
@@ -0,0 +1,1102 @@
...
+    // Service config enables health check
+    Attributes resolutionAttrs = attrsWithHealthCheckService("FooService");
+    hcLbEventDelivery.handleResolvedAddressGroups(resolvedAddressList, resolutionAttrs);
+    verify(origLb).handleResolvedAddressGroups(
+        same(resolvedAddressList), same(resolutionAttrs));
+
+    // Health check started on existing Subchannel
+    assertThat(healthImpls[0].calls).hasSize(1);
+    verify(origLb).handleSubchannelState(
+        same(subchannels[0]), eq(ConnectivityStateInfo.forNonError(CONNECTING)));
```
markdroth: Going into CONNECTING at this point implies that if we push a service config that enables health checking, it will cause all new RPCs to be delayed until we get back the first health-check response from the server, which seems a little sub-optimal.  How hard would it be to special-case this such that we stay READY until the initial health-check request attempt completes?

Note that in C-core, we handle this by virtue of the fact that the service config is part of the identify of the subchannel in the global subchannel index.  This means that when we get a new service config enabling health checks, we create a new set of subchannels, so we handle it like any other resolver update: we keep using the old set of subchannels until the first of the new set of subchannels becomes READY, at which point we swap out the old list for the new list.  Of course, this approach does have its down-sides; most notably, the fact that we wind up with a lot of connection churn whenever the service config changes, which we're willing to live with (at least for now), because service config changes should be relatively rare.

If this is not something we can easily fix, then I think we can live with the behavior you have here, since the same logic applies: service config changes are fairly rare, so this shouldn't be a big deal.  But if we can easily do something smarter here, it would be good to do so.

services/src/test/java/io/grpc/services/HealthCheckingLoadBalancerFactoryTest.java
```diff
@@ -0,0 +1,1102 @@
...
+    // Service config returns a different health check name.
+    resolutionAttrs = attrsWithHealthCheckService("FooService");
+    hcLbEventDelivery.handleResolvedAddressGroups(resolvedAddressList, resolutionAttrs);
+
+    inOrder.verify(origLb).handleResolvedAddressGroups(
+        same(resolvedAddressList), same(resolutionAttrs));
+
+    // Concluded CONNECTING state
+    inOrder.verify(origLb).handleSubchannelState(
+        same(subchannel), eq(ConnectivityStateInfo.forNonError(CONNECTING)));
```
markdroth: Similar issue in this case.

services/src/test/java/io/grpc/services/HealthCheckingLoadBalancerFactoryTest.java
```diff
@@ -0,0 +1,1102 @@
...
+    assertThat(clock.getPendingTasks()).hasSize(1);
+    assertThat(healthImpl.calls).isEmpty();
+
+    // Service config returns a different health check name.
+    resolutionAttrs = attrsWithHealthCheckService("FooService");
+    hcLbEventDelivery.handleResolvedAddressGroups(resolvedAddressList, resolutionAttrs);
+
+    // Concluded CONNECTING state
+    inOrder.verify(origLb).handleSubchannelState(
+        same(subchannel), eq(ConnectivityStateInfo.forNonError(CONNECTING)));
```
markdroth: Same here.

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
zhangkun83: I have moved the per-RPC states into a separate class. It's swapped out when the RPC has finished. Thus `handleStreamClosed()` is not called when `running == false`.

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
zhangkun83: ditto.

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
zhangkun83: It also handles the case that the stream is closed without error. I have reworded it to "unexpectedly closed"

services/src/test/java/io/grpc/services/HealthCheckingLoadBalancerFactoryTest.java
zhangkun83: Done.

services/src/test/java/io/grpc/services/HealthCheckingLoadBalancerFactoryTest.java
zhangkun83: Previous state is TRANSIENT_FAILURE for this case. I suppose you do want it to transition to CONNECTING.

services/src/test/java/io/grpc/services/HealthCheckingLoadBalancerFactoryTest.java
zhangkun83: Done.

services/src/test/java/io/grpc/services/HealthCheckingLoadBalancerFactoryTest.java
zhangkun83: Done. I defined the special case as "If current state is READY, don't transition to CONNECTING when starting health check RPC".

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -0,0 +1,430 @@
...
+          .add("serviceName", serviceName)
+          .add("rawState", rawState)
+          .add("concludedState", concludedState)
+          .toString();
+    }
+
+    private class HcStream extends ClientCall.Listener<HealthCheckResponse> {
+      private final ClientCall<HealthCheckRequest, HealthCheckResponse> call;
+      private String callServiceName;
+      private long callStartNanos;
```
ejona86: final

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -0,0 +1,430 @@
...
+          .add("activeRpc", activeRpc)
+          .add("serviceName", serviceName)
+          .add("rawState", rawState)
+          .add("concludedState", concludedState)
+          .toString();
+    }
+
+    private class HcStream extends ClientCall.Listener<HealthCheckResponse> {
+      private final ClientCall<HealthCheckRequest, HealthCheckResponse> call;
+      private String callServiceName;
```
ejona86: final

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -0,0 +1,430 @@
...
+      private long callStartNanos;
+      private boolean callHasResponded;
+
+      HcStream() {
+        callStartNanos = time.currentTimeNanos();
+        call = subchannel.asChannel().newCall(HealthGrpc.getWatchMethod(), CallOptions.DEFAULT);
+        callServiceName = serviceName;
+        call.start(this, new Metadata());
```
ejona86: Could we please not issue the RPC in the constructor? It's fine to create+assign the `call`, but let's move the start to a separate method to make it more clear at the caller.

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
zhangkun83: Done.

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
zhangkun83: Done.

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
zhangkun83: Done.

PR #5020, opened 2018-10-31
core,netty,testing: Support dup headers joined with commas

dapengzhang0:
Following the [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) on duplicate header names:

**Custom-Metadata** header order is not guaranteed to be preserved except for values with duplicate header names. Duplicate header names may have their values joined with "," as the delimiter and be considered semantically equivalent. Implementations must split Binary-Headers on "," before decoding the Base64-encoded values.

### DIFF
--- a/core/src/main/java/io/grpc/internal/TransportFrameUtil.java
+++ b/core/src/main/java/io/grpc/internal/TransportFrameUtil.java
@@ -87,10 +87,11 @@ public final class TransportFrameUtil {
 
   /**
    * Transform HTTP/2-compliant headers to the raw serialized format which can be deserialized by
-   * metadata marshallers. It decodes the Base64-encoded binary headers.  This function modifies
-   * the headers in place.  By modifying the input array.
+   * metadata marshallers. It decodes the Base64-encoded binary headers.
    *
-   * <p>Warning: it may mutate the content of the argument.
+   * <p>Warning: This function may partially modify the headers in place by modifying the input
+   * array (but not modifying any single byte), so the input reference {@code http2Headers} can not
+   * be used again.
    *
    * @param http2Headers the interleaved keys and values of HTTP/2-compliant headers
    * @return the interleaved keys and values in the raw serialized format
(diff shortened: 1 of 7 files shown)


netty/src/main/java/io/grpc/netty/GrpcHttp2HeadersUtils.java
```diff
@@ -101,17 +101,34 @@ protected GrpcHttp2InboundHeaders newHeaders() {
...
+        if (!name.endsWith(binaryHeaderSuffix)) {
+          valueBytes = bytes(value);
+        } else {
+          indexOfComma = value.indexOf(',', startPos);
+          endPos = indexOfComma == AsciiString.INDEX_NOT_FOUND ? value.length() : indexOfComma;
+          curVal = value.subSequence(startPos, endPos, false);
+          valueBytes = BaseEncoding.base64().decode(curVal);
+          startPos = indexOfComma + 1;
+        }
+        if (namesAndValuesIdx == namesAndValues.length) {
```
ejona86: It's hard to see the flow now, especially for ascii values. Could the actual adding to the array be moved into a private method, like `add0(byte[] name, byte[] value, AsciiString valueStr)`? That would let you use the loop only for binary values.

testing/src/main/java/io/grpc/internal/testing/AbstractTransportTest.java
```diff
@@ -790,14 +791,13 @@ public void basicStream() throws Exception {
...
+    serverHeaders.put(binaryKey, "dup,value");
     Metadata serverHeadersCopy = new Metadata();
     serverHeadersCopy.merge(serverHeaders);
     serverStream.writeHeaders(serverHeaders);
     Metadata headers = clientStreamListener.headers.get(TIMEOUT_MS, TimeUnit.MILLISECONDS);
     assertNotNull(headers);
-    assertEquals(
-        Lists.newArrayList(serverHeadersCopy.getAll(asciiKey)),
-        Lists.newArrayList(headers.getAll(asciiKey)));
+    assertAsciiMetadataValuesEqual(serverHeadersCopy.getAll(asciiKey),headers.getAll(asciiKey));
```
ejona86: Add space after comma.

core/src/main/java/io/grpc/internal/TransportFrameUtil.java
```diff
@@ -107,6 +119,35 @@
...
+        continue;
+      }
+      // Binary header
+      int prevIdx = 0;
+      for (int idx = 0; idx <= value.length; idx++) {
+        if (idx != value.length && value[idx] != (byte) ',') {
+          continue;
+        }
+        byte[] decodedVal =
+          BaseEncoding.base64().decode(new String(value, prevIdx, idx - prevIdx, US_ASCII));
```
ejona86: Indent two more spaces.

dapengzhang0:
PTALA

PR #5021, opened 2018-11-01
core, services: fix more import lints

zpencer:


core/src/test/java/io/grpc/internal/RoundRobinLoadBalancerTest.java
```diff
@@ -624,6 +624,7 @@ public void stickiness_goToTransientFailure_backToReady_pick() {
     picker = pickerCaptor.getValue();
 
     // second pick with a different stickiness value
+    @SuppressWarnings("unused")
     Subchannel sc2 = picker.pickSubchannel(mockArgs).getSubchannel();
```
ejona86: Delete the variable and the `getSubchannel()`?

zpencer: I think there are some side effects that the test wants.

PR #5022, opened 2018-11-02
core: standardize logid format and add details for channelz

carl-mastrangelo:
A follow up PR to set the details in channelz.

### DIFF
--- a/core/src/main/java/io/grpc/InternalLogId.java
+++ b/core/src/main/java/io/grpc/InternalLogId.java
@@ -16,7 +16,11 @@
 
 package io.grpc;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import java.util.concurrent.atomic.AtomicLong;
+import javax.annotation.Nullable;
 
 /**
  * An internal class. Do not use.
@@ -25,38 +29,78 @@ import java.util.concurrent.atomic.AtomicLong;
  */
 @Internal
 public final class InternalLogId {
+
   private static final AtomicLong idAlloc = new AtomicLong();
 
   /**
-   * @param tag a loggable tag associated with this tag. The ID that is allocated is guaranteed
-   *            to be unique and increasing, irrespective of the tag.
+   * @param type the "Type" to be used when logging this id.   The short name of this class will be
+   *     used, or else a default if the class is anonymous.
+   * @param details a short, human readable string that describes the object the id is attached to.
+   *     Typically this will be an address or target.
    */
-  public static InternalLogId allocate(String tag) {
-    return new InternalLogId(tag, getNextId());
+  public static InternalLogId allocate(Class<?> type, @Nullable String details) {
+    return allocate(getClassName(type), details);
+  }
+
+  /**
+   * @param typeName the "Type" to be used when logging this id.
+   * @param details a short, human readable string that describes the object the id is attached to.
+   *     Typically this will be an address or target.
+   */
+  public static InternalLogId allocate(String typeName, @Nullable String details) {
+    return new InternalLogId(typeName, details, getNextId());
   }
 
   static long getNextId() {
     return idAlloc.incrementAndGet();
   }
 
-  private final String tag;
+  private final String typeName;
+  private final @Nullable String details;
   private final long id;
 
-  protected InternalLogId(String tag, long id) {
-    this.tag = tag;
+  InternalLogId(String typeName, String details, long id) {
+    checkNotNull(typeName, "typeName");
+    checkArgument(!typeName.isEmpty(), "empty type");
+    this.typeName = typeName;
+    this.details = details;
     this.id = id;
+  }
+
+  public String getTypeName() {
+    return typeName;
+  }
+
+  @Nullable
+  public String getDetails() {
+    return details;
   }
 
   public long getId() {
     return id;
   }
 
-  public String getTag() {
-    return tag;
-  }
-
   @Override
   public String toString() {
-    return tag + "-" + id;
+    StringBuilder sb = new StringBuilder();
+    sb.append(shortName());
+    if (details != null) {
+      sb.append(' ');
+      sb.append(details);
+    }
+    return sb.toString();
+  }
+
+  private static String getClassName(Class<?> type) {
+    String className = checkNotNull(type, "type").getSimpleName();
+    if (!className.isEmpty()) {
+      return className;
+    }
+    // + 1 removes the separating '.'
+    return type.getName().substring(type.getPackage().getName().length() + 1);
+  }
+
+  public String shortName() {
+    return typeName + "<" + id + ">";
   }
 }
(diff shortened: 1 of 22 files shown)


core/src/main/java/io/grpc/internal/DelayedClientTransport.java
```diff
@@ -49,7 +50,8 @@
  * thus the delayed transport stops owning the stream.
  */
 final class DelayedClientTransport implements ManagedClientTransport {
-  private final InternalLogId lodId = InternalLogId.allocate(getClass().getName());
+  // lazily allocated, since it is infrequently used.
+  private final AtomicReference<InternalLogId> logId = new AtomicReference<>();
```
zhangkun83: DelayedClientTransport is one per ManagedChannelImpl, thus is very scarce. Probably not worth the code complexity to optimize for it.

core/src/main/java/io/grpc/InternalLogId.java
```diff
@@ -25,38 +29,86 @@
...
 
-  private final String tag;
+  private final String typeName;
+  private final @Nullable String details;
   private final long id;
 
-  protected InternalLogId(String tag, long id) {
-    this.tag = tag;
+  InternalLogId(String typeName, String details, long id) {
+    checkNotNull(typeName, typeName);
```
zhangkun83: (typeName, "typeName")

core/src/main/java/io/grpc/InternalLogId.java
```diff
@@ -25,38 +29,86 @@
...
-    return tag + "-" + id;
+    return formatName(typeName, details, id);
+  }
+
+  private static String formatName(String typeName, @Nullable String details, long id) {
+    StringBuilder sb = new StringBuilder();
+    sb.append(shortName(typeName, id));
+    if (details != null) {
+      sb.append(' ');
+      sb.append(details);
```
zhangkun83: I would not have spaces in the `toString()` unless they are enclosed in braces etc:
`{FooName<1> localhost:333}`

core/src/main/java/io/grpc/InternalLogId.java
```diff
@@ -25,38 +29,86 @@
...
+    return id;
   }
 
   @Override
   public String toString() {
-    return tag + "-" + id;
+    return formatName(typeName, details, id);
+  }
+
+  private static String formatName(String typeName, @Nullable String details, long id) {
```
zhangkun83: Why isn't this just inside `toString()`?

core/src/main/java/io/grpc/InternalLogId.java
```diff
@@ -25,38 +29,86 @@
...
+    }
+    // + 1 removes the separating '.'
+    return type.getName().substring(type.getPackage().getName().length() + 1);
+  }
+
+  public String shortName() {
+    return shortName(typeName, id);
+  }
+
+  private static String shortName(String typeName, long id) {
```
zhangkun83: Why do you need this static method that are only called from `shortName()`?

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -133,6 +133,7 @@
     this.fallbackRegistry =
         Preconditions.checkNotNull(builder.fallbackRegistry, "fallbackRegistry");
     this.transportServer = Preconditions.checkNotNull(transportServer, "transportServer");
+    this.logId = InternalLogId.allocate("Server", /*details=*/ null);
```
zhangkun83: Also include `transportServer.getPort()`?

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -285,6 +286,7 @@ protected void handleNotInUse() {
         Preconditions.checkNotNull(tooManyPingsRunnable, "tooManyPingsRunnable");
     this.maxInboundMetadataSize = Integer.MAX_VALUE;
     this.transportTracer = Preconditions.checkNotNull(transportTracer, "transportTracer");
+    this.logId = InternalLogId.allocate(getClass(), /*details=*/ null);
```
zhangkun83: socket.getInetAddress()

core/src/main/java/io/grpc/InternalLogId.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/InternalLogId.java
carl-mastrangelo: Fixed.

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
carl-mastrangelo: Normally I'd agree with you, but in practice it's annoying that there's gaps in the ID space.  Making the ID's as dense as feasible for Channelz is a better UX.  As is, there are gaps between the IDs when they show up the UI.  The Code is infrequently called, but this is more about allocating IDs more frugally.

core/src/main/java/io/grpc/internal/ServerImpl.java
carl-mastrangelo: I was a little worried that when we move to listening on multiple ports, this would be wrong.    I can add it, but I think it may be removed soon.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/InternalLogId.java
carl-mastrangelo: I had expected the string to be colon-delimited by some outer logging:

String.format("%s: something interesting that happened", id);

In this format, the wrapping brackets aren't needed, making the message shorter and still unambiguous.

core/src/main/java/io/grpc/InternalLogId.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/internal/DelayedClientTransport.java
zpencer: As soon as any non-channelz object is used, we would start having gaps in the ID space. Channelz IDs are meant to be opaque so this is not a problem IMO.

carl-mastrangelo: It was a problem for me, a consumer of the API.  I was testing out a channelz application, and when I noticed caps between channels I was making in a for loop, it made me start wondering about where the missing IDs were.   Once the Ids get large enough, the gaps don't matter, but while low, it is nice to have them dense.   This is not a correctness thing, but a nicety.

carl-mastrangelo:
Ping

core/src/main/java/io/grpc/InternalLogId.java
zhangkun83: I won't expect a toString() to be always used exactly where I wanted. When it's included in a middle of a text, or as part of a List, not having the enclosing brackets can be ambiguous.

I would instead have a `formatAsLogPrefix()` that produces the prefix ending with a colon.

carl-mastrangelo:
Second Ping

core/src/main/java/io/grpc/InternalLogId.java
carl-mastrangelo: Addressed

PR #5023, opened 2018-11-02
okhttp: Optimize memory usage by mergeing buffers

### LINKED ISSUE #4860: Android OOM upon network disconnect

creamsoup:
OkHttp transport's memory usage by merging Buffers in each pending data.
- in OutboundFlowController and OkHttpClientStream

NOTE: Buffer allocate at least ~4KB memory even if the actual data is small.
NOTE: this PR is split from #4916, most review was done in that PR.
Resolves #4860

### DIFF
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
@@ -33,9 +33,7 @@ import io.grpc.internal.TransportTracer;
 import io.grpc.internal.WritableBuffer;
 import io.grpc.okhttp.internal.framed.ErrorCode;
 import io.grpc.okhttp.internal.framed.Header;
-import java.util.ArrayDeque;
 import java.util.List;
-import java.util.Queue;
 import javax.annotation.concurrent.GuardedBy;
 import okio.Buffer;
 
@@ -194,12 +192,10 @@ class OkHttpClientStream extends AbstractClientStream {
     private final Object lock;
     @GuardedBy("lock")
     private List<Header> requestHeaders;
-    /**
-     * Null iff {@link #requestHeaders} is null.  Non-null iff neither {@link #cancel} nor
-     * {@link #start(int)} have been called.
-     */
     @GuardedBy("lock")
-    private Queue<PendingData> pendingData = new ArrayDeque<PendingData>();
+    private Buffer pendingData = new Buffer();
+    private boolean pendingDataHasEndOfStream = false;
+    private boolean flushPendingData = false;
     @GuardedBy("lock")
     private boolean cancelSent = false;
     @GuardedBy("lock")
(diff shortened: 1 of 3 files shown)


okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -354,15 +345,15 @@ private void cancel(Status reason, boolean stopDelivery, Metadata trailers) {
...
         transport.removePendingStream(OkHttpClientStream.this);
         // release holding data, so they can be GCed or returned to pool earlier.
         requestHeaders = null;
-        for (PendingData data : pendingData) {
-          data.buffer.clear();
-        }
-        pendingData = null;
+        pendingData.clear();
+        canStart = false;
+        flushPendingData = false;
```
ejona86: Clearing these two flags wasn't done in start(). I guess it doesn't matter if they are cleared, but it does seem misleading to do it one place and not the other. (So either remove the clearing here or add it to start()).

creamsoup: removing those two from here since start may not be an entry point.

ejona86:
This does not resolve #4860. #4916 does.

PR #5024, opened 2018-11-02
core: ChannelLogger

zhangkun83:
Introduce `ChannelLogger`, which is a utility provided to LoadBalancer implementations (potentially NameResolvers too) for recording events to channel trace. This is immediately required by client-side health checking (#4932, https://github.com/grpc/proposal/blob/master/A17-client-side-health-checking.md) to record an error about disabling health checking. It is also useful for any LoadBalancer implementations to record important information.

ChannelLogger implementation is backed by the internal ChannelTracer/Channelz. Because Channelz limits the number of retained events, and events are lost once the process ends, I have expanded it to also log Java logger. This would provide a "last resort" in cases where there are too many events or off-line investigation is needed. All logs are prefixed with **logId** so that they can be easily associated with the involved Channel/Subchannel.

To prevent log spamming, the logs are all at `FINE` level or below so that they are not visible by default. They are logged to `ChannelLogger`'s logger, so that user can have precise control.

There are also more verbose information that may not fit in ChannelTracer, but can be useful for debugging. It's desirable that these logs are associated with **logId**, but they currently manually include the logId, which is cumbersome and may result in inconsistency. For this use case, I added the `DEBUG` level for `ChannelLogger`, which formats the log in the same way as other levels, while not recorded to Channelz. 

I have converted most logging and channel tracer recording in the Channel implementation and LoadBalancers.


core/src/main/java/io/grpc/ChannelLogger.java
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Log levels.  See the table below for the mapping from the ChannelLogger levels to Channelz
+   * severity level (see {@code ChannelTraceEvent} from <a
+   * href="https://github.com/grpc/grpc-java/blob/master/services/src/main/proto/grpc/channelz/v1/channelz.proto">channelz.proto</a>)
+   * and Java logger levels.  Note that {@code DEBUG} level is not recorded on Channelz.
+   * <pre>
+   * +---------------------+-------------------+-------------------+
+   * | ChannelLogger Level | Channelz Severity | Java Logger Level |
+   * +---------------------+-------------------+-------------------+
+   * | DEBUG               | N/A               | FINEST            |
```
carl-mastrangelo: This table should be encoded as arguments to the enum valuies:


```java
public enum Level {
  DEBUG(null, "CT_INFO", Level.FINEST),
}
```

That way the mapping can be without a switch statement below.

core/src/main/java/io/grpc/internal/ChannelTracer.java
```diff
@@ -45,24 +54,28 @@
...
-      public boolean add(Event event) {
-        if (size() == maxEvents) {
-          removeFirst();
-        }
-        eventsLogged++;
-        return super.add(event);
-      }
-    };
+  ChannelTracer(
+      InternalLogId logId, final int maxEvents, long channelCreationTimeNanos, String description) {
```
carl-mastrangelo: You should document the behavior change of maxEvents

core/src/main/java/io/grpc/internal/ChannelTracer.java
```diff
@@ -45,24 +54,28 @@
...
-      @Override
-      public boolean add(Event event) {
-        if (size() == maxEvents) {
-          removeFirst();
-        }
-        eventsLogged++;
-        return super.add(event);
-      }
-    };
+  ChannelTracer(
```
carl-mastrangelo: Can description be null?

core/src/main/java/io/grpc/ChannelLogger.java
```diff
@@ -0,0 +1,51 @@
...
+   * | INFO                | CT_INFO           | FINEST            |
+   * | WARNING             | CT_WARNING        | FINER             |
+   * | ERROR               | CT_ERROR          | FINE              |
+   * +---------------------+-------------------+-------------------+
+   * </pre>
+   */
+  public enum Level {
```
carl-mastrangelo: This probably deserves a name like ChannelLevel, since it behaves differently than other "Levels".

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -32,12 +32,13 @@
 import com.google.errorprone.annotations.ForOverride;
 import io.grpc.Attributes;
 import io.grpc.CallOptions;
+import io.grpc.ChannelLogger;
+import io.grpc.ChannelLogger.Level;
```
ejona86: I'd suggest not importing this and always referencing it via `ChannelLogger.Level`. Otherwise it is pretty confusing and you can't import `java.util.logging.Level`.

grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
```diff
@@ -320,8 +314,7 @@ void shutdown() {
   }
 
   void propagateError(Status status) {
-    logger.log(Level.FINE, "[{0}] Had an error: {1}; dropList={2}; backendList={3}",
-        new Object[] {logId, status, dropList, backendList});
+    logger.log(Level.DEBUG, "Error: " + status);
```
ejona86: We really want to know whether or not this log message will be used, to avoid the string handling when useless.

zhangkun83: Since ChannelLogger is not on data path and is called pretty infrequently, I am not sure it's worth the code clutter to wrap it with a check.

ejona86: Things like `logger.log(Level.DEBUG, "Got an LB response: " + response);` could still be large and run on the network thread. Also, some LBs may do RPCs on cache misses, which could want to log to DEBUG.

I'm a bit worried about how much is getting pulled into this API. There's a lot of data being logged. Like every time the picker changes with `logger.log(Level.INFO, state + ": picks=" + picker.pickList + ", drops=" + picker.dropList);`. I know that isn't on the critical path, but that still happens frequently. And most of the time nobody will see it.

zhangkun83: What about `log(Level level, String template, Object... args)`?. For now I can check the loggable level in the implementation and that would save the string generation for `DEBUG`. We could optimize `ChannelTracer` even later to generate strings only when Channelz is requested.

ejona86: `log(Level, String, Object...)` seems fine to me. It'd probably be an overload (so `log(Level, String)` would still exist) to avoid allocating the `Object[]` when unnecessary.

core/src/main/java/io/grpc/ChannelLogger.java
carl-mastrangelo: It's suffiicient to make this a String instead of a hard ref, and then have a test that assert they all work.

zhangkun83: For these fields to be accessible from `ChannelLoggerImpl`, they would need to be public, which technically exposes these `CT_` enums even not referencing them directly. Doesn't seem enough benefit from doing that.

carl-mastrangelo: These could be exposed via an internal accessor.  

The benefit is that we can use hard references to see exactly how they map, rather than relying on switch statements.

core/src/main/java/io/grpc/ChannelLogger.java
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Logs a message.
+   */
+  public abstract void log(ChannelLogLevel level, String message);
+
+  /**
+   * Logs a message, using a message template and a list of arguments used to generate the log
+   * message with {@link String#format}.
+   */
+  public abstract void log(ChannelLogLevel level, String template, Object... args);
```
carl-mastrangelo: Isn't this format different than what normal loggers use?  (i.e. `{}`) ?

I think this should be something that fits in with other, similar log statements.  (I.e. it should do what `SimpleFormatter` does)

ejona86: It should use `MessageFormat`.

zhangkun83: Used MessageFormat. PTAL.

core/src/main/java/io/grpc/internal/ChannelLoggerImpl.java
```diff
@@ -0,0 +1,76 @@
...
+  @Override
+  public void log(ChannelLogLevel level, String messageFormat, Object... args) {
+    if (level == ChannelLogLevel.DEBUG) {
+      // DEBUG logs can be expensive to generate (e.g., large proto messages), and when not logged,
+      // go nowhere.  We will skip the generation if it's not logged.
+      if (ChannelTracer.logger.isLoggable(Level.FINEST)) {
+        tracer.logOnly(Level.FINEST, MessageFormat.format(messageFormat, args));
+      }
+    } else {
+      reportToTracer(level, MessageFormat.format(messageFormat, args));
```
ejona86: This will create the event and format the message even when tracing is disabled.

PR #5026, opened 2018-11-02
services: HealthCheckingLoadBalancerUtil and HealthCheckingLoadBalancerProvider

zhangkun83:
HealthCheckingLoadBalancerUtil is the public wrapper utility that helps
turn a LoadBalancerFactory into a health-checking capable one.

HealthCheckingRoundRobinLoadBalancerProvider overrides the
RoundRobinLoadBalancerProvider from grpc-core.

Tracking issue: #4932


services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerProvider.java
```diff
@@ -0,0 +1,77 @@
...
+import io.grpc.internal.ExponentialBackoffPolicy;
+import io.grpc.internal.TimeProvider;
+
+/**
+ * A LoadBalancerProvider that wraps another provider and adds <a
+ * href="https://github.com/grpc/proposal/blob/master/A17-client-side-health-checking.md">client-side
+ * health checking</a>.
+ */
+@ExperimentalApi("https://github.com/grpc/grpc-java/issues/5025")
+public abstract class HealthCheckingLoadBalancerProvider extends LoadBalancerProvider {
```
ejona86: I don't think this should be a provider and I don't think it should be extended. Extending means you can't mix-and-match other similar solutions. Instead, I think each LB Provider that wants this should call a utility during `newLoadBalancer()`.

zhangkun83: I have change the PR to expose a utility instead. PTAL.

PR #5033, opened 2018-11-06
benchmarks: driverServer graceful shutdown

jtattermusch:
Fixes https://github.com/grpc/grpc/issues/17101.

According to the docs the `shutdownNow()` kills all the inflight calls (including quitWorker method) before they can finish and hence the C++ benchmark driver reports `driver.cc:550]  Worker 0 could not be properly quit because Received RST_STREAM with error code 8`  (I misinterpreted the msg earlier, the error code 8 seems to mean "cancelled" according to http/2 spec).

Before I was able to reproduce reliably locally and after this change, I'm unable to reproduce, so I assume it fixes the problem.


jtattermusch:
Merging as `driver.awaitTermination()` is already invoked here: https://github.com/grpc/grpc-java/blob/f8f86da480a607f26b205da73d5c3945d699ce78/benchmarks/src/main/java/io/grpc/benchmarks/driver/LoadWorker.java#L110

PR #5035, opened 2018-11-06
core: Explicitly mention MCB.intercept's execution order

ejona86:
New users are much more likely to use MCB to add an interceptor instead
of ClientInterceptors, so may not be aware of the interesting execution
order.

CC @xCASx

### DIFF
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -104,7 +104,9 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   /**
    * Adds interceptors that will be called before the channel performs its real work. This is
    * functionally equivalent to using {@link ClientInterceptors#intercept(Channel, List)}, but while
-   * still having access to the original {@code ManagedChannel}.
+   * still having access to the original {@code ManagedChannel}. Interceptors run in the reverse
+   * order in which they are added, just as with consecutive calls to {@code
+   * ClientInterceptors.intercept()}.
    *
    * @return this
    * @since 1.0.0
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -104,7 +104,9 @@
   /**
    * Adds interceptors that will be called before the channel performs its real work. This is
    * functionally equivalent to using {@link ClientInterceptors#intercept(Channel, List)}, but while
-   * still having access to the original {@code ManagedChannel}.
+   * still having access to the original {@code ManagedChannel}. Interceptors run in the reverse
```
carl-mastrangelo: not  `{@link }` ? :P

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -115,6 +117,8 @@
    * Adds interceptors that will be called before the channel performs its real work. This is
    * functionally equivalent to using {@link ClientInterceptors#intercept(Channel,
    * ClientInterceptor...)}, but while still having access to the original {@code ManagedChannel}.
+   * Interceptors run in the reverse order in which they are added, just as with consecutive calls
```
carl-mastrangelo: Optional: ClientCall.Listener events happen in the same order as interceptors though.  It's only reverse going down, which may be not obvious.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
ejona86: It is linked earlier in the same paragraph, so no.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
ejona86: I'm not sure how to discuss that without it being even more confusing. To me, just the fact that we mention this should clue most users in, even if they aren't 100% certain what it implies when they read it first.

PR #5037, opened 2018-11-06
all: Emit bin-headers with unpadded encoding

dapengzhang0:


dapengzhang0:
I though this change was breaking okhttp, but it turned out my env issue.

core/src/main/java/io/grpc/internal/TransportFrameUtil.java
```diff
@@ -61,7 +61,8 @@
       if (endsWith(key, binaryHeaderSuffixBytes)) {
         // Binary header.
         serializedHeaders[k] = key;
-        serializedHeaders[k + 1] = BaseEncoding.base64().encode(value).getBytes(US_ASCII);
+        serializedHeaders[k + 1]
+            = BaseEncoding.base64().omitPadding().encode(value).getBytes(US_ASCII);
```
ejona86: This will now do an allocation. Let's configure the base encoder once outside of the for loop.

dapengzhang0: Can I declare a public static final instance in this class, and use it everywhere (except for core)?

dapengzhang0: Now I use an static instance in `InternalMetadata`.

dapengzhang0:
Done for the comment that couldn't be found any more. PTALA

PR #5042, opened 2018-11-07
services: HealthCheckingLoadBalancer logs to ChannelLogger

zhangkun83:
Log the event that health check is disabled due to UNIMPLEMENTED as required in the spec:
https://github.com/grpc/proposal/blob/master/A17-client-side-health-checking.md

Also log every Subchannel state change that is affected by health-checking, i.e., the state changes when the raw Subchannel state is READY and health-check is running.

Tracking issue: #4932


services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -393,8 +393,9 @@ void handleResponse(HealthCheckResponse response) {
 
       void handleStreamClosed(Status status) {
         if (Objects.equal(status.getCode(), Code.UNIMPLEMENTED)) {
-          // TODO(zhangkun83): record this to channel tracer
           disabled = true;
+          subchannel.getChannelLogger().log(
```
carl-mastrangelo: is this really an error?

zhangkun83: Spec says log it as ERROR.

PR #5047, opened 2018-11-08
Add missing j2objc dependency to Bazel build

### LINKED ISSUE #5046: "unknown enum constant" warnings when building with Bazel
If I run `bazel build //...` in grpc-java at HEAD (or v1.16.1) I see many warnings like this:

```
INFO: From Building core/libinternal.jar (110 source files):
warning: unknown enum constant ReflectionSupport$Level.FULL
  reason: class file for com.google.j2objc.annotations.ReflectionSupport$Level not found
```

This was discussed on https://github.com/bazelbuild/bazel/issues/6414 and is due to a missing compile-time dependency of guava on j2objc.

HEAD

No warnings when building.

drigz:
This avoids a warning when building artifacts that depend on Guava.

Fixes #5046.


carl-mastrangelo:
@drigz why doesn't guava pull this in?

drigz:
@carl-mastrangelo Bazel doesn't have built-in support for extracting transitive dependency information from Maven. The grpc-java build manually lists all external JARs required by each artifact.

An alternative to grpc-java's current approach would be to use [bazel-deps](https://github.com/johnynek/bazel-deps) to autogenerate the correct dependencies.

ejona86:
Hmm... it looks like we should really use a java_library to add the necessary deps to guava. But I guess that's a bit too much work at the moment, and the only use of j2objc will be because of Guava, so this should be easy to fix-up in the future.

PR #5051, opened 2018-11-09
services: cancel health-check when LoadBalancer.shutdown() is called.

zhangkun83:
The health checking balancer won't receive an update about Subchannel
shutdown via handleSubchannelState(), because no more callback will be
called after LoadBalancer.shutdown() is called.

Tracking issue: #4932

### DIFF
--- a/services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
+++ b/services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
@@ -95,6 +95,7 @@ final class HealthCheckingLoadBalancerFactory extends Factory {
     
     private LoadBalancer delegateBalancer;
     @Nullable String healthCheckedService;
+    private boolean balancerShutdown;
 
     final HashSet<HealthCheckState> hcStates = new HashSet<HealthCheckState>();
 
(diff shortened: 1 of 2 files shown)


services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -183,6 +183,18 @@ public void handleSubchannelState(
...
     }
 
+    @Override
+    public void shutdown() {
+      super.shutdown();
+      for (HealthCheckState hcState : helper.hcStates) {
+        // ManagedChannel will stop calling handleSubchannelState() after shutdown() is called,
+        // which is required by LoadBalancer API semantics. We need to deliver the final SHUTDOWN
+        // signal to health checkers so that they can cancel the streams.
+        hcState.updateRawState(ConnectivityStateInfo.forNonError(SHUTDOWN));
```
ejona86: This seems like it would notify `handleSubchannelState()` of our delegate after `shutdown()`. Unfortunately, putting `super.shutdown()` after this loop would also be confusing, because normally they wouldn't see the shutdown state. Do we need a "shutdown" flag to prevent notifying delegate of these transitions? Or would we prefer a flag so that the failed RPCs don't retry?

zhangkun83:
cla please be happy.

PR #5063, opened 2018-11-14
stub: disable exception if onCancelHandler set

### LINKED ISSUE #5061: Add option to disable cancellation causing exception on Server
As mentioned some in #4351, if there is a cancellation on server-side we end up throwing an exception when sending from the async stub. That's very surprising and isn't great.

That used to be the only way of receiving cancellation notification with the async stub. However, we now have ways to let users add their own cancellation listeners. So we could either disable the throwing when using cancellation callback or have an option that will disable the throwing behavior.

venilnoronha:
This restrains a cancellation `Exception` when an `onCancelHandler` is set in `ServerCallStreamObserverImpl`. Fixes #5061.

### DIFF
--- a/stub/src/main/java/io/grpc/stub/ServerCalls.java
+++ b/stub/src/main/java/io/grpc/stub/ServerCalls.java
@@ -331,14 +331,17 @@ public final class ServerCalls {
 
     @Override
     public void onNext(RespT response) {
-      if (cancelled && onCancelHandler == null) {
-        throw Status.CANCELLED.withDescription("call already cancelled").asRuntimeException();
+      if (cancelled) {
+        if (onCancelHandler == null) {
+          throw Status.CANCELLED.withDescription("call already cancelled").asRuntimeException();
+        }
+      } else {
+        if (!sentHeaders) {
+          call.sendHeaders(new Metadata());
+          sentHeaders = true;
+        }
+        call.sendMessage(response);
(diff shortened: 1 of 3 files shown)


venilnoronha:
ping

stub/src/main/java/io/grpc/stub/ServerCalls.java
```diff
@@ -332,13 +332,16 @@ public void setCompression(String compression) {
     @Override
     public void onNext(RespT response) {
       if (cancelled) {
-        throw Status.CANCELLED.withDescription("call already cancelled").asRuntimeException();
-      }
-      if (!sentHeaders) {
-        call.sendHeaders(new Metadata());
-        sentHeaders = true;
+        if (onCancelHandler == null) {
```
ejona86: nit: Could this have an `else` that `return`s? That would allow removing the current `else` and returning the code to look a bit more like it did before. (I do see that this code as written mirrors onCompleted, but it also has a single line in the else, so it doesn't matter as much.)

venilnoronha: Addressed in 8276d64.

ejona86:
@venilnoronha, Thank you!

venilnoronha:
@ejona86 thanks for reviewing/merging!

PR #5067, opened 2018-11-17
netty: support listening on multiple port

carl-mastrangelo:
I did my best to make sure this was thread safe, please be forgiving!

### DIFF
--- a/core/src/main/java/io/grpc/InternalChannelz.java
+++ b/core/src/main/java/io/grpc/InternalChannelz.java
@@ -311,7 +311,7 @@ public final class InternalChannelz {
       private long callsSucceeded;
       private long callsFailed;
       private long lastCallStartedNanos;
-      public List<InternalInstrumented<SocketStats>> listenSockets = Collections.emptyList();
+      public List<InternalInstrumented<SocketStats>> listenSockets = new ArrayList<>();
 
       public Builder setCallsStarted(long callsStarted) {
         this.callsStarted = callsStarted;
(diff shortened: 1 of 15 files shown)


ST-DDT:
Having multiple connectors is a nice feature for services that serve both internal and external requests. If I also have the ability to omit the encryption or use a different certificate set for internal services that would be awesome.

creamsoup:
Discussed offline with @beschmi and confirmed that this change can solve an internal requirement. For this requirement, it don't really need the multiple InternalServers.

I don't expect API changes for the multiple InternalServer approach (at least for the existing APIs like getPorts). if this is true, we can support the multiple InternalServer on top of this change.

ST-DDT:
@carl-mastrangelo Can I somehow check which endpoint was used to connect to the server/receive a request?
Is there an attribute like `Metadata.Key<SocketAddress> TARGET_ADDRESS`?

ST-DDT:
Can we assign special names or attributes for each address? Like `internal` or `external`?

ejona86:
This change includes invasive changes to NettyServer that would need to be reverted if we go with the alternative approach, and this change won't likely be automatically revertable due to conflicts. So I want to get some basic understanding of how hard it would be to do it in the ServerImpl.

carl-mastrangelo:
@ejona86 This is ready to be reviewed again.    NSB is a little more awkward because it has to create a list of transport servers, but aside from that it should be pretty straight forward.

Still not exposing an API, since I haven't thought what it should look like for InProcess.

PR #5069, opened 2018-11-17
Update ClientInterceptor for kotlin integration

fb64:
Add nonnull indication on channel parameter to improve integration with kotlin.
This allows avoid kotlin consider channel as nullable when implementing ClientInterceptor.


ST-DDT:
Shouldn't the other parameters also be annotated `@Nonnull`? What about the `ServerInterceptor`?

fb64:
I think you're right, both in ServerInterceptor and ClientInterceptor interfaces, all method parameters should not be null. I'll update my pull request ...

ST-DDT:
I guess the line needs to be wrapped now (Checkstyle is complaining)

fb64:
Sorry I didn't see that ... 
I just updated the PR, normally everything is well formatted for checkstyle.

fb64:
Sorry I didn't see that ... 
I just updated the PR, normally everything is well formatted for checkstyle.

fb64:
My responses:
1. You’re right, and it can be interesting to continue to add non-null check when it’s needed.
2.  Effectively this annotation is part of [JSR 305](https://jcp.org/en/jsr/detail?id=305) which is actually “dormant”. But it's already used in [ServiceConfigInterceptor](https://github.com/grpc/grpc-java/blob/e9b656845045fd596c2b7a2c0425b9b80ef40848/core/src/main/java/io/grpc/internal/ServiceConfigInterceptor.java#L76) method for example. For information kotlin supports [others annotations ](https://kotlinlang.org/docs/reference/java-interop.html#nullability-annotations)
3. Yes, but actually are those parameters can be really null ? i think no…
4.  Yes I'm agree, like channel in  [ClientInterceptors](https://github.com/grpc/grpc-java/blob/b0f423295b4674cb5247a6143fd211b050ef0065/core/src/main/java/io/grpc/ClientInterceptors.java#L87) for example
5. This annotation already exists in the code base and it would be a shame not to continue using it. I think this kind of indications help both java and kotlin users.

ejona86:
The main difference is that the current annotations are more of documentation and don't change the behavior of anything. I'm not aware of having `javax.annotation.Nonnull` in a public API. (Actually... looks like `GrpcCleanupRule` and `LoadBalancer` include such an annotation, although it is for 'in' parameters to our API.)

There is a lot more effort and maintenance necessary to make our implementation Nullability-checker-safe. Unfortunately, I've not figured out a way good way to have reliable Nullability indicators on the public API without also having a checker enabled.

If we move `io.grpc.internal` to its own artifact, then we could enable the checker for io.grpc. That may not be too onerous...

ST-DDT:
Currently every java implementation I know implicitly assumes that the parameters in the interceptors are non-null. In fact I don't even know what to do if I ever get a null there (except for throwing an exception).

So stating it explicitly doesn't do any harm. If we ever change the nullability contract, we would break the interceptors anyway regardless of the annotations. Having the annotations will also reduce the chance of accidental nullability changes.

If you consider the null parameters a possibility at all, then you have to change the examples/existing implementations to have at least a comment such as `// should check for null here...`.

The only concern I have is the (missing) consistency of having the annotations.

ejona86:
I'm not disagreeing that these instances should be non-null. Although, technically, these are only non-null if Channel.newCall's arguments are non-null. So we also need to mark those non-null. And there's many, many more APIs that would need the non-null marking. I'm worried of getting into the "business" of marking nullability within a Java API, because I feel like we would need a nullability checker and that then imposes nullability checks on the rest of the library. I'm worried about maintaining API stability, as we have little experience with the gotchas involved here nor what changes would maintain stability.

For example, can we mark Channel.newCall's arguments as non-null? They _should_ be non-null, and you'd get a NPE today if they were null. But is it _possible_ to break existing Kotlin applications by adding such an annotation? (That seems likely...) For that matter, I'm not even 100% confident that adding non-null annotation to ClientInterceptor is API-safe. For example, if Kotlin fails to compile if you use `?.` on a non-null argument, it could be API-breaking. If Kotlin requires the Kotlin method definition to match exactly the Java annotations, that would be API-breaking.

It is really bizarre supporting Kotlin-isms from a Java library, as we don't know Kotlin! I'm sort of feeling like doing this "right" involves having someone who knows (or learns) Kotlin and then develop a _comprehensive_ plan on how to best support the language, with rules the rest of us could follow. At that point we could decide whether it was worth it, or whether that is the approach we wanted. (Vs, say, a Kotlin-specific API.)

fb64:
The point is no to “support" kotlin, it’s already the case because kotlin if fully interoperable with java. It’s just add indication/check to improve kotlin integration, that can improve java usage too. It was just a proposition, I understand your point of view and I don’t know what approach is really the best. 
I tried just to begin by a concrete case because I am actually using grpc in Android app written in kotlin ...
For information If you use “?." on a non-null variable in kotlin it does not  break anything but the inverse does ..

joelpedraza:
> But is it possible to break existing Kotlin applications by adding such an annotation? 

No. 

 For that matter, I'm not even 100% confident that adding non-null annotation to ClientInterceptor is API-safe. For example, if Kotlin fails to compile if you use ?. on a non-null argument, it could be API-breaking. If Kotlin requires the Kotlin method definition to match exactly the Java annotations, that would be API-breaking.

joelpedraza:
> * Adding these needs to be done carefully, because it seems removing them is an API breakage for Kotlin users

It can be, yes.

> * But is it possible to break existing Kotlin applications by adding such an annotation? 

Yes. In fact this PR will.

As for why it's a breaking change: it has to do with subtypes and variance.

In kotlin any non null `T` is a subtype of nullable `T?`

Adding @NotNull breaks invariance and contravariance.
Adding @Nullable breaks covariance.

The inverse also holds

Removing @NotNull breaks covariance
Removing @Nullable breaks invariance and contravariance.

Java Linters or IDEs can make nullability annotation violations warning or errors. If errors they have the same net effect as the kotlin compiler errors.

Opinion from here on out:

>   * As a Java user, I'm sort of offended by feeling like I have to drink the nullability cool-aid because of Kotlin's design choices

In java all subtypes of `Object` are nullable. Nullability is there, it's just _always_ there. Without annotations it's just not _compile time checked_ nullability.

Ultimately I think adding nullability annotations improves this project's own codebase (if linting is enabled), and also documents for users, and  improves their usages of it. Regardless of language.

Since this can break source compatibility all kotlin users, or fail builds for java users with lint warningsAsErrors set to true: adding or removing the annotations should probably only happen in major versions.

If you're tempted to think "ugh, I can never change nullability without a breaking change" it might be helpful to reframe it like "oh, were preventing introducing new bugs in existing client code because they held old assumptions that are no longer true". Annotations or no, it's a breaking change! It's a question of whether it breaks at compile time or run time. The safest code is the code that does not compile. =)

PR #5070, opened 2018-11-17
core: add (de)register() in LoadBalancerRegistry.

zhangkun83:


zhangkun83:
Can you elaborate the classloader leak issue?

ejona86:
When "plugin" systems are used the application will load additional Java classes within their own classloader. They could possibly make changes to the global system. When the plugins are removed they need to de-register themselves and remove all references to themselves from the system, otherwise their class loader will remain retained.

This normally comes up with servlets, and they need to do things like shut down any threads they created. Servlets+gRPC wouldn't have this problem, but the general concept remains. Basically, it is best not to assume that an application is a single static thing at runtime. For example, IDEs are made almost entirely pluggable pieces of code.

zhangkun83:
@ejona86  PTAL

core/src/main/java/io/grpc/LoadBalancerRegistry.java
```diff
@@ -35,48 +38,56 @@
...
-        logger.fine(provider + " found but not available");
-        continue;
-      }
+  /**
+   * Register a provider.
+   *
+   * <p>If the provider's {@link LoadBalancerProvider#isAvailable isAvailable()} returns
+   * {@code false}, this method will throw {@link IllegalArgumentException}.
+   *
+   * <p>If more than one providers with the same {@link LoadBalancerProvider#getPolicyName policy
```
ejona86: s/providers/provider/

core/src/main/java/io/grpc/LoadBalancerRegistry.java
```diff
@@ -89,27 +100,33 @@ public static synchronized LoadBalancerRegistry getDefaultRegistry() {
           HARDCODED_CLASSES,
           LoadBalancerProvider.class.getClassLoader(),
           new LoadBalancerPriorityAccessor());
-      instance = new LoadBalancerRegistry(providerList);
+      instance = new LoadBalancerRegistry();
+      for (LoadBalancerProvider provider : providerList) {
+        logger.fine("Service loader found " + provider);
+        if (provider.isAvailable()) {
+          instance.register(provider);
```
ejona86: addAll + refresh would be more ideal, to avoid the N^2 (not that that really matters given the sizes involved, except this is done at startup and will probably be interpreted).

core/src/main/java/io/grpc/LoadBalancerRegistry.java
```diff
@@ -35,48 +38,56 @@
...
-      }
+  /**
+   * Register a provider.
+   *
+   * <p>If the provider's {@link LoadBalancerProvider#isAvailable isAvailable()} returns
+   * {@code false}, this method will throw {@link IllegalArgumentException}.
+   *
+   * <p>If more than one providers with the same {@link LoadBalancerProvider#getPolicyName policy
+   * name} are registered, the one with the highest {@link LoadBalancerProvider#getPriority
+   * priority} will be effective.  If there are more than one name-sake providers ranking the
```
ejona86: s/ranking/rank/

core/src/main/java/io/grpc/ServiceProviders.java
```diff
@@ -71,11 +71,16 @@ private ServiceProviders() {
       list.add(current);
     }
 
-    // Sort descending based on priority.
+    // Sort descending based on priority.  If priorities are equal, compare the class names to
```
ejona86: This is contrary to "If there are more than one name-sake providers ranking the highest priority, the one registered first will be effective."

If it returns zero, then a stable sort will produce the correct order.

core/src/main/java/io/grpc/LoadBalancerRegistry.java
zhangkun83: Done.

core/src/main/java/io/grpc/LoadBalancerRegistry.java
zhangkun83: Done.

core/src/main/java/io/grpc/ServiceProviders.java
zhangkun83: > This is contrary to "If there are more than one name-sake providers ranking the highest priority, the one registered first will be effective."

I don't think so. This determines the order of providers being registered in the default register. Sorting by class name gives an deterministic registration order.

core/src/main/java/io/grpc/LoadBalancerRegistry.java
zhangkun83: Done in a slightly different way.

PR #5073, opened 2018-11-19
core: support LoadBalancingConfig from Service Config

zhangkun83:
LoadBalancingConfig is a JSON object (map) where the only key is the policy name, and the only value is a JSON object (map) which is the config for that policy. It supersedes LoadBalancingPolicy which is just a string.

While the Service Config contains one LoadBalancingPolicy, it contains a list of LoadBalancingConfig objects. The first LoadBalancingConfig that is usable at runtime is the effective one.

AutoConfiguredLoadBalancer passes the parsed policy config (the value corresponding to the policy name)  to the delegate LoadBalancer as an attribute via `handleResolvedAddressGroups()`. In cases where a policy creates child policies, the parent policy is expected to pass on this attributes to its children.

/cc @markdroth


core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -138,76 +155,106 @@ void setDelegate(LoadBalancer lb) {
...
+   *
+   * @param servers The list of servers reported
+   * @param config the service config object
+   * @return the new load balancer factory, never null
+   */
+  @VisibleForTesting
+  static PolicySelection decideLoadBalancerProvider(
+      List<EquivalentAddressGroup> servers, @Nullable Map<String, Object> config,
+      ChannelLogger channelLogger)
+    throws PolicyException {
```
ejona86: Indent more or move to previous line.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -138,76 +155,106 @@ void setDelegate(LoadBalancer lb) {
...
+      LinkedHashSet<String> policiesTried = new LinkedHashSet<String>();
+      for (Map<String, Object> lbConfig : lbConfigs) {
+        if (lbConfig.size() != 1) {
+          throw new PolicyException(
+              "There are " + lbConfig.size()
+              + " load-balancing configs in a list item. Exactly one is expected. Config="
+              + lbConfig);
         }
+        Entry<String, Object> entry = lbConfig.entrySet().iterator().next();
+        // Internally the service config is written in Protobuf and converted to JSON.
```
ejona86: Can this be done in the internal NameResolver instead?

I'd prefer not to let users configure the "wrong" names in DNS.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -138,76 +155,106 @@ void setDelegate(LoadBalancer lb) {
...
+          return new PolicySelection(provider, (Map) entry.getValue());
+        }
+        policiesTried.add(policy);
       }
-      return getProviderOrThrow(DEFAULT_POLICY, "Using default policy");
+      throw new PolicyException(
+          "None of " + policiesTried + " specified by Service Config are available.");
     }
+    return new PolicySelection(
+        getProviderOrThrow(DEFAULT_POLICY, "Using default policy"), null);
```
ejona86: Given how the string is combined, it seems this should be s/U/u/ or the string should be quoted.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
zhangkun83: While I'd prefer the same, there is a cost if we do the conversion in the internal NameResolver. Because the policy name can appear anywhere in the service config, and it can be recursive, we'd need to replace all policy names in the config. There isn't a generic and reliable way to locate them. Because the proto is defined, we could enumerate known locations, but every time a new policy is added to the config proto, we'd need to update the conversion code, in all languages, which could be a hassle.

Alternatively, we could add an internal flag the control whether to accept camel case policy names. We will make it false by default, but turn it on internally. If we decide to go this route, we should probably do the same across all languages.

@markdroth @dfawley @menghanl @AspirinSJL WDTY?

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -81,11 +86,16 @@ public void shutdown() {}
...
+      }
       Map<String, Object> configMap = attributes.get(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG);
-      LoadBalancerProvider newlbp;
+      PolicySelection selection;
       try {
-        newlbp = decideLoadBalancerProvider(servers, configMap);
-      } catch (PolicyNotFoundException e) {
+        selection = decideLoadBalancerProvider(servers, configMap, helper.getChannelLogger());
+      } catch (PolicyException e) {
         Status s = Status.INTERNAL.withDescription(e.getMessage());
```
carl-mastrangelo: chain the cause?

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -138,76 +155,106 @@ void setDelegate(LoadBalancer lb) {
...
-            ServiceConfigUtil.getLoadBalancingPolicyFromServiceConfig(config);
-        if (serviceConfigChoiceBalancingPolicy != null) {
-          // Handle ASCII specifically rather than relying on the implicit default locale of the str
-          return getProviderOrThrow(
-              Ascii.toLowerCase(serviceConfigChoiceBalancingPolicy),
-              "service-config specifies load-balancing policy");
+    if (config != null) {
+      List<Map<String, Object>> lbConfigs =
+          ServiceConfigUtil.getLoadBalancingConfigsFromServiceConfig(config);
+      LinkedHashSet<String> policiesTried = new LinkedHashSet<String>();
```
carl-mastrangelo: `s/new LinkedHashSet<String>();/new LinkedHashSet<>();/`

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
zhangkun83: I intentionally do not attach the exception to the status because the PolicyException's message has all information.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
zhangkun83: done.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
zhangkun83: done.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
zhangkun83: Per offline discussion, we have found a way to keep the underscored names in the json converted from proto, and we will not do any normalization in here. "round_robin" will always be "round_robin" everywhere, except for the deprecated LoadBalancingPolicy enum which will still be "ROUND_ROBIN" which is toLowerCase()'ed before used for registry lookup.

PR #5079, opened 2018-11-21
core/services: test/debug friendlier

zhangkun83:
Raise visibility of AutoConfiguredLoadBalancerFactory as internal
tests need to access it from a different package.

Rename HealthCheckingLoadBalancerFactory.LoadBalancerImpl to
*.HealthCheckingLoadBalancer so that its toString() output is more
informative.


zhangkun83:
I have a test that takes a LoadBalancingConfig proto, converts it to JSON and feed it to AutoConfiguredLoadBalancer, and verifies that it's interpreted correctly.

ejona86:
Ah, makes sense.

PR #5088, opened 2018-11-27
core: allow ClientStreamTracer to intercept trailers.

zhangkun83:
This would allow LoadBalancers to intercept trailers which could carry
load information.

### DIFF
--- a/core/src/main/java/io/grpc/ClientStreamTracer.java
+++ b/core/src/main/java/io/grpc/ClientStreamTracer.java
@@ -37,6 +37,15 @@ public abstract class ClientStreamTracer extends StreamTracer {
   }
 
   /**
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/grpc/ClientStreamTracer.java
```diff
@@ -36,6 +36,15 @@ public void outboundHeaders() {
   public void inboundHeaders() {
   }
 
+  /**
+   * Trailing metadata has been received from the server.
+   *
+   * @param trailers the mutable trailing metadata.  Modifications to it will be seen by
+   *                 interceptors and the application.
+   */
```
carl-mastrangelo: please add `@since`

Issue #4084, opened 2018-02-14
Swap to java-library gradle plugin


ejona86:
It should speed builds and rebuilds, while also getting us closer to having a (clean) Java 9 module.

https://docs.gradle.org/current/userguide/java_library_plugin.html

ejona86:
This looks like a prerequisite for enabling minimization for the gradle shadow plugin. It's unclear if we want minimization, though.

PR #5098, opened 2018-11-29
examples: prefer implementation to compile in build.gradle

dapengzhang0:
It's a lot of work to tackle #4084 but we can first do some cleanup in examples.

### DIFF
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -72,63 +72,63 @@ task routeGuideServer(type: CreateStartScripts) {
     mainClassName = 'io.grpc.examples.routeguide.RouteGuideServer'
     applicationName = 'route-guide-server'
     outputDir = new File(project.buildDir, 'tmp')
-    classpath = project.startScripts.classpath
+    classpath = startScripts.classpath
 }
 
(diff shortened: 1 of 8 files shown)


examples/build.gradle
```diff
@@ -31,22 +28,22 @@ def protobufVersion = '3.5.1'
 def protocVersion = '3.5.1-1'
 
 dependencies {
-    compile "com.google.api.grpc:proto-google-common-protos:1.0.0"
-    compile "io.grpc:grpc-alts:${grpcVersion}"
-    compile "io.grpc:grpc-netty-shaded:${grpcVersion}"
-    compile "io.grpc:grpc-protobuf:${grpcVersion}"
-    compile "io.grpc:grpc-stub:${grpcVersion}"
-    compileOnly "javax.annotation:javax.annotation-api:1.2"
```
ericgribkoff: `compileOnly` is still valid, as I understand it: https://developer.android.com/studio/build/dependencies

dapengzhang0: Right, I replaced it by mistake, actually I wanted to keep it.

dapengzhang0: Done.

PR #5105, opened 2018-12-03
all: sync tcnative version

dapengzhang0:
The helloworld TLS example can not run properly with the outdated `examples/build.gradle`.


dapengzhang0:
The helloworld TLS example can not run properly with the outdated `examples/build.gradle`.

PR #5118, opened 2018-12-05
StreamObservers: make sure onComplete only gets called once from onReadyHandler

zklapow:
re: #4558

I encountered the same issue as @biran0079 was seeing where `onReady` was getting called again after `onComplete` had already been called, resulting in an exception because the channel had already been half closed. This simply tracks whether the stream has previously been completed and refuses to call `onComplete`.

I opted not to make this thread-safe because it seems like many of these interfaces are generally not thread-safe (and AFAICT this will only get called from one thread).

### DIFF
--- a/stub/src/main/java/io/grpc/stub/StreamObservers.java
+++ b/stub/src/main/java/io/grpc/stub/StreamObservers.java
@@ -44,12 +44,15 @@ public final class StreamObservers {
     Preconditions.checkNotNull(target, "target");
 
     final class FlowControllingOnReadyHandler implements Runnable {
+      private boolean completed;
+
       @Override
       public void run() {
         while (target.isReady() && source.hasNext()) {
           target.onNext(source.next());
         }
-        if (!source.hasNext()) {
+        if (!source.hasNext() && !completed) {
+          completed = true;
           target.onCompleted();
         }
       }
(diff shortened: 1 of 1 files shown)


thelinuxfoundation:

* User @zklapow isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

zklapow:
CLA signed

stub/src/main/java/io/grpc/stub/StreamObservers.java
```diff
@@ -44,13 +44,18 @@
...
+      private boolean completed;
+
       @Override
       public void run() {
-        while (target.isReady() && source.hasNext()) {
-          target.onNext(source.next());
-        }
-        if (!source.hasNext()) {
-          target.onCompleted();
+        if (!completed) {
```
ejona86: nit: Instead, do the following to avoid increasing the indentation.

```java
if (completed) {
  return;
}
```

zklapow: 👍 updated

zklapow:
CLA signed correctly this time :)

ejona86:
@zklapow, thank you!

PR #5121, opened 2018-12-05
core: add LoadBalancer.Helper.refreshNameResolution()

### LINKED ISSUE #4006: Have LoadBalancers trigger NameResolver.refresh()
Today we implicitly refresh the NameResolver when a transport disconnects. But we don't actually know if that disconnect is important. For example, with GRPCLB only the OOB connection to the GRPCLB server really impacts the addresses we use; GRPCLB will be telling us when a service backend should be removed from our list. Thus, the refresh is useless for all disconnections except from the OOB channel.

C today has the LB decide when to refresh. That seems pretty easy to implement in the normal cases and helpful in the complex cases. So we should consider doing the same thing.

Unfortunately, it does cause some problem for migrating existing LB implementations. Maybe if we see the LB manually call refresh() we disable the implicit refreshing we do today. Maybe something else.

CC @zhangkun83

zhangkun83:
Resolves #4006

### DIFF
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -394,10 +394,10 @@ final class ManagedChannelImpl extends ManagedChannel implements
 
   // Run from syncContext
   @VisibleForTesting
-  class NameResolverRefresh implements Runnable {
+  class DelayedNameResolverRefresh implements Runnable {
     @Override
     public void run() {
-      nameResolverRefresh = null;
+      scheduledNameResolverRefresh = null;
       if (nameResolver != null) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -984,7 +994,7 @@ void remove(RetriableStream<?> retriableStream) {
     // Must be called from syncContext
     private void handleInternalSubchannelState(ConnectivityStateInfo newState) {
       if (newState.getState() == TRANSIENT_FAILURE || newState.getState() == IDLE) {
-        nr.refresh();
```
ejona86: Move `nr` to `NameResolverListenerImpl`?

PR #5122, opened 2018-12-05
context: make Deadline toString() more readable

carl-mastrangelo:


context/src/main/java/io/grpc/Deadline.java
```diff
@@ -142,11 +145,36 @@ public long timeRemaining(TimeUnit unit) {
...
+    long remainingNanos = timeRemaining(TimeUnit.NANOSECONDS);
+    long seconds = Math.abs(remainingNanos) / NANOS_PER_SECOND;
+    long nanos = Math.abs(remainingNanos) % NANOS_PER_SECOND;
+
+    StringBuilder buf = new StringBuilder();
+    if (remainingNanos < 0) {
+      buf.append('-');
+    }
+    buf.append(seconds);
+    if (nanos > 0) {
```
zhangkun83: While this block is beautifully done for its purpose, the chance of having N trailing zeros in real life is (1/10^N). `String.format(".%09d", nanos)` does a slightly less nice job but it's probably a worthy trade-off for simpler code.

zhangkun83: A deadline would be whole seconds in most cases, which can be handled by keeping this `if`.

carl-mastrangelo: Done.

PR #5126, opened 2018-12-05
file download example

nddipiazza:
add an example of how to download a file from grpc.

### DIFF
--- a/examples/src/main/java/io/grpc/examples/filedownload/FileDownloadServer.java
+++ b/examples/src/main/java/io/grpc/examples/filedownload/FileDownloadServer.java
@@ -19,9 +19,6 @@ package io.grpc.examples.filedownload;
 import com.google.protobuf.ByteString;
 import io.grpc.Server;
 import io.grpc.ServerBuilder;
-import io.grpc.examples.helloworld.GreeterGrpc;
-import io.grpc.examples.helloworld.HelloReply;
-import io.grpc.examples.helloworld.HelloRequest;
 import io.grpc.stub.StreamObserver;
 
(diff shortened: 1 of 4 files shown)


examples/src/main/java/io/grpc/examples/filedownload/FileDownloadServer.java
```diff
@@ -0,0 +1,119 @@
...
+          int length;
+          while ((length = bis.read(buffer, 0, bufferSize)) != -1) {
+            responseObserver.onNext(
+              DataChunk.newBuilder().setData(ByteString.copyFrom(buffer, 0, length)).build()
+            );
+          }
+          responseObserver.onCompleted();
+        }
+      } catch (IOException e) {
+        e.printStackTrace();
```
carl-mastrangelo: This should probably be an RPC Error.    Somethign like `responseObserver.onError` ?

examples/src/main/proto/file_download.proto
```diff
@@ -0,0 +1,36 @@
+// Copyright 2015 The gRPC Authors
```
carl-mastrangelo: 2018

dapengzhang0:
@nddipiazza There is [a good place](https://github.com/saturnism/grpc-java-by-example) for various such examples mentioned in [grpc-ecosystem](https://github.com/grpc-ecosystem/awesome-grpc/blob/master/readme.md#res-examples). And they are accepting pull requests.

dapengzhang0:
@nddipiazza Thank you for your pull request. However, we would recommend adding it to https://github.com/saturnism/grpc-java-by-example rather than here.

PR #5127, opened 2018-12-05
grpclb: refresh name resolution when balancer stream closed.

zhangkun83:


zhangkun83:
`refreshNameResolution()` will be no-op if resolver is shutdown, because `ManagedChannelImpl.nameResolver` is set to null right after it's shutdown, and `ManagedChannelImpl.refreshNameResolutionNow()` checks nameResolver for null.

PR #5129, opened 2018-12-06
core: do not fail RPCs if balancer address exists but grpclb is unavailable

### LINKED ISSUE #4602: Audit code path when GrpcLB is not in ClassPath
We need to guarantee that if GRPCLB addresses are provided by the NameResolver, they won't make it to a LB instance other than GRPCLB, as other LB types will not properly ignore the GRPCLB addresses and try to route traffic to them and never realize all such RPCs will fail.

Today it seems AutoConfiguredLoadBalancerFactory will throw, which isn't probably appropriate because it will cause the channel to go into panic mode. Instead, it could choose a LB instance that fails all RPCs with a clear Status. Or it could consider stripping the LB addresses and continue like normal, passing the non-LB addresses to the PF/RR/etc. If there are no non-LB addresses, maybe it could still fail RPCs with a clear status saying gRPCLB is required for this target.

We should also add a comment to the GRPCLB Attribute Key noting that LB policies won't receive it.

zhangkun83:
When service owner turns on grpclb through service config, it
shouldn't break existing clients that don't have grpclb in their
classpath.

Resolves #4602

### DIFF
--- a/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
+++ b/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
@@ -20,7 +20,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.annotations.VisibleForTesting;
 import io.grpc.Attributes;
-import io.grpc.ChannelLogger;
 import io.grpc.ChannelLogger.ChannelLogLevel;
 import io.grpc.ConnectivityState;
 import io.grpc.ConnectivityStateInfo;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -155,73 +169,97 @@ void setDelegate(LoadBalancer lb) {
...
+            throw new PolicyException(
+                "There are " + lbConfig.size()
+                + " load-balancing configs in a list item. Exactly one is expected. Config="
+                + lbConfig);
+          }
+          Entry<String, Object> entry = lbConfig.entrySet().iterator().next();
+          String policy = entry.getKey();
+          if (policy.equals(GRPCLB_POLICY)) {
+            throw new PolicyException(
+                GRPCLB_POLICY + " specified by Service Config but no balancer addresses exist");
```
ejona86: This looks like it should exist in gRPC LB instead. And it seems like something like that is already in GrpclbState.

PR #5132, opened 2018-12-06
doc: restructure examples/README.md

dapengzhang0:
Try to make `examples/README.md` better organized. There may be multiple ways to re-structure it, and I'm not opposed to alternatives.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -25,7 +25,7 @@ before trying out the examples.
 
 - [Json serialization](src/main/java/io/grpc/examples/advanced)
 
-#### To build the examples
+### To build the examples
 
 1. **[Install gRPC Java library SNAPSHOT locally, including code generation plugin](../COMPILING.md) (Only need this step for non-released versions, e.g. master HEAD).**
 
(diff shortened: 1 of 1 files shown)


examples/README.md
```diff
@@ -9,7 +9,23 @@ You may want to read through the
...
+
+- [Error handling](src/main/java/io/grpc/examples/errorhandling)
+
+- [Compression](src/main/java/io/grpc/examples/experimental)
+
+- [Flow control](src/main/java/io/grpc/examples/manualflowcontrol)
+
+- [Json serialization](src/main/java/io/grpc/examples/advanced)
+
+#### To build the above examples,
```
ejona86: This is an abuse of a header. 1) choosing a header size instead of a header depth (why is there no `###`) and 2) having a header that is part of a sentence that is continued below.

I'd accept `### To build the examples`

examples/README.md
```diff
@@ -41,7 +57,7 @@ Please refer to gRPC Java's [README](../README.md) and
 [tutorial](https://grpc.io/docs/tutorials/basic/java.html) for more
 information.
 
-## Maven
+#### Maven
```
ejona86: Please use three instead of four. Ditto below.

dapengzhang0: Done.

examples/README.md
dapengzhang0: Done.

PR #5144, opened 2018-12-07
services: add method to make health checker permanently go unhealthy

carl-mastrangelo:


services/src/main/java/io/grpc/services/HealthServiceImpl.java
```diff
@@ -29,28 +29,35 @@
...
 
 final class HealthServiceImpl extends HealthGrpc.HealthImplBase {
 
+  private static final Logger logger = Logger.getLogger(HealthServiceImpl.class.getName());
+
   // Due to the latency of rpc calls, synchronization of the map does not help with consistency.
   // However, need use ConcurrentHashMap to allow concurrent reading by check().
-  private final ConcurrentHashMap<String, ServingStatus> statusMap =
-      new ConcurrentHashMap<String, ServingStatus>();
+  private final Map<String, ServingStatus> statusMap = new ConcurrentHashMap<>();
```
carl-mastrangelo: This change is needed to make Animal Sniffer not get mad.    ConcurrentHashMap.keySet is apparently a Java 8 method.

services/src/main/java/io/grpc/services/HealthServiceImpl.java
```diff
@@ -103,22 +110,48 @@ public void cancelled(Context context) {
...
       if (prevStatus != null) {
         notifyWatchers(service, null);
       }
     }
   }
 
+  void enterTerminalState() {
+    synchronized (watchLock) {
+      if (terminal) {
+        logger.warning("Already terminating");
```
zhangkun83: If you consider this a bug, you should probably log a stack trace here to help debugging.
If it's not a bug, a warning wouldn't be necessary.

carl-mastrangelo: Done.

PR #5145, opened 2018-12-07
interop-testing: Sample to reduce ProxyTest flakiness

### LINKED ISSUE #2951: io.grpc.testing.integration.ProxyTest.smallLatency is flaky
This is a re-appearance of #2176, but that one is so old I feel bad with reopening it.

ejona86:
This also reduces the time of the bandwidth tests from ~6 seconds to ~2,
each (they have about 1s of overhead).

Fixes #2951. The flake is very frequent in #4980.

### DIFF
--- a/interop-testing/src/test/java/io/grpc/testing/integration/ProxyTest.java
+++ b/interop-testing/src/test/java/io/grpc/testing/integration/ProxyTest.java
@@ -24,6 +24,9 @@ import java.io.DataOutputStream;
 import java.io.IOException;
 import java.net.ServerSocket;
 import java.net.Socket;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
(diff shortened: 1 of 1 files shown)


interop-testing/src/test/java/io/grpc/testing/integration/ProxyTest.java
```diff
@@ -79,13 +82,16 @@ public void smallLatency() throws Exception {
...
-    long rtt = (stop - start);
-    assertEquals(latency, rtt, latency);
+    List<Long> rtts = new ArrayList<>();
+    for (int i = 0; i < 3; i++) {
+      long start = System.nanoTime();
+      clientOut.write(message, 0, 1);
+      clientIn.read(message);
+      rtts.add(System.nanoTime() - start);
+    }
+    Collections.sort(rtts);
```
dapengzhang0: What about 
```
long minRtt = Long.MAX_VALUE;
for {
  long rtt = System.nanoTime() - start;
  minRtt = rtt < minRtt ? rtt : minRtt;
}
assertEquals(latency, minRtt, .5 * latency);

Comment (ejona86) [reply interop-testing/src/test/java/io/grpc/testing/integration/ProxyTest.java]: I could. At times I've put in a `System.out.println(rtts)` and it is nice to see all the values. I'd say this seems like a toss up, where it doesn't matter much which approach is taken.

File: interop-testing/src/test/java/io/grpc/testing/integration/ProxyTest.java [hunk 2]
```diff
@@ -135,11 +144,17 @@ public void smallBandwidth() throws Exception {
...
-    long start = System.nanoTime();
-    clientIn.readFully(new byte[5 * bandwidth]);
-    long stop = System.nanoTime();
-
-    long bandUsed = ((5 * bandwidth) / ((stop - start) / TimeUnit.SECONDS.toNanos(1)));
+    int sample = bandwidth / 5;
+    List<Double> bandwidths = new ArrayList<>();
+    for (int i = 0; i < 5; i++) {
+      long start = System.nanoTime();
+      clientIn.readFully(new byte[sample]);
```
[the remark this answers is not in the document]
ejona86: The reason that "bandwidth" was chosen here is because it was roughly "1 second worth of data". In order to prevent this test from taking 5 seconds, I do 5 tests each 1/5 the size. I could make 'int trials = 5' variable if that would make it more clear.

PR #5146, opened 2018-12-07
interop-testing: Simply ProxyTest and shutdown handling

ejona86:
This is low priority


interop-testing/src/test/java/io/grpc/testing/integration/ProxyTest.java
```diff
@@ -203,43 +181,46 @@ public int init() throws IOException {
...
-          }
-        } else if (mode.equals("stream")) {
-          serverIn.read(response);
-          byte[] message = new byte[16 * 1024];
-          while (!shutDown) {
-            serverOut.write(message, 0, message.length);
-          }
-          serverIn.close();
-          serverOut.close();
+        Socket rcv = server.accept();
```
dapengzhang0: Are we allow to use try-with-resource now? At least in tests.

dapengzhang0: It seems we can override sourceCompatibility in test
```
compileTestJava {
  sourceCompatibility = "1.8"
}
```

ejona86:
@dapengzhang0, friendly ping

PR #5148, opened 2018-12-08
core: add LoadBalancer.canHandleEmptyAddressListFromNameResolution()

zhangkun83:
Currently ManagedChannelImpl will interpret empty address list from
NameResolver as an error, and LoadBalancer will NEVER receive an empty
list from handleResolvedAddressGroups().  There is a case in the
request-routing design where a LoadBalancer only receives service
config with which it constructs children NameResolver/LoadBalancer
pairs, thus no address is expected at this level.

canHandleEmptyAddressListFromNameResolution() is a signal to Channel
(and to the parent LoadBalancer in case of hierachical LoadBalancers)
about whether it accepts empty address lists.  The default is false,
which is the current behavior.

The logic is currently duplicated in ManagedChannelImpl and
AutoConfiguredLoadBalancer.  The one in ManagedChannelImpl will be
removed once we delete ManagedChannelBuilder.loadBalancerFactory() as
AutoConfiguredLoadBalancer will always be the top-level LoadBalancer
by then.

### DIFF
--- a/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
+++ b/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
@@ -111,7 +111,6 @@ public final class AutoConfiguredLoadBalancerFactory extends LoadBalancer.Factor
       try {
         selection = decideLoadBalancerProvider(servers, configMap);
       } catch (PolicyException e) {
-        System.err.println("XXX: " + e);
         Status s = Status.INTERNAL.withDescription(e.getMessage());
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -111,6 +111,7 @@ public void handleResolvedAddressGroups(
       try {
         selection = decideLoadBalancerProvider(servers, configMap);
       } catch (PolicyException e) {
+        System.err.println("XXX: " + e);
```
ejona86: Remove

PR #5164, opened 2018-12-13
examples: polish metadata example

zhangkun83:
- Added trailer operations
- Moved from `header` to `metadata`
  - The interceptors now capture received metadata instead of logging them.
   - On the client-side, received metadata is checked by application
   - On the server-side, received headers is injected into Context and checked by application
- Moved away from mocks in the tests
- Generate binaries for the example client and server. They have been verified to be working.

### DIFF
--- a/examples/src/main/java/io/grpc/examples/metadata/CustomMetadataClient.java
+++ b/examples/src/main/java/io/grpc/examples/metadata/CustomMetadataClient.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.grpc.examples.header;
+package io.grpc.examples.metadata;
 
 import io.grpc.Channel;
 import io.grpc.ClientInterceptor;
--- a/examples/README.md
+++ b/examples/README.md
@@ -15,7 +15,7 @@ before trying out the examples.
 
 - [Route guide](src/main/java/io/grpc/examples/routeguide)
 
-- [Metadata](src/main/java/io/grpc/examples/header)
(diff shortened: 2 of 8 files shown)


examples/src/main/java/io/grpc/examples/metadata/CustomMetadataClient.java
```diff
@@ -59,6 +59,7 @@ private void shutdown() throws InterruptedException {
    * A simple client method that like {@link io.grpc.examples.helloworld.HelloWorldClient}.
    */
   private void greet(String name) {
+    interceptor.outgoingHeader.set("Value of client->server header");
```
ejona86: I don't think we should encourage behavior like this. And if we did do behavior like this, we should make a new interceptor each call.

I think we need to have a conversation about what this example should do, because right now I see most of the code being for things I don't want our users doing just "for the sake of example."

Darkest: Ive been struggling with things implemented in this PR for several days.
Could you please write your opinion on how the following logic should be implemented using the library?

Let's say we have client and a server and we want:

- enrich each client request to server with unique id(let it be UUID.random())
- enrich each client request to server with user information, such as security token
- enrich each server response with arbitrary Map[String, String] with different values for each call
- read all the data mentioned above on both sides

What is the recommended way for doing this? Can it be added to examples and docs?

For now all I see on the web suggests I should create a new stub, or add interceptors for each call, is it adequate in terms of performance and jvm resources usage(new interceptor instances created for each call)

Issue #5113, opened 2018-12-04
Make proxy configuration channel specific


idelvall:
We are using gRPC in a java agent that is attached to our customers JVMs. The agent communicates via gRPC to our server for sending app usage metrics. 

The principal contract for our agent is to not adding any side effect into the instrumented application.

We have faced a customer environment, where the outgoing gRPC connection needs to go though a HTTPS proxy, in order to reach the internet. The customer application, on the other hand, establishes outgoing HTTP connections inside their local network that don't go though that proxy.

Given that gRPC proxy configuration is done via the standard system properties (https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html) that are system wide, we are unable to properly configure the gRPC client to use the proxy without adding the (terminal) side effect of forcing the other HTTP outgoing connections to go though that proxy.

This could be solved if the proxy configuration could be done specific to the gRPC channel.

- Any idea if/on-how i can make this work right now?
- Would you consider reducing the scope of the configuration to the channel instead using system wide configuration? If yes, i could work on it if you give me some guidance

Note: Proxy authenticator config is also system wide

### What version of gRPC are you using?
1.13.1

idelvall:
Ok I was able to resolve this. I answer myself so maybe someone can benefit from this in the future.

After looking at the code I realized that this can be achieved by two means:
- By passing a `ProxySocketAddress` to the channel builder, for example via [`NettyChannelBuilder.forAddress(SocketAddress serverAddress)`](https://github.com/grpc/grpc-java/blob/b5acbedd55ff80142c5480e210cf452d81fdaf00/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java#L93)
- Or preferably, by registering (via the system loader mechanism) a custom [`DnsNameResolverProvider` ](https://github.com/grpc/grpc-java/blob/b7dc501bbe2a71273b782931d0e96d528946300a/core/src/main/java/io/grpc/internal/DnsNameResolverProvider.java) with max priority (`10`)  that uses a custom proxy detector, rather than the default one `GrpcUtil.getDefaultProxyDetector()` (that is configured via the system-wide standard parameters).

idelvall:
If that information is accurate and complete, it would be a good idea to add it to the documentation

idelvall:
Thanks, yes that is exactly the same issue. I will resume this effort and work on a PR

zhangkun83:
Closed in favor of #5279

Issue #5138, opened 2018-12-07
Kokoro fails with address already in use


creamsoup:
https://source.cloud.google.com/results/invocations/9ea09e43-6e49-4ded-a7be-0e4ce849fb96 (windows)

```
io.grpc.okhttp.OkHttpTransportTest > openStreamPreventsTermination FAILED
    java.io.IOException: Failed to bind
        Caused by:
        java.net.BindException: Address already in use: bind
```

i've been observed this on many platform, on both netty and okhttp.

larry-safran:
Not happening anymore

Issue #4858, opened 2018-09-13
Allow custom NameResolvers to support CONNECT Proxies


ejona86:
This requires `ProxySocketAddress` being public. It seems we'll also need to inject `ProxyDetector` to NameResolvers.

ejona86:
API review discussion:
ProxyDetector to return SocketAddress; Rename ProxySocketAddress to HttpConnectProxySocketAddress. Implement equality and toString(). Merge ProxyParameters into HttpConnectProxySocketAddress. Use Builder for HttpConnectProxySocketAddress

Related: #5279

creamsoup:
i believe this is fixed by #5344

PR #5173, opened 2018-12-17
Allow specifying a ProxyDetector per ManagedChannel

idelvall:
Fixes https://github.com/grpc/grpc-java/issues/5113

This PR is similar to [this other one](https://github.com/grpc/grpc-java/pull/3957) but exposes `ProxyDetector` in public API rather than `java.net.Proxy`. This makes implementation cleaner, and also has the advantage of supporting proxy authorization configuration.

cc @lukaszx0 @ejona86 @carl-mastrangelo


idelvall:
Signed CLA

idelvall:
Ok, just read about `skipCodegen=true`, let's run the tests locally....

idelvall:
I just realized `ProxyParams` also need to be moved to the public API.
I will work tomorrow on it.

core/src/main/java/io/grpc/ProxyDetector.java
```diff
@@ -14,8 +14,9 @@
  * limitations under the License.
  */
 
-package io.grpc.internal;
+package io.grpc;
 
+import io.grpc.internal.ProxyParameters;
```
carl-mastrangelo: ProxyParameters would need to to be promoted to io.grpc, since io.grpc cannot depend on io.grpc.internal.

core/src/main/java/io/grpc/NameResolver.java
```diff
@@ -120,6 +120,27 @@ public void refresh() {}
...
+        ProxyDetector proxyDetector, Factory factory) {
+      if (factory instanceof ProxyAwareFactory) {
+        return ((ProxyAwareFactory)factory).newNameResolver(targetUri, params, proxyDetector);
+      } else {
+        return factory.newNameResolver(targetUri, params);
+      }
+    }
+  }
+
+  public static interface ProxyAwareFactory {
```
carl-mastrangelo: We typically shy away from interfaces on the public API because they aren't easy to modify later.   This would probably need to become an abstract class, refactored to compose with the nameresolver.

idelvall: I see. My intention was to add the new method to the factory instead of changing the current one, and breaking the API compatibility. Pushed another approach that keeps the same goal without the interface

idelvall:
Done changes, waiting for your review

idelvall:
@carl-mastrangelo I think it's ready, can you TAL? 
Btw seems like the "Macos" test doesn't report status, it's that normal?

creamsoup:
kokoro not reporting back to github is marked as fixed yesterday. but maybe not deployed yet. i'll manually trigger the kokoro

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -530,12 +533,15 @@ ClientStream newSubstream(ClientStreamTracer.Factory tracerFactory, Metadata new
...
+      final TimeProvider timeProvider,
+      ProxyDetector proxyDetector) {
     this.target = checkNotNull(builder.target, "target");
     this.nameResolverFactory = builder.getNameResolverFactory();
     this.nameResolverParams = checkNotNull(builder.getNameResolverParams(), "nameResolverParams");
-    this.nameResolver = getNameResolver(target, nameResolverFactory, nameResolverParams);
+    this.nameResolver = getNameResolver(target, nameResolverFactory, nameResolverParams,
+        proxyDetector);
     this.timeProvider = checkNotNull(timeProvider, "timeProvider");
+    this.proxyDetector = proxyDetector;
```
ejona86: Based on my comments in the BuilderImpl, this should get a checkNotNull.

ejona86: This should be `builder.proxyDetector`. We "simplified" it to avoid passing all the parameters individually. That would also mean you would revert some of the changes to the tests.

idelvall:
Addressed your comments @ejona86. I was not aware of the `Attributes` API, it's much cleaner now

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -524,6 +525,18 @@ public T maxTraceEvents(int maxTraceEvents) {
     throw new UnsupportedOperationException();
   }
 
+  /**
+   * Sets the proxy detector to be used in addresses name resolution. If <code>null</code> is passed
+   * default proxy detector {@link GrpcUtil#getDefaultProxyDetector()} will be used
```
ejona86: nit: just "default proxy detector". The GrpcUtil thing is internal, and it isn't really pertinent, so let's not mention it. If you _really_ wanted, you could mention that ProxySelector/Authenticator will be used instead. But meh.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -602,6 +606,19 @@ public CallTracer create() {
     channelz.addRootChannel(this);
   }
 
+  private static Attributes addProxyToAttributes(ProxyDetector proxyDetector,
+      Attributes attributes) {
+    if (attributes == null) {
```
ejona86: Nit: won't ever be null, so remove. And if it was null, I'd sort of be happy if it crashed so we could see where the faulty value was coming from (but there's a checkNotNull already in place).

ejona86:
@carl-mastrangelo, would you mind merging after review?

For posterity: ProxyDetector was likely to be exposed already for #4858. While I don't get excited about making a new public API for proxies, it is a _much_ simpler API than any of the obvious alternatives (ProxySelector/Authenticator, I'm looking at you). I sort of assume @zhangkun83 and I will do a look-over of the API after the new year. The API is simple enough that an interface doesn't worry me too much, although it is obviously non-extensible.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -17,6 +17,7 @@
 package io.grpc;
 
 import com.google.common.base.Preconditions;
+import io.grpc.internal.GrpcUtil;
```
ejona86: The (Bazel) CI is failing because of this. We have a rule not to have io.grpc depend on io.grpc.internal (long-term we have hopes of splitting out a grpc-api which just contains io.grpc).

ejona86: Fixed. (Since you can't trigger the CI, I figured it would just be easiest if I fixed this up.)

core/src/main/java/io/grpc/ManagedChannelBuilder.java
ejona86: Fixed (to remove reference on `internal`)

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
idelvall: done

core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java
```diff
@@ -258,12 +260,15 @@ private ProxyParameters detectProxy(InetSocketAddress targetAddr) throws IOExcep
     }
 
     if (auth == null) {
-      return new ProxyParameters(resolvedProxyAddr, null, null);
+      return ProxyParameters.forAddress(resolvedProxyAddr).build();
     }
 
     // TODO(spencerfang): users of ProxyParameters should clear the password when done
```
idelvall: I changed `ProxyParameters` to use [`char[]` for passwords](https://stackoverflow.com/questions/8881291/why-is-char-preferred-over-string-for-passwords) for this reason. 
What remains now in this direction is propagating this change to the Netty and OKHttp transports (the first one seems to expose a `String` password parameter in its public API)

idelvall:
Last failure seems due to a known issue https://github.com/grpc/grpc-java/issues/5138

idelvall:
Done. Also changed password type to char array

core/src/main/java/io/grpc/ProxyParameters.java
```diff
@@ -51,11 +72,48 @@ public boolean equals(Object o) {
...
+    private Builder(InetSocketAddress proxyAddress) {
+      this.proxyAddress = Preconditions.checkNotNull(proxyAddress, "proxyAddress");
+    }
+
+    public Builder username(String username) {
+      this.username = username;
+      return this;
+    }
+
+    public Builder password(char[] password) {
```
zhangkun83: Everything on the propagation path, down to netty, would need to switch to `char[]` to achieve the goal of erasing the password after use. All components along the path would also need to coordinate to make sure the `byte[]` will not be used after it's erased, and I am a little worried that it may become too invasive of a change, while it's unclear to me how much more security it will bring. @ejona86 and @carl-mastrangelo what's your opinions?

carl-mastrangelo: I agree with @zhangkun83 here.  It looks like the password is stored inside the Netty handler down the line, so deleting our ref is not sufficient.  Also, I think the bugs introduced by having a mutable array are more likely than the password field being leaked (this isn't C++, after all).

idelvall: Yeah, I agree too. Too much changes for an unclear benefit. I will undo the `char[]` changes on Monday

idelvall:
Done. Password is of type `String` now.
Also removed [TODO about erasing passwords](https://github.com/grpc/grpc-java/blob/5b17b2d6537584536cfadf43c8be1830b075a4b3/core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java#L266) cc/ @zpencer

core/src/main/java/io/grpc/NameResolver.java
```diff
@@ -99,14 +99,21 @@ public void refresh() {}
     public static final Attributes.Key<Integer> PARAMS_DEFAULT_PORT =
         Attributes.Key.create("params-default-port");
 
+    /**
+     * Proxy detector used in name resolution.
+     */
+    @ExperimentalApi("https://github.com/grpc/grpc-java/issues/5113")
```
carl-mastrangelo: @zhangkun83 should this be a TransportAttr?

core/src/main/java/io/grpc/ProxyParameters.java
```diff
@@ -24,13 +24,19 @@
...
  */
+@ExperimentalApi("https://github.com/grpc/grpc-java/issues/5113")
 public final class ProxyParameters {
-  public final InetSocketAddress proxyAddress;
-  @Nullable public final String username;
-  @Nullable public final String password;
 
-  /** Creates an instance. */
-  public ProxyParameters(
+  private final InetSocketAddress proxyAddress;
```
carl-mastrangelo: This seems like it needs to be a SocketAddress, rather than an InetSocketAddress.  If there was a UDS http proxy, it wouldn't be able to use this.

core/src/main/java/io/grpc/ProxyParameters.java
```diff
@@ -58,4 +78,41 @@ public boolean equals(Object o) {
   public int hashCode() {
     return Objects.hashCode(proxyAddress, username, password);
   }
+
+  /**
+   * Create a new builder.
+   */
+  public static Builder forAddress(InetSocketAddress proxyAddress) {
```
carl-mastrangelo: Same comment here.  It seems this needs to be a SocketAddress.

core/src/main/java/io/grpc/ProxyParameters.java
```diff
@@ -58,4 +78,41 @@ public boolean equals(Object o) {
...
+    @Nullable
+    private String username;
+    @Nullable
+    private String password;
+
+    private Builder(InetSocketAddress proxyAddress) {
+      this.proxyAddress = Preconditions.checkNotNull(proxyAddress, "proxyAddress");
+    }
+
+    public Builder username(String username) {
```
carl-mastrangelo: `@Nullable` ?

core/src/main/java/io/grpc/internal/ProxyDetectorImpl.java
```diff
@@ -258,12 +260,14 @@ private ProxyParameters detectProxy(InetSocketAddress targetAddr) throws IOExcep
...
+      return ProxyParameters.forAddress(resolvedProxyAddr).build();
     }
 
-    // TODO(spencerfang): users of ProxyParameters should clear the password when done
-    return new ProxyParameters(
-        resolvedProxyAddr, auth.getUserName(), new String(auth.getPassword()));
+    return ProxyParameters
+        .forAddress(resolvedProxyAddr)
+        .username(auth.getUserName())
+        .password(new String(auth.getPassword()))
```
carl-mastrangelo: hmmm, do you know why user name can be null, but not password?

idelvall: fixed

core/src/main/java/io/grpc/ProxyParameters.java
idelvall: fixed

core/src/main/java/io/grpc/ProxyParameters.java
idelvall: Ok, after reading this https://github.com/grpc/grpc-java/issues/1539 I am starting to understand...

idelvall: Addressed too

core/src/main/java/io/grpc/NameResolver.java
idelvall: Waiting for your feedback on this

core/src/main/java/io/grpc/ProxyParameters.java
ejona86: :-/ It should have stayed a InetSocketAddress, since this is for CONNECT and CONNECT requires IP addresses.

carl-mastrangelo: @ejona86 when it was an internal detail, yeah, but now that it is becomign public, it should be based on the transport.  A different transport may not use InetSocketAddress.

idelvall: > ... CONNECT requires IP addresses

But, technically the connection from client to the proxy could still be over UDS and the contents sent use the correct HTTP CONNECT semantics (with a destination IP address), so the connection from the proxy to server being over TPC right?

> ... it should be based on the transport. A different transport may not use InetSocketAddress.

Not only in the transport, given that the proxy is an external entity. 

I think the key question is: is there any HTTP proxy implementation that supports UDS incoming connections? So far, I couldn't find any

ejona86: Oh, right, this is the proxy's address. Okay, fair, it can be a `SocketAddress` in some ways. But this will cause havoc with #3085 since a NameResolver would no longer know which address types it might return, as ProxyDetector chooses this proxy address. This wasn't a problem before because proxy selection would only happen for addresses that could have been InetSocketAddress. (That is to say, a NameResolver that only cares about UnixDomainSocketAddress wouldn't even need to consult the ProxyDetector because a proxy will always use inet addresses to connect to the end host.) So even we want to support HTTP proxies for UDS incoming connections, I don't know how we could support it because there's no real way to choose a transport that supports UDS at that point. Even telling the ProxyDetector the permitted address types doesn't really solve it because the dns resolver wouldn't support UDS.

But I guess we can just have a "don't do that" policy. `java.net.Proxy` does have a `SocketAddress`, although it also is "extensible" as there could be new proxy types added in the future.

Maybe we can just say that `ProxyDetector.proxyFor()` should return the same `SocketAddress` type as provided. We'd add a new method in the future if we needed something more advanced.

I really didn't want to get into all this... The API still needs work and we agree it needs work, but there will be very few users of this part of the API (they will receive/pass ProxyParameters, but they won't really create or inspect them).

carl-mastrangelo: I think we can punt on this if ProxyDetector becomes an abstract class.  I'm not certain that we can enforce the socket types being the same, since InetSocketAddress is nonfinal.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -493,14 +493,21 @@ public void close() {}
...
+              throw new AssertionError();
+            } else if (proxy.getProxyAddress() instanceof InetSocketAddress) {
+              sock = createHttpProxySocket(
+                  address,
+                  (InetSocketAddress)proxy.getProxyAddress(),
+                  proxy.getUsername(),
+                  proxy.getPassword()
+              );
+            } else {
+              throw new Error(
```
ejona86: This should not be an error. Given the catch that's a bit lower, you can do `throw Status.INTERNAL.withDescription("Unsupported SocketAddress implementation " + proxy.getProxyAddress().getClass()).asException();`

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -493,14 +493,21 @@ public void close() {}
...
+              // This should never happen, case added for readability
+              throw new AssertionError();
+            } else if (proxy.getProxyAddress() instanceof InetSocketAddress) {
+              sock = createHttpProxySocket(
+                  address,
+                  (InetSocketAddress)proxy.getProxyAddress(),
+                  proxy.getUsername(),
+                  proxy.getPassword()
+              );
+            } else {
```
ejona86: Make this an if above the createHttpProxySocket, that way there's no need for the else:

```
if (!(proxy.getProxyAddress() instanceof InetSocketAddress)) {
  throw ...;
}
sock = createHttpProxySocket(...):
```

idelvall: added these last changes @ejona86

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -493,14 +493,21 @@ public void close() {}
           if (proxy == null) {
             sock = new Socket(address.getAddress(), address.getPort());
           } else {
-            sock = createHttpProxySocket(
-                address,
-                proxy.getProxyAddress(),
-                proxy.getUsername(),
-                proxy.getPassword()
-            );
+            if (proxy.getProxyAddress() == null) {
```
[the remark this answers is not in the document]
idelvall: It's curious, i thought it was indeed avoiding it.... Removed

core/src/main/java/io/grpc/ProxyParameters.java
ejona86: The nullable here is still in the wrong location. Ditto below.

core/src/main/java/io/grpc/ProxyParameters.java
```diff
@@ -58,4 +81,43 @@ public boolean equals(Object o) {
...
+    private String username;
+    @Nullable
+    private String password;
+
+    private Builder(SocketAddress proxyAddress) {
+      this.proxyAddress = Preconditions.checkNotNull(proxyAddress, "proxyAddress");
+    }
+
+    @Nullable
+    public Builder username(String username) {
```
ejona86: Arg my other comment is showing the wrong code. Re-creating to show `@Nullable` is in wrong spot. Ditto below.

ejona86:
@carl-mastrangelo or @zhangkun83, PTAL

core/src/main/java/io/grpc/ProxyParameters.java
idelvall: Oh, yes, I missed that. Fixed

core/src/main/java/io/grpc/ProxyParameters.java
```diff
@@ -14,35 +14,58 @@
...
+  private ProxyParameters(
+      SocketAddress proxyAddress,
       @Nullable String username,
       @Nullable String password) {
     Preconditions.checkNotNull(proxyAddress);
     // The resolution must be done by the ProxyParameters producer, because consumers
     // may not be allowed to do IO.
-    Preconditions.checkState(!proxyAddress.isUnresolved());
+    if (proxyAddress instanceof InetSocketAddress) {
+      Preconditions.checkState(!((InetSocketAddress)proxyAddress).isUnresolved());
```
carl-mastrangelo: Sorry to be a hassle, but this should be a checkArgument, not checkState

core/src/main/java/io/grpc/internal/DnsNameResolverProvider.java
```diff
@@ -49,12 +51,16 @@ public DnsNameResolver newNameResolver(URI targetUri, Attributes params) {
       Preconditions.checkArgument(targetPath.startsWith("/"),
           "the path component (%s) of the target (%s) must start with '/'", targetPath, targetUri);
       String name = targetPath.substring(1);
+      ProxyDetector proxyDetector = params.get(Factory.PARAMS_PROXY_DETECTOR);
+      if (proxyDetector == null) {
+        proxyDetector = GrpcUtil.NOOP_PROXY_DETECTOR;
```
carl-mastrangelo: GrpcUtil.getDefaultProxyDetector  ?

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -493,10 +493,19 @@ public void close() {}
...
+              sock = createHttpProxySocket(
+                  address,
+                  (InetSocketAddress)proxy.getProxyAddress(),
+                  proxy.getUsername(),
+                  proxy.getPassword()
+              );
+            } else {
+              throw Status.INTERNAL.withDescription(
+                  "Unsupported SocketAddress implementation " + proxy.getProxyAddress().getClass())
+                  .asException();
```
carl-mastrangelo: asRuntimeException

core/src/main/java/io/grpc/internal/DnsNameResolverProvider.java
ejona86: I had suggested the NOOP_PROXY_DETECTION here as a temporary thing (some logic was moved from the Builder). I figured @zhangkun83 (or whomever does #4159) would remove it when we replace the Attributes with a concrete class.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
ejona86: It has to be asException(), because of the catch it is within.

core/src/main/java/io/grpc/internal/DnsNameResolverProvider.java
idelvall: No problem @ejona86 , applied @zhangkun83 suggestions

idelvall:
Applied latest suggestions. Waiting for your feedback

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -524,6 +524,18 @@ public T maxTraceEvents(int maxTraceEvents) {
     throw new UnsupportedOperationException();
   }
 
+  /**
+   * Sets the proxy detector to be used in addresses name resolution. If <code>null</code> is passed
+   * the default proxy detector will be used.
+   *
+   * @return this
+   * @since 1.18.0
```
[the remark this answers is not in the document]
ejona86: Okay. I fixed that. Merging after tests complete.

ejona86:
@idelvall, thank you! I'm sorry that turned into so much back-and-forth!

idelvall:
@ejona86 np, glad to have added my two cents.
Thanks to you all for sharing this great project!

PR #5175, opened 2018-12-17
core,netty: refactor server builder

carl-mastrangelo:
This change:

* makes Census tracing factories at the end of the user added ones
* makes more vars in AbstractServerImplBuilder package private
* annotates methods in ASIB to be clearer
* simplifies several of the setters to be single line
* Makes the generics on the Tracer factories proper


carl-mastrangelo:
This change is a precursor to normalizing the construction of servers.

PR #5184, opened 2018-12-19
services: add BUILD.bazel

Ubehebe:
At HEAD, the java sources in services/ are not accessible from any bazel target. Most top-level directories have both a BUILD.bazel and a build.gradle, but services/ has only build.gradle.

(Note that https://github.com/grpc/grpc-java/pull/5183 must be merged before this can be merged, to fix the re2j jar.)

### DIFF
--- a/services/BUILD.bazel
+++ b/services/BUILD.bazel
@@ -1,0 +1,24 @@
+java_library(
+    name = "services",
+    srcs = glob([
+        "src/main/java/io/grpc/**/*.java",
+        "src/generated/main/**/*.java",
+    ]),
+    visibility = [
+        "//visibility:public",
+    ],
+    deps = [
+        "//context",
+        "//core",
+        "//core:internal",
+        "//core:util",
+        "//protobuf",
+        "//stub",
+        "@com_google_code_findbugs_jsr305//jar",
+        "@com_google_guava_guava//jar",
+        "@com_google_protobuf//:protobuf_java",
+        "@com_google_protobuf//:protobuf_java_util",
+        "@javax_annotation_javax_annotation_api//jar",
+        "@com_google_re2j//jar",
+    ],
+)
--- a/services/BUILD.bazel
+++ b/services/BUILD.bazel
@@ -1,11 +1,50 @@
+package(default_visibility = ["//visibility:public"])
+
 java_library(
-    name = "services",
-    srcs = glob([
-        "src/main/java/io/grpc/**/*.java",
-        "src/generated/main/**/*.java",
-    ]),
-    visibility = [
-        "//visibility:public",
+    name = "binarylog",
+    srcs = [
+        "src/main/java/io/grpc/services/BinaryLogProvider.java",
+        "src/main/java/io/grpc/services/BinaryLogProviderImpl.java",
+        "src/main/java/io/grpc/services/BinaryLogSink.java",
+        "src/main/java/io/grpc/services/BinlogHelper.java",
+        "src/main/java/io/grpc/services/InetAddressUtil.java",
+        "src/main/java/io/grpc/services/TempFileSink.java",
+    ],
+    deps = [
+        "//context",
+        "//core",
+        "//services/src/main/proto/grpc/binlog/v1:binarylog_java_proto",
+        "@com_google_code_findbugs_jsr305//jar",
+        "@com_google_guava_guava//jar",
+        "@com_google_protobuf//:protobuf_java",
+        "@com_google_protobuf//:protobuf_java_util",
+        "@com_google_re2j//jar",
+    ],
+)
+
+java_library(
+    name = "channelz",
+    srcs = [
+        "src/main/java/io/grpc/services/ChannelzProtoUtil.java",
+        "src/main/java/io/grpc/services/ChannelzService.java",
+    ],
+    deps = [
+        "//context",
+        "//core",
+        "//services/src/main/proto/grpc/channelz/v1:channelz_grpc",
+        "//services/src/main/proto/grpc/channelz/v1:channelz_java_proto",
+        "//stub",
+        "@com_google_code_findbugs_jsr305//jar",
+        "@com_google_guava_guava//jar",
+        "@com_google_protobuf//:protobuf_java",
+        "@com_google_protobuf//:protobuf_java_util",
+    ],
+)
+
+java_library(
+    name = "reflection",
+    srcs = [
+        "src/main/java/io/grpc/protobuf/services/ProtoReflectionService.java",
     ],
     deps = [
         "//context",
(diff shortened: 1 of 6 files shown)


services/BUILD.bazel
```diff
@@ -0,0 +1,24 @@
+java_library(
+    name = "services",
+    srcs = glob([
+        "src/main/java/io/grpc/**/*.java",
+        "src/generated/main/**/*.java",
```
carl-mastrangelo: We usually dont link to generated sources, since bazel usually  generates the the source ephemerally.

services/BUILD.bazel
```diff
@@ -0,0 +1,24 @@
+java_library(
+    name = "services",
+    srcs = glob([
+        "src/main/java/io/grpc/**/*.java",
```
carl-mastrangelo: This is a little too widely scoped.   While each of the services in are grouped into this directory, they are mostly independent.  For example, we have the following java_libraries internally:

* ProtoReflectionService.java
* ChannelzProtoUtil.java, ChannelzService.java
* binary logging

Splitting up helps silo dependencies and speed up builds.   It also prevents files from accidentally get pulled in, which `**` does.

services/BUILD.bazel
Ubehebe: Good point. I added proto_library, java_proto_library, and java_grpc_library targets in c91d2004b49e2f3262abb0ac843c004f0bbfb6c8, so that the top-level services targets can depend on them instead of the gencode. PTAL.

services/BUILD.bazel
Ubehebe: Done. The services package now exposes :binarylog, :channelz, and :reflection java_library targets.

dapengzhang0:
@Ubehebe I think should place `services/src/main/proto/grpc/binlog/v1/BUILD.bazel` at `services/src/main/proto/BUILD.bazel`. Because: if there are multiple proto files in the directory importing one another, the former will not work. Though right now there is only single proto file. Same for other BUILD.bazel files.

services/BUILD.bazel
```diff
@@ -0,0 +1,65 @@
...
+        "src/main/java/io/grpc/services/BinaryLogProviderImpl.java",
+        "src/main/java/io/grpc/services/BinaryLogSink.java",
+        "src/main/java/io/grpc/services/BinlogHelper.java",
+        "src/main/java/io/grpc/services/InetAddressUtil.java",
+        "src/main/java/io/grpc/services/TempFileSink.java",
+    ],
+    deps = [
+        "//context",
+        "//core",
+        "//services/src/main/proto/grpc/binlog/v1:binarylog_java_proto",
```
ejona86: Instead of creating these BUILD files, we actually want to refer to the ones in https://github.com/grpc/grpc-proto , right @carl-mastrangelo? We would need the java_proto_library and java_grpc_library rules in our repo, though.

ejona86:
(Note: if a particular proto is not in the grpc-proto repo, it is fine to depend on the proto in grpc-java; I was mainly just referring to those protos that exist in grpc-proto.)

services/BUILD.bazel
Ubehebe: Adding a dependency on grpc-proto would introduce the risk that the bazel and gradle builds of grpc-java would compile against different versions of the grpc protos. I think this would be a confusing situation (I can imagine people filing bugs that some API worked in the bazel release but not the gradle one, or vice versa).

I don't have visibility into how you're generating the gencode when you check it into grpc-java. Judging from the generated/ commit history, it's a manual process. I think before we can add a dep on grpc-proto, we need a mechanism to ensure that the checked-in gencode is always in sync with the grpc-proto version. Thoughts?

ejona86: > Adding a dependency on grpc-proto would introduce the risk that the bazel and gradle builds of grpc-java would compile against different versions of the grpc protos.

In some ways, yes. In some ways no. These protos must remain backward compatible for many reasons, so there should be limited risk of breakage. Also, internally at Google we use the grpc-proto version (not the grpc-java copy), and this isn't all that different.

> I think before we can add a dep on grpc-proto, we need a mechanism to ensure that the checked-in gencode is always in sync with the grpc-proto version.

This is already done as part of our .travis.yml (we run `test -z "$(git status --porcelain)"`).

The process for copying the protos is still improving, and we should eventually have a script that will copy over the current version of grpc-proto. We can then make sure to bump the bazel version at the same time we run that script. So the concerns you raised are more short-term and things aren't even that poor short-term.

ejona86: Created script to sync protos in #5190

Ubehebe: OK, I think depending on grpc-proto is reasonable. I looked into this and came up with three problems.

1. The `:reflection` target in the PR doesn't compile against grpc-proto's [:reflection_proto](https://github.com/grpc/grpc-proto/blob/master/BUILD.bazel#L108). That's because the reflection proto had a breaking rewrite (in particular, its package changed from grpc.reflection.v1alpha to grpc.reflection.v1). [ProtoReflectionService still imports the v1alpha gencode](https://github.com/grpc/grpc-java/blob/81da3eb95be37fa0647ce8da2e19de96ab84c600/services/src/main/java/io/grpc/protobuf/services/ProtoReflectionService.java#L34), so it's broken.
1. The `:binarylog` target in the PR doesn't compile against grpc-proto's [:binarylog_proto](https://github.com/grpc/grpc-proto/blob/master/BUILD.bazel#L89). The binarylog proto has also had a package-breaking rewrite (from grpc.binarylog.v1alpha to grpc.binarylog.v1), but in this case grpc-java uses a newer version than what grpc-proto exposes ([example](https://github.com/grpc/grpc-java/blob/41c8d8020f787a1f898a08177e1cd428a60512c2/services/src/main/java/io/grpc/services/BinlogHelper.java#L55)).
1. The `:channelz` target in the PR compiles against grpc-proto's channelz, but generates a warning:

```
DEBUG: grpc-java/java_grpc_library.bzl:11:7: in srcs attribute of //services:channelz_grpc__do_not_reference__srcjar: Proto source with label @io_grpc_grpc_proto//:channelz_proto should be in same package as consuming rule
```

This comes from the [java_grpc_library rule implementation](https://github.com/grpc/grpc-java/blob/41c8d8020f787a1f898a08177e1cd428a60512c2/java_grpc_library.bzl#L11). This happens because I defined a java_grpc_library target in the services package to wrap grpc-proto's channelz library.

For 1, I suggest checking the new version of the reflection gencode into this repo and updating package references. (grpc-proto still exposes a target for the alpha version of the reflection proto, but depending on that seems like it would just create more work down the line.)

For 2, I suggest changing grpc-proto's binarylog target to point to the new version. I'll start a discussion at that repo.

For 3, I'm not sure if this is a real problem. The debug message doesn't explain why java_grpc_library targets need to be in the same package as their proto_library targets. (Defining the java_grpc_library in grpc-proto would create a circular dep anyway.)

ejona86: For (1), use `:reflection_proto_deprecated` instead of `:reflection_proto`.

For (2), we can use our copy here if you want. I don't want to make too much more work for you. But if you're happy updating that other repo, go for it. (You can CC us on an PR; we have some control in that repo as well.)

For (3), yes, it will generate that warning. That's expected. At some point I think I'll be removing that warning from java_grpc_library but for the while we can live with it.

perezd:
Friendly ping! Would love to see this merged!

dapengzhang0:
@perezd now you can use [grpc-proto](https://github.com/grpc/grpc-proto/blob/master/BUILD.bazel) for most of the services. Please resolve branch conflict.

perezd:
@Ubehebe Are you planning to resolve this? If not I can create a new PR.

dapengzhang0:
@Ubehebe Thank you for your PR, but #5384 has been done with grpc-proto, closing this one.

PR #5187, opened 2018-12-19
Mark CallOptions' deadline/executor as @Nullable

deadmoose:
They are, but weren't consistently marked like the other members


dapengzhang0:
@deadmoose Thank you! Will merge.

PR #5191, opened 2018-12-20
alts,services: Sync proto sources from grpc-proto

ejona86:
This syncs to grpc/grpc-proto@f9af5e19f73

----

The sync was performed with the script from #5190.


alts/src/main/proto/grpc/gcp/handshaker.proto
```diff
@@ -229,6 +229,6 @@ service HandshakerService {
   // service expects to respond. Client does not have to wait for service's
   // response before sending next request.
   rpc DoHandshake(stream HandshakerReq)
-    returns (stream HandshakerResp) {
+      returns (stream HandshakerResp) {
```
sanjaypujare: This diff is superfluous - due to whitespace changes. Can we get rid of it?

ejona86: This is a pure copy of the proto in https://github.com/grpc/grpc-proto. If we revert this whitespace change it will cause a pure-whitespace change in other repos.

dapengzhang0:
@ejona86 Sometimes I do want to check what grpc-proto version (the commint number in grpc-proto) this proto file in grpc-java is. Seems I can only find the commit number in your PR description?

ejona86:
Yes. We need to discuss how to see what version we are using, but I figured this is still a step forward as-is.

One option would be to specify the SHA in the sync-protos.sh instead of "master" (or in a file just containing the sha). I'm not sure what other ideas we have.

ejona86:

PR #5196, opened 2018-12-21
Downgrade to Guava 25.1

### LINKED ISSUE #5166: Downgrade Guava to 21.0?
It would be helpful for downstream consumers of grpc-java if the Guava dependency could be downgraded to a prior release. We are using grpc-java and have been happy with how it has integrated in with existing codebases with minimal dependencies. Starting with the 1.16.x release however, Guava was upgraded from 19.0 to 26.0-android. This complicates integration with existing codebases, as going from 19.0 to 26.0-android has removed quite a few older APIs which are still used in open source and internal code. Additionally, for our use cases we'd prefer to use the `-jre` variant instead of the `-android` variant.

We are using Maven's requireUpperBoundDeps (https://maven.apache.org/enforcer/enforcer-rules/requireUpperBoundDeps.html) in all of our projects to catch common dependency issues, which requires us to do one of two things:

* Add exclusions to grpc-java dependencies or disable warnings on Guava to quiet the check. This isn't great because if grpc-java starts depending on new APIs introduced in Guava, we'll fail at runtime with MethodNotFound/ClassNotFound exceptions.
* Update all third party and internal codebases to use Guava 26.0-android or later.

I've been able to successfully compile grpc-java with Guava 21.0. This is also the version in which no non-`@Beta` methods are removed going forward (https://github.com/google/guava/wiki/UseGuavaInYourBuild#how-do-i-avoid-conflicts-between-multiple-versions-of-guava). Would the grpc-java maintainers consider moving to this release to make it easier to integrate into existing codebases?

1.17.1

ejona86:
This is temporary for the 1.18 release to give users a bit more time to
get past the breaking changes to Beta APIs in Guava 26.

Fixes #5166

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -106,7 +106,9 @@ subprojects {
         javaPluginPath = "$rootDir/compiler/build/exe/java_plugin/$protocPluginBaseName$exeSuffix"
 
         nettyVersion = '4.1.32.Final'
-        guavaVersion = '26.0-android'
+        // Temporarily use a slightly older version of Guava. Will want to bump
+        // to 27.0.1+ with grpc 1.19 or 1.20 for the listenablefuture dep changes.
+        guavaVersion = '25.1-android'
         protobufVersion = '3.5.1'
         protocVersion = '3.5.1-1'
         protobufNanoVersion = '3.0.0-alpha-5'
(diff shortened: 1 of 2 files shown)


dapengzhang0:
@ejona86 Are you going to backport?

ejona86:
@dapengzhang0, yes. I started working on it this morning and then got sidetracked.

PR #5203, opened 2019-01-02
Auth: Make IOException non retriable

ajaaym:
Fix [#3573](https://github.com/googleapis/google-cloud-java/issues/3573) Making IOException non retriable. Since most of Auth implementation based on network are already retrying for common problems and there are more IOException that are not retriable.

### DIFF
--- a/auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
+++ b/auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
@@ -135,16 +135,11 @@ final class GoogleAuthLibraryCallCredentials extends CallCredentials2 {
 
       @Override
       public void onFailure(Throwable e) {
-        if (e instanceof IOException) {
-          // Since it's an I/O failure, let the call be retried with UNAVAILABLE.
-          applier.fail(Status.UNAVAILABLE
-              .withDescription("Credentials failed to obtain metadata")
-              .withCause(e));
-        } else {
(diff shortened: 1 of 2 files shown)


thelinuxfoundation:

* User @ajaaym isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

ajaaym:
I signed it

ejona86:
NAK. This was done on purpose in #3757 to fix #3267. This is cross-language gRPC behavior. The more common case is simply there was an I/O failure getting the token and retrying will fix it. If the client has no valid token then things are broken anyway; if an application-level retry forever retries, that seems like its own problem.

ajaaym:
@ejona86 Thanks for the review and background. Reason I thought to put back this change is bcz google-auth-library-java already retries the common cause so when it comes here its normally not-retriable. Closing this and will take up the discussion on google-auth-library-java.

ejona86:
@ajaaym, hmmm... That was added in https://github.com/googleapis/google-auth-library-java/pull/104 . I wasn't aware of that (I reviewed many of the initial changes in google-auth-library-java and was an old maintainer for google-{http,oauth,api}-java-client). Go ahead and include me in the conversation with google-auth-library-java. It seems the problem is that google-auth-library-java is retrying for 18 minutes, if I'm reading ExponentialBackOff correctly (see DEFAULT_MAX_ELAPSED_TIME_MILLIS and nextBackOffMillis(); 900000 ms is 18 minutes, not 15...).

PR #5209, opened 2019-01-03
Adding Bom

### LINKED ISSUE #3954: Create BOM for grpc-java
A BOM provides versions for multiple libraries to aid keeping the versions consistent.

See https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-bom/pom.xml and GoogleCloudPlatform/google-cloud-java#2307

creamsoup:
adapted from gax-java's BOM: googleapis/gax-java#607

some of examples with maven pom is also converted to use bom.

resolves #3954

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -34,6 +34,8 @@ $ VERSION_FILES=(
   build.gradle
   android/build.gradle
   android-interop-testing/app/build.gradle
+  bom/build.gradle
+  bom/pom.xml
   core/src/main/java/io/grpc/internal/GrpcUtil.java
   cronet/build.gradle
   documentation/android-channel-builder.md
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -34,8 +34,6 @@ $ VERSION_FILES=(
   build.gradle
   android/build.gradle
   android-interop-testing/app/build.gradle
-  bom/build.gradle
-  bom/pom.xml
   core/src/main/java/io/grpc/internal/GrpcUtil.java
   cronet/build.gradle
   documentation/android-channel-builder.md
--- a/bom/build.gradle
+++ b/bom/build.gradle
@@ -1,18 +1,5 @@
 description = 'gRPC: BOM'
 
-buildscript {
-    repositories {
-        mavenLocal()
-        maven {
-            url 'https://plugins.gradle.org/m2/'
-        }
-    }
-}
-
-archivesBaseName = "bom"
-
-project.version = "1.19.0-SNAPSHOT" // CURRENT_GRPC_VERSION
-
 publishing {
     publications {
         maven(MavenPublication) {
@@ -26,12 +13,13 @@ publishing {
                 def dependencyManagement = asNode().appendNode('dependencyManagement')
                 def dependencies = dependencyManagement.appendNode('dependencies')
                 rootProject.subprojects.each { subproject ->
-                    if (!internalProjects.contains(subproject.name)) {
-                        def dependencyNode = dependencies.appendNode('dependency')
-                        dependencyNode.appendNode('groupId', subproject.group)
-                        dependencyNode.appendNode('artifactId', subproject.name)
-                        dependencyNode.appendNode('version', subproject.version)
-                    }
+                    if (internalProjects.contains(subproject.name)) {
+		        return
+		    }
+                    def dependencyNode = dependencies.appendNode('dependency')
+                    dependencyNode.appendNode('groupId', subproject.group)
+                    dependencyNode.appendNode('artifactId', subproject.name)
+                    dependencyNode.appendNode('version', subproject.version)
                 }
 	    }
         }
--- a/bom/build.gradle
+++ b/bom/build.gradle
@@ -23,7 +23,8 @@ publishing {
                 // Generate bom using subprojects
                 def internalProjects = [project.name, 'grpc-xds', 'grpc-gae-interop-testing-jdk8']
 
-                def dependencies = asNode().appendNode('dependencies')
+                def dependencyManagement = asNode().appendNode('dependencyManagement')
+                def dependencies = dependencyManagement.appendNode('dependencies')
                 rootProject.subprojects.each { subproject ->
                     if (!internalProjects.contains(subproject.name)) {
                         def dependencyNode = dependencies.appendNode('dependency')
--- a/bom/build.gradle
+++ b/bom/build.gradle
@@ -1,0 +1,38 @@
+description = 'gRPC: BOM'
+
+buildscript {
+    repositories {
+        mavenLocal()
+        maven {
+            url 'https://plugins.gradle.org/m2/'
+        }
+    }
+}
+
+archivesBaseName = "bom"
+
+project.version = "1.19.0-SNAPSHOT" // CURRENT_GRPC_VERSION
+
+publishing {
+    publications {
+        maven(MavenPublication) {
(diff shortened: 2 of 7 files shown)


creamsoup:
can you take a close look at the scope of dependency, and also list of dependencies in BOM? 
I followed dropwizard like bom which includes all dependencies already in the build.gradle and all the modules.
for gradle projects, the dependencies of each module may not be needed? With gradle5 [officially support BOM](https://docs.gradle.org/5.0/userguide/managing_transitive_dependencies.html#sec:bom_import),  those can be useful if we consume this in the gradle build file.

bom/pom.xml
```diff
@@ -0,0 +1,374 @@
...
+        <artifactId>grpc-context</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.grpc</groupId>
+        <artifactId>grpc-stub</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.grpc</groupId>
```
dapengzhang0: nit: sort the list

dapengzhang0:
Is `<artifactId>maven-enforcer-plugin</artifactId>` in `examples/pom.xml` needed anymore when BOM is used?

bom/pom.xml
```diff
@@ -0,0 +1,374 @@
...
+      </dependency>
+      <dependency>
+        <groupId>com.google.code.gson</groupId>
+        <artifactId>gson</artifactId>
+        <version>2.7</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>${guava.version}</version>
```
TheIndifferent: Please do not pin versions of transitive dependencies, this will make the BOM useless because it will enforce everyone to use `andoid` and not `jre` version of Guava, Protobuf 3.5.x instead of 3.6.x and so on.

Please see Netty and Jackson as examples of well-formed BOMs:

https://github.com/netty/netty/blob/4.1/bom/pom.xml

https://github.com/FasterXML/jackson-bom/blob/master/pom.xml

ejona86: I agree we should only include `io.grpc` artifacts. For netty the general solution is to use grpc-netty-shaded. If someone specifies grpc-netty directly they will need to manage their netty version.

ejona86:
@dapengzhang0, maven-enforcer-plugin is still necessary after this to ensure the BOM is not accidentally downgrading a version, there may be mulitple BOMs, and because all Maven users should really be using requireUpperBoundDeps for their own sanity.

bom/pom.xml
garrettjonesgoogle: +1 to not including transitive dependencies - for the purposes of cloud-oss-bom (https://github.com/GoogleCloudPlatform/cloud-opensource-java/tree/master/boms/cloud-oss-bom), which combines multiple BOMs together, we need each BOM to be limited to the artifacts of a single library.

bom/build.gradle
```diff
@@ -0,0 +1,105 @@
...
+    mavenCentral()
+    jcenter()
+  }
+}
+
+archivesBaseName = "bom"
+
+project.version = "1.19.0-SNAPSHOT" // CURRENT_GRPC_VERSION
+
+apply plugin: 'maven-publish'
```
ejona86: :-/ Now we have both publishing mechanisms. That's not great, and would lead to lots of confusion. Does this mean we should begin the migration to maven-publish (which would be good)?

bom/build.gradle
```diff
@@ -0,0 +1,105 @@
...
+  archives(mavenJavaBomOutputFile) {
+    builtBy copyPom
+  }
+}
+
+afterEvaluate {
+  // We can't use the `publishing` section from the main build.gradle because
+  // we don't want all the Java artifacts, and we want to use our own pom.xml
+  // instead of the generated one.
+  publishing {
```
ejona86: This doesn't have the same "publish to a local repo" (`-PrepositoryDir`) support that we use for our releases.

bom/build.gradle
```diff
@@ -0,0 +1,105 @@
...
+  }
+}
+
+archivesBaseName = "bom"
+
+project.version = "1.19.0-SNAPSHOT" // CURRENT_GRPC_VERSION
+
+apply plugin: 'maven-publish'
+
+ext {
```
ejona86: No need to use ext when you're in a particular project; just define a variable.

bom/pom.xml
ejona86: Repeating myself that only io.grpc should be included (so that it will be part of the review).

bom/pom.xml
```diff
@@ -43,322 +43,118 @@
...
       <dependency>
         <groupId>io.grpc</groupId>
-        <artifactId>grpc-stub</artifactId>
+        <artifactId>grpc-benchmarks</artifactId>
         <version>${project.version}</version>
       </dependency>
       <dependency>
         <groupId>io.grpc</groupId>
-        <artifactId>grpc-auth</artifactId>
+        <artifactId>grpc-bom</artifactId>
```
garrettjonesgoogle: I don't think `grpc-bom` should be included here (it's self-referential)

bom/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+        mavenLocal()
+        maven {
+            url 'https://plugins.gradle.org/m2/'
+        }
+    }
+}
+
+archivesBaseName = "bom"
+
+project.version = "1.19.0-SNAPSHOT" // CURRENT_GRPC_VERSION
```
ejona86: This is unnecessary. It is specified in the main build.gradle.

RELEASING.md
```diff
@@ -34,6 +34,8 @@ $ VERSION_FILES=(
   build.gradle
   android/build.gradle
   android-interop-testing/app/build.gradle
+  bom/build.gradle
+  bom/pom.xml
```
ejona86: Delete

bom/build.gradle
```diff
@@ -0,0 +1,39 @@
+description = 'gRPC: BOM'
+
+buildscript {
```
ejona86: We don't need this, since we're not pulling down any plugins (other than those provided by gradle)

creamsoup: done

examples/example-tls/pom.xml
```diff
@@ -9,44 +9,55 @@
...
     <maven.compiler.source>1.7</maven.compiler.source>
     <maven.compiler.target>1.7</maven.compiler.target>
   </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>io.grpc</groupId>
+        <artifactId>grpc-bom</artifactId>
+        <version>${project.version}</version>
```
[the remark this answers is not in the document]
creamsoup: yes i agree that grpc.version makes more sense. changed all examples' pom.

examples/example-tls/pom.xml
```diff
@@ -9,44 +9,55 @@
...
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
   <dependencies>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-netty</artifactId>
```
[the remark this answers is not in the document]
creamsoup: yes, converted back to original order.

Issue #4755, opened 2018-08-14
ALTS: release handshaker service channel if it is not used.


jiangtaoli2016:
In current ALTS java implementation, we create a shared ALTS thread and a gRPC channel to the handshaker service when we build ALTS channel. The channel to the handshaker service will already be there until application exit. It is more efficient if we can ref count the channel and free this gRPC channel if nobody is using it.

Note that ALTS C++ code does the same resource management: creates a thread and a channel to the handshaker service and only destroys them during grpc shutdown. https://github.com/grpc/grpc/blob/master/src/core/tsi/alts_transport_security.cc#L53

zpencer:
@jiangtaoli2016 any updates on this one?

srini100:
@jiangtaoli2016, can this issue be closed since https://github.com/grpc/grpc-java/pull/5210 is merged?

jiangtaoli2016:
Yes, Let's close this.

PR #5210, opened 2019-01-04
ALTS: release handshaker channel if no longer needed

jiangtaoli2016:
In ALTS and Google Default Channel, we create a shared channel to handshaker service lazily. This PR releases the channel when ProtocolNegotiator close() is called.

### DIFF
--- a/alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
+++ b/alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
@@ -20,7 +20,6 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.protobuf.Any;
 import io.grpc.Attributes;
-import io.grpc.CallCredentials;
 import io.grpc.Channel;
 import io.grpc.Grpc;
 import io.grpc.InternalChannelz.OtherSecurity;
@@ -29,6 +28,7 @@ import io.grpc.SecurityLevel;
 import io.grpc.Status;
 import io.grpc.alts.internal.RpcProtocolVersionsUtil.RpcVersionsCheckResult;
 import io.grpc.alts.internal.TsiHandshakeHandler.TsiHandshakeCompletionEvent;
+import io.grpc.internal.GrpcAttributes;
 import io.grpc.internal.ObjectPool;
 import io.grpc.netty.GrpcHttp2ConnectionHandler;
 import io.grpc.netty.ProtocolNegotiator;
@@ -192,7 +192,7 @@ public abstract class AltsProtocolNegotiator implements ProtocolNegotiator {
                     .set(ALTS_CONTEXT_KEY, altsContext)
                     .set(Grpc.TRANSPORT_ATTR_REMOTE_ADDR, ctx.channel().remoteAddress())
                     .set(Grpc.TRANSPORT_ATTR_LOCAL_ADDR, ctx.channel().localAddress())
-                    .set(CallCredentials.ATTR_SECURITY_LEVEL, SecurityLevel.PRIVACY_AND_INTEGRITY)
+                    .set(GrpcAttributes.ATTR_SECURITY_LEVEL, SecurityLevel.PRIVACY_AND_INTEGRITY)
                     .build(),
                 new Security(new OtherSecurity("alts", Any.pack(altsContext.context))));
           }
--- a/alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
+++ b/alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
@@ -58,9 +58,7 @@ public abstract class AltsProtocolNegotiator implements ProtocolNegotiator {
 
   /** Creates a negotiator used for ALTS client. */
   public static AltsProtocolNegotiator createClientNegotiator(
-      final TsiHandshakerFactory handshakerFactory,
-      final ObjectPool<Channel> handshakerChannelPool,
-      final LazyChannel lazyHandshakerChannel) {
+      final TsiHandshakerFactory handshakerFactory, final LazyChannel lazyHandshakerChannel) {
     final class ClientAltsProtocolNegotiator extends AltsProtocolNegotiator {
 
       @Override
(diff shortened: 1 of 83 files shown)


jiangtaoli2016:
#4755

alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
```diff
@@ -68,20 +80,28 @@ public Handler newHandler(GrpcHttp2ConnectionHandler grpcHandler) {
...
   }
 
   /** Creates a negotiator used for ALTS server. */
   public static AltsProtocolNegotiator createServerNegotiator(
-      final TsiHandshakerFactory handshakerFactory) {
+      final TsiHandshakerFactory handshakerFactory,
+      final ObjectPool<Channel> handshakerChannelPool) {
     final class ServerAltsProtocolNegotiator extends AltsProtocolNegotiator {
+
+      private Channel handshakerChannel = null;
```
carl-mastrangelo: ```s/ = null;/;```

it's implicit in java.

alts/src/main/java/io/grpc/alts/internal/TsiHandshakerFactory.java
```diff
@@ -16,11 +16,12 @@
...
 
+import io.grpc.Channel;
 import javax.annotation.Nullable;
 
 /** Factory that manufactures instances of {@link TsiHandshaker}. */
 public interface TsiHandshakerFactory {
 
   /** Creates a new handshaker. */
-  TsiHandshaker newHandshaker(@Nullable String authority);
+  TsiHandshaker newHandshaker(@Nullable Channel handshakerChannel, @Nullable String authority);
```
carl-mastrangelo: I fear this may have internal repercussions.

jiangtaoli2016: It is not an issue. I have done it before. Let me prepare an import patch.

alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
jiangtaoli2016: done.

alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
```diff
@@ -68,20 +80,28 @@ public Handler newHandler(GrpcHttp2ConnectionHandler grpcHandler) {
       @Override
       public void close() {
         logger.finest("ALTS Client ProtocolNegotiator Closed");
-        // TODO(jiangtaoli2016): release resources
+        handshakerChannelPool.returnObject(handshakerChannel);
```
ejona86: This could release the object before all the Handlers have finished using it.

alts/src/main/java/io/grpc/alts/internal/TsiHandshakerFactory.java
ejona86: This looks wrong. The reason authority is passed is because it could be different every time. This channel is also specific to one handshaker implementation.

Instead, you should pass some object when creating the TsiHandshakerFactory. Problably an ObjectPool.

alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
jiangtaoli2016: If we don't return the channel here, where shall we return the channel?

ejona86: I was thinking we would return it here and in the handshakers.

However, I checked and this close is called on termination, which means it is guaranteed no handshakes are still occurring. So there's nothing needed to be changed here.

alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
```diff
@@ -47,18 +49,28 @@
...
+      final ObjectPool<Channel> handshakerChannelPool) {
     final class ClientAltsProtocolNegotiator extends AltsProtocolNegotiator {
+
+      private Channel handshakerChannel = null;
+
       @Override
       public Handler newHandler(GrpcHttp2ConnectionHandler grpcHandler) {
-        TsiHandshaker handshaker = handshakerFactory.newHandshaker(grpcHandler.getAuthority());
+        if (handshakerChannel == null) {
+          handshakerChannel = handshakerChannelPool.getObject();
```
ejona86: This is not thread-safe.

alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
```diff
@@ -91,13 +101,34 @@ public Handler newHandler(GrpcHttp2ConnectionHandler grpcHandler) {
       @Override
       public void close() {
         logger.finest("ALTS Server ProtocolNegotiator Closed");
-        // TODO(jiangtaoli2016): release resources
+        handshakerChannelPool.returnObject(lazyHandshakerChannel.get());
```
ejona86: This creates it if it wasn't used. If that's being done there's no point in creating it lazily.

Add a close/shutdown/release to `LazyChannel`? It would be a synchronized method and release the object only if it was created. That also means you wouldn't need to pass `ObjectPool` to `AltsProtocolNegotiator` which assumed it is the same instance as `LazyChannel` is using.

alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
```diff
@@ -91,13 +101,34 @@ public Handler newHandler(GrpcHttp2ConnectionHandler grpcHandler) {
...
       }
     }
 
     return new ServerAltsProtocolNegotiator();
   }
 
+  /** Channel created from a channel pool lazily. */
+  public static class LazyChannel {
+    private ObjectPool<Channel> channelPool;
```
ejona86: `final`

alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
jiangtaoli2016: Good suggestion! done.

alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
jiangtaoli2016: done

jiangtaoli2016:
Thanks Eric and Carl for helpful discussion and suggestions! It is much nicer than the original implementation.

PR #3921, opened 2018-01-05
Upgrade to Protobuf 3.5.1 and Protoc 3.5.1-1.

zhangkun83:
protoc-3.5.0-linux-x86_64 introduced GLIBC_2.14 dependency and broke
gRPC release process (https://github.com/google/protobuf/issues/4138).
3.5.1-1 is the proper re-build.

PR #5212, opened 2019-01-07
bazel: update protobuf dependency

drigz:
This makes grpc-java compatible with recent Bazel releases by pulling
in changes including:

https://github.com/protocolbuffers/protobuf/pull/4650
https://github.com/protocolbuffers/protobuf/commit/c2f88de3318ff13c104da2bed3c197efab215e30

Is it OK to update the Bazel dependency separately, or should I update references, as in https://github.com/grpc/grpc-java/pull/3921?

### DIFF
--- a/repositories.bzl
+++ b/repositories.bzl
@@ -175,18 +175,18 @@ def com_google_protobuf():
     # This statement defines the @com_google_protobuf repo.
     http_archive(
         name = "com_google_protobuf",
-        sha256 = "1f8b9b202e9a4e467ff0b0f25facb1642727cdf5e69092038f15b37c75b99e45",
-        strip_prefix = "protobuf-3.5.1",
-        urls = ["https://github.com/google/protobuf/archive/v3.5.1.zip"],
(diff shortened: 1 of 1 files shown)


ejona86:
@drigz, sorry for the late reply. Go ahead and update all references of protobuf to the 3.6.1 release (like in the other PR you mentioned). It's safe since gRPC dropped support for Java 6 (#3961).

In the past updating protobuf version has been a bit of a pain because of compiler warnings. If it becomes too painful for you just tell us and we will work on bumping the version.

drigz:
@ejona86 I gave it a shot, but it seems quite involved - especially because protobuf dropped support for javanano ([source](https://github.com/protocolbuffers/protobuf/issues/5288#issuecomment-433150290)). Should I file an issue for the update and close this PR?

ejona86:
@drigz, oh, you leave javanano in the past. Sorry for the confusion. We can take it from here. Let's leave this open until we make a replacement PR.

ejona86:
Closing in favor of #5320

Issue #5002, opened 2018-10-25
CallCredentials in 1.16.0


njhill:
This may be a non-issue but I wanted to verify the precise intention of the staged changes described in https://github.com/grpc/grpc-java/issues/4901.

The first steps imply a transition release which doesn't break existing code, presumably this is 1.16.0.
> 1. Introduce `CallCredentials2`, which is the new API, which implements the old `CallCredentials`. Change all stock implementations to it. Deprecate the old `CallCredentials`.
> 2. Make a release, and ask all implementations (including [FirestoreCallCredentials](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/FirestoreCallCredentials.java)) to implement `CallCredentials2` instead.

But in its current form, the 1.16.0 `CallCredentials` breaks binary compatibility for implementations since `CallCredentials.MetadataApplier` has changed from an interface to an abstract class, and this is in a parameter of its only method `applyRequestMetadata()`. I guess maybe the answer is that because that method is an `@ExperimentalApi` then you are SOL. But since implementing libraries need to be rebuilt to work with this anyhow, why bother staging it at all?

And in the subsequent release, _users_ of `CallCredentials` will suffer the same fate despite the interface itself being non-experimental. I am curious why this break needs to happen - couldn't it be kept as an interface with an abstract `BaseCallCredentials` implementation added that implementors can extend to be insulated from further API evolution?

njhill:
It's also not clear why `CallCredentials.MetadataApplier` needs to be changed from an interface at all - if I understand correctly this will only ever have an internal implementation?

zhangkun83:
/cc @ejona86 

We didn't realize changing `MetadataApplier` from interface to abstract class would break binary compatibility.  This is indeed an issue.

The proper way may be keeping `CallCredentials.MetadataApplier` an interface and making `CallCredentials2.MetadataApplier` abstract class instead. GRPC would bridge those two.

We may need a patch release for this.

njhill:
Thanks @zhangkun83! What is the rationale for changing it from an interface at all? Just so that it can have a package-private constructor to make it clear that it's not something that users would ever implement themselves?

And what are your thoughts on keeping `CallCredentials` an interface?

ejona86:
[...]

PR #5216, opened 2019-01-09
core: promote the new API in CallCredentials2

zhangkun83:
This is the 3rd step of #4901

- The deprecated `CC.applyRequestMetadata(... Attributes ...)` is now **replaced** by the new API `CC.applyRequestMetadata(... CC.MetadataApplier ...)` transformed from `CC2.applyRequestMetadata(... CC2.MetadataApplier ...)`.
- The Attributes keys in `CallCredentials` were deprecated, and now deleted.
- The deprecated interface `CC.MetadataApplier` is **replaced** by an equivalent abstract class.
- `CallCredentials2` is now marked as deprecated, while keeping its interface intact so that it won't break current implementations that are still on `CallCredentials2`.
- From this point on, implementations should do a one-line change from `extends CallCredentials2` to `extends CallCredentials`
- `GoogleAuthLibraryCallCredentials` is kept as `CallCredentials2` for now, as there is an internal consumer that expects it to be `CallCredentials2`.

/cc @njhill

### DIFF
--- a/core/src/main/java/io/grpc/CallCredentials.java
+++ b/core/src/main/java/io/grpc/CallCredentials.java
@@ -16,7 +16,6 @@
 
 package io.grpc;
 
-import io.grpc.Attributes.Key;
 import java.util.concurrent.Executor;
 
 /**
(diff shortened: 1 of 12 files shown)


core/src/main/java/io/grpc/CallCredentials.java
```diff
@@ -35,83 +34,51 @@
...
-   * The authority string used to authenticate the server. Usually it's the server's host name. It
-   * is guaranteed to be present in the {@code attrs} passed to {@link #applyRequestMetadata}. It is
-   * by default from the channel, but can be overridden by the transport and {@link
-   * io.grpc.CallOptions} with increasing precedence.
-   */
-  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1914")
-  @Grpc.TransportAttr
-  @Deprecated
-  public static final Key<String> ATTR_AUTHORITY = Key.create("io.grpc.CallCredentials.authority");
+public abstract class CallCredentials {
```
sanjaypujare: The comments starting on line 31 should be changed to reflect this is no longer an interface but an abstract class.

zhangkun83:
@njhill, since you reported #5002, you may want to take a closer look, or test it locally to verify there is no binary compatibility issue.

PR #1844, opened 2016-05-19
netty: don't revalidate when converting between Metadata and Http2Headers

carl-mastrangelo:
This change doesn't cause, but rather unearths a potential performance gain.  Consider the following:

```
Benchmark                         (direct)  (transport)  (validate)    Mode     Cnt       Score     Error  Units
TransportBenchmark.unaryCall1024      true        NETTY        true  sample  557618  178882.761 ▒ 429.097  ns/op
TransportBenchmark.unaryCall1024      true        NETTY       false  sample  613258  162598.208 ▒ 377.362  ns/op
TransportBenchmark.unaryCall1024     false        NETTY        true  sample  366755  272032.442 ▒ 629.053  ns/op
TransportBenchmark.unaryCall1024     false        NETTY       false  sample  397020  251252.659 ▒ 726.254  ns/op

Also 

Benchmark                         (direct)  (transport)  (validate)    Mode     Cnt       Score     Error  Units
TransportBenchmark.unaryCall1024      true        NETTY        true  sample  558486  178687.649 ▒ 353.134  ns/op
TransportBenchmark.unaryCall1024      true        NETTY       false  sample  624353  159732.303 ▒ 384.432  ns/op
TransportBenchmark.unaryCall1024     false        NETTY        true  sample  371358  268634.961 ▒ 711.681  ns/op
TransportBenchmark.unaryCall1024     false        NETTY       false  sample  391621  254763.291 ▒ 668.236  ns/op
```

Running with the plain Netty Transport to make sure headers go all the way through.  Trying to repeat this experiment several times seems to vary between no perf difference and 15-25us.  I still haven't nailed what exactly the 16% speed up is from.  Running with ASM dumping shows identical x86 assembly codes.  

In any case, the validation Netty does is redundant with that of Metadata

PR #5217, opened 2019-01-09
netty: fix lint error

creamsoup:
Remove @VisibleForTesting, because NettyServerTransport is referencing a method Utils#getSocketOptions.


ejona86:
The reason for this change is a bit wrong. The real reason is because the annotation was wrong; it is a util package and is supposed to be used by other classes in the same package. It looked like the annotation was mistakenly added in https://github.com/grpc/grpc-java/pull/1844 (the class seems like it was made public during the development of that PR and then reverted, especially given the discussion at https://github.com/grpc/grpc-java/pull/1844#discussion-diff-64117939).

PR #5219, opened 2019-01-09
Upgrade to Mockito 1.10.19

ejona86:


ejona86:
This is triggering a problem with serverNotListening on Windows. I wouldn't be surprised if this is a legitimate race or similar. I'll have to investigate before merging.

ejona86:
It looks like the flakiness has disappeared. I've done 4 (maybe 5) runs recently and they all succeeded.

PR #5221, opened 2019-01-09
all: upgrade mockito

carl-mastrangelo:


ejona86:
I got there first! #5219

It looks like this is missing two files, so let's just go with my PR?

PR #5223, opened 2019-01-09
core: fix a channel panic bug caused by calling NameResolver.refresh() when it's not started

### LINKED ISSUE #5222: ManagedChannelImpl should not call NameResolver.refresh() if it's not started
This was discovered by internal user. ManagedChannelImpl calling `refresh()` on a not started NameResolver produced an exception and led the channel into panic mode.
 
```
Caused by: java.lang.IllegalStateException: not started
	at com.google.common.base.Preconditions.checkState(Preconditions.java:510)
	at io.grpc.internal.DnsNameResolver.refresh(DnsNameResolver.java:193)
	at io.grpc.internal.ManagedChannelImpl.refreshNameResolutionNow(ManagedChannelImpl.java:429)
	at io.grpc.internal.ManagedChannelImpl.access$3900(ManagedChannelImpl.java:98)
	at io.grpc.internal.ManagedChannelImpl$LbHelperImpl.handleInternalSubchannelState(ManagedChannelImpl.java:993)
	at io.grpc.internal.ManagedChannelImpl$LbHelperImpl.access$4500(ManagedChannelImpl.java:987)
	at io.grpc.internal.ManagedChannelImpl$LbHelperImpl$1ManagedInternalSubchannelCallback.onStateChange(ManagedChannelImpl.java:1031)
	at io.grpc.internal.InternalSubchannel$2.run(InternalSubchannel.java:356)
	at io.grpc.SynchronizationContext.drain(SynchronizationContext.java:101)
	at io.grpc.internal.InternalSubchannel$TransportListener.transportShutdown(InternalSubchannel.java:616)
	at io.grpc.okhttp.OkHttpClientTransport.startGoAway(OkHttpClientTransport.java:792)
	at io.grpc.okhttp.OkHttpClientTransport.onException(OkHttpClientTransport.java:778)
	at io.grpc.okhttp.OkHttpClientTransport$4.run(OkHttpClientTransport.java:538)
	... 4 more
```

This looks like a pre-existing issue, while made more visible (or rather, disruptive) by #5089 where `refresh()` started to be called from the sync context. Prior to that, the exception would be caught by the scheduled executor service, which would only end up in logs rather than panicking the channel. We should probably back port the fix to 1.17.x

zhangkun83:
Resolves #5222

### DIFF
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -308,16 +308,20 @@ final class ManagedChannelImpl extends ManagedChannel implements
   }
 
   // Must be called from syncContext
-  private void shutdownNameResolverAndLoadBalancer(boolean verifyActive) {
-    if (verifyActive) {
-      checkState(nameResolver != null, "nameResolver is null");
+  private void shutdownNameResolverAndLoadBalancer(boolean channelIsActive) {
+    if (channelIsActive) {
+      checkState(nameResolverStarted, "nameResolver is not started");
       checkState(lbHelper != null, "lbHelper is null");
     }
     if (nameResolver != null) {
       cancelNameResolverBackoff();
       nameResolver.shutdown();
-      nameResolver = null;
       nameResolverStarted = false;
+      if (channelIsActive) {
+        nameResolver = getNameResolver(target, nameResolverFactory, nameResolverParams);
+      } else {
+        nameResolver = null;
+      }
     }
     if (lbHelper != null) {
       lbHelper.lb.shutdown();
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -421,11 +419,19 @@ private void cancelNameResolverBackoff() {
     }
   }
 
-  // Must be run from syncContext
-  private void refreshNameResolutionNow() {
+  /**
+   * Force name resolution refresh to happen immediately and reset refresh back-off. Must be run
+   * from syncContext.
+   */
+  private void forceNameResolutionRefresh() {
```
ejona86: "force" and "now" are both not great names, since `refreshNameResolution` always refreshes. Maybe `refreshAndResetNameResolution`? Although that's not perfect, at least people may read the javadoc instead of assuming they know what it means. Dunno.

PR #5227, opened 2019-01-10
examples: Not to promise a supported bazel version

dapengzhang0:
It's not maintainable.


ejona86:
We may want a "tested with Bazel 0.blah", but yeah.

PR #5230, opened 2019-01-10
kokoro: Reduce gradle workers on Mac to reduce flakes

ejona86:
We've had a long run of test timeout failures on Mac with timeouts in
the seconds. Multiple of these haven't seemed like races, but simply
that the Mac machines were too slow. The most recent case for this is a
set of timeouts for
OkHttpTransportTest.earlyServerClose_serverFailure_withClientCancelOnListenerClosed
where it took over a second for a server stream to be created ("Timed
out waiting for server stream").

This commit reduces the number of workers to reduce the overall load,
hoping this lets tests complete within a reasonable amount of time.

### DIFF
--- a/buildscripts/kokoro/macos.cfg
+++ b/buildscripts/kokoro/macos.cfg
@@ -4,6 +4,14 @@
 build_file: "grpc-java/buildscripts/kokoro/unix.sh"
 timeout_mins: 45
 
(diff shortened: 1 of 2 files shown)


dapengzhang0:
Macos failed because the env variable key is not allowed.

ejona86:
@dapengzhang0, PTAL

PR #5232, opened 2019-01-11
core: disable SRV records lookup

ejona86:
SRV has not yet been enabled in a release. Since work is rapidly
underway to replace GRPC-LB with a service config+XDS-based solution,
there's now thoughts that we won't ever enable grpclb by default (but
may allow it to be automatically enabled when using GoogleDefaultChannel
or similar). Since things are being worked out, disable it.

This effectively reverts c729a0f.

CC @carl-mastrangelo

### DIFF
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -90,7 +90,7 @@ final class DnsNameResolver extends NameResolver {
   private static final String JNDI_LOCALHOST_PROPERTY =
       System.getProperty("io.grpc.internal.DnsNameResolverProvider.enable_jndi_localhost", "false");
   private static final String JNDI_SRV_PROPERTY =
-      System.getProperty("io.grpc.internal.DnsNameResolverProvider.enable_grpclb", "true");
+      System.getProperty("io.grpc.internal.DnsNameResolverProvider.enable_grpclb", "false");
   private static final String JNDI_TXT_PROPERTY =
       System.getProperty("io.grpc.internal.DnsNameResolverProvider.enable_service_config", "false");
 


blazej24:
@ejona86 what do you mean by:
> work is rapidly underway to replace GRPC-LB with a service config+XDS-based solution 

I am currently writing a loadbalancer using GRPCLB and wanted to use it - I thought that GRPCLB will be enabled by default soon.

ejona86:
@blazej24, there is a new design coming for an xDS-based API, as is mentioned a bit in https://github.com/grpc/proposal/blob/master/A24-lb-policy-config.md . For gRPC-LB implementations, it shouldn't be too much work to change things to support the new protocol (as there's a version of the protocol that's similar to gRPC-LB), but realize the bootstrapping on the client changes from SRV to TXT. There will be more information coming.

ejona86:
FWIW, we thought it was going to be enabled soon as well, but timelines sort of morphed and the very slow roll-out of SRV in C and Java coupled with some risk-reduction changes we added made us re-evaluate. Again, there will be more information coming.

ejona86:
@blazej24, you'll want to read [Evolution of gRPCLB](https://groups.google.com/forum/#!msg/grpc-io/0yGihF-EFQo/A4QKdXffBwAJ) just posted to the mailing list.

blazej24:
@ejona86 thanks for info.  

> For gRPC-LB implementations, it shouldn't be too much work to change things to support the new protocol (as there's a version of the protocol that's similar to gRPC-LB)

Which version of xDS do you mean as similar to gRPC-LB?

ejona86:
@blazej24, apparently [it is ADS](https://github.com/grpc/grpc-java/blob/master/xds/third_party/envoy/src/main/proto/envoy/service/discovery/v2/ads.proto). But that doesn't seem very clear. There will be a gRFC coming, which should be more clear. All I really understand is there is a mode that will return the addresses directly to the client, like was done in gRPC-LB.

PR #5247, opened 2019-01-16
netty: add a channel active handler

carl-mastrangelo:
This is the next part in building the initial pipeline, and I expect to be the common first handler for any protocol negotiation.  In the expected use, all (real) protocol negotiation logic will expect the channel to already be active when handlers are added to the pipeline.  They will not deal with channelActive or channelInactive.   Instead, those events will be handled by  `WaitUntilActiveHandler`, and `WriteBufferingAndExceptionHandler`, respectively.  

This class is trivial in order to be reusable by all protocol negotiation setups.  In a subsequent PR, the Plaintext protocol negotiator will be modified to use this setup.


ejona86:
Sorry for the delay in review, this was lost in my inbox.

I'm not wild about the code not being used. Could you just swap Plaintext over to using it now? Otherwise it's hard to tell if it is a good solution.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -787,4 +780,152 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
...
+      } else {
+        super.userEventTriggered(ctx, evt);
+      }
+    }
+
+    private void fireProtocolNegotiationEvent(ChannelHandlerContext ctx) {
+      if (protocolNegotiationEvent == null) {
+        protocolNegotiationEvent = ProtocolNegotiationEvent.DEFAULT;
+      }
+      Attributes attrs = protocolNegotiationEvent.getAttributes().toBuilder()
```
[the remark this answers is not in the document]
carl-mastrangelo: It's a little awkward to make active imply insecure, but it's reasonable to expect later handlers to replace it.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -374,17 +377,7 @@ public void close() {}
    * is active.
    */
   public static ProtocolNegotiator plaintext() {
-    return new PlaintextNegotiator();
-  }
-
-  static final class PlaintextNegotiator implements ProtocolNegotiator {
-    @Override
-    public Handler newHandler(GrpcHttp2ConnectionHandler handler) {
-      return new BufferUntilChannelActiveHandler(handler);
```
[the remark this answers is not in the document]
carl-mastrangelo: Done.

netty/src/test/java/io/grpc/netty/ProtocolNegotiatorsTest.java
```diff
@@ -99,6 +112,98 @@ public void setUp() throws Exception {
...
+    };
+
+    LocalAddress addr = new LocalAddress("local");
+    ChannelFuture cf = new Bootstrap()
+        .channel(LocalChannel.class)
+        .handler(lateAddingHandler)
+        .group(group)
+        .register();
+    chan = cf.channel();
+    cf.sync();
```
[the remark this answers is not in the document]
carl-mastrangelo: Done.

netty/src/test/java/io/grpc/netty/ProtocolNegotiatorsTest.java
```diff
@@ -99,6 +112,98 @@ public void setUp() throws Exception {
...
+        });
+
+    LocalAddress addr = new LocalAddress("local");
+    ChannelFuture cf = new Bootstrap()
+        .channel(LocalChannel.class)
+        .handler(handler)
+        .group(group)
+        .register();
+    chan = cf.channel();
+    cf.sync();
```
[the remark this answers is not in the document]
carl-mastrangelo: Done.

netty/src/test/java/io/grpc/netty/ProtocolNegotiatorsTest.java
```diff
@@ -383,53 +515,148 @@ public void httpProxy_500() throws Exception {
...
-    } finally {
-      channel.close();
+    assertThat(req.method()).isEqualTo(HttpMethod.CONNECT);
+    assertThat(req.uri()).isEqualTo("specialHost:314");
+    assertThat(req.headers().get(HttpHeaderNames.HOST)).isEqualTo("specialHost:314");
+
+    Object second = serverReads.poll(5, TimeUnit.SECONDS);
+    assertThat(second).isInstanceOf(LastHttpContent.class);
+
+    if (true) {
```
ejona86: Just use a plain block:
```
{
  Throwable cause ...
  ...
}
```

(Although in this case it seems the extra block doesn't get you much.)

netty/src/test/java/io/grpc/netty/ProtocolNegotiatorsTest.java
```diff
@@ -319,60 +446,65 @@ public void httpProxy_completes() throws Exception {
     LocalAddress proxy = new LocalAddress("httpProxy_completes");
     SocketAddress host = InetSocketAddress.createUnresolved("specialHost", 314);
 
-    ChannelInboundHandler mockHandler = mock(ChannelInboundHandler.class);
+    final BlockingQueue<Object> serverReads = new LinkedBlockingQueue<>();
```
ejona86: Why is the server code changing? This is effectively being rewritten and it isn't clear to my why.

carl-mastrangelo: The server code is pretty unusual.   Using mocks to pull out the ctx and read the strings is surprising.  Also, by making the netty handler capture the HttpObject types, the connect request can be checked more precisely (rather than prefix matching the strings).  

I think the new version is much clearer, and more netty-like.  If you still feel it should be reverted I can.

ejona86: I don't really mind the blocking queue to get reads, although it is still unclear why this was changed. What is the behavior change this test was updated for?

I did prefer not to use the Http decoder stuff because it is less obvious we are using them correctly. Like we could receive the headers, but the request wasn't terminated. And for the encoder we'd need to make sure that there are no poor interactions with the CONNECT response.

carl-mastrangelo: Originally this was failing due to grpchandler, and to check my bases I removed the mocks after getting some spurious channelInactives.  Later I added in the FakeGrpcHttp2Handler.  I think this is the root of the problem.  

I reverted these changes and added back in the mocks.

carl-mastrangelo:
I'll merge this if you don't have any other comments.

Issue #5103, opened 2018-12-03
When does grpc support okhttp 3.0 


YilinYuan:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?
grpc 1.6.1

### What did you expect to see?
grpc-okhttp of  grpc 1.6.1  contains okhttp 2.5.0 ,but I want to use okhttp 3.0 because I need use it for  andriod glide.     I can't compaile different  okhttp's version  for andriod 's app

dapengzhang0:
There is no problem to use okhttp 3.0 and grpc-okhttp-1.6.1 in the same app. That is because grpc-okhttp manually shaded all necessary okhttp2 code that could conflict with okhttp3. Refer to discussions in #2999, #2549, #1628

Issue #2999, opened 2017-05-12
Any plan upgrade to OkHttp 3?


b95505017:
As title.

ejona86:
We didn't really. We've manually "shaded" most of OkHttp that we use such that the okhttp dependency is just for our surface API for specifying TLS settings (via ConnectionSpec).

We could probably have OkHttpChannelBuilder support both ConnectionSpec from OkHttp 2 and 3 simultaneously. It would mean depending on both versions, but when using ProGuard only one would be used.

It's not clear whether that just gives us a "nice feeling" or whether it actually would benefit someone.

ejona86:
It doesn't sound like this provides a tangible benefit, so closing. It'd be easy to convince me otherwise, though. So if this would benefit you, say so and we will consider it. Also mention whether depending on both OkHttp 2 and OkHttp 3 would be a solution for you.

kraghu:
In Android We are so used to okhttp for handling all our rest call needs . It will benefit us from learning point of view.

Issue #2549, opened 2016-12-22
 The version of okhttp is not up-to-date


trywarrior:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?
1.0.2

### What JVM are you using (`java -version`)?
1.8.0_51

### What did you do?
If possible, provide a recipe for reproducing the error.

grpc used the old version of okhttp 2.5.0

### What did you expect to see?

Update the version of okhttp to the latest version

carl-mastrangelo:
I believe we have vendored copies of OkHttp in our code, which have since been modified.  It may be difficult to update OkHttp.

ejona86:
The code is vendored, which means it doesn't really change much the version in the dependency chain. See #2999

Issue #1628, opened 2016-04-02
Support OkHttp3?


njovy:
It's been a while since stable OkHttp3 has been released and libraries depending on OkHttp started to support OkHttp3 as well. Is there any plan to support OkHttp3?

ejona86:
We've not discussed it too much. I believe that most gRPC users are still on OkHttp 2. I'm not against moving to OkHttp3, but we wouldn't want to support both versions.

adrien-aubel:
All for Okhttp3 here :)
Okhttp3 has a much improved API and backend, and is obviously where Square is focused at this point.
So this would probably benefit users of Okhttp backend, and reduce overhead for Android apps that already use Okhttp3. For users that didn't upgrade yet to Okhttp3, they'll still get the benefits, but they will have both versions as dependencies (which still works well, all Okhttp3 classes are in `okhttp3` package name, while Okhttp2 classes are in `com.squareup.okhttp`).

carl-mastrangelo:
I have been using OkHttp3 on my own personal projects and I have to agree I like the API of it as well.  It probably won't be as prioritized as Netty since there isn't as much Android demand, but I would be interested in having this in the future.

eveliotc:
Any update here? We would love to start using gRPC but have been using okhttp3 since July last year, fork wouldn't be ideal.

ejona86:
Closing, for same reason as in #2999. I guess comment there if you want, since I describe a possible option. The only way I see it being an improvement for _users_ is by being able to use OkHttp 3's ConnectionSpec. But I've also not heard anyone say that's involved.

PR #5248, opened 2019-01-16
okhttp: update okhttp and okio dependencies

joelpedraza:
Update OkHttp to 3.x, OkIo to 2.x
===

This has been denied when requested in #5103, #2999, #2549, #1628

The case for upgrading
===

okhttp 2.5.0 is ~5 years old. The major version upgrade previously denied is now just over 3 years old.

Many libraries have migrated dependencies to okhttp3 in recent versions. 
---

General Libs

 * Retrofit
 * Picasso
 * Glide
 * react-native

Vendor SDKs

 * Parse
 * Mapbox
 * Google Maps
 * Branch
 * Paypal

It would be preferable to not need to include multiple versions of the same http lib in our project, especially for those apps targeting Google Instant Apps: which have a hard 4MB apk size limit (an oversimplification, but the http and grpc libs are likely to be in the 'base' module of an app)

Changelog entries relevant to my project
---

(paraphrased)

OkHttp
 * The DNS service is now pluggable
 * Many fixes related to certificate pinning
 * Use newer TLS primitives when available (android api 17+)
 * Gracefully recover when Android 7.0's sockets throw an unexpected NullPointerException
 * Recover gracefully when Android's DNS crashes with an unexpected NullPointerException.
 * Recover gracefully when Android's socket connections crash with an unexpected ClassCastException.

OkIo
 * Perf improvements. 

OkIo 2.x  is binary compatible with 1.x

Vendoring both versions
===

Could work for my project. 2.x would fall out when tree shaking.

Motivations on doing so are unclear to me given the age of 2.x branch. Looking at the project history guava got a major version upgrade from 19 to 20, though admittedly netty has only had revision bumps.

This PR does not try to vendor both 2.x and 3.x versions. It's only a version bump.


thelinuxfoundation:

* User @joelpedraza isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

joelpedraza:
> Thank you for your pull request. Before we can look at your contribution, we need to ensure all contributors are covered by a Contributor License Agreement.
> 
> After the following items are addressed, please respond with a new comment here, and the automated system will re-verify.
> 
>     * User @joelpedraza isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf
> 
> 
> _Regards, CLA GitHub bot_

I signed it

ejona86:
I'll note that if we removed DEFAULT_CONNECTION_SPEC (which could be done immediately in a separate PR) we really wouldn't be using many okhttp methods.

OkHttpChannelBuilder:
* We should remove DEFAULT_CONNECTION_SPEC. We can do that in a separate PR
* We should continue supporting OkHttp 2's ConnectionSpec for a while. We should do the same convertSpec() goo for it, but just another copy of it (overloading the current method). Proguard will strip them when unused. This alone should be all that's necessary to support both OkHttp 2 and 3 simultaneously for migration.

OkHttpClientTransport changes should be fine.

I'm not wild about the piecemeal copying of CipherSuite and TlsVersion (without copying more from OkHttp 3.12). But I see why it was done. @ericgribkoff, thoughts? 

> The DNS service is now pluggable

gRPC doesn't use this.

> Many fixes related to certificate pinning

I'm not sure if this works with gRPC in the normal way.

> Use newer TLS primitives when available (android api 17+)

Doh, I thought we had already modified ConnectionSpec to add chacha20 support, but apparently not. We'd need to update OkHttpChannelBuilder's cipher suite list to use the new suites. We should be able to just look at Netty's list (as the comment mentions).

> Gracefully recover when Android 7.0's sockets throw an unexpected NullPointerException
> Recover gracefully when Android's socket connections crash with an unexpected ClassCastException.

I'd have to see the specific locations these are done to tell whether gRPC would see the changes.

> Recover gracefully when Android's DNS crashes with an unexpected NullPointerException.

gRPC doesn't use okhttp's DNS.

> OkIo
> Perf improvements.

Upgrading okio is fine. I am sad about okio swapping languages though; reading the okio 2 code is hard so I always swap to looking at the old Java code and assuming it works basically the same way.

We will need to make sure we can upgrade okio internally.

space-pope:
Noticed this was still open and that the OkHttp version is still at 2.5. The linked issue that's the most recent item in this thread (closed because this one is still open) is another reason to upgrade, IMO.

ericgribkoff:
> Noticed this was still open and that the OkHttp version is still at 2.5. The linked issue that's the most recent item in this thread (closed because this one is still open) is another reason to upgrade, IMO.

I think the reporter of https://github.com/googleapis/google-cloud-java/issues/4727 may have misdiagnosed their problem. At least, I don't see anything in the referenced OkHttp 3 bug fix (https://github.com/square/okhttp/pull/4365) to indicate that it has anything to do with uncaught RejectExecutionExceptions in gRPC or that switching to OkHttp 3 would do anything to alleviate it. The OkHttp issue is about an executor that has shutdown: the issue reported in gRPC's stack is about executors occasionally rejecting new tasks, which is normal non-shutdown executor behavior. Rather, it looks like the issue reported there is more to do with https://github.com/grpc/grpc-java/issues/636

ejona86:
I added a comment to https://github.com/googleapis/google-cloud-java/issues/4727 talking about the things that may be related. This PR is unrelated to that issue (it was misdiagnosed).

space-pope:
Sorry for the misguided necro, but thanks for the quick reply! In principle, I'd still be for bumping dependencies, but I've read a bit and realize it isn't that simple/doesn't add a ton of size overhead after proguard's done its thing.

ejona86:
Most of this isn't very relevant after https://github.com/grpc/grpc-java/pull/8971 . We could still expose the okhttp 3+ ConnectionSpec API on OkHttpChannelBuilder, and we would accept a contribution for that. But it should be a new method. We want to leave the okhttp 2 ConnectionSpec API on OkHttpChannelBuilder for a while after the new method is added (to allow easy migration). See #8732

PR #5255, opened 2019-01-18
netty: handle racy stream cancellation on already failed transports

carl-mastrangelo:


netty/src/main/java/io/grpc/netty/NettyClientStream.java
```diff
@@ -230,14 +232,29 @@ public TransportState(
 
     @Override
     public int id() {
+      assert id > 0 : "id must be positive " + id;
```
ejona86: It seems this should be a `checkState()`, since this class can't guarantee the invariant. Did you make it an assert for performance?

carl-mastrangelo: Yeah, I didn't know how frequently this method would be called.   I can make it a checkState if you would still like.

carl-mastrangelo:
poke

netty/src/main/java/io/grpc/netty/NettyClientStream.java
ejona86: I think I'm just against the assert. I'm okay with a checkState and I'm okay with just letting consumers see the -1. But I'm not wild about this half-way thing.

carl-mastrangelo: I treat asserts as compiled comments, which is what I intended here.  meh.

PR #5257, opened 2019-01-18
netty: handle reentrant exception in WBAEH

carl-mastrangelo:


carl-mastrangelo:
This occurs currently due to `AbstractBufferingHandler` playing poorly with `WriteBufferingAndExceptionHandler`.   When the ABH encounters a call to `fail()`, it propagates the exception up to WBAEH, which calls `close()`.  This calls back down to ABH inline, resulting in a loop.

Since ABH will eventually be removed, I don't think this is on going, but for the time being it's needed.

carl-mastrangelo:
Travis breaks for jdk11, which I have submitted a patch for [here](https://github.com/sormuras/bach/pull/41).  Submitting now since there will be some delay between when travis picks it up.

PR #5261, opened 2019-01-18
services: log an error to java logger when health-check disabled.

zhangkun83:
This is required by the spec:
https://github.com/grpc/proposal/blob/master/A17-client-side-health-checking.md


services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -421,6 +425,9 @@ void handleResponse(HealthCheckResponse response) {
       void handleStreamClosed(Status status) {
         if (Objects.equal(status.getCode(), Code.UNIMPLEMENTED)) {
           disabled = true;
+          logger.log(
+              Level.SEVERE, "Health-check with {0} is disabled: {1}",
```
ejona86: Make it clear the status came from the server. For example, "Health-check with {0} is disabled. Server returned: {1}"

zhangkun83: Done.

PR #5264, opened 2019-01-22
core: do not lose status when RST_STREAM with NO_ERROR received

ericgribkoff:
The current behavior (RST_STREAM w/ NO_ERROR results in client seeing INTERNAL error, even if gRPC status code was already sent in trailing headers) seems to have been introduced in https://github.com/grpc/grpc-java/pull/4485. The added call to `transportReportStatus` within `deframerClosed` would have no effect when `statusReported` had already been set to true upon receipt of the rst stream.

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractClientStream.java
+++ b/core/src/main/java/io/grpc/internal/AbstractClientStream.java
@@ -240,7 +240,8 @@ public abstract class AbstractClientStream extends AbstractStream
      * #listenerClosed} because there may still be messages buffered to deliver to the application.
      */
     private boolean statusReported;
-    private boolean deframerClosedWithPartialMessage;
+    /** True if the status to report (set via {@link #transportReportStatus}) is OK. */
+    private boolean statusToReportIsOk;
 
     protected TransportState(
         int maxMessageSize,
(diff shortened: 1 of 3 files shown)


dapengzhang0:
When fixing #3264, alternative approaches were considered, but @ejona86 convinced me to take the current approach.  @ejona86 do you remember why did we fix #3264 by moving `transportReportStatus()` from inside `inboundTrailersReceived()` to inside `closeDeframer()`, rather than fixing in `closeListener() `?

ejona86:
@ericgribkoff and I spoke offline. Something similar to this seems okay, but I don't want to change closeListener.

ericgribkoff:
Updated to set `statusToReportIsOk` in `#transportReportStatus`, and removed the changes from `closeListener`.

core/src/main/java/io/grpc/internal/AbstractClientStream.java
```diff
@@ -240,8 +240,8 @@ protected TransportTracer getTransportTracer() {
      * #listenerClosed} because there may still be messages buffered to deliver to the application.
      */
     private boolean statusReported;
-    private Metadata trailers;
-    private Status trailerStatus;
+    /** True if the status to report (set via {@link #transportReportStatus}) is OK. */
+    private boolean statusToReportIsOk;
```
ejona86: Why is this "to" report. `statusReportedIsOk`?

ericgribkoff: Done

PR #4857, opened 2018-09-13
BROKEN on Windows: Avoid using Jetty ALPN

ejona86:
We still want Netty to test that it continues working with Jetty ALPN,
but otherwise we don't want to use it.

Fixes #2266

PR #5272, opened 2019-01-23
Fix text to indicate ALPN availability

sanjaypujare:
ALPN now available starting release 9 of Java.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -229,7 +229,7 @@ dependencies {
 
 If not using the Netty transport (or you are unable to use OpenSSL for some reason) another alternative is to use the JDK for TLS.
 
-No standard Java release has built-in support for ALPN today ([there is a tracking issue](https://bugs.openjdk.java.net/browse/JDK-8051498) so go upvote it!) so we need to use the [Jetty-ALPN](https://github.com/jetty-project/jetty-alpn) (or [Jetty-NPN](https://github.com/jetty-project/jetty-npn) if on Java < 8) bootclasspath extension for OpenJDK. To do this, add an `Xbootclasspath` JVM option referencing the path to the Jetty `alpn-boot` jar.
+Starting with release 9, Java has built-in support for ALPN ([check this issue](https://bugs.openjdk.java.net/browse/JDK-8051498)) so we need to use the [Jetty-ALPN](https://github.com/jetty-project/jetty-alpn) (or [Jetty-NPN](https://github.com/jetty-project/jetty-npn) if on Java < 8) bootclasspath extension for earlier OpenJDK releases. To do this, add an `Xbootclasspath` JVM option referencing the path to the Jetty `alpn-boot` jar.
 
 ```sh
 java -Xbootclasspath/p:/path/to/jetty/alpn/extension.jar ...


sanjaypujare:
@carl-mastrangelo No. Are you saying it is not okay to merge this doc only change without that test?

dapengzhang0:
Then we should document gRPC with TLS won't work with java 9's ALPN support directly or anything that's more accurate.

SECURITY.md
```diff
@@ -229,7 +229,7 @@ dependencies {
 
 If not using the Netty transport (or you are unable to use OpenSSL for some reason) another alternative is to use the JDK for TLS.
 
-No standard Java release has built-in support for ALPN today ([there is a tracking issue](https://bugs.openjdk.java.net/browse/JDK-8051498) so go upvote it!) so we need to use the [Jetty-ALPN](https://github.com/jetty-project/jetty-alpn) (or [Jetty-NPN](https://github.com/jetty-project/jetty-npn) if on Java < 8) bootclasspath extension for OpenJDK. To do this, add an `Xbootclasspath` JVM option referencing the path to the Jetty `alpn-boot` jar.
+Starting with release 9, Java has built-in support for ALPN ([check this issue](https://bugs.openjdk.java.net/browse/JDK-8051498)) so we need to use the [Jetty-ALPN](https://github.com/jetty-project/jetty-alpn) (or [Jetty-NPN](https://github.com/jetty-project/jetty-npn) if on Java < 8) bootclasspath extension for earlier OpenJDK releases. To do this, add an `Xbootclasspath` JVM option referencing the path to the Jetty `alpn-boot` jar.
```
ericgribkoff: With JDK9, we actually don't need the Jetty-ALPN anymore, as it's available from the JDK itself.

Netty (https://github.com/grpc/grpc-java/pull/3555) and OkHttp (https://github.com/grpc/grpc-java/pull/4136) both support JDK9's ALPN, so it looks like this doc does need to be updated to reflect that change.

dapengzhang0: Also for java 8, we could avoid using Jetty-ALPN agent but use conscrypt library instead, but somehow windows is not working well (#4857).

sanjaypujare: @ericgribkoff thanks. Are you confirming that someone has tested with JDK9's ALPN and everything works so I can just go ahead and merge this PR?

ericgribkoff: I am confirming that JDK9's ALPN works with those two transports, but not that this PR should be merged: the PR says that JDK9 has built-in support for ALPN so we need to use Jetty, and the "so we need to use" part is incorrect.

dapengzhang0: @ericgribkoff "So we need ... for earlier OpenJDK releases" is correct.

sanjaypujare: Actually I added "...for earlier OpenJDK releases" towards the end of the sentence. But I admit the sentence can be reworded and improved. Let me try that.

sanjaypujare: BTW, I am also thinking of rearranging the `TLS with JDK` section with these changes:

- remove the `(Jetty ALPN/NPN)` from the section title
- move my new note to the top of the section since the rest of the section only applies when you have to use Jetty jars
- remove the "WARNING: DON'T DO THIS!!" or reword it clarify it applies only to the Jetty case. Does anybody know that the JDK's own ALPN doesn't have the drawbacks mentioned there (slower and not supporting the necessary ciphers for HTTP2)
- reword the "the moral of the story" sentence to clarify it applies to the Jetty jars

sanjaypujare: How about the "Configuring Jetty ALPN in Web Containers" section? It needs to be changed too?

I wonder if we should just point to existing docs like https://www.eclipse.org/jetty/documentation/9.4.x/alpn-chapter.html for various ALPN related caveats instead of trying to keep our content up to date?

ejona86: I think it needs a lot more work to make it more clear. I've rewritten most of these top sections in #5281 and removed a lot of the documentation for Jetty ALPN. I think we might should drop this PR and work on #5281 instead.

I removed much of this Jetty documentation because it isn't the best way of using Jetty ALPN. If we feel we need it, then we should instead describe how to use the Jetty ALPN agent, which came around after we had already decided against Jetty ALPN.

sanjaypujare: I am okay with dropping this PR. Will close this after I finish reviewing #5281 .

PR #5274, opened 2019-01-23
services: set the default serving state of all services to SERVING

carl-mastrangelo:


carl-mastrangelo:
Filed https://github.com/grpc/grpc/issues/17805

markdroth:
This seems incorrect to me.  We should not set any service other than the empty string to SERVING by default.  None of the other languages do that.

markdroth:
Ah, okay.  Carry on, then. :)

carl-mastrangelo:
Added a comment.

services/src/main/java/io/grpc/services/HealthStatusManager.java
```diff
@@ -26,6 +26,11 @@
  * created in the constructor of {@code HealthStatusManager}, and it can be retrieved by the
  * {@link #getHealthService()} method.
  * The health status manager can update the health statuses of the server.
+ *
+ * <p>
```
ejona86: What's with this formatting? Don't intent because of the `<p>`. Don't include the `</p>`. End in a period.

```
* <p>The default, empty-string, service name, {@link #SERVICE_NAME_ALL_SERVICES}, is initialized to
* {@link ServingStatus#SERVING}.
```

carl-mastrangelo: It's what IntelliJ likes.  Seems arbitrary.

services/src/main/java/io/grpc/services/HealthStatusManager.java
```diff
@@ -26,6 +26,9 @@
  * created in the constructor of {@code HealthStatusManager}, and it can be retrieved by the
  * {@link #getHealthService()} method.
  * The health status manager can update the health statuses of the server.
+ *
+ * <p>The default, empty-string, service name, {@link #SERVICE_NAME_ALL_SERVICES}, is initialized to
+ *    {@link ServingStatus#SERVING}.
```
ejona86: Remove the extra indentation here as well.

It's part of the style guide. https://google.github.io/styleguide/javaguide.html#s7.1.2-javadoc-paragraphs

PR #5281, opened 2019-01-25
SECURITY.md: Rewrite/reflow our recommendations

ejona86:
Much of these sections pre-dated gRPC 1.0 and Java 9. Modernize them to be more
on-point and clear.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,9 +4,13 @@ gRPC supports a number of different mechanisms for asserting identity between an
 
 # Transport Security (TLS)
 
-HTTP/2 over TLS mandates the use of [ALPN](https://tools.ietf.org/html/draft-ietf-tls-applayerprotoneg-05) to negotiate the use of the h2 protocol. ALPN is a fairly new standard and (where possible) gRPC also supports protocol negotiation via [NPN](https://tools.ietf.org/html/draft-agl-tls-nextprotoneg-04) for systems that do not yet support ALPN.
+HTTP/2 over TLS mandates the use of [ALPN](https://tools.ietf.org/html/rfc7301)
+to negotiate the use of the h2 protocol and support for the GCM mode of AES.
 
-On Android, use the [Play Services Provider](#tls-on-android). For non-Android systems, use [OpenSSL](#tls-with-openssl).
+There are multiple options available, but on Android we recommend using the
+[Play Services Provider](#tls-on-android) and for non-Android systems we
+recommend [netty-tcnative with
+BoringSSL](#tls-with-netty-tcnative-on-boringssl).
 
 ## TLS on Android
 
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -215,39 +209,46 @@ dependencies {
...
+gRPC historically supported Jetty ALPN for ALPN on Java 8. While functional, it
+suffers from poor performance and breakages when the JRE is upgraded.
+When mis-matched to the JRE version, it can also produce unpredictable errors
+that are hard to diagnose. When using it, it became common practice that any
+time we saw a TLS failure that made no sense we would blame a Jetty ALPN/JRE
+version mismatch and we were overwhelmingly correct. The Jetty ALPN agent makes
+it much easier to use, but we still strongly discourage Jetty ALPN's use.
 
-Users should be aware that GCM is [_very_ slow (1 MB/s)](https://bugzilla.redhat.com/show_bug.cgi?id=1135504) before Java 8u60. With Java 8u60 GCM is 10x faster (10-20 MB/s), but that is still slow compared to OpenSSL (~200 MB/s), especially with AES-NI support (~1 GB/s). GCM cipher suites are the only suites available that comply with HTTP2's cipher requirements.
+When using Jetty ALPN with Java 8, realize that performance will be 2-10% that
```
sanjaypujare: Java 8 => Java 8 or less

(unless we don't work with pre-8)

ejona86: Jetty ALPN only supported Java 8. You had to use Jetty NPN for Java 7. I'm not mentioning NPN at all because we "officially" require ALPN.

SECURITY.md
```diff
@@ -42,108 +46,99 @@ import java.security.Security;
...
+supported platforms. For Maven:
 
 ```xml
-<project>
   <dependencies>
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-tcnative-boringssl-static</artifactId>
-      <version>2.0.20.Final</version>
+      <version>2.0.20.Final</version> <!-- See table for correct version -->
```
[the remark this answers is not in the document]
ejona86: Yeah, I know that was a problem. The table should probably be moved out of the debugging section, but I wasn't ready to make that change.

SECURITY.md
ejona86: (Also, Java 7 performance was "zero" because it didn't support AES GCM.)

SECURITY.md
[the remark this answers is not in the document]
sanjaypujare: Can we use an anchor point and hyperlink to the anchor point? So your text here will be
```
<!-- See [table](#version_table) for correct version -->
```

See my comment below for the anchor point on the version table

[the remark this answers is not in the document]
sanjaypujare: Okay github won't let me put a comment on unmodified content. But on line 363 you can insert
`<a name="version_table"></a>` on a line by itself to create the anchor.

SECURITY.md
```diff
@@ -42,108 +46,99 @@ import java.security.Security;
...
 ```gradle
-buildscript {
-  repositories {
-    mavenCentral()
-  }
-}
-
 dependencies {
-    compile 'io.netty:netty-tcnative-boringssl-static:2.0.20.Final'
+  // See table for correct version
```
sanjaypujare: Same as above - if you can make `table` a hyperlink...

SECURITY.md
```diff
@@ -42,108 +46,99 @@ import java.security.Security;
...
+  compile 'io.netty:netty-tcnative-boringssl-static:2.0.20.Final'
 }
 ```
 
-### OpenSSL: Dynamically Linked (netty-tcnative)
-
-If for any reason you need to dynamically link against OpenSSL (e.g. you need control over the version of OpenSSL), you can instead use the `netty-tcnative` artifact.
+For projects sensitive to binary size, specify the classifier for the precise
```
sanjaypujare: How to specify the classifier? Is it just appended to the end? Can we give an example `io.netty:netty-tcnative-boringssl-static-osx-x86_64:2.0.20.Final` ?

sanjaypujare: Interestingly our own `examples/example-tls/build.gradle` uses `runtimeOnly` instead of `compile` as here . Which one is correct and do we need to fix the other one?

SECURITY.md
```diff
@@ -42,108 +46,99 @@ import java.security.Security;
...
+platform you need: `windows-x86_64`, `osx-x86_64`, `linux-x86_64`. You can also
+use [os-maven-plugin](https://github.com/trustin/os-maven-plugin) or
+[osdetector-gradle-plugin](https://github.com/google/osdetector-gradle-plugin),
+to choose the classifier for the platform running the build.
```
sanjaypujare: Our own `examples/example-tls/build.gradle` does not seem to include the osdetector plugin explicitly. So it looks like there isn't anything for the user to do to use the osdetector plugin?

SECURITY.md
```diff
@@ -42,108 +46,99 @@ import java.security.Security;
...
+the dependency.
 
 ```xml
 <project>
   <dependencies>
     <dependency>
       <groupId>io.netty</groupId>
       <artifactId>netty-tcnative</artifactId>
-      <version>2.0.20.Final</version>
+      <version>2.0.20.Final</version> <!-- see table for correct version -->
```
sanjaypujare: hyperlink "table" to "version_table"

SECURITY.md
```diff
@@ -42,108 +46,99 @@ import java.security.Security;
...
 
-This is currently the recommended approach for using gRPC over TLS (on non-Android systems).
+JDK versions prior to Java 9 do not support ALPN and are either missing AES GCM
+support or have 2% the performance of OpenSSL.
 
-The main benefits of using OpenSSL are:
+We recommend most users use grpc-netty-shaded, which includes netty-tcnative on
+BoringSSL. It includes pre-built libraries for 64 bit Windows, OS X, and 64 bit
+Linux. For 32 bit Windows, Conscrypt may be an option. For all other platforms,
+Java 9+ is required.
```
sanjaypujare: Except for this single reference, there isn't any documentation about the option of built-in ALPN on JDK-9 or higher. It is not immediately clear that it's a valid option

sanjaypujare:
closes #5272

SECURITY.md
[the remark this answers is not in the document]
ejona86: The table should not be in the middle of the troubleshooting section. I was expecting that it'd get moved to its own section.

But once it gets its own section, I can't put a link within a code sample, as markdown is disabled there. So it'd have to be in words like "See Compatibility Table for correct version".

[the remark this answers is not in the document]
sanjaypujare: Correct - I missed the part about code sample/quoted text.

PR #5282, opened 2019-01-25
all: fix formating on bzl file

carl-mastrangelo:
The internal linter complains about the formatting of this file.


carl-mastrangelo:
I can change it to `none` if you'd like.

ejona86:
I'm not too concerned about this particular PR, but in the future no prefix or `java_grpc_library.bzl` (or `java_grpc_library`) would probably be better. A "none" prefix is worse than useless.

PR #5287, opened 2019-01-25
all: fix lint errors

carl-mastrangelo:


netty/src/test/java/io/grpc/netty/NettyChannelBuilderTest.java
```diff
@@ -40,8 +40,6 @@
 
   @Rule public final ExpectedException thrown = ExpectedException.none();
   private final SslContext noSslContext = null;
-  private final ProxyParameters noProxy = null;
```
dapengzhang0: Unused import ProxyParameters?

carl-mastrangelo: Done.

PR #5289, opened 2019-01-26
all: migrate gradle publish from maven to use maven-publish plugin

creamsoup:
You can see sample snapshot repo at http://jihuncho0.svl.corp.google.com:8081/#browse/browse:jihun_staging


netty/build.gradle
```diff
@@ -33,9 +33,3 @@ test {
     // Allow testing Jetty ALPN in TlsTest
     jvmArgs "-javaagent:" + configurations.alpnagent.asPath
 }
-
-jmh {
```
creamsoup: this caused an error during :grpc-netty:animalsnifferJmh task.
error is about "Failed to check signatures"

creamsoup:
some differences are 
1. grpc-gae-interop-testing-jdk8-1.0-SNAPSHOT.pom packaging "war" is removed.  (probably need to fix?)
2. runtime scope is changed to compile.
3. test dependencies are removed.

ejona86:
grpc-gae-interop-testing-jdk8 should not be uploaded.

compile scope instead of runtime is unfortunate, but probably okay. I think I remember seeing some Gradle discussion about this. It might be fixed if we swapped to implementation and api dependencies.

build.gradle
```diff
@@ -348,79 +342,101 @@ subprojects {
...
+                artifact sourcesJar
 
-            licenses {
-                license {
-                    name 'Apache 2.0'
-                    url 'https://opensource.org/licenses/Apache-2.0'
-                }
-            }
+                pom {
+                    name = "$project.group:$project.name".toString()
```
ejona86: Why is toString() necessary?

build.gradle
```diff
@@ -348,79 +342,101 @@ subprojects {
...
+                        }
+                    }
+
+                    developers {
+                        developer {
+                            id = "grpc.io"
+                            name = "gRPC Contributors"
+                            email = "grpc-io@googlegroups.com"
+                            url = "https://grpc.io/"
+                            // https://issues.gradle.org/browse/GRADLE-2719
```
ejona86: This may now be out-of-date.

netty/shaded/build.gradle
```diff
@@ -46,14 +46,6 @@ shadowJar {
     mergeServiceFiles()
 }
 
-// This is a hack to have shadow plugin modify the uploadArchives POM's
```
ejona86: grpc-netty-shaded's pom doesn't look right. It shouldn't have dependencies on netty and tcnative.

settings.gradle
```diff
@@ -1,44 +1,45 @@
 rootProject.name = "grpc"
-include ":grpc-core"
```
ejona86: Why were these lines changed?

netty/build.gradle
ejona86: Check whether "./gradlew :grpc-netty:jmh"  runs. You don't need to do a full benchmark, but if a benchmark starts, it is probably fine.

build.gradle
```diff
@@ -348,79 +342,101 @@ subprojects {
         from sourceSets.main.allSource
     }
 
-    artifacts { archives javadocJar, sourcesJar }
```
ejona86: Why is the artifacts block no longer used here, but we still use it other places (like compiler)?

compiler/build.gradle
```diff
@@ -1,5 +1,6 @@
 apply plugin: "cpp"
 apply plugin: "com.google.protobuf"
+apply plugin: "java"
```
ejona86: This looks fishy.

netty/shaded/build.gradle
```diff
@@ -46,13 +41,15 @@ shadowJar {
...
-// but that's broken anyway with install/uploadArchives
-uploadShadow.repositories.addAll(uploadArchives.repositories)
-// And then we use a further hack to share that POM with install
-install.repositories.mavenInstaller.pom = uploadArchives.repositories.mavenDeployer.pom
+publishing {
+    publications {
+        maven(MavenPublication) {
+            // Ideally swap to project.shadow.component(it) when it isn't broken for project deps
+            artifact shadowJar
+            // TODO: manually add deps
```
ejona86: TODO: here and elsewhere

build.gradle
```diff
@@ -348,79 +342,105 @@ subprojects {
...
+                            username = rootProject.ossrhUsername
+                            password = rootProject.ossrhPassword
+                        }
+                    }
+                    def releaseUrl = stagingUrl
+                    def snapshotUrl = 'https://oss.sonatype.org/content/repositories/snapshots/'
+                    url = version.endsWith('SNAPSHOT') ? snapshotUrl : releaseUrl
+                }
             }
+            mavenLocal()
```
ejona86: Remove (unnecessary)

build.gradle
creamsoup: those are unused and removed.

build.gradle
creamsoup: GString(Impl)? seems to have issue (there are few open issues and some are closed but still doesn't quite work)
manual string concat works without toString.

build.gradle
creamsoup: yes. the issue is marked as closed. removing the note.

build.gradle
```diff
@@ -348,79 +342,101 @@ subprojects {
...
-            "grpc-protobuf",
-            "grpc-protobuf-lite",
-            "grpc-protobuf-nano"
-        ])) {
-            def core = pom.dependencies.find {dep -> dep.artifactId == 'grpc-core'}
-            if (core != null) {
-                // Depend on specific version of grpc-core because internal package is unstable
-                core.version = "[" + core.version + "]"
+        repositories {
+            maven {
```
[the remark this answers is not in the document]
creamsoup: so publishToMavenLocal is not using the repositoryDir to specify
./gradlew publishToMavenLocal -Dmaven.repo.local=$MY_M2_HOME
or 
./gradlew publish -PrepositoryDir=$MY_M2_HOME

compiler/build.gradle
creamsoup: oops

netty/build.gradle
creamsoup: it was broken (it maybe due to some hack in the previous version of PR). now it works as expected.

netty/shaded/build.gradle
creamsoup: discussed offline. NOTE for others: because shadow plugin is filtering project deps, we had to manually generate dependencies.

settings.gradle
creamsoup: reverted to reduce review complexity. i'll issue other PR for sorting the order.

creamsoup:
TODO: i think some checks will fail (especially android), need to fix the some build scripts (may still need to install since android ones are not converted)

PR #5295, opened 2019-01-30
interop-testing:  Add GoogleDefaultCreds test case for java

apolcyn:
This PR adds the [google default credentials](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#google_default_credentials) test case for Java. As follow up, will also need to add thist into `run_interop_tests.py` on GCE linux and also on mac.

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
@@ -16,6 +16,9 @@
 
 package io.grpc.testing.integration;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.io.Files;
 import io.grpc.ManagedChannel;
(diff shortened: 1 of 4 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
```diff
@@ -16,6 +16,9 @@
 
 package io.grpc.testing.integration;
 
+import static org.junit.Assert.assertEquals;
```
ejona86: This isn't a test file, so don't use asserts. Instead you could use things like IllegalArgumentException or IllegalStateException.

interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
```diff
@@ -304,6 +307,13 @@ private void runTest(TestCases testCase) throws Exception {
         break;
       }
 
+      case GOOGLE_DEFAULT_CREDENTIALS: {
+        assertNotNull(customCredentialsType);
```
ejona86: Hmm... It doesn't seem right to require the user to specify customCredentialsType. Although I do see now that the test causes trouble when creating the channel.

Let's move the test from AbstractInteropTest to Tester in this file, and just have it manually create the Channel for itself. You'll need to shutdown the channel though, and that should probably be in a try-finally:

```
ManagedChannel channel = GoogleDefaultChannelBuilder.forAddress(serverHost, serverPort).build();
try {
  // do test
} finally {
  channel.shutdownNow();
}
```

apolcyn: done

interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
apolcyn: removed per below comment

apolcyn:
thanks. Let me know if I should squash. I don't have write access so I can't merge :)

PR #5300, opened 2019-01-31
all: fix lint

dapengzhang0:


context/src/test/java/io/grpc/ThreadLocalContextStorageTest.java
```diff
@@ -25,7 +25,7 @@
 @RunWith(JUnit4.class)
 public final class ThreadLocalContextStorageTest {
   private static final Context.Key<Object> KEY = Context.key("test-key");
-  private ThreadLocalContextStorage storage = new ThreadLocalContextStorage();
+  private final ThreadLocalContextStorage storage = new ThreadLocalContextStorage();
```
dapengzhang0: @ejona86 Did you purposely set it non-final for GC reason?

Issue #5301, opened 2019-01-31
Remove  compileOnly libraries.truth


elharo:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?

HEAD

```
    // Use compileOnly to avoid dependencyConvergence problem with the Guava
    // pulled in via Truth, for users that don't use Truth. Truth requires a
    // more up-to-date Guava than we support elsewhere, which would trigger
    // convergence failures in tests that only our users could resolve. Using
    // compileOnly means only users using Truth would have the problem and
    // they'd have to resolve it like normal anyway.
    compileOnly libraries.truth
```

This is out of date. Truth 0.42 and HEAD are now both at 25.1-android.

ejona86:
We don't want to add a compile dependency on truth because basically none of our users need it and it can cause the Guava version to be different during tests vs production.

Instead, the plan is for us to move the internal bits of grpc-testing into core, and depend on core's tests directly when needed.

suztomo:
I created a demo project to produce an issue of grpc-testing.
https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/447/files

suztomo:
Thank you for the enhancement.

PR #5302, opened 2019-01-31
remove obsolete truth guava version workaround

elharo:


ejona86:
We happen to use it only minorly in the classes in testing/ today. But the usages are appropriate and we do use Truth elsewhere. ErrorProne is getting increasingly pushy to use Truth as well.

The reason the dependency matters here is that users will receive the dependency for this library, but none of the others. And we wouldn't want our users to accidentally use a different version of Guava when testing vs running in production.

I'm going to talk with others on the team to figure out our preferred approach.

dapengzhang0:
Tracking issue #5301

elharo:
Where does this stand? Should I close this pull request?

ejona86:
Closing, since we don't want to expose a dependency on truth. Fixing the compileOnly will be tracked by #5301

PR #5307, opened 2019-02-01
alts: fix missing dependencies

creamsoup:


build.gradle
```diff
@@ -346,8 +346,8 @@ subprojects {
         publications {
             // do not use mavenJava, as java plugin will modify it via "magic"
             maven(MavenPublication) {
-                if (!project.name.contains("grpc-netty-shaded") && !name.contains("grpc-alts")
-                      && !name.contains("grpc-compiler")) {
+                def nonJavaPluginProjects = ['grpc-netty-shaded', 'grpc-alts', 'grpc-compiler']
```
ejona86: It seems we should remove grpc-alts from this list instead of modifying alts/build.gradle. Then it will match the previous configuration (before maven-publish) by only renaming grpc-netty to grpc-netty-shaded.

creamsoup: done.

PR #5312, opened 2019-02-02
netty: add channelFactory to NettyChannelBuilder

creamsoup:


carl-mastrangelo:
I find the API a little surprising.  I would expect the last call to win (see `Bootstrap`) and how it uses `ReflectiveChannelFactory` to accomplish this.

PR #5316, opened 2019-02-04
Fix usages of any(<Primitive>.class) matchers

TimvdLippe:
In Mockito 2, if a method expects a primitive type, but an any(<Primitive>.class) matcher is used in its place, it will throw an error. To prepare for this upcoming breakage, change
all existing any(<Primitive>.class) matchers to use the correct any<Primitive>() matcher.


carl-mastrangelo:
@TimvdLippe You'll need to sign the CNCF CLA (all Googlers do).  Once done, I can merge.  Thanks for the fast response!

TimvdLippe:
CLA should be signed now.

PR #5320, opened 2019-02-05
Upgrade to protobuf 3.6.1

### LINKED ISSUE #5310: Bazel building is broken with protobuf<3.6.1.3
1.18.0

Bazel build is working; currently it is broken since old protobuf versions uses the removed `REPOSITORY_NAME` and `PACKAGE_NAME` (see https://github.com/protocolbuffers/protobuf/commit/af4810663bf1fb2388e88ad9038b80551aefd6da).

### Solution
Bump protobuf to 3.6.1.3?

ejona86:
For Bazel, we upgrade to protobuf 3.6.1.2 and javalite HEAD to fix
incompatibilities in newer Bazel releases.

compiler/Dockerfile is unused, so it was removed instead of being updated.

protoc no longer includes codegen for nano, so we remain on the older protoc
any time nano is used.

Protobuf now requires C++11 when compiling, so windows was swapped to
VC 14.

-----

CC @drigz 

Fixes #5310


ejona86:
Windows is failing because Protobuf now requires C++11. We'll have to update to a newer version of VS.

ejona86:
@zhangkun83, PTAL, the changes are each in their own separate commit.

PR #5322, opened 2019-02-05
Sample of sychronous blocking stub

carl-mastrangelo:


carl-mastrangelo:
The core idea here is to be synchronous, but generally non-blocking.  There are blocking and non-blocking variants of each method.

stub/src/main/java/io/grpc/stub/SyncCall.java
```diff
@@ -0,0 +1,794 @@
+/*
+ * Copyright 201, gRPC Authors All rights reserved.
```
sanjaypujare: 201 => 2019

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -324,7 +326,14 @@ protected boolean metricsExpected() {
 
   @Test
   public void emptyUnary() throws Exception {
```
sanjaypujare: Why not keep the old test and add a new test? A direct call like `blockingStub.emptyCall(EMPTY)` and another one via `SyncCall` are testing 2 different code paths, right?

sanjaypujare: Also the test does not need to be in interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java, right? Instead could it be moved to stub/src/test/java/io/grpc/stub/ClientCallsTest.java or  (even better) a new file stub/src/test/java/io/grpc/stub/SyncCallsTest.java

carl-mastrangelo: The point was to try out the new API, not so much to be a full switch.  Mainly a sample of how it can be used.

PR #5324, opened 2019-02-06
netty: fix lint errors

creamsoup:


creamsoup:
do we "fix" those?

ejona86:
We generally ignore it. We have configured ErrorProne to ignore FutureReturnValueIgnored failures, both internally and externally. But unfortunately the code review tool ignores our settings. b/37215436 is tracking the issue.

creamsoup:
Okay. Closing it, thanks for the bug link.

PR #5328, opened 2019-02-06
core: Update gRPC to use span kind and raw full method.

bogdandrutu:


carl-mastrangelo:
Is anyone depending on the format of these ?

bogdandrutu:
@carl-mastrangelo on our exporters to Stackdriver we add the prefix in opencensus and google internal.

ejona86:
This is being reverted in #5355

PR #5353, opened 2019-02-13
netty: limit access to ProtocolNegotiators

carl-mastrangelo:
It doesn't appear that there are any internal users of ProtocolNegotiators.  Limit access to it using an Internal accessor.


carl-mastrangelo:
Thanks for the super fast review @ejona86

PR #5356, opened 2019-02-13
xds: fallback handling

dapengzhang0:
I made XdsLbState not abstract for now.
I did not include graceful swapping balancers when service config change, for now just shutdown the old one and use the new one.


xds/src/main/java/io/grpc/xds/FallbackManager.java
```diff
@@ -0,0 +1,138 @@
...
+import io.grpc.LoadBalancerRegistry;
+import io.grpc.SynchronizationContext.ScheduledHandle;
+import io.grpc.internal.ServiceConfigUtil;
+import io.grpc.xds.XdsLbState.SubchannelStore;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+
+final class FallbackManager {
```
zhangkun83: Is this going to be only used in `XdsLoadBalancer` as a final field? I don't see a benefit of making it a separate class. Usually the valid reasons for making a class is:
1. Reusability.
1. Complex implementation underneath a simple interface. A separate class would make it easy to test it separately, and its user will mock this interface in their own tests.

It's debatable whether Case 2 is the case here, but I feel it's a quite large interface (5 methods) compared to its implementation. Even if you defined an interface for it and mocked it in `XdsLoadBalancerTest`, just having the two classes unit-tested separately wouldn't provide enough confidence of them working as a whole, which calls for the need of an integration test (of the two classes). Maybe just merge it into `XdsLoadBalancer`.

xds/src/main/java/io/grpc/xds/XdsLbState.java
```diff
@@ -85,6 +105,35 @@ final XdsComms shutdownAndReleaseXdsComms() {
     return xdsComms;
   }
 
+  /**
+   * Manages EAG and locality info for a collection of subchannels, not including subchannels
+   * created by the fallback balancer.
+   */
+  static final class SubchannelStore {
```
zhangkun83: Is this essentially the round-robin list, because I see "drop" is indicated here?
If so, it should probably be named more specific, say, "RoundRobinList".

xds/src/main/java/io/grpc/xds/XdsLoadBalancer.java
```diff
@@ -17,49 +17,67 @@
...
+  private final SubchannelStore subchannelStore = new SubchannelStore();
+  private final Helper helper;
+  private final FallbackManager fallbackManager;
 
   @Nullable
   private XdsLbState xdsLbState;
 
+  private Map<String, Object> fallbackPolicy;
+
   XdsLoadBalancer(Helper helper) {
```
zhangkun83: It's better to inject `LoadBalancerRegistry` through the constructor, so that tests can construct `LoadBalancerRegistry` with fake child policies and will not collide with the ones from default registry. The XDS balancer factory will pass the default registry.

ditto for the registry in `FallbackManager`.

xds/src/main/java/io/grpc/xds/FallbackManager.java
```diff
@@ -0,0 +1,138 @@
...
+    if (fallbackTimer == null) {
+      fallbackTimer = helper.getSynchronizationContext().schedule(
+          fallbackTask, FALLBACK_TIMEOUT_MS, TimeUnit.MILLISECONDS,
+          helper.getScheduledExecutorService());
+    }
+  }
+
+  private final class FallbackTask implements Runnable {
+
+    FallbackTask() {}
```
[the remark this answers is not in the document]
dapengzhang0: I was concerning about synthetic constructor if its private. Yes, inside `maybeStartFallbackTimer()` is better, and allocation seems only once. Not using anonymous class for better reading exception stacktrace.

xds/src/main/java/io/grpc/xds/FallbackManager.java
dapengzhang0: The `fallbackManager` instance in my mind will be passed to `AdsStream` response handler. Once `AdsStream` is active, it will use the `fallbackManager` to cancel fallback. And once `AdsStream` is closed and no ready subchannels, it may use the `fallbackManager` to `maybeUseFallbackPolicy()`.

xds/src/main/java/io/grpc/xds/XdsLbState.java
dapengzhang0: Uncertain yet. It may have hierarchies, because it may be grouped by locality. Will see how later implementation goes. Also, it may be weighted RR over localities then intra-locality.

xds/src/main/java/io/grpc/xds/FallbackManager.java
zhangkun83: `AdsStream` only calls the 2 methods, compared to 5 methods on the interface. Alternatively, this could be done by `XdsLoadBalancer` passing a callback to `AdsStream`, which has `onWorking()` and `onClosed()`, and do the fallback handling inside `XdsLoadBalancer`.

zhangkun83:
Another comment on `AdsStream`. It seems `onNext()` etc are public just because it implements `StreamObserver`, but `AdsStream` is not actually used by **its user** as a `StreamObserver`. It may be better to create a `StreamObserver` inside the class so that those `StreamObserver` methods don't show up on the `AdsStream` interface.

xds/src/main/java/io/grpc/xds/FallbackManager.java
dapengzhang0: Done.

dapengzhang0:
Some test cases will be added once AdsStream handling logic is implemented.

xds/src/test/java/io/grpc/xds/XdsLoadBalancerTest.java
```diff
@@ -91,54 +96,52 @@ public LoadBalancer newLoadBalancer(Helper helper) {
...
-  }
-
-  @After
-  public void tearDown() {
-    LoadBalancerRegistry.getDefaultRegistry().deregister(lbProvider1);
-    LoadBalancerRegistry.getDefaultRegistry().deregister(lbProvider2);
+    lbRegistry.register(lbProvider1);
+    lbRegistry.register(lbProvider2);
+    lb = new XdsLoadBalancer(helper, lbRegistry);
+    doReturn(new SynchronizationContext(Thread.currentThread().getUncaughtExceptionHandler()))
```
zhangkun83: You probably want to create the exception handler like you did in FallbackManagerTest.

xds/src/test/java/io/grpc/xds/FallbackManagerTest.java
```diff
@@ -0,0 +1,152 @@
...
+import org.junit.runners.JUnit4;
+import org.mockito.Matchers;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+/**
+ * Unit test for {@link FallbackManager}.
+ */
+@RunWith(JUnit4.class)
+public class FallbackManagerTest {
```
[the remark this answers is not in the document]
dapengzhang0: Made a big refactoring (merged `XdsComms` with `AdsStream`), which plumbs `adsStreamCallback` into `XdsComms` (formerly `AdsStream`), and added tests for the non-trivial interaction between the balancer and the fallback manager.

xds/src/main/java/io/grpc/xds/XdsLbState.java
```diff
@@ -85,26 +113,65 @@ final XdsComms shutdownAndReleaseXdsComms() {
...
+    }
 
-    XdsComms(ManagedChannel channel, AdsStream adsStream) {
-      this.channel = channel;
-      this.adsStream = adsStream;
+    // ** This is wrong **
+    // TODO: use name resolve to resolve addresses for balancerName, and create xdsComms in
+    // name resolver listener callback
+    ManagedChannel oobChannel = helper.createOobChannel(
+        new EquivalentAddressGroup(ImmutableList.<SocketAddress>of(new SocketAddress() {})),
```
zhangkun83: Maybe expose this fake EAG as a static final field visible to tests, and verify `createOobChannel()` with this EAG in tests. This way later when you pass real EAGs, those tests will fail and you can update them accordingly, so you are sure that you don't miss one.

xds/src/test/java/io/grpc/xds/XdsLoadBalancerTest.java
```diff
@@ -149,121 +307,306 @@ public void canHandleEmptyAddressListFromNameResolution() {
...
+
+    ArgumentCaptor<Attributes> captor = ArgumentCaptor.forClass(Attributes.class);
+    verify(fakeBalancer1).handleResolvedAddressGroups(
+        Matchers.<List<EquivalentAddressGroup>>any(), captor.capture());
+    assertThat(captor.getValue().get(ATTR_LOAD_BALANCING_CONFIG))
+        .containsExactly("supported_1", new HashMap<String, Object>());
+  }
+
+  @Test
+  public void fallback_AdsWorkingTimerExpired() throws Exception {
```
zhangkun83: The timer is cancelled, not expired.

xds/src/test/java/io/grpc/xds/XdsLoadBalancerTest.java
```diff
@@ -149,121 +307,306 @@ public void canHandleEmptyAddressListFromNameResolution() {
...
+
+    serverResponseWriter.onError(new Exception("fake error"));
+
+    ArgumentCaptor<Attributes> captor = ArgumentCaptor.forClass(Attributes.class);
+    verify(fakeBalancer1).handleResolvedAddressGroups(
+        Matchers.<List<EquivalentAddressGroup>>any(), captor.capture());
+    assertThat(captor.getValue().get(ATTR_LOAD_BALANCING_CONFIG))
+        .containsExactly("supported_1", new HashMap<String, Object>());
+
+    fakeClock.forwardTime(10, TimeUnit.SECONDS);
```
zhangkun83: Maybe 
```java
assertThat(fakeClock.forwardTime(10, TimeUnit.SECONDS)).isEqualTo(1);
assertThat(fakeClock.getPendingTasks()).isEmpty();
```
to verify that the timer did expire, but had no effect.

xds/src/main/java/io/grpc/xds/XdsComms.java
```diff
@@ -0,0 +1,130 @@
...
+        public void onError(Throwable t) {
+          helper.getSynchronizationContext().execute(
+              new Runnable() {
+                @Override
+                public void run() {
+                  if (cancelled) {
+                    return;
+                  }
+                  closed = true;
+                  adsStreamCallback.onClosed();
```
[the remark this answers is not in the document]
dapengzhang0: Done.

xds/src/main/java/io/grpc/xds/XdsComms.java
```diff
@@ -0,0 +1,130 @@
...
+        .streamAggregatedResources(xdsResponseReader);
+  }
+
+  void shutdownChannel() {
+    channel.shutdown();
+    shutdownLbRpc("Loadbalancer client shutdown");
+  }
+
+  @CheckReturnValue
+  XdsComms maybeRestart() {
```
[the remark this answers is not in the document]
dapengzhang0: Renamed to `refreshAdsStream()`. Added inner class `AdsStream`.

xds/src/main/java/io/grpc/xds/XdsComms.java
```diff
@@ -0,0 +1,130 @@
...
+
+  void shutdownLbRpc(String message) {
+    if (cancelled) {
+      return;
+    }
+    cancelled = true;
+    xdsRequestWriter.onError(Status.CANCELLED.withDescription(message).asRuntimeException());
+  }
+
+  interface AdsStreamCallback {
```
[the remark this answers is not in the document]
dapengzhang0: Done.

xds/src/main/java/io/grpc/xds/XdsLbState.java
```diff
@@ -85,26 +113,65 @@ final XdsComms shutdownAndReleaseXdsComms() {
     return xdsComms;
   }
 
-  static final class XdsComms {
-    private final ManagedChannel channel;
-    private final AdsStream adsStream;
+  private void maybeStartXdsComms() {
```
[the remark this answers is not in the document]
dapengzhang0: Inlined the method.

xds/src/main/java/io/grpc/xds/XdsLbState.java
dapengzhang0: Added TODO because the current PR focus on fallback, and this method eventually will have the correct implementation and be tested.

xds/src/test/java/io/grpc/xds/XdsLoadBalancerTest.java
dapengzhang0: Renamed.

xds/src/test/java/io/grpc/xds/XdsLoadBalancerTest.java
dapengzhang0: Done.

dapengzhang0:
@zhangkun83 Thanks a lot for the review!

PR #5357, opened 2019-02-13
netty: prevent IndexOutOfBoundsException when no handler is passed to BufferingHandler

creamsoup:


creamsoup:
I was originally thought i need to pass empty list. I don't need that change anymore, but it is still relevant to fix the condition.

PR #5361, opened 2019-02-13
xds: use shadow plugin for generated code

dapengzhang0:
- Import envoy proto file to the latest internal version, which has correct java proto options. (The PGV proto, `validate.proto`, doesn't have the correct and up-to-date java_package proto option yet, but as long as we don't use those generated classes, it seems fine.)
- Stop modifying java proto options by import.sh.
- Apply shadow plugin when publishing.


dapengzhang0:
jar tvf xds/build/libs/grpc-xds-1.20.0-SNAPSHOT.jar > jar-tvf.txt
[jar-tvf.txt](https://github.com/grpc/grpc-java/files/2862641/jar-tvf.txt)

xds/build/publications/maven/pom-default.xml
[pom-default.xml.txt](https://github.com/grpc/grpc-java/files/2862653/pom-default.xml.txt)

xds/third_party/protoc-gen-validate/import.sh
```diff
@@ -25,36 +25,26 @@ FILES=(
 validate/validate.proto
 )
 
-# clone the protoc-gen-validate github repo in /tmp directory
-pushd /tmp
-rm -rf $GIT_BASE_DIR
+# clone the protoc-gen-validate github repo in a tmp directory
+tmpdir="$(mktemp -d)"
+pushd "${tmpdir}" || exit 1
```
zhangkun83: Add `set -e` at the beginning of the script, then the whole script would exit if any line fails. Then you wouldn't need `|| exit 1`

xds/build.gradle
```diff
@@ -42,3 +49,25 @@ if (rootProject.childProjects.containsKey('grpc-compiler')) {
...
+    }
+    relocate 'io.envoyproxy', 'io.grpc.xds.shaded.io.envoyproxy'
+    relocate 'validate', 'io.grpc.xds.shaded.com.lyft.pgv.validate'
+    exclude "**/*.proto"
+}
+
+publishing {
+    publications {
+        maven(MavenPublication) {
+            artifact shadowJar
```
ejona86: We'd normally also remove the original jar, like done for alts. Although I understand we aren't publishing for this project yet.

xds/build.gradle
```diff
@@ -53,6 +53,7 @@ configureProtoCompilation()
 jar {
     classifier = 'original'
 }
+configurations.archives.artifacts.removeAll { it.classifier == "original" }
```
ejona86: This doesn't look right. With maven-publish artifacts should no longer really do anything/we shouldn't be using. I would have expected this in the `publishing` section, like seen in alts.

dapengzhang0: Oops, I was exactly copying alts/build.gradle, but the version of my branch was earlier than 1c3432c3fb055e40beeb8aee38e114834ca10da5

PR #5364, opened 2019-02-13
core: make NameResolver not thread-safe

### LINKED ISSUE #2649: NameResolver can be NotThreadSafe

zhangkun83:
Resolves #2649

As a prerequisite, added `getSynchronizationContext()` to `NameResolver.Helper`.

`DnsNameResolver` has gone through a small refactor around the `Resolve` runnable, which makes it a little simpler.

### DIFF
--- a/core/src/main/java/io/grpc/NameResolver.java
+++ b/core/src/main/java/io/grpc/NameResolver.java
@@ -37,10 +37,14 @@ import javax.annotation.concurrent.ThreadSafe;
  * {@link Listener} is responsible for eventually (after an appropriate backoff period) invoking
  * {@link #refresh()}.
  *
+ * <p>Implementations <string>don't need to be thread-safe</strong>.  All methods are guaranteed to
+ * be called sequentially.  Additionally, all methods that have side-effects, i.e., {@link #start},
+ * {@link #shutdown} and {@link #refresh} are called from the same {@link SynchronizationContext} as
+ * returned by {@link Helper#getSynchronizationContext}.
+ *
  * @since 1.0.0
  */
 @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1770")
-@ThreadSafe
 public abstract class NameResolver {
   /**
    * Returns the authority used to authenticate connections to servers.  It <strong>must</strong> be
@@ -214,13 +218,27 @@ public abstract class NameResolver {
     /**
      * The port number used in case the target or the underlying naming system doesn't provide a
      * port number.
+     *
+     * @since 1.19.0
      */
     public abstract int getDefaultPort();
 
     /**
      * If the NameResolver wants to support proxy, it should inquire this {@link ProxyDetector}.
      * See documentation on {@link ProxyDetector} about how proxies work in gRPC.
+     *
+     * @since 1.19.0
      */
     public abstract ProxyDetector getProxyDetector();
+
+    /**
+     * Returns the {@link SynchronizationContext} where {@link #start}, {@link #shutdown} and {@link
+     * #refresh} are run from.
+     *
+     * @since 1.20.0
+     */
+    public SynchronizationContext getSynchronizationContext() {
+      throw new UnsupportedOperationException("Not implemented");
+    }
   }
 }
--- a/core/src/main/java/io/grpc/NameResolver.java
+++ b/core/src/main/java/io/grpc/NameResolver.java
@@ -37,7 +37,7 @@ import javax.annotation.concurrent.ThreadSafe;
  * {@link Listener} is responsible for eventually (after an appropriate backoff period) invoking
  * {@link #refresh()}.
  *
- * <p>Implementations <string>don't need to be thread-safe</strong>.  All methods are guaranteed to
+ * <p>Implementations <strong>don't need to be thread-safe</strong>.  All methods are guaranteed to
  * be called sequentially.  Additionally, all methods that have side-effects, i.e., {@link #start},
  * {@link #shutdown} and {@link #refresh} are called from the same {@link SynchronizationContext} as
  * returned by {@link Helper#getSynchronizationContext}.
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -138,17 +138,17 @@
...
+
+  // Following fields must be accessed from synContext
+  private ResolutionResults cachedResolutionResults = null;
   private boolean shutdown;
-  @GuardedBy("this")
   private Executor executor;
-  @GuardedBy("this")
   private boolean resolving;
-  @GuardedBy("this")
   private Listener listener;
```
ejona86: It looks like this can be accessed outside of the syncContext; it's only changed in start().

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -138,17 +138,17 @@
   private final String host;
   private final int port;
   private final Resource<Executor> executorResource;
-  @GuardedBy("this")
+  private final long cacheTtlNanos;
+  private final Stopwatch stopwatch;
```
ejona86: This should only be accessed from syncContext.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -138,17 +138,17 @@
...
+  private final SynchronizationContext syncContext;
+
+  // Following fields must be accessed from synContext
+  private ResolutionResults cachedResolutionResults = null;
```
carl-mastrangelo: remove the null.   Its the default and surprising to see it.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -167,7 +167,10 @@
       port = nameUri.getPort();
     }
     this.proxyDetector = Preconditions.checkNotNull(helper.getProxyDetector(), "proxyDetector");
-    this.resolveRunnable = new Resolve(this, stopwatch, getNetworkAddressCacheTtlNanos(isAndroid));
+    this.cacheTtlNanos = getNetworkAddressCacheTtlNanos(isAndroid);
+    this.stopwatch = Preconditions.checkNotNull(stopwatch, "stopwatch");
+    this.syncContext =
```
carl-mastrangelo: This is a little uncomfortable to me.  It would be better if the ctor was dumber, and the DNRP constructed the dependencies for this class, rather than DNR constructing its own.  Same with proxyDetector.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -176,72 +179,53 @@ public final String getServiceAuthority() {
...
+  public final void start(Listener listener) {
     Preconditions.checkState(this.listener == null, "already started");
     executor = SharedResourceHolder.get(executorResource);
     this.listener = Preconditions.checkNotNull(listener, "listener");
     resolve();
   }
 
   @Override
-  public final synchronized void refresh() {
+  public final void refresh() {
```
carl-mastrangelo: This does not need to be final, since the class is final.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -256,37 +240,42 @@ void resolveInternal(Listener savedListener) {
...
+            host);
+        resolutionResults = results;
+        syncContext.execute(new Runnable() {
+            @Override
+            public void run() {
+              cachedResolutionResults = results;
+            }
+          });
         if (cacheTtlNanos > 0) {
           stopwatch.reset().start();
```
carl-mastrangelo: Can stopwatch be accessed by multiple threads?

core/src/main/java/io/grpc/internal/DnsNameResolver.java
ejona86: This new form was much clearer for me, because they are the same for every resolver. Having them in the resolver is harder to reason about because I then have to check if they change. I'd actually prefer if Listener was removed from Resolver as well.

core/src/main/java/io/grpc/NameResolver.java
```diff
@@ -214,13 +218,27 @@ void onAddresses(
     /**
      * The port number used in case the target or the underlying naming system doesn't provide a
      * port number.
+     *
+     * @since 1.19.0
```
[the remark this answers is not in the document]
zhangkun83: Sure: #5367

core/src/main/java/io/grpc/internal/DnsNameResolver.java
ejona86: I'm saying that listener can actually be accessed from any thread. It doesn't change.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -138,16 +138,21 @@
   private final String host;
   private final int port;
   private final Resource<Executor> executorResource;
-  @GuardedBy("this")
+  private final long cacheTtlNanos;
+  private final SynchronizationContext syncContext;
+
+  // Must only be called from syncContext
+  private final Stopwatch stopwatch;
```
ejona86: It seems better to include this in the "must be accessed from the syncContext" section.

zhangkun83: Done.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
zhangkun83: Done.

PR #5368, opened 2019-02-17
Add link to grpc-spring-boot-starter

ST-DDT:
There is the [yidongnan/grpc-spring-boot-starter](https://github.com/yidongnan/grpc-spring-boot-starter) library that connects gRPC with spring. IMO it would be useful, if we could link to that project to allow new users of this framework to discover its integration with the spring framework.

Or don't you want to link to third party libraries?

**See also:**
* https://github.com/grpc/grpc.github.io/pull/832
* https://github.com/spring-projects/spring-framework/issues/20905
* https://github.com/spring-cloud/spring-cloud-sleuth/pull/1139#issuecomment-444302714

**Alternatives considered:**
* Create an official gRPC project for spring integration

**Disclosure:** I'm one of the core maintainers of [yidongnan/grpc-spring-boot-starter](https://github.com/yidongnan/grpc-spring-boot-starter).

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,3 +218,9 @@ gRPC comes with three Transport implementations:
    and is for client only.
 3. The in-process transport is for when a server is in the same process as the
    client. It is useful for testing, while also being safe for production use.
+
+Third-Party Integration
+-----------------------
+
+* [grpc-spring-boot-starter](https://github.com/yidongnan/grpc-spring-boot-starter)
+  (unofficial) fully integrates gRPC with [Spring](https://spring.io/)


README.md
```diff
@@ -218,3 +218,9 @@ gRPC comes with three Transport implementations:
    and is for client only.
 3. The in-process transport is for when a server is in the same process as the
    client. It is useful for testing, while also being safe for production use.
+
+Third-Party Integration
+-----------------------
+
+* [grpc-spring-boot-starter](https://github.com/yidongnan/grpc-spring-boot-starter)
```
dapengzhang0: Just out of curiosity, I saw at least two grpc-spring-boot-starters on github: this one and https://github.com/LogNet/grpc-spring-boot-starter, what is the difference and their relation? Which one is more widely used?

ST-DDT: **Differences**

The two projects have two different philosophies.

* [LogNet/grpc-spring-boot-starter](https://github.com/LogNet/grpc-spring-boot-starter) is server only and focuses on being light weight
  (Size: ~15KB)
* [yidongnan/grpc-spring-boot-starter](https://github.com/yidongnan/grpc-spring-boot-starter) is for both server and client side and aims at full integration with spring, thus it also supports additional features such as Spring-Security and Actuator
  (Size: ~160KB; roughly ~90KB per side)

**Relation**

The only relation they have is their topic "combining gRPC with spring", thus I would call them competitors for lack of better names. I'm not aware of any other relationships between them.

**Usage**

I don't have any numbers. 

The following is very opinionated:
* LogNet is currently preferred by the Google Search. LogNet started earlier and does not have the eastern name.
* yidongnan has an increasing number of visitors and contributors lately.

dapengzhang0: Thanks a lot for your explanation. As discussed in https://github.com/grpc/grpc.github.io/pull/832, I agree that [Awesome gRPC](https://github.com/grpc-ecosystem/awesome-grpc) is a better place for the link (already there) . You may also want to move/mirror your github repo to https://github.com/grpc-ecosystem, we can consider that too.

Issue #5369, opened 2019-02-17
Could not find TLS ALPN provider; no working netty-tcnative, Conscrypt, or Jetty NPN/ALPN available


ankitshubham97:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?
1.13.1

### What did you expect to see?
The jar should be running fine.

I am using java 8 to build an executable jar. Below is the java version:
```
$ /usr/lib/jvm/java-1.8.0-openjdk-amd64/jre/bin/java -version
openjdk version "1.8.0_171"
OpenJDK Runtime Environment (build 1.8.0_171-8u171-b11-2~14.04-b11)
OpenJDK 64-Bit Server VM (build 25.171-b11, mixed mode)
```
I am using gradle 3.4.1 to generate the jar as follows:
`/opt/gradle-3.4.1/bin/gradle jar -Dorg.gradle.java.home=/usr/lib/jvm/java-1.8.0-openjdk-amd64/ 
`
Below are the dependencies defined in the module-level build.gradle file:
```
dependencies {
  compile files("$TOOLCHAIN_VERSION_DIR/lib/commons-io-2.6.jar")
  compile files("$TOOLCHAIN_VERSION_DIR/lib/grpc-all.jar")
  compile files("$TOOLCHAIN_VERSION_DIR/lib/java-protobuf.jar")
  compile files("$TOOLCHAIN_VERSION_DIR/lib/jetty.jar")
  compile files("$TOOLCHAIN_VERSION_DIR/lib/log4j-core-2.8.jar")
  compile files("$TOOLCHAIN_VERSION_DIR/lib/log4j-slf4j-impl-2.8.jar")
  compile files("$TOOLCHAIN_VERSION_DIR/lib/netty-tcnative-boringssl-static-2.0.20.Final.jar")
  compile files("$TOOLCHAIN_VERSION_DIR/lib/netty-tcnative-2.0.20.Final.jar")
  compile files("$TOOLCHAIN_VERSION_DIR/lib/picocli-3.8.2.jar")
  compile group: 'com.google.guava', name: 'guava', version: '20.0'
  compile files("$TOOLCHAIN_VERSION_DIR/lib/javassist-3.19.0-GA.jar")
  compile project(':annotation')
}
```
After building, I am running the jar on an AIX7.2 machine as follows:
`/usr/java8_64/jre/bin/java -jar agent-1.0.jar`

The java version on the AIX machine is as follows:
```
[...]

PR #5374, opened 2019-02-20
io,grpc,netty: Add support for IBMJSSE2 provider

ankitshubham97:
This commit addresses the issue: [https://github.com/grpc/grpc-java/issues/5369](issue#5369)
This commit improves gRPC to run gRPC server with TLS enabled on
AIX machines with IBMJSSE2 as their JDK SSL provider.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
+++ b/netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
@@ -95,6 +95,7 @@ public class GrpcSslContexts {
       NEXT_PROTOCOL_VERSIONS);
 
   private static final String SUN_PROVIDER_NAME = "SunJSSE";
+  private static final String IBM_PROVIDER_NAME = "IBMJSSE2";
   private static final Method IS_CONSCRYPT_PROVIDER;
 
   static {
@@ -225,6 +226,18 @@ public class GrpcSslContexts {
         throw new IllegalArgumentException(
             SUN_PROVIDER_NAME + " selected, but Jetty NPN/ALPN unavailable");
       }
+    } else if (IBM_PROVIDER_NAME.equals(jdkProvider.getName())) {
+      // Jetty ALPN/NPN only supports one of NPN or ALPN
+      if (JettyTlsUtil.isJettyAlpnConfigured()) {
+        apc = ALPN;
+      } else if (JettyTlsUtil.isJettyNpnConfigured()) {
+        apc = NPN;
+      } else if (JettyTlsUtil.isJava9AlpnAvailable()) {
+        apc = ALPN;
+      } else {
+        throw new IllegalArgumentException(
+          IBM_PROVIDER_NAME + " selected, but Jetty NPN/ALPN unavailable");
+      }
     } else if (isConscrypt(jdkProvider)) {
       apc = ALPN;
     } else {
@@ -268,6 +281,12 @@ public class GrpcSslContexts {
             || JettyTlsUtil.isJava9AlpnAvailable()) {
           return provider;
         }
+      } else if (IBM_PROVIDER_NAME.equals(provider.getName())) {
+        if (JettyTlsUtil.isJettyAlpnConfigured()
+            || JettyTlsUtil.isJettyNpnConfigured()
+            || JettyTlsUtil.isJava9AlpnAvailable()) {
+          return provider;
+        }
       } else if (isConscrypt(provider)) {
         return provider;
       }
@@ -305,3 +324,4 @@ public class GrpcSslContexts {
         alpnNegotiator.protocols());
   }
 }
+


thelinuxfoundation:

_Regards,
CLA GitHub bot_

thelinuxfoundation:

* User @ankitshubham97 isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

ankitshubham97:
Filled in the CLA form.

sanjaypujare:
The link in the PR comment doesn't work - the text is correct but the actual link is `https://github.com/grpc/grpc-java/pull/issue#5369` which obviously is wrong.

netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
```diff
@@ -225,6 +226,18 @@ public static SslContextBuilder configure(SslContextBuilder builder, Provider jd
         throw new IllegalArgumentException(
             SUN_PROVIDER_NAME + " selected, but Jetty NPN/ALPN unavailable");
       }
+    } else if (IBM_PROVIDER_NAME.equals(jdkProvider.getName())) {
+      // Jetty ALPN/NPN only supports one of NPN or ALPN
```
sanjaypujare: So the Security Provider here is `IBMJSSE2` and after falling thru to line 248 you are still passing `Http2SecurityUtil.CIPHERS` which according to its documentation at https://netty.io/4.1/api/io/netty/handler/codec/http2/Http2SecurityUtil.html says "The following list is derived from SunJSSE Supported Ciphers and Mozilla Modern Cipher Suites in accordance with the HTTP/2 Specification."

The actual list of ciphers for IBMJSSE2 is listed at https://www.ibm.com/support/knowledgecenter/en/SSYKE2_7.0.0/com.ibm.java.security.component.70.doc/security-component/jsse2Docs/ciphersuites.html . 

May be they match or there is no problem with this code but it will be good to confirm that and may be add a comment to that effect?

netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
```diff
@@ -305,3 +324,4 @@ static void ensureAlpnAndH2Enabled(
         alpnNegotiator.protocols());
   }
 }
+
```
sanjaypujare: Is the newline needed? I prefer to see only changes that are needed in the PR.

netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
```diff
@@ -268,6 +281,12 @@ private static Provider findJdkProvider() {
             || JettyTlsUtil.isJava9AlpnAvailable()) {
           return provider;
         }
+      } else if (IBM_PROVIDER_NAME.equals(provider.getName())) {
+        if (JettyTlsUtil.isJettyAlpnConfigured()
+            || JettyTlsUtil.isJettyNpnConfigured()
+            || JettyTlsUtil.isJava9AlpnAvailable()) {
```
sanjaypujare: lines 285-287 are identical to lines 279-281. The checks seem to be independent of `provider` (the for loop iterator) and loop invariant. Can we create a helper method and also initialize the value outside the loop?

sanjaypujare:
I agree with @carl-mastrangelo : it will be very desirable to have a couple of tests. One for SunJSSE (with the Sun JDK) and another one for IBMJSSE2 (with the IBM JDK?). If there are problems adding tests I would like to know

ankitshubham97:
@sanjaypujare I would be happy to add the tests. May I know what exactly should we be covering in the tests? Also, as @carl-mastrangelo earlier said that you do not have IBM test bed set up, how am I supposed to verify if the tests pass?

carl-mastrangelo:
@ankitshubham97 Were you able to run our tests on this JVM?  (I assume you have access to it).   You should be able to run `./gradlew build` and see the tests pass, making sure your JAVA_HOME is set properly.

As for what we want to see, the coverage numbers from codecov would be good.

Lastly, since we don't have access to this JVM ourselves, it would have to be clear up front that we may break support (accidentally, or from technical compromise) for it.   We do want gRPC to run everywhere, but short of regular testing it would be best effort.  Is that okay with you?

sanjaypujare:
@ankitshubham97 I was thinking you could add tests that are enabled only for the IBM JVM and you are able to verify those tests. So you can verify that `findProvider` returns the right provider and `configure` returns the right `SslContextBuilder`. It will be really nice to be able to mock the `SslContextBuilder` so unit testing becomes easier.

netty/src/main/java/io/grpc/netty/GrpcSslContexts.java
```diff
@@ -225,6 +226,18 @@ public static SslContextBuilder configure(SslContextBuilder builder, Provider jd
         throw new IllegalArgumentException(
             SUN_PROVIDER_NAME + " selected, but Jetty NPN/ALPN unavailable");
       }
+    } else if (IBM_PROVIDER_NAME.equals(jdkProvider.getName())) {
+      // Jetty ALPN/NPN only supports one of NPN or ALPN
+      if (JettyTlsUtil.isJettyAlpnConfigured()) {
```
ejona86: Does it actually support Jetty ALPN? I don't see anything about Jetty ALPN working for IBM and based on the [IBM documentation](https://www.ibm.com/support/knowledgecenter/en/SSEQTP_liberty/com.ibm.websphere.wlp.doc/ae/cwlp_alpnsupport.html), it seems like Jetty ALPN only works if using Oracle's JDK on IBM hardware. It looks like it has [its own API for Java 8](https://www.ibm.com/support/knowledgecenter/en/SSYKE2_8.0.0/com.ibm.java.security.api.doc/jsse2/com/ibm/jsse2/ext/ALPNJSSEExt.html).

I would _really_ like to know which of these is being selected when it works for you. I hope it is the `isJava9AlpnAvailable()` case.

My eventual goal in this part of the code was to use the Java 9+ API, when available, and check whether this specific provider supports ALPN by creating an sslEngine and [calling `getApplicationProtocol()`](https://docs.oracle.com/javase/9/docs/api/javax/net/ssl/SSLEngine.html#getApplicationProtocol--) (or similar); if it throws UnsupportedOperationException, then it doesn't support ALPN. That way we don't have to hard-code provider names.

SunJSSE is hard-coded today because it _must_ be hard-coded for Jetty ALPN/NPN. Java 9+ support was a community contribution and just followed most of the current flow. Unfortunately, just being on Java 9+ doesn't mean the provider supports ALPN, thus the need for eventually adding a `getApplicationProtocol()` check. But this had been "good enough" up until now since we know Java 9+'s SunJSSE implementation supports ALPN.

ejona86:
It also looks like we can add [`ibmjdk8`](https://github.com/michaelklishin/jdk_switcher#jdk-aliases) to the list of JDKs in `.travis.yml`. That way TlsTest could actually run against the IBM JRE.

sanjaypujare:
What's the status of this PR? Any interest?

cmuchinsky:
I just stumbled upon this and have a need for it as I'm using the IBM JDK, it seems like the solution might simply be to use `SunJSSE` and `IBMJSSE2` interchangeably. I will try that out and if successful will open another PR.

ejona86:
Closing in favor of #7422

PR #5378, opened 2019-02-20
okhttp: add socketFactory method to channel builder

ericgribkoff:
This is mainly aimed at allowing Android users to force the OkHttp channel to use a particular network, via providing a socket factory obtained from [`android.net.Network#getSocketFactory()`](https://developer.android.com/reference/android/net/Network.html#getSocketFactory()). The corresponding feature request from an internal user is at b/119564668.

### DIFF
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
@@ -160,6 +160,8 @@ public class OkHttpChannelBuilder extends
 
   /**
    * Override the default {@link SocketFactory} used to create sockets.
+   *
+   * @since 1.20.0
    */
   public final OkHttpChannelBuilder socketFactory(@Nullable SocketFactory socketFactory) {
     this.socketFactory = socketFactory;
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
@@ -159,7 +159,8 @@ public class OkHttpChannelBuilder extends
   }
 
   /**
-   * Override the default {@link SocketFactory} used to create sockets.
+   * Override the default {@link SocketFactory} used to create sockets. If the socket factory is not
+   * set, the value of {@link SocketFactory#getDefault} will be used.
    *
    * @since 1.20.0
    */
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
@@ -49,6 +49,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
+import javax.net.SocketFactory;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSocketFactory;
@@ -120,6 +121,7 @@ public class OkHttpChannelBuilder extends
   private Executor transportExecutor;
   private ScheduledExecutorService scheduledExecutorService;
 
+  private SocketFactory socketFactory;
   private SSLSocketFactory sslSocketFactory;
   private HostnameVerifier hostnameVerifier;
   private ConnectionSpec connectionSpec = INTERNAL_DEFAULT_CONNECTION_SPEC;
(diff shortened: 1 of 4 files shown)


okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
```diff
@@ -234,6 +245,7 @@ protected void handleNotInUse() {
     // Client initiated streams are odd, server initiated ones are even. Server should not need to
     // use it. We start clients at 3 to avoid conflicting with HTTP negotiation.
     nextStreamId = 3;
+    this.socketFactory = socketFactory;
```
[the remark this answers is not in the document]
ericgribkoff: The [javadoc](https://docs.oracle.com/javase/7/docs/api/javax/net/SocketFactory.html#getDefault()) says:

> Factory classes are specified by environment-specific configuration mechanisms. For example, the getDefault method could return a factory that was appropriate for a particular user or applet, and a framework could use a factory customized to its own purposes.

While I believe that in general the return value of `SocketFactory.getDefault()` is [hard-coded](https://stackoverflow.com/questions/803511/setting-the-default-socketfactory/804607) and cannot change at runtime, this doesn't seem to be guaranteed by the API - so I went with the safer (?) approach of invoking it whenever used, which should mimic the existing behavior of `new Socket(...)`

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -156,6 +158,14 @@ public final OkHttpChannelBuilder transportExecutor(@Nullable Executor transport
     return this;
   }
 
+  /**
+   * Override the default {@link SocketFactory} used to create sockets.
+   */
```
dapengzhang0: Add `@since 1.20.0`

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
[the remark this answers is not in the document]
dapengzhang0: That field can still be final. You might also need to test `SocketFactory.getDefault()` is used.

I think every single okhttp transport creates for only once a socket almost immediately after the transport is instantiated. It does not make much sense that the transport should be responsible to reflect any runtime change during the uncertain small gap between transport instantiation and transport start. Also if the runtime change is not synchronized with the `serializingExecutor`, socketFactoryToUse().createSocket() is not thread safe with the runtime change.

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -156,6 +158,17 @@ public final OkHttpChannelBuilder transportExecutor(@Nullable Executor transport
     return this;
   }
 
+  /**
+   * Override the default {@link SocketFactory} used to create sockets. If the socket factory is not
+   * set, the value of {@link SocketFactory#getDefault} will be used.
```
dapengzhang0: Let's hide the implementation detail. What about "If the socket factory is not set or set to null, a default one will be used"?

ericgribkoff: Done

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
ericgribkoff: Done

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientTransport.java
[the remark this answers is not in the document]
ericgribkoff: It is now `final`

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -156,6 +158,14 @@ public final OkHttpChannelBuilder transportExecutor(@Nullable Executor transport
     return this;
   }
 
+  /**
+   * Override the default {@link SocketFactory} used to create sockets.
```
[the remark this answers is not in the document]
ericgribkoff: Done

okhttp/src/main/java/io/grpc/okhttp/OkHttpChannelBuilder.java
```diff
@@ -463,8 +487,8 @@ SSLSocketFactory createSocketFactory() {
     private final boolean usingSharedExecutor;
     private final boolean usingSharedScheduler;
     private final TransportTracer.Factory transportTracerFactory;
-    @Nullable
-    private final SSLSocketFactory socketFactory;
+    @Nullable private final SocketFactory socketFactory;
```
dapengzhang0: Remove `@Nullable`, it's just like a random NPE from any library if ` SocketFactory.getDefault()` were null.

ericgribkoff: Done

PR #5391, opened 2019-02-26
all: when publishing to maven/mavenLocal excludes xds and gae-interops

creamsoup:
Build scripts are copying artifacts in maven local directory, so previous filter doesn't work

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -428,14 +428,7 @@ subprojects {
         sign publishing.publications.maven
     }
 
-    // skip publishing incubatingArtifacts when -PskipIncubatingArtifactsToMavenLocal=true
-    if (rootProject.properties.get('skipIncubatingArtifactsToMavenLocal', false).toBoolean()) {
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -428,8 +428,11 @@ subprojects {
         sign publishing.publications.maven
     }
 
-    publishMavenPublicationToMavenRepository.onlyIf {
-        !name.contains("grpc-gae-interop-testing") && !name.contains("grpc-xds")
+    // skip publishing incubatingArtifacts when -PskipIncubatingArtifacts=true
+    if (rootProject.properties.get('skipIncubatingArtifacts', false).toBoolean()) {
+        publishMavenPublicationToMavenLocal.onlyIf {
```
ejona86: I'm surprised that only MavenLocal is included.

I'm also surprised we don't do this by default.

creamsoup: discussed offline, for current release skip generating artifacts for xds and gae. Filed an issue(#5394) to move this logic in to each subproject.

PR #5395, opened 2019-02-26
Rename source jar outputs to avoid conflicts with java_common.compile

cushon:
More information: https://github.com/bazelbuild/bazel/issues/5824

### DIFF
--- a/buildscripts/kokoro/bazel.sh
+++ b/buildscripts/kokoro/bazel.sh
@@ -3,7 +3,7 @@
 set -exu -o pipefail
 cat /VERSION
 
-use_bazel.sh 0.22.0
+use_bazel.sh 0.23.1
 bazel version
 
 cd github/grpc-java
--- a/java_grpc_library.bzl
+++ b/java_grpc_library.bzl
@@ -20,15 +20,17 @@ def _java_rpc_library_impl(ctx):
     if flavor == "normal":
         flavor = ""
 
+    srcjar = ctx.actions.declare_file("%s-proto-gensrc.jar" % ctx.label.name)
+
     args = ctx.actions.args()
     args.add(ctx.executable._java_plugin.path, format = "--plugin=protoc-gen-grpc-java=%s")
-    args.add("--grpc-java_out={0}:{1}".format(flavor, ctx.outputs.srcjar.path))
+    args.add("--grpc-java_out={0}:{1}".format(flavor, srcjar.path))
     args.add_all(includes, map_each = _create_include_path)
     args.add_all(srcs, map_each = _path_ignoring_repository)
 
     ctx.action(
         inputs = depset([ctx.executable._java_plugin] + srcs, transitive = [includes]),
-        outputs = [ctx.outputs.srcjar],
+        outputs = [srcjar],
         executable = ctx.executable._protoc,
         arguments = [args],
     )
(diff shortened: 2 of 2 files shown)


thelinuxfoundation:

* User @cushon isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

cushon:
I should be covered by a CLA now.

ejona86:
As you can see, this is broken because it also requires an upgrade of Bazel. That's going to need some separate work.

ejona86:
Damn it. This is _only_ compatible with 0.23 and it is broken without it in 0.23, so the fix requires a flag day...

cushon:
Another option that is compatible with 0.22 is to just rename the srcjar output to not conflict with `java_common`, e.g.:

```diff
diff --git a/java_grpc_library.bzl b/java_grpc_library.bzl
index 76264a0d2..5435e7437 100644
--- a/java_grpc_library.bzl
+++ b/java_grpc_library.bzl
@@ -91,7 +91,7 @@ _java_rpc_library = rule(
     fragments = ["java"],
     outputs = {
         "jar": "lib%{name}.jar",
-        "srcjar": "lib%{name}-src.jar",
+        "srcjar": "lib%{name}-source.jar",
     },
     provides = [JavaInfo],
     implementation = _java_rpc_library_impl,
```

This PR is the recommended fix for >=0.23, but the other fix is fine unless the name of that artifact is part of this rule's API.

The incompatible change was `--incompatible_generate_javacommon_source_jar` / https://github.com/bazelbuild/bazel/issues/5824, which was available in 0.22. The `output_source_jar` of `java_common.compile` was a late addition, but using it isn't mandatory.

ejona86:
I tried with `--incompatible_generate_javacommon_source_jar` in 0.22 and it still complained about lack of output_source_jar.

I've generally considered the source jar as part of the rule's API, just given how many users of srcjar I've seen for proto_library and java_library over the years. And that seems reinforced by the fact that your Bazel change is trying to make the srcjar consistent across rules...

Well... it seems like renaming the srcjar from `-src.jar` to `-source.jar` is the best approach. It could still break users on 0.22 and before, but if they are on 0.23+ they should see a `-src.jar`, right? And breaking only users depending on `-src.jar` seems better than breaking everyone.

cushon:
> I tried with `--incompatible_generate_javacommon_source_jar` in 0.22 and it still complained about lack of output_source_jar.

My comment wasn't very clear, but I think that's what I expected. You should see an error building current master (without this PR) using 0.22 and `--incompatible_generate_javacommon_source_jar`.

> It could still break users on 0.22 and before, but if they are on 0.23+ they should see a `-src.jar`, right?

I think that's right. They could still see a difference because there's an additional source jar, but it seems like a fairly safe change.

Another option is to do version or feature detection and use output_source_jar only if it's supported.

ejona86:
> I tried with `--incompatible_generate_javacommon_source_jar` in 0.22 and it still complained about lack of output_source_jar.

Oh, wait. I see. Just passing --incompatible_generate_javacommon_source_jar would generate the -src.jar in 0.22. That's more workable than a full flag day.

ejona86:
> > I tried with `--incompatible_generate_javacommon_source_jar` in 0.22 and it still complained about lack of output_source_jar.
>
> My comment wasn't very clear, but I think that's what I expected. You should see an error building current master (without this PR) using 0.22 and `--incompatible_generate_javacommon_source_jar`.

Sorry, mine was too. I had tried it with this PR. I think we're on the same page now.

smparkes:
Locally patched `java_grpc_library.bz` `0.23.0` until this gets merged and it's working great. Thanks, @cushon.

ejona86:
My plan:
1. merge this to master. v1.20 will require Bazel 0.23+
2. Add a change like https://github.com/grpc/grpc-java/pull/5395#issuecomment-467650439 to v1.19.x for Bazel users that aren't using the srcjar

PR #5396, opened 2019-02-26
all: when publishing to maven/mavenLocal excludes xds and gae-interops (backport of #5391)

creamsoup:


dapengzhang0:
I think we should allow publishing xds locally for testing purpose, at least to see the result of the shadow plugin in xds.

dapengzhang0:
Oh, this is backport, then I don't care.

PR #5397, opened 2019-02-27
core: refactor load-balancing config handling

zhangkun83:
The LoadBalancingConfig message, which looks like
```json
{
  "policy_name" : {
    "config_key1" : "config_value1",
    "config_key2" : "config_value2"
   }
}
```
appears multiple times. It gets super tedious and confusing to handle, because both the whole config and the value (in the above example is `{ "config_key1" : "config_value1" }`) are just `Map<String, Object>`, and each user needs to do the following validation:
 1. The whole config must have exactly one key
 2. The value must be a map

Here I define `LbConfig` that holds the policy name and the config value, and a method in `ServiceConfigUtil` that converts the parsed JSON format into `LbConfig`.

There is also multiple cases where you need to handle a list of configs (top-level balancing policy, child and fallback policies in xds, grpclb child policies). I also made another helper method in `ServiceConfigUtil` to convert them into `List<LbConfig>`.

Found and fixed a bug in the xds code, where the top-level balancer should pass the config value (excluding the policy name), not the whole config to the child balancers. Search for "supported_1_option" in the diff to see it in the tests.


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1318,7 +1319,7 @@ public void run() {
               if (retryEnabled) {
                 throttle = getThrottle(config);
               }
-            } catch (RuntimeException re) {
+            } catch (MalformedConfigException re) {
```
carl-mastrangelo: I don't think this is a safe change to make.  Previously, RuntimeException was pretty wide, and any error processing service config would be logged.   This makes gRPC resilient in case of unforeseen service config problems.  And able to be fixed if a bad service config was pushed.

Won't this make panic mode much more likely?

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -345,52 +364,86 @@ static Integer getMaxResponseMessageBytesFromMethodConfig(Map<String, Object> me
...
+   * Given a JSON list of LoadBalancingConfigs, and convert it into a list of LbConfig.
+   */
+  @SuppressWarnings("rawtypes")
+  public static List<LbConfig> unwrapLoadBalancingConfigList(Object listObject)
+      throws MalformedConfigException {
+    List list;
+    try {
+      list = (List) listObject;
+    } catch (ClassCastException e) {
+      throw new MalformedConfigException("List expected, but is " + listObject);
```
carl-mastrangelo: the previous exception needs to be chained here.

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -642,4 +697,64 @@ private static long saturatedAdd(long a, long b) {
     // we did over/under flow, if the sign is negative we should return MAX otherwise MIN
     return Long.MAX_VALUE + ((naiveSum >>> (Long.SIZE - 1)) ^ 1);
   }
+
+  /**
+   * A LoadBalancingConfig that includes the policy name (the key) and its raw config value (parsed
+   * JSON).
+   */
+  public static final class LbConfig {
```
carl-mastrangelo: package privatae

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -642,4 +697,64 @@ private static long saturatedAdd(long a, long b) {
...
+    private final String policyName;
+    private final Map<String, Object> rawConfigValue;
+
+    public LbConfig(String policyName, Map<String, Object> rawConfigValue) {
```
carl-mastrangelo: same

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -642,4 +697,64 @@ private static long saturatedAdd(long a, long b) {
...
+    private static final long serialVersionUID = 1L;
+
+    private MalformedConfigException(String msg) {
+      super(msg);
+    }
+
+    private MalformedConfigException(String msg, Throwable cause) {
+      super(msg, cause);
+    }
+  }
```
carl-mastrangelo: ServiceConfigUtil doesn't seem like an appropriate name any more.  I think you should rename the class, or move the inner types out.

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -642,4 +697,64 @@ private static long saturatedAdd(long a, long b) {
...
+          .add("policyName", policyName)
+          .add("rawConfigValue", rawConfigValue)
+          .toString();
+    }
+  }
+
+  /**
+   * Thrown when the config is a valid JSON, but doesn't conform to the config scheme.
+   */
+  public static final class MalformedConfigException extends Exception {
```
carl-mastrangelo: package private

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -642,4 +697,64 @@ private static long saturatedAdd(long a, long b) {
...
+    public String toString() {
+      return MoreObjects.toStringHelper(this)
+          .add("policyName", policyName)
+          .add("rawConfigValue", rawConfigValue)
+          .toString();
+    }
+  }
+
+  /**
+   * Thrown when the config is a valid JSON, but doesn't conform to the config scheme.
```
carl-mastrangelo: s/scheme/schema/

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -171,8 +172,13 @@ public void handleResolvedAddressGroups(
         List<EquivalentAddressGroup> servers, Attributes attributes) {
       Map<String, Object> serviceConfig =
           attributes.get(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG);
-      String serviceName = ServiceConfigUtil.getHealthCheckedServiceName(serviceConfig);
-      helper.setHealthCheckedService(serviceName);
+      try {
+        String serviceName = ServiceConfigUtil.getHealthCheckedServiceName(serviceConfig);
+        helper.setHealthCheckedService(serviceName);
+      } catch (MalformedConfigException e) {
```
carl-mastrangelo: I dont think this was right.   Exceptions in service config should be treated more carefully.  A runtimeException here could cause a channel to stop working and never recover, which is bad.


Service config needs to be able to be fully validated before replacing the existing one, incase of problems.  If an exception comes out due to user provided data, and that causes the channel to explode, then it has the same problem as replacing the existing config.   A new service config would result in an unrecoverable state.

zhangkun83:
> I'm pretty uncomfortable exposing any more of `internal` to `services`. I would prefer a top level exception instead, since basically all load balancers / name resolver would need to access it.

I was about to move the exception to `io.grpc` but I recalled @ejona86 was planning on returning a {object, error} tuple from "parse()" (as part of the service config error-handling APIs) rather than throwing. @ejona86, can you remind me the reason? I don't quite remember. Throwing fits very well in `ServiceConfigUtil`, because even the parsing of a single primitive could throw. If they returned the {object, error} tuple instead, the code would become quite tedious.

ejona86:
> I was about to move the exception to `io.grpc` but I recalled @ejona86 was planning on returning a {object, error} tuple from "parse()" (as part of the service config error-handling APIs) rather than throwing. @ejona86, can you remind me the reason? I don't quite remember.

The NameResolver will be passing the result as-is in the Attributes of onAddresses. The result is the tuple "(cookie, status)", since ChannelImpl needs to know precisely what the results are. So an exception simply wouldn't work.

The aversion against throwing is also because throwing is "very bad" in the rest of our code. It seems a bad idea to have "you must not throw here" in some parts of the API and "you must throw here" in others. It will be hard for application authors to know when to do what.

> Throwing fits very well in `ServiceConfigUtil`, because even the parsing of a single primitive could throw. If they returned the {object, error} tuple instead, the code would become quite tedious.

I think the code will be a bit tedious in any case, to have good enough error messages. But I do agree exceptions are still easier to do that with. I sort of imagine using exceptions within a module, but the tuple between modules.

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
ejona86: If this will be public for use by LB policies, then it shouldn't be in `internal`. No implementation should be required to use `internal` code. Since hierarchical policies are core to the design, that means we need a non-internal way to handle subpolicies. 

We could expose the parser as a utility function that returns the normal `(cookie, status)`. We would then have a utility (on Helper?) to construct a LB policy provided the cookie. The cookie would be this LbConfig, but LbConfig would be private.

If parent policies are allowed to special-case certain names (which seems like it may be a bad idea, but I expect there are arguments on both sides), then we would be required to make the class public and it would be part of our API in `io.grpc` or say those use-cases must parse the LbConfig themselves (and access the LB registry manually).

(For the gRPC-LB v1 goo, it will special-case certain names, but in that case it isn't actually going to need the pre-built LbConfig parsing since it will be ignoring the subpolicy configuration anyway.)

-----

Note that this wouldn't be used as-is with the API I was talking about. It wouldn't be a `rawConfigValue` but instead the child policy's cookie. But I understand we can't plumb that today.

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -345,52 +367,86 @@ static Integer getMaxResponseMessageBytesFromMethodConfig(Map<String, Object> me
...
+    return new LbConfig(entry.getKey(), configValue);
+  }
+
+  /**
+   * Given a JSON list of LoadBalancingConfigs, and convert it into a list of LbConfig.
+   */
+  @SuppressWarnings("rawtypes")
+  public static List<LbConfig> unwrapLoadBalancingConfigList(Object listObject)
+      throws MalformedConfigException {
+    List list;
```
dapengzhang0: Use `@SuppressWarnings("unchecked")` and `List<Object>` or `List<?>` instead of `@SuppressWarnings("rawtypes")` and `List`?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1316,13 +1316,14 @@ public void run() {
             try {
               serviceConfigInterceptor.handleUpdate(serviceConfig);
               if (retryEnabled) {
-                throttle = getThrottle(config);
+                throttle = ServiceConfigUtil.getThrottlePolicy(
+                    config.get(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG));
```
dapengzhang0: This can be simply `throttle = ServiceConfigUtil.getThrottlePolicy(serviceConfig)`

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1316,13 +1316,14 @@ public void run() {
...
               }
-            } catch (RuntimeException re) {
+            } catch (Exception e) {
```
dapengzhang0: Move the `//TODO` in the doc of `getThrottlePolicy` here. The //TODO is not only for `getThrottlePolicy`, but also for `getDouble()`, `getList()` etc.

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
dapengzhang0: If it shouldn't be `internal`, it might be `@Internal` for the moment.

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -642,4 +700,64 @@ private static long saturatedAdd(long a, long b) {
...
+          .add("policyName", policyName)
+          .add("rawConfigValue", rawConfigValue)
+          .toString();
+    }
+  }
+
+  /**
+   * Thrown when the config is a valid JSON, but doesn't conform to the config schema.
+   */
+  public static final class MalformedConfigException extends Exception {
```
dapengzhang0: This should be a public API, because `ManageChannelBuilder` may throw it if we pass a default config to it.

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
ejona86: I don't want `ServiceConfigUtil` to be public. Maybe a particular method/class may need to be split out, but there's far too many random things going on here. For example, things like getFallbackPolicyFromXdsConfig should really live in XDS.

dapengzhang0: >getFallbackPolicyFromXdsConfig should really live in XDS.

Yeah, I was considering putting `getFallbackPolicyFromXdsConfig` in XDS, but it reuses too many methods in `ServiceConfigUtil`, and need large refactoring so I gave up.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -94,8 +97,12 @@ public static String getHealthCheckedServiceName(@Nullable Map<String, Object> s
     return getString(healthCheck, "serviceName");
   }
 
+  // TODO(zhangkun83): ideally this method should throw only MalformedConfigException when the
+  // config is malformed, so that if there is a bug in the code, a RuntimException would be thrown
+  // and make the channel panic as expected.
   @Nullable
-  static Throttle getThrottlePolicy(@Nullable Map<String, Object> serviceConfig) {
+  static Throttle getThrottlePolicy(@Nullable Map<String, Object> serviceConfig)
+      throws Exception {
```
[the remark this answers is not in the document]
zhangkun83: I removed the TODO since I get rid of MalformedConfigException.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
zhangkun83: I removed the TODO since I got rid of MalformedConfigException.

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
zhangkun83: Done.

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
zhangkun83: I meant that this **internal** class has to have **public access** as of now, but I expect it to evolve into a public API in a possibly different form. This PR is **not** going to propose any public API. It is merely an internal refactor to keep our own LoadBalancingConfig handling path sane.

zhangkun83:
@ejona86 @carl-mastrangelo @dapengzhang0 PTAL

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -411,101 +455,101 @@ public static String getStickinessMetadataKeyFromServiceConfig(
...
    */
   @SuppressWarnings("unchecked")
-  static List<Object> getList(Map<String, Object> obj, String key) {
+  static List<Object> getList(Map<String, Object> obj, String key) throws Exception {
     assert obj.containsKey(key);
     Object value = checkNotNull(obj.get(key), "no such key %s", key);
     if (value instanceof List) {
       return (List<Object>) value;
     }
-    throw new ClassCastException(
```
carl-mastrangelo: ClassCastException was right before.  These getter methods imply a specific type, which ClassCastException is for.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
```diff
@@ -97,18 +101,24 @@ public void handleResolvedAddressGroups(
...
+            helper.getChannelLogger().log(
+                ChannelLogLevel.WARNING,
+                "Binary stickiness header is not supported. The header \"{0}\" will be ignored",
+                stickinessMetadataKey);
+          } else if (stickinessState == null
+              || !stickinessState.key.name().equals(stickinessMetadataKey)) {
+            stickinessState = new StickinessState(stickinessMetadataKey);
+          }
         }
+      } catch (Exception e) {
```
carl-mastrangelo: I dont think you can catch this here.  Even if you *eventually* plan on removing this, that doesn't mean you can break it now.  This should rethrow as a RuntimeException.

core/src/test/java/io/grpc/internal/ServiceConfigUtilTest.java
```diff
@@ -51,72 +41,186 @@ public void getBalancerNameFromXdsConfig() throws Exception {
...
+    assertThat(config.getPolicyName()).isEqualTo("xds_experimental");
+    assertThat(config.getRawConfigValue()).isEqualTo(JsonParser.parse(
+            "{\"balancerName\" : \"dns:///balancer.example.com:8080\","
+            + "\"childPolicy\" : [{\"round_robin\" : {}}, {\"lbPolicy2\" : {\"key\" : \"val\"}}]"
+            + "}"));
+  }
+
+  @Test
+  public void unwrapLoadBalancingConfig_tooManyFields() throws Exception {
+    // A LoadBalancingConfig should not have more than one fields.
```
carl-mastrangelo: s/fields/field/

core/src/test/java/io/grpc/internal/ServiceConfigUtilTest.java
```diff
@@ -51,72 +41,186 @@ public void getBalancerNameFromXdsConfig() throws Exception {
...
+    String lbConfig = "{\"xds_experimental\" : { "
+        + "\"balancerName\" : \"dns:///balancer.example.com:8080\","
+        + "\"childPolicy\" : [{\"round_robin\" : {}}, {\"lbPolicy2\" : {\"key\" : \"val\"}}]"
+        + "},"
+        + "\"grpclb\" : {} }";
+    try {
+      ServiceConfigUtil.unwrapLoadBalancingConfig(JsonParser.parse(lbConfig));
+      fail("Should throw");
+    } catch (Exception e) {
+      assertThat(e.getMessage()).contains("There are 2 fields");
```
carl-mastrangelo: This should be `assertThat(e).hasMessageThat().contains("There are 2 fields")`

The difference is that if the test fails, you get the whole exception as context.

core/src/test/java/io/grpc/internal/ServiceConfigUtilTest.java
```diff
@@ -51,72 +41,186 @@ public void getBalancerNameFromXdsConfig() throws Exception {
...
+    try {
+      ServiceConfigUtil.unwrapLoadBalancingConfig(JsonParser.parse(lbConfig));
+      fail("Should throw");
+    } catch (Exception e) {
+      assertThat(e.getMessage()).contains("There are 0 fields");
+    }
+  }
+
+  @Test
+  public void unwrapLoadBalancingConfig_config_is_list() throws Exception {
```
carl-mastrangelo: The name of this function is wrong.  Underscores separate the test dimensions, not sentences.  Please use CamelCase.

carl-mastrangelo: Furthermore, the name should be more of an expectation.   Like, `unwrapLoadBalancingConfig_failsOnList`

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
```diff
@@ -171,8 +171,13 @@ public void handleResolvedAddressGroups(
         List<EquivalentAddressGroup> servers, Attributes attributes) {
       Map<String, Object> serviceConfig =
           attributes.get(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG);
-      String serviceName = ServiceConfigUtil.getHealthCheckedServiceName(serviceConfig);
-      helper.setHealthCheckedService(serviceName);
+      try {
+        String serviceName = ServiceConfigUtil.getHealthCheckedServiceName(serviceConfig);
```
carl-mastrangelo: same comment, I think the best you can do is rethrow as RE, and have it logged top level in LbHelperImpl in ManagedChannelImpl

xds/src/main/java/io/grpc/xds/XdsLoadBalancer.java
```diff
@@ -89,18 +89,23 @@ public void onError() {
...
-    xdsLbState.handleResolvedAddressGroups(servers, attributes);
+    try {
+      LbConfig newLbConfig = ServiceConfigUtil.unwrapLoadBalancingConfig(newRawLbConfig);
+      fallbackPolicy = selectFallbackPolicy(newLbConfig, lbRegistry);
+      fallbackManager.updateFallbackServers(servers, attributes, fallbackPolicy);
+      fallbackManager.maybeStartFallbackTimer();
+      handleNewConfig(newLbConfig);
+      xdsLbState.handleResolvedAddressGroups(servers, attributes);
+    } catch (Exception e) {
+      // TODO: handle it
```
carl-mastrangelo: Sorry, this is not acceptable.

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
zhangkun83: Done.

core/src/test/java/io/grpc/internal/ServiceConfigUtilTest.java
zhangkun83: Done.

core/src/test/java/io/grpc/internal/ServiceConfigUtilTest.java
zhangkun83: Done.

core/src/test/java/io/grpc/internal/ServiceConfigUtilTest.java
zhangkun83: Done.

services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
zhangkun83: We are not going to have `ManagedChannelImpl` blindly catch and log all RE. It won't be the case either once we have the 2-step parse/apply implementation where the parsing will be moved out of this method. It is wrong for `ManagedChannelImpl` to catch-and-log REs thrown from `handleResolvedAddressGroups()` instead of panicking.

According to @markdroth, C++ just ignores malformed health check config, which is what I am doing in this PR. If you prefer to keep the current behavior, I can rethrow it and let the channel panic. Let me know which one you'd like.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer.java
zhangkun83: I have changed it to rethrowing so it's keeping the current behavior.

zhangkun83:
@carl-mastrangelo I have reverted the exception-related changes and other two minor unrelated changes. Now the PR only contains `LbConfig`-related stuff. PTAL.

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -345,52 +347,81 @@ static Integer getMaxResponseMessageBytesFromMethodConfig(Map<String, Object> me
...
    */
-  public static String getBalancerPolicyNameFromLoadBalancingConfig(Map<String, Object> lbConfig) {
-    return lbConfig.entrySet().iterator().next().getKey();
+  @SuppressWarnings("unchecked")
+  public static LbConfig unwrapLoadBalancingConfig(Object lbConfig) {
+    Map<String, Object> map;
+    try {
+      map = (Map<String, Object>) lbConfig;
+    } catch (ClassCastException e) {
+      throw new ClassCastException("Invalid type. Config=" + lbConfig);
```
carl-mastrangelo: chain the error?

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -345,52 +347,81 @@ static Integer getMaxResponseMessageBytesFromMethodConfig(Map<String, Object> me
...
+      throw new RuntimeException(
+          "There are " + map.size() + " fields in a LoadBalancingConfig object. Exactly one"
+          + " is expected. Config=" + lbConfig);
+    }
+    Map.Entry<String, Object> entry = map.entrySet().iterator().next();
+    Map<String, Object> configValue;
+    try {
+      configValue = (Map<String, Object>) entry.getValue();
+    } catch (ClassCastException e) {
+      throw new ClassCastException("Invalid value type.  value=" + entry.getValue());
```
carl-mastrangelo: same comment

core/src/main/java/io/grpc/internal/ServiceConfigUtil.java
```diff
@@ -345,52 +347,81 @@ static Integer getMaxResponseMessageBytesFromMethodConfig(Map<String, Object> me
...
-  public static String getBalancerNameFromXdsConfig(
-      Map<String, Object> xdsConfig) {
-    Object entry = xdsConfig.entrySet().iterator().next().getValue();
-    return getString((Map<String, Object>) entry, XDS_CONFIG_BALANCER_NAME_KEY);
+  public static List<LbConfig> unwrapLoadBalancingConfigList(Object listObject) {
+    List<?> list;
+    try {
+      list = (List<?>) listObject;
+    } catch (ClassCastException e) {
+      throw new ClassCastException("List expected, but is " + listObject);
```
carl-mastrangelo: and here

PR #5398, opened 2019-02-27
all: fix publish not having checksum

creamsoup:


build.gradle
```diff
@@ -428,7 +432,7 @@ subprojects {
         sign publishing.publications.maven
     }
 
-    [publishMavenPublicationToMavenRepository, publishMavenPublicationToMavenLocal]*.onlyIf {
```
ejona86: Go ahead and leave this. I find it unexpected when maven local installs produce different results.

creamsoup: done

PR #5358, opened 2019-02-13
Add an example of server side compression support 

AmiDavidW:
I was looking for a way to enable the compression of server responses, but cannot find a direct answer, as mentioned [here](https://github.com/grpc/grpc-java/issues/2204).

I figured out a way to get it working and would like to have some comments and feedbacks.

It would be great if we have an officially recommended way to implement it.

PR #5401, opened 2019-02-27
core,stub: Document grpc performs compression nego on server

ejona86:
CC @AmiDavidW, as this came up in #5358


core/src/main/java/io/grpc/ServerCall.java
```diff
@@ -180,6 +180,9 @@ public void setMessageCompression(boolean enabled) {
    * The compressor to use will be looked up in the {@link CompressorRegistry}.  Default gRPC
    * servers support the "gzip" compressor.
    *
+   * <p>It is safe to call this even if the client does not support the compression format chosen.
```
carl-mastrangelo: Optional: you might call out that the analogous client side call (CallOptions.withCompression) does not safely negotiate.

ejona86:
Added documentation to CallOptions. There was already documentation on AbstractStub.

PR #5405, opened 2019-02-27
testing: Move AbstractTransportTest to core to avoid Truth dep

### LINKED ISSUE #5301: Remove  compileOnly libraries.truth

ejona86:
Fixes #5301 

----

CC @elharo @suztomo. This will be present in v1.20.

### DIFF
--- a/core/src/test/java/io/grpc/inprocess/InProcessTransportTest.java
+++ b/core/src/test/java/io/grpc/inprocess/InProcessTransportTest.java
@@ -18,10 +18,10 @@ package io.grpc.inprocess;
 
 import com.google.common.collect.ImmutableList;
 import io.grpc.ServerStreamTracer;
+import io.grpc.internal.AbstractTransportTest;
 import io.grpc.internal.GrpcUtil;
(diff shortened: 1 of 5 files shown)


dapengzhang0:
Why not move the entire package to core?

ejona86:
It became a much bigger task. The Test{,Client,Server}Tracers are used by interop-testing/src/main which is published as an artifact, and parts of TestUtils are using by benchmarks, as well as something else is used by benchmarks. It would be good to work on these issues, but there was no need to yak shave at this point in time.

PR #5407, opened 2019-02-27
SECURITY.md: Don't suggest users to use Jetty ALPN

ejona86:
I fixed many of these elsewhere in the document via 4ba168f28, but I
didn't update the troubleshooting section previously.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -53,8 +53,8 @@ support or have 2% the performance of OpenSSL.
 
 We recommend most users use grpc-netty-shaded, which includes netty-tcnative on
 BoringSSL. It includes pre-built libraries for 64 bit Windows, OS X, and 64 bit
-Linux. For 32 bit Windows, Conscrypt may be an option. For all other platforms,
-Java 9+ is required.
+Linux. For 32 bit Windows, Conscrypt is an option. For all other platforms, Java
+9+ is required.
 
 For users of grpc-netty we recommend [netty-tcnative with
 BoringSSL](#tls-with-netty-tcnative-on-boringssl), although using the built-in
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -329,7 +329,10 @@ If you aren't using gRPC on Android devices, you are most likely using `grpc-net
 
 If you are developing for Android and have a dependency on `grpc-netty`, you should remove it as `grpc-netty` is unsupported on Android. Use `grpc-okhttp` instead.
 
-If you are on a 32-bit operating system, or not on a [Transport Security supported platform](#transport-security-tls), you should use Jetty ALPN (and beware of potential issues), or you'll need to build your own 32-bit version of `netty-tcnative`.
+If you are on a 32-bit operating system, using Java 11+ may be the easiest
+solution, as ALPN was added to Java in Java 9. If on 32-bit Windows, Conscrypt
+may be an option. Otherwise you need to build your own 32-bit version of
```
sanjaypujare: Instead of the tentative "may be" wording can we say something like this:

If on 32-bit Windows and you cannot use Java 11+, you have the option of using [Conscrypt](#TLS with Conscrypt).

SECURITY.md
```diff
@@ -329,7 +329,10 @@ If you aren't using gRPC on Android devices, you are most likely using `grpc-net
...
+`netty-tcnative`.
```
sanjaypujare: `netty-tcnative` -> [`netty-tcnative`](https://github.com/netty/netty-tcnative)

ejona86: We reference netty-tcnative _lots_ of times in this doc. It is a bad idea to make each one a link. We have a link to it in the TLS on non-Android section. I would be okay with linking to a "compiling netty-tcnative" section, but we don't have such a section and really the users are on their own at that point.

SECURITY.md
ejona86: I guess I should just say "If on 32-bit Windows, Conscrypt is an option" instead of "may be". And make it a link. Does that sound good?

sanjaypujare: Yes. The "may be" wording creates unnecessary uncertainty (I think).

ejona86: Done.

SECURITY.md
sanjaypujare: How about then linking to https://netty.io/wiki/forked-tomcat-native.html#wiki-h2-6 ? Assuming that's the correct page for compiling netty-tcnative.

ejona86: Good idea. I had forgotten that documentation existed.

Issue #5383, opened 2019-02-22
Bazel incompatible changes


laurentlb:
This repository doesn't build with Bazel future incompatible changes, and will break with future releases of Bazel.

In order to detect the issues in advance, you can try Bazelisk:
  [`bazelisk --migrate test //...`](https://github.com/philwo/bazelisk)

Some of the issues can be fixed automatically using Buildifier:
  [`buildifier --lint=fix path/to/file.bzl`](https://github.com/bazelbuild/buildtools)

ejona86:
@laurentlb, what version of Bazel should we test with?

laurentlb:
In general, the latest version of Bazel. This would be 0.22 (or 0.23, released a few minutes ago, the announcement is pending).

If you use bazelisk, it will automatically fetch the latest version.

ejona86:
I asked because 1) there was no statement as to what is broken and 2) I had run with 0.22 already. In fact, it still is working. Please in the future provide at least a subset of the errors you see.

```
grpc-java$ bazel version
INFO: Invocation ID: 8e1df78a-d4ba-4462-9100-3d7cfc3aae25
Build label: 0.22.0
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Mon Jan 28 12:58:08 2019 (1548680288)
Build timestamp: 1548680288
Build timestamp as int: 1548680288

grpc-java$ bazel build ...
INFO: Invocation ID: fbcf25a5-1e7e-4e53-b155-154e9d773273
INFO: Analysed 30 targets (0 packages loaded, 0 targets configured).
INFO: Found 30 targets...
INFO: Elapsed time: 0.329s, Critical Path: 0.00s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
```

laurentlb:
```
$ USE_BAZEL_VERSION=0.22.0 bazelisk --migrate build //... --nobuild
...
Migration is needed for the following flags:
  --incompatible_disable_legacy_proto_provider
  --incompatible_no_support_tools_in_action_inputs
  --incompatible_new_actions_api
  --incompatible_disable_deprecated_attr_params
```

If you run Bazel with any of these 4 flags, the build fails. We plan to enable these flags by default in a future version of Bazel, so fixing the issue know will prevent a future breakage.

[...]

PR #5411, opened 2019-02-28
bazel: update to new ProtoInfo provider

drigz:
This is part of #5383. As Bazel 0.22.0 is the only supported version
now, use this in the Kokoro build.


drigz:
@ejona86 Is there a way to test this on Kokoro before merging?

PR #5413, opened 2019-03-01
netty: replace TLS protocol negotiator with new style handlers

carl-mastrangelo:


netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -268,70 +270,125 @@ public void close(ChannelHandlerContext ctx, ChannelPromise future) throws Excep
...
+        }
+      } else {
+        super.userEventTriggered(ctx, evt);
+      }
+    }
+
+    private void fireProtocolNegotiationEvent(ChannelHandlerContext ctx, SSLSession session) {
+      Security security = new Security(new Tls(session));
+      Attributes attrs = pne.getAttributes().toBuilder()
+          .set(GrpcAttributes.ATTR_SECURITY_LEVEL, SecurityLevel.PRIVACY_AND_INTEGRITY)
```
creamsoup: couple attrs (TRANSPORT_ATTR_REMOTE_ADDR, TRANSPORT_ATTR_LOCAL_ADDR) are missing compare to BufferUntilTlsNegotiatedHandler, doesn't look like critical ones, was this intended?

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -268,70 +270,125 @@ public void close(ChannelHandlerContext ctx, ChannelPromise future) throws Excep
...
+      super.handlerAdded(ctx);
     }
 
     @Override
-    public void close() {}
+    public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+      if (evt instanceof ProtocolNegotiationEvent) {
+        pne = (ProtocolNegotiationEvent) evt;
+      } else if (evt instanceof SslHandshakeCompletionEvent) {
+        SslHandshakeCompletionEvent handshakeEvent = (SslHandshakeCompletionEvent) evt;
```
creamsoup: the ProtocolNegotiationEvent must be fired before SslHandshakeCompletionEvent? if it sees the SslHandshakeCompletionEvent before ProtocolNegotiationEvent, should it fail?

carl-mastrangelo: It may, but I wanted to avoid too much state in the handler.  The idea for these handlers is that they are pluggable and single purpose.  In that regard, they don't assume that a previous PNE, or else they would assume a previous handler, which is tight coupling.

PNE cannot be fired after SHCE, because it removed this handler from the pipeline.  Thus, the invariant is held that if there is is a PNE, it must happen before the SHCE.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
carl-mastrangelo: Yes.  These come from the WUAH firing an event later.   I can add them in the test.

netty/src/test/java/io/grpc/netty/ProtocolNegotiatorsTest.java
```diff
@@ -585,6 +582,7 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
     elg.shutdownGracefully();
   }
 
+
```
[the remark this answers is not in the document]
carl-mastrangelo: Done.

netty/src/test/java/io/grpc/netty/ProtocolNegotiatorsTest.java
```diff
@@ -596,24 +594,77 @@ public void bufferingHandler_shouldNotThrowForEmptyHandler() throws Exception {
         .register().sync();
     ChannelFuture sf = new ServerBootstrap()
         .channel(LocalServerChannel.class)
-        .childHandler(new ChannelHandlerAdapter() {})
+        .childHandler(new ChannelHandlerAdapter() {
+        })
```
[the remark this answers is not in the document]
carl-mastrangelo: Whoops, done.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
creamsoup: oh yeah, i missed that one.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
creamsoup: okay, i see the point. keeping default to null vs default attribute doesn't seems like having much of state than allows unexpected behavior in some weird race condition. since i don't see the race can happen, i am fine with that.

PR #5415, opened 2019-03-01
core: add LoadBalancer.Helper#createResolvingOobChannel()

zhangkun83:
This can be used by xds LoadBalancer to create a channel to the XDS
traffic director, as the service config will only specify the target
name of the balancer.

This PR only adds the interface to unblock the xds work.
Implementation would take some time thus will come later.

I selected a slightly different name from the original `createOobChannel()`
to signify their difference. I am open to naming it just `createOobChannel()`.

### DIFF
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -551,9 +551,10 @@ public abstract class LoadBalancer {
     public abstract ManagedChannel createOobChannel(EquivalentAddressGroup eag, String authority);
 
     /**
-     * Updates the addresses used for connections in the {@code Channel}. This is supperior to
-     * {@link #createOobChannel} when the old and new addresses overlap, since the channel can
-     * continue using an existing connection.
+     * Updates the addresses used for connections in the {@code Channel} that was created by {@link
+     * #createOobChannel(EquivalentAddressGroup, String)}. This is supperior to {@link
+     * #createOobChannel(EquivalentAddressGroup, String)} when the old and new addresses overlap,
+     * since the channel can continue using an existing connection.
      *
      * @throws IllegalArgumentException if {@code channel} was not returned from {@link
      *     #createOobChannel}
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -563,6 +564,24 @@ public void updateOobChannelAddresses(ManagedChannel channel, EquivalentAddressG
...
     }
 
+    /**
+     * Creates an out-of-band channel for LoadBalancer's own RPC needs, e.g., talking to an external
+     * load-balancer service, that is specified by a target string.  See the documentation on
+     * {@link ManagedChannelBuilder#forTarget} for the format of a target string.
+     *
+     * <p>The target string will be resolved by a {@link NameResolver} created according to the
+     * target string.  The out-of-band channel doesn't have load-balancing. If multiple addresses
+     * are resolved for the target, the "pick_first" policy will be applied.
```
dapengzhang0: `the "pick_first" policy will be applied` seems implementation detail, or an instruction for implementors, but the users don't care what's under the hood.

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -563,6 +564,24 @@ public void updateOobChannelAddresses(ManagedChannel channel, EquivalentAddressG
...
+     *
+     * <p>The LoadBalancer is responsible for closing unused OOB channels, and closing all OOB
+     * channels within {@link #shutdown}.
+     *
+     * <P>NOT IMPLEMENTED: this method is currently a stub and not yet implemented by gRPC.
+     */
+    public ManagedChannel createResolvingOobChannel(String target) {
```
dapengzhang0: I'm not against a slight different name, but "createResolvingOobChannel" may not be better than "createOobChannel".

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -563,6 +564,24 @@ public void updateOobChannelAddresses(ManagedChannel channel, EquivalentAddressG
...
+     *
+     * <p>The target string will be resolved by a {@link NameResolver} created according to the
+     * target string.  The out-of-band channel doesn't have load-balancing. If multiple addresses
+     * are resolved for the target, the "pick_first" policy will be applied.
+     *
+     * <p>The LoadBalancer is responsible for closing unused OOB channels, and closing all OOB
+     * channels within {@link #shutdown}.
+     *
+     * <P>NOT IMPLEMENTED: this method is currently a stub and not yet implemented by gRPC.
+     */
```
dapengzhang0: nit: Add `@since`

core/src/main/java/io/grpc/LoadBalancer.java
zhangkun83: Fair. I changed it to a descriptive phrasing.

core/src/main/java/io/grpc/LoadBalancer.java
zhangkun83: Done.

PR #5416, opened 2019-03-02
services: handle malformed health-check config gracefully.

zhangkun83:
Previously a bad config will make the channel panic. According to
@markdroth, gRPC C++ just ignores it. Here we will log a warning in
addition.


YifeiZhuang (approved):

markdroth:
I don't remember this conversation.  Can you remind me of the context here?

Are we talking about the health check config in service config?  If so, it's not clear to me how it can be malformed, because invalid JSON would be rejected by the service config parser.

zhangkun83:
After 5 years, the context has left my brain ...

shivaspeaks:
Closing this because nobody remembers the context.

PR #5419, opened 2019-03-02
core/grpclb: propagation and parsing of grpclb config

zhangkun83:
Make sure the config for grpclb is passed to the `GrpclbLoadBalancer`, which will support two child policies -- "round_robin" (default) and "pick_first".

Previously the presence of balancer addresses would dictate "grpclb" policy, despite of the service config. Service config will now take precedence instead.

Implement config parsing logic in `GrpclbLoadBalancer`. Per offline discussions with @markdroth and @ejona86, we will ignore configuration errors for now. The more appropriate [config error handling](https://github.com/grpc/proposal/pull/100) is upcoming.

### DIFF
--- a/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
+++ b/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
@@ -257,9 +257,7 @@ public final class AutoConfiguredLoadBalancerFactory extends LoadBalancer.Factor
             LoadBalancerProvider provider = registry.getProvider(policy);
             if (provider == null) {
               policiesTried.add(policy);
-            }
-
-            if (provider != null) {
+            } else {
               if (!policiesTried.isEmpty()) {
                 // Before returning, log all previously tried policies
                 helper.getChannelLogger().log(
--- a/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
+++ b/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
@@ -16,7 +16,6 @@
 
 package io.grpc.internal;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -223,25 +222,23 @@ public final class AutoConfiguredLoadBalancerFactory extends LoadBalancer.Factor
         lbConfigs = ServiceConfigUtil.unwrapLoadBalancingConfigList(rawLbConfigs);
       }
       if (lbConfigs != null && !lbConfigs.isEmpty()) {
-        // If there isn't a preselecbted, try to get the provider as configured
         LinkedHashSet<String> policiesTried = new LinkedHashSet<>();
         for (LbConfig lbConfig : lbConfigs) {
           String policy = lbConfig.getPolicyName();
-          if (policy.equals(GRPCLB_POLICY_NAME)) {
-            return selectGrpclb(servers, backendAddrs, lbConfig.getRawConfigValue());
+          LoadBalancerProvider provider = registry.getProvider(policy);
+          if (provider == null) {
+            policiesTried.add(policy);
           } else {
-            LoadBalancerProvider provider = registry.getProvider(policy);
-            if (provider == null) {
-              policiesTried.add(policy);
-            } else {
-              if (!policiesTried.isEmpty()) {
-                // Before returning, log all previously tried policies
-                helper.getChannelLogger().log(
-                    ChannelLogLevel.DEBUG,
-                    "{0} specified by Service Config are not available", policiesTried);
-              }
-              return new PolicySelection(provider, backendAddrs, lbConfig.getRawConfigValue());
+            if (!policiesTried.isEmpty()) {
+              // Before returning, log all previously tried policies
+              helper.getChannelLogger().log(
+                  ChannelLogLevel.DEBUG,
+                  "{0} specified by Service Config are not available", policiesTried);
             }
+            return new PolicySelection(
+                provider,
+                policy.equals(GRPCLB_POLICY_NAME) ? servers : backendAddrs,
+                lbConfig.getRawConfigValue());
           }
         }
         if (!haveBalancerAddress) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -247,22 +249,41 @@ PolicySelection decideLoadBalancerProvider(
         lbConfigs = ServiceConfigUtil.unwrapLoadBalancingConfigList(rawLbConfigs);
       }
       if (lbConfigs != null && !lbConfigs.isEmpty()) {
```
ejona86: lbConfig parsing wins out over haveBalancerAddress. So this order looks wrong. It looks like this code should move before the haveBalancerAddress code.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -84,7 +95,42 @@ public void handleResolvedAddressGroups(
...
+    Map<String, Object> rawLbConfigValue = attributes.get(ATTR_LOAD_BALANCING_CONFIG);
+    Mode newMode = retrieveModeFromLbConfig(rawLbConfigValue, helper.getChannelLogger());
+    grpclbState.handleAddresses(newLbAddressGroups, newBackendServers, newMode);
+  }
+
+  @VisibleForTesting
+  static Mode retrieveModeFromLbConfig(
+      @Nullable Map<String, Object> rawLbConfigValue, ChannelLogger channelLogger) {
+    try {
+      if (rawLbConfigValue != null) {
```
ejona86: Inverting these conditions would make the code more clear avoiding lots of indentation:

```java
if (rawLbConfigValue == null) {
  return DEFAULT_MODE;
}
Object rawChildPolicies = rawLbConfigValue.get("childPolicy");
if (rawChildPolicies == null) {
  return DEFAULT_MODE;
}
...
```

zhangkun83:
@ejona86 PTAL

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -214,59 +216,82 @@ PolicySelection decideLoadBalancerProvider(
...
-      roundRobinDueToGrpclbDepMissing = false;
-
       List<LbConfig> lbConfigs = null;
       if (config != null) {
         List<Map<String, Object>> rawLbConfigs =
             ServiceConfigUtil.getLoadBalancingConfigsFromServiceConfig(config);
         lbConfigs = ServiceConfigUtil.unwrapLoadBalancingConfigList(rawLbConfigs);
       }
       if (lbConfigs != null && !lbConfigs.isEmpty()) {
+        // If there isn't a preselecbted, try to get the provider as configured
```
ejona86: Delete comment?

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -214,59 +216,82 @@ PolicySelection decideLoadBalancerProvider(
...
+          helper.getChannelLogger().log(ChannelLogLevel.ERROR, errorMsg);
+          logger.warning(errorMsg);
+        }
+        return new PolicySelection(
+            getProviderOrThrow(
+                "round_robin", "received balancer addresses but grpclb runtime is missing"),
+            backends, null);
+      }
+      roundRobinDueToGrpclbDepMissing = false;
+      if (addresses.size() == backends.size()) {
```
ejona86: I thought this would move to gRPC-LB (and probably be a warning).

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -214,59 +216,82 @@ PolicySelection decideLoadBalancerProvider(
...
+        // This is a special case where the existence of balancer address in the resolved address
+        // selects "grpclb" policy if the service config couldn't select a policy
+        return selectGrpclb(servers, backendAddrs, null);
       }
+      // No config nor balancer address. Use default.
       return new PolicySelection(
           getProviderOrThrow(defaultPolicy, "using default policy"), servers, null);
     }
+
+    private PolicySelection selectGrpclb(
```
ejona86: This looks like it should only be used in the `if (haveBalancerAddress) {` case.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -214,59 +216,82 @@ PolicySelection decideLoadBalancerProvider(
...
         LinkedHashSet<String> policiesTried = new LinkedHashSet<>();
         for (LbConfig lbConfig : lbConfigs) {
           String policy = lbConfig.getPolicyName();
-          LoadBalancerProvider provider = registry.getProvider(policy);
-          if (provider != null) {
-            if (!policiesTried.isEmpty()) {
-              helper.getChannelLogger().log(
-                  ChannelLogLevel.DEBUG,
-                  "{0} specified by Service Config are not available", policiesTried);
+          if (policy.equals(GRPCLB_POLICY_NAME)) {
```
ejona86: Get rid of this check here and instead do a check after `registry.getProvider(policy)`. The check will just swap out and use `servers` instead of `backendAddrs`.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -214,59 +216,82 @@ PolicySelection decideLoadBalancerProvider(
...
-          }
-          return new PolicySelection(
-              getProviderOrThrow(
-                  "round_robin", "received balancer addresses but grpclb runtime is missing"),
-              backendAddrs, null);
-        } else {
-          return new PolicySelection(grpclbProvider, servers, null);
-        }
-      }
-      roundRobinDueToGrpclbDepMissing = false;
```
ejona86: This line won't work quite the same way when moved. It looks like it can be removed though.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
zhangkun83: GRPCLB is handling it correctly by failing RPCs.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
zhangkun83: I think it still works. Keeping it.

zhangkun83:
@ejona86  PTAL

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
ejona86: We should use the fallback policy instead of failing RPCs. Go ahead and file a bug to handle this later.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -234,36 +268,15 @@ PolicySelection decideLoadBalancerProvider(
               getProviderOrThrow(
                   "round_robin", "received balancer addresses but grpclb runtime is missing"),
               backendAddrs, null);
-        } else {
-          return new PolicySelection(grpclbProvider, servers, null);
         }
+        return new PolicySelection(grpclbProvider, servers, null);
+      } else {
```
ejona86: It's confusing to have the else _and_ continue, since the if always returns. I'd suggest moving this out of the else and just put it after the if.

PR #5420, opened 2019-03-03
core: add internal transport attribute ATTR_CLIENT_EAG_ATTRS

zhangkun83:
This is needed for GRPCLB pick_first support, which needs to attach
tokens to headers, and the tokens are per server.  In pick_first, all
addresses are in a single Subchannel, thus the LoadBalancer needs to
know which backend is used for a new stream.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
@@ -24,13 +24,13 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Stopwatch;
 import com.google.common.base.Supplier;
 import io.grpc.Attributes;
-import io.grpc.Grpc;
 import io.grpc.InternalChannelz;
 import io.grpc.Metadata;
 import io.grpc.Status;
 import io.grpc.StatusException;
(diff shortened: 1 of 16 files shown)


zhangkun83:
@ejona86, I have moved the attribute to internal per our offline discussion.

netty/src/main/java/io/grpc/netty/NettyClientHandler.java
```diff
@@ -439,7 +442,7 @@ public boolean visit(Http2Stream stream) throws Http2Exception {
   @Override
   public void handleProtocolNegotiationCompleted(
       Attributes attributes, InternalChannelz.Security securityInfo) {
-    this.attributes = attributes;
+    this.attributes = this.attributes.toBuilder().setAll(attributes).build();
```
ejona86: I don't understand this change.

zhangkun83: This merges the new attributes from the protocol negotiator into `this.attributes`.
`this.attributes` includes the EAG attrs. `attributes` is from the protocol negotiator and doesn't include EAG attrs.

Issue #5422, opened 2019-03-04
Android lint InvalidPackage: javax.naming


bubenheimer:
### What version of gRPC are you using?
1.19.0

Sample project: https://github.com/bubenheimer/grpclint

Running `./gradlew :app:lintDebug` triggers errors:

    /Users/uli/.gradle/caches/modules-2/files-2.1/io.grpc/grpc-core/1.19.0/48b280ef2c8f42989c65bd61665926c212379660/grpc-core-1.19.0.jar: Error: Invalid package reference in library; not included in Android: javax.naming.directory. Referenced from io.grpc.internal.JndiResourceResolverFactory.JndiRecordFetcher. [InvalidPackage]
    /Users/uli/.gradle/caches/modules-2/files-2.1/io.grpc/grpc-core/1.19.0/48b280ef2c8f42989c65bd61665926c212379660/grpc-core-1.19.0.jar: Error: Invalid package reference in library; not included in Android: javax.naming. Referenced from io.grpc.internal.JndiResourceResolverFactory.JndiRecordFetcher. [InvalidPackage]

carl-mastrangelo:
@bubenheimer those files are loaded reflectively, and only if present.  There aren't any hard references from public API to those classes.  They are never loaded on Android, but usually loaded on server side.

bubenheimer:
Understood. This does not interfere with correct operation of the app. It does trigger lint errors, which I need to address.

I tried shuffling around the imports for the two Attributes classes in JndiRecordFetcher (swapping the import for one class for an import of the other), but that did not help. I still suspect the lint errors relate to having two classes present which are both named "Attributes", it's an unusual coincidence.

carl-mastrangelo:
@bubenheimer in #5424 I posted our proguard flags for building core, which would exclude the javax.naming packages.  Would those help you at all?  (I am not terribly familiar with how proguard fits into peoples build processes).

ejona86:
@carl-mastrangelo, I think the lint is separate from proguard.

dapengzhang0:
Proguard/dontwarn does not fix lint errors. To suppress the lint error, you may either add `lintOptions` as in https://github.com/grpc/grpc-java/blob/v1.19.0/examples/android/helloworld/app/build.gradle#L23 in your `build.gradle` or use lint.xml as in https://github.com/4thline/cling/issues/145#issuecomment-140642446 .

[...]

PR #5424, opened 2019-03-04
core,okhttp,testing: add proguard flags

carl-mastrangelo:
@ericgribkoff  had suggesting these flags files could live in OSS, and in the scope of #5422 it seems there would be some use of them.  

Currently these have no effect, but they serve as references we can point people to.  Also, these match the internal copies (in content and file path).


carl-mastrangelo:
Alright, I think this would simplify our import script slightly, so if you don't have any other concerns I would like to commit this.  In the future we can consolidate the proguard-rules.txt file with the ones here as you suggested.

PR #5426, opened 2019-03-05
all: Stop committing generated protobuf messages

ejona86:
This commit swaps to using a Sync task to place generated code in the
src/generated folder instead of the gradle-protobuf-plugin's
generatedFilesBaseDir. This provides much nicer results on failed
builds, and you will no longer see all the generated files deleted.

But at the same time the Sync task makes it easy to only copy the
grpc-generated code. This was not previously done because we were lazy
and using generatedFilesBaseDir, which made it difficult to treat the
services differently from the messages.

------

There's a lot of noise because of deletions. But I only modified build.gradle, xds/build.gradle, and .travis.yml.

CC @dapengzhang0, this will reduce the size of the xDS files.


dapengzhang0:
>There's a lot of noise because of deletions. But I only modified build.gradle, xds/build.gradle, and .travis.yml.

This will help: In `Files changed` tab, select `file filter...`, and filter out .java files.

PR #5427, opened 2019-03-05
netty, core: pass log-only channel logger into transport

carl-mastrangelo:


core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -258,9 +258,13 @@ private void startNewTransport() {
...
           .setUserAgent(userAgent)
           .setHttpConnectProxiedSocketAddress(proxiedAddr);
+    TransportLogger transportLogger = new TransportLogger();
+    // In case the transport logs in the constructor, use the subchannel logId
+    transportLogger.logId = getLogId();
     ConnectionClientTransport transport =
         new CallTracingTransport(
-            transportFactory.newClientTransport(address, options, channelLogger), callsTracer);
+            transportFactory.newClientTransport(address, options, transportLogger), callsTracer);
+    transportLogger.logId = transport.getLogId();
```
zhangkun83: I was wrong when saying new transports were logged to the subchannel's logger. Probably worth a log to `channelLogger` here to associate the subchannel and the transport.

carl-mastrangelo: Done.

PR #5430, opened 2019-03-05
netty: add a proxy handler

carl-mastrangelo:
New Pieces:

* Added an Http connect proxy.  
* Record the target proxy address in an attribute key (in the new handler)
* Added docs about new protocol negotiator style

Missing pieces:

* New handler is not wired up, ProxiedSocketAddress is not changed

Where this is going:

Each negotiator in ProtocolNegotiators will insert ClientHttpProxyHandler before WaitUntilActiveHandler.  This will make each negotiator capable of handling proxy addresses, (but not yet required to handle them).  This will remove the chaining of the ProxyProtocolNegotiator.  The motivation is that these new style negotiators expect to be have total knowledge of the handlers that will be added.  The handlers are composable, rather than the negotiators.

The new attributed is currently used for testing, but will eventually be represented in the Channelz data.  I don't expect this will have any use to applications, so we could remove it from the attributes, if desired.

### DIFF
--- a/core/src/main/java/io/grpc/Grpc.java
+++ b/core/src/main/java/io/grpc/Grpc.java
@@ -47,6 +47,14 @@ public final class Grpc {
       Attributes.Key.create("local-addr");
 
   /**
+   * Attribute key for the proxied remote address of a transport.
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1710")
+  @TransportAttr
+  public static final Attributes.Key<SocketAddress> TRANSPORT_ATTR_PROXIED_REMOTE_ADDR =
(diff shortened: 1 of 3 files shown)


ejona86:
We should discuss this offline.

ejona86:
I'm not quite sure about the history for this, but I think we decided it wasn't necessary: composing negotiators works okay. Or we reworked something somewhere else to make it okay (like what inserts the write queuing handler and such). I expect I wasn't bothered by this approach as much as I was worried about why it was necessary and preventing it from being necessary. (That is to say, if this is how the code was written to begin with, that would have been fine. But to _have_ to write it this was was worrying.)

PR #5432, opened 2019-03-05
netty: expose some of the reusable handlers in ProtocolNegotiators

carl-mastrangelo:


netty/src/main/java/io/grpc/netty/InternalProtocolNegotiators.java
```diff
@@ -69,4 +71,12 @@ public void close() {
     
     return new TlsNegotiator();
   }
+
+  public static ChannelHandler waitUntilActiveHandler(ChannelHandler next) {
```
creamsoup: some javadoc would be helpful

carl-mastrangelo: Added a dummy comment, but this is an accessor and not meant for normal usage.

PR #5438, opened 2019-03-06
grpclb: support "pick_first" child policy

zhangkun83:
The PICK_FIRST mode puts all backend addresses in a single Subchannel. There are a few points where it's different from the default ROUND_ROBIN mode:
1. PICK_FIRST doesn't eagerly connect to backends like ROUND_ROBIN does. Instead, it requests for connections when the Subchannel is picked.
2. PICK_FIRST adds tokens to the headers via a different code path (`TokenAttachingTracerFactory`) than ROUND_ROBIN
3. For simple implementation, when the mode is changed by service config when the LoadBalancer is working, we will shut down `GrpclbState` and starts a new one with the new mode. All connections will be closed during the transition. We don't expect this to happen in practice given the specific use case of PICK_FIRST.


grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
```diff
@@ -740,34 +823,45 @@ public boolean equals(Object other) {
...
 
-  private interface RoundRobinEntry {
+  @VisibleForTesting
+  interface RoundRobinEntry {
     PickResult picked(Metadata headers);
   }
 
   @VisibleForTesting
   static final class BackendEntry implements RoundRobinEntry {
+    final Subchannel subchannel;
```
ejona86: Why is this being stored now?

zhangkun83: Because it's always non-null in `BackendEntry`, and there are quite a few places that need to retrieve the subchannel from `BackendEntry`. `entry.subchannel` is easier to reason than `entry.result.getSubchannel()` that you won't get a `null`.

PR #5446, opened 2019-03-08
core: separate service config parsing and add NR.Helper method

carl-mastrangelo:


carl-mastrangelo:
This is part of the necessary change, limited in scope.  The next part will be moving the rest of the code that parses the config in `NameResolverListenerImpl.onAddresses` into the `ManagedChannelServiceConfig`.  Then the NR can pass the service config into the Managed channel and apply it directly.

core/src/main/java/io/grpc/NameResolver.java
```diff
@@ -242,5 +247,92 @@ void onAddresses(
...
+      public static <T> ConfigOrStatus<T> fromConfig(T config) {
+        return new ConfigOrStatus<>(config);
+      }
+
+      /**
+       * Returns a {@link ConfigOrStatus} for the failure to parse the config.
+       *
+       * @since 1.20.0
+       */
+      public static <T> ConfigOrStatus<T> fromStatus(Status status) {
```
zhangkun83: ditto `fromError()`

core/src/main/java/io/grpc/NameResolver.java
```diff
@@ -242,5 +247,92 @@ void onAddresses(
...
+      @Nullable
+      public T getConfig() {
+        return config;
+      }
+
+      /**
+       * @since 1.20.0
+       */
+      @Nullable
+      private Status getStatus() {
```
zhangkun83: ditto `getError()`

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -546,25 +546,17 @@ ClientStream newSubstream(ClientStreamTracer.Factory tracerFactory, Metadata new
...
-
-        @Override
-        public SynchronizationContext getSynchronizationContext() {
-          return syncContext;
-        }
-      };
+    this.retryEnabled = builder.retryEnabled && !builder.temporarilyDisableRetry;
+    this.nameResolverHelper =
+        new NrHelper(
+        builder.getDefaultPort(),
```
zhangkun83: more indent

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1666,4 +1658,63 @@ public void execute(Runnable command) {
...
+    }
+
+    @Override
+    public SynchronizationContext getSynchronizationContext() {
+      return syncCtx;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public ConfigOrStatus<ManagedChannelServiceConfig> parseServiceConfig(
```
zhangkun83: Please add a test for this method, to verify the success and the error cases.

carl-mastrangelo: Done.   I expect more tests will move over to calling this method in the next CLs too.

core/src/main/java/io/grpc/internal/ManagedChannelServiceConfig.java
```diff
@@ -0,0 +1,282 @@
...
+
+    // Try and do as much validation here before we swap out the existing configuration.  In case
+    // the input is invalid, we don't want to lose the existing configuration.
+
+    List<Map<String, ?>> methodConfigs =
+        ServiceConfigUtil.getMethodConfigFromServiceConfig(serviceConfig);
+
+    if (methodConfigs == null) {
+      // this is surprising, but possible.
+      logger.fine("no methods configs in service config");
```
zhangkun83: A fine-level log is hardly noticeable. If you want this to be more useful, I would pass the ChannelLogger to `fromServiceConfig()`, and log this information with `DEBUG`. It'd have a better chance of being noticed as it will show up when someone greps the channel logs.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -3376,6 +3381,59 @@ public void getAuthorityAfterShutdown() throws Exception {
...
+        proxyDetector,
+        syncCtx,
+        retryEnabled,
+        maxRetryAttemptsLimit,
+        maxHedgedAttemptsLimit);
+
+    ConfigOrError<ManagedChannelServiceConfig> coe =
+        nrh.parseServiceConfig(ImmutableMap.<String, Object>of());
+
+    assertThat(coe.getConfig()).isNotNull();
```
zhangkun83: Should be useful to have `(coe.getError()).isNull()`.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -3376,6 +3381,59 @@ public void getAuthorityAfterShutdown() throws Exception {
...
+        proxyDetector,
+        syncCtx,
+        retryEnabled,
+        maxRetryAttemptsLimit,
+        maxHedgedAttemptsLimit);
+
+    ConfigOrError<ManagedChannelServiceConfig> coe =
+        nrh.parseServiceConfig(ImmutableMap.<String, Object>of("methodConfig", "bogus"));
+
+    assertThat(coe.getError()).isNotNull();
```
zhangkun83: Should also be useful to have `(coe.getConfig()).isNull()`.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
carl-mastrangelo: This is impossible, because of ConfigOrError.  There can't be both.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
carl-mastrangelo: Same comment.  CoE is supposed to be one or other, never both.   it doesn't make sense to check for both here.

core/src/main/java/io/grpc/internal/ManagedChannelServiceConfig.java
carl-mastrangelo: Removed.

PR #5453, opened 2019-03-08
core: move all NameResolver.Listener impl to syncContext

### LINKED ISSUE #5450: Data race in NameResolve.Listener.onError
`NameResolve.Listener.onError` can be called concurrently in different threads, so the following code in `onError()` impl may have data race.

```
    haveBackends = false;
}
```

zhangkun83:
Resolves #5450


dapengzhang0:
Actually I wanted the `onError()` called from inside `onAddress()` to be re-entrant in the sense that it is executed prior to any other `onError()` after this `onAddress()`. I propose a different fix. Maybe let's discuss offline.

PR #5456, opened 2019-03-09
grpc_java_repositories: fix com_google_protobuf_javalite sha256.

Ubehebe:
`curl -sL https://github.com/google/protobuf/archive/384989534b2246d413dbcd750744faab2607b516.zip | shasum -a 256`
gives `79d102c61e2a479a0b7e5fc167bcfaa4832a0c6aad4a75fa7da0480564931bcc`.

### DIFF
--- a/repositories.bzl
+++ b/repositories.bzl
@@ -226,7 +226,7 @@ def com_google_protobuf_javalite():
     # java_lite_proto_library rules implicitly depend on @com_google_protobuf_javalite
     http_archive(
         name = "com_google_protobuf_javalite",
-        sha256 = "d8a2fed3708781196f92e1e7e7e713cf66804bd2944894401057214aff4f468f",
+        sha256 = "79d102c61e2a479a0b7e5fc167bcfaa4832a0c6aad4a75fa7da0480564931bcc",
         strip_prefix = "protobuf-384989534b2246d413dbcd750744faab2607b516",
         urls = ["https://github.com/google/protobuf/archive/384989534b2246d413dbcd750744faab2607b516.zip"],
     )


Ubehebe:
This appears to have been broken since eaca73473c272a0883f144d702144fbbc54ff355 (early February), which bumped the javalite version but suspiciously only changed the last digit of the sha. I'm not sure why this went unnoticed; `bazel query rdeps` shows 14 targets in this repo depend on it.

carl-mastrangelo:
@ejona86 PTAL

ejona86:
Hmmm... that means we don't actually have a reference to it anywhere. I added a trash java_lite_proto_library and built and it detected the mismatch. protobuf-lite does have a dependency on it, but only for android builds...

It seems there should be some sort of addition to prevent regressions in the future or it should be removed, but it's not obvious what form it should take. Whatever.

ejona86:
(FYI: a common way that I figure out the new SHA is to modify the old sha and rebuilt. The build will fail and tell me the sha of what it downloaded. I then copy-paste it into the build file. That's why the last digit changed; I had expected the build to fail.)

ejona86:
@Ubehebe, thank you!

Ubehebe:
@ejona86 no problem. FYI, the reason I noticed this was because I use [ibazel](https://github.com/bazelbuild/bazel-watcher) for local development. `ibazel build ...` was broken on this bad sha, even though nothing in my repo depends on the javalite artifact (and `bazel build ...` worked). So ibazel must be doing some extra analysis.

zhangkun83:
This needs to be backported to 1.19.

PR #5461, opened 2019-03-11
core fix 5450 (alternative approach)

### LINKED ISSUE #5450: Data race in NameResolve.Listener.onError

dapengzhang0:
Resolves #5450

### DIFF
--- a/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
+++ b/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
@@ -1339,8 +1339,6 @@ final class ManagedChannelImpl extends ManagedChannel implements
     @Override
     public void onError(final Status error) {
       checkArgument(!error.isOk(), "the error status must not be OK");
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1339,41 +1341,45 @@ public void onError(final Status error) {
       checkArgument(!error.isOk(), "the error status must not be OK");
       logger.log(Level.WARNING, "[{0}] Failed to resolve name. status={1}",
```
zhangkun83: I would move this line into the runnable to be consistent with the logging in `onAddresses()`.

dapengzhang0: Done.

PR #5463, opened 2019-03-11
core: channelBuilder.defaulServiceConfig() and lookUpServiceConfig()

dapengzhang0:
Note that the default config follows the [spec](https://github.com/grpc/proposal/blob/master/A21-service-config-error-handling.md#behavior-on-receiving-a-new-grpc-config) that

```
4. If the service config is invalid, the clients should reject the received
config in its entirety and do different things depending on its state and
configuration.
	1. If it previously had selected a valid service config or no service
	config, it should continue using that previous configuration from that
	config. An empty service config is valid.
	2. If the client has not previously selected a valid service config or no
	service config, for example a new client, then:
		1. It uses the default service config, if configured and valid. Note
		that an empty default service config is a valid service config.
		2. If it does not have a default service config, it should treat the
		resolution attempt as having failed (e.g., for a polling-based
		resolver, it should retry the query after appropriate backoff). In
		other words,the channel will remain in state TRANSIENT_FAILURE until a
		valid service config is received.
```

### DIFF
--- a/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
@@ -245,7 +245,7 @@ public abstract class ForwardingChannelBuilder<T extends ForwardingChannelBuilde
   }
 
   @Override
-  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig) throws BuilderException {
+  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig) {
     delegate().defaultServiceConfig(serviceConfig);
     return thisT();
   }
--- a/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
 
 /**
  * A {@link ManagedChannelBuilder} that delegates all its builder method to another builder by
@@ -244,7 +245,7 @@ public abstract class ForwardingChannelBuilder<T extends ForwardingChannelBuilde
   }
 
   @Override
-  public T defaultServiceConfig(Map<String, ?> serviceConfig) throws BuilderException {
+  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig) throws BuilderException {
     delegate().defaultServiceConfig(serviceConfig);
     return thisT();
   }
--- a/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
@@ -251,8 +251,8 @@ public abstract class ForwardingChannelBuilder<T extends ForwardingChannelBuilde
   }
 
   @Override
-  public T lookupServiceConfig(boolean enable) {
-    delegate().lookupServiceConfig(enable);
+  public T lookUpServiceConfig(boolean enable) {
+    delegate().lookUpServiceConfig(enable);
     return thisT();
   }
 
--- a/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
@@ -18,6 +18,7 @@ package io.grpc;
 
 import com.google.common.base.MoreObjects;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 
@@ -239,6 +240,18 @@ public abstract class ForwardingChannelBuilder<T extends ForwardingChannelBuilde
   @Override
   public T proxyDetector(ProxyDetector proxyDetector) {
     delegate().proxyDetector(proxyDetector);
+    return thisT();
+  }
+
+  @Override
+  public T defaultServiceConfig(Map<String, Object> serviceConfig) throws BuilderException {
+    delegate().defaultServiceConfig(serviceConfig);
+    return thisT();
+  }
+
+  @Override
+  public T discardServiceConfigFromNameResolver() {
+    delegate().discardServiceConfigFromNameResolver();
     return thisT();
   }
 
--- a/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
@@ -244,7 +244,7 @@ public abstract class ForwardingChannelBuilder<T extends ForwardingChannelBuilde
   }
 
   @Override
-  public T defaultServiceConfig(Map<String, Object> serviceConfig) throws BuilderException {
+  public T defaultServiceConfig(Map<String, ?> serviceConfig) throws BuilderException {
     delegate().defaultServiceConfig(serviceConfig);
     return thisT();
   }
--- a/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ForwardingChannelBuilder.java
@@ -251,8 +251,8 @@ public abstract class ForwardingChannelBuilder<T extends ForwardingChannelBuilde
   }
 
   @Override
-  public T discardServiceConfigFromNameResolver() {
-    delegate().discardServiceConfigFromNameResolver();
+  public T lookupServiceConfig(boolean enable) {
+    delegate().lookupServiceConfig(enable);
     return thisT();
   }
 
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -540,9 +540,10 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   }
 
   /**
-   * Provides a service config to the channel.
-   *
-   * @param serviceConfig A nested map representing a Json object in the most natural way:
+   * Provides a service config to the channel. The channel will use the default service config when
+   * the name resolver provides no service config or if the channel disables lookup service config
+   * from name resolver. The argument {@code serviceConfig} is a nested map representing a Json
+   * object in the most natural way:
    *
    *        <table border="1">
    *          <tr>
@@ -568,9 +569,8 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
    *          </tr>
    *        </table>
    *
-   *        If null is passed, then there will be no default service config.
+   * <p>If null is passed, then there will be no default service config.
    *
-   * @throws UnsupportedOperationException If not implemented.
    * @throws IllegalArgumentException When the given serviceConfig is invalid or the current version
    *         of grpc library can not parse it gracefully. The state of the builder is unchanged if
    *         an exception is thrown.
@@ -583,7 +583,8 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   }
 
   /**
-   * Lets the channel look up or not look up service config from the name resolver.
+   * Lets the channel look up or not look up service config from the name resolver. The look-up is
+   * enabled by default.
    *
    * @return this
    * @since 1.20.0
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -542,8 +542,8 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   /**
    * Provides a service config to the channel. The channel will use the default service config when
    * the name resolver provides no service config or if the channel disables lookup service config
-   * from name resolver. The argument {@code serviceConfig} is a nested map representing a Json
-   * object in the most natural way:
+   * from name resolver (see {@link #lookUpServiceConfig(boolean)}). The argument
+   * {@code serviceConfig} is a nested map representing a Json object in the most natural way:
    *
    *        <table border="1">
    *          <tr>
@@ -583,8 +583,7 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   }
 
   /**
-   * Lets the channel look up or not look up service config from the name resolver. The look-up is
-   * enabled by default.
+   * Enables or disables service config look-up from the naming system. Enabled by default.
    *
    * @return this
    * @since 1.20.0
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -571,14 +571,14 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
    *        If null is passed, then there will be no default service config.
    *
    * @throws UnsupportedOperationException If not implemented.
-   * @throws BuilderException When the given serviceConfig is invalid or the current version of grpc
-   *         library can not parse it gracefully. The state of the builder is unchanged if an
-   *         exception is thrown.
+   * @throws IllegalArgumentException When the given serviceConfig is invalid or the current version
+   *         of grpc library can not parse it gracefully. The state of the builder is unchanged if
+   *         an exception is thrown.
    * @return this
    * @since 1.20.0
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/5189")
-  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig) throws BuilderException {
+  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig) {
     throw new UnsupportedOperationException();
   }
 
@@ -591,17 +591,6 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/5189")
   public T lookupServiceConfig(boolean enable) {
     throw new UnsupportedOperationException();
-  }
-
-  /**
-   * Checked exception when there is an error in the configuration of the builder.
-   */
-  public static class BuilderException extends Exception {
-    private static final long serialVersionUID = 1347259L;
-
-    public BuilderException(Exception cause) {
-      super(cause);
-    }
   }
 
   /**
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -583,13 +583,13 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   }
 
   /**
-   * Lets the channel ignore any service config from the name resolver.
+   * Lets the channel look up or not look up service config from the name resolver.
    *
    * @return this
    * @since 1.20.0
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/5189")
-  public T discardServiceConfigFromNameResolver() {
+  public T lookupServiceConfig(boolean enable) {
     throw new UnsupportedOperationException();
   }
 
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -18,6 +18,7 @@ package io.grpc;
 
 import com.google.common.base.Preconditions;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 
@@ -535,6 +536,69 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/5113")
   public T proxyDetector(ProxyDetector proxyDetector) {
     throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Provides a service config to the channel.
+   *
+   * @param serviceConfig A nested map representing a Json object in the most natural way:
+   *
+   *        <table border="1">
+   *          <tr>
+   *            <td>Json entry</td><td>Java Type</td>
+   *          </tr>
+   *          <tr>
+   *            <td>object</td><td>{@link Map}</td>
+   *          </tr>
+   *          <tr>
+   *            <td>array</td><td>{@link List}</td>
+   *          </tr>
+   *          <tr>
+   *            <td>string</td><td>{@link String}</td>
+   *          </tr>
+   *          <tr>
+   *            <td>number</td><td>{@link Double}</td>
+   *          </tr>
+   *          <tr>
+   *            <td>boolean</td><td>{@link Boolean}</td>
+   *          </tr>
+   *          <tr>
+   *            <td>null</td><td>{@code null}</td>
+   *          </tr>
+   *        </table>
+   *
+   * @throws UnsupportedOperationException If not implemented.
+   * @throws BuilderException When the given serviceConfig is invalid or the current version of grpc
+   *         library can not parse it gracefully. The state of the builder is unchanged if an
+   *         exception is thrown.
+   * @return this
+   * @since 1.20.0
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/5189")
+  public T defaultServiceConfig(Map<String, Object> serviceConfig) throws BuilderException {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Lets the channel ignore any service config from the name resolver.
+   *
+   * @return this
+   * @since 1.20.0
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/5189")
+  public T discardServiceConfigFromNameResolver() {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Checked exception when there is an error in the configuration of the builder.
+   */
+  public static class BuilderException extends Exception {
+    private static final long serialVersionUID = 1347259L;
+
+    public BuilderException(Exception cause) {
+      super(cause);
+    }
   }
 
   /**
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -590,7 +590,7 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
    * @since 1.20.0
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/5189")
-  public T lookupServiceConfig(boolean enable) {
+  public T lookUpServiceConfig(boolean enable) {
     throw new UnsupportedOperationException();
   }
 
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
 
 /**
  * A builder for {@link ManagedChannel} instances.
@@ -567,6 +568,8 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
    *          </tr>
    *        </table>
    *
+   *        If null is passed, then there will be no default service config.
+   *
    * @throws UnsupportedOperationException If not implemented.
    * @throws BuilderException When the given serviceConfig is invalid or the current version of grpc
    *         library can not parse it gracefully. The state of the builder is unchanged if an
@@ -575,7 +578,7 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
    * @since 1.20.0
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/5189")
-  public T defaultServiceConfig(Map<String, ?> serviceConfig) throws BuilderException {
+  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig) throws BuilderException {
     throw new UnsupportedOperationException();
   }
 
--- a/core/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/core/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -575,7 +575,7 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
    * @since 1.20.0
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/5189")
-  public T defaultServiceConfig(Map<String, Object> serviceConfig) throws BuilderException {
+  public T defaultServiceConfig(Map<String, ?> serviceConfig) throws BuilderException {
     throw new UnsupportedOperationException();
   }
 
--- a/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
@@ -386,14 +386,13 @@ public abstract class AbstractManagedChannelImplBuilder
   }
 
   @Override
-  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig)
-      throws BuilderException {
+  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig) {
     defaultServiceConfig = parseMap(serviceConfig);
     return thisT();
   }
 
   @Nullable
-  private static Map<String, ?> parseMap(@Nullable Map<?, ?> map) throws BuilderException {
+  private static Map<String, ?> parseMap(@Nullable Map<?, ?> map) {
     if (map == null) {
       return null;
     }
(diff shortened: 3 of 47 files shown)


core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -537,6 +539,71 @@ public T proxyDetector(ProxyDetector proxyDetector) {
     throw new UnsupportedOperationException();
   }
 
+  /**
+   * Provides a service config to the channel.
```
zhangkun83: Also document the cases this default config will be used. My understanding of the [spec](https://github.com/grpc/proposal/blob/master/A21-service-config-error-handling.md) is the default config is used only if the Channel has not seen any config from NameResolver, including the case config lookup is disabled.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1289,24 +1307,54 @@ public void run() {
...
+            channelLogger.log(ChannelLogLevel.INFO, "Service config discarded by channel settings");
+            effectiveConfig = Attributes.newBuilder()
+                .setAll(config)
+                .set(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG, lastServiceConfig)
+                .build();
+          } else {
+            Map<String, ?> effectiveServiceConfig =
+                config.get(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG);
+            if (effectiveServiceConfig != null
+                && effectiveServiceConfig.containsKey(DnsNameResolver.SERVICE_CONFIG_ERROR)) {
```
zhangkun83: @carl-mastrangelo is implementing the error handling in the proper way. Whatever hack you put here for error-handling will be thrown away.  The code as of now will just ignore the update if the config parsing failed, or the resolver didn't return a config. You may want to just keep it that way.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -379,6 +385,74 @@ public T proxyDetector(@Nullable ProxyDetector proxyDetector) {
...
+  @Nullable
+  private static Map<String, ?> parseMap(@Nullable Map<?, ?> map) throws BuilderException {
+    if (map == null) {
+      return null;
+    }
+    // Not using ImmutableMap.Builder because of extra guava dependency for Android.
+    Map<String, Object> parsedMap = new LinkedHashMap<>();
+    for (Map.Entry<?, ?> entry : map.entrySet()) {
+      if (!(entry.getKey() instanceof String)) {
+        throw new BuilderException(null); // TODO(zdapeng): cause
```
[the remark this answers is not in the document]
dapengzhang0: ClassCastException will be wrapped. Let's wait for the decision on what Exception class should be thrown.

core/src/main/java/io/grpc/ForwardingChannelBuilder.java
```diff
@@ -242,6 +244,18 @@ public T proxyDetector(ProxyDetector proxyDetector) {
...
   }
 
+  @Override
+  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig) throws BuilderException {
+    delegate().defaultServiceConfig(serviceConfig);
+    return thisT();
+  }
+
+  @Override
+  public T discardServiceConfigFromNameResolver() {
```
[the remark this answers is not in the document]
dapengzhang0: I'm OK with `lookupServiceConfig(boolean)` as suggested by @zhangkun83 , let me know if any objection.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -291,15 +292,19 @@ public void run() {
...
-            if (serviceConfig != null) {
+              // reject service configs in its entirety
+              serviceConfig = Collections.singletonMap(SERVICE_CONFIG_ERROR, (Object) e);
               break;
             }
           }
-        } catch (RuntimeException e) {
+        } catch (IOException | RuntimeException e) {
           logger.log(Level.WARNING, "Can't parse service Configs", e);
+          serviceConfig = Collections.singletonMap(SERVICE_CONFIG_ERROR, (Object) e);
```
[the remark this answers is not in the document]
dapengzhang0: I need to distinguish empty config (valid) and error config. In error config case, will try to use the default config; and if the default config is missing, then go to transient failure. Are you also modifying the exception handling at the original line#294?

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -405,27 +410,24 @@ static ResolutionResults resolveAll(
...
-          }
-          List<?> listChoices = (List<?>) rawChoices;
-          for (Object obj : listChoices) {
-            if (!(obj instanceof Map)) {
-              throw new IOException("wrong element type " + rawChoices);
-            }
+
+        Object rawChoices = JsonParser.parse(txtRecord.substring(SERVICE_CONFIG_PREFIX.length()));
+        if (!(rawChoices instanceof List)) {
+          throw new IOException("wrong type " + rawChoices);
```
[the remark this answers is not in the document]
dapengzhang0: I'm trying to minimize the change. So I would leave it unchanged in this PR.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: So can I just assume the attributes passed here is with either no config or a valid config?

dapengzhang0:
I reverted service config resolution error handling part,  and for now assume receiving null config is always a non-error case.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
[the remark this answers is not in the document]
dapengzhang0: reverted.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: Reverted the hack.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -537,6 +539,60 @@ public T proxyDetector(ProxyDetector proxyDetector) {
...
+   *
+   * @param serviceConfig A nested map representing a Json object in the most natural way:
```
ejona86: Don't include the table in the `@param`. Move that up to the general description. `@param` should be short, like 2 sentences max, and ideally just a sentence fragment.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -379,6 +385,74 @@ public T proxyDetector(@Nullable ProxyDetector proxyDetector) {
...
   }
 
+  @Override
+  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig) {
+    defaultServiceConfig = parseMap(serviceConfig);
+    return thisT();
+  }
+
+  @Nullable
+  private static Map<String, ?> parseMap(@Nullable Map<?, ?> map) {
```
ejona86: This doesn't look to be parsing anything. It seems to just be making an immutable copy. Name it as such?

ejona86: Note: if you instead had `parseObject`, this would be much shorter as it would have less duplication between map and list. But it seems this will be deleted almost immediately, so it is fine as-is.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1289,24 +1310,51 @@ public void run() {
...
+                  "Service config from name resolver discarded by channel settings");
+            }
+            if (defaultServiceConfig != null) {
+              channelLogger.log(ChannelLogLevel.INFO, "Using default service config");
+            } else if (lastServiceConfig != null && serviceConfig == null) {
+              // no default config, no current service config, last config is non-trivial
+              channelLogger.log(ChannelLogLevel.INFO, "Service config changed to null");
+            }
+
+            if (!Objects.equal(defaultServiceConfig, serviceConfig)) {
```
ejona86: Don't bother with this condition. The equals check could be pretty expensive itself and there's virtually no chance the configs match (unless they are empty, I guess). If you want, you could check to see if they are both null (which is basically equivalent to reference equality, but more clear).

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1289,24 +1310,51 @@ public void run() {
...
+              effectiveConfig = Attributes.newBuilder()
```
ejona86: `config.toBuilder()` is more efficient than `newBuilder().setAll(config)` and more clear.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1289,24 +1310,51 @@ public void run() {
...
+                  .setAll(config)
+                  .set(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG, defaultServiceConfig)
+                  .build();
+            }
+            lastServiceConfig = defaultServiceConfig;
+          } else {
+            if (!serviceConfig.equals(lastServiceConfig)) {
```
ejona86: Just use reference equality.

core/src/main/java/io/grpc/ForwardingChannelBuilder.java
```diff
@@ -242,6 +244,18 @@ public T proxyDetector(ProxyDetector proxyDetector) {
     return thisT();
   }
 
+  @Override
+  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig) {
```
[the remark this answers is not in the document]
dapengzhang0: The  problem of `Map<String, Object>` might be: If the user has a `Map<String, Map>` object at hand, they have to cast it to `Map<String, Object>` first and then pass it to the API. Actually I'm not too strong about it. @carl-mastrangelo you want to debate about it?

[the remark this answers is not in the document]
ejona86: I don't disagree, but given the service config options, what is the likelihood that they set values that are the same type? It seems they would create the object just for passing to this call, so it would also be trivial for them to use Object instead of Map<String, Object> as the value.

I could accept having it, it just seems like it would be more confusing to users without much benefit.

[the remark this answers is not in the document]
carl-mastrangelo: I found that `Map<String, ?>` involved less internal casting, and was easier to pass in.  It also means that it isn't possible to write to the map since no type matches the wildcard, implying it is immutable.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
ejona86: It looks this is preserving existing behavior. It can stay as-is. (Although in the future I expect it will be changed.)

core/src/main/java/io/grpc/ManagedChannelBuilder.java
dapengzhang0: Done.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
dapengzhang0: Done.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -537,6 +539,60 @@ public T proxyDetector(ProxyDetector proxyDetector) {
...
+   *            <td>boolean</td><td>{@link Boolean}</td>
+   *          </tr>
+   *          <tr>
+   *            <td>null</td><td>{@code null}</td>
+   *          </tr>
+   *        </table>
+   *
+   *        If null is passed, then there will be no default service config.
+   *
+   * @throws UnsupportedOperationException If not implemented.
```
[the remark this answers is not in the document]
dapengzhang0: Done.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -537,6 +539,60 @@ public T proxyDetector(ProxyDetector proxyDetector) {
...
+   * @return this
+   * @since 1.20.0
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/5189")
+  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig) {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Lets the channel look up or not look up service config from the name resolver.
```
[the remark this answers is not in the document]
dapengzhang0: Done.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
dapengzhang0: Renamed to `checkMapEntryTypes()`

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: Done.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: Done.

core/src/main/java/io/grpc/ForwardingChannelBuilder.java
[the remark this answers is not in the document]
ejona86: How could `Map<String, ?>` involve less internal casting? Can you give an example?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -237,9 +237,17 @@ public void uncaughtException(Thread t, Throwable e) {
...
+  // Must be mutated and read from constructor or syncContext
+  // TODO: check this value when error in service config resolution
   @Nullable
   private Map<String, ?> lastServiceConfig; // used for channel tracing when value changed
+  @Nullable
+  private Map<String, ?> defaultServiceConfig;
+  // Must be mutated and read from constructor or syncContext
+  // See service config error handling spec for reference.
+  // TODO: check this value when error in service config resolution
+  private boolean selectedServiceConfigOrNoServiceConfig;
```
zhangkun83: A more comprehensible name for this variable may be "waitingForServiceConfig" which defaults to true.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1289,24 +1308,50 @@ public void run() {
...
           }
-          final Map<String, ?> serviceConfig =
+
+          // Assuming no error in config resolution for now.
+          Map<String, ?> serviceConfig =
               config.get(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG);
-          if (serviceConfig != null && !serviceConfig.equals(lastServiceConfig)) {
-            channelLogger.log(ChannelLogLevel.INFO, "Service config changed");
+          Attributes effectiveConfig = config;
+          if (!lookUpServiceConfig || serviceConfig == null) {
```
zhangkun83: Maybe simplify the conditions and make it flow better:

```java
Map<String, ?> effectiveConfig = null;

// Try to use config if returned from name resolver and allowed
if (serviceConfig != null) {
  if (lookUpServiceConfig) {
    effectiveConfig = serviceConfig;
  } else {
    channelLogger.log(ChannelLogLevel.INFO, "Service config from name resolver discarded by channel settings");
  }
}

// Otherwise, try to use the default config if available
if (effectiveConfig == null && defaultServiceConfig != null) {
  channelLogger.log(ChannelLogLevel.INFO, "Using default service config");
  effectiveConfig = defaultServiceConfig;
}

if (effectiveConfig != lastServiceConfig) {
  channelLogger.log(ChannelLogLevel.INFO,
      "Service config changed" + (effectiveConfig == null ? " to null" : ""));
  lastServiceConfig = effectiveConfig;
}

config = config.toBuilder()
    .set(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG, effectiveConfig)
    .build();

waitingForServiceConfig = false;

```

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1289,24 +1308,50 @@ public void run() {
...
+            }
+            if (defaultServiceConfig != null) {
+              channelLogger.log(ChannelLogLevel.INFO, "Using default service config");
+            } else if (lastServiceConfig != null && serviceConfig == null) {
+              // no default config, no current service config, last config is non-trivial
+              channelLogger.log(ChannelLogLevel.INFO, "Service config changed to null");
+            }
+
+            if (!(defaultServiceConfig == null && serviceConfig == null)) {
+              effectiveConfig = config.toBuilder()
```
zhangkun83: Even service config is already `null` in the attributes, it would make the implementation simpler to always update the attributes with the selected config (possibly null), because performance/allocation is not a concern for this code path. Please see my suggestion above.

dapengzhang0: That will break an existing test (Attributes.EMPTY is not the same as attributes with a key and a null value), I might need to change the test.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -537,6 +539,61 @@ public T proxyDetector(ProxyDetector proxyDetector) {
     throw new UnsupportedOperationException();
   }
 
+  /**
+   * Provides a service config to the channel. The channel will use the default service config when
+   * the name resolver provides no service config or if the channel disables lookup service config
+   * from name resolver. The argument {@code serviceConfig} is a nested map representing a Json
```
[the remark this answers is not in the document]
dapengzhang0: Done.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -537,6 +539,61 @@ public T proxyDetector(ProxyDetector proxyDetector) {
...
+   * @return this
+   * @since 1.20.0
+   */
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/5189")
+  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig) {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Lets the channel look up or not look up service config from the name resolver. The look-up is
```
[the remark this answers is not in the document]
dapengzhang0: Done.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
zhangkun83: If it's just one test, please go ahead and update that test.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -379,6 +385,74 @@ public T proxyDetector(@Nullable ProxyDetector proxyDetector) {
     return thisT();
   }
 
+  @Override
+  public T defaultServiceConfig(@Nullable Map<String, ?> serviceConfig) {
+    defaultServiceConfig = checkMapEntryTypes(serviceConfig);
```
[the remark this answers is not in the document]
dapengzhang0: Yes, this will be thrown away. I'm adding a TODO.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -237,9 +237,17 @@ public void uncaughtException(Thread t, Throwable e) {
   // Must be mutated and read from syncContext
   @CheckForNull
   private Boolean haveBackends; // a flag for doing channel tracing when flipped
-  // Must be mutated and read from syncContext
+  // Must be mutated and read from constructor or syncContext
+  // TODO: check this value when error in service config resolution
   @Nullable
   private Map<String, ?> lastServiceConfig; // used for channel tracing when value changed
+  @Nullable
+  private Map<String, ?> defaultServiceConfig;
```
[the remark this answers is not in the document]
dapengzhang0: Done.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: Done.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: Done.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
dapengzhang0: There are 6 tests, each will need introduce an ArgumentCapture, so added a check `if (effectiveServiceConfig != serviceConfig)` instead.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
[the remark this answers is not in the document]
zhangkun83: The `parseServiceConfig()` checks the config against the schema, which is more useful than just checking whether the map is a valid JSON. The missing part is parsing LoadBalancingConfig, but the immediate user of the default config is mobile and they presumably don't need load-balancing. Why check in the throw-away code that does less?

core/src/main/java/io/grpc/ForwardingChannelBuilder.java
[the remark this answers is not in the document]
carl-mastrangelo: Here is a similar case with `List<Object>`:  https://github.com/grpc/grpc-java/blob/v1.19.x/core/src/main/java/io/grpc/internal/ServiceConfigUtil.java#L501

It's not always going to happen that way, but still it was nice to avoid that.  

One other thing, when passing in this type, it's nice to pass values into Map<String, ?>, especially in unit tests where the type comes from an inlined value:

```java
defaultServiceConfig(ImmutableMap.of("methodConfig", myTestMethodConfig));
```

It involves a cast to the `Map<String, Object>`.  I think we should try it as `Map<String, ?>` and consider if it should be changed in API review.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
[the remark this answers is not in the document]
carl-mastrangelo: I think the parsing is still going to be happening, even with my code checked in.  Service Config is useable by more than grpc-java's code.  It supposed to be for application code to use.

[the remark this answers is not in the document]
dapengzhang0: > Why check in the throw-away code that does less

This code is extremely easy to throw-away, no need to resolve merge conflict. There are also test cases, which don't need to be thrown away when the code is replaced.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -3433,6 +3433,215 @@ public void nameResolverHelper_badConfigFails() {
     assertThat(coe.getError().getCause()).isInstanceOf(ClassCastException.class);
   }
 
+  @Test
+  public void disableNameResolverConfig_noDefaultConfig() throws Exception {
```
zhangkun83: `disableServiceConfigLookUp_`

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -3433,6 +3433,215 @@ public void nameResolverHelper_badConfigFails() {
...
+      assertThat(attributesCaptor.getValue().get(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG))
+          .isNull();
+      verify(mockLoadBalancer, never()).handleNameResolutionError(any(Status.class));
+    } finally {
+      LoadBalancerRegistry.getDefaultRegistry().deregister(mockLoadBalancerProvider);
+    }
+  }
+
+  @Test
+  public void disableNameResolverConfig_withDefaultConfig() throws Exception {
```
zhangkun83: ditto

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -3433,6 +3433,215 @@ public void nameResolverHelper_badConfigFails() {
...
+      assertThat(attributesCaptor.getValue().get(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG))
+          .isEqualTo(defaultServiceConfig);
+      verify(mockLoadBalancer, never()).handleNameResolutionError(any(Status.class));
+    } finally {
+      LoadBalancerRegistry.getDefaultRegistry().deregister(mockLoadBalancerProvider);
+    }
+  }
+
+  @Test
+  public void noDiscardNameResolverConfig_noDefaultConfig() throws Exception {
```
zhangkun83: `enableServiceConfigLookUp_`

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -3433,6 +3433,215 @@ public void nameResolverHelper_badConfigFails() {
...
+      assertThat(attributesCaptor.getValue().get(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG))
+          .isEqualTo(serviceConfig);
+      verify(mockLoadBalancer, never()).handleNameResolutionError(any(Status.class));
+    } finally {
+      LoadBalancerRegistry.getDefaultRegistry().deregister(mockLoadBalancerProvider);
+    }
+  }
+
+  @Test
+  public void noDiscardNameResolverConfig_withDefaultConfig() throws Exception {
```
zhangkun83: ditto

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -3433,6 +3433,215 @@ public void nameResolverHelper_badConfigFails() {
...
+          eq(ImmutableList.of(addressGroup)),
+          attributesCaptor.capture());
+      assertThat(attributesCaptor.getValue().get(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG))
+          .isEqualTo(serviceConfig);
+      verify(mockLoadBalancer, never()).handleNameResolutionError(any(Status.class));
+    } finally {
+      LoadBalancerRegistry.getDefaultRegistry().deregister(mockLoadBalancerProvider);
+    }
+  }
+
```
zhangkun83: There should also be `enableServiceConfigLookUp_resolverReturnsNoConfig_withDefaultConfig` and `enableServiceConfigLookUp_resolverReturnsNoConfig_noDefaultConfig`

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1311,58 +1319,51 @@ public void run() {
...
-          Attributes effectiveAttrs = attrs;
-          if (effectiveServiceConfig != serviceConfig) {
-            effectiveAttrs = attrs.toBuilder()
-                .set(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG, effectiveServiceConfig)
-                .build();
-          }
-
-          waitingForServiceConfig = false;
+          } else {
+            // Try to use config if returned from name resolver and allowed
```
zhangkun83: "and allowed" should be removed now.

dapengzhang0: Done.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -537,6 +539,60 @@ public T proxyDetector(ProxyDetector proxyDetector) {
...
+   * the name resolver provides no service config or if the channel disables lookup service config
+   * from name resolver (see {@link #lookUpServiceConfig(boolean)}). The argument
+   * {@code serviceConfig} is a nested map representing a Json object in the most natural way:
+   *
+   *        <table border="1">
+   *          <tr>
+   *            <td>Json entry</td><td>Java Type</td>
+   *          </tr>
+   *          <tr>
+   *            <td>object</td><td>{@link Map}</td>
```
carl-mastrangelo: `Map<String, ?>`

core/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -537,6 +539,60 @@ public T proxyDetector(ProxyDetector proxyDetector) {
...
+   *          </tr>
+   *          <tr>
+   *            <td>array</td><td>{@link List}</td>
```
carl-mastrangelo: `List<?>`

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -379,6 +385,75 @@ public T proxyDetector(@Nullable ProxyDetector proxyDetector) {
...
+
+  @Nullable
+  private static Map<String, ?> checkMapEntryTypes(@Nullable Map<?, ?> map) {
+    if (map == null) {
+      return null;
+    }
+    // Not using ImmutableMap.Builder because of extra guava dependency for Android.
+    Map<String, Object> parsedMap = new LinkedHashMap<>();
+    for (Map.Entry<?, ?> entry : map.entrySet()) {
+      if (!(entry.getKey() instanceof String)) {
```
carl-mastrangelo: This can be a `checkArgument`

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -379,6 +385,75 @@ public T proxyDetector(@Nullable ProxyDetector proxyDetector) {
...
+      } else if (value instanceof List) {
+        parsedMap.put(key, checkListEnttryTypes((List<?>) value));
+      } else if (value instanceof String) {
+        parsedMap.put(key, value);
+      } else if (value instanceof Double) {
+        parsedMap.put(key, value);
+      } else if (value instanceof Boolean) {
+        parsedMap.put(key, value);
+      } else {
+        throw new IllegalArgumentException();
```
carl-mastrangelo: Include the type name and key

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -379,6 +385,75 @@ public T proxyDetector(@Nullable ProxyDetector proxyDetector) {
...
+      }
+
+      String key = (String) entry.getKey();
+      Object value = entry.getValue();
+      if (value == null) {
+        parsedMap.put(key, null);
+      } else if (value instanceof Map) {
+        parsedMap.put(key, checkMapEntryTypes((Map<?, ?>) value));
+      } else if (value instanceof List) {
+        parsedMap.put(key, checkListEnttryTypes((List<?>) value));
```
carl-mastrangelo: checkListEnttryTypes is misspelled

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -379,6 +385,75 @@ public T proxyDetector(@Nullable ProxyDetector proxyDetector) {
...
+    return thisT();
+  }
+
+  @Nullable
+  private static Map<String, ?> checkMapEntryTypes(@Nullable Map<?, ?> map) {
+    if (map == null) {
+      return null;
+    }
+    // Not using ImmutableMap.Builder because of extra guava dependency for Android.
+    Map<String, Object> parsedMap = new LinkedHashMap<>();
```
carl-mastrangelo: `new LinkedHashMap<>(map.size());`

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -237,9 +237,17 @@ public void uncaughtException(Thread t, Throwable e) {
   // Must be mutated and read from syncContext
   @CheckForNull
   private Boolean haveBackends; // a flag for doing channel tracing when flipped
-  // Must be mutated and read from syncContext
+  // Must be mutated and read from constructor or syncContext
+  // TODO: check this value when error in service config resolution
```
carl-mastrangelo: can you add my name for this todo?

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
dapengzhang0: The default load factor is 0.75, it will be rehashed.

core/src/main/java/io/grpc/ManagedChannelBuilder.java
dapengzhang0: Generic types, including wildcards sometimes are very inconvenient to use. I would let the users be flexible to choose whatever generic type parameter that's most convenient for them, (either the users want to minimize the times to cast and suppress unchecked, or want to minimize errorprone at compile time), as long as it follows the table as a whole.

dapengzhang0:
@carl-mastrangelo PTALA, hope this PR will not introduce too many merge conflicts for yours.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -379,6 +385,78 @@ public T proxyDetector(@Nullable ProxyDetector proxyDetector) {
...
+    for (Map.Entry<?, ?> entry : map.entrySet()) {
+      checkArgument(
+          entry.getKey() instanceof String,
+          "The key of the entry '" + entry + "' is not of String type");
```
carl-mastrangelo:
```java
checkArgument(entry.getKey() instanceof String, "The key of the entry '%s' is not of String type", entry);
```

dapengzhang0:
Thank you @zhangkun83 @carl-mastrangelo @ejona86 for the review!

PR #5464, opened 2019-03-11
all: fix lint errors

creamsoup:


core/src/main/java/io/grpc/NameResolver.java
```diff
@@ -308,8 +304,6 @@ private ConfigOrError(Status status) {
 
       /**
        * Returns config if exists, otherwise null.
-       *
-       * @since 1.20.0
```
ejona86: Hmm... This looks differently than the import. I think there was already a follow-up PR to clean them up (by adding a description). So I guess the changes to this file can just be dropped.

ejona86: Sounds fine either way.

PR #5473, opened 2019-03-13
alts: add ComputeEngineChannelBuilder

jiangtaoli2016:
Add ComputeEngineChannelBuilder. This is similar to GoogleDefaultChannelBuilder, except that ComputeEngineChannelBuilder only runs on GCE and only fetches call credentials from metadata server.

### DIFF
--- a/alts/src/main/java/io/grpc/alts/CallCredentialsInterceptor.java
+++ b/alts/src/main/java/io/grpc/alts/CallCredentialsInterceptor.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2019 The gRPC Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


jiangtaoli2016:
Hold on review. Let me fix presubmit error first.

ejona86:
@jiangtaoli2016, those don't look like failures with your code. Try to rebuild them. Although the linux_artifacts looks broken for an unrelated reason; rebuilding won't fix it (but you can ignore it for now).

jiangtaoli2016:
Thanks @ejona86 for quick review. 
@apolcyn Could you start using this new channel builder in your staging test?

apolcyn:
Thanks for the quick turnaround on this. I'll update tests to start using this

apolcyn:
Actually for tests, can we follow up this PR to add support for using this new channel builder in the interop client? (probably around https://github.com/grpc/grpc-java/blob/master/interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java#L372)

jiangtaoli2016:
I forgot about that. Let me update the interop client with a new customCredentialTypes = "compute_engine_credentials" on Java side. @apolcyn can you update the python script with this new type after I submit java PR?

apolcyn:
yeah, I can update tests once that's in

Issue #5130, opened 2018-12-06
document for alts examples


dapengzhang0:
Because [ALTS examples](https://github.com/grpc/grpc-java/tree/master/examples/example-alts) can only run in ALTS-enabled environments, might need a README to document the requirements.

YifeiZhuang:
added via #7861

PR #5475, opened 2019-03-13
alts: add ComputeEngineChannelBuilder to interop test

jiangtaoli2016:
Add compute_engine_credentials as custom_credentials_type option.

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
@@ -20,6 +20,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.io.Files;
 import io.grpc.ManagedChannel;
 import io.grpc.alts.AltsChannelBuilder;
+import io.grpc.alts.ComputeEngineChannelBuilder;
 import io.grpc.alts.GoogleDefaultChannelBuilder;
 import io.grpc.internal.AbstractManagedChannelImplBuilder;
 import io.grpc.internal.GrpcUtil;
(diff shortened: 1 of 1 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java
```diff
@@ -373,6 +374,10 @@ protected ManagedChannel createChannel() {
           && customCredentialsType.equals("google_default_credentials")) {
         return GoogleDefaultChannelBuilder.forAddress(serverHost, serverPort).build();
       }
+      if (customCredentialsType != null
+          && customCredentialsType.equals("compute_engine_credentials")) {
```
apolcyn: sorry, mind changing to `compute_engine_channel_creds` ?

`compute_engine_credentials` is very similar to the `compute_engine_creds` test case, and it also may look like it corresponds to compute engine "call creds".

jiangtaoli2016: Done.

jiangtaoli2016:
@apolcyn revised. PTAL.

dapengzhang0:
A side note to @jiangtaoli2016: Could you please add an one paragraph README in examples/example-alts in what environment and what settings are needed so that external users can use alts channel?  #5130

sanjaypujare:
I second @dapengzhang0's suggestion. Good to have a README that's enough for a user to execute the example

jiangtaoli2016:
@dapengzhang0 @sanjaypujare I will open an email thread to discuss.  Thanks!

PR #5477, opened 2019-03-14
Move LB parsing logic into LB.Factory

carl-mastrangelo:
Sorry for the large commit, but a lot of things had to change for this to work.   At a high level, this is what has changed:

- LB.Factory now has a parse method.
- All known LB.Factories implement the parse, 
- Each factory still uses its existing config logic, but using the new parser
- AutoConfiguredLoadBalancerFactory is invoked from ManagedChannelImpl to parse and select a config

This is not a complete PR, and some follow on work is needed.  Please recognize this when reviewing.  I'm afraid if I keep making changes I will keep getting merge conflicts and the review will become impossible.

### DIFF
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -795,7 +795,6 @@ public abstract class LoadBalancer {
   @ThreadSafe
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1771")
   public abstract static class Factory {
-
     /**
      * Creates a {@link LoadBalancer} that will be used inside a channel.
      *
--- a/core/src/main/java/io/grpc/LoadBalancer.java
+++ b/core/src/main/java/io/grpc/LoadBalancer.java
@@ -795,6 +795,7 @@ public abstract class LoadBalancer {
   @ThreadSafe
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1771")
   public abstract static class Factory {
+
     /**
      * Creates a {@link LoadBalancer} that will be used inside a channel.
      *
--- a/core/src/main/java/io/grpc/LoadBalancerProvider.java
+++ b/core/src/main/java/io/grpc/LoadBalancerProvider.java
@@ -52,8 +52,8 @@ public abstract class LoadBalancerProvider extends LoadBalancer.Factory {
   public abstract String getPolicyName();
 
   /**
-   * Parses the Load Balancing policy for the service config.  This will return a
-   * {@link ConfigOrError} which contains either the successfully parsed config, or the
+   * Parses the config for the Load Balancing policy unpacked from the service config.  This will
+   * return a {@link ConfigOrError} which contains either the successfully parsed config, or the
    * {@link Status} representing the failure to parse.  Implementations are expected to not throw
    * exceptions but return a Status representing the failure.
    *
@@ -63,9 +63,9 @@ public abstract class LoadBalancerProvider extends LoadBalancer.Factory {
    * @since 1.20.0
    * @see https://github.com/grpc/proposal/blob/master/A24-lb-policy-config.md
    */
-  public ConfigOrError<?> parseLoadBalancingConfigPolicy(
-      Map<String, ?> rawLoadBalancingConfigPolicy) {
-    return ConfigOrError.UNSUPPORTED_CONFIG;
+  public ConfigOrError<?> parseLoadBalancingPolicyConfig(
+      Map<String, ?> rawLoadBalancingPolicyConfig) {
+    return ConfigOrError.UNKNOWN_CONFIG;
   }
 
   @Override
--- a/core/src/main/java/io/grpc/LoadBalancerProvider.java
+++ b/core/src/main/java/io/grpc/LoadBalancerProvider.java
@@ -57,7 +57,7 @@ public abstract class LoadBalancerProvider extends LoadBalancer.Factory {
    * {@link Status} representing the failure to parse.  Implementations are expected to not throw
    * exceptions but return a Status representing the failure.
    *
-   * @param rawLoadBalancingConfigPolicy The {@link Map} representation of the load balancing
+   * @param rawLoadBalancingPolicyConfig The {@link Map} representation of the load balancing
    *     policy choice.
    * @return a tuple of the fully parsed and validated balancer configuration, else the Status.
    * @since 1.20.0
--- a/core/src/main/java/io/grpc/LoadBalancerProvider.java
+++ b/core/src/main/java/io/grpc/LoadBalancerProvider.java
@@ -17,6 +17,8 @@
 package io.grpc;
 
 import com.google.common.base.MoreObjects;
+import io.grpc.NameResolver.Helper.ConfigOrError;
+import java.util.Map;
 
 /**
  * Provider of {@link LoadBalancer}s.  Each provider is bounded to a load-balancing policy name.
@@ -48,6 +50,23 @@ public abstract class LoadBalancerProvider extends LoadBalancer.Factory {
    * and can only start with letters.
    */
   public abstract String getPolicyName();
+
+  /**
+   * Parses the Load Balancing policy for the service config.  This will return a
+   * {@link ConfigOrError} which contains either the successfully parsed config, or the
+   * {@link Status} representing the failure to parse.  Implementations are expected to not throw
+   * exceptions but return a Status representing the failure.
+   *
+   * @param rawLoadBalancingConfigPolicy The {@link Map} representation of the load balancing
+   *     policy choice.
+   * @return a tuple of the fully parsed and validated balancer configuration, else the Status.
+   * @since 1.20.0
+   * @see https://github.com/grpc/proposal/blob/master/A24-lb-policy-config.md
+   */
+  public ConfigOrError<?> parseLoadBalancingConfigPolicy(
+      Map<String, ?> rawLoadBalancingConfigPolicy) {
+    return ConfigOrError.UNSUPPORTED_CONFIG;
+  }
 
   @Override
   public final String toString() {
--- a/core/src/main/java/io/grpc/NameResolver.java
+++ b/core/src/main/java/io/grpc/NameResolver.java
@@ -20,7 +20,6 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.base.MoreObjects;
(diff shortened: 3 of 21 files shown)


xds/src/main/java/io/grpc/xds/XdsLoadBalancer.java
```diff
@@ -91,17 +88,23 @@ public void handleResolvedAddressGroups(
...
     Map<String, ?> newRawLbConfig = checkNotNull(
         attributes.get(ATTR_LOAD_BALANCING_CONFIG), "ATTR_LOAD_BALANCING_CONFIG not available");
-    LbConfig newLbConfig = ServiceConfigUtil.unwrapLoadBalancingConfig(newRawLbConfig);
-    fallbackPolicy = selectFallbackPolicy(newLbConfig, lbRegistry);
+
+    ConfigOrError<XdsConfig> cfg =
+        XdsLoadBalancerProvider.parseLoadBalancingConfigPolicy(newRawLbConfig, lbRegistry);
+    if (cfg.getError() != null) {
+      throw cfg.getError().asRuntimeException();
+    }
```
zhangkun83: Do you expect all balancers to have this check-and-throw? If so, maybe we should add a `getConfigOrThrow()` to `ConfigOrError`.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer.java
```diff
@@ -318,4 +292,47 @@ public void run() {
...
     }
   }
+
+  /**
+   * Represents a successfully parsed and validated LoadBalancingConfig for XDS.
+   */
+  static final class XdsConfig {
+    private final String newBalancerName;
+    @Nullable
+    private final LbConfig childPolicy;
```
zhangkun83: `LbConfig` contains unparsed raw value. You should probably use a different type that is a `(policy, parsedConfig)` tuple. If you plan to do it later, please leave a TODO.

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -795,6 +795,7 @@ public ChannelLogger getChannelLogger() {
   @ThreadSafe
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1771")
   public abstract static class Factory {
+
```
zhangkun83: You should probably revert the whole file.

core/src/main/java/io/grpc/LoadBalancerProvider.java
```diff
@@ -49,6 +51,23 @@
...
+   * {@link Status} representing the failure to parse.  Implementations are expected to not throw
+   * exceptions but return a Status representing the failure.
+   *
+   * @param rawLoadBalancingConfigPolicy The {@link Map} representation of the load balancing
+   *     policy choice.
+   * @return a tuple of the fully parsed and validated balancer configuration, else the Status.
+   * @since 1.20.0
+   * @see https://github.com/grpc/proposal/blob/master/A24-lb-policy-config.md
+   */
+  public ConfigOrError<?> parseLoadBalancingConfigPolicy(
```
zhangkun83: `parseLoadBalancingPolicyConfig()`

core/src/main/java/io/grpc/NameResolver.java
```diff
@@ -272,6 +272,24 @@ public SynchronizationContext getSynchronizationContext() {
...
+          return "service config parsing not supported";
+        }
+      }
+
+      /**
+       * A sentinel value indicating that service config is not supported.   This can be used to
+       * indicate that parsing of the service config is neither right nor wrong, but doesn't have
+       * any meaning.
+       */
+      public static final ConfigOrError<?> UNSUPPORTED_CONFIG =
```
zhangkun83: I don't think "unsupported" is accurate here. Service config is not something a LB policy **supports** or not, it's a mechanism that the policy **uses** or not. If `DUMMY_CONFIG` doesn't sound good to you, maybe `NO_CONFIG` or `UNUSED_CONFIG`?

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -795,11 +796,28 @@ public ChannelLogger getChannelLogger() {
...
+     * {@link ConfigOrError} which contains either the successfully parsed config, or the
+     * {@link Status} representing the failure to parse.  Implementations are expected to not throw
+     * exceptions but return a Status representing the failure.
+     *
+     * @param rawLoadBalancingConfigPolicy The {@link Map} representation of the load balancing
+     *     policy choice.
+     * @return a tuple of the fully parsed and validated balancer configuration, else the Status.
+     * @since 1.20.0
+     */
+    public ConfigOrError<?> parseLoadBalancingConfigPolicy(
```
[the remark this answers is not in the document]
zhangkun83: Although the spec doesn't call out clearly, the documentation on `message LoadBalancingConfig`  "Selects LB policy and provides corresponding configuration." Each entry in the `oneof` corresponds to the policy, while the value of that entry is its configuration.

core/src/main/java/io/grpc/NameResolver.java
```diff
@@ -259,8 +259,7 @@ public SynchronizationContext getSynchronizationContext() {
      * @since 1.20.0
      */
     public ConfigOrError<?> parseServiceConfig(Map<String, ?> rawServiceConfig) {
-      return ConfigOrError.fromError(
-          Status.INTERNAL.withDescription("service config parsing not supported"));
+      throw new UnsupportedOperationException("should have been implemented");
```
carl-mastrangelo: updated

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancerFactory.java
```diff
@@ -37,7 +37,6 @@
 @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1782")
 @Deprecated
 public final class GrpclbLoadBalancerFactory extends LoadBalancer.Factory {
-
```
zhangkun83: Revert this line.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancerProvider.java
```diff
@@ -51,4 +59,68 @@ public LoadBalancer newLoadBalancer(LoadBalancer.Helper helper) {
         helper, new CachedSubchannelPool(), TimeProvider.SYSTEM_TIME_PROVIDER,
         new ExponentialBackoffPolicy.Provider());
   }
+
+  @Override
+  public ConfigOrError<?> parseLoadBalancingPolicyConfig(
+      Map<String, ?> rawLoadBalancingConfigPolicy) {
+    return parseLoadBalancingConfigPolicyInternal(rawLoadBalancingConfigPolicy);
```
zhangkun83: try-catch ?

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -291,8 +312,46 @@ private LoadBalancerProvider getProviderOrThrow(String policy, String choiceReas
...
+        }
+        return ConfigOrError.fromError(
+            Status.UNKNOWN.withDescription(
+                "None of " + policiesTried + " specified by Service Config are available."));
+      }
+      return ConfigOrError.fromError(
+          Status.UNKNOWN.withDescription("Empty Load Balancing config"));
+    } catch (RuntimeException e) {
+      return ConfigOrError.fromError(
+          Status.UNKNOWN.withDescription("can't parse load balancer configuration").withCause(e));
```
dapengzhang0: I think "Empty load balancing config"/"None of policiesTried are available" should be treated differently from "can't parse load balancer configuration". The former should not be considered as service config resolution error.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancerFactory.java
carl-mastrangelo: Done.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancerProvider.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
carl-mastrangelo: This is existing behavior.  See `decideLoadBalancerProvider`

zhangkun83: No, it's not the existing behavior. Look at the last `return` of `decideLoadBalancerProvider()`. `defaultPolicy` should be returned if config is empty and there is no balancer address.

carl-mastrangelo: You are right, I am mistaken.   I changed this to return null because CoE can't wrap null, and a sentinel CoE would be difficult to construct.  

PTAL

zhangkun83: I think you forgot to push your change.

zhangkun83: The caller of this method (NrHelper.parseServiceConfig()) doesn't expect `null` from this method. Because `NrHelper.parseServiceConfig()` cannot return `null`, it would have to either return a sentinel result for "no policy configured / use default policy", or just return the default policy directly.

Returning the default policy seems easier, and I would just do it in `selectLoadBalancerPolicy()` instead of `NrHelper.parseServiceConfig()`. It should be outside of the `try-catch`, because the absence of the default policy should be considered a bug rather than a configuration error.

carl-mastrangelo: There is no default policy, because the default LB is not available yet.  Only the ACLBF knows that.

carl-mastrangelo: resolved offline

zhangkun83: The `autoConfiguredLoadBalancerFactory != null` path of `NrHelper.parseServiceConfig()` is not covered in tests. Can you add tests for that?

carl-mastrangelo: Added a test

Issue #1770, opened 2016-05-03
Tracking Issue for NameResolution being experimental.


carl-mastrangelo:
Specific usages:

* ``

b-hoyt:
I'm interested Context-aware name resolution. This doesn't seem obviously possible with the current NameResolver API and usage, but I'm not familiar with the broader set of APIs. Thoughts?

ejona86:
@b-hoyt I don't know what you mean. Context-aware in what way? Why is it not possible with the current API?

b-hoyt:
Say for example each client making an RPC call wants to provide hints to name resolution for how the NameResolver might resolve some targetUri that represents an abstract destination "||map_service||". The clients provide these hints, which may vary between requests, as value in the current context.

I can picture how an interceptor, which has access to the Context, may have a reference to the NameResolver in order to supply it with the hints. The NameResolver could do its thing using those hints and updated the associated Listener with the fully ResolvedServerInfos. 

But there is no distinction between the ResolvedServerInfos for in-flight-request-A vs. in-flight-request-B, in other words Context scope < ResolvedServerInfos scope

zhangkun83:
@b-hoyt Name resolution is in the Channel scope, which outlives the Call scope. Allowing individual RPC calls to influence name resolution simply doesn't fit in our architecture.

ejona86:
I'd suggest writing a `Channel` (not even interrceptor) that creates Channels on-the-fly based on the Context, and re-uses them as appropriate. The other option is to plug into the LB API some, since in some ways GRPCLB does something similar to this.

But I don't see NameResolver supporting this case. As @zhangkun83 mentioned, it is pretty incompatible with the architecture.

carl-mastrangelo:
API meeting notes:

* Make UNKNOWN_CONFIG package private

ejona86:
Actually, it was that UNKNOWN_CONFIG would be private, within LoadBalancerProvider.

ejona86:
#7133 is for removing NameResolver.Factory.

PR #5486, opened 2019-03-20
 core: deprecate NR.L and add NR.Results

carl-mastrangelo:


core/src/main/java/io/grpc/NameResolver.java
```diff
@@ -262,6 +320,29 @@ public SynchronizationContext getSynchronizationContext() {
...
+    /**
+     * Updates the {@link Channel} with an error from the resolver. The caller is responsible for
+     * eventually invoking {@link NameResolver#refresh()} to re-attempt resolution.
+     *
+     * @param error a non-OK status
+     * @since 1.20.0
+     */
+    public void updateNameResolutionError(Status error) {
+      throw new UnsupportedOperationException("Not implemented");
+    }
```
zhangkun83: Since we have the `Observer`, these two methods should not be added.

core/src/main/java/io/grpc/NameResolver.java
```diff
@@ -350,4 +431,143 @@ public String toString() {
...
+    public List<EquivalentAddressGroup> getServers() {
+      return servers;
+    }
+
+    /**
+     * Gets the attributes associated with the servers resolved by name resolution.
+     *
+     * @since 1.20.0
+     */
+    @ResolutionResultAttr
```
zhangkun83: Maybe rename `Result` to `ResolutionResult` to match the annotation name.

core/src/main/java/io/grpc/internal/ForwardingNameResolver.java
```diff
@@ -38,10 +38,16 @@ public String getServiceAuthority() {
   }
 
   @Override
+  @SuppressWarnings("deprecation")
```
zhangkun83: Better be `@Deprecated`

core/src/test/java/io/grpc/internal/ForwardingNameResolverTest.java
```diff
@@ -72,4 +74,22 @@ public void onError(Status error) { }
     forwarder.start(listener);
     verify(delegate).start(listener);
   }
+
+
+  @Test
+  @SuppressWarnings("deprecation") // this will be removed in 1.21.0
```
zhangkun83: This shouldn't be necessary.

core/src/main/java/io/grpc/NameResolver.java
carl-mastrangelo: Sorry, removed.

core/src/main/java/io/grpc/NameResolver.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/internal/ForwardingNameResolver.java
carl-mastrangelo: Done.

core/src/test/java/io/grpc/internal/ForwardingNameResolverTest.java
carl-mastrangelo: Done.

core/src/main/java/io/grpc/NameResolver.java
```diff
@@ -350,4 +408,143 @@ public String toString() {
       }
     }
   }
+
+  /**
+   * Represents the results from a Name Resolver.
+   *
+   * @since 1.20.0
+   */
+  @ExperimentalApi("FIXME")
```
zhangkun83: You can use `NameResolver`'s tracking issue: https://github.com/grpc/grpc-java/issues/1770

carl-mastrangelo: done.

PR #5502, opened 2019-03-26
core: remove Type from ConfigOrError

carl-mastrangelo:
The type are turned out to not be very useful, so removing it.  

cc @zhangkun83

### DIFF
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -292,14 +292,16 @@ final class DnsNameResolver extends NameResolver {
 
       Attributes.Builder attrs = Attributes.newBuilder();
       if (!resolutionResults.txtRecords.isEmpty()) {
-        ConfigOrError<Map<String, ?>> serviceConfig =
+        ConfigOrError serviceConfig =
             parseServiceConfig(resolutionResults.txtRecords, random, getLocalHostname());
         if (serviceConfig != null) {
           if (serviceConfig.getError() != null) {
             savedObserver.onError(serviceConfig.getError());
             return;
           } else {
-            attrs.set(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG, serviceConfig.getConfig());
+            @SuppressWarnings("unchecked")
+            Map<String, ?> config = (Map<String, ?>) serviceConfig.getConfig();
+            attrs.set(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG, config);
           }
         }
       } else {
@@ -312,7 +314,7 @@ final class DnsNameResolver extends NameResolver {
   }
 
   @Nullable
-  static ConfigOrError<Map<String, ?>> parseServiceConfig(
+  static ConfigOrError parseServiceConfig(
       List<String> rawTxtRecords, Random random, String localHostname) {
     List<Map<String, ?>> possibleServiceConfigChoices;
     try {
@@ -337,7 +339,7 @@ final class DnsNameResolver extends NameResolver {
     if (possibleServiceConfig == null) {
       return null;
     }
-    return ConfigOrError.<Map<String, ?>>fromConfig(possibleServiceConfig);
+    return ConfigOrError.fromConfig(possibleServiceConfig);
   }
 
   private void resolve() {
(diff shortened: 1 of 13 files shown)


dapengzhang0:
It's unclear to see why generic type is not useful, because the `LoadBalancerProvider` API 
`ConfigOrError parseLoadBalancingPolicyConfig(Map<String, ?> rawLoadBalancingPolicyConfig)` is (essentially) unused yet. For the static method `static ConfigOrError<XdsConfig> parseLoadBalancingConfigPolicy()`, the generic type does seem beneficial.

carl-mastrangelo:
When there is an error, the CoE cannot be reused.  For example:

```java
ConfigOrError<ManagedChannelServiceConfig() {
  ConfigOrError<?> coe = func();
  if (coe.getError() != null) {
     return coe;
  }
  // ...
}
```

Also, in overloads of the LoadBalancerProvider subclasses, they frequently have to return an internal type, but they dont want to expose the type since the method is public.   For example Grpclb does not want to expose `ConfigOrError<Mode>`.    Lastly, a cast is always going to be needed cast by the LB so the type param is not that useful on the factory, since it cant be passed to the LB.

dapengzhang0:
>For example Grpclb does not want to expose ConfigOrError<Mode>.

But GrpclbLoadBalancerProvider as well as all other provider impls is internal.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -292,14 +292,16 @@ public void run() {
...
             parseServiceConfig(resolutionResults.txtRecords, random, getLocalHostname());
         if (serviceConfig != null) {
           if (serviceConfig.getError() != null) {
             savedObserver.onError(serviceConfig.getError());
             return;
           } else {
-            attrs.set(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG, serviceConfig.getConfig());
+            @SuppressWarnings("unchecked")
+            Map<String, ?> config = (Map<String, ?>) serviceConfig.getConfig();
+            attrs.set(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG, config);
```
dapengzhang0: This introduces an unchecked cast, which is definitely not better.

zhangkun83: This is probably the only case where the generic is beneficial. But it's in the implementation, not on API surface. Arguably `parseServiceConfig()` doesn't have to return `ConfigOrError`. It can be changed to return `Map<String, ?>` and throw on errors instead.

I am with @carl-mastrangelo, in that the generic is not useful on the public API, namely `NameResolver` and `LoadBalancer`:

1. `NameResolver` doesn't look at the config. It gets `ConfigOrError` from `NameResolver.Helper#parseServiceConfig`, check the error, and passes it to `NameResolver.Observer.onResult()`. `NameResolver.Helper#parseServiceConfig` would have to return `ConfigOrError<?>`, thus `NameResolver.Observer.onResult()` would only accept the same.

2. `LoadBalancerProvider` produces `ConfigOrError` and `LoadBalancer` accepts it. Intuitively it could benefit from having a generic `ConfigOrError`, but in practice that would require the generic type to be on `LoadBalancerProvider` and `LoadBalancer` too. That sounds like too much. Even if we made that change, the type information is lost in `LoadBalancerRegistry`, and there would be unchecked conversions in `ManagedChannelImpl`, which is not nice either.

dapengzhang0: There are tow cases a concrete type is beneficial: here and in `XdsConfig`. I'm OK with not using generic, I hate generic too, and I'm OK with losing type information somewhere, but still like not to lose concrete type information completely anywhere. So I would prefer either
1. The API not to use `ConfigOrError` but use `Object parseConfig(...) throws Exception` so that the override change to a concrete type
    ```
    @Override
    SpecificConfigType parseConfig(...) throws SpecificException
   ```

2. Making `ConfigOrError` non-final
   ```
    @Override
    SpecificConfigOrError parseConfig(...) throws SpecificException

    class SpecificConfigOrError extends ConfigOrError {
      @Override
      SpecificConfig getConfig()
    }
   ```

dapengzhang0: Update: Discussed with @zhangkun83 offline, the type information is completely lost at `ManagedChannelImpl`. LGTM

Issue #5496, opened 2019-03-24
Extract name-resolution logic from ManagedChannelImpl for re-use


zhangkun83:
First mentioned in [another comment](https://github.com/grpc/grpc-java/issues/1599#issuecomment-444595810), and then discussed with @ejona86 and @dapengzhang0, it was concluded that we need to extract the following logic:

1. Parse the target string and convert it into a URI
2. Manage NameResolver retries
3. Wire up NameResolver and LoadBalancer

out of `ManagedChannelImpl`, and make it usable by:
1. `ManagedChannelImpl`
2. `LoadBalancer.Helper#createResolvingOobChannel` (blocking #5495)
3. `XdsLoadBalancer` "custom mode": `XdsLoadBalancer` receives the target name for each locality, and needs to resolve the target name and start a child `LoadBlanacer` for it.

zhangkun83:
For **CHILD_POLICY** case, our preliminary API design looks like:
```java
// LbConfig class name is a placeholder for whatever type we choose for passing 
// parsed LoadBalancingConfig

abstract class LoadBalancer.Helper {
  RequestDirector createRequestDirector(String target, LbConfig initialConfig, Helper helper);
}

abstract class RequestDirector {
  abstract void handleSubchannelState(Subchannel subchannel, ConnectivityStateInfo newState);
  abstract void shutdown();
  abstract void handleLbConfigChange(LbConfig lbConfig);
}
```

zhangkun83:
With the changes made in #5722, we can get rid of `handleSubchannelState()` from the utility API. I also I come up with a new name because "RequestDirector" is not great. The up-to-date API will look like:
```java
final class ResolverBalancerPair {
  // Resolves a target and use the given LbConfig to create the LB.
  static ResolverBalancerPair newPair(
      NameResolverRegistry, String target, ResolutionResultFilter, NameResolver.Args, Helper);

  // Shuts down both NR and LB
  void shutdown();

  interface ResolutionResultFilter {
    LoadBalancer.ResolvedAddresses filter(NameResolver.ResolutionResult result);
  }
}
```

`ResolutionResultFilter` is called on every `ResolutionResult` from `NameResolver`. It will have different behaviors in the three cases:
- In `ChannelImpl`, this will implement service config handling, extract LB config, parse it, and include it in the `ResolvedAddresses`.
- In `OOB`, will only take the addresses and put it in `ResolvedAddresses`. The service config will be discarded.
[...]

PR #5503, opened 2019-03-26
core: pass Subchannel state updates to SubchannelStateListener rather than LoadBalancer

### LINKED ISSUE #5497: Move LoadBalancer.handleSubchannelState() to a per-Subchannel listener interface
In hierarchical `LoadBalancer`s (e.g., `XdsLoadBalancer`) or wrapped `LoadBalancer`s (e.g., `HealthCheckingLoadBalancerFactory`, the top-level `LoadBalancer` receives `Subchannel` state updates from the Channel impl, and they almost always pass it down to its children `LoadBalancer`s.

Sometimes the children `LoadBalancer`s are not directly created by the parent, thus requires whatever API in the middle to also pass Subchannel state updates, complicating that API. For example, the proposed [`RequestDirector`](https://github.com/grpc/grpc-java/issues/5496#issuecomment-476008051) includes `handleSubchannelState()` solely to plumb state updates to where they are used.

If we instead pass a listener when creating a `Subchannel` to accept state updates, those updates could be directly passed to where the `Subchannel` is created, skipping the explicit chaining in the middle:
```java
abstract class SubchannelStateListener {
  void onSubchannelState(ConnectivityStateInfo newState);
}

abstract class LoadBalancer.Helper {
  abstract Subchannel createSubchannel(
      List<EquivalentAddressGroup> addrs, Attributes attrs, SubchannelStateListener stateListener);
}
```

If a wrapping `LoadBalancer` wants to look at or alter the Subchannel state updates for its delegate (like in `HealthCheckingLoadBalancerFactory`), it can still do so in the wrapping `LoadBalancer.Helper` passed to the delegate, by intercepting the `SubchannelStateListener`.

zhangkun83:
Resolves #5497

## Notes for reviewers
The API and the needed changes in `LoadBalancer` implementations are relatively simple.  The large changes are mostly in tests. Please focus on non-test files first.

## Motivation

In hierarchical `LoadBalancer`s (e.g., `XdsLoadBalancer`) or wrapped `LoadBalancer`s (e.g., `HealthCheckingLoadBalancerFactory`, the top-level `LoadBalancer` receives `Subchannel` state updates from the Channel impl, and they almost always pass it down to its children `LoadBalancer`s.

Sometimes the children `LoadBalancer`s are not directly created by the parent, thus requires whatever API in the middle to also pass Subchannel state updates, complicating that API. For example, the proposed [`RequestDirector`](https://github.com/grpc/grpc-java/issues/5496#issuecomment-476008051) includes `handleSubchannelState()` solely to plumb state updates to where they are used. We also see this pattern in `HealthCheckingLoadBalancerFactory`, `GrpclbState` and `SubchannelPool`.

Another minor issue is, the parent `LoadBalancer` would need to intercept the `Helper` passed to its children to map Subchannels to the children `LoadBalancer`s, so that it pass updates about relevant Subchannels to the children.  Otherwise, a child `LoadBalancer` may be surprised by seeing Subchannel not created by it, and it's not efficient to broadcast Subchannel updates to all children.

## API Proposal
We will pass a `SubchannelStateListener` when creating a `Subchannel` to accept state updates, those updates could be directly passed to where the `Subchannel` is created, skipping the explicit chaining in the middle.

Also define a first-class object `CreateSubchannelArgs` to pass arguments for the reasons below:
1. It may avoid breakages when we add new arguments to `createSubchannel()`.  For example, a `LoadBalancer` may wrap `Helper` and intercept `createSubchannel()` in a hierarchical case. It may not be interested in all arguments. Passing a single `CreateSubchannelArgs` will not break the parent `LoadBalancer` if we add new fields later.
2. This also reduces the eventual size of Helper interface, as the convenience `createSubchannel()` that accepts one EAG instead of a List is no longer necessary, since that convenience is moved into `CreateSubchannelArgs`.

```java
interface SubchannelStateListener {
  void onSubchannelState(Subchannel subchannel, ConnectivityStateInfo newState);
}

abstract class LoadBalancer.Helper {
  abstract Subchannel createSubchannel(CreateSubchannelArgs args);
}

final class CreateSubchannelArgs {
  final List<EquivalentAddressGroup> getAddresses();
  final Attributes getAttributes();
  final SubchannelStateListener getStateListener();
  final class Builder () {
    ...
  }
}
```

The new `createSubchannel()` must be called from synchronization context, as a step towards #5015.

## How the new API helps

Most hierarchical `LoadBalancer`s would just let the listener from the child `LoadBalancer`s directly pass through to gRPC core, which is less boilerplate than before.

Without any effort by the parent, each child will only see updates for the Subchannels it has created, which is clearer and more efficient.

If a parent `LoadBalancer` does want to look at or alter the Subchannel state updates for its delegate (like in `HealthCheckingLoadBalancerFactory`), it can still do so in the wrapping `LoadBalancer.Helper` passed to the delegate by intercepting the `SubchannelStateListener`.

## Migration implications
Existing `LoadBalancer` implementations will continue to work, while they will see deprecation warnings when compiled:
 - The old `LoadBalancer.Helper#createSubchannel` variants are now deprecated, but will work until they are deleted. They create a `SubchannelStateListener` that delegates to `LoadBalancer#handleSubchannelState`.
 - `LoadBalancer#handleSubchannelState` is now deprecated, and will throw if called and the implementation doesn't override it. It will be deleted in a future release.

The migration for most `LoadBalancer` implementation would be moving the logic from `LoadBalancer#handleSubchannelState` into a `SubchannelStateListener`.


carl-mastrangelo:
Quick thoughts before diving in:

* Listeners in our code typically get passed in at `start`, not construction.  (e.g. NameResolver.Listener, ClientCall.Listener, ServerCall.Listener, InternalServer.Listener, ManagedClientTransport.Listener)   I think you should keep with this pattern, and attach the listener to the subchannel later.  This would fill a missing gap in Subchannels API, which already has a shutdown() method.

* In context of the comment above, I think that would allow the `Subchannel` arg from the listener to go away, since the LB can late-associate the listener with the Subchannel.

zhangkun83:
> Listeners in our code typically get passed in at start, not construction. (e.g. NameResolver.Listener, ClientCall.Listener, ServerCall.Listener, InternalServer.Listener, ManagedClientTransport.Listener) I think you should keep with this pattern, and attach the listener to the subchannel later. This would fill a missing gap in Subchannels API, which already has a shutdown() method.
> In context of the comment above, I think that would allow the Subchannel arg from the listener to go away, since the LB can late-associate the listener with the Subchannel.

I thought about that option, but didn't go that route. The Subchannel arg from the listener actually fits naturally. A `LoadBalancer` that creates more than one `Subchannel`s will almost always have one listener that looks at the states of all `Subchannel`s and update the picker. If we stop passing `Subchannel` to the listener, they would end up doing something like:
```java
...
subchannel.start((state) -> handleState(subchannel, state));
...
private void handleState(Subchannel subchannel, ConnectivityStateInfo state) {
...
}
```

For the sake of API style consistency, and a relatively smaller churn on the API (which mostly affects unit tests), it may be the right thing to do.

zhangkun83:
Another reason that may justify passing the listener to `createSubchannel()`: prevent bugs in LoadBalancer implementations.

The interfaces that has `start()`, e.g.,`ClientCall`, are **active objects** in the sense they act on their own, and the listener is the only way to receive outputs from it. Thus it's impossible to make the mistake of "forgetting to call start()". It would be very obvious in the code.

`Subchannel` is a passive object, as it's returned as a result of a pick. It by its own doesn't produce any output, and the listener is for making decisions related to it (include it or not). It's possible that a `LoadBalancer` forgets to call `start()` and misses out all the updates from that Subchannel.

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -471,6 +481,141 @@ public boolean equals(Object other) {
...
+   */
+  public static final class CreateSubchannelArgs {
+    private final List<EquivalentAddressGroup> addrs;
+    private final Attributes attrs;
+    private final SubchannelStateListener stateListener;
+
+    private CreateSubchannelArgs(
+        List<EquivalentAddressGroup> addrs, Attributes attrs,
+        SubchannelStateListener stateListener) {
+      this.addrs = checkNotNull(addrs, "addresses are not set");
```
carl-mastrangelo: This needs to be a copy

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -471,6 +481,141 @@ public boolean equals(Object other) {
...
+
+      /**
+       * The addresses to connect to.  All addresses are considered equivalent and will be tried
+       * in the order they are provided.
+       */
+      public Builder setAddresses(EquivalentAddressGroup addrs) {
+        this.addrs = Collections.singletonList(addrs);
+        return this;
+      }
+
```
carl-mastrangelo: make Builders ctor package private

grpclb/src/main/java/io/grpc/grpclb/CachedSubchannelPool.java
```diff
@@ -23,65 +23,77 @@
...
   @Override
   public Subchannel takeOrCreateSubchannel(
-      EquivalentAddressGroup eag, Attributes defaultAttributes) {
+      EquivalentAddressGroup eag, Attributes defaultAttributes, SubchannelStateListener listener) {
     final CacheEntry entry = cache.remove(eag);
     final Subchannel subchannel;
     if (entry == null) {
-      subchannel = helper.createSubchannel(eag, defaultAttributes);
+      final Attributes attrs = defaultAttributes.toBuilder()
+          .set(STATE_LISTENER, new AtomicReference<>(listener)).build();
```
carl-mastrangelo: nit: .build() should be on the next line

grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
```diff
@@ -150,6 +152,25 @@ public String toString() {
   private RoundRobinPicker currentPicker =
       new RoundRobinPicker(Collections.<DropEntry>emptyList(), Arrays.asList(BUFFER_ENTRY));
 
+  private final SubchannelStateListener subchannelStateListener = new SubchannelStateListener() {
```
carl-mastrangelo: One thing you should consider:  Make these listeners a named private class.  It makes debugging much easier in a stack trace.  Also, the named classes can be made final, which avoids letting Mockito.spy extend them.

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
```diff
@@ -137,20 +141,25 @@ public Subchannel answer(InvocationOnMock invocation) throws Throwable {
   }
 
   @After
+  @SuppressWarnings({"deprecation", "unchecked"})
   public void tearDown() throws Exception {
     verifyNoMoreInteractions(mockArgs);
+    verify(mockHelper, never()).createSubchannel(any(List.class), any(Attributes.class));
```
dapengzhang0: nit: `ArgumentMatchers.<List<EAG>>any()` does not need suppress  unchecked.

grpclb/src/main/java/io/grpc/grpclb/CachedSubchannelPool.java
```diff
@@ -23,65 +23,77 @@
...
+            })
+          .build());
     } else {
       subchannel = entry.subchannel;
       entry.shutdownTimer.cancel();
-      // Make the balancer up-to-date with the latest state in case it has changed while it's
-      // in the cache.
+      subchannel.getAttributes().get(STATE_LISTENER).set(listener);
+      // Make the listener up-to-date with the latest state in case it has changed while it's in the
+      // cache.
```
dapengzhang0: Is the whole method in syncContext? Why not run `listener.onSubchannelState(subchannel, entry.state)` directly, so that there will be no state change?

zhangkun83: `onSubchannelState()` is not re-entrant with the other logic from `GrpclbState` when called by channel impl. Although reentrancy may not cause a problem, it's better to keep the baheavior closer to with channel impl.

That said, I think there is a problem with the current code. If channel impl schedules an update before the task on the next line while this method is running, the update from the channel, which is newer, would be overwritten with the older value by the latter task. I went with a different approach of the implementation that always track Subchannels and their states, whether in or out of the pool, which ends up simpler.

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
zhangkun83: Done.

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -963,6 +1137,39 @@ public ChannelLogger getChannelLogger() {
     }
   }
 
+  /**
+   * Receives state changes for one or more {@link Subchannel}s. All methods are run under {@link
+   * Helper#getSynchronizationContext}.
+   *
+   * @since 1.21.0
+   */
+  public abstract static class SubchannelStateListener {
```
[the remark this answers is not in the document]
zhangkun83: Sure. Changed to interface. As we discussed, if we ever want to plumb more Subchannel information back to the LoadBalancer, we can add another listener in `CreateSubchannelArgs`.

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -648,6 +657,144 @@ public boolean equals(Object other) {
...
+       * The addresses to connect to.  All addresses are considered equivalent and will be tried
+       * in the order they are provided.
+       */
+      public Builder setAddresses(EquivalentAddressGroup addrs) {
+        this.addrs = Collections.singletonList(addrs);
+        return this;
+      }
+
+      /**
+       * Required.  The addresses to connect to.  All addresses are considered equivalent and will
```
ejona86: Having "Required." as its own sentence here hurts the readability of the javadoc. Javadoc using the first sentence as its summary, so basically this won't have any documentation in the summary view at the top of the page.

core/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -648,6 +657,144 @@ public boolean equals(Object other) {
...
+      /**
+       * Creates a new args object.
+       */
+      public CreateSubchannelArgs build() {
+        return new CreateSubchannelArgs(addrs, attrs, stateListener);
+      }
+    }
+
+    @Override
+    public String toString() {
```
ejona86: It'd be easier to read this class if these methods weren't split by the Builder class in-between.

PR #5504, opened 2019-03-26
core: Avoid locks in SynchronizationContext

njhill:
Similar to what's done in `SerializingExecutor`, it's easy to make `SynchronizationContext` non-blocking.

### DIFF
--- a/core/src/main/java/io/grpc/SynchronizationContext.java
+++ b/core/src/main/java/io/grpc/SynchronizationContext.java
@@ -20,13 +20,13 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 
 import java.lang.Thread.UncaughtExceptionHandler;
-import java.util.ArrayDeque;
 import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
-import javax.annotation.concurrent.GuardedBy;
+import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/grpc/SynchronizationContext.java
```diff
@@ -113,13 +105,11 @@ public final void drain() {
    * #executeLater} in the lock, and call {@link #drain} outside the lock.
    */
   public final void executeLater(Runnable runnable) {
-    synchronized (lock) {
-      queue.add(checkNotNull(runnable, "runnable is null"));
-    }
+    queue.add(checkNotNull(runnable, "runnable is null"));
```
dapengzhang0: This may change the current behavior because the ordering of the runnables may be changed. There seems no harm for the ordering change (May need confirm from @zhangkun83 ). This change seems better, because `executeLater` becomes always not blocking after the change.

core/src/main/java/io/grpc/SynchronizationContext.java
```diff
@@ -80,29 +77,24 @@ public SynchronizationContext(UncaughtExceptionHandler uncaughtExceptionHandler)
...
+        Runnable runnable;
+        while ((runnable = queue.poll()) != null) {
+          try {
+            runnable.run();
+          } catch (Throwable t) {
+            uncaughtExceptionHandler.uncaughtException(Thread.currentThread(), t);
+          }
+        }
+      } finally {
+        drainingThread.set(null);
```
zhangkun83: There seems to be a race:

T1: call drain(): set drainingThread. Finish all tasks, exit loop
T2: call executeLater(): add a new task
T2: call drain(): try to set drainingThread, but is still set, give up
T1: still in drain(): reset drainingThread, then exit

End result: the task added by T2 will not be run unless someone calls drain() again.

zhangkun83: Nm, I see you have another loop to catch that.

dapengzhang0: >T1: still in drain(): reset drainingThread, then exit

For T1, because there is  `while (!queue.isEmpty())` in the end, it will drain again, not exit.

dapengzhang0: I agree without lock it's harder to prove the correctness and easier to break the correctness in the future though. But this seems to have the same complexity as the `SerializingExecutor`.

njhill:
Thanks @carl-mastrangelo  @dapengzhang0 @zhangkun83. A possible variation would be to use an `AtomicBoolean` and `ThreadLocal<Boolean>` instead of the `AtomicReference<Thread>`, i.e. no need to use `Thread.currentThread()`. This would make `throwIfNotInThisSynchronizationContext` cheaper - just a TL instead of volatile read. But as @carl-mastrangelo said, not on request path so I guess not that perf-critical.

core/src/main/java/io/grpc/SynchronizationContext.java
dapengzhang0: Discussed offline, the behavior is more or less the same. So no concern here.

PR #5512, opened 2019-03-28
GoogleAuthLibraryCallCredentials uses Credentials Builder

### LINKED ISSUE #5509: Use the JWT service builder in GoogleAuthLibraryCallCredentials
`GoogleAuthLibraryCallCredentials.JwtHelper` uses a deprecated constructor to create a `ServiceAccountJwtAccessCredentials`.  The deprecated constructor was removed in `google-auth-library-java`'s `0.14.0` release, but given its use in grpc, the Constructor was re-added in `0.15.0`.  The deprecated constructor will be removed in future releases.

There are`*Credentials.Builder` which is the preferred method of creating any `Credentials`, and have been available since the `0.8.0` release.  Please use `ServiceAccountJwtAccessCredentials.Builder` instead of the `ServiceAccountJwtAccessCredentials` constructor.

NOTE: I will submit a PR for this.

sduskis:
Replacing use of a deprecated constructor with a builder.  The constructor will be removed soon.

Fixes #5509


sduskis:
Signed it.

sduskis:
I signed it.

auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java
```diff
@@ -245,31 +244,59 @@ static JwtHelper createJwtHelperOrNull(ClassLoader loader) {
...
           .asSubclass(Credentials.class);
-      jwtConstructor
-          = jwtClass.getConstructor(String.class, String.class, PrivateKey.class, String.class);
+      newJwtBuilder = jwtClass.getDeclaredMethod("newBuilder");
+      Class<?> builderClass = newJwtBuilder.getReturnType();
+      build = builderClass.getMethod("build");
+
+      methodPairs = new ArrayList<>();
+      for (String key : new String[] { "ClientId", "ClientEmail", "PrivateKey", "PrivateKey" }) {
+        Method getter = serviceAccountClass.getMethod("get" + key);
```
ejona86: The string generation here a bit worrying, since things like ProGuard can rename methods, and they won't be able to notice references like this. We've had to be explicit and repeat ourselves many times in the past. That generally means the string arguments have to be literals.

```java
methodPairs.add(new MethodPair(
    serviceAccountClass.getMethod("getClientId"),
    builderClass.getMethod("setClientId", getter.getReturnType())));
```

sduskis: Will do.  Thanks for the catch.

sduskis:
@ejona86, this PR should be ready.

ejona86:
@sduskis, thank you! BTW: I did verify that the pre-existing code wouldn't break horribly when you do the API change. It will only issue a warning when being created and then use oauth instead of JWT.

Issue #5510, opened 2019-03-28
There should be a way to disable OpenCensus stats and tracing


njhill:
Currently, out-of-the-box client and server config includes a very expensive "no-op" in the form of the enabled-by-default opencensus tracing and stats interceptors. This is the case even if no implementation library is on the classpath, and disabling them is difficult/discouraged/impossible (without reflection) depending on the transport.

```
Benchmark                         (direct)  (statsAndTracing)  (transport)  Mode  Cnt      Score      Error  Units
TransportBenchmark.unaryCall1024      true               true    INPROCESS  avgt   10   3313.595 ±  107.360  ns/op
TransportBenchmark.unaryCall1024      true              false    INPROCESS  avgt   10   2142.610 ±   99.678  ns/op
TransportBenchmark.unaryCall1024      true               true  NETTY_LOCAL  avgt   10  64780.414 ± 1564.422  ns/op
TransportBenchmark.unaryCall1024      true              false  NETTY_LOCAL  avgt   10  60969.867 ± 1447.113  ns/op
TransportBenchmark.unaryCall1024      true               true        NETTY  avgt   10  85561.145 ± 3765.635  ns/op
TransportBenchmark.unaryCall1024      true              false        NETTY  avgt   10  81556.086 ± 2136.174  ns/op
TransportBenchmark.unaryCall1024      true               true  NETTY_EPOLL  avgt   10  79758.331 ± 3005.046  ns/op
TransportBenchmark.unaryCall1024      true              false  NETTY_EPOLL  avgt   10  75667.486 ± 2734.508  ns/op
TransportBenchmark.unaryCall1024      true               true       OKHTTP  avgt   10  74714.350 ± 2848.833  ns/op
TransportBenchmark.unaryCall1024      true              false       OKHTTP  avgt   10  70527.535 ± 2945.292  ns/op
```
There's also significantly more garbage produced when enabled.

It would be great to enable these only when the impl library is detected, however:
- There are circumstances when it might be there but not used
- It may be the case that only one of stats or tracing is being used
- Even if it is being used, instrumentation overhead of this magnitude may be considered undesirable (and these numbers are just the api plumbing without any additional cost from the impl itself)

Could these be changed to be disabled by default? I'm unsure why such settings are considered internal when they pertain to external functionality. There's not much danger of anyone "accidentally" leaving it disabled since it would be immediately apparent that the stats aren't working but the converse is of course true and I would assume most common.

dapengzhang0:
[...]

PR #5516, opened 2019-03-28
stub: optimize ThreadlessExecutor used for blocking calls

njhill:
The `ThreadlessExecutor` currently used for blocking calls uses `LinkedBlockingQueue` which is relatively heavy both in terms of allocations and synchronization overhead (e.g. when compared to `ConcurrentLinkedQueue`). It accounts for ~10% of allocations and ~5% of allocated bytes per-call in the `TransportBenchmark` when using in-process transport with [stats and tracing disabled](https://github.com/grpc/grpc-java/issues/5510).

Changing to use a `ConcurrentLinkedQueue` results in a ~5% speedup of that benchmark.

### DIFF
--- a/stub/src/main/java/io/grpc/stub/ClientCalls.java
+++ b/stub/src/main/java/io/grpc/stub/ClientCalls.java
@@ -640,14 +640,18 @@ public final class ClientCalls {
 
     /**
      * Waits until there is a Runnable, then executes it and all queued Runnables after it.
+     * Must only be called by one thread at a time.
      */
     public void waitAndDrain() throws InterruptedException {
+      final Thread currentThread = Thread.currentThread();
+      throwIfInterrupted(currentThread);
       Runnable runnable = poll();
       if (runnable == null) {
-        waiter = Thread.currentThread();
+        waiter = currentThread;
         try {
           while ((runnable = poll()) == null) {
             LockSupport.park(this);
+            throwIfInterrupted(currentThread);
           }
         } finally {
           waiter = null;
@@ -662,13 +666,16 @@ public final class ClientCalls {
       } while ((runnable = poll()) != null);
     }
 
+    private static void throwIfInterrupted(Thread currentThread) throws InterruptedException {
+      if (currentThread.isInterrupted()) {
+        throw new InterruptedException();
+      }
+    }
+
     @Override
     public void execute(Runnable runnable) {
       add(runnable);
-      final Thread waitingThread = waiter;
-      if (waitingThread != null) {
-        LockSupport.unpark(waitingThread);
-      }
+      LockSupport.unpark(waiter); // no-op if null
     }
   }
 }
--- a/stub/src/main/java/io/grpc/stub/ClientCalls.java
+++ b/stub/src/main/java/io/grpc/stub/ClientCalls.java
@@ -34,10 +34,11 @@ import java.util.Iterator;
 import java.util.NoSuchElementException;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
-import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.locks.LockSupport;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
@@ -627,10 +628,12 @@ public final class ClientCalls {
     }
   }
 
-  private static final class ThreadlessExecutor implements Executor {
+  @SuppressWarnings("serial")
+  private static final class ThreadlessExecutor extends ConcurrentLinkedQueue<Runnable>
+      implements Executor {
     private static final Logger log = Logger.getLogger(ThreadlessExecutor.class.getName());
 
-    private final BlockingQueue<Runnable> queue = new LinkedBlockingQueue<>();
+    private volatile Thread waiter;
 
     // Non private to avoid synthetic class
     ThreadlessExecutor() {}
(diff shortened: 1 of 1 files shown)


stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -639,20 +642,33 @@ public void onClose(Status status, Metadata trailers) {
      * Waits until there is a Runnable, then executes it and all queued Runnables after it.
      */
     public void waitAndDrain() throws InterruptedException {
-      Runnable runnable = queue.take();
-      while (runnable != null) {
+      Runnable runnable = poll();
```
dapengzhang0: Does it throw anymore?

dapengzhang0: Maybe mark `waitAndDrain()` with `@NotThreadSafe` because there must not be two concurrent callers of it.

stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -639,20 +642,33 @@ public void onClose(Status status, Metadata trailers) {
...
+      if (runnable == null) {
+        waiter = Thread.currentThread();
+        try {
```
dapengzhang0: `try` seems not needed?

stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -639,20 +642,33 @@ public void onClose(Status status, Metadata trailers) {
...
+      } while ((runnable = poll()) != null);
     }
 
     @Override
     public void execute(Runnable runnable) {
-      queue.add(runnable);
+      add(runnable);
+      final Thread waitingThread = waiter;
+      if (waitingThread != null) {
+        LockSupport.unpark(waitingThread);
```
dapengzhang0: Why not just `LockSupport.unpark(waiter)`? If waiter is null, there is no effect.

stub/src/main/java/io/grpc/stub/ClientCalls.java
njhill: This line doesn't but `LockSupport.park()` does (that's where the blocking now happens). The net behaviour should be equivalent to before including w.r.t. interruptions.

njhill: Sure, the class is only for internal use in an SPSC context.

carl-mastrangelo: Are you sure?  Interrupt can wake up park, but it will go back to sleep, making interruption a noop.

carl-mastrangelo: One possible thing you could try is to make this SPSC.  I have a POC (originally for SerializingExecutor) here:  https://github.com/grpc/grpc-java/pull/3778/files

dapengzhang0: `InterruptedException` is a checked Exception. It's not in the signature of `LockSupport.park()` so it's never thrown by it. The the thread is interrupted when `LockSupport.park()` is blocking, `LockSupport.park()` just exits.

njhill: @dapengzhang0 @carl-mastrangelo apologies you're right of course :) need to check interrupt flag after wake-up like you say and throw explicitly

stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -639,20 +642,33 @@ public void onClose(Status status, Metadata trailers) {
...
      */
     public void waitAndDrain() throws InterruptedException {
-      Runnable runnable = queue.take();
-      while (runnable != null) {
+      Runnable runnable = poll();
+      if (runnable == null) {
+        waiter = Thread.currentThread();
+        try {
+          while ((runnable = poll()) == null) {
+            LockSupport.park(this);
```
dapengzhang0: There is no synchronization, so the argument is misleading?

stub/src/main/java/io/grpc/stub/ClientCalls.java
njhill: @carl-mastrangelo sure, I remember seeing that before, could be worth a try here too. I thought (possibly mistakenly) this would be a simpler change just to circumvent `LinkedBlockingQueue` which was the main goal.

stub/src/main/java/io/grpc/stub/ClientCalls.java
njhill: @dapengzhang0 from [the docs](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/LockSupport.html):
> This object is recorded while the thread is blocked to permit monitoring and diagnostic tools to identify the reasons that threads are blocked. ... The normal argument to supply as a `blocker` within a lock implementation is `this`.

njhill:
Thanks @dapengzhang0 @carl-mastrangelo, have addressed the comments, PTAL

dapengzhang0:
LGTM

carl-mastrangelo:
@njhill Can you include your before and after JMH numbers for the commit?  We typically include them when making performance optimizations.

njhill:
@carl-mastrangelo I thought I had observed a bigger difference in the non-direct case in other runs, when the system was noisier. I know it's not a huge delta but there's a couple more similar changes I have in mind which cumulatively add up to maybe ~15% (to be confirmed!)

Before:
```
Benchmark                         (direct)  (transport)  Mode  Cnt      Score     Error  Units
TransportBenchmark.unaryCall1024      true    INPROCESS  avgt   60   1877.339 ±  46.309  ns/op
TransportBenchmark.unaryCall1024     false    INPROCESS  avgt   60  12680.525 ± 208.684  ns/op
```

After:
```
Benchmark                         (direct)  (transport)  Mode  Cnt      Score     Error  Units
TransportBenchmark.unaryCall1024      true    INPROCESS  avgt   60   1779.188 ±  36.769  ns/op
TransportBenchmark.unaryCall1024     false    INPROCESS  avgt   60  12532.470 ± 238.271  ns/op
```

This is with the following changes to default config:
- Set `tracingEnabled` and `statsEnabled` to false in channel and server builders
- Bumped forks 1 -> 2, iterations 10 -> 30 and changed mode to `AverageTime`

carl-mastrangelo:
@njhill merged, thanks!

dapengzhang0:
Thanks a lot for your PR @njhill

PR #5517, opened 2019-03-29
Revert " core: deprecate NR.L and add NR.Results "

zhangkun83:
This reverts commit f4a31ec62d038abf6faea7abaf377e2df3d308f0.

This change is part of a series of unfinished API changes and refactors for [Service Config Error Handling](https://github.com/grpc/proposal/blob/master/A21-service-config-error-handling.md#service-config-error-handling). It by its own should not go into the release.


dapengzhang0:
Why master branch not v1.20.x?

zhangkun83:
@dapengzhang0 Thanks for the catch!

PR #5525, opened 2019-04-01
 android-interop-testing/examples: upgrade android plugin

### LINKED ISSUE #5523: Android Interopt Test is breaking due to mockito upgrade
See https://github.com/grpc/grpc-java/pull/5484#issuecomment-478637238

cc @TimvdLippe

dapengzhang0:
This resolves #5523

While bumping `com.android.tools.build:gradle:3.1.2` to `3.3.0`, some other plugins/artifacts/maven repo/buildscripts have to be updated:

- gradle (wrapper) need to upgrade to 4.10.x
- protobuf gradle plugin need to bump a version compatible with gradle version.
- need add `google()` and `jcenter()` repos for android (otherwise `com.android.tools.build:aapt2:3.3.0x` and `trove4j` will not be found resp.)
- need to accept license for Android "build-tools;28.0.3" in kokoro env.


android/build.gradle
```diff
@@ -36,6 +36,8 @@ android {
 }
 
 repositories {
+    google()
```
ejona86: Why were these added?

dapengzhang0: I had a build error "Could not find com.android.tools.build:aapt2:3.3.0-5013011".

buildscripts/kokoro/android-interop.sh
```diff
@@ -15,6 +15,9 @@ export LDFLAGS=-L/tmp/protobuf/lib
 export CXXFLAGS=-I/tmp/protobuf/include
 export LD_LIBRARY_PATH=/tmp/protobuf/lib
 export OS_NAME=$(uname)
+export ANDROID_SDK_HOME=/opt/android-sdk/current
```
ejona86: ANDROID_HOME is already defined with this value.

buildscripts/kokoro/android-interop.sh
```diff
@@ -15,6 +15,9 @@ export LDFLAGS=-L/tmp/protobuf/lib
...
+
+echo y | ${ANDROID_SDK_HOME}/tools/bin/sdkmanager "build-tools;28.0.3"
```
ejona86: Hmm... this will actually install this particular version. I sort of question if it would be better to drop the right MD5s into licenses/android-sdk-license

buildscripts/kokoro/android-interop.sh
dapengzhang0: Is `ANDROID_HOME` already defined for both `android` and `android-interop-test` in kokoru env?

ejona86: Kokoro defines ANDROID_HOME. I think the android gradle plugin itself requires ANDROID_HOME to be set.

buildscripts/kokoro/android-interop.sh
dapengzhang0: linux-artifacts is also failing. The same issue. http://g/kokoro-users/_P4EAOWi-ko/PRFQvowPDgAJ

dapengzhang0:
Now Android Interop is passing in kokoro.

android/build.gradle
ejona86: That explains `google`, but why jcenter?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.9-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.1-all.zip
```
ejona86: -bin (to match the others, and because it is smaller)

android/build.gradle
dapengzhang0: After fixing  "Could not find com.android.tools.build:aapt2:3.3.0-5013011" by adding `google()`, there is another artifact missing `Could not find org.jetbrains.trove4j:trove4j`.

buildscripts/kokoro/android-interop.sh
dapengzhang0: >I sort of question if it would be better to drop the right MD5s into licenses/android-sdk-license

I don't understand how to do that/ how that works.

gradle/wrapper/gradle-wrapper.properties
dapengzhang0: Done.

buildscripts/kokoro/android-interop.sh
ejona86: Hmm... it seems the license handling is done differently these days. Previously after you accepted a licenses you could get any of the binaries with that license. But that doesn't seem to work any more. So this seems fine.

PR #5535, opened 2019-04-03
testing: move mockito from compile to testCompile

dapengzhang0:


dapengzhang0:
This breaks android-interopt-test again because AbstractInteropTest depends on mockito. Will add mockito compile dep in android-interopt-test/build.gradle.

PR #5536, opened 2019-04-03
 core: add client and server call perfmark annotations

carl-mastrangelo:
This code is highly experimental, so we can change it at will later.  This PR is to go ahead and try it out.

Perfmark task calls are added to the client and server calls public APIs, recording when the calls begin and end.  They use separate scope IDs (roughly these are Thread IDs) for the listener and the call due to no synchronization between them.  However, they both use the same PerfTags, which allows them to be associated.  

In the future, we can plumb the tag down into the stream to include deeper information about whats going on in a call.

### DIFF
--- a/core/BUILD.bazel
+++ b/core/BUILD.bazel
@@ -40,6 +40,7 @@ java_library(
     visibility = ["//:__subpackages__"],
     deps = [
         ":core",
+        ":perfmark",
         "//context",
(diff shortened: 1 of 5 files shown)


carl-mastrangelo:
(the CI should pick up binary size and method count differences from this PR as well.  I will be running the benchmarks shortly to see if there is any discernible overhead while disabled.)

PR #5531, opened 2019-04-01
buildscripts: add android-interop-test build (without runing test) in kokoro android

dapengzhang0:

PR #5538, opened 2019-04-03
kokoro: Add android-interop-test to android build

ejona86:
It is already being built as part of android-interop, but android-interop is
only run post-submit. We want to notice compilation errors before that point.

-----

Since the android-interop-test build is currently broken, this depends on #5537.


dapengzhang0:
Is this the same as #5531?

PR #5542, opened 2019-04-03
netty, alts: expose ProtectedPromise, and writeBufferedAndRemove methods

creamsoup:
ProtectedPromise will be used in frameProtector in PSP.
PSP also needs to call writeBufferedAndRemove.


alts/src/main/java/io/grpc/alts/internal/ProtectedPromise.java
```diff
@@ -18,6 +18,7 @@
 
 import static com.google.common.base.Preconditions.checkState;
 
+import io.grpc.Internal;
```
carl-mastrangelo: This is not sufficient.  You need to expose this via a class named Internal...

creamsoup: done. PTAL

netty/src/main/java/io/grpc/netty/WriteBufferingAndExceptionHandlerUtils.java
```diff
@@ -0,0 +1,40 @@
...
+
+package io.grpc.netty;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
+
+/** Utility class for {@link WriteBufferingAndExceptionHandler}. */
+public final class WriteBufferingAndExceptionHandlerUtils {
```
carl-mastrangelo: this needs to be hidden too

creamsoup: done.

netty/src/main/java/io/grpc/netty/NettyClientHandler.java
```diff
@@ -445,12 +445,7 @@ public void handleProtocolNegotiationCompleted(
     this.attributes = this.attributes.toBuilder().setAll(attributes).build();
     this.securityInfo = securityInfo;
     super.handleProtocolNegotiationCompleted(attributes, securityInfo);
-    // Once protocol negotiator is complete, release all writes and remove the buffer.
-    ChannelHandlerContext handlerCtx =
-        ctx().pipeline().context(WriteBufferingAndExceptionHandler.class);
-    if (handlerCtx != null) {
-      ((WriteBufferingAndExceptionHandler) handlerCtx.handler()).writeBufferedAndRemove(handlerCtx);
-    }
+    InternalWriteBufferingAndExceptionHandlerUtils.writeBufferingAndRemove(ctx().channel());
```
carl-mastrangelo: nit: This cannot call into an internal package.   Youll need to expose writeBufferingAndRemove as a package-private static on the NCH itself, and make this call into NCH

creamsoup: done.

PR #5532, opened 2019-04-01
xds: implement ADS request and response handling in standard mode

dapengzhang0:
Hi @zhangkun83 , I would like to get some review feedback on what I currently have first before sending out a complete version, because the latter might be hard to change course.

Summary of PR: 
- XdsLbState now assumes standard mode only.
- Will not send CDS request. A EDS request will be sent at the constructor of `AdsStream`.
- Added a method to `LocalityStore`
  - `void updateLocalityStore(Map<Locality, LocalityInfo> localityInfoMap);`
- When a EDS response is received. `LocalityStore.updateLocalityStore()` will be called.
- `LocalityStoreImpl` maintains a map `Map<Locality, LocalityLbInfo> localityMap`.
- `LocalityStoreImpl.updateLocalityStore()` will create a child balancer for each locality, with a `ChildHelper`. Then each child balancer will call `handleResolvedAddresses()`.
- `LocalityStoreImpl.updateLocalityStore()` will update `childPickers`.
- `ChildHelper.updateBalancingState()` will update `childPickers` and then delegate to parent `helper.updateBalancingState()`.
- `XdsLbState.handleSubchannelState()` will delegate to `childBalancer.handleSubchannelState()` where the subchannel belongs to the childBalancer's locality.

PR #5550, opened 2019-04-05
xds: add InterLocalityPicker

dapengzhang0:
This is the WrrCompositingPicker as suggested by https://github.com/grpc/grpc-java/pull/5532#discussion_r272397994

### DIFF
--- a/xds/src/main/java/io/grpc/xds/XdsPicker.java
+++ b/xds/src/main/java/io/grpc/xds/XdsPicker.java
@@ -19,120 +19,87 @@ package io.grpc.xds;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.common.base.MoreObjects;
-import com.google.common.base.Objects;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import io.grpc.ConnectivityState;
 import io.grpc.LoadBalancer.PickResult;
 import io.grpc.LoadBalancer.PickSubchannelArgs;
 import io.grpc.LoadBalancer.SubchannelPicker;
 import java.util.List;
 import java.util.Map;
-import javax.annotation.Nullable;
-import javax.annotation.concurrent.Immutable;
+import java.util.Random;
 
-abstract class XdsPicker extends SubchannelPicker {
+final class XdsPicker<LocalityT> extends SubchannelPicker {
 
-  protected final List<LocalityState> wrrList;
-  private final Map<Locality, SubchannelPicker> childPickers;
+  private final List<LocalityT> localities;
+  private final int[] accumulatedWeights;
+  private final Map<LocalityT, SubchannelPicker> childPickers;
+  private final Random random;
+  private final int totalWeight;
 
-  abstract Locality pickLocality();
+  XdsPicker(
+      List<LocalityT> localities,
+      List<Integer> weights,
+      Map<LocalityT, SubchannelPicker> childPickers) {
+    this(localities, weights, childPickers, new Random());
+  }
 
-  XdsPicker(List<LocalityState> wrrList, Map<Locality, SubchannelPicker> childPickers) {
-    checkArgument(!checkNotNull(wrrList, "wrrList in null").isEmpty(), "wrrList is empty");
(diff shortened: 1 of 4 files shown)


xds/src/main/java/io/grpc/xds/XdsPicker.java
```diff
@@ -0,0 +1,160 @@
...
+
+  @Override
+  public PickResult pickSubchannel(PickSubchannelArgs args) {
+    Locality locality = wrrAlgorithm.pickLocality(wrrList);
+    return childPickers.get(locality).pickSubchannel(args);
+  }
+
+  @VisibleForTesting // Interface of weighted Round-Robin algorithm that is convenient for test.
+  interface WrrAlgorithm {
+    Locality pickLocality(List<LocalityState> wrrList);
```
zhangkun83: This interface doesn't make sense. The wrrList should be passed once (for example at construction time).

xds/src/main/java/io/grpc/xds/XdsPicker.java
```diff
@@ -0,0 +1,160 @@
...
+import io.grpc.ConnectivityState;
+import io.grpc.LoadBalancer.PickResult;
+import io.grpc.LoadBalancer.PickSubchannelArgs;
+import io.grpc.LoadBalancer.SubchannelPicker;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+class XdsPicker extends SubchannelPicker {
```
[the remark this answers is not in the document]
dapengzhang0: abstract for now to test with fake wrr algorithm

xds/src/main/java/io/grpc/xds/XdsPicker.java
```diff
@@ -0,0 +1,160 @@
...
+  private final Map<Locality, SubchannelPicker> childPickers;
+
+  XdsPicker(List<LocalityState> wrrList, Map<Locality, SubchannelPicker> childPickers) {
+    this(wrrList, childPickers, wrrAlgo);
+  }
+
+  @VisibleForTesting
+  XdsPicker(
+      List<LocalityState> wrrList,
+      Map<Locality, SubchannelPicker> childPickers,
```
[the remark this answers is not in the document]
dapengzhang0: Done.

xds/src/main/java/io/grpc/xds/XdsPicker.java
dapengzhang0: Done.

xds/src/main/java/io/grpc/xds/XdsPicker.java
```diff
@@ -0,0 +1,105 @@
...
+
+    this.random = random;
+  }
+
+  @Override
+  public final PickResult pickSubchannel(PickSubchannelArgs args) {
+    LocalityT locality = null;
+
+    if (totalWeight == 0) {
+      locality = localities.get(random.nextInt(localities.size()));
```
zhangkun83: `Random` is not thread-safe. You need to use `ThreadLocalRandom`. Maybe define an interface `ThreadSafeRandom` with only `nextInt()`, and pass it in instead of `Random`.

dapengzhang0:
@zhangkun83 PTAL

xds/src/main/java/io/grpc/xds/XdsPicker.java
dapengzhang0: Done.

xds/src/main/java/io/grpc/xds/XdsPicker.java
```diff
@@ -0,0 +1,108 @@
...
+    checkNotNull(localities, "localities in null");
+    checkArgument(
+        !localities.isEmpty(),
+        "localities list is empty");
+    this.localities = ImmutableList.copyOf(localities);
+
+    checkNotNull(weights, "weights in null");
+    checkArgument(
+        weights.size() == localities.size(),
+        "localities and weights are of different sizes");
```
zhangkun83: It may be simpler to define a `WeightedLocality` class:
```java
class WeightedLocality {
  int getWeight();
  SubchannelPicker getPicker();
}
```
Then you wouldn't need most of the checks here and you wouldn't need the map either.

dapengzhang0:
@zhangkun83 Thanks for the suggests. PTALA.

PR #5554, opened 2019-04-06
interop-test: add test case for "pick_first" picking behavior

zhangkun83:
New fields are added to the test protos according to https://github.com/grpc/grpc-proto/pull/51

This test needs to be run in an environment where "pick_first" or "grpclb" with child policy "pick_first" is configured.

/cc @apolcyn

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -491,6 +491,26 @@ public abstract class AbstractInteropTest {
         Collections.singleton(goldenResponse));
   }
 
+  /**
+   * Assuming "pick_first" policy is used, tests that all requests are sent to the same server.
+   */
+  public void pickFirstUnary() throws Exception {
+    SimpleRequest request = SimpleRequest.newBuilder()
+        .setResponseSize(1)
+        .setFillServerId(true)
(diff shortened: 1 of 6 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -491,6 +491,26 @@ public void serverCompressedUnary() throws Exception {
...
+        .setResponseSize(1)
+        .setFillServerId(true)
+        .setPayload(Payload.newBuilder().setBody(ByteString.copyFrom(new byte[1])))
+        .build();
+
+    SimpleResponse firstResponse = blockingStub.unaryCall(request);
+    // Increase the chance of all servers are connected, in case the channel should be doing
+    // round_robin instead.
+    Thread.sleep(5000);
+    for (int i = 0; i < 100; i++) {
```
ejona86: 100 worries me a little bit, since it is tended to be assumed these are cheap tests and they can be run automatically lots of places. This is fine for now, but I might appreciate a followup later that reduces it once we are more certain things are working.

zhangkun83:
> Did you use buildscripts/sync-protos.sh to sync from grpc-proto?

No I didn't. I tried it but it doesn't seem to do anything.

zhangkun83:
`sync-protos.sh` did nothing because we have not been sync'ing `messages.proto`.

PR #5555, opened 2019-04-06
core: Migrate InternalSubchannel to use SynchronizedContext 

creamsoup:
one of the main motivation was trying to fix a deadlock in OkHttpClientTransport.
FYI. InProcessTransport.start returns Runnable, so can't remove the runnable.


core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -275,33 +246,19 @@ private void startNewTransport() {
...
-          syncContext.drain();
-        }
+        reconnectTask = null;
+        channelLogger.log(ChannelLogLevel.INFO, "CONNECTING after backoff");
+        gotoNonErrorState(CONNECTING);
+        startNewTransport();
       }
     }
 
+    syncContext.throwIfNotInThisSynchronizationContext();;
```
zhangkun83: Move it to the beginning of the method?

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -313,20 +270,21 @@ public void run() {
         "TRANSIENT_FAILURE ({0}). Will reconnect after {1} ns",
         printShortStatus(status), delayNanos);
     Preconditions.checkState(reconnectTask == null, "previous reconnectTask is not done");
-    reconnectCanceled = false;
-    reconnectTask = scheduledExecutor.schedule(
+    reconnectTask = syncContext.schedule(
         new LogExceptionRunnable(new EndOfCurrentBackoff()),
         delayNanos,
-        TimeUnit.NANOSECONDS);
+        TimeUnit.NANOSECONDS,
```
zhangkun83: We probably don't need to use `LogExceptionRunnable` anymore. It was necessary because the runnable was scheduled on `ScheduledExecutorService` which doesn't handle exceptions as we would like to (6a04022d7aceb73d9cf8096b107930eac0f6c477)

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -335,43 +293,43 @@ void resetConnectBackoff() {
...
-      syncContext.drain();
-    }
+    });
   }
 
-  @GuardedBy("lock")
-  private void gotoNonErrorState(ConnectivityState newState) {
-    gotoState(ConnectivityStateInfo.forNonError(newState));
+  private void gotoNonErrorState(final ConnectivityState newState) {
+    syncContext.execute(new Runnable() {
```
zhangkun83: `gotoNonErrorState()` is only run from synContext. Why is this necessary?

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -555,10 +492,10 @@ private static void checkListHasNoNulls(List<?> list, String msg) {
...
     public void transportReady() {
       channelLogger.log(ChannelLogLevel.INFO, "READY");
-      Status savedShutdownReason;
-      try {
-        synchronized (lock) {
-          savedShutdownReason = shutdownReason;
+      syncContext.execute(new Runnable() {
+        @Override
+        public void run() {
+          Status savedShutdownReason = shutdownReason;
```
zhangkun83: We can safely use `shutdownReason` directly within this scope.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
creamsoup: iirc, this was hard to figure out. the SyncContext execution order is different than lock. because it doesn't allow the re-entrance.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
creamsoup: done.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -156,10 +133,8 @@ protected void handleNotInUse() {
   @Nullable
```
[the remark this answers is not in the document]
creamsoup: done.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -79,53 +77,33 @@
...
-  // 3. Every synchronized("lock") must be inside a try-finally which calls drain() in "finally".
   private final SynchronizationContext syncContext;
 
   /**
    * The index of the address corresponding to pendingTransport/activeTransport, or at beginning if
    * both are null.
    */
-  @GuardedBy("lock")
-  private Index addressIndex;
+  private volatile Index addressIndex;
```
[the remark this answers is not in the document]
creamsoup: good point. if we adapt more SynchronizedContext, it would be nice to have something like GuardedBy.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
creamsoup: right. syncContext is already handling exception.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
creamsoup: yes thanks.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
zhangkun83: Did it cause any problem? Calling inline is closer to the previous behavior under lock.

creamsoup: yes, 3 tests are failing (io.grpc.internal.ManagedChannelImplTest.{channelsAndSubchannels_oob_instrumented_fail,oobchannels,channelsAndSubchannels_oob_instrumented_success}.

zhangkun83: These tests fail because `transportReady()` delivers a `READY` state to the listener before updating  `activeTransport`. `OobChannel` calls `DelayedClientTransport.reprocess()` as soon as receiving a `READY`, which will fail because `activeTransport` is not yet set. Previous code got away with this issue because it called `syncContext.drain()` after `activeTransport` is set, but the root cause, which is transitioning to a new state prematurely, was already there. Using `syncContext.excute()` inside `gotoNonErrorState()` only covers up the root cause.

The correct implementation is to transition to a new state **after** all necessary state changes have been made. If you move all `gotoNonErrorState()` and `gotoState()` calls to after their prerequisite state changes, those tests will pass.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -80,52 +78,36 @@
...
+   */
   private final SynchronizationContext syncContext;
 
   /**
    * The index of the address corresponding to pendingTransport/activeTransport, or at beginning if
    * both are null.
    */
-  @GuardedBy("lock")
-  private Index addressIndex;
+  private volatile Index addressIndex;
```
ejona86: Why is this volatile? It is never changed. It seems there are reads from the index without synchronization: tostring, getAddressGroups (just missing throwIfNotInThisSynchronizationContext()?). That needs to be resolved.

creamsoup: @zhangkun83 and I discussed this before, this was tricky one iirc. Those two methods are called from elsewhere (getAddressGroups is called from ManagedChannelImpl.SubchannelImpl and OobChannel). if it returns Future, we need to create a thread otherwise it will hang. Slightly delay for the getAddressGroups looked acceptable according to Kun. @zhangkun83, can you explain again?

zhangkun83: My fault. Making `addressIndex` volatile doesn't help. It should in fact be final.
`Index.addressGroups` can be made volatile to help `toString()` and `getAddressGroups()`. They can't have `throwIfNotInThisSynchronizationContext()` because they didn't have such requirement when they were first introduced.

ejona86: `Index.addressGroups` being volatile would fix it, although may be a bit confusing to see all the unnecessary volatile reads (making you think synchronization is going on). We could also keep another field that the same as `Index.addressGroups` except being volatile. In can be in InternalSubchannel (vs Index) if we'd like.

carl-mastrangelo: @zhangkun83 that's definitely worth a comment.

zhangkun83: +1 for having a `volatile addressGroups` in `InternalSubchannel`.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -86,8 +86,19 @@
   /**
    * The index of the address corresponding to pendingTransport/activeTransport, or at beginning if
    * both are null.
+   *
+   * <p>Note: any {@link Index#updateAddresses(List)} should also update {@link #addressGroups}.
    */
-  private volatile Index addressIndex;
+  private Index addressIndex;
```
zhangkun83: This should be final.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
creamsoup: yeah just added the volatile accessor for `addressGroups`, PTAL.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -86,8 +86,19 @@
...
+
+  /**
+   * A volatile accessor to {@link Index#getAddressGroups()}. There are few methods ({@link
+   * #getAddressGroups()} and {@link #toString()} access this value where they supposed to access
+   * in the {@link #syncContext}. Ideally {@link Index#getAddressGroups()} can be volatile, so we
+   * don't need to maintain this volatile accessor. Although, having this accessor can reduce
```
ejona86: For posterity (it is fine to leave as-is): for me it was more of an argument for readability. With this approach the intent of the code is clearer.

PR #5561, opened 2019-04-08
core: make the NameResolver.start() change backward compatible for callers

zhangkun83:
This will give time for pre-existing external callers (e.g.,
forwarding NameResolvers) to migrate to the new start().


zhangkun83:
I was wrong about the milestone. The original API change was not in 1.20 thus this change doesn't need to be backported.

PR #5566, opened 2019-04-09
services: Remove dependency on re2j

ejona86:
re2j is a fairly unnecessary dependency. Our usage of Pattern is quite limited
and isn't all that hard to do manually. Our usage would be safe to use normal
java.util.regex, but it's sort of annoying to keep re-explaining to others who
are (rightly) concerned with java.util.regex's poor pathological behavior.


anuraaga:
Thanks a lot for this :)

PR #5572, opened 2019-04-09
context: try out static final storage

carl-mastrangelo:
I'm not totally convinced this is better than previously, but I believe it avoids the class initialization problem.  

The performance difference is minimal, but it does avoid the extra stack frame, and avoids the atomic compiler barrier (cheap).   storage() is called quite frequently, so there may be an aggregate benefit to making this fast.

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -116,42 +116,42 @@ public class Context {
    */
   public static final Context ROOT = new Context(null, EMPTY_ENTRIES);
 
+  // Visible For testing
+  static Storage storage() {
+    return LazyStorage.storage;
+  }
+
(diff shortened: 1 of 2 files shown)


context/src/main/java/io/grpc/Context.java
```diff
@@ -116,42 +116,42 @@
...
-    Storage tmp = storage.get();
-    if (tmp == null) {
-      tmp = createStorage();
+  private static final class LazyStorage {
+    static final Storage storage;
+
+    static {
+      AtomicReference<Throwable> deferredStorageFailure = new AtomicReference<>();
+      storage = createStorage(deferredStorageFailure);
+      Throwable failure;
```
ejona86: `Throwable failure = deferredStorageFailure.get();` ?

carl-mastrangelo: Done.

PR #5580, opened 2019-04-11
core: declare Metadata reads are threadsafe

carl-mastrangelo:
Fixes b/118299078


ejona86:
I'm still not certain we want this.

carl-mastrangelo:
What is making you uncertain?

PR #5581, opened 2019-04-11
netty: change default transport to Epoll if available, otherwise using Nio

creamsoup:
Motivation:
To support TCP_USER_TIMEOUT([proposal](https://github.com/grpc/proposal/blob/master/A18-tcp-user-timeout.md)). Nio doesn't support TCP_USER_TIMEOUT while Epoll and KQueue supports TCP_USER_TIME. Since most users/servers are using linux based system, adding Epoll is necessary. KQueue maybe supported later, but not in this PR.

To make it backward compatible, for cases when channelType or eventLoop is mixed in with default and user provided object(s). We will fallback to Nio when not all the values (channelType, eventLoop) are provided. Which should be same as current behavior. In later version (possibly 1.22.0) will throw exception since this is error prone.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
+++ b/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
@@ -35,9 +35,11 @@ import io.grpc.internal.AbstractManagedChannelImplBuilder;
 import io.grpc.internal.AtomicBackoff;
 import io.grpc.internal.ClientTransportFactory;
 import io.grpc.internal.ConnectionClientTransport;
+import io.grpc.internal.FixedObjectPool;
 import io.grpc.internal.GrpcUtil;
 import io.grpc.internal.KeepAliveManager;
-import io.grpc.internal.SharedResourceHolder;
+import io.grpc.internal.ObjectPool;
+import io.grpc.internal.SharedResourcePool;
 import io.grpc.internal.TransportTracer;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFactory;
@@ -65,14 +67,16 @@ import javax.net.ssl.SSLException;
 @CanIgnoreReturnValue
 public final class NettyChannelBuilder
     extends AbstractManagedChannelImplBuilder<NettyChannelBuilder> {
-  public static final Logger logger = Logger.getLogger(NettyChannelBuilder.class.getName());
+  private static final Logger logger = Logger.getLogger(NettyChannelBuilder.class.getName());
 
   public static final int DEFAULT_FLOW_CONTROL_WINDOW = 1048576; // 1MiB
 
   private static final long AS_LARGE_AS_INFINITE = TimeUnit.DAYS.toNanos(1000L);
 
-  private static final ChannelFactory<? extends Channel> DEFAULT_CHANNEL_FACTORY
-      = new ReflectiveChannelFactory<>(Utils.defaultClientChannelType());
+  private static final ChannelFactory<? extends Channel> DEFAULT_CHANNEL_FACTORY =
+      new ReflectiveChannelFactory<>(Utils.defaultClientChannelType());
+  private static final ObjectPool<? extends EventLoopGroup> DEFAULT_EVENT_LOOP_GROUP_POOL =
+      SharedResourcePool.forResource(Utils.DEFAULT_WORKER_EVENT_LOOP_GROUP);
 
   private final Map<ChannelOption<?>, Object> channelOptions =
       new HashMap<>();
@@ -80,8 +84,7 @@ public final class NettyChannelBuilder
   private NegotiationType negotiationType = NegotiationType.TLS;
   private OverrideAuthorityChecker authorityChecker;
   private ChannelFactory<? extends Channel> channelFactory = DEFAULT_CHANNEL_FACTORY;
-  @Nullable
-  private EventLoopGroup eventLoopGroup;
+  private ObjectPool<? extends EventLoopGroup> eventLoopGroupPool = DEFAULT_EVENT_LOOP_GROUP_POOL;
   private SslContext sslContext;
   private int flowControlWindow = DEFAULT_FLOW_CONTROL_WINDOW;
   private int maxHeaderListSize = GrpcUtil.DEFAULT_MAX_HEADER_LIST_SIZE;
@@ -145,18 +148,18 @@ public final class NettyChannelBuilder
   }
 
   /**
-   * Specifies the channel type to use, by default we use {@link EpollSocketChannel} if available,
+   * Specifies the channel type to use, by default we use {@code EpollSocketChannel} if available,
    * otherwise using {@link NioSocketChannel}.
    *
    * <p>You either use this or {@link #channelFactory(io.netty.channel.ChannelFactory)} if your
    * {@link Channel} implementation has no no-args constructor.
    *
+   * <p>It's an optional parameter. If the user has not provided an Channel type or ChannelFactory
+   * when the channel is built, the builder will use the default one which is static.
+   *
    * <p>You must also provide corresponding {@link #eventLoopGroup(EventLoopGroup)}. For example,
    * {@link NioSocketChannel} must use {@link io.netty.channel.nio.NioEventLoopGroup}, otherwise
    * your application won't start.
-   *
-   * <p>It's an optional parameter. If the user has not provided an Channel type or ChannelFactory
-   * when the channel is built, the builder will use the default one which is static.
    */
   public NettyChannelBuilder channelType(Class<? extends Channel> channelType) {
     checkNotNull(channelType, "channelType");
@@ -214,7 +217,11 @@ public final class NettyChannelBuilder
    * to shut it down when it's desired.
    */
   public NettyChannelBuilder eventLoopGroup(@Nullable EventLoopGroup eventLoopGroup) {
-    this.eventLoopGroup = eventLoopGroup;
+    if (eventLoopGroup != null) {
+      this.eventLoopGroupPool = new FixedObjectPool<>(eventLoopGroup);
+    } else {
+      this.eventLoopGroupPool = DEFAULT_EVENT_LOOP_GROUP_POOL;
+    }
     return this;
   }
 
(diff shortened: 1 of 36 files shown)


creamsoup:
i am not sure why gradle spotless plugin starts to complain about gradle build file check

netty/src/main/java/io/grpc/netty/Utils.java
```diff
@@ -176,6 +184,78 @@ public static Status statusFromThrowable(Throwable t) {
     return s;
   }
 
+  /**
+   * Returns a {@link ChannelFactory} of default channel; {@code
+   * io.netty.channel.epoll.EpollSocketChannel} when Epoll is available, otherwise using {@link
+   * NioSocketChannel}.
+   */
+  @SuppressWarnings("unchecked")
+  static ChannelFactory<? extends Channel> defaultChannelFactory() {
```
ejona86: Instead of splitting the `isEpollAvailable()` logic between each of these methods, I'd much rather it was checked and processed once or twice (for client vs server). We also don't want to be doing the reflection every Channel creation.

It'd probably be best/easiest to put the logic in a `static` block.

netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -110,7 +110,7 @@
       boolean permitKeepAliveWithoutCalls, long permitKeepAliveTimeInNanos,
       InternalChannelz channelz) {
     this.address = address;
-    this.channelType = checkNotNull(channelType, "channelType");
+    this.channelType = channelType == null ? Utils.defaultServerChannel() : channelType;
```
ejona86: Why not initialize `channelType` to `defaultServerChannel()`?

netty/src/main/java/io/grpc/netty/Utils.java
```diff
@@ -176,6 +184,78 @@ public static Status statusFromThrowable(Throwable t) {
...
+    if (isEpollAvailable()) {
+      try {
+        Class<Channel> channel =
+            (Class<Channel>) Class.forName("io.netty.channel.epoll.EpollSocketChannel");
+        logger.log(Level.FINE, "Using EpollSocketChannel");
+        return new ReflectiveChannelFactory<>(channel);
+      } catch (ClassNotFoundException e) {
+        logger.log(Level.INFO, "netty-epoll is not available (this may be normal)");
```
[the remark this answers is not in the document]
creamsoup: done.

netty/src/main/java/io/grpc/netty/NettyServer.java
creamsoup: no longer relevant, but i think the channelType in the server is final.  default value for eventloop was handled here. which is moved to the builder in newer commit. PTAL

netty/src/main/java/io/grpc/netty/Utils.java
creamsoup: done.

netty/src/main/java/io/grpc/netty/Utils.java
```diff
@@ -65,11 +72,50 @@
...
+  public static final Resource<EventLoopGroup> NIO_BOSS_EVENT_LOOP_GROUP
+      = new DefaultEventLoopGroupResource(1, "grpc-nio-boss-ELG", NioEventLoopGroup.class);
+  public static final Resource<EventLoopGroup> NIO_WORKER_EVENT_LOOP_GROUP
+      = new DefaultEventLoopGroupResource(0, "grpc-nio-worker-ELG", NioEventLoopGroup.class);
+  public static final Resource<EventLoopGroup> DEFAULT_BOSS_EVENT_LOOP_GROUP;
+  public static final Resource<EventLoopGroup> DEFAULT_WORKER_EVENT_LOOP_GROUP;
 
-  public static final Resource<EventLoopGroup> DEFAULT_WORKER_EVENT_LOOP_GROUP =
-      new DefaultEventLoopGroupResource(0, "grpc-default-worker-ELG");
+  private static final Class<? extends ServerChannel> defaultServerChannelType;
```
ejona86: DEFAULT_SERVER_CHANNEL_TYPE? Ditto next line.

netty/src/main/java/io/grpc/netty/Utils.java
```diff
@@ -65,11 +72,50 @@
...
+
+    try {
+      if (isEpollAvailable()) {
+        serverChannelType = epollServerChannelType();
+        clientChannelType = epollChannelType();
+        // load EventLoopGroup to make sure all defaults are using same type.
+        eventLoopGroupType = epollEventLoopGroupType();
+        epollSuccessful = true;
+      }
+    } catch (Exception e) {
```
ejona86: Log this. It looks like it should be a WARNING, since the code above should not throw.

Although to make the code more obvious, we could really just let the exceptions propagate (again, since it shouldn't throw). That would allow you to assign directly to the `final` fields, as if something bad happens you don't need to worry about assigning the rest or getting into a weird state (because the program will crash).

netty/src/main/java/io/grpc/netty/Utils.java
```diff
@@ -176,20 +222,109 @@ public static Status statusFromThrowable(Throwable t) {
...
+  }
+
+  // Must call when epoll is available
+  private static Class<? extends ServerChannel> epollServerChannelType() {
+    try {
+      Class<? extends ServerChannel> serverSocketChannel =
+          Class
+              .forName("io.netty.channel.epoll.EpollServerSocketChannel")
+              .asSubclass(ServerChannel.class);
+      logger.log(Level.FINE, "Using EpollServerSocketChannel");
```
ejona86: Remove this log? It seems a bit weird in this context.

netty/src/main/java/io/grpc/netty/Utils.java
```diff
@@ -176,20 +222,109 @@ public static Status statusFromThrowable(Throwable t) {
...
+      boolean available = (Boolean)
+          Class
+              .forName("io.netty.channel.epoll.Epoll")
+              .getDeclaredMethod("isAvailable")
+              .invoke(null);
+      if (!available) {
+        logger.log(Level.FINE, "Epoll is not available", getEpollUnavailabilityCause());
+      }
+      return available;
+    } catch (Exception e) {
```
ejona86: ClassNotFound should probably be special-cased, since that is quite normal (although it would still be fair to log it at FINE). Otherwise this exception should be logged, probably at WARNING.

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -406,11 +430,40 @@ protected ClientTransportFactory buildTransportFactory() {
...
+
+  private void populateDefaultChannelAndGroup() {
+    boolean groupOrChannelProvided = channelFactory != DEFAULT_CHANNEL_FACTORY
+        || eventLoopGroup != null;
+
+    if (groupOrChannelProvided) {
+      logger.log(
+          Level.WARNING,
+          "Both EventLoopGroup and ChannelType should be provided, otherwise client may not start "
+              + "depends on the gRPC default value.");
```
ejona86: The sentence is broken between "start" and "depends".

Include in the message that we are falling back to Nio for compatibility, and that this will cause an exception in the future.

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -406,11 +430,40 @@ protected ClientTransportFactory buildTransportFactory() {
...
+
+      if (eventLoopGroup == null) {
+        eventLoopGroup = SharedResourceHolder.get(Utils.NIO_WORKER_EVENT_LOOP_GROUP);
```
ejona86: We shouldn't be mutating the builder. That can cause quite a bit of confusion if `build()` is called twice. For example, the way this code is, the second call to build() would not acquire a reference to the event loop and so it may be shut down while still in use.

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -406,11 +430,40 @@ protected ClientTransportFactory buildTransportFactory() {
...
+        logger.log(Level.FINE, String.format("Using fall back EventLoopGroup %s", eventLoopGroup));
```
ejona86: This will always call String.format(), even if not logging at FINE level. That should be avoided. Ditto next log.

netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -108,17 +107,17 @@
       long maxConnectionIdleInNanos,
       long maxConnectionAgeInNanos, long maxConnectionAgeGraceInNanos,
       boolean permitKeepAliveWithoutCalls, long permitKeepAliveTimeInNanos,
-      InternalChannelz channelz) {
+      InternalChannelz channelz, boolean usingSharedBossGroup, boolean usingSharedWorkerGroup) {
```
ejona86: Having these bools passed in this deep can be sort of annoying. You may consider passing in ObjectPools instead. That way this code wouldn't need any conditionals and some of the handling elsewhere looks like it would improve.

njhill:
Suggestion: Why not just have a fixed runtime dependency on `netty-transport-native-epoll` from `grpc-netty`? Without this, the benefit is kind of limited in that many users won't "automatically" use epoll. I expect many of those who would know/think to add the dependency would already know to customize their server/channel configuration. Yes it is one additional library pulled in for everyone which might not be used, but:
- It's not very big
- There are already other hardwired, possibly less-used libs such as the opencensus apis, and this would only be there if they already depend on `grpc-netty`, unlike those which have a dep from `grpc-core`
- I assume the majority of production deployments run on linux x86_64 anyhow

An additional minor benefit would be to remove the need for reflection. Also, there is no harm in this being there if they are running on non linux-x84_64 platforms. In fact [as questioned a while back](https://github.com/netty/netty/issues/8097), using the os selector for this [in the build](https://github.com/grpc/grpc-java/blob/5b2187d6bd6a03cd0dfe09d1505ca88d86e01ab8/build.gradle#L188) seems incorrect - the build env in general is arbitrary in relation to the runtime platform.

netty/src/main/java/io/grpc/netty/Utils.java
creamsoup: oops done.

netty/src/main/java/io/grpc/netty/NettyServer.java
creamsoup: done. it looks much simpler

ejona86:
@njhill, hmm... We were thinking most users should be using `grpc-netty-shaded`, which means they would get what they need.

I agree that it may make sense to depend on native-epoll for the Java classes. And I didn't realize the native transport was so small. Although we may also want kqueue support. It seems both of them depend on netty-unix-common, which is fairly small. There's no "uber" jar for it though. It looks like the linux/osx versions of netty-unix-common isn't for normal use, because it only contains a `.a`?

Hmm... I'm not sure how Maven/Gradle handle mixing dependencies with different classifiers.

I think I'm favoring this current more conservative approach for the moment. We've seen small performance gains from epoll; we're mainly doing this for better socket option support when available, like TCP_USER_TIMEOUT. Once we see the world doesn't fall over, we might make it a stronger dependency; and go through the effort of making sure Maven/Gradle will behave in a useful way.

ejona86:
@creamsoup, labeling this PR as "all" seems misleading. There's literally a single line that isn't in the `netty` module. I'd suggest just labeling it "netty", as it also makes the rest of the description more clear.

The PR should also include some sort of description as to _why_ it is being made.

ejona86:
@creamsoup, @njhill brought up a good point about auto-detection logic. We don't want any auto-detection logic for our epoll dependency from grpc-netty-shaded. We can probably just always use the "linux-x86_64" classifier; see the [main build.gradle](https://github.com/grpc/grpc-java/blob/v1.20.0/build.gradle#L188).

njhill:
@ejona86 `netty-transport-unix-common` is a common dep of epoll and kqueue but itself doesn't contain any native libs and the artifact doesn't have a classifier.  What might be misleading (it confused me in the past) is the existence of the non-classifier `netty-transport-native-epoll` jar. This is identical to the `linux-x86_64` one minus the single `.so` file, and as such kind of useless. I guess if there was a hard dep but you weren't using epoll you could save 24K by using it instead (i.m.o. this isn't worth the confusion it introduces where folks might think they are using epoll by including the dep, even though it won't work without the classifier specified).

So I think it wouldn't cause problems to similarly depend on kqueue in addition to epoll (no problem having diff classifiers for diff artifacts), though I would expect it's rarer for performance to be such a concern in macos contexts.

ejona86:
@njhill, `netty-transport-unix-common` [does have platform-specific classifier versions](https://search.maven.org/remotecontent?filepath=io/netty/netty-transport-native-unix-common/4.1.34.Final/netty-transport-native-unix-common-4.1.34.Final-linux-x86_64.jar). I was saying though that it doesn't look like you need to use them; they just contain platform-specific goo used during compilation of other modules (like `/META-INF/native/lib/libnetty-unix-common.a`).

My bringing up of Kqueue is that it is another "useless dependency" for certain people. It's also small, though. Such small unnecessary dependencies could potentially add up, but they do seem minor compared to the rest of the netty artifacts (gosh it is big).

njhill:
@ejona86 apologies, you are right. But yes doesn't seem they are used at runtime.

netty/src/main/java/io/grpc/netty/Utils.java
```diff
@@ -65,11 +72,34 @@
...
+  public static final Resource<EventLoopGroup> NIO_BOSS_EVENT_LOOP_GROUP
+      = new DefaultEventLoopGroupResource(1, "grpc-nio-boss-ELG", NioEventLoopGroup.class);
+  public static final Resource<EventLoopGroup> NIO_WORKER_EVENT_LOOP_GROUP
+      = new DefaultEventLoopGroupResource(0, "grpc-nio-worker-ELG", NioEventLoopGroup.class);
+  public static final Resource<EventLoopGroup> DEFAULT_BOSS_EVENT_LOOP_GROUP;
+  public static final Resource<EventLoopGroup> DEFAULT_WORKER_EVENT_LOOP_GROUP;
 
-  public static final Resource<EventLoopGroup> DEFAULT_WORKER_EVENT_LOOP_GROUP =
-      new DefaultEventLoopGroupResource(0, "grpc-default-worker-ELG");
+  private static final Class<? extends ServerChannel> DEFAULT_SERVER_CHANNEL_TYPE;
```
ejona86: Make these two public and get rid of the accessor methods?

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -406,13 +437,46 @@ protected ClientTransportFactory buildTransportFactory() {
...
       negotiator = createProtocolNegotiatorByType(negotiationType, localSslContext);
     }
+
+    // TODO(jihuncho) throw exception if not groupOrChannelProvided after 1.22.0
+    ObjectPool<? extends EventLoopGroup> resolvedEventLoopGroupPool = eventLoopGroupPool;
+    ChannelFactory<? extends Channel> resolvedChannelFactory = channelFactory;
+    if (shouldFallBackToNio()) {
+      logger.log(
+          Level.WARNING,
+          "Both EventLoopGroup and ChannelType should be provided, otherwise client may not start. "
```
ejona86: "Both EventLoopGroup and ChannelType should be provided or neither should be, otherwise the client..."

Ditto for server.

netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -50,6 +50,7 @@
 import io.netty.util.ReferenceCounted;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
+
```
ejona86: Remove.

ejona86:
@njhill, I may not have made it clear: I'm really happy you pointed out that the epoll jars were small. Although we'll use reflection for the moment and we need to do some more research into Maven/Guava behavior, I will likely try to swap us to a strong dependency in the future to avoid the reflection.

netty/src/main/java/io/grpc/netty/Utils.java
creamsoup: done.

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
creamsoup: thanks, done.

netty/src/main/java/io/grpc/netty/NettyServer.java
creamsoup: done.

PR #5586, opened 2019-04-12
core: Add NameResolverRegistry

### LINKED ISSUE #5562: Add support for custom attributes in NameResolver.Helper

1.19

### Related

* Based on https://github.com/grpc/grpc-java/issues/5556#issuecomment-480953694

### Feature-Request

The java service loader API has the drawback that it creates instances without context.
In my case it creates [`NameResolverProvider`](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/NameResolverProvider.java)s without spring's dependency injection.
With the Attributes API that present up to and including 1.18, I was able to pipe the potentially required beans/config to the actual implementations. Now this path is blocked.

call-hierarchy

Lets assume we have a config which contains the "service discovery server URL".
Somehow we need to push this tiny piece of runtime only information into the ServiceDiscoveryNameResolver(Factory).

We now (1.19 or later) have basically three options:

1. Encode everything in the URI
   This will soon become a nightmare for every maintainer as more and more properties needs to be encoded for the potential case that one provider needs them. This option cannot pass more complex instances such as a `RestTemplate` or `HttpRequestOptions`.
2. Use static fields to pass the variables
   Well you can pass anything through static variables, but using them to transfer something stateful sounds like a bad idea and fails as soon as the load order is slightly of or multiple distinct instances of gprc-services run in the same JVM (to save some memory).
3. Wrap the NameResolverProvider (registry) in another (spring-managed) registry that will first try to use the contextual name resolvers before falling back to the grpc-managed ones. Well this works but somewhat defeats the purpose of specifying priorities and availability in the first place.

Prior to 1.19 we also had a forth option:

4. Pass the context through the attributes

In [yidongnan/grpc-spring-boot-starter](https://github.com/yidongnan/grpc-spring-boot-starter/) I use the third variant, but also considered using the forth variant because it would free my code from [conditional wrappers](https://github.com/yidongnan/grpc-spring-boot-starter/blob/master/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcDiscoveryClientAutoConfiguration.java#L49) and would allow adding other contextual name resolvers more easily.

**ConfigMappedNameResolverFactory** (springNameResolver)

````java
    @Nullable
    @Override
    public NameResolver newNameResolver(final URI targetUri, final Attributes params) {
        final String clientName = targetUri.toString();
        final GrpcChannelProperties clientConfig = this.config.getChannel(clientName);
        URI remappedUri = clientConfig.getAddress();
        if (remappedUri == null) {
            remappedUri = this.defaultUriMapper.apply(clientName);
            if (remappedUri == null) {
                throw new IllegalStateException("No targetUri provided for '" + clientName + "'"
                        + " and defaultUri mapper returned null.");
            }
        }
        log.debug("Remapping target URI for {} to {} via {}", clientName, remappedUri, this.delegate);
        final Attributes extendedParas = params.toBuilder() // Enrich attributes with the context
                .set(NameResolverConstants.PARAMS_CLIENT_NAME, clientName)
                .set(NameResolverConstants.PARAMS_CLIENT_CONFIG, clientConfig)
                // .set(NameResolverConstants.PARAMS_APPLICATION_CONTEXT, applicationContext) // Full spring context
                .build();
        return this.delegate.newNameResolver(remappedUri, extendedParas);
    }
````

I currently don't use the extra attributes, but I considered it for future simplifications.

The [`DiscoveryClientResolverFactory`](https://github.com/yidongnan/grpc-spring-boot-starter/blob/master/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientResolverFactory.java#L68) is currently my only contextual name resolver. In this example I could request the `DiscoveryClient` bean from the `applicationContext` instead of passing it in the constructor.

IMO it would be best to revert the helper back to the unspecific attributes or add an optional, mutable and unspecific attributes to the helper because:
* A single `getCustomAttributes() : Object` would fail as soon as another layer of context is added on top.
* At least the default port is useless when other service discovery mechanisms are used. I'm not sure whether I would call it an implementation detail but at least it's related.
* There is no factory method/forwarder class to overwrite one value with another without risking to become fragile to changes to the helper class which seems to be the case for 1.21 again.

### PR DESCRIPTION
**Pull request by ejona86**
NameResolverRegistry takes on all the logic previously in
NameResolverProvider. But it also allows manual registration of
NameResolvers, which is useful when the providers have complex
construction or need objects injected into them.

This also avoids a circular dependency during class loading since
previously loading any Provider searched for all Providers via
ClassLoader since ClassLoader handling was static within the parent
class.

Fixes #5562

CC @ST-DDT

### CODE REVIEW

File: core/src/test/java/io/grpc/NameResolverRegistryTest.java [hunk 1]
```diff
@@ -0,0 +1,209 @@
...
+
+import io.grpc.internal.DnsNameResolverProvider;
+import java.net.URI;
+import java.util.List;
+import org.junit.After;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Unit tests for {@link NameResolverProvider}. */
```
ST-DDT: Maybe you forgot to update the test javadocs.

ejona86: Yep. Fixed.

core/src/main/java/io/grpc/NameResolverRegistry.java
```diff
@@ -0,0 +1,183 @@
...
+    refreshProviders();
+  }
+
+  private synchronized void refreshProviders() {
+    List<NameResolverProvider> providers = new ArrayList<NameResolverProvider>(allProviders);
+    // sort() must be stable, as we prefer first-registered providers
+    Collections.sort(providers, new Comparator<NameResolverProvider>() {
+      @Override
+      public int compare(NameResolverProvider p1, NameResolverProvider p2) {
+        return p1.priority() - p2.priority();
```
zhangkun83: I was under the impression that higher priority should be used first, and this is how `LoadBalancerRegistry` does.

ejona86: I was originally going to use the comparator in ServiceProviders. As things changed around I ended up missing a Collections.reverseOrder. I had forgotten higher preferred, so didn't notice in the tests. Fixed.

core/src/main/java/io/grpc/NameResolverRegistry.java
```diff
@@ -83,12 +83,12 @@ public synchronized void deregister(NameResolverProvider provider) {
   private synchronized void refreshProviders() {
     List<NameResolverProvider> providers = new ArrayList<NameResolverProvider>(allProviders);
     // sort() must be stable, as we prefer first-registered providers
-    Collections.sort(providers, new Comparator<NameResolverProvider>() {
+    Collections.sort(providers, Collections.reverseOrder(new Comparator<NameResolverProvider>() {
```
zhangkun83: Why not just reverse the expression in the body?

ejona86: The comparator is the natural ordering. It can be subtle to notice a - b vs b - a. Using reverseOrder makes it more explicit.

ejona86:
Rebased past grpc-api creation

core/src/main/java/io/grpc/NameResolverRegistry.java
ST-DDT: Maybe add a comment here that sums it up.
````java
// Higher values/priorities first
````
IMO thats way more expressive than a reverse order somewhere in between.

ejona86: Added a comment (mirroring what is in ServiceProviders). My concern here is that the comment could be wrong... In fact, I considered a comment before and it _would_ have been wrong because I lost the reverseOrder during refactoring.

Issue #5359, opened 2019-02-13
bazel: migrate to deps of maven_install


ejona86:
jvm_maven_import_external (and jvm_import_external) allow providing `deps` which will be added to the java_import in the generated workspace. You can refer to the target that includes deps via `@com_google_guava_guava` instead of `@com_google_guava_guava//jar`.

Unfortunately, if we use the non-//jar targets, they aren't available with maven_jar so it'd break users that haven't migrated from maven_jar to jvm_import_external/jvm_maven_import_external. See also #5214

We need to notify our users we will require the deps to be provided appropriately and give them some time to migrate. This includes dependent tools, like https://github.com/johnynek/bazel-deps.

ejona86:
The rules_jvm_external project provides a maven_install rule, which [Bazel published about on their blog](https://blog.bazel.build/2019/03/31/rules-jvm-external-maven.html).

I was uncertain how we'd upgrade to it, as it has _yet another_ way of referencing dependencies. However, it's since added the `generate_compat_repositories` option which generates compatibility workspaces/labels for the old-style. This provides a clear upgrade path. We swap to using maven_install in our grpc_java_repositories while continuing to use the old label. Once the ecosystem as a whole is on the new-style system we swap to the new labels.

ejona86:
There's [pretty clear documentation now](https://github.com/bazelbuild/rules_jvm_external#exporting-and-consuming-artifacts-from-external-repositories) on how libraries like us should interact with maven_install(). We should add the new `*_ARTIFACTS` symbol for users to use with maven_install and rely on `generate_compat_repositories` until drop support for the old style in `grpc_java_repositories()`.

We will also [need to define a map for `override_targets`](https://github.com/census-instrumentation/opencensus-java/issues/1963#issuecomment-522696875).

PR #5590, opened 2019-04-13
core: Move io.grpc to grpc-api

### LINKED ISSUE #1447: Create a grpc-api artifact
After #1446 we could make a grpc-api artifact that does not include `io.grpc.internal`. From the other issue:

> It would be useful to be able to depend on the API but not the implementation to not pull in unnecessary dependencies and provide cleaner builds that don't accidentally depend on the implementation. For instance, a library could have optional gRPC support, but depend on grpc-api even when grpc was not selected.

ejona86:
io.grpc has fewer dependencies than io.grpc.internal. Moving it to a
separate artifact lets users use the API without bringing in the deps.
If the library has an optional dependency on grpc, that can be quite
convenient.

We now version-pin both grpc-api and grpc-core, since both contain
internal APIs.

I had to change a few tests in grpc-api to avoid FakeClock. Moving
FakeClock to grpc-api was difficult because it uses
io.grpc.internal.TimeProvider, which can't be moved since it is a
production class. Having grpc-api's tests depend on grpc-core's test
classes would be weird and cause a circular dependincy. Having
grpc-api's tests depend on grpc-core is likely possible, but weird and
fairly unnecessary at this point. So instead I rewrote the tests to
avoid FakeClock.

Fixes #1447

------

Most of the non-trivial changes are dealing with the FakeClock goo.
Note that Bazel has //core:core export //api:api. I chose not to make
//core:core deprecated at this time, but we can if we'd like.

CC @anuraaga, @adriancole


api/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    testCompile project(':grpc-context').sourceSets.test.output,
+            project(':grpc-testing'),
+            project(':grpc-grpclb'),
+            libraries.guava_testlib
+
+    jmh project(':grpc-core')
+
+    signature "org.codehaus.mojo.signature:java17:1.0@signature"
```
adriancole: note: this matches grpc-context

api/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
...
+    srcs = glob([
+        "src/main/java/**/*.java",
+    ]),
+    visibility = ["//visibility:public"],
+    deps = [
+        "//context",
+        "@com_google_code_findbugs_jsr305//jar",
+        "@com_google_guava_failureaccess//jar",  # future transitive dep of Guava. See #5214
+        "@com_google_guava_guava//jar",
+        "@com_google_j2objc_j2objc_annotations//jar",
```
adriancole: this doesn't obviously match the build.gradle. is that intentional? I can see these deps are mostly the same as what context brings though.. 

```
    deps = [
        "@com_google_code_findbugs_jsr305//jar",
        "@com_google_errorprone_error_prone_annotations//jar",
        "@com_google_guava_guava//jar",
        "@com_google_j2objc_j2objc_annotations//jar",
],
```

ejona86: Uhhh... grpc-context should have no deps. Those others should be removed. They aren't there in context/build.gradle, so at least that's right. It looks like it was added with too many deps when it was initially created in 6d9e149dcb. Created #5591 to clean it up.

The only non-annotation dependencies of grpc-api are grpc-context and Guava.

The dependencies between gradle and Bazel seem to match here. What may be confusing you is animalsniffer and errorprone. Those are included in build.gradle because we want to use our versions instead of the versions used by Guava (since our versions are newer, and we use those dependencies elsewhere in grpc). So you will see errorprone listed, but then there is the exclude in the `compile (libraries.guava)` section. This is Maven dependencyConvergence goo (although we technically don't support dependencyConvergence, but we still want to support requireUpperBoundDeps and don't have a good tool yet to do so).

Transitive vs direct dependencies are messed up still in Bazel, so that adds confusion (it just confused me). j2objc_annotations, for example, I think is only needed by guava. But the Maven Central glue for Bazel hasn't historically done transitive dependencies well (#5359).

adriancole:
would it be too much to ask to have this work on guava 19 signatures, even if that means inlining some Preconditions? I'm not sure the scope of what's used in the grpc-api package, but it would be nice if this makes it more flexible.

api/src/test/java/io/grpc/ForwardingExecutorService.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Forwards all methods to delegate.
+ */
+abstract class ForwardingExecutorService implements ExecutorService {
```
anuraaga: Since this is in the `test` source set, does it make sense to just go ahead and use Guava? I guess dependencies aren't as much of an issue in the `test` scope.

ejona86:
Moving discussion of Guava 19 support to #5592.

api/src/test/java/io/grpc/ForwardingExecutorService.java
ejona86: Doh, yes. I checked Guava for ForwardingScheduledExecutorService and so started making my own, forgetting that Guava has ForwardingExecutorService.

ejona86:
Rebased to address LoadBalancerTest conflicting file in github. Rebasing handled it without manual interaction. Now waiting for CIs to run.

adriancole:
thanks all!

PR #5594, opened 2019-04-15
context: avoid synthetic methods on Context

carl-mastrangelo:
This change avoids an extra stack frame by removing synthetic methods like `io.grpc.Context$access100`.  These are removed by making the fields and methods package-private, which avoids the methods (and should have a minor improvement on method count and binary size).

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -185,8 +185,8 @@ public class Context {
     return current;
   }
 
-  ArrayList<ExecutableListener> listeners;
-  CancellationListener parentListener = new ParentListener();
+  private ArrayList<ExecutableListener> listeners;
+  private CancellationListener parentListener = new ParentListener();
   final CancellableContext cancellableAncestor;
   final PersistentHashArrayMappedTrie<Key<?>, Object> keyValueEntries;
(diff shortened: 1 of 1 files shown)


context/src/main/java/io/grpc/Context.java
```diff
@@ -185,8 +185,8 @@ public static Context current() {
     return current;
   }
 
-  private ArrayList<ExecutableListener> listeners;
-  private CancellationListener parentListener = new ParentListener();
+  ArrayList<ExecutableListener> listeners;
+  CancellationListener parentListener = new ParentListener();
```
zhangkun83: The fields of private inner classes are fine, since this PR doesn't make them visible to outside of the file. `checkNotNull()` and `lookup()` and also fine, since no harm can be done with it. I am a little uneasy about making these two mutable fields package-private, although we are unlikely to mess with it outside of the class. @ejona86 WDYT?

ejona86: Why are these being changed? `listeners` and `parentListener` is only referenced directly by Context.

carl-mastrangelo: Accidentally marked public while trying to reverse the field from javap.  PTAL

PR #5595, opened 2019-04-15
docs: add note about retrying UNAVAILABLE

jadekler:
Per recent discussions.

Go: https://github.com/grpc/grpc-go/pull/2774
C: https://github.com/grpc/grpc/pull/18761

### DIFF
--- a/api/src/main/java/io/grpc/Status.java
+++ b/api/src/main/java/io/grpc/Status.java
@@ -182,7 +182,8 @@ public final class Status {
     /**
      * The service is currently unavailable.  This is a most likely a
      * transient condition and may be corrected by retrying with
-     * a backoff.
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/grpc/Status.java
```diff
@@ -182,7 +182,8 @@
     /**
      * The service is currently unavailable.  This is a most likely a
      * transient condition and may be corrected by retrying with
-     * a backoff.
+     * a backoff. Note that it is not always safe to retry
+     * non-idempotent operations.
```
sanjaypujare: Can we reword to avoid double negation? Such as:

"Note that it could be unsafe to retry non-idempotent operations" 

or even better

"Note that it is safe to retry only idempotent operations."

Issue #5593, opened 2019-04-15
OpenCensus grpc method is not set


TheMarex:
### What version of gRPC are you using?

We are using 1.19.0 in combination with OpenCensus 0.19.2.

### What did you expect to see?

When using the `RpcViewConstants.GRPC_SERVER_SERVER_LATENCY_VIEW` from `opencensus-java` I would expect to see the metric split up by gRPC method, as is intended by them grouping the metric by the tag `grpc_server_method`.

However, `grpc-java` still sets the deprecated `method` tag which means this grouping is broken when using `grpc-java:0.19.0` with `opencensus-java:0.19.2`.

The easy fix would be to set both tags, to ensure a wider compatibility with opencensus. I can prepare a PR if that sounds like a good approach.

TheMarex:
I did some digging in the history and this regression basically boils down to the decision in https://github.com/grpc/grpc-java/pull/4494 to not also support the new constants. Seems like OpenCensus decided to switch to using the new constants in their default views anyway, which broke the metrics for clients using these views.

#4494 mentioned a performance regression, I would like to start a discussion about quantifying the problem and potentially taking the slow-down if necessary. Right now OpenCensus integration with gRPC is broken in my opinion, since there is no current gRPC version for which OpenCensus will work.

gRPC seems like the right place for the fix, since it would allow a wide range of OpenCensus versions to be compatible and give people time to migrate in any code that references the old `method` while partially porting to the new `grpc_{client,server}_method`.

ejona86:
CC @songy23, @sebright

songy23:
Thanks @TheMarex for looking into this. Initially I was taking the same approach in https://github.com/grpc/grpc-java/pull/5601, but it has caused performance problem because it's expensive to attach more tags to every client and server RPC. At that time we suggested to keep using the old constants for backwards-compatibility, and after a certain time of deprecation window we'll just replace the old constants with the new ones. 

Since you brought this up, maybe it's a good time to revisit the approach that adding both new and old OpenCensus tags. @zhangkun83 would you be able to benchmark https://github.com/grpc/grpc-java/pull/5601? 

If performance is still a problem, I think we can try migrating to the new tags in gRPC code first, and then do some tricks in OpenCensus to map the new tags to old ones in order to keep backwards-compatibility.

[...]

PR #5601, opened 2019-04-16
core: Add new OpenCensus tags for method names

TheMarex:
This PR fixes https://github.com/grpc/grpc-java/issues/5593 by adding support for the new gRPC method to the `CensusModule`.

I think it will be necessary to have this overlap for at least one versions to give clients a chance to upgrade without breaking changes. Since gRPC already assumes OpenCensus >0.17, the required migration steps for a client would be:

1. Upgrade gRPC to a new version >0.19 (with this change).
2. Update applications to report the new `OpenCensus` gRPC `View`s provided by `opencensus-java` that use these new tags
3. Update the frontend/dashboard to the new views
4. Update applications to stop reporting metrics using the old OpenCensus gRPC `View`s

### DIFF
--- a/core/src/main/java/io/grpc/internal/CensusStatsModule.java
+++ b/core/src/main/java/io/grpc/internal/CensusStatsModule.java
@@ -41,6 +41,8 @@ import io.opencensus.stats.Measure.MeasureLong;
 import io.opencensus.stats.MeasureMap;
 import io.opencensus.stats.Stats;
 import io.opencensus.stats.StatsRecorder;
+import io.opencensus.tags.InternalUtils;
+import io.opencensus.tags.Tag;
 import io.opencensus.tags.TagContext;
 import io.opencensus.tags.TagValue;
 import io.opencensus.tags.Tagger;
--- a/core/src/main/java/io/grpc/internal/CensusStatsModule.java
+++ b/core/src/main/java/io/grpc/internal/CensusStatsModule.java
@@ -41,8 +41,6 @@ import io.opencensus.stats.Measure.MeasureLong;
 import io.opencensus.stats.MeasureMap;
 import io.opencensus.stats.Stats;
 import io.opencensus.stats.StatsRecorder;
-import io.opencensus.tags.InternalUtils;
-import io.opencensus.tags.Tag;
 import io.opencensus.tags.TagContext;
 import io.opencensus.tags.TagValue;
 import io.opencensus.tags.Tagger;
--- a/core/src/test/java/io/grpc/internal/CensusModulesTest.java
+++ b/core/src/test/java/io/grpc/internal/CensusModulesTest.java
@@ -289,14 +289,16 @@ public class CensusModulesTest {
     StatsTestUtils.MetricsRecord record = statsRecorder.pollRecord();
     assertNotNull(record);
     TagValue methodTagOld = record.tags.get(DeprecatedCensusConstants.RPC_METHOD);
+    TagValue methodTagNew = record.tags.get(RpcMeasureConstants.GRPC_CLIENT_METHOD);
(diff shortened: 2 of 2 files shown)


thelinuxfoundation:

* User @TheMarex isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

thelinuxfoundation:

* User @TheMarex isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

ejona86:
CC @songy23, @sebright

zhangkun83:
@carl-mastrangelo has been profiling gRPC with Census. Unless there has been performance optimization in Census tagging, I don't expect the performance overhead to be different from the previous results. It may be better to first do the mapping mentinoed in https://github.com/grpc/grpc-java/issues/5593#issuecomment-484267801 under the hood, then migrate gRPC to the new tags.

songy23:
> It may be better to first do the mapping mentinoed in #5593 (comment) under the hood, then migrate gRPC to the new tags.

Make sense to me, added the mapping in https://github.com/census-instrumentation/opencensus-java/pull/1854. With that those who use the deprecated constants will continue getting the correct metrics.

TheMarex:
@songy23 with the change in census-instrumentation/opencensus-java#1854 in mind, should I adapt this PR to only set the non-deprecated constant?

My understanding then would be that this would mean all following gRPC versions would depend on a OpenCensus version > 0.19 - is that correct?

Thank you for looking at this so fast. :+1: Sorry about the CLA noise, figuring this out with my org.

songy23:
> should I adapt this PR to only set the non-deprecated constant?

If there's no performance regress, I would still like to set both the deprecated and new tags. That way we can ensure both old and new RPC metrics work for all OpenCensus versions.

On the other hand if only the new tags are set, users who depend on a lower version of OpenCensus (<0.21) and use the old RPC metrics will lose data of the old "method"/"status" tag. They'll have to either migrate to the new RPC metrics, or upgrade OpenCensus to 0.21.

> My understanding then would be that this would mean all following gRPC versions would depend on a OpenCensus version > 0.19 - is that correct?

That's right, https://github.com/census-instrumentation/opencensus-java/pull/1854 will be released in OpenCensus 0.21.0. After the release I'll update gRPC to use the newer version.

ejona86:
> On the other hand if only the new tags are set, users who depend on a lower version of OpenCensus (<0.21) and use the old RPC metrics will lose data of the old "method"/"status" tag. They'll have to either migrate to the new RPC metrics, or upgrade OpenCensus to 0.21.

gRPC does not support users downgrading versions of our dependencies. If we require opencensus 0.21, the user can use opencensus 0.21 or newer.

songy23:
> gRPC does not support users downgrading versions of our dependencies. If we require opencensus 0.21, the user can use opencensus 0.21 or newer.

That's great, then please hold off this PR a bit until I upgraded the OC version.

thelinuxfoundation:

* User @TheMarex isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

dapengzhang0:
@TheMarex please fix build failures.

igorbernstein2:
I propose to close this PR in favor of https://github.com/grpc/grpc-java/pull/5996

TheMarex:
@igorbernstein2 thanks for picking this up again. I agree, please use #5996 in favor of this.

FYI I wasn't able to figure out the CLA thing and Linux Foundation is unresponsive.

PR #5604, opened 2019-04-16
netty: deflake ping flow control logic

carl-mastrangelo:
The last ping time was not set, leading to the bandwidth since last ping in `FlowControlPinger.updateWindow()` to always be zero.  

Fixes failures of the form:
```
io.grpc.netty.NettyClientHandlerTest > windowShouldNotExceedMaxWindowSize FAILED
    java.lang.AssertionError: expected:<8388608> but was:<65535>
```


carl-mastrangelo:
This is still technically flaky, but I don't feel like wiring Stopwatch all the way through to the pinger.  This change should fix most of the flake.

PR #5606, opened 2019-04-16
Allow TCP_USER_TIMEOUT for Netty client and ComputeEngineChannelBuilder.

creamsoup:
moved Epoll related methods to EpollUtils. which can expose some of its utility methods especially ChannelOption TCP_USER_TIMEOUT. 
Alternatively, we can open the channel option in ComputeEngineChannelBuilder. 

This introduces conflict between TCP_USER_TIMEOUT using keepAliveTime. Current approach is using smaller value of this. This can be debatable, should we always favor ChannelOption?


creamsoup:
we decided to use KeepAlive to implicitly set TCP_USER_TIMEOUT.

PR #5611, opened 2019-04-18
netty: add internal API to fall back to NIO transport

creamsoup:
channel type in the NettyServerBuilder is using null which in unfortunate. it wasn't able to tell if the user provided same type. 
Converting to channel factory was considered, but in the Server#start checks instance of channel type to set SocketChannel specific channel options. if we want to use channelFactory route, we can just blindly set those options which is probably not a big deal.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
+++ b/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
@@ -218,10 +218,14 @@ public final class NettyChannelBuilder
    */
   public NettyChannelBuilder eventLoopGroup(@Nullable EventLoopGroup eventLoopGroup) {
     if (eventLoopGroup != null) {
-      this.eventLoopGroupPool = new FixedObjectPool<>(eventLoopGroup);
-    } else {
-      this.eventLoopGroupPool = DEFAULT_EVENT_LOOP_GROUP_POOL;
+      return eventLoopGroupPool(new FixedObjectPool<>(eventLoopGroup));
     }
+    return eventLoopGroupPool(DEFAULT_EVENT_LOOP_GROUP_POOL);
+  }
+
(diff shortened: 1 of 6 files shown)


netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -218,10 +218,14 @@ public NettyChannelBuilder negotiationType(NegotiationType type) {
...
-      this.eventLoopGroupPool = new FixedObjectPool<>(eventLoopGroup);
-    } else {
-      this.eventLoopGroupPool = DEFAULT_EVENT_LOOP_GROUP_POOL;
+      return eventLoopGroupPool(new FixedObjectPool<>(eventLoopGroup));
     }
+    return eventLoopGroupPool(DEFAULT_EVENT_LOOP_GROUP_POOL);
+  }
+
+  NettyChannelBuilder eventLoopGroupPool(ObjectPool<? extends EventLoopGroup> eventLoopGroupPool) {
+    checkNotNull(eventLoopGroupPool, "eventLoopGroupPool");
```
ejona86: nit: assign at same time?

netty/src/main/java/io/grpc/netty/NettyServerBuilder.java
```diff
@@ -522,6 +531,8 @@ public NettyServerBuilder permitKeepAliveWithoutCalls(boolean permit) {
             + "BossEventLoopGroup is missing. Fall back to NioEventLoopGroup.");
       }
     }
+    resolvedChannelType =
+        resolvedChannelType != null ? resolvedChannelType : Utils.DEFAULT_SERVER_CHANNEL_TYPE;
```
ejona86:
```java
if (resolvedChannelType == null) {
  resolvedChannelType = Utils.DEFAULT_SERVER_CHANNEL_TYPE;
}
```

PR #5618, opened 2019-04-19
java_grpc_library: Use toolchain to enable DexArchiveAspect

ejona86:
This removes the "java_import to work around error with android_binary"
hack; see the b/78647825 references to see the really important parts.
But in general using a toolchain cleans up the code considerably and
allows us to reduce the visibility of our protoc plugin. It also is
useful to enable using aspects in the future to avoid using
make_non_strict() which would improve compilation speed avoid avoid many
unnecessary rebuilds.

This is an export of cl/229763103 and brings external up-to-sync with
internal.

### DIFF
--- a/compiler/BUILD.bazel
+++ b/compiler/BUILD.bazel
@@ -1,3 +1,5 @@
+load("//:java_grpc_library.bzl", "java_rpc_toolchain")
+
 cc_binary(
     name = "grpc_java_plugin",
     srcs = [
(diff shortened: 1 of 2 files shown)


compiler/BUILD.bazel
```diff
@@ -1,12 +1,53 @@
+load("//:java_grpc_library.bzl", "java_rpc_toolchain")
+
```
carl-mastrangelo: nit: I think you should set a default visibility of private here, and explicitly declare the vis on the rest of the targets.

ejona86: Private is the default, and why repeat it again to each of the targets?

Issue #4896, opened 2018-10-01
Using proto_source_root in proto_library breaks java_grpc_library


haikalpribadi:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?

v1.15.0

### What did you expect to see?

Say I have 2 proto files sitting in `//some-package/`
- `//some-package/A.proto` which `imports B.proto` (without `some-package` as prefix)
- `//some-package/B.proto`

Then I compile the `proto_library in` //some-package/BUILD`, like the following:
```
proto_library(
    name = "some-proto",
    srcs = glob(["*.proto"])
    proto_source_root = "some-package"
)
```
The above works well because `proto_source_root`, which allows us to not have to import one proto file into another using the full path relative from WORKSPACE.

As instructed on GRPC docs (https://grpc.io/docs/reference/java/generated-code.html) Next I compile `java_proto_library`:
```
java_proto_library(
    name = "java-proto",
    deps = ["some-proto"],
)
```
This also works fine. But finally, I compile `java_grpc_library`:
```
java_grpc_library(
    name = "java-grpc",
    srcs = [":some-proto"],
    deps = [":java-proto"],
)
```
And here it breaks. `java_grpc_library` failes to compile with error:
```
B.proto: file not found.
some-package/A.proto: Import "B.proto" was not found or had errors.
```
So basically, `A.proto` cannot find `B.proto`. However, when I replace the import statement in `A.proto` with `import some-package/B.proto` and remove usage of `proto_source_root` in `proto_library` everything works, but we really don't want to do this for many reasons.

Does anyone know what could be the cause and how to work around this?

haikalpribadi:
To help reproduce this problem easily, I've isolated the problem on a branch on my repo: https://github.com/haikalpribadi/grakn/tree/bazel-init-grpc-java-broken-with-proto-source-root

Thank you!

haikalpribadi:
Thans, @ejona86. Is there any examples on how to use file/transitive_descriptor_sets ?

Although, I do think adding `proto_source_root` to `java_grpc_library` (and every other `*_grpc_library` for that matter) is a more elegant solution. And an even better solution is for `java_grpc_library` to not need `proto_source_root` at all, and instead rely on the fact that the `proto_library` has already encapsulated this information. Is that possible?

ejona86:
> And an even better solution is for java_grpc_library to not need proto_source_root at all, and instead rely on the fact that the proto_library has already encapsulated this information.

[...]

PR #5621, opened 2019-04-21
java_grpc_library: enable (strip_)import_prefix in proto dependencies

mjduijn:
Enables strip_import_prefix and import_prefix in proto dependencies by:
* Relying on descriptor sets for transitive dependencies instead of the actual proto files
* Using `proto_path(..)` from bazelbuild/rules_go instead of `_path_ignoring_repository(..)`

As far as I can tell `java_grpc_library` does not have regression tests (yet) so I have a set of proto files locally that test the behavior.
I would be more than happy to share them.

### DIFF
--- a/java_grpc_library.bzl
+++ b/java_grpc_library.bzl
@@ -58,11 +58,49 @@ java_rpc_toolchain = rule(
     implementation = _java_rpc_toolchain_impl,
 )
 
-# "repository" here is for Bazel builds that span multiple WORKSPACES.
-def _path_ignoring_repository(f):
-    if len(f.owner.workspace_root) == 0:
-        return f.short_path
-    return f.path[f.path.find(f.owner.workspace_root) + len(f.owner.workspace_root) + 1:]
+# Taken from bazelbuild/rules_go license: Apache 2
+# https://github.com/bazelbuild/rules_go/blob/528f6faf83f85c23da367d61f784893d1b3bd72b/proto/compiler.bzl#L94
+# replaced `prefix = paths.join(..` with `prefix = "/".join(..`
+def _proto_path(src, proto):
+    """proto_path returns the string used to import the proto. This is the proto
+    source path within its repository, adjusted by import_prefix and
+    strip_import_prefix.
+
+    Args:
+        src: the proto source File.
+        proto: the ProtoInfo provider.
+
+    Returns:
+        An import path string.
+    """
+    if not hasattr(proto, "proto_source_root"):
+        # Legacy path. Remove when Bazel minimum version >= 0.21.0.
+        path = src.path
+        root = src.root.path
+        ws = src.owner.workspace_root
+        if path.startswith(root):
+            path = path[len(root):]
+        if path.startswith("/"):
+            path = path[1:]
+        if path.startswith(ws):
+            path = path[len(ws):]
+        if path.startswith("/"):
+            path = path[1:]
+        return path
+
+    if proto.proto_source_root == ".":
+        # true if proto sources were generated
+        prefix = src.root.path + "/"
+    elif proto.proto_source_root.startswith(src.root.path):
+        # sometimes true when import paths are adjusted with import_prefix
+        prefix = proto.proto_source_root + "/"
+    else:
+        # usually true when paths are not adjusted
+        prefix = "/".join([src.root.path, proto.proto_source_root]) + "/"
+    if not src.path.startswith(prefix):
+        # sometimes true when importing multiple adjusted protos
+        return src.path
+    return src.path[len(prefix):]
 
 def _java_rpc_library_impl(ctx):
     if len(ctx.attr.srcs) != 1:
@@ -72,8 +110,9 @@ def _java_rpc_library_impl(ctx):
                "same package as consuming rule").format(ctx.label, ctx.attr.srcs[0].label))
 
     toolchain = ctx.attr._toolchain[_JavaRpcToolchainInfo]
-    srcs = ctx.attr.srcs[0][ProtoInfo].direct_sources
-    descriptor_set_in = ctx.attr.srcs[0][ProtoInfo].transitive_descriptor_sets
+    proto = ctx.attr.srcs[0][ProtoInfo]
+    srcs = proto.direct_sources
+    descriptor_set_in = proto.transitive_descriptor_sets
 
     srcjar = ctx.actions.declare_file("%s-proto-gensrc.jar" % ctx.label.name)
 
@@ -81,10 +120,12 @@ def _java_rpc_library_impl(ctx):
     args.add(toolchain.plugin, format = "--plugin=protoc-gen-rpc-plugin=%s")
     args.add("--rpc-plugin_out={0}:{1}".format(toolchain.plugin_arg, srcjar.path))
     args.add_joined("--descriptor_set_in", descriptor_set_in, join_with = ":")
-    args.add_all(srcs, map_each = _path_ignoring_repository)
+    for src in srcs:
+        args.add(_proto_path(src, proto))
 
     ctx.actions.run(
-        inputs = depset([toolchain.plugin] + srcs, transitive = [descriptor_set_in]),
+        inputs = descriptor_set_in,
+        tools = [toolchain.plugin],
         outputs = [srcjar],
         executable = toolchain.protoc,
         arguments = [args],


thelinuxfoundation:

* User @mjduijn isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

mjduijn:
I signed it

ejona86:
> Relying on descriptor sets for transitive dependencies

I've tried this twice time before, and it's introduced compatibility issues. I've not had a chance to figure out why (but I did spend time on it; it is just non-obvious). The incompatibilities exposed themselves when testing at scale internally at Google, and caused build breakages.

I fought `proto_source_root` before and described my issues in https://github.com/grpc/grpc-java/issues/4896#issuecomment-457648775 . I've not looked in-depth at `strip_import_prefix`, since I've not had time after wasting a bunch of effort on `proto_source_root`. It's not entirely clear to me how `strip_import_prefix` is different than `proto_source_root`.

ejona86:
I've got a commit with some targets to build as a test at https://github.com/ejona86/grpc-java/tree/builddefstest . It's too ugly to merge at the moment (mainly the directory tree; it needs a better home).

mjduijn:
I would have thought using descriptor sets would reduce compatibility issues. Using an encapsulated descriptor intuitively seems more robust than relying on proto files and then setting the correct `proto_paths`.

I can't comment on google internal compatibility issues, you definitely have more experience here and I'm fine with closing the PR in favor of a future solution. Based on the referenced comment, your builddefstest commit and some of my earlier tests I've tried to disprove my implementation and succeeded in the end (branch: https://github.com/mjduijn/grpc-java/tree/wip-enable-proto-import-prefix). This was what got me:

> Part of the problem is that when using proto_source_root you can still do imports based on the full path. (That is, it seems you can take any proto_library and add proto_source_root to it without breaking the build.)

Your follow-up about the need to access unexposed internals seems dead-on as well since the problem lies with the imports not mapping the correct descriptors. I do not see a way to implement `(strip_)import_prefix` without using descriptors at the moment but I think I'll waste a little more time before closing the PR.

ejona86:
> I would have thought using descriptor sets would reduce compatibility issues.

Oh, definitely. It is mostly just is _a change_, neither good nor bad. And whatever java_grpc_library is doing today "obviously works" because if it didn't work then an engineer would have never submitted the code that used it. I just need to go through and figure out what sort of issues the change tickles and either fix some bug, workaround it, or fix them manually.

> I've tried to disprove my implementation and succeeded in the end

Shoot. I wished you would have failed :-).  I had a _small_ hope that the new strip_import_prefix wouldn't have that problem.

> I do not see a way to implement (strip_)import_prefix without using descriptors at the moment

IIRC, descriptors were mostly unrelated. If you have a solution that works with descriptors, that is a solution in my mind. I _want_ to use descriptors, but just haven't found the time and it's not been important since nothing depends on it. If descriptors solves strip_import_prefix I'd have a better chance of finding the time. As I said elsewhere, I want to use strip_import_prefix with gRPC's protos.

ejona86:
@mjduijn, I poked again at using the descriptor set. The problem is the transitive descriptor set is finding duplicate messages in some cases and failing. There are indeed duplicate (different) messages, but the duplicate messages aren't actually used for the codegen so the pre-existing execution of protoc succeeds (for both java_proto_library and java_grpc_library). This would be hard to fix.

mjduijn:
The new version no longer uses descriptor sets, instead it uses an aspect to calculate both the adjusted import path (adjusted by `proto_source_root`, `import_prefix` and `strip_import_prefix`) as well as the import path from the repository root. I could make the aspect return the transitive files as well so that `ctx.attr.srcs[0][ProtoInfo].transitive_imports` is no longer needed. Since the protoc proto_path argument needs to point to a proto file and not a descriptor (correct?) I couldn't find a way to make this work using the descriptor set.

This is purely for the benefit of being able to set both of the imports below in a proto file. In my opinion the previous version (https://github.com/mjduijn/grpc-java/blob/c00e2e52b186fe08357bc8f26f4772e192cd3500/java_grpc_library.bzl) was cleaner and I'm afraid the added proto_path will cause new compatibility issues.

    import "test1/A.proto";                     // path adjusted by proto_source_root argument
    import "some-other-package/test1/A.proto";  // path from repository root

The `strip_import_prefix` argument works different from `proto_source_root` in that it seems impossible to add the argument to a `proto_library` rule and keep the import paths from the root of the project. This eliminates the need for the extra import path and thus allows the use of the descriptor sets. Since `proto_source_root` will be deprecated it might be worth to just wait it out, or to force all google internal protos to use `strip_import_prefix` instead of it (have no idea how much work this will be, but I'm guessing quite a bit :D) 

I don't understand your comment regarding duplicate messages. Is the duplication error silenced, the generated file discarded and thus missing? Do you have an example of this happening? How can I reproduce this?

ejona86:
Give me some time to take a closer look at what you've done. Here are some responses, but I've not yet looked at the code.

> The strip_import_prefix argument works different from proto_source_root in that it seems impossible to add the argument to a proto_library rule and keep the import paths from the root of the project.

Oh, that's good.

> Since proto_source_root will be deprecated it might be worth to just wait it out, or to force all google internal protos to use strip_import_prefix instead of it

Yeah, in my mind there is no need to support `proto_source_root`. If it was easy to have it work, that would be fine. But there's no point in going out of our way to support it. Neither of these options are used much at Google, as they are relatively new and the monorepo means there's less need for them.

> I don't understand your comment regarding duplicate messages. Is the duplication error silenced, the generated file discarded and thus missing? Do you have an example of this happening? How can I reproduce this?

When using `--descriptor_set_in` protoc does some processing that notices the duplication, and then aborts with an error like `Symbol name "package.Foo" conflicts with the existing symbol "package.Foo".`

Reproduced at https://github.com/ejona86/grpc-java/tree/duplicate-descriptor_set_in . Yesterday I was already thinking it might not be as wide-spread as I feared. I meant to do a batch job last night to check how many things would break, but forgot. After the minimal reproduction today, I think it can only be triggered if an unused dependency is listed in `proto_library`, as otherwise protoc complains while parsing the .proto files; that would make this 1) pretty rare and 2) easy to fix.

mjduijn:
@ejona86 Did you have a chance to look at the implementations?

The change for transitive descriptors is not currently in the PR but available here https://github.com/mjduijn/grpc-java/commit/20e79364295829a4f3fba5ece75861a622141598. 
I would rather not start flip-flopping but I can place it on top of the current commit if you are leaning in that direction.

The rules work within my company repo but we only have 10 protos with grpc java rules so the results may not be accurate. 
I do have a set of gazelle rules to generate them so if it helps I could probably add the rules to [googleapis/googleapis](https://github.com/googleapis/googleapis) or [envoyproxy/envoy](https://github.com/envoyproxy/envoy).

> When using --descriptor_set_in protoc does some processing that notices the duplication, and then aborts with an error like Symbol name "package.Foo" conflicts with the existing symbol "package.Foo".
Reproduced at https://github.com/ejona86/grpc-java/tree/duplicate-descriptor_set_in . Yesterday I was already thinking it might not be as wide-spread as I feared. I meant to do a batch job last night to check how many things would break, but forgot. After the minimal reproduction today, I think it can only be triggered if an unused dependency is listed in proto_library, as otherwise protoc complains while parsing the .proto files; that would make this 1) pretty rare and 2) easy to fix.

Thank you for this explanation! This was a bit of an eye opener in how using descriptors differs from using actual protos.

ejona86:
> Did you have a chance to look at the implementations?

:( no. _However_, I _did_ dig into the descriptor_set_in issues and know what minor changes are necessary. I have a one-line fix to another project internally and then I can swap to descriptor_set_in.

dapengzhang0:
ping @ejona86

mjduijn:
@ejona86 I've adjusted based on the changes in #5837.
I'm not too happy about iterating over the srcs but I could not find a better way to add the arguments.

ejona86:
https://github.com/bazelbuild/bazel/pull/8814 is advertised to fix this problem. It's a surprisingly small change. I'm looking into it (I was sent a change internally to apply that to grpc's codebase; I'd then also export it to this repo).

mjduijn:
Cool! Did not know about the `_virtual_imports`

PR #5630, opened 2019-04-23
netty: not using reflection to create NioEventLoopGroup

creamsoup:


netty/src/main/java/io/grpc/netty/Utils.java
```diff
@@ -268,16 +269,15 @@ private static Throwable getEpollUnavailabilityCause() {
...
 
-  private static EventLoopGroup createEventLoopGroup(
-      Class<? extends EventLoopGroup> eventLoopGroupType,
+  private static EventLoopGroup createEpollEventLoopGroup(
       int parallelism,
       ThreadFactory threadFactory) {
     try {
-      return eventLoopGroupType
+      return epollEventLoopGroupType()
           .getConstructor(Integer.TYPE, ThreadFactory.class)
```
ejona86: Go ahead and store the constructor as a static field. That also removes the need for the `unused` variable.

creamsoup: done. not sure this is what you meant

creamsoup:
> We're not using this syntax yet.

right, fixed.

PR #5634, opened 2019-04-24
api: expose if waitForReady has been set on CallOptions

carl-mastrangelo:
This is needed to properly implement the Service Config waitForReady underride.

### DIFF
--- a/api/src/main/java/io/grpc/CallOptions.java
+++ b/api/src/main/java/io/grpc/CallOptions.java
@@ -370,7 +370,7 @@ public final class CallOptions {
    * calls and 'wait for ready' is the opposite to it.
    */
   public boolean isWaitForReady() {
-    return waitForReady;
+    return Boolean.TRUE.equals(waitForReady);
   }
 
   Boolean getWaitForReady() {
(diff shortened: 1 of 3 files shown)


ejona86:
@carl-mastrangelo  This is not necessary to implement the service config waitForReady logic. That logic can only change the value to 'true'.

carl-mastrangelo:
The https://github.com/grpc/grpc/blob/master/doc/service_config.md says  the client set value can override the service config one.

ejona86:
@markdroth and I talked about this in the past. @markdroth, since our clients won't be explicitly disabling waitForReady, we had talked about the limitations of applying the service config. What was the verdict?

markdroth:
You are both correct about the intended semantics.

The intended semantic is that the value specified via the client API will trump the value specified via the service config.  This means that we only look at the service config value when the client API does not specify any value.  There are therefore only two cases here:

1. API:unset (default false), service_config:false: wait_for_ready=false
2. API:unset (default false), service_config:true: wait_for_ready=true

Since case 1 is effectively a no-op, case 2 is the only one in which the service config value gets used.

In C-core, we plumb down the value from the client API as a trinary value: unset, true, or false.  Only if it's unset do we look at the service config value.

Issue #5636, opened 2019-04-25
GC friendly long live server streaming


Gordiychuk:
In my application, I use server streaming, when client send request and receive responses until connection will not close(days or more). At heapdump i found that initial request never released until stream not close.

https://github.com/grpc/grpc-java/blob/master/stub/src/main/java/io/grpc/stub/ServerCalls.java#L132
![Screenshot from 2019-04-25 17:06:10](https://user-images.githubusercontent.com/10699594/56742163-e493ae00-677c-11e9-9223-b9e144403084.png)

Why not release it as only method was invoked at this place https://github.com/grpc/grpc-java/blob/master/stub/src/main/java/io/grpc/stub/ServerCalls.java#L171

ejona86:
Fixed by #5638

PR #5638, opened 2019-04-25
server-streaming: release reference on request after method call

Gordiychuk:
fixing https://github.com/grpc/grpc-java/issues/5636

### DIFF
--- a/stub/src/main/java/io/grpc/stub/ServerCalls.java
+++ b/stub/src/main/java/io/grpc/stub/ServerCalls.java
@@ -129,7 +129,7 @@ public final class ServerCalls {
       private final ServerCall<ReqT, RespT> call;
       private final ServerCallStreamObserverImpl<ReqT, RespT> responseObserver;
       private boolean canInvoke = true;
-      private boolean messageReceived = false;
+      private boolean messageReceived;
       private ReqT request;
 
       // Non private to avoid synthetic class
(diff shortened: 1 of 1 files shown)


thelinuxfoundation:

* User @Gordiychuk isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

Gordiychuk:
Signed

stub/src/main/java/io/grpc/stub/ServerCalls.java
```diff
@@ -129,6 +129,7 @@ private ServerCalls() {
       private final ServerCall<ReqT, RespT> call;
       private final ServerCallStreamObserverImpl<ReqT, RespT> responseObserver;
       private boolean canInvoke = true;
+      private boolean messageReceived = false;
```
carl-mastrangelo: optional: drop the `= false`, as it's the default.

stub/src/main/java/io/grpc/stub/ServerCalls.java
```diff
@@ -169,6 +171,7 @@ public void onHalfClose() {
         }
 
         method.invoke(request, responseObserver);
+        request = null;
```
ejona86: It looks like you only need to add this line and not bother with the others. It doesn't seem `messageReceived` is improving any behavior.

ejona86:
This was a good find. Thank you!

stub/src/main/java/io/grpc/stub/ServerCalls.java
Gordiychuk: done

ejona86:
@Gordiychuk, thank you!

PR #5642, opened 2019-04-25
core: factor service config state management out of ManagedChannelImpl

carl-mastrangelo:
This class was an internal class in another branch, but it is somewhat too big.  I have factored it out into a reviewable piece.  It enforces all the ServiceConfig state transitions.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ServiceConfigState.java
+++ b/core/src/main/java/io/grpc/internal/ServiceConfigState.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright 2019 The gRPC Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.internal;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+import io.grpc.Status;
+import io.grpc.SynchronizationContext;
+import javax.annotation.Nullable;
+
+/**
+ * {@link ServiceConfigState} holds the state of the current service config.  It must be mutated
+ * and read from {@link ManagedChannelImpl} constructor or the provided syncContext.
+ */
+final class ServiceConfigState {
+  @Nullable private final ManagedChannelServiceConfig defaultServiceConfig;
+  private final boolean lookUpServiceConfig;
+  private final SynchronizationContext syncCtx;
+
+  // mutable state
+  @Nullable private ManagedChannelServiceConfig currentServiceConfig;
+  @Nullable private Status currentError;
+
+  /**
+   * @param defaultServiceConfig The initial service config, or {@code null} if absent.
+   * @param lookUpServiceConfig {@code true} if service config updates might occur.
+   * @param syncCtx The synchronization context that this is accessed from.
+   */
+  ServiceConfigState(
+      @Nullable ManagedChannelServiceConfig defaultServiceConfig,
+      boolean lookUpServiceConfig,
+      SynchronizationContext syncCtx) {
+    this.defaultServiceConfig = defaultServiceConfig;
+    this.lookUpServiceConfig = lookUpServiceConfig;
+    this.syncCtx = checkNotNull(syncCtx, "syncCtx");
+    if (!lookUpServiceConfig) {
+      this.currentServiceConfig = defaultServiceConfig;
+    }
+  }
+
+  /**
+   * Returns {@code true} if it RPCs should wait on a service config resolution.  This can return
+   * {@code false} if:
+   *
+   * <ul>
+   *   <li>There is a valid service config from the name resolver
+   *   <li>There is a valid default service config and a service config error from the name
+   *       resolver
+   *   <li>No service config from the name resolver, and no intent to lookup a service config.
+   *
+   * In the final case, the default service config may be present or absent, and will be the
+   * current service config.
+   */
+  boolean waitOnServiceConfig() {
+    syncCtx.throwIfNotInThisSynchronizationContext();
+    if (currentServiceConfig != null || currentError != null) {
+      return false;
+    } else {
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/ServiceConfigState.java
```diff
@@ -0,0 +1,135 @@
...
+   */
+  @Nullable Status getCurrentError() {
+    checkState(!waitOnServiceConfig(), "still waiting on service config");
+    return currentError;
+  }
+
+  /**
+   * Returns {@code true} if the update was successfully applied, else {@code false}.
+   */
+  boolean update(Status error) {
```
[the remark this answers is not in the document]
carl-mastrangelo: I plan on either logging it if the return value is false, or not and letting the channel go into transient failure.

core/src/main/java/io/grpc/internal/ServiceConfigState.java
```diff
@@ -0,0 +1,135 @@
...
+   * <ul>
+   *   <li>There is a valid service config from the name resolver
+   *   <li>There is a valid default service config and a service config error from the name
+   *       resolver
+   *   <li>No service config from the name resolver, and no intent to lookup a service config.
+   *
+   * In the final case, the default service config may be present or absent, and will be the
+   * current service config.
+   */
+  boolean waitOnServiceConfig() {
```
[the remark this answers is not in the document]
carl-mastrangelo: Done.

carl-mastrangelo:
I have updated the service config state transitions, and included a test case for each possible transition.  PTAL

carl-mastrangelo:
@zhangkun83 friendly ping

core/src/main/java/io/grpc/internal/ServiceConfigState.java
```diff
@@ -0,0 +1,130 @@
...
+    checkState(!shouldWaitOnServiceConfig(), "still waiting on service config");
+    return currentServiceConfig;
+  }
+
+  /**
+   * Gets the current service config error.
+   *
+   * @throws IllegalStateException if the service config has not yet been updated.
+   */
+  @Nullable Status getCurrentError() {
```
zhangkun83: I wonder if the caller would care about the error only when currentServiceConfig is null. In other words, I wonder if it would be sufficient to have a `ConfigOrError` field rather than `currentServiceConfig` and `currentError`.

core/src/main/java/io/grpc/internal/ServiceConfigState.java
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * Returns {@code true} if the update was successfully applied, else {@code false}.
+   */
+  boolean update(Status error) {
+    checkNotNull(error, "error");
+    checkArgument(!error.isOk(), "can't use OK error");
+    checkState(expectUpdates(), "unexpected service config update");
+    boolean firstUpdate = !updated;
+    updated = true;
+    if ((firstUpdate && defaultServiceConfig == null) || currentError != null) {
```
zhangkun83: The flow of control is a little difficult to follow. It would be more readable if you structure it around the input conditions rather than output values:
```java
if (firstUpdate) {
  if (defaultServieConfig != null) {
    currentCoe = ConfigOrError.fromConfig(defaultServiceConfig);
  } else {
    currentCoe = ConfigOrError.fromError(error);
  }
} else {
  // Subsequent updates. Update the error only if already in error
  // Discard it otherwise.
  if (currentCoe != null && currentCoe.error != null) {
    currentCoe = ConfigOrError.fromError(error);
  }
}
```

core/src/main/java/io/grpc/internal/ServiceConfigState.java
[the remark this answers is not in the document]
carl-mastrangelo: Theres only two possible ways this method can work:

1.  The error is installed 
2.  The error is ignored.

If the error is installed into this state, the ManagedChannel can enter transient failure as described in the gRFC.    If it is not installed, it is notable to log.  The return value of this indicates if the installation was successful.  The Channel is the actual consumer of this error, not the ConfigServiceState.  It would be a violation of responsibility to handle the error here.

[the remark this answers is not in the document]
zhangkun83: I don't necessarily agree with you about the split of responsibilities, but it's minor.

core/src/main/java/io/grpc/internal/ServiceConfigState.java
carl-mastrangelo: ConfigOrError involves unnecessary casting in MCI.

carl-mastrangelo: Also, right now, ConfigOrError doesn't allow null, so then the ConfigOrError field would have to be nullable.   Leaving it as a MCSC simplifies MCI's usage of  getCurrentServiceConfig()

core/src/main/java/io/grpc/internal/ServiceConfigState.java
carl-mastrangelo: I tried it out as the folllowing, but I think its much more complicated.  It's not defined in terms of inputs or outputs, but as the state diagram.  You suggestion is in terms of edges, while this is in terms of the nodes.  The first condition (as is) says if it is in state 1 or state 2, go to state 2.  

```java
    boolean firstUpdate = !updated;
    updated = true;
    if (firstUpdate) {
      if (defaultServiceConfig == null) {
        assert currentServiceConfig == null;
        currentError = error;
        return true;
      } else if (currentServiceConfig == null) {
        currentServiceConfig = defaultServiceConfig;

      }
      return false;
    } else {
      if (currentError != null) {
        assert currentServiceConfig == null;
        currentError = error;
        return true;
      } else {
        return false;
      }
    }
```

core/src/main/java/io/grpc/internal/ServiceConfigState.java
zhangkun83: It's natural for a `ConfigOrError` field to be nullable, because it's the case in `ResolutionResult`. On the other hand, the config part of `ConfigOrError` is not nullable in the no-error case, while you keep it as a nullable field here, which I think is a mismatch and that involves conversion.

If you made both `update()` and `get()` deal with `ConfigOrError`, that would simplify the interface and bring better consistency.

Casting is unfortunate, but it's already the case on the API so it's not a new issue.

core/src/main/java/io/grpc/internal/ServiceConfigState.java
zhangkun83: I find edge-oriented is easier to reason about, because you can easily put comments on the edges. Node-oriented is harder to read, unless you have the state diagram at hand, which is not the case here.

carl-mastrangelo: I updated the code as requested, but I think it is much more verbose, more complicated, and not as obviously correct.  I left the old functions in place to work with the unit tests.

PR #5645, opened 2019-04-26
netty: fix TCP_USER_TIMEOUT to use keepAliveTimeout instead of keepAliveTime

creamsoup:


creamsoup:
yes, the keepAliveTimeNanos is not usued (every caller is using default), but not accepting keepAliveTime while accepting keepAliveTimeout is weird. But since it is a test code, i am fine with either way.

PR #5651, opened 2019-04-30
netty: set default socket options for all SocketChannels

creamsoup:


netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -157,7 +157,7 @@ public void start(ServerListener serverListener) throws IOException {
     ServerBootstrap b = new ServerBootstrap();
     b.group(bossGroup, workerGroup);
     b.channel(channelType);
-    if (NioServerSocketChannel.class.isAssignableFrom(channelType)) {
+    if (ServerSocketChannel.class.isAssignableFrom(channelType)) {
```
ejona86: Actually, why do we need this `if` at all?

creamsoup: not really. it could suppress some warning on netty side but since NCT is not checking, i'll remove it.

PR #5656, opened 2019-05-01
stub: improve error message for Client/ServerCalls 

creamsoup:
when StreamObserver.onNext is called after the stream is terminated.
resolves #5656

### DIFF
--- a/stub/src/main/java/io/grpc/stub/ClientCalls.java
+++ b/stub/src/main/java/io/grpc/stub/ClientCalls.java
@@ -17,6 +17,7 @@
 package io.grpc.stub;
 
 import static com.google.common.base.Preconditions.checkNotNull;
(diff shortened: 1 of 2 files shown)


creamsoup:
fyi. the force push is just rebase of current master for travis to pick up the java8 fix.

ejona86:
FYI: Travis builds from a merge commit of this PR to master. So just going to travis and hitting the "rebuild" button should have been enough.

creamsoup:
i actually tried the rebuild couple times, but it didn't work (config from the ui looked staled). i'll try again next time.

ejona86:
Okay. That's fine. I'm not bothered by the rebase to address this, I was just noting it shouldn't be necessary since that would be easier in the future.

PR #5657, opened 2019-05-01
 all: Upgrade OpenCensus versions.

songy23:
Also updated CensusModule to use the new helper methods `ContextUtils.withValue()` instead of directly manipulating the context keys. See https://github.com/census-instrumentation/opencensus-java/pull/1864.

/cc @bogdandrutu @dinooliva @zhangkun83

### DIFF
--- a/core/src/main/java/io/grpc/internal/CensusStatsModule.java
+++ b/core/src/main/java/io/grpc/internal/CensusStatsModule.java
@@ -18,7 +18,6 @@ package io.grpc.internal;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
-import static io.opencensus.tags.unsafe.ContextUtils.TAG_CONTEXT_KEY;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Stopwatch;
@@ -48,6 +47,7 @@ import io.opencensus.tags.Tagger;
 import io.opencensus.tags.Tags;
 import io.opencensus.tags.propagation.TagContextBinarySerializer;
 import io.opencensus.tags.propagation.TagContextSerializationException;
+import io.opencensus.tags.unsafe.ContextUtils;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
@@ -654,7 +654,7 @@ public final class CensusStatsModule {
     @Override
     public Context filterContext(Context context) {
       if (!module.tagger.empty().equals(parentCtx)) {
-        return context.withValue(TAG_CONTEXT_KEY, parentCtx);
+        return ContextUtils.withValue(context, parentCtx);
       }
       return context;
     }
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/grpc/internal/CensusStatsModule.java
```diff
@@ -356,7 +356,7 @@ public void outboundMessage(int seqNo) {
       this.parentCtx = checkNotNull(parentCtx);
       TagValue methodTag = TagValue.create(fullMethodName);
       this.startCtx = module.tagger.toBuilder(parentCtx)
-          .put(DeprecatedCensusConstants.RPC_METHOD, methodTag)
+          .putPropagating(DeprecatedCensusConstants.RPC_METHOD, methodTag)
```
songy23: For backwards-compatibility these tags are still treated as propagating tags. We could make them non-propagating (they should be non-propagating eventually) though that may be breaking. WDYT?

core/src/main/java/io/grpc/internal/CensusTracingModule.java
```diff
@@ -382,7 +382,8 @@ public ServerStreamTracer newServerStreamTracer(String fullMethodName, Metadata
       // Safe usage of the unsafe trace API because CONTEXT_SPAN_KEY.get() returns the same value
       // as Tracer.getCurrentSpan() except when no value available when the return value is null
       // for the direct access and BlankSpan when Tracer API is used.
-      final ClientCallTracer tracerFactory = newClientCallTracer(CONTEXT_SPAN_KEY.get(), method);
+      final ClientCallTracer tracerFactory =
+          newClientCallTracer(ContextUtils.getValue(Context.current()), method);
```
songy23: Note `ContextUtils.getValue()` will always return non-null, while previously `CONTEXT_SPAN_KEY.get()` may return null.

core/src/main/java/io/grpc/internal/CensusStatsModule.java
zhangkun83: Although I don't know anyone who depends on this being propagated (gRPC doesn't), it's safer to keep the current behavior in this PR, and change it to non-propagating in a separate PR.

core/src/main/java/io/grpc/internal/CensusTracingModule.java
zhangkun83: The returned `Span` eventually goes to `spanBuilderWithExplicitParent()`. Would it behavior the same with `null` vs the non-null value returned by `ContextUtils.getValue()`?

core/src/main/java/io/grpc/internal/CensusStatsModule.java
songy23: SGTM, I'll send a follow-up PR for it.

core/src/main/java/io/grpc/internal/CensusTracingModule.java
songy23: Yes the behavior will be the same. Instead of returning `null` previously, `ContextUtils.getValue()` now returns a blank (invalid) span. Both `null` and invalid span are handled in the same way in `spanBuilderWithExplicitParent()`: https://github.com/census-instrumentation/opencensus-java/blob/7b9ea2251c63852bf5e0633d402600c90822de04/impl_core/src/main/java/io/opencensus/implcore/trace/SpanBuilderImpl.java#L76.

songy23:
CI failure seems unrelated to this change: 
```bash
io.grpc.testing.integration.ProxyTest > largeBandwidth FAILED
    java.lang.AssertionError: expected:<1.048576E7> but was:<1.6053392668863518E7>
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:553)
        at org.junit.Assert.assertEquals(Assert.java:683)
        at io.grpc.testing.integration.ProxyTest.largeBandwidth(ProxyTest.java:184)
``` 

Please help me merge this PR - thanks.

PR #5663, opened 2019-05-02
core,grpclb: refactor SubchannelPool to SubchannelPoolHelper

dapengzhang0:
- `SubchannelPool` class is moved from `grpclb` to `internal`.
- Add `Helper.shutdown()` that is equivalent to `SubchannelPool.clear()`
- Right now still expose `SubchannelPool` class to `grpclb` (for testing only) because `GrpclbLoadBalancerTest` was heavily testing actions on `SubchannelPool`. We could do further so that `SubchannelPool` is trashed/package private and `grpclb` only uses `SubchannelPoolHelper`, but that would be more work than just refactoring because we need to change the way of testing `GrpclbLoadBalancer` and `GrpclbState` (namely, "Test behaviors, not methods").

### DIFF
--- a/core/src/main/java/io/grpc/internal/CachedSubchannelPool.java
+++ b/core/src/main/java/io/grpc/internal/CachedSubchannelPool.java
@@ -14,14 +14,13 @@
  * limitations under the License.
  */
 
-package io.grpc.grpclb;
+package io.grpc.internal;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.annotations.VisibleForTesting;
-import io.grpc.Attributes;
 import io.grpc.ConnectivityStateInfo;
 import io.grpc.EquivalentAddressGroup;
 import io.grpc.LoadBalancer.CreateSubchannelArgs;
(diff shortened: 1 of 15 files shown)


core/src/main/java/io/grpc/internal/CachedSubchannelPool.java
```diff
@@ -30,14 +29,16 @@
...
 import java.util.concurrent.TimeUnit;
 
 /**
  * A {@link SubchannelPool} that keeps returned {@link Subchannel}s for a given time before it's
  * shut down by the pool.
  */
-final class CachedSubchannelPool implements SubchannelPool {
-  private final HashMap<EquivalentAddressGroup, CacheEntry> cache =
+// TODO: make it package private. Right now it has to be public for grpclb testing.
+public final class CachedSubchannelPool implements SubchannelPool {
```
zhangkun83: I would prefer keep `CachedSubchannelPool` in place and copy the implementation in `SubchannelPoolHelper`. There is more needed to implement in the Helper, like `updateSubchannelAddresses()`, so that it's a semantic-complete `Helper` implementation.

dapengzhang0:
Close the current PR and track this in #5674.

>I would prefer keep CachedSubchannelPool in place and copy the implementation in SubchannelPoolHelper. There is more needed to implement in the Helper, like updateSubchannelAddresses(), so that it's a semantic-complete Helper implementation.

PR #5664, opened 2019-05-02
 api: NameResolver.Args deprecates NameResolver.Helper.

zhangkun83:
Having two `Helper`s (the other being `LoadBalancer.Helper`) is proven to
be confusing, and the one in `NameResolver` is fundamentally different
from the one from `LoadBalancer`, as the latter mutates the states while
the former doesn't.  Renaming it to `Args` make it less awkward to
expose it to `LoadBalancer`, which is needed for creating `NameResolver`s
for per-locality routing in the `XdsLoadBalancer`.


api/src/main/java/io/grpc/NameResolver.java
```diff
@@ -360,6 +399,186 @@ public ConfigOrError parseServiceConfig(Map<String, ?> rawServiceConfig) {
...
+      return new Builder();
+    }
+
+    /**
+     * Builder for {@link Args}.
+     *
+     * @since 1.21.0
+     */
+    public static final class Builder {
+      private Integer defaultPort;
```
ejona86: We may want to set defaults for multiple of these, like defaultPort and proxyDetector. But this is good at this point.

Issue #5665, opened 2019-05-03
Authentication example is presently harmful


ejona86:
The example is currently equivalent to the "header" example. So there's no possible value-add it could contribute at the moment.

But unfortunately it is recommending [making a ClientInterceptor](https://github.com/grpc/grpc-java/blob/v1.20.0/examples/src/main/java/io/grpc/examples/authentication/JwtClientInterceptor.java) for sending JWT to the server. That's not what we recommend. We would recommend using `MoreCallCredentials.from(Credential)` along with [ServiceAccountJwtAccessCredentials](https://github.com/googleapis/google-auth-library-java/blob/v0.15.0/oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java) (which is Google-centric, but users can potentially use the `fromPkcs8()` construction method) or to implement `CallCredentials` directly. `CallCredentials` has the advantage that the credential 1) can verify its security requirements, like how all JWTs should be sent on an encrypted connection, and 2) will be fresher, as it is called just before the RPC is sent (a waitForReady RPC could be delayed for _days_).

The example also needs to incorporate how to propagate the authenticated user to the application via Context. Ideally it would at least note how to fail the RPC if the user is unauthenticated. Examples of how to do this is [available on SO](https://stackoverflow.com/questions/40112374/how-do-i-access-request-metadata-for-a-java-grpc-service-i-am-defining/40113309#40113309).

I'm considering deleting the example for the upcoming v1.21.0 release.

ejona86:
It also is using the Metadata Key name "jwt" for the JWT. However, it seems JWT typically uses the Authentication header with the Bearer prefix.

sanjaypujare:
@ejona86 [io.grpc.ClientInterceptor](https://github.com/grpc/grpc-java/blob/master/api/src/main/java/io/grpc/ClientInterceptor.java) has this comment:

>Implementers use this mechanism to add cross-cutting behavior to Channel and stub implementations. Common examples of such behavior include:
...
Adding credentials to header metadata

Whether you call it a recommendation or not, it documents a use case for client-interceptors especially the cross-cutting aspect, so the example was used to illustrate that aspect more than how to properly and securely implement authentication. Granted the example is simplistic and functionally somewhat similar to the "header" example (although that example shows the server sending a header in the response to the client). 

[...]

PR #5666, opened 2019-05-03
api: Discourage using ClientInterceptor for credentials

ejona86:
As brought up in #5665

### DIFF
--- a/api/src/main/java/io/grpc/ClientInterceptor.java
+++ b/api/src/main/java/io/grpc/ClientInterceptor.java
@@ -24,10 +24,14 @@ import javax.annotation.concurrent.ThreadSafe;
  * <p>Implementers use this mechanism to add cross-cutting behavior to {@link Channel} and
  * stub implementations. Common examples of such behavior include:
  * <ul>
- * <li>Adding credentials to header metadata</li>
  * <li>Logging and monitoring call behavior</li>
+ * <li>Adding metadata for proxies to observe</li>
  * <li>Request and response rewriting</li>
  * </ul>
+ *
+ * <p>Providing authentication credentials is better served by {@link
+ * CallCredentials}. But a {@code ClientInterceptor} could set the {@code
+ * CallCredentials} within the {@link CallOptions}.
  */
 @ThreadSafe
 public interface ClientInterceptor {


api/src/main/java/io/grpc/ClientInterceptor.java
```diff
@@ -24,10 +24,14 @@
...
  * <ul>
- * <li>Adding credentials to header metadata</li>
  * <li>Logging and monitoring call behavior</li>
+ * <li>Adding metadata for proxies to observe</li>
  * <li>Request and response rewriting</li>
  * </ul>
+ *
+ * <p>Providing authentication credentials is better served by {@link
+ * CallCredentials}. But a {@code ClientInterceptor} could set the {@code
+ * CallCredentials} within the {@link CallOptions}.
```
sanjaypujare: Should we say something about whether this is undesirable or not?

api/src/main/java/io/grpc/ClientInterceptor.java
```diff
@@ -24,10 +24,14 @@
  * <p>Implementers use this mechanism to add cross-cutting behavior to {@link Channel} and
  * stub implementations. Common examples of such behavior include:
  * <ul>
- * <li>Adding credentials to header metadata</li>
  * <li>Logging and monitoring call behavior</li>
+ * <li>Adding metadata for proxies to observe</li>
```
sanjaypujare: Metadata can be added for anything - why limit it to "for proxies to observe"?  (I know it's only an example but still...)

api/src/main/java/io/grpc/ClientInterceptor.java
ejona86: ClientInterceptor can work. But using CallCredentials is easier and generally better. It's sort of in-depth, so didn't want to really get into it in this documentation.

We could end up putting some benefits of CallCredentials in the CallCredentials class though.

api/src/main/java/io/grpc/ClientInterceptor.java
ejona86: These are just examples. There are other possibilities. The listed items are use-cases, not just functionality that is exposed. When someone starts reading this class, I wouldn't expect them to know what metadata is and why it is used.

Issue #3084, opened 2017-06-09
Potential deadlock due to calling callbacks while holding a lock


zhangkun83:
`InProcessClientStream` and `InProcessServerStream` are synchronized on their own. `InProcessClientStream.serverStreamListener` is called under `synchronized (InProcessClientStream.this)`, and vice versa.

If the application tries to call methods on `ClientCall` or `ServerCall` from within the callbacks (assuming that it has already taken care of the thread-safety of the method calls on "Call" objects), a deadlock is possible when direct executor is used. For example:
- Thread1
  1. `InProcessClientStream.serverRequested` (locks `InProcessClientStream.this`)
  1. `InProcessClientStream.serverStreamListener.messageRead()`
  1. Eventually reaches application callback, which calls `ServerCall.close()`
  1. `InProcessServerStream.close()` (locks `InProcessServerStream.this`)

- Thread2
  1. `InProcessServerStream.clientRequested` (locks `InProcessServerStream.this`)
  1. `InProcessServerStream.clientStreamListener.messageRead()`
  1. Eventually reaches application callback, which calls `ClientCall.close()`
  1. `InProcessClientStream.close()` (locks `InProcessClientStream.this`)

As locks are acquired in reverse orders from two threads, a deadlock is possible.

The fundamental issue is that we should not call into application code while holding a lock, because we don't know what application code can do thus we can't control the order of subsequent locking.

OkHttp has the same issue, because `OkHttpClientStream.transportDataReceived()`, which will call into application code, is called under lock.

We could use `ChannelExecutor` (maybe renamed) to prevent calling into callbacks while holding a lock.

ArvinDevel:
I recently ran into a problem which is related to `InProcessServerStream ` and `InProcessClientStream ` deadlock.  When my server app thread response to client request, it will first lock `InProcessServerStream`, then try to lock `InProcessClientStream` which is locked by grpc-thread which wants to lock  `InProcessServerStream` at the same time. 
Does this mean I can only use the same grpc thread to response?

ArvinDevel:
Output of jstack related to deadlock is below.
===================================================
"grpc-default-executor-1":
	at io.grpc.inprocess.InProcessTransport$InProcessStream$InProcessServerStream.request(InProcessTransport.java:324)
	- waiting to lock <0x00000007b7f8d6f8> (a io.grpc.inprocess.InProcessTransport$InProcessStream$InProcessServerStream)
[...]

Issue #2263, opened 2016-09-13
Update examples to use Executor


carl-mastrangelo:
Our hello world examples don't show the right way to create a server.   Fix them

nevi-me:
I keep getting the following error when trying to start grpc from a jar file:

```
Sep 15, 2016 8:25:01 AM io.grpc.netty.NettyServerHandler onStreamError
WARNING: Stream Error
io.netty.handler.codec.http2.Http2Exception$StreamException: com.google.common.util.concurrent.MoreExecutors.directExecutor()Ljava/util/concurrent/Executor;
        at io.netty.handler.codec.http2.Http2Exception.streamError(Http2Exception.java:147)
        at io.grpc.netty.NettyServerHandler.newStreamException(NettyServerHandler.java:437)
        at io.grpc.netty.NettyServerHandler.onHeadersRead(NettyServerHandler.java:210)
        at io.grpc.netty.NettyServerHandler.access$300(NettyServerHandler.java:93)
        at io.grpc.netty.NettyServerHandler$FrameListener.onHeadersRead(NettyServerHandler.java:459)
        at io.netty.handler.codec.http2.DefaultHttp2ConnectionDecoder$FrameReadListener.onHeadersRead(DefaultHttp2ConnectionDecoder.java:335)
        at io.netty.handler.codec.http2.DefaultHttp2ConnectionDecoder$FrameReadListener.onHeadersRead(DefaultHttp2ConnectionDecoder.java:285)
        at io.netty.handler.codec.http2.Http2InboundFrameLogger$1.onHeadersRead(Http2InboundFrameLogger.java:56)
        at io.netty.handler.codec.http2.DefaultHttp2FrameReader$2.processFragment(DefaultHttp2FrameReader.java:467)
        at io.netty.handler.codec.http2.DefaultHttp2FrameReader.readHeadersFrame(DefaultHttp2FrameReader.java:475)
        at io.netty.handler.codec.http2.DefaultHttp2FrameReader.processPayloadState(DefaultHttp2FrameReader.java:247)
        at io.netty.handler.codec.http2.DefaultHttp2FrameReader.readFrame(DefaultHttp2FrameReader.java:155)
        at io.netty.handler.codec.http2.Http2InboundFrameLogger.readFrame(Http2InboundFrameLogger.java:41)
        at io.netty.handler.codec.http2.DefaultHttp2ConnectionDecoder.decodeFrame(DefaultHttp2ConnectionDecoder.java:113)
[...]

PR #5672, opened 2019-05-06
docs: add note about directExecutor in example tests.

rubensayshi:
Since the HelloWorld tests are atm the "documentation" for how to do unit tests for grpc I think it would be good to add a note about the danger of the `directExecutor()`, particularly in bidirectional streaming.

I considered removing the `directExecutor ` but it seems a common pattern in the grpc test suite and has it's merits so warning devs seems like a better solution.

related: 
 - https://github.com/grpc/grpc-java/issues/3084
 - https://github.com/grpc/grpc-java/issues/2263


rubensayshi:
@ejona86 okay, updated with your suggestion.

ejona86:
@zhangkun83, could you review as well?

ejona86:
@rubensayshi, thank you!

PR #5673, opened 2019-05-06
[on hold] xds: use SubchannelStateListener for fallback balancer

dapengzhang0:


dapengzhang0:
Close current PR and track by #5675

PR #5678, opened 2019-05-07
context: drop call to actual() in DeadlineSubject

carl-mastrangelo:


carl-mastrangelo:
The generics on Subject are going to be changing; this is a mirror of the downstream change internally.

PR #5682, opened 2019-05-07
api: Rename NameResolver.Observer to Listener2

ejona86:
This follows the sort of changes we've done in the past, where the '2'
implies "version 2." We can end up reclaiming the original name if we
wish in the future.

The main reason for this change is to avoid changing to Observer since
the rest of io.grpc consistently uses Listener.

CC @zhangkun83


dapengzhang0:
Is it true that FooBar2 is never going to be a stable API, and that it will eventually be renamed to `FooBar` and then go to stable?

ejona86:
@dapengzhang0, my current plan is to reclaim the original name, and end up trashing FooBar2. But FooBar2 could live for a year in a deprecated state and not hurt anyone. I can also live with staying with the Listener2 name.

PR #5684, opened 2019-05-08
Revert "core: pass Subchannel state updates to SubchannelStateListene…

### LINKED ISSUE #5676: Subchannel identity crisis
Since we are doing lot of wrapping on the LoadBalancer API, and Subchannels are passed through various interfaces, a wrapped Subchannel will require a lot of care to make sure it's consistent on all APIs, e.g.,
1. The Subchannel from `createSubchannel()` should be the same objects that's passed to `SubchannelStateListener` on every wrapping level.
1. An upcoming ORCA load reporting API is another example (https://github.com/grpc/grpc-java/pull/5639#discussion_r280992684).
1. The `Subchannel` from `PickResult` will have to be unwrapped as well. `ManagedChannelImpl` assumes the `Subchannel` from `PickResult` is the real "SubchannelImpl".

Possible solutions:
1. Move `LoadBalancer.Helper` methods that accepts `Subchannel` to the `Subchannel` class, e.g., `updateSubchannelAddresses()`. Stop passing `Subchannel` to the listeners. But that requires a `start()`, which is a non-trivial API change on the current `Subchannel`, but could be considered for new APIs.
2. The opposite of above -- move all state-mutating methods, e.g., `shutdown()`, `requestConnection()` to `Helper`, eliminating the need for wrapping `Subchannel`.
3. Provide a util to help correctly wrap `Subchannel` while keeping the consistency.
4. Add another identity object to `Subchannel`. This could also used by `ManagedChannelImpl` to retrieve the real "SubchannelImpl" from `PickResult`.

Solutions 2 and 5 wouldn't go along with our plan of changing Subchannel attributes to an object, because a wrapping Helper can only add its own stuff on that object by wrapping it, which ends up the need for unwrapping it on the Subchannel. Otherwise they seem to work well.

@ejona86 @voidzcy

zhangkun83:
…r rather than LoadBalancer (#5503)"

This reverts commit 62b03fd7e6936b7b1e78e4f13e9944d5a6da0771.

Effectively reverts its follow-up commits:
dc218b6d4d212d52688b7b0be7a13db7f1de1e7f
405d8c3865fa31132db08dc5b5609dff3b2fa0f8
44840fe813764167fc10102bf1fbe9fa1a8f5f5a

In order to fix #5676, we need to make a tweak to `SubchannelStateListener` and some other APIs. Since it has never been in any releases, it's better to avoid it in the upcoming 1.21 so that we don't have to deal with yet another migration.

### DIFF
--- a/api/src/main/java/io/grpc/LoadBalancer.java
+++ b/api/src/main/java/io/grpc/LoadBalancer.java
@@ -16,14 +16,12 @@
 
 package io.grpc;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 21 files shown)


zhangkun83:
@voidzcy, this will break #5639. Don't try to fix it yet. I will re-introduce `SubchannelStateListener` in the better way and you can use that.

/cc @dapengzhang0

PR #5686, opened 2019-05-08
PerfMark:  added for specifically supported packages to access PerfMark functions

ran-su:


carl-mastrangelo:
You'll need to update the commit message to be in the same format as the rest of our commits.  Otherwise LGTM

core/src/main/java/io/grpc/perfmark/InternalPerfMark.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private InternalPerfMark() {}
+
+    /** Expose class to allow packages that utilize PerfMark to get PerfMarkTask instances */
+    public abstract static class InternalPerfMarkTask extends PerfMarkTask {
+        public InternalPerfMarkTask() {}
+    }
+
+    /** Expose methods that create PerfTag to packages that utilize PerfMark */
+    private static final long NULL_NUMERIC_TAG = 0;
```
carl-mastrangelo: was this supposed to be public?

ran-su: We only need these two constants within this class.

PR #5689, opened 2019-05-09
core: Make OpenCensus tags non-propagating.

songy23:
The `method` and `status` tags shouldn't be propagated in the first place,
but in previous OpenCensus implementation all tags are propagating by
default. Now with the TagMetadata it may make sense to change them to
local tags.

This will be a breaking change to users who depend on the behavior that
these tags propagate through process boundaries.


songy23:
Just rebased against master. @zhangkun83 please help me trigger Kokoro build again, thanks.

dapengzhang0:
@zhangkun83 Merge?

songy23:
Thanks!

PR #5715, opened 2019-05-09
examples: make tls example easier to run

carl-mastrangelo:
* Make the server cert able to be verified by the ca cert in openssl
* Make the port number consistent in each example (easy to copy paste wrong one)
* use correct netty-tcnative

cc: @jtattermusch

### DIFF
--- a/examples/example-tls/README.md
+++ b/examples/example-tls/README.md
@@ -42,8 +42,8 @@ You can use the following script to generate self-signed certificates for grpc-j
 ```bash
 mkdir -p /tmp/sslcert
 pushd /tmp/sslcert
-# Changes these CN's to match your hosts in your environment if needed.
-SERVER_CA_CN=localhost-ca # must be different than SERVER_CN
+# Change these CN's to match your hosts in your environment if needed.
+SERVER_CA_CN=localhost-ca
 SERVER_CN=localhost
 CLIENT_CN=localhost # Used when doing mutual TLS
 
--- a/examples/example-tls/README.md
+++ b/examples/example-tls/README.md
@@ -43,6 +43,7 @@ You can use the following script to generate self-signed certificates for grpc-j
 mkdir -p /tmp/sslcert
 pushd /tmp/sslcert
 # Changes these CN's to match your hosts in your environment if needed.
+SERVER_CA_CN=localhost-ca # must be different than SERVER_CN
 SERVER_CN=localhost
 CLIENT_CN=localhost # Used when doing mutual TLS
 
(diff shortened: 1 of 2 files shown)


examples/example-tls/README.md
```diff
@@ -43,14 +43,15 @@ You can use the following script to generate self-signed certificates for grpc-j
 mkdir -p /tmp/sslcert
 pushd /tmp/sslcert
 # Changes these CN's to match your hosts in your environment if needed.
```
sanjaypujare: Changes => Change

examples/example-tls/README.md
```diff
@@ -43,14 +43,15 @@ You can use the following script to generate self-signed certificates for grpc-j
...
+SERVER_CA_CN=localhost-ca # must be different than SERVER_CN
```
sanjaypujare: Do we need the comment "# must be different than SERVER_CN" ?

The comment is the justification for this change but why leave it in the README file?

carl-mastrangelo: I suppose not.

examples/example-tls/README.md
carl-mastrangelo: Done.

examples/example-tls/README.md
```diff
@@ -43,14 +43,15 @@ You can use the following script to generate self-signed certificates for grpc-j
...
 
 echo Generate CA key:
 openssl genrsa -passout pass:1111 -des3 -out ca.key 4096
 echo Generate CA certificate:
 # Generates ca.crt which is the trustCertCollectionFile
-openssl req -passin pass:1111 -new -x509 -days 365 -key ca.key -out ca.crt -subj "/CN=${SERVER_CN}"
+openssl req -passin pass:1111 -new -x509 -days 365 -key ca.key -out ca.crt -subj "/CN=${SERVER_CA_CN}"
 echo Generate server key:
 openssl genrsa -passout pass:1111 -des3 -out server.key 4096
 echo Generate server signing request:
```
sanjaypujare: Lines 59 and 68 say "Self-signed..." which is incorrect, so we can fix that?

carl-mastrangelo: what would you prefer?   I think self-signed is pretty clear, it means someone didnt give it to us.

sanjaypujare:
> * Make the ca cert able to be verified by the server cert in openssl

Shouldn't this be "Make the server cert able to be verified by the ca cert in openssl" ?

Also this change is not making any material difference to the code, right? Because each cert is in its own file (server.crt vs ca.crt) and we are separately passing ca.crt as the trust-store it doesn't really matter that we used the same name "localhost" as the CN for both certs. In any case the example wouldn't have worked if that was the problem. I am okay with this change, but just trying to clarify.

carl-mastrangelo:
> Shouldn't this be "Make the server cert able to be verified by the ca cert in openssl" ?

done.

> Also this change is not making any material difference to the code, right? Because each cert is in its own file (server.crt vs ca.crt) and we are separately passing ca.crt as the trust-store it doesn't really matter that we used the same name "localhost" as the CN for both certs.

In java it worked.   The origin of this PR is @jtattermusch trying out our example and finding it confusing and not working out of the box.  The certs could not be used by the C version of gRPC.

examples/example-tls/README.md
sanjaypujare: See https://en.wikipedia.org/wiki/Self-signed_certificate 

> In technical terms a self-signed certificate is one signed with its own private key.

Here server and client certs are signed by the ca cert's private key.

PR #5719, opened 2019-05-10
buildscripts: Specify Bazel version to use (v1.19.x)

ejona86:

------

This was done for v1.20.x and later in #5411

### DIFF
--- a/buildscripts/kokoro/bazel.sh
+++ b/buildscripts/kokoro/bazel.sh
@@ -4,6 +4,8 @@ set -exu -o pipefail
 cat /VERSION
 bazel version
 
+use_bazel.sh 0.19.0
+
 cd github/grpc-java
 bazel build ...
 


ejona86:
I've not sent this out for review yet. I was going to update 1.17 and earlier to Bazel 0.11.0, since that is already installed on Kokoro. But that version currently fails to start on Kokoro (complains about corruption). So now I think I'll have all the versions 1.19 and earlier use Bazel 0.19. But I had to create a PR to test it.

ejona86:
0.19 seems to be working in Kokoro. It'll just be auto-downloaded (vs being pre-installed like 0.11). I've rebased this on v1.19.x. I plan to backport this all the way back to v1.10.x

ejona86:
Didn't change anything except for the commitdate, to kick Travis out of a funk. It seems at least it was impacted by me changing the pr to be against v1.19 instead of v1.17.

PR #5736, opened 2019-05-15
api: make Subchannel.requestConnection() thread-safe.

zhangkun83:
This is called by pick-first LoadBalancer from the data-path, thus has
to be thread-safe.

This is a revision to #5718 thus has to be back-ported to 1.21.x as well.


zhangkun83:
Closing in favor of #5738

PR #5740, opened 2019-05-15
buildscripts: Specify Bazel version to use (1.16.x backport)

ejona86:

Backport of #5739

### DIFF
--- a/buildscripts/kokoro/bazel.sh
+++ b/buildscripts/kokoro/bazel.sh
@@ -3,6 +3,8 @@
 set -exu -o pipefail
 cat /VERSION
 
+use_bazel.sh 0.19.0
+
 cd github/grpc-java
 bazel build ...
 
--- a/buildscripts/kokoro/bazel.sh
+++ b/buildscripts/kokoro/bazel.sh
@@ -3,7 +3,7 @@
 set -exu -o pipefail
 cat /VERSION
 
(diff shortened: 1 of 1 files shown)


nicolasnoble:
Wait... bazel 0.11? Why so old?

ejona86:
@nicolasnoble, because it fails to build (build --nobuild works though) with 0.19.0 because of compiler warnings. And IIRC Kokoro was on 0.11.0 for a long time and 0.11.0 is still installed by default on the Kokoro machines. So I think at the time these were using 0.11.0 to build. (Which... now that Kokoro fails I remember that the most recent Kokoro image has a corrupt Bazel 0.11.0 installation. /sigh)

I'm trying to fix things all the way back to grpc-java 0.10.0, which was Feb 2018. Bazel 0.11.0 was released Feb 2018. So the versions line up about right.

Doing a binary search, it appears 0.16 is the most recent version that builds for grpc v1.16. But it fails for grpc v1.10. And bazel 0.14 also fails for grpc v1.10. I think I'm going to swap to 0.12.0, which is the closest I can get to 0.11.0 and works for both grpc v1.10 and v1.16.

ejona86:
@dapengzhang0, okay. Let's try this again. Once this is merged I'll update the other PRs to bazel 0.12.0.

PR #5745, opened 2019-05-15
buildscripts: Specify Bazel version to use (1.11.x backport)

ejona86:

Backport of #5719


dapengzhang0:
Bazel build is failing though.

PR #5753, opened 2019-05-16
examples: Use encouraged APIs and avoid over-specifying configuration

ejona86:
Also fix tcnative, which was the wrong version and would cause a crash
with NoSuchMethodError.

As noticed when looking into #5737

### DIFF
--- a/examples/example-tls/README.md
+++ b/examples/example-tls/README.md
@@ -24,7 +24,7 @@ Running the hello world with TLS is the same as the normal hello world, but take
 **hello-world-tls-server**:
 
 ```text
-USAGE: HelloWorldServerTls host port certChainFilePath privateKeyFilePath [trustCertCollectionFilePath]
+USAGE: HelloWorldServerTls port certChainFilePath privateKeyFilePath [trustCertCollectionFilePath]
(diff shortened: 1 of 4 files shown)


sanjaypujare:
Originally the example used "localhost" as the hostname forcing client and server to run on the same host (and where localhost actually resolves to your local machine which is not always the case). I thought the right enhancement for the example would have been to remove that restriction and allow client and server to be run on different hosts (or docker containers etc) to make the example more useful and general. But this change goes in the opposite direction where localhost is assumed and removed as a configuration parameter so I think that makes the example somewhat limited.

ejona86:
> Originally the example used "localhost" as the hostname forcing client and server to run on the same host (and where localhost actually resolves to your local machine which is not always the case).

You change one line when generating the certificates (which is commented as a place to change) and execute the binaries with the right name and it works.

> But this change goes in the opposite direction where localhost is assumed and removed as a configuration parameter so I think that makes the example somewhat limited.

Not so. It actually makes it easier. The server was only binding to one IP address. Now it binds to all addresses. The server doesn't need to know its own hostname.

PR #5784, opened 2019-05-23
xds: handle 100% drop for fallback mode

dapengzhang0:
- Cancel fallback timer and/or exit fallback mode once receiving an EDS response indicating 100% drop.
- Also update balancing state once receiving the first EDS response with drop information when the channel is at the initial IDLE state.

### DIFF
--- a/xds/src/main/java/io/grpc/xds/LocalityStore.java
+++ b/xds/src/main/java/io/grpc/xds/LocalityStore.java
@@ -289,6 +289,9 @@ interface LocalityStore {
 
       if (dropOverloads != null && !dropOverloads.isEmpty()) {
         picker = new DroppablePicker(dropOverloads, picker, random);
+        if (state == null) {
+          state = IDLE;
+        }
       }
 
       if (state != null) {
--- a/xds/src/main/java/io/grpc/xds/LocalityStore.java
+++ b/xds/src/main/java/io/grpc/xds/LocalityStore.java
@@ -53,6 +53,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import javax.annotation.Nullable;
 
 /**
  * Manages EAG and locality info for a collection of subchannels, not including subchannels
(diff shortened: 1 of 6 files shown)


xds/src/main/java/io/grpc/xds/LocalityStore.java
```diff
@@ -285,6 +285,9 @@ private void updatePicker(ConnectivityState state,  List<WeightedChildPicker> ch
 
       if (!dropOverloads.isEmpty()) {
         picker = new DroppablePicker(dropOverloads, picker, random);
+        if (state == null) {
```
zhangkun83: Should probably add a `@Nullable` to the `state` argument

xds/src/test/java/io/grpc/xds/LocalityStoreTest.java
```diff
@@ -274,8 +278,17 @@ public PickResult pickSubchannel(PickSubchannelArgs args) {
     verify(random, never()).nextInt(1000_000);
   }
 
+  @Test
+  public void updateLoaclityStore_withEmptyDropList() {
+    localityStore.updateDropPercentage(ImmutableList.<DropOverload>of());
```
zhangkun83: Isn't empty list the default value already? Should `updateLoaclityStore()` be renamed to `updateLoaclityStore_withEmptyDropList()`?

xds/src/test/java/io/grpc/xds/XdsCommsTest.java
```diff
@@ -284,8 +290,11 @@ public void standardMode_sendEdsRequest_getEdsResponse_withNoDrop() {
         .build();
     responseWriter.onNext(edsResponse);
 
-    verify(localityStore, times(2)).updateDropPercentage(ImmutableList.<DropOverload>of());
-    verify(localityStore, times(2)).updateLocalityStore(localityEndpointsMappingCaptor.capture());
+    verify(adsStreamCallback, times(1)).onWorking();
+    verify(adsStreamCallback, never()).onAllDrop();
+    verify(adsStreamCallback, never()).onError();
```
zhangkun83: Can the above two lines be replaced by `verifyNoMoreInteractions()`? (Ditto for the same cases below)

Issue #1549, opened 2016-03-14
Simplify implementation of back-pressure in generated stubs


jhump:
Pending API changes can allow reactive/async pattern for interacting with flow control and applying back pressure: https://github.com/grpc/grpc-java/pull/1545/files

In many cases, automatic back-pressure in generated stubs could be very useful -- e.g. having calls to `StreamObserve#onNext(T)` block instead of queueing.

It's been pointed out that this could cause deadlock for bidi-streaming operations, so perhaps we can just not expose this functionality for bidi-streaming calls?

It may also be worth pointing out that most other runtimes (wrapped languages and Go) already expose streams via blocking operations and already require that apps be aware of and work-around possible deadlock issues resulting therefrom. So maybe providing similar mechanisms in Java is fine, with said caveats.

Another possible alternative could possibly be done in an extension/add-on instead of in GRPC. For example, wrapping streaming requests and responses with RxJava `Observables` may further simplify the async case enough to make the synchronous (and possibly-deadlock-prone) case unnecessary.

ejona86:
> It may also be worth pointing out that most other runtimes (wrapped languages and Go) already expose streams via blocking operations and already require that apps be aware of and work-around possible deadlock issues resulting therefrom. So maybe providing similar mechanisms in Java is fine, with said caveats.

This could possibly be as an alternative to StreamObserver. For example, we could make a more robust blocking API that could handle bidi. To my knowledge, the various languages' APIs are either completely blocking or completely async; they don't "randomly" block.

But alternatively, simply having the blocking behavior be opt-in could be enough to address deadlock concerns.

ejona86:
The StreamObserver used for sending messages currently completely ignores outbound flow control. The original argument for this is that most applications don't care. While not debating that point, it does seem enough applications do care that they experience a rude awakening.

In discussing with @louiscryan, it seems we are agreeing that it may have been a misstep to ignore flow control by default. If doing it again, we may have had blocking for sending enabled by default. However, changing the behavior now would impact existing applications and it isn't clear how much pain that would cause our existing users. So it may be prudent to introduce something new that we encourage users to migrate to.

In any case, we do still feel the need for a non-blocking variant, where the application should use `isReady()`/`onReady()` (which is not yet available on the async client stub).

[...]

PR #5799, opened 2019-05-29
api,stub: Clarify isReady()/onReady() interaction semantics

ejona86:
https://github.com/grpc/grpc-java/issues/1549#issuecomment-496564794 brought up that the documentation didn't really specify what it guaranteed.

CC @ulfjack


ulfjack:
Looks good, thanks!

PR #5802, opened 2019-05-29
api: deprecate Helper.updateSubchannelAddresses() and add equivalent on Subchannel

### LINKED ISSUE #5676: Subchannel identity crisis

zhangkun83:
Resolves #5676

### DIFF
--- a/api/src/main/java/io/grpc/LoadBalancer.java
+++ b/api/src/main/java/io/grpc/LoadBalancer.java
@@ -963,7 +963,9 @@ public abstract class LoadBalancer {
      * href="https://github.com/grpc/grpc-java/issues/5015">#5015</a> for the background.
      *
      * @since 1.4.0
+     * @deprecated use {@link Subchannel#updateAddresses} instead
      */
+    @Deprecated
     public final void updateSubchannelAddresses(
         Subchannel subchannel, EquivalentAddressGroup addrs) {
       checkNotNull(addrs, "addrs");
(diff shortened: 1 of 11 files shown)


api/src/main/java/io/grpc/LoadBalancer.java
```diff
@@ -1301,6 +1305,19 @@ public ChannelLogger getChannelLogger() {
       throw new UnsupportedOperationException();
     }
 
+    /**
+     * Replaces the existing addresses used with this {@code Subchannel}. If the new and old
+     * addresses overlap, the Subchannel can continue using an existing connection.
+     *
+     * <p>It must be called from the Synchronization Context or will will throw.
```
ejona86: s/will will/will/

PR #5803, opened 2019-05-30
core: only let ManagedChannelImpl convert empty resolution result to error

### LINKED ISSUE #5692: NameResolver refresh not triggered if empty addresses returned

1.20.0

I've implemented a custom NameResolver to resolve addresses in Consul (DNS or HTTP). We're pairing this name resolver with the RoundRobinLoadBalancer to provide simple client-side load balancing.

While integration testing, I found that if we return an empty list of addresses in the name resolver, no refresh is ever triggered on the NameResolver (leading to a permanent error of UNAVAILABLE for all calls). If I change the NameResolver implementation to call Listener.onError() instead when no addresses are found, refresh() will eventually be triggered and the problem can resolve itself.

When writing the name resolver, I believe I may have gotten tripped up on this comment (which makes it sound like onAddresses() with an empty list is equivalent to onError()): https://github.com/grpc/grpc-java/blob/980dd9671d26ab3dfb0cadfe5f226d840d4ac124/api/src/main/java/io/grpc/NameResolver.java#L285

For reference, the DNS name resolver doesn't pass an empty list of addresses: https://github.com/grpc/grpc-java/blob/206a2e39333e991963adc7da74cb05dd2410c1a5/core/src/main/java/io/grpc/internal/DnsNameResolver.java#L282-L286

I'm not sure if this is expected behavior and I know this is an experimental API, but wanted to report in case this wasn't the desired behavior.

zhangkun83:
Previously, AutoConfiguredLoadBalancer was also handling it, but it
doesn't trigger retries.  By returning true for
canHandleEmptyAddressListFromNameResolution(),
AutoConfiguredLoadBalancer effectively by-passed the empty-result
handling logic in ManagedChannelImpl, thus resolution retries were
never triggered.

This change requires AutoConfiguredLoadBalancer to stop being a
LoadBalancer, for its tryHandleResolvedAddresses().  It doesn't
cause any trouble because AutoConfiguredLoadBalancer has become
less and less like a LoadBalancer during the service config changes.

Resolves #5692

### DIFF
--- a/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
+++ b/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
@@ -17,6 +17,7 @@
 package io.grpc.internal;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static io.grpc.LoadBalancer.ATTR_LOAD_BALANCING_CONFIG;
 
 import com.google.common.annotations.VisibleForTesting;
 import io.grpc.Attributes;
@@ -28,6 +29,8 @@ import io.grpc.LoadBalancer;
 import io.grpc.LoadBalancer.Helper;
 import io.grpc.LoadBalancer.PickResult;
 import io.grpc.LoadBalancer.PickSubchannelArgs;
+import io.grpc.LoadBalancer.ResolvedAddresses;
+import io.grpc.LoadBalancer.Subchannel;
 import io.grpc.LoadBalancer.SubchannelPicker;
 import io.grpc.LoadBalancerProvider;
 import io.grpc.LoadBalancerRegistry;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -144,48 +148,38 @@ public void handleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
...
-    @Override
-    public void handleNameResolutionError(Status error) {
+    void handleNameResolutionError(Status error) {
       getDelegate().handleNameResolutionError(error);
     }
 
     @Deprecated
-    @Override
-    public void handleSubchannelState(Subchannel subchannel, ConnectivityStateInfo stateInfo) {
+    void handleSubchannelState(Subchannel subchannel, ConnectivityStateInfo stateInfo) {
```
carl-mastrangelo: why cant this be removed yet?

zhangkun83: LoadBalancers still using the old `createSubchannel()` will still receive subchannel state updates from this path. It will be removed once `LoadBalancer#handleSubchannel` is removed.

PR #5823, opened 2019-06-03
xds: allow grpclb balancer addresses for backward compatibility

dapengzhang0:
During migration, the name resolver may not know when the client has been upgraded to xds, so it may still send grpclb v1 addresses with a list of policies including both grpclb v1 and xds.

### DIFF
--- a/xds/src/main/java/io/grpc/xds/XdsLoadBalancer.java
+++ b/xds/src/main/java/io/grpc/xds/XdsLoadBalancer.java
@@ -316,7 +316,7 @@ final class XdsLoadBalancer extends LoadBalancer {
       fallbackBalancer = lbRegistry.getProvider(fallbackPolicy.getPolicyName())
           .newLoadBalancer(fallbackBalancerHelper);
       fallbackBalancerHelper.balancer = fallbackBalancer;
-      handleFallbackAddressesOrNameResolutionError();
+      propagateFallbackAddresses();
     }
 
     void updateFallbackServers(
@@ -331,7 +331,7 @@ final class XdsLoadBalancer extends LoadBalancer {
       this.fallbackPolicy = fallbackPolicy;
       if (fallbackBalancer != null) {
         if (fallbackPolicy.getPolicyName().equals(currentFallbackPolicy.getPolicyName())) {
-          handleFallbackAddressesOrNameResolutionError();
+          propagateFallbackAddresses();
         } else {
           fallbackBalancer.shutdown();
           fallbackBalancer = null;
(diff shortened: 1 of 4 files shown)


zhangkun83:
This work is done in the XdsLoadBalancer -- doesn't it require client to be upgraded to xDS in the first place?

xds/src/main/java/io/grpc/xds/XdsLoadBalancer.java
```diff
@@ -326,14 +328,24 @@ void updateFallbackServers(
         List<EquivalentAddressGroup> servers, Attributes attributes,
         LbConfig fallbackPolicy) {
       this.fallbackServers = servers;
+      String fallbackPolicyName = fallbackPolicy.getPolicyName();
+      if (!fallbackPolicyName.equals("grpclb") && !fallbackPolicyName.equals(XDS_POLICY_NAME)) {
```
zhangkun83: Probably worth a comment here.

dapengzhang0: Done.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer.java
```diff
@@ -360,6 +340,38 @@ void updateFallbackServers(
...
+        ImmutableList.Builder<EquivalentAddressGroup> backends = ImmutableList.builder();
+        for (EquivalentAddressGroup eag : fallbackServers) {
+          if (eag.getAttributes().get(GrpcAttributes.ATTR_LB_ADDR_AUTHORITY) == null) {
+            backends.add(eag);
+          }
+        }
+        servers = backends.build();
+      }
+
+      // FIXME: this is a temporary hack.
```
zhangkun83: Please add TODO(zhangkun83) and refer to https://github.com/grpc/grpc-java/issues/5496

xds/src/main/java/io/grpc/xds/XdsLoadBalancer.java
```diff
@@ -360,6 +340,38 @@ void updateFallbackServers(
       }
     }
 
+    private void handleFallbackAddressesOrNameResolutionError() {
```
zhangkun83: Why does the method name contain "NameResolutionError"? It's only called from fallback-related methods. What this method does is passing the latest fallback addresses to the fallback balancer. The handling of empty addresses is just part of it. Maybe "propagateFallbackAddresses()".

xds/src/main/java/io/grpc/xds/XdsLoadBalancer.java
dapengzhang0: Done.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer.java
dapengzhang0: Renamed

PR #5825, opened 2019-06-04
Add io.perfmark Annotations

carl-mastrangelo:
This add perfmark annotations in some key places, notably on transport/application boundaries, and thread hop locations.   Perfmark records to a thread-local buffer the events that happen in each thread.  Perfmark is disabled by default, and will compile to a noop unless `Perfmark.setEnabled` is invoked.  This should make it free when disable, and pretty fast when it is enabled. 

It is important that started tasks are ended, so several places in our code are moved to either try-finally blocks, or moved into a private method.  I realize this is ugly, but I think it is manageable.  In the future, we can look at making an agent or compiler plugin that simplifies the recording.   

Linking between threads is done with a `Link` object, which is created on the "outbound" task, and used on the "inbound" task.  This is slightly more verbose, and does has a small amount of runtime overhead, even when disabled.    (for null checks, slightly higher memory usage, etc.)   I think this is okay to, because it makes other optimizations much easier.

cc: @ran-su


core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -463,82 +466,100 @@ public void transportTerminated() {
...
-        Decompressor decompressor = decompressorRegistry.lookupDecompressor(encoding);
-        if (decompressor == null) {
-          stream.close(
-              Status.UNIMPLEMENTED.withDescription(
-                  String.format("Can't find decompressor for %s", encoding)),
-              new Metadata());
-          return;
+      final Tag tag = PerfMark.createTag(methodName, stream.hashCode());
+      PerfMark.startTask("ServerTransportListener.streamCreated", tag);
+      try {
```
ejona86: This is a lot to put into the try-finally. Could we pass tag as an argument to a function instead?

Would probably be a good idea for ClientStreamListener.closed as well. And for some of the Netty changes.

carl-mastrangelo: Done.

carl-mastrangelo:
Thanks @ejona86 !

PR #5827, opened 2019-06-04
examples: hedging example

dapengzhang0:
See README change for details.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -25,9 +25,72 @@ before trying out the examples.
 
 - [Json serialization](src/main/java/io/grpc/examples/advanced)
 
-- [Google Authentication](example-gauth/GOOGLE_AUTH_EXAMPLE.md)
+- <details>
+  <summary>Hedging</summary>
 
-### To build the examples
+  The [hedging example](src/main/java/io/grpc/examples/hedging) demonstrates that enabling hedging
+  can reduce tail latency. (Users should note that enabling hedging may introduce other overhead;
+  and in some scenarios, such as when some server resource gets exhausted for a period of time and
+  almost every RPC during that time has high latency or fails, hedging may make things worse.
+  Setting a throttle in the service config is recommended to protect the server from too many
+  inappropriate retry or hedging requests.)
+
+  The server and the client in the example are basically the same as those in the
+  [hello world](src/main/java/io/grpc/examples/helloworld) example, except that the server mimics a
+  long tail of latency, and the client sends 2000 requests and can turn on and off hedging.
+
+  To mimic the latency, the server randomly delays the RPC handling by 2 seconds at 10% chance, 5
+  seconds at 5% chance, and 10 seconds at 1% chance.
+
+  When running the client enabling the following hedging policy
+
+  ```json
+        "hedgingPolicy": {
+          "maxAttempts": 3,
+          "hedgingDelay": "1s",
+          "nonFatalStatusCodes": [
+            "UNAVAILABLE"
+          ]
+        }
+  ```
+  A summary of latency in the client log is typically like the following
+
+  ```text
+  Total RPCs sent: 2,000. Total RPCs failed: 0
+  [Hedging enabled]
+  ========================
+  50% latency: 0ms
+  90% latency: 6ms
+  95% latency: 1,003ms
+  99% latency: 2,002ms
+  99.9% latency: 2,011ms
+  Max latency: 5,272ms
+  ========================
+  ```
+
+  See [the section below](#to-build-the-examples) for how to build and run the example. The
+  executables for the server and the client are `hedging-hello-world-server` and
+  `hedging-hello-world-client`.
+
+  To disable hedging, set environment variable `DISABLE_HEDGING=true` before
+  running the client. A summary of latency in the client log is typically like the following
+
+  ```text
+  Total RPCs sent: 2,000. Total RPCs failed: 0
+  [Hedging disabled]
+  ========================
+  50% latency: 0ms
+  90% latency: 2,002ms
+  95% latency: 5,002ms
+  99% latency: 10,004ms
+  99.9% latency: 10,007ms
+  Max latency: 10,007ms
+  ========================
+  ```
+
+</details>
+
+### <a name="to-build-the-examples"></a> To build the examples
 
 1. **[Install gRPC Java library SNAPSHOT locally, including code generation plugin](../COMPILING.md) (Only need this step for non-released versions, e.g. master HEAD).**
 
@@ -93,6 +156,8 @@ $ bazel-bin/hello-world-client
   + [TLS examples](example-tls)
 
   + [ALTS examples](example-alts)
+
+- [Google Authentication](example-gauth/GOOGLE_AUTH_EXAMPLE.md)
 
 - [Kotlin examples](example-kotlin)
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -53,7 +53,7 @@ before trying out the examples.
           ]
         }
   ```
-  a summary of latency in the client log is typically like the following
+  Then the latency summary in the client log is like the following
 
   ```text
   Total RPCs sent: 2,000. Total RPCs failed: 0
@@ -73,7 +73,7 @@ before trying out the examples.
   `hedging-hello-world-client`.
 
   To disable hedging, set environment variable `DISABLE_HEDGING_IN_HEDGING_EXAMPLE=true` before
-  running the client. A summary of latency in the client log is typically like the following
+  running the client. That produces a latency summary in the client log like the following
 
   ```text
   Total RPCs sent: 2,000. Total RPCs failed: 0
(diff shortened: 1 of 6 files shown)


examples/src/main/java/io/grpc/examples/hedging/HedgingHelloWorldClient.java
```diff
@@ -0,0 +1,158 @@
+package io.grpc.examples.hedging;
```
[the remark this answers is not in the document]
dapengzhang0: Done.

examples/src/main/java/io/grpc/examples/hedging/HedgingHelloWorldClient.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * A client that requests a greeting from the {@link HedgingHelloWorldServer} with a hedging policy.
+ */
+public class HedgingHelloWorldClient {
+  static final String DISABLE_HEDGING = "DISABLE_HEDGING";
```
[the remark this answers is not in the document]
dapengzhang0: Done.

examples/src/main/java/io/grpc/examples/hedging/HedgingHelloWorldClient.java
```diff
@@ -0,0 +1,158 @@
...
+    HelloRequest request = HelloRequest.newBuilder().setName(name).build();
+    HelloReply response = null;
+    StatusRuntimeException statusRuntimeException = null;
+    long startTime = System.nanoTime();
+    try {
+      response = blockingStub.sayHello(request);
+    } catch (StatusRuntimeException e) {
+      failedRpcs.incrementAndGet();
+      statusRuntimeException = e;
+      logger.log(Level.WARNING, "RPC failed: {0}", e.getStatus());
```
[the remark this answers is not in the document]
dapengzhang0: Done.

examples/README.md
```diff
@@ -53,7 +53,7 @@ before trying out the examples.
           ]
         }
   ```
-  A summary of latency in the client log is typically like the following
+  a summary of latency in the client log is typically like the following
```
ejona86: Revert?

dapengzhang0: It's not the beginning of the sentence, or the previous sentence isn't complete.

ejona86: It is the beginning of a sentence. It is basically just missing a `:` or `.` after "following." "The following" is a noun here.

ejona86: Oh, I see. How about: "Then the latency summary in the client log is like the following"

examples/README.md
```diff
@@ -25,9 +25,72 @@ before trying out the examples.
...
+  Max latency: 5,272ms
+  ========================
+  ```
+
+  See [the section below](#to-build-the-examples) for how to build and run the example. The
+  executables for the server and the client are `hedging-hello-world-server` and
+  `hedging-hello-world-client`.
+
+  To disable hedging, set environment variable `DISABLE_HEDGING_IN_HEDGING_EXAMPLE=true` before
+  running the client. A summary of latency in the client log is typically like the following
```
ejona86: Wordsmithing. Might consider "That produces a latency summary in the client log like the following"

examples/README.md
```diff
@@ -25,9 +25,72 @@ before trying out the examples.
...
+  seconds at 5% chance, and 10 seconds at 1% chance.
+
+  When running the client enabling the following hedging policy
+
+  ```json
+        "hedgingPolicy": {
+          "maxAttempts": 3,
+          "hedgingDelay": "1s",
+          "nonFatalStatusCodes": [
+            "UNAVAILABLE"
```
dapengzhang0: @markdroth The [design](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#validation-of-hedgingpolicy) requires that the nonFatalStatusCodes for Hedging must be non-empty. But I found it really really awkward to HAVE TO add a status here. To me all statuses can be fatal, but I do still want hedging. This is unlike the normal retry, which you must have a (retryable) status to retry. I don't remember the rationale that for Hedging the set of nonFatalStatusCodes must be non-empty, but I really think this requirement dose not apply in my usecase. Would you mind I make a change to the design doc to delete this requirement?

markdroth: I think the rationale for requiring the statuses to be specified for hedging is exactly the same as for retries: the overall goal is to get the call to succeed, and we need to know which statuses we should keep trying vs. which ones to consider the call complete and return the error.  I don't think it's actually useful for this to be optional.

dapengzhang0: In my usecase, only to cut long tail latency (to get the call to succeed early), every status is fatal: I have multiple hedges, if I receive any status, including UNAVAILABLE, before receiving initial headers, I will cancel all the hedges, and commit the RPC as UNAVAILABLE; and if I receive initial headers from any one of the hedges first, I will cancel all the other hedges.

dapengzhang0: In case I know we should not keep trying for any status. Could I still do hedging to reduce tail latency? I think we still could.
>we need to know which statuses we should keep trying

markdroth: I don't think it's a realistic use-case to say that you care about reducing tail latency but don't care about RPC success.  If you are going to all the effort of supporting hedging, then you will also want to wait for a call that either succeeds or fails in a way that you know will not succeed for other attempts that are already in flight.  Otherwise, you're just wasting all of the resources that you used for the other attempts that are already in flight.

dapengzhang0: >I don't think it's a realistic use-case to say that you care about reducing tail latency but don't care about RPC success. 

I do care about RPC success.  But why at least one status for hedging? In this example, I have to make up a non-fatal status that is useless and risky for me, such as UNAVAILABLE (sometimes users may know most of but don't know ALL the scenarios for UNAVAILABLE, so they can't say UNAVAILABLE is non-fatal for sure), otherwise I can't do hedging at all?

>If you are going to all the effort of supporting hedging, then you will also want to wait for a call that either succeeds or fails in a way that you know will not succeed for other attempts that are already in flight

I think that's exactly what I do to allow empty nonFatalStatusCodes.

> Otherwise, you're just wasting all of the resources that you used for the other attempts that are already in flight.

If that is "waste", it wastes resources but saves time, because of backoff and long tail, the waste doesn't happen in 90% or 99%. For non-empty nonFatalStatusCodes it also wastes resources, just more or less. If you don't want to "waste" at all, why not just do normal retry instead of hedging. If you want to save time, you more or less waste some resources.

markdroth: Okay, I think I see what you're saying.  The problem here is that the hedging `nonFatalStatusCodes` field is actually the *inverse* of the retry `retriableStatusCodes` -- having the former be empty is the equivalent of having the latter list every single status code except for `OK`.

I'm open to making this change if everyone agrees it makes sense.  I'd like input from @ejona86 and @dfawley on this.

Also, if we are going to do this, then we should immediately fix the service config error handling code to support this, since we just cut the release branch yesterday.  If this release goes out with code that considers an empty list to be invalid, then it will not be possible to change this later without breaking people.  @yashykt, I assume that's an easy change?

dapengzhang0: Thanks, I will sent out a PR for the spec change for @markdroth @ejona86  @dfawley to review. There are currently no people using hedging, so the chance to break people (only new hedging users in the future && using the yesterday's branch cut && with empty nonfatalStatusCodes) is quite low, but still it's best to change before release goes out.

ejona86: I support having the list be empty. Hedging should allow issuing multiple RPCs and first one back wins, whatever the status.

> If this release goes out with code that considers an empty list to be invalid, then it will not be possible to change this later without breaking people

@markdroth:
1. DNS TXT is still disable though, right? So it shouldn't impact most users
2. I don't see that as a breaking change; it is a relaxation. It would mean that service owners couldn't use the empty list unless they were certain their clients were upgraded to a release where it worked. Although I agree it is good to avoid problems like that

dfawley: SGTM.  I don't see any reason to require this to be non-empty.

markdroth: I had thought that we had already enabled TXT lookups by default, but it looks like I was remembering wrong.  So yeah, I agree, this isn't that time-critical, but it is something we need to make sure to follow up on before we do that.

yashykt: C-Core does not parse hedging policy as of yet, probably because it has not implemented hedging support.

markdroth: Good point -- I had forgotten that we had not yet added parsing for this!

Okay, so there's nothing more to be done here yet.  Thanks!

ejona86: @markdroth, we must not enable TXT lookups until the service config error handling design+implementation is settled and tested.

markdroth: Agreed.  But that's been done in C-core.

PR #5833, opened 2019-06-05
all: update to error prone 2.3.3

carl-mastrangelo:


context/src/test/java/io/grpc/ContextTest.java
```diff
@@ -967,7 +967,7 @@ public void close() throws SecurityException {
         assertNull(logRef.get());
         ctx = ctx.fork();
       }
-      ctx = ctx.fork();
+      ctx.fork();
```
dapengzhang0: Why?

carl-mastrangelo: ctx is unused which ErrorProne complains about

dapengzhang0: I see, we had `@SuppressWarnings("CheckReturnValue") ` for `ContextTest`, so there is no problem.

PR #5834, opened 2019-06-05
core,services: cleanup io.grpc.internal.IoUtils

dapengzhang0:


core/src/main/java/io/grpc/internal/MessageFramer.java
```diff
@@ -269,7 +270,8 @@ private static int writeToOutputStream(InputStream message, OutputStream outputS
     } else {
       // This makes an unnecessary copy of the bytes when bytebuf supports array(). However, we
       // expect performance-critical code to support flushTo().
-      long written = IoUtils.copy(message, outputStream);
+      @SuppressWarnings("BetaApi") // ByteStreams is not Beta in v27
```
carl-mastrangelo: I don't think you can use it then.

dapengzhang0: There is no change for the API from v26 to v27, so the the API is technically stable since v26. We can safely use it without risk of guava incompatible changes in future releases.

PR #5835, opened 2019-06-05
examples: add example of how to set log scoped data with log4j

carl-mastrangelo:
This came up on [Stack Overflow](https://stackoverflow.com/questions/56396647/what-is-the-proper-way-to-set-mdc-for-logging-in-grpc-java-with-serverintercepto/).  The example shows how to use non-`io.grpc.Context`   scoped values.  In this case, it adds data that will be included in the log statement.

cc @ejona86

### DIFF
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -29,6 +29,7 @@ dependencies {
     implementation "io.grpc:grpc-protobuf:${grpcVersion}"
     implementation "io.grpc:grpc-stub:${grpcVersion}"
     compileOnly "javax.annotation:javax.annotation-api:1.2"
+    compile "org.apache.logging.log4j:log4j-api:2.11.2"
 
     // examples/advanced need this for JsonFormat
     implementation "com.google.protobuf:protobuf-java-util:${protobufVersion}"
--- a/examples/src/main/java/io/grpc/examples/logcontext/Client.java
+++ b/examples/src/main/java/io/grpc/examples/logcontext/Client.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2015 The gRPC Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.examples.logcontext;
+
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.ForwardingClientCall.SimpleForwardingClientCall;
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import io.grpc.StatusRuntimeException;
+import io.grpc.examples.helloworld.GreeterGrpc;
+import io.grpc.examples.helloworld.HelloReply;
+import io.grpc.examples.helloworld.HelloRequest;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * A simple client that like {@link io.grpc.examples.helloworld.HelloWorldClient}.
+ * This client can help you create custom headers.
+ */
+public class Client {
+  private static final Logger logger = Logger.getLogger(Client.class.getName());
+
+  private final ManagedChannel originChannel;
+  private final GreeterGrpc.GreeterBlockingStub blockingStub;
+
+  /**
+   * A custom client.
+   */
+  private Client(String host, int port) {
+    originChannel = ManagedChannelBuilder.forAddress(host, port)
+        .usePlaintext()
+        .build();
+
+    blockingStub = GreeterGrpc.newBlockingStub(originChannel).withInterceptors(
+        new ClientInterceptor() {
+          @Override
+          public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+              MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+            return new SimpleForwardingClientCall<ReqT, RespT>(next.newCall(method, callOptions)) {
+              @Override
+              public void start(Listener<RespT> responseListener, Metadata headers) {
+                try {
+                  headers.put(
+                      HeaderServerInterceptor.CLIENT_NAME_KEY,
+                      InetAddress.getLocalHost().getHostName());
+                } catch (UnknownHostException e) {
+                  throw new RuntimeException(e);
+                }
+                super.start(responseListener, headers);
+              }
+            };
+          }
+        });
+  }
+
+  private void shutdown() throws InterruptedException {
+    originChannel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
+  }
+
+  /**
+   * A simple client method that like {@link io.grpc.examples.helloworld.HelloWorldClient}.
+   */
+  private void greet(String name) {
+    logger.info("Will try to greet " + name + " ...");
+    HelloRequest request = HelloRequest.newBuilder().setName(name).build();
+    HelloReply response;
+    try {
+      response = blockingStub.sayHello(request);
+    } catch (StatusRuntimeException e) {
+      logger.log(Level.WARNING, "RPC failed: {0}", e.getStatus());
+      return;
+    }
+    logger.info("Greeting: " + response.getMessage());
+  }
+
+  /**
+   * Main start the client from the command line.
+   */
+  public static void main(String[] args) throws Exception {
+    Client client = new Client("localhost", 50051);
+    try {
+      /* Access a service running on the local machine on port 50051 */
+      String user = "world";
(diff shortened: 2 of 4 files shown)


examples/build.gradle
```diff
@@ -29,6 +29,7 @@ dependencies {
     implementation "io.grpc:grpc-protobuf:${grpcVersion}"
     implementation "io.grpc:grpc-stub:${grpcVersion}"
     compileOnly "javax.annotation:javax.annotation-api:1.2"
+    compile "org.apache.logging.log4j:log4j-api:2.11.2"
```
ejona86: We've been putting examples that require additional dependencies in their own project.

carl-mastrangelo: examples-log4j okay?

ejona86: SGTM

dapengzhang0: We started using `implementation` as replacement for `compile` in examples.

carl-mastrangelo: Done.

carl-mastrangelo: What's the difference?

dapengzhang0: I've heard `compile` is deprecated.

examples/examples-log4j/src/main/java/io/grpc/examples/logcontext/Client.java
```diff
@@ -0,0 +1,114 @@
...
+import io.grpc.examples.helloworld.HelloReply;
+import io.grpc.examples.helloworld.HelloRequest;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * A simple client that like {@link io.grpc.examples.helloworld.HelloWorldClient}.
```
sanjaypujare: s/that like/that is like/

examples/examples-log4j/src/main/java/io/grpc/examples/logcontext/Client.java
```diff
@@ -0,0 +1,114 @@
...
+          }
+        });
+  }
+
+  private void shutdown() throws InterruptedException {
+    originChannel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
+  }
+
+  /**
+   * A simple client method that like {@link io.grpc.examples.helloworld.HelloWorldClient}.
```
sanjaypujare: s/that like/that is like/

sanjaypujare:
I would like to see a readme file for this example with a link to the readme file from the main README.md under .../examples

examples/examples-log4j/src/main/proto/helloworld/helloworld.proto
```diff
@@ -0,0 +1,37 @@
+// Copyright 2015 The gRPC Authors
```
sanjaypujare: This file is identical to .../examples/src/main/proto/helloworld.proto. I understand the examples directory and package structure makes this necessary but can we think of eliminating this duplication? Also that one is under .../proto but this one is under .../proto/helloworld . Wonder why the difference.

examples/examples-log4j/src/main/java/io/grpc/examples/logcontext/CustomLogServer.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private static class GreeterImpl extends GreeterGrpc.GreeterImplBase {
+
+    @Override
+    public void sayHello(HelloRequest req, StreamObserver<HelloReply> responseObserver) {
+      logger.info("Got a request");
+      // outputs something like:
+      // 2019/06/05 15:22:12:686 PDT INFO Got a request
+      // {requestId=3e6c256d-6e87-411e-8bf3-fbf81e7ce0e6, clientname=my.domain.name}
```
sanjaypujare: s/clientname/clientName/

examples/examples-log4j/src/main/java/io/grpc/examples/logcontext/CustomLogServer.java
```diff
@@ -0,0 +1,107 @@
...
+      HelloReply reply = HelloReply.newBuilder().setMessage("Hello " + req.getName()).build();
```
sanjaypujare: Is there any way we can copy the requestId and clientName values into the reply text so on the client side user can see what the serverInterceptor added in the ThreadContext?

examples/examples-log4j/src/main/java/io/grpc/examples/logcontext/HeaderServerInterceptor.java
```diff
@@ -0,0 +1,108 @@
...
+      final Metadata requestHeaders,
+      ServerCallHandler<ReqT, RespT> next) {
+
+    final String requestId = UUID.randomUUID().toString();
+    final String clientName = String.valueOf(requestHeaders.get(CLIENT_NAME_KEY));
+
+    return new SimpleForwardingServerCallListener<ReqT>(next.startCall(call, requestHeaders)) {
+      @Override
+      public void onCancel() {
+        ThreadContext.put(REQUEST_ID_NAME, requestId);
```
Nipel-Crumple: Maybe it's appropriate to use here `org.apache.logging.log4j.CloseableThreadContext` with _try-with-resources_?

examples/examples-log4j/src/main/java/io/grpc/examples/logcontext/HeaderServerInterceptor.java
```diff
@@ -0,0 +1,108 @@
...
+import io.grpc.ServerCall;
+import io.grpc.ServerCallHandler;
+import io.grpc.ServerInterceptor;
+import java.util.UUID;
+import org.apache.logging.log4j.ThreadContext;
+
+/**
+ * A interceptor to handle server header.
+ */
+public class HeaderServerInterceptor implements ServerInterceptor {
```
Nipel-Crumple: In case of interceptors-chain the context won't be set in next interceptors because it is set only in callbacks, will it? For example if the next interceptor has some non-trivial logic or calculation based on request meta-information, its logs won't be printed in log files with the given `requestId` because `ThreadContext` is not set for interceptors chain.

examples/examples-log4j/src/main/java/io/grpc/examples/logcontext/CustomLogServer.java
sanjaypujare: Also the only one that matters here is the one you set in `onComplete` (or is it `onMessage`) inside the `Listener` in `HeaderServerInterceptor.interceptCall`, right? For the ones set in `onCancel`, `onHalfClose` etc, is there any way to capture and verify them?

examples/build.gradle
carl-mastrangelo: Change made

creamsoup:
don't forget to add the new build.gradle to VERSION_FILES in RELEASING.md

examples/examples-log4j2/BUILD.bazel
```diff
@@ -0,0 +1,53 @@
...
+        "@io_grpc_grpc_java//api",
+        "@io_grpc_grpc_java//protobuf",
+        "@io_grpc_grpc_java//stub",
+    ],
+)
+
+java_binary(
+    name = "hello-world-alts-client",
+    testonly = 1,
+    main_class = "io.grpc.examples.alts.HelloWorldAltsClient",
```
sanjaypujare: Why does the build file have these classes? Shouldn't this be related to CustomLogServer and Client? Looks like file copied from another example?

examples/examples-log4j2/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+2. Run in this directory:
+```
+$ ../gradlew installDist
+```
+
+This creates the scripts `hello-world-server`, `hello-world-client`,
+`route-guide-server`, and `route-guide-client` in the
+`build/install/examples/bin/` directory that run the examples. Each
+example requires the server to be running before starting the client.
```
sanjaypujare: This seems wrong. This only creates hello-world-logcontext-server and hello-world-logcontext-client and in build/install/examples-log4j2/bin directory?

examples/examples-log4j2/README.md
```diff
@@ -0,0 +1,62 @@
+grpc Log4j2 example
+==============================================
+
+The examples require grpc-java to already be built. You are strongly encouraged
```
sanjaypujare: I want to see some verbiage about what this example is supposed to illustrate - possibly at the top of the file. How about adding information like (feel free to edit or add):

"This example illustrates how to properly set the ThreadContext in a server-interceptor that is then used by the logger in your server. The server-interceptor in this example sets a requestId (a randomly generated string) and the client name from the request header in the ThreadContext. The log statement in the server method will include these 2 things in the log output".

examples/examples-log4j2/gradlew.bat
```diff
@@ -0,0 +1,84 @@
+@if "%DEBUG%" == "" @echo off
```
sanjaypujare: Is this file used? Most example sub-directories don't contain gradlew.bat and gradlew (Linux shell script). Seems to be redundant?

examples/examples-log4j2/gradlew
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env sh
```
sanjaypujare: Is this file used? Most example sub-directories don't contain gradlew (Linux shell script). Redundant?

examples/examples-log4j2/build.gradle
```diff
@@ -0,0 +1,78 @@
...
+def protocVersion = '3.7.1'
+
+dependencies {
+    implementation "org.apache.logging.log4j:log4j-api:2.11.2"
+    runtime "org.apache.logging.log4j:log4j-core:2.11.2"
+    implementation "io.grpc:grpc-protobuf:${grpcVersion}"
+    implementation "io.grpc:grpc-stub:${grpcVersion}"
+    implementation "io.grpc:grpc-netty:${grpcVersion}"
+    compileOnly "javax.annotation:javax.annotation-api:1.2"
+}
```
larry-safran, line 35: All versions need to be updated to current values.

PR #5837, opened 2019-06-06
java_grpc_library: Swap to descriptor_set_in to protoc

ejona86:
This avoids re-parsing the proto files and allows proto_library
to enforce more checks.

This is an export of cl/197343148

------

I discovered the incompatibilities with this internally were few and were due to extraneous dependencies. They have been fixed, so this is now safe.

CC @mjduijn


mjduijn:
Great to see the issues resolved! I'll update my PR accordingly

PR #5842, opened 2019-06-06
android: Convert to maven-publish

ejona86:
com.github.dcendents:android-maven-gradle-plugin is incompatible with
Gradle 5 and the project hasn't seen any activity in over a year, so it
seems unlikely to get fixed.

We want to use maven-publish anyway, since that's what we use elsewhere.

--------

CC @creamsoup

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -14,12 +14,13 @@ buildscript {
     dependencies {
         classpath 'com.android.tools.build:gradle:3.3.0'
         classpath "net.ltgt.gradle:gradle-errorprone-plugin:0.6"
-        classpath 'com.github.dcendents:android-maven-gradle-plugin:2.1'
+        classpath "digital.wup:android-maven-publish:3.6.2"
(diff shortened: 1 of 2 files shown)


android/build.gradle
```diff
@@ -81,69 +77,83 @@ task sourcesJar(type: Jar) {
...
-            developer {
-                id "grpc.io"
-                name "gRPC Contributors"
-                email "grpc-io@googlegroups.com"
-                url "https://grpc.io/"
-                organization = "gRPC Authors"
-                organizationUrl "https://www.google.com"
+    repositories {
+        maven {
+            if (rootProject.hasProperty('repositoryDir')) {
```
dapengzhang0: where is `repositoryDir` set? `android.sh`?

ejona86: It is within linux_artifacts.sh. It's more obvious now because I needed to change the line that actually includes it.

PR #5849, opened 2019-06-07
core,netty: use PerfMark tags with the HTTP/2 stream ids

carl-mastrangelo:
This change removes the WriteQueue linking and splits it out into each
of the commands, so that the trace is more precise, and the tag
information is correct.

It is still unclear what the initial Tag should be for ClientCallImpl,
since it should not access the TransportState to get the HTTP/2 stream id.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
@@ -74,6 +74,8 @@ import io.netty.handler.codec.http2.Http2StreamVisitor;
 import io.netty.handler.codec.http2.StreamBufferingEncoder;
 import io.netty.handler.codec.http2.WeightedFairQueueByteDistributor;
 import io.netty.handler.logging.LogLevel;
+import io.perfmark.PerfMark;
+import io.perfmark.Tag;
 import java.nio.channels.ClosedChannelException;
 import java.util.concurrent.Executor;
 import java.util.logging.Level;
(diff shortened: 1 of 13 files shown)


carl-mastrangelo:
Still to do is tracing of OkHTTP and Cronet, but after those there shouldn't be any more of these PRs.

carl-mastrangelo:
Blocked on roll forward of https://github.com/grpc/grpc-java/pull/5853

netty/src/main/java/io/grpc/netty/NettyClientHandler.java
```diff
@@ -539,54 +544,71 @@ private void createStream(final CreateStreamCommand command, final ChannelPromis
       return;
     }
 
-    final NettyClientStream.TransportState stream = command.stream();
-    final Http2Headers headers = command.headers();
+    NettyClientStream.TransportState stream = command.stream();
+    Http2Headers headers = command.headers();
     stream.setId(streamId);
 
+    PerfMark.startTask("NettyClientHandler.createStream", stream.tag());
```
creamsoup: not related to this PR, but returning task (or closable for startTask method) is maybe cleaner than passing string argument.

netty/src/test/java/io/grpc/netty/NettyClientHandlerTest.java
```diff
@@ -789,7 +789,7 @@ public TransportStateImpl(
         EventLoop eventLoop,
         int maxMessageSize,
         TransportTracer transportTracer) {
-      super(handler, eventLoop, maxMessageSize, StatsTraceContext.NOOP, transportTracer);
+      super(handler, eventLoop, maxMessageSize, StatsTraceContext.NOOP, transportTracer, "name");
```
ejona86: s/name/methodname/ ? or method like below?

Ditto for NettyClientStreamTest.

netty/src/main/java/io/grpc/netty/NettyClientHandler.java
carl-mastrangelo: Ack

netty/src/test/java/io/grpc/netty/NettyClientHandlerTest.java
carl-mastrangelo: Done!

PR #5832, opened 2019-06-05
interop-testing: disable timeout when debugging

carl-mastrangelo:

PR #5851, opened 2019-06-07
interop-testing: fix tests on Android

carl-mastrangelo:
Updates #5832

cc: @ejona86


dapengzhang0:
Did you try to use `android.support.test.rule.DisableOnAndroidDebug` instead of `DisableOnDebug` directly? I wonder if that works.

carl-mastrangelo:
@dapengzhang0 can we even pull in that dependency?

dapengzhang0:
That would be as painful as the deps of `android.annotation.SuppressLint` that I added [before](https://github.com/grpc/grpc-java/commit/b7dd92e7bb3fbc476a186177aee39230440238bd).

>can we even pull in that dependency?

I checked the source code of `DisableOnAndroidDebug`, seems it will not work for non-android.

carl-mastrangelo:
@dapengzhang0 I don't really debug on Android.  I think it's okay for the timeout to fallback to how it works now.

dapengzhang0:
LGTM

PR #5875, opened 2019-06-13
Revert "services: fix HealthCheckingLoadBalancer.shutdown(). (#5848)"

ejona86:
This reverts commit c6f15162ff3d5e21c3cb037079ddf8292fb68140 (#5848). It broke
an internal health checking test because the server wouldn't shut down.
We assume the health checking RPC isn't getting closed.

### DIFF
--- a/services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
+++ b/services/src/main/java/io/grpc/services/HealthCheckingLoadBalancerFactory.java
@@ -121,7 +121,7 @@ final class HealthCheckingLoadBalancerFactory extends Factory {
       HealthCheckState hcState = new HealthCheckState(
           this, originalSubchannel, syncContext, delegate.getScheduledExecutorService());
       hcStates.add(hcState);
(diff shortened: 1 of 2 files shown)


zhangkun83:
The reason it doesn't work is because `HealthCheckingLoadBalancer.shutdown()` calls `super.shutdown()` first, which triggers the real balancer to shutdown all Subchannels, which will remove them from `hcStates` before `HealthCheckingLoadBalancer.shutdown()` iterates them and deliver the fake SHUTDOWN, resulting in HC RPCs not being cancelled.

Currently we have the restriction that `SubchannelStateListener` aren't called after LoadBalancer is shutdown. I think this restriction is unnecessary and complicates our logic here. Without that restriction, we could completely rely on `onSubchannelState(SHUTDOWN)` for cancelling the HC RPCs, and no need to fake the notifications.

PRs are on the way.

dapengzhang0:
@zhangkun83 Could you fix this soon, because out internal lb interop tests with java massage client is flaky.

PR #5879, opened 2019-06-14
core/test: re-enable tests for panic mode.

zhangkun83:
panic mode was temporarily disabled by #4152 and re-enabled by #4245,
but the tests were not.  This has caused a few test code that was
broken but executed at all.

Also verify that panic mode is not triggered in the tests unexpectedly.


core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -2214,7 +2227,7 @@ public void run() {
     verifyPanicMode(panicReason);
 
     // No new resolver or balancer are created
```
ejona86: The code disagrees with this comment.

zhangkun83: Clarified the comment.

PR #5880, opened 2019-06-14
core: Fix FINE deadline logging

ejona86:
We were logging when withDeadline() was used, not when the Context was used. As
discovered while looking at https://stackoverflow.com/q/56593692/4690866 .

In e19e8f7d updateTimeoutHeaders was removed and logIfContextNarrowedTimeout
was called directly. However, the two methods had reverse ordering of
callDeadine/outerCallDeadline and the caller did not get their arguments
swapped.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ClientCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ClientCallImpl.java
@@ -248,7 +248,7 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
     boolean deadlineExceeded = effectiveDeadline != null && effectiveDeadline.isExpired();
     if (!deadlineExceeded) {
       logIfContextNarrowedTimeout(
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -248,7 +248,7 @@ public void runInContext() {
     boolean deadlineExceeded = effectiveDeadline != null && effectiveDeadline.isExpired();
     if (!deadlineExceeded) {
       logIfContextNarrowedTimeout(
-          effectiveDeadline, callOptions.getDeadline(), context.getDeadline());
```
dapengzhang0: This is a bug not caught by tests! But I'm not so bothered without test coverage because it's not convenient to test logging.

ejona86: I started to add a test, but then trashed it because I really don't want to start having tests for debug log statements.

PR #5886, opened 2019-06-14
Revert "interop-testing: Observe flow control in TestServiceImpl"

ejona86:
This reverts commit e795f14bedc56af2f2ba60eef1b23add6106ab99. It breaks some of
our internal tests. At the very least the cast fails because we are reusing the
service for something without going through the normal stub. There may also
have been a test hang, but it'd take more effort to figure out.


ejona86:
Kokoro is hosed currently. We'd like this in to help with the import, so using my administrator powers to ignore Windows CI.

Issue #4776, opened 2018-08-17
ClientCallImpl.toString() could be more helpful


carl-mastrangelo:
ClientCall current only prints out the method descriptor, which is not so helpful.  It would be useful to include some more detail.  Here are some ideas:

* The deadline
* whether the call has been started, half closed, or cancelled
* whether initial headers have been received, or trailers recieved
* what was the context for the call
* the channel id of the call
* if the channel has a subchannel assigned, the subchannel id or remote host name
* Security level (if available)
* Wait for ready or not

These would be useful when printing out the call, or including the call details in error messages, or generally.

carl-mastrangelo:
cc @zhangkun83

carl-mastrangelo:
Getting some thoughts down after talking with a few people:

* initial deadline and time remaining  are useful.  
* representing the string a a textproto would be terse and human readable.  It could be parsed automatically
* include the channelid or socketid (per channelz) in the string, which can be grepped for in the INFO level logs  This would alleviate putting a lot of debug info into the string, as it could be queried via channelz
* RPC state would be useful, possible as short string (headerssent|messagesent|closesent|headersrecv|messagesrecv|closerecv) 
* Include the call string in the StatusRuntimeException produced by io.grpc.stub would be nice, possibly in the exception's message (not the status's)

bgguna:
I can pick up this issue and create a PR. This is going to be my first PR to this community so if anything is incorrect, please shout out.

ejona86:
Closing, since nobody's found a need for this recently.

I don't think most of the details mentioned are a great idea, as they expose internals I'd rather not expose and require thread-safety to implement. The exceptions would be deadline and wait-for-ready; those would be fine, but given the improvements to DEADLINE_EXCEEDED error reporting I'm not sure it adds much value at this point.

Issue #4740, opened 2018-08-07
Attach channel state information to locally generated DEADLINE_EXCEEDED errors


zhangkun83:
Additional information can be added to description to help debug DEADLINE_EXCEEDED errors.

Things to consider:
 - Current connectivity state
 - Latest events from channel trace
 - Last error from LoadBalancer (saved even after LoadBalancer got out of TRANSIENT_FAILURE state)

ejona86:
This would have helped debugging a mobile user today.

PR #5892, opened 2019-06-17
core: attach debug information about stream to DEADLINE_EXCEEDED

zhangkun83:
Works for #4740 

- Subclasses of `AbstractClientStream` include remote address in insight if available.
- `DelayedStream` adds buffered time, and the insight of real stream if it's set.
- `RetriableStream` insights outputs of Substreams.
- ~~`ManagedChannelImpl` adds channel state. Other information, like last error status, would need API changes to `LoadBalancer`~~.

Example error message:
```
deadline exceeded after 8112071ns. [buffered_nanos=24763, remote_addr=foo.test.google.fr/127.0.0.1:44749]
```
or
```
deadline exceeded after 8112071ns. [buffered_nanos=22344324763, waiting_for_connection]
```

This is related to #4776 but taking a more usage-specific approach.

### DIFF
--- a/core/src/main/java/io/grpc/inprocess/InProcessTransport.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessTransport.java
@@ -21,6 +21,7 @@ import static io.grpc.internal.GrpcUtil.TIMEOUT_KEY;
 import static java.lang.Math.max;
 
 import com.google.common.base.MoreObjects;
+import com.google.common.base.MoreObjects.ToStringHelper;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
 import io.grpc.Attributes;
@@ -808,8 +809,7 @@ final class InProcessTransport implements ServerTransport, ConnectionClientTrans
       }
 
       @Override
-      public String getDebugString() {
-        return getClass().getSimpleName();
+      public void appendTimeoutDetails(ToStringHelper toStringHelper) {
       }
     }
   }
--- a/core/src/main/java/io/grpc/inprocess/InProcessTransport.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessTransport.java
@@ -806,6 +806,11 @@ final class InProcessTransport implements ServerTransport, ConnectionClientTrans
         long effectiveTimeout = max(0, deadline.timeRemaining(TimeUnit.NANOSECONDS));
         headers.put(TIMEOUT_KEY, effectiveTimeout);
       }
+
+      @Override
+      public String getDebugString() {
+        return getClass().getSimpleName();
+      }
     }
   }
 
--- a/core/src/main/java/io/grpc/inprocess/InProcessTransport.java
+++ b/core/src/main/java/io/grpc/inprocess/InProcessTransport.java
@@ -21,7 +21,6 @@ import static io.grpc.internal.GrpcUtil.TIMEOUT_KEY;
 import static java.lang.Math.max;
 
 import com.google.common.base.MoreObjects;
-import com.google.common.base.MoreObjects.ToStringHelper;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
 import io.grpc.Attributes;
@@ -45,6 +44,7 @@ import io.grpc.internal.ConnectionClientTransport;
 import io.grpc.internal.GrpcAttributes;
 import io.grpc.internal.GrpcUtil;
 import io.grpc.internal.InUseStateAggregator;
+import io.grpc.internal.InsightBuilder;
 import io.grpc.internal.ManagedClientTransport;
 import io.grpc.internal.NoopClientStream;
 import io.grpc.internal.ObjectPool;
@@ -809,7 +809,7 @@ final class InProcessTransport implements ServerTransport, ConnectionClientTrans
       }
 
       @Override
-      public void appendTimeoutDetails(ToStringHelper toStringHelper) {
+      public void appendTimeoutInsight(InsightBuilder insight) {
       }
     }
   }
--- a/core/src/main/java/io/grpc/internal/ClientCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ClientCallImpl.java
@@ -30,7 +30,6 @@ import static java.lang.Math.max;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
-import com.google.common.base.MoreObjects.ToStringHelper;
 import io.grpc.Attributes;
 import io.grpc.CallOptions;
 import io.grpc.ClientCall;
@@ -349,12 +348,12 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
 
     @Override
     public void run() {
-      ToStringHelper timeoutDetails = MoreObjects.toStringHelper("");
-      stream.appendTimeoutDetails(timeoutDetails);
+      InsightBuilder insight = new InsightBuilder();
+      stream.appendTimeoutInsight(insight);
       // DelayedStream.cancel() is safe to call from a thread that is different from where the
       // stream is created.
       stream.cancel(DEADLINE_EXCEEDED.augmentDescription(
-          String.format("deadline exceeded after %dns. %s", remainingNanos, timeoutDetails)));
+          String.format("deadline exceeded after %dns. %s", remainingNanos, insight)));
     }
   }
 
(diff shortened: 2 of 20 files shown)


core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -347,10 +349,12 @@ private void removeContextListenerAndCancelDeadlineFuture() {
 
     @Override
     public void run() {
+      ToStringHelper timeoutDetails = MoreObjects.toStringHelper("");
```
carl-mastrangelo: a comment saying why this is blank would be helpful.  It's unusual, and not immediately obvious.

core/src/main/java/io/grpc/inprocess/InProcessTransport.java
```diff
@@ -806,6 +807,10 @@ public void setDeadline(Deadline deadline) {
         long effectiveTimeout = max(0, deadline.timeRemaining(TimeUnit.NANOSECONDS));
         headers.put(TIMEOUT_KEY, effectiveTimeout);
       }
+
+      @Override
+      public void appendTimeoutDetails(ToStringHelper toStringHelper) {
```
carl-mastrangelo: TBH, I think StringBuilder is a better fit.  

```java
public void appendTimeoutDetails(StringBuilder buffer) {
  buffer.append(key).append('=').append(value).append(',')
}
```

The code that finally does toString can trim the last character.  Alternatively, this could just be a `Map<String, Object>`.

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -480,7 +480,6 @@ public void run() {
     return channelStatsFuture;
   }
 
-  @VisibleForTesting
```
ejona86: Revert?

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -89,6 +92,10 @@
    */
   private final AtomicBoolean noMoreTransparentRetry = new AtomicBoolean();
 
+  /** Keeps track of all substreams for informative purposes. */
+  private final ConcurrentLinkedQueue<Substream> allSubstreams =
```
ejona86: I'm concerned about how long we'll keep these streams alive. We could potentially be retaining an entire transport for "no reason."

I think we should add a new `Collection<Status>` or `Collection<Status.Code>` to `State` for the completed streams. The stream's Status can be easily passed to `State.substreamClosed(Substream)`, and so when the stream is removed from `drainedSubstreams` its results will be present in the other list. There's no need for the other stream data for completed streams (like address) since it isn't normally provided.

zhangkun83: I have updated this PR according to your suggestion. Now it includes the status codes from all closed substreams. @dapengzhang0, according to javadoc on the `drainedSubstreams` field, it doesn't seem to always include unfinished substreams. Please advice on how to proceed.

core/src/main/java/io/grpc/internal/InsightBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+ * resulted string is made up of a list of short strings, each of which gives out a piece of
+ * information.
+ */
+public final class InsightBuilder {
+  private final ArrayList<String> buffer = new ArrayList<String>();
+
+  /**
+   * Appends a piece of information which is a plain string.
+   */
+  public InsightBuilder append(String insight) {
```
carl-mastrangelo: If you made this Object, you could call String.valueof() inside and avoid having to call toString() at the call sites.

core/src/main/java/io/grpc/internal/InsightBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.ArrayList;
+import javax.annotation.Nullable;
+
+/**
+ * Builds a concise and readable string that gives insight of the concerned part of the system.  The
+ * resulted string is made up of a list of short strings, each of which gives out a piece of
+ * information.
+ */
+public final class InsightBuilder {
+  private final ArrayList<String> buffer = new ArrayList<String>();
```
carl-mastrangelo: List

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -642,6 +644,20 @@ public final Attributes getAttributes() {
     return Attributes.EMPTY;
   }
 
+  @Override
+  public void appendTimeoutInsight(InsightBuilder insight) {
+    synchronized (lock) {
+      insight.appendKeyValue("finished", closedSubstreamsInsight);
+    }
+    InsightBuilder drainedSubstreamsInsight = new InsightBuilder();
+    for (Substream sub : state.drainedSubstreams) {
```
dapengzhang0: I don't think `drainedSubstreams` provide useful information, a previously failed attempt could be either drained or not-yet-drained, and a drained message could be either be buffered in DelayedStream or in RealStream, there is not much difference from the user's point of view, it only marks a progress of internal implementation.

core/src/main/java/io/grpc/internal/RetriableStream.java
zhangkun83: Per offline discussion, I have updated the code and added comments.

core/src/main/java/io/grpc/internal/InsightBuilder.java
zhangkun83: I don't think it matters here... it would add one more import line.

zhangkun83:
PTAL

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -646,16 +646,32 @@ public final Attributes getAttributes() {
...
+    State currentState;
     synchronized (lock) {
-      insight.appendKeyValue("finished", closedSubstreamsInsight);
+      insight.appendKeyValue("closed", closedSubstreamsInsight);
+      currentState = state;
     }
-    InsightBuilder drainedSubstreamsInsight = new InsightBuilder();
-    for (Substream sub : state.drainedSubstreams) {
+    if (currentState.winningSubstream != null) {
+      // TODO(zhangkun83): in this case while other drained substreams have been cancelled in favor
```
carl-mastrangelo: This is not a todo.  There is nothing in this comment explaining what is to be done.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -646,16 +646,32 @@ public final Attributes getAttributes() {
...
+      // of the winning substream, they may not have received closed() notifications yet, thus they
+      // may be missing from closedSubstreamsInsight.  This may be a little confusing to the user.
       InsightBuilder substreamInsight = new InsightBuilder();
-      sub.stream.appendTimeoutInsight(substreamInsight);
-      drainedSubstreamsInsight.append(substreamInsight.toString());
+      currentState.winningSubstream.stream.appendTimeoutInsight(substreamInsight);
+      insight.appendKeyValue("committed", substreamInsight);
+    } else {
+      InsightBuilder openSubstreamsInsight = new InsightBuilder();
+      // drainedSubstreams doesn't include all open substreams.  Those who have just been created
```
carl-mastrangelo: s/who/which/

core/src/test/java/io/grpc/internal/RetriableStreamTest.java
```diff
@@ -217,6 +217,15 @@ public void retry_everythingDrained() {
     ClientStream mockStream3 = mock(ClientStream.class);
 
     doReturn(mockStream1).when(retriableStreamRecorder).newSubstream(0);
+    doAnswer(new Answer<Void>() {
+        @Override
+        public Void answer(InvocationOnMock in) {
+          InsightBuilder insight = (InsightBuilder) in.getArguments()[0];
+          insight.appendKeyValue("server_addr", "2.2.2.2:443");
```
carl-mastrangelo: existing terminology in our code I think points to this being a "backend_addr", not a server_addr.   (or possibly remote_addr)

core/src/main/java/io/grpc/internal/DelayedStream.java
```diff
@@ -94,6 +98,22 @@ public void run() {
...
+  public void appendTimeoutInsight(InsightBuilder insight) {
+    synchronized (this) {
+      if (listener == null) {
+        return;
+      }
+      if (realStream != null) {
+        insight.appendKeyValue("buffered_nanos", streamSetTimeNanos - startTimeNanos);
+        realStream.appendTimeoutInsight(insight);
+      } else {
+        insight.append("waiting_for_connection");
```
ejona86: Nit: it'd be nice if these cases were more symmetric in the output string, moving this to the next line.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -642,6 +644,36 @@ public final Attributes getAttributes() {
     return Attributes.EMPTY;
   }
 
+  @Override
+  public void appendTimeoutInsight(InsightBuilder insight) {
+    State currentState;
+    synchronized (lock) {
+      insight.appendKeyValue("closed", closedSubstreamsInsight);
```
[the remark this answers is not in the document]
zhangkun83: `InsightBuilder` already does that and that's documented in the javadoc.

core/src/main/java/io/grpc/internal/RetriableStream.java
```diff
@@ -642,6 +644,36 @@ public final Attributes getAttributes() {
...
+      // drainedSubstreams doesn't include all open substreams.  Those who have just been created
+      // and are still catching up with buffered requests (in other words, still draining) will not
+      // show up.  We think this is benign, because the draining should be typically fast, and it'd
+      // be indistinguishable from the case where those streams are to be created a little late due
+      // to delays in the timer.
+      for (Substream sub : currentState.drainedSubstreams) {
+        InsightBuilder substreamInsight = new InsightBuilder();
+        sub.stream.appendTimeoutInsight(substreamInsight);
+        openSubstreamsInsight.append(substreamInsight);
+        insight.appendKeyValue("open", openSubstreamsInsight);
```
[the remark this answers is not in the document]
zhangkun83: Good catch! Fixed.

core/src/main/java/io/grpc/internal/DelayedStream.java
zhangkun83: Done.

PR #5897, opened 2019-06-18
alts: temporarily revert protocol negotiation to work with servers

carl-mastrangelo:


carl-mastrangelo:
They do, but I believe it's safe, as long as its just TSI that breaks the pattern.

PR #5900, opened 2019-06-19
alts: fix server hang

carl-mastrangelo:
Followup to #5893 which causes a server side hang.  This is a hack.


carl-mastrangelo:
@jiangtaoli2016 it would be super helpful if there was a mock nanny handshaker so we could test the client and server together.   It's pretty painful to test this internally, but then make the changes externally.

jiangtaoli2016:
@carl-mastrangelo We have fake handshaker service in gRPC c++ (grpc/test/core/tsi/alts/fake_handshaker/) used by envoy. Can grpc-java use this by running fake handshaker as a separate binary? Or we have to write a fake handshaker in Java.

carl-mastrangelo:
@jiangtaoli2016 can it be used in a bazel rule?  we could make one here.

PR #5901, opened 2019-06-19
netty-shaded: Use compile instead of runtime dependency in pom

### LINKED ISSUE #5881: Add dependency on grpc-core to the main README.md
The "Download" section starts with a Maven pom file excerpt containing some dependencies. Note that grpc-netty-shaded doesn't include some classes that are available in grpc-core, for example, `io.grpc.internal.AbstractServerImplBuilder`. I suggest to add dependency on grpc-core to avoid confusion. 
(I've spent quite some time trying to figure out why example-tls builds with Maven but my project doesn't... Turned out the difference is that my project uses netty-shaded.)

ejona86:
Maven does not include transitive runtime dependencies in the
compile-time classpath (testing shows Gradle 4 does). So if a user
references the shaded NettyServerBuilder without also depending on
grpc-core directly, compilation will fail because
AbstractServerImplBuilder couldn't be found.

This isn't technically a problem, since we're not wanting to encourage
users to reference the shaded classes directly. But some users will
certainly reference the classes anyway and the error is pretty confusing
while also being trivially worked around. In other words: it justs
wastes people's time and benefits nobody.

Fixes #5881

### DIFF
--- a/netty/shaded/build.gradle
+++ b/netty/shaded/build.gradle
@@ -56,7 +56,7 @@ publishing {
                     dependencyNode.appendNode('artifactId', dep.name)
                     def version = (dep.name == 'grpc-core') ? '[' + dep.version + ']' : dep.version
                     dependencyNode.appendNode('version', version)
-                    dependencyNode.appendNode('scope', 'runtime')
+                    dependencyNode.appendNode('scope', 'compile')
                 }
             }
         }


ejona86:
For Gradle's behavior, see: https://stackoverflow.com/questions/35066318/gradle-includes-transitive-runtime-dependency-as-compile-dependency

Although it appears it may have been fixed in 5.0: https://github.com/gradle/gradle/issues/977

I tested on grpc 1.21.0, which used gradle 4.10.1. But our example didn't use `enableFeaturePreview('IMPROVED_POM_SUPPORT') `, thus I wouldn't have seen the new behavior. I'm glad this has been fixed in Gradle (although I haven't verified by testing).

PR #5905, opened 2019-06-19
core,netty: block server shutdown until the socket is unbound

carl-mastrangelo:


api/src/main/java/io/grpc/Server.java
```diff
@@ -102,6 +104,8 @@ public int getPort() {
 
   /**
    * Initiates an orderly shutdown in which preexisting calls continue but new calls are rejected.
+   * After calling this method, clients may no longer connect to the listening socket(s).
```
ejona86: Nit: I feel like this first sentence is mostly covered by the first. And the user couldn't really notice if we accepted another connection anyway. How about "After this call returns, this server has released the listening socket(s) and it may be reused by another server."

netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -283,6 +283,15 @@ public void operationComplete(ChannelFuture future) throws Exception {
         eventLoopReferenceCounter.release();
       }
     });
+    if (Thread.currentThread().isInterrupted()) {
```
ejona86: Why this condition? Is it an optimization? Honestly, I'd prefer not to optimize this case, which also has benefits like the FINE logging happening independent of how interrupt races.

carl-mastrangelo: IIRC normal futures throw if they try to block while the thread is interrupted, but I wasn't sure Netty's would do the same.  I'm happy to remove it

api/src/main/java/io/grpc/Server.java
carl-mastrangelo: Done.

PR #5910, opened 2019-06-20
Revert "api: expose if waitForReady has been set on CallOptions"

ejona86:
This reverts commit 53a3f12666f00d941660cefe55df27b87174f415. It is not
needed by service config. If it was needed then we'd have a problem
because there's virtually no user calling withoutWaitForReady even
though virtually every user wanting fail fast.

------

This reverts #5634. CC @zhangkun83


carl-mastrangelo:
Until you get the spec changed, this method is needed. 

 @markdroth: you examples on https://github.com/grpc/grpc-java/pull/5634  don't cover the case where waitForReady is explicitly set to `false` on the call, but `true` in the service config.

markdroth:
If it is explicitly set via the API, we use that value, regardless of what the service config says.

carl-mastrangelo:
@markdroth then you would have to agree: the ClientCall has a tri-state waitForReady value: unset, false, and true.  Unset implies false, unless the service config says otherwise.  Hence, I don't think this PR can proceed.

ejona86:
> If it is explicitly set via the API, we use that value, regardless of what the service config says.

None of our users set fail fast explicitly via the API. If that is mandatory, then the feature needs to be removed from service config; we can't support it.

ejona86:
The point is that our clients only explicitly set "waitForReady". None of them explicitly set "fail fast", because there is no point. There are _zero_ users of `withoutWaitForReady()` in Google.

markdroth:
Just because no one does this today doesn't mean they can't.  In a deployment where the service owner is in control of their clients and wants most calls to be wait_for_ready but has some specific calls for which they don't want wait_for_ready, it makes perfect sense to set wait_for_ready=true in the service config and use the client API to explicitly set wait_for_ready=false on calls for which they don't want wait_for_ready.

Supporting this field does not mean that people have to use it, and I do agree that people have to understand the implications before choosing to use it.  But the defined behavior has not changed, and we should support it in Java, just like we do in other languages.

ejona86:
> just like we do in other languages.

@markdroth, Go does not use a ternary value for wait for ready. So they are either not implementing it at all or implementing it like I was suggesting.
https://github.com/grpc/grpc-go/blob/712624e6869ea374ad1d2a57071e7de7310dbfbd/rpc_util.go#L157

And C# is the same (stored as a bit in a bit field):
https://github.com/grpc/grpc/blob/559e328799c524b8b39988f07f08ba62a56b7b69/src/csharp/Grpc.Core.Api/CallOptions.cs#L121

Node, Ruby, and ObjC don't seem to support wait for ready. Python added wait for ready in 1.21 and it uses ternary. It's unclear if PHP supports wait for ready.

> In a deployment where the service owner is in control of their clients and wants most calls to be wait_for_ready but has some specific calls for which they don't want wait_for_ready, it makes perfect sense to set wait_for_ready=true in the service config and use the client API to explicitly set wait_for_ready=false on calls for which they don't want wait_for_ready

Uh, huh. And why do they need to push changes to that behavior in service config? In Java they would just call withWaitForReady when they create the stub and they'd get the same behavior.

It sounds like Java can accept their feature request when it comes and they will control their client's rollout.

ejona86:
@larry-safran, that's because nobody (probably literally zero) users are using the feature this was done for. And if we do find a user using it, I'd tell them to stop because it is fundamentally wrong.

Issue #5812, opened 2019-05-31
Add retry logic for XdsComms


dapengzhang0:
Please answer these questions before submitting your issue.

PR #5920, opened 2019-06-21
xds: ADS RPC retry

dapengzhang0:


dapengzhang0:
Resolves #5812

xds/src/main/java/io/grpc/xds/XdsComms.java
```diff
@@ -280,6 +330,7 @@ public void onCompleted() {
           .streamAggregatedResources(xdsResponseReader);
       this.localityStore = localityStore;
 
+      cancelRetryTimer();
```
zhangkun83: A bit weird for an inner class to mutate outer class' state in the constructor.
Might be better to move it to `refreshAdsStream()`, and let the retry timer call `refreshAdsStream()` too.

xds/src/main/java/io/grpc/xds/XdsComms.java
```diff
@@ -269,6 +286,39 @@ public void onCompleted() {
...
+          // run in SynchronizationContext
+          void scheduleRetry() {
+            if (channel.isShutdown()) {
+              return;
+            }
+
+            checkState(
+                cancelled || closed,
+                "Scheduling retry while the stream is neither cancelled nor closed");
+
```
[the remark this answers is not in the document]
dapengzhang0: Done.

xds/src/main/java/io/grpc/xds/XdsComms.java
dapengzhang0: This was exactly my first draft version. Because besides `refreshAdsStream()`, the `XdsComms` constructor also calls `new AdsStream()`, I wanted it to be always safe to call `new AdsStream()` so I put `cancelRetryTimer()` in `AdsStream()`. Reverted.

PR #5930, opened 2019-06-25
util: a util to gracefully switch load balancer when lb policy changes

dapengzhang0:
Following the section "Gracefully Switching LB Policies" in the spec go/grpc-client-channel-spec

### DIFF
--- a/core/src/main/java/io/grpc/util/GracefulLbSwitch.java
+++ b/core/src/main/java/io/grpc/util/GracefulLbSwitch.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2019 The gRPC Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package io.grpc.util;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.grpc.ConnectivityState.READY;
+
+import io.grpc.ConnectivityState;
+import io.grpc.Internal;
+import io.grpc.LoadBalancer;
+import io.grpc.LoadBalancerProvider;
+import io.grpc.Status;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.NotThreadSafe;
+
+/**
+ * A forwarding load balancer and holder of currentLb and pendingLb. The pendingLb's helper will not
+ * update balancing state until a subchannel managed by the pendingLB is READY, whence the pendingLb
+ * becomes to current.
+ */
+@Internal
+@NotThreadSafe // Must be accessed in SynchronizationContext
+public final class GracefulLbSwitch extends ForwardingLoadBalancer {
+
+  // never null
+  private final LoadBalancer newLb;
+  // never null
+  private LoadBalancer currentLb;
+  @Nullable // set to null once the new lb becomes current
+  private LoadBalancer pendingLb;
+
+  /**
+   * Constructor. Example usage:
+   *
+   * <pre>
+   *   private GracefulLbSwitch lbSwitch;
+   *
+   *   void onLbPolicyChange() {
+   *     lbSwitch = new GracefulLbSwitch(lbSwitch, newLbProvider, helper);
+   *     lbSwitch.handleResolvedAddresses(resolvedAddresses);
+   *   }
+   * </pre>
+   */
+  public GracefulLbSwitch(
+      @Nullable GracefulLbSwitch previousSwitch, LoadBalancerProvider lbProvider,
+      final Helper helper) {
+    checkNotNull(lbProvider, "lbProvider");
+    checkNotNull(helper, "helper");
+
+    if (previousSwitch == null) {
+      newLb = lbProvider.newLoadBalancer(helper);
+      currentLb = newLb;
+      return;
+    }
+
+    if (previousSwitch.pendingLb != null) {
+      previousSwitch.pendingLb.shutdown();
+    }
+
+    currentLb = previousSwitch.currentLb;
+
+    class PendingHelper extends ForwardingLoadBalancerHelper {
+
+      @Override
+      protected Helper delegate() {
+        return helper;
+      }
+
+      @Override
+      public void updateBalancingState(
+          ConnectivityState newState, SubchannelPicker newPicker) {
+        if (newState == READY) {
+          if (pendingLb != null) {
+            currentLb.shutdown();
+            currentLb = pendingLb;
+            pendingLb = null;
+          }
+        }
+
+        if (pendingLb == null) {
+          helper.updateBalancingState(newState, newPicker);
+        }
+      }
+    }
+
+    newLb = lbProvider.newLoadBalancer(new PendingHelper());
+    pendingLb = newLb;
+  }
+
+  @Override
+  protected LoadBalancer delegate() {
+    return newLb;
+  }
+
+  @Override
+  public void shutdown() {
+    currentLb.shutdown();
+    if (pendingLb != null) {
+      pendingLb.shutdown();
+    }
+  }
+}
--- a/core/src/main/java/io/grpc/util/GracefulLbSwitch.java
+++ b/core/src/main/java/io/grpc/util/GracefulLbSwitch.java
@@ -24,7 +24,6 @@ import io.grpc.ConnectivityState;
 import io.grpc.Internal;
 import io.grpc.LoadBalancer;
 import io.grpc.LoadBalancerProvider;
-import io.grpc.Status;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 
--- a/core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
+++ b/core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
@@ -17,8 +17,10 @@
 package io.grpc.util;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 import static io.grpc.ConnectivityState.READY;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.grpc.ConnectivityState;
 import io.grpc.ConnectivityStateInfo;
 import io.grpc.Internal;
@@ -44,6 +46,19 @@ public final class GracefulSwitchLoadBalancer extends ForwardingLoadBalancer {
     public void shutdown() {}
   };
 
+  @VisibleForTesting
+  static final SubchannelPicker BUFFER_PICKER = new SubchannelPicker() {
+    @Override
+    public PickResult pickSubchannel(PickSubchannelArgs args) {
+      return PickResult.withNoResult();
+    }
+
+    @Override
+    public String toString() {
+      return "BUFFER_PICKER";
+    }
+  };
+
   private final Helper helper;
   private LoadBalancer currentLb = NOOP_BALANCER;
   private LoadBalancer pendingLb = NOOP_BALANCER;
@@ -52,6 +67,8 @@ public final class GracefulSwitchLoadBalancer extends ForwardingLoadBalancer {
   @Nullable
   private String pendingPolicyName;
   private boolean isReady;
+  private ConnectivityState pendingState;
+  private SubchannelPicker pendingPicker;
 
   public GracefulSwitchLoadBalancer(Helper helper) {
     this.helper = checkNotNull(helper, "helper");
@@ -68,6 +85,9 @@ public final class GracefulSwitchLoadBalancer extends ForwardingLoadBalancer {
     pendingLb.shutdown();
     pendingLb = NOOP_BALANCER;
     pendingPolicyName = null;
+    pendingState = ConnectivityState.IDLE;
+    pendingPicker = BUFFER_PICKER;
+
     if (newPolicyName.equals(currentPolicyName)) {
       return;
     }
@@ -82,14 +102,19 @@ public final class GracefulSwitchLoadBalancer extends ForwardingLoadBalancer {
 
       @Override
       public void updateBalancingState(ConnectivityState newState, SubchannelPicker newPicker) {
-        if (pendingLb == lb && newState == READY) {
-          swap();
-        }
-        if (currentLb == lb) {
-          helper.updateBalancingState(newState, newPicker);
+        if (lb == pendingLb) {
+          checkState(isReady, "there is pending lb while current lb has been out of ready");
+          pendingState = newState;
+          pendingPicker = newPicker;
+          if (newState == READY) {
+            swap();
+          }
+        } else if (lb == currentLb) {
           isReady = newState == READY;
           if (!isReady && pendingLb != NOOP_BALANCER) { // current policy exits READY, swap
             swap();
+          } else {
+            helper.updateBalancingState(newState, newPicker);
           }
         }
       }
@@ -110,6 +135,7 @@ public final class GracefulSwitchLoadBalancer extends ForwardingLoadBalancer {
     currentPolicyName = pendingPolicyName;
     pendingLb = NOOP_BALANCER;
     pendingPolicyName = null;
+    helper.updateBalancingState(pendingState, pendingPicker);
   }
 
   @Override
--- a/core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
+++ b/core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
@@ -23,7 +23,7 @@ import io.grpc.ConnectivityState;
 import io.grpc.Internal;
 import io.grpc.LoadBalancer;
 import io.grpc.LoadBalancerProvider;
-import javax.annotation.Nullable;
+import io.grpc.Status;
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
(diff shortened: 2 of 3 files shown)


dapengzhang0:
@zhangkun83, @ejona86  This is how the lb switcher looks like.

core/src/main/java/io/grpc/util/GracefulLbSwitch.java
```diff
@@ -0,0 +1,116 @@
...
+  private LoadBalancer pendingLb;
+
+  /**
+   * Constructor. Example usage:
+   *
+   * <pre>
+   *   private GracefulLbSwitch lbSwitch;
+   *
+   *   void onLbPolicyChange() {
+   *     lbSwitch = new GracefulLbSwitch(lbSwitch, newLbProvider, helper);
```
zhangkun83: Both the user and this class are doing swap-outs. It might be easier to use if you let the user keep the reference and only this class do the swap-out:
```java
class GracefulSwitchLoadBalancer extends ForwardingLoadBalancer {
  // Assuming changing policy doesn't swap out the child helper
  // If it does, the new helper would be passed here.
  void switchTo(LoadBalancer newLoadBalancer);
  ...
}
```

dapengzhang0: `void switchTo(LoadBalancer newLoadBalancer)` returns void, then there is only one instance of `GracefulSwitchLoadBalancer`? When to shutdown which?

ejona86: +1 to reusing the same instance to doing future switches.

@dapengzhang0, yes, each code location that wants graceful switching would have one instance of the class. I don't understand the shutdown question.

dapengzhang0: @ejona86 `GracefulSwitchLoadBalancer` in @zhangkun83 's code snippet is a `LoadBalancer`, when to shutdown `GracefulSwitchLoadBalancer` instance?

ejona86: When shutdown is called.

dapengzhang0: Then user must not reuse or call `switchTo` after it's shutdown.

ejona86: That seems fine. I'd expect that to largely happen when the LB is being shutdown already.

dapengzhang0: `instance.shutdown()` -> `instance.switchTo(newDelegate)` -> `instance.handleResolvedAddresses()` on the same instance breaks the contract in javadoc of `LoadBalancer#shutdown()`

```
  /**
   * ...  No more methods on this class will be called after this method.  ...
   */
  public abstract void shutdown();
```

ejona86: When you shut down, you throw away the instance. But you're only going to need to shut down the instance when you yourself are shutting down. So there won't be any additional method calls after the shutdown.

dapengzhang0: Done.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
```diff
@@ -0,0 +1,110 @@
...
+  private LoadBalancer delegate;
+  @Nullable // never null after init()
+  private LoadBalancer currentLb;
+  @Nullable // set to null once the new lb becomes current
+  private LoadBalancer pendingLb;
+
+  private GracefulSwitchLoadBalancer() {}
+
+  /** Initializes an instance. */
+  public static GracefulSwitchLoadBalancer init(LoadBalancerProvider lbProvider, Helper helper) {
```
zhangkun83: Alternatively, you can assign `delegate` and `currentLb` to no-op LoadBalancers initially, and let the user to call `switchTo()` even for the first LB. That will be a smaller API surface and won't be harder to use.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
```diff
@@ -0,0 +1,110 @@
...
+    delegate = lbProvider.newLoadBalancer(pendingHelper);
+    pendingHelper.lb = delegate;
+    pendingLb = delegate;
+  }
+
+  @Override
+  protected LoadBalancer delegate() {
+    return delegate;
+  }
+
```
zhangkun83: All other methods should be propagated to both `currentLb` and `pendingLb` (maybe there is no benefit of extending `ForwardingLoadBalancer`).

Handling of name resolution results / errors is a bit interesting. First, `canHandleEmptyAddressListFromNameResolution()` should probably return the value from the latest LB (either `currentLb` or `pendingLb`). If we do that, we should also propagate name resolution results and errors to the same LB. That means a LB being swapped out won't receive resolution updates, but that seems fine.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
```diff
@@ -0,0 +1,110 @@
...
+import io.grpc.LoadBalancerProvider;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.NotThreadSafe;
+
+/**
+ * A forwarding load balancer and holder of currentLb and pendingLb. The pendingLb's helper will not
+ * update balancing state until a subchannel managed by the pendingLB is READY, whence the pendingLb
+ * becomes current.
+ */
+@Internal
```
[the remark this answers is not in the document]
dapengzhang0: This is because `ForwardingLoadBalancer` and `ForwardingLoadBalancerHelper` are in `io.grpc.util`. We had the same problem for `CachedSubchannelPool` https://github.com/grpc/grpc-java/pull/5663#discussion_r280648764

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
dapengzhang0: I don't agree with that. The new name resolution result should assume the latest lb policy, and the old policy does not apply.

zhangkun83: `handleSubchannelState()` and `requestConnection()` should routed to both.
I agree that name resolution results should be sent to the latest (looks you are already doing that).

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
dapengzhang0: Good idea!

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
dapengzhang0: Yes, that makes sense.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
```diff
@@ -0,0 +1,114 @@
...
+  private LoadBalancer currentLb = NOOP_BALANCER;
+  private LoadBalancer pendingLb = NOOP_BALANCER;
+
+  /** Gracefully switch to a new balancer. */
+  public void switchTo(LoadBalancerProvider lbProvider, final Helper helper) {
+    checkNotNull(lbProvider, "lbProvider");
+    checkNotNull(helper, "helper");
+
+    if (currentLb == NOOP_BALANCER) {
+      delegate = lbProvider.newLoadBalancer(helper);
```
zhangkun83: Maybe always wrap the helper? When this `delegate` is swapped out in the future, we will want to discard any updates from it.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
```diff
@@ -0,0 +1,114 @@
...
+public final class GracefulSwitchLoadBalancer extends ForwardingLoadBalancer {
+  private static final LoadBalancer NOOP_BALANCER = new LoadBalancer() {
+    @Override
+    public void handleNameResolutionError(Status error) {}
+
+    @Override
+    public void shutdown() {}
+  };
+
+  private LoadBalancer delegate = NOOP_BALANCER;
```
zhangkun83: Both `currentLb` and `pendingLb` are delegates, thus naming this field `delegate` is confusing. Since it's the one that receives name resolution results, I would suggest renaming it to `resolutionResultRecipient`.

You may also consider not extending `ForwardingLoadBalancer`, but instead override all resolution-related methods that explicitly forward to `resolutionResultRecipient`. Right now the name resolution path is hidden behind `ForwardingLoadBalancer`'s not overridden methods and that's not very readable.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
dapengzhang0: Now implementation follows as go/grpc-client-channel-spec, `requestConnection()` will be only forwarded to the new policy. 

`handleSubchannelState()` is deprecated and will not be used by this class, so not overriding it now.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
dapengzhang0: Now changed implementation, following  go/grpc-client-channel-spec .

dapengzhang0:
PTAL

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
```diff
@@ -0,0 +1,124 @@
...
+
+      @Override
+      public void updateBalancingState(ConnectivityState newState, SubchannelPicker newPicker) {
+        if (pendingLb == lb && newState == READY) {
+          swap();
+        }
+        if (currentLb == lb) {
+          helper.updateBalancingState(newState, newPicker);
+          isReady = newState == READY;
+          if (!isReady && pendingLb != NOOP_BALANCER) { // current policy exits READY, swap
```
zhangkun83: When the old policy is swapped out due to non-readiness, the last reported state from the new policy should take over. This will make a difference on fail-fast RPCs if old policy is TRANSIENT_FAILURE while the new one is CONNECTING. How about:
```java
if (pendingLb == lb) {
  pendingState = newState;
  pendingPicker = newPicker;
  if (newState == READY) {
    swap();
  }
}
if (currentLb == lb) {
  isReady = newState == READY;
  if (!isReady && pendingLb != NOOP_BALANCER) {
    swap();
    helper.updateBalancingState(pendingState, pendingPicker);
  } else {
    helper.updateBalancingState(newState, newPicker);
  }
}
```

dapengzhang0: What is the initial pendingPicker?

A. null, on swap helper will not update picker for new policy.
B. BufferPicker, on swap helper will update BufferPicker for new policy.

zhangkun83: The initial pendingPicker is relevant only when `currentLb` is swapped out before `pendingLb` emits an update. In that case initial state should be IDLE and the picker should be BufferPicker, which is consistent with the initial states of ManagedChannel before LoadBalancer emits the first update.

dapengzhang0: Done, with a little difference.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java

dapengzhang0:
PTALA

core/src/test/java/io/grpc/util/GracefulSwitchLoadBalancerTest.java
```diff
@@ -329,19 +336,51 @@ public void switchWhileOldPolicyIsReadyThenOldPolicyGoesToNotReady() {
...
+
+    LoadBalancer lb1 = balancers.get(lbPolicies[1]);
+    Helper helper1 = helpers.get(lb1);
 
     picker = mock(SubchannelPicker.class);
     helper0.updateBalancingState(CONNECTING, picker);
-    verify(lb0).shutdown();
+
+    inOrder.verify(lb0).shutdown();
+    inOrder.verify(mockHelper, never()).updateBalancingState(CONNECTING, picker);
```
zhangkun83: Does `inOrder.verify(, never())` mean this never happened after `lb0.shutdown()`? This could pass unexpectedly if the action did happen but was before `lb0.shutdown()`. I always verify without `inOrder` just to be sure.

core/src/test/java/io/grpc/util/GracefulSwitchLoadBalancerTest.java
```diff
@@ -0,0 +1,481 @@
...
+    LoadBalancer lb1 = balancers.get(lbPolicies[1]);
+
+    gracefulSwitchLb.switchTo(lbProviders.get(lbPolicies[1]));
+    assertThat(balancers.get(lbPolicies[1])).isSameInstanceAs(lb1);
+
+    verifyNoMoreInteractions(lb0, lb1);
+  }
+
+  @Test
+  public void newPolicyNameTheSameAsCurrentPolicy() {
```
zhangkun83: Test method names should include both the premises and the expected outcome. For example: "switchToSamePolicyLb_hasNoEffect()".

core/src/test/java/io/grpc/util/GracefulSwitchLoadBalancerTest.java
dapengzhang0: Good point. Didn't check carefully when adding inOrder. Removed inOrder for this line.

core/src/test/java/io/grpc/util/GracefulSwitchLoadBalancerTest.java
dapengzhang0: Done. However, did not apply to all methods, some have multiple verification points. Normally I use _suffix for dimensions, such as testMethod1_option1On_option2Off.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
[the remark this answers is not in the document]
ejona86: We should either move the Forwarding tools to `io.grpc` or to `io.grpc.internal`. We can have a shim in `io.grpc.util` that extends from its "real" location.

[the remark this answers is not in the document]
ejona86: In talking with Kun more:
 * This is a utility that hierarchical LBs will need. Long-term it should be a public utility (in io.grpc.util)
 * We will need to use it in `io.grpc.internal` for AutoConfiguredLoadBalancer

So that means the core implementation will need to be in `io.grpc.internal` and then we'll have a shim in `io.grpc.util`. The shim can be a ForwardingLoadBalancer where delegate() returns the instance from `io.grpc.internal` and defines switchTo().

So this current location is "fine", but shouldn't be `@Internal`.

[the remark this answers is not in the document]
dapengzhang0: I had following choices in mind (out of scope of this PR):
- A) Move Forwarding tools from util to `io.grpc`. This need deprecate the existing API.
- B) Move Forwarding tools from util to `io.grpc.internal`. This need deprecate the existing API, and  after deleted there will be no external public API anymore.
- C) Add an InternalAccessor shim in `io.grpc.util`. Then `io.grpc.util` will be more dependent on `io.grpc.internal`. (Currently `io.grpc.util` does depend on `io.grpc.internal`, but technically not necessary).
- D) Let `io.grpc.internal` depend on `io.grpc.util`, not currently the opposite. Will impact Bazel users.
- E) Move RoundRobinBalancer out of `io.grpc.util` to `io.grpc.roundrobin` (this will make `io.grpc.util` lightweight), and let `io.grpc.internal` depend on the lightweight `io.grpc.util`, not the opposite. This will have behavior change: 
  + people using round_robin will add `io.grpc.roundrobin` to runtime dependency; bazel users using round_robin will not add `io.grpc.util` to runtime dependency, which currently they do and does not make sense.
  + those who only want to use the public APIs in `io.grpc.util` will not have a side effect of loading round_robin.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Override
+    public String toString() {
+      return "BUFFER_PICKER";
+    }
+  };
+
+  private final Helper helper;
+  private LoadBalancer currentLb = NOOP_BALANCER;
```
ejona86: Have a comment somewhere here what "current" and "pending" mean and their relationship? Something "current" is the LB whose picker is being used and pending is the policy we are waiting to become ready. And that current guaranteed to be set after the initial switchTo and that the pending fields are cleared when it become current.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
```diff
@@ -0,0 +1,158 @@
...
+    checkNotNull(newLbProvider, "newLbProvider");
+
+    String newPolicyName = newLbProvider.getPolicyName();
+    if (newPolicyName.equals(pendingPolicyName)) {
+      return;
+    }
+    pendingLb.shutdown();
+    pendingLb = NOOP_BALANCER;
+    pendingPolicyName = null;
+    pendingState = ConnectivityState.IDLE;
```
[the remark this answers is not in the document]
zhangkun83: I thought it was created with IDLE. `ManagedChannelImpl.exitIdleMode()` doesn't change the state from IDLE to CONNECTING.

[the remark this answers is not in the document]
ejona86: Hmmm... you're right, it doesn't. That seems like a bug. The initial name resolution should be considered part of CONNECTING. And when we create a LB we do expect it to connect to addresses as soon as it gets the list, without any RPC being issued.

Maybe that's what this comment is referring to (although the `updateBalancingState()` seems unnecessary, because the SubchannelStateListener is going to fire immediately and we'll update the balancer state again):
https://github.com/grpc/grpc-java/blob/8a9afd618aec0e1bf5858232cac70c89a90fb96e/core/src/main/java/io/grpc/internal/PickFirstLoadBalancer.java#L62-L64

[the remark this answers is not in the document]
dapengzhang0: >although the updateBalancingState() seems unnecessary, because the SubchannelStateListener is going to fire immediately and we'll update the balancer state again

So because of above, can we set initial pendingState = null, and if old policy goes out of READY state while pendingState is null, we don't update balancing state immediately, it will be updated automatically on the new policy's inital handleResolvedAddress()?

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+  }
+
+  private void swap() {
+    currentLb.shutdown();
+    currentLb = pendingLb;
+    currentPolicyName = pendingPolicyName;
+    pendingLb = NOOP_BALANCER;
+    pendingPolicyName = null;
+    helper.updateBalancingState(pendingState, pendingPicker);
```
ejona86: Let's move the `shutdown()` after this line (alternatively move this before the shutdown).

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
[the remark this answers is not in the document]
ejona86: If we don't update the state, that implies we are still using the old policy. We should not continue using the picker for a shutdown() policy. So I do think it is important to make sure we replace the picker. I made a small comment to make that clearer in swap.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
```diff
@@ -0,0 +1,158 @@
...
+
+    @Override
+    public String toString() {
+      return "BUFFER_PICKER";
+    }
+  };
+
+  private final Helper helper;
+  private LoadBalancer currentLb = NOOP_BALANCER;
+  private LoadBalancer pendingLb = NOOP_BALANCER;
```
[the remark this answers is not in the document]

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
```diff
@@ -0,0 +1,158 @@
...
+  @Nullable
+  private String currentPolicyName;
+  @Nullable
+  private String pendingPolicyName;
+  private boolean isReady;
```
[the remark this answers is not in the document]

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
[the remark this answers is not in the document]

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java

PR #5933, opened 2019-06-26
interop-testing: support just ServerBuilder types

carl-mastrangelo:


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -233,14 +234,17 @@ private void startServer() {
...
         .addStreamTracerFactory(serverStreamTracerFactory);
-    io.grpc.internal.TestingAccessor.setStatsImplementation(
-        builder,
-        new CensusStatsModule(
-            tagger,
-            tagContextBinarySerializer,
-            serverStatsRecorder,
-            GrpcUtil.STOPWATCH_SUPPLIER,
-            true, true, true, false /* real-time metrics */));
+    if (metricsExpected()) {
```
zhangkun83: Even if metrics are not expected, census context propagation may still happen, e.g., in `InProcessTest`. You need to inject fake Census whenever builder is `AbstractServerImplBuilder`, regardless of `metricsExpected()`, and then have the assert:
```java
if (metricsExpected()) {
  assertThat(builder).isInstanceOf(AbstractServerImplBuilder.class).
}
```

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -1472,6 +1476,7 @@ public void customMetadata() throws Exception {
   @Test(timeout = 10000)
   public void censusContextsPropagated() {
     Assume.assumeTrue("Skip the test because server is not in the same process.", server != null);
+    Assume.assumeTrue(metricsExpected());
```
zhangkun83: Same reason as above. This should be `assumeTrue(serverBuilderIsAbstractServerImplBuilder)`.

interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
carl-mastrangelo: I changed it to just check the class.  PTAL

zhangkun83: Please add the assert I suggested. It's much easier to debug this assert failing than metrics not being recorded.

carl-mastrangelo: Done, (sorry, I didn't understand why you wanted it at first).

PR #5934, opened 2019-06-26
Protono

carl-mastrangelo:
Upstream ProtocolNegotiatiorHandler, and swap the appropriate classes to it.

ALTS is not switched yet, since it is shared between client and server.  Once the server is changed to use WBAEH, it can be moved too.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
+++ b/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
@@ -322,17 +322,14 @@ final class ProtocolNegotiators {
     public void close() {}
   }
 
-  static final class ClientTlsHandler extends ChannelDuplexHandler {
+  static final class ClientTlsHandler extends ProtocolNegotiationHandler {
 
-    private final ChannelHandler next;
     private final SslContext sslContext;
     private final String host;
     private final int port;
 
-    private ProtocolNegotiationEvent pne;
-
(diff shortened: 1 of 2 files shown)


netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -851,54 +843,113 @@ public AsciiString scheme() {
...
+          .build();
+      replaceProtocolNegotiationEvent(existingPne.withAttributes(attrs));
+    }
+  }
+
+  /**
+   * ProtocolNegotiationHandler is a convenience handler that makes it easy to follow the rules for
+   * protocol negotiation.  Handlers should strongly consider extending this handler.
+   */
+  static class ProtocolNegotiationHandler extends ChannelDuplexHandler {
```
creamsoup: should this be abstract class to force override protocolNegotiationEventTriggered? 
(handlerAdded0 and other ForOverride ones could remain same)

carl-mastrangelo: That's actually a good idea.  Let me change it.

carl-mastrangelo: Hmm, well after trying it out, it doesn't seem so easy.  `protocolNegotiationEventTriggered` doesn't actually include the event (by design) so classes like ClientTlsHandler have to override it, even though they may not use it.

PR #5936, opened 2019-06-27
netty: don't use InternalProtocolNegotiators in ProtocolNegotiators

carl-mastrangelo:
copy paste error from adding ProtcolNegotiatorHandler

### DIFF
--- a/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
+++ b/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
@@ -904,8 +904,7 @@ final class ProtocolNegotiators {
 
     @Override
     public final void handlerAdded(ChannelHandlerContext ctx) throws Exception {
-      InternalProtocolNegotiators.negotiationLogger(ctx)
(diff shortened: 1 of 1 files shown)


netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -904,8 +904,7 @@ protected ProtocolNegotiationHandler(ChannelHandler next) {
 
     @Override
     public final void handlerAdded(ChannelHandlerContext ctx) throws Exception {
-      InternalProtocolNegotiators.negotiationLogger(ctx)
-          .log(ChannelLogLevel.DEBUG, negotiatorName + " started");
+      negotiationLogger(ctx).log(ChannelLogLevel.DEBUG, negotiatorName + " started");
```
dapengzhang0: Avoid string concat. But no need do it in this PR.

carl-mastrangelo: I'll do a followup.

PR #5941, opened 2019-06-28
netty: bump to 4.1.37

carl-mastrangelo:


carl-mastrangelo:
Since netty_tcnative was updated to be compatible, I will update it separately.

PR #5946, opened 2019-07-01
core: make libraries.android_annotations compileOnly

dapengzhang0:
This will not add 
```
    <dependency>
      <groupId>com.google.android</groupId>
      <artifactId>annotations</artifactId>
      <version>4.1.1.4</version>
      <scope>compile</scope>
    </dependency>
```

into `/grpc-core-version.pom`.

Tested by commenting on `@SuppressLint` and off in `JndiResourceResolverFactory` then `gradlew :app:lintDebug` for `examples/android/helloworld`, still works.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -27,6 +27,7 @@ dependencies {
             project(':grpc-api').sourceSets.test.output,
             project(':grpc-testing'),
             project(':grpc-grpclb'),
+            libraries.android_annotations,
             libraries.guava_testlib
 
     signature "org.codehaus.mojo.signature:java17:1.0@signature"
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -3,7 +3,6 @@ description = 'gRPC: Core'
 dependencies {
     compile project(':grpc-api'),
             libraries.gson,
-            libraries.android_annotations,
             libraries.perfmark
     compile (libraries.opencensus_api) {
         // prefer 3.0.2 from libraries instead of 3.0.1
@@ -21,6 +20,8 @@ dependencies {
         // prefer 20.0 from libraries instead of 19.0
         exclude group: 'com.google.guava', module: 'guava'
     }
+    
+    compileOnly libraries.android_annotations
 
     testCompile project(':grpc-context').sourceSets.test.output,
             project(':grpc-api').sourceSets.test.output,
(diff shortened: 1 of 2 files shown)


dapengzhang0:
`compileTestJava` had a warning, so I had to add `testCompile libraries.android_annotations`; however, I didn't see warnings in helloworld example. Would users see warning?

ejona86:
What was the warning you saw for `compileTestJava`?

dapengzhang0:
> What was the warning you saw for compileTestJava?

https://travis-ci.org/grpc/grpc-java/jobs/552905381

```
> Task :grpc-core:compileTestJava
/home/travis/build/grpc/grpc-java/core/build/classes/java/main/io/grpc/internal/JndiResourceResolverFactory$JndiRecordFetcher.class: warning: Cannot find annotation method 'value()' in type 'SuppressLint': class file for android.annotation.SuppressLint not found
error: warnings found and -Werror specified
1 error
1 warning
> Task :grpc-core:compileTestJava FAILED
```

ejona86:
It seems like the dependency is needed for consumers during compilation, since the annotation has CLASS retention. I don't think there's a better scope than compile in Maven.

What is the objective here?

dapengzhang0:
>It seems like the dependency is needed for consumers during compilation, since the annotation has CLASS retention.

But I didn't see any warnings in example/helloword, seems we didn't suppress any warnings in jvmargs in examples/build.gradle

https://source.cloud.google.com/results/invocations/d64b4d0c-cbe5-4999-9de9-f867797cc0a0/targets/grpc%2Fjava%2Fmaster%2Fpresubmit%2Flinux_artifacts/log

>What is the objective here?
Just try to make grpc-core pom file cleaner.

ejona86:
> But I didn't see any warnings in example/helloword, seems we didn't suppress any warnings in jvmargs in examples/build.gradle

There's plenty of tools that will scan all classes. Those would be impacted. In effect, that is not a sufficient test and the warning message you provided shows that certain tools may notice the missing annotation.

dapengzhang0:
I tried with grpc-java-api-checker, there was no breakage. But I understand the risk that some compile time scanners may try to load all the annotation classes and then they will fail at this one. Abort.

PR #5951, opened 2019-07-03
Add `@Nullable` to getter for trailers on `StatusRuntimeException` si…

mkobit:
…nce the constructors allows it to be null


dapengzhang0:
Thank you @mkobit

PR #5947, opened 2019-07-02
Make //compiler:grpc_java_plugin publicly visible again

aaliddell:
Fixes #5942

Prior to 745aa0a2f570d74e7d7fe0c72334c24e21b8ab17, this target was visible publicly and is useful for external projects that depend on the protoc plugin.

PR #5952, opened 2019-07-03
bazel: Add comment for public protoc plugin

### LINKED ISSUE #5942: //compiler:grpc_java_plugin no longer visible
v1.21.0

Prior to 745aa0a2f570d74e7d7fe0c72334c24e21b8ab17, the `//compiler:grpc_java_plugin` target was publicly visible for use by external bazel workspaces. This allowed use of the protoc plugin in Bazel workspaces that use grpc-java as a dependency, in particular the rules_proto rules. Since this commit, it it no longer possible.

Is there a particular motivation for marking it as non-public?

ejona86:
The 3rd-party rules_proto project is referencing our compiler directly
and not using our java_grpc_library. This target is fine for them to
use (although we'd prefer using our java_grpc_library), but most users
shouldn't be touching it.
    
Related to #5942 and #5947

### DIFF
--- a/compiler/BUILD.bazel
+++ b/compiler/BUILD.bazel
@@ -1,5 +1,6 @@
 load("//:java_grpc_library.bzl", "java_rpc_toolchain")
 
+# This should not generally be referenced. Users should use java_grpc_library
 cc_binary(
     name = "grpc_java_plugin",
     srcs = [


dapengzhang0:
There was already a PR for this, #5947. The only difference is comment.

ejona86:
Hmm... Well, the comment is important. We could merge the other one and then this one could just become the comment. Or the other changes to include some comment of some sort.

dapengzhang0:
>We could merge the other one and then this one could just become the comment. 

Agree on this.

ejona86:
Converted this PR to just a comment, to be merged after #5947

PR #5955, opened 2019-07-04
core: fix terminated status check in TransportListener

creamsoup:
when we queue and execute the runnable in transport terminated, it may not be drained immediately if there's a already drain is running on the other thread. 
Adding a boolean to check shutdown is called before terminated.

### DIFF
--- a/core/src/main/java/io/grpc/internal/InternalSubchannel.java
+++ b/core/src/main/java/io/grpc/internal/InternalSubchannel.java
@@ -494,6 +494,7 @@ final class InternalSubchannel implements InternalInstrumented<ChannelStats>, Tr
   private class TransportListener implements ManagedClientTransport.Listener {
     final ConnectionClientTransport transport;
     final SocketAddress address;
+    boolean shutdownInitiated = false;
 
     TransportListener(ConnectionClientTransport transport, SocketAddress address) {
       this.transport = transport;
@@ -530,6 +531,7 @@ final class InternalSubchannel implements InternalInstrumented<ChannelStats>, Tr
     public void transportShutdown(final Status s) {
       channelLogger.log(
           ChannelLogLevel.INFO, "{0} SHUTDOWN with {1}", transport.getLogId(), printShortStatus(s));
+      shutdownInitiated = true;
       syncContext.execute(new Runnable() {
         @Override
         public void run() {
@@ -573,7 +575,7 @@ final class InternalSubchannel implements InternalInstrumented<ChannelStats>, Tr
           }
         }
       });
-      Preconditions.checkState(activeTransport != transport,
+      Preconditions.checkState(shutdownInitiated || activeTransport != transport,
           "activeTransport still points to this transport. "
           + "Seems transportShutdown() was not called.");
     }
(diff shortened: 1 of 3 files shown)


creamsoup:
internally tracked bug b/136496833

core/src/main/java/io/grpc/internal/InternalSubchannel.java
```diff
@@ -573,7 +575,7 @@ public void run() {
           }
         }
       });
-      Preconditions.checkState(activeTransport != transport,
+      Preconditions.checkState(shutdownInitiated || activeTransport != transport,
```
ejona86: Just check `shutdownInitiated`, and you can do it at the beginning of the method.

creamsoup: thanks done. had to fix few tests PTAL

core/src/test/java/io/grpc/internal/InternalSubchannelTest.java
```diff
@@ -1145,6 +1146,8 @@ public void channelzMembership() throws Exception {
 
     MockClientTransportInfo t0 = transports.poll();
     assertTrue(channelz.containsClientSocket(t0.transport.getLogId()));
+    t0.listener.transportShutdown(Status.RESOURCE_EXHAUSTED);
+    fakeClock.forwardNanos(10);
```
ejona86: Why is advancing the time necessary?

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -521,8 +521,10 @@ public void channelzMembership_subchannel() throws Exception {
     MockClientTransportInfo transportInfo = transports.poll();
     assertNotNull(transportInfo);
     assertTrue(channelz.containsClientSocket(transportInfo.transport.getLogId()));
+    transportInfo.listener.transportReady();
```
ejona86: Adding this to this test looks fine/natural (so no need for a change). I will just note that it shouldn't be _necessary_ for the transport to become ready before being shutdown. That happens during negotiation failure, for example.

core/src/test/java/io/grpc/internal/InternalSubchannelTest.java
creamsoup: shutdown triggers connection backoff which won't be executed if not advancing the time. this fails the check in After block (checking any pending tasks)

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
creamsoup: this is also related to reconnect, the shutdown triggers backoff and reconnect so the subchannel is not removed from channelz which makes the test fail.

core/src/test/java/io/grpc/internal/InternalSubchannelTest.java
ejona86: Go ahead and issue the ready, so the shutdown doesn't trigger a backoff. That is the more "normal" flow.

creamsoup: okay. done.

PR #5958, opened 2019-07-07
Bump PerfMark to 0.17.0

carl-mastrangelo:
The main changes how linking is done.  Linking is now always done
through the `PerfMark` entry class.   This is for two reasons:

1.  It make instrumenting the linking calls *much* easier.
2.  It follows the API pattern of "verbNoun()".  Previous callsites
    would have `Link link = PerfMark.link(); link.link()`.  This
    stuttering is not quick to follow.

Generated using:

```
find -name \*.java -exec sed -i 's#link = PerfMark.link();#link = PerfMark.linkOut();#g' {} \;
find -name \*.java -exec sed -i 's#link.link();#PerfMark.linkIn(link);#g' {} \;
find -name \*.java -exec sed -i 's#command.getLink().link();#PerfMark.linkIn(command.getLink());#g' {} \;
find -name \*.java -exec sed -i 's#cmd.getLink().link();#PerfMark.linkIn(cmd.getLink());#g' {} \;
find -name \*.java -exec sed -i 's#msg.getLink().link();#PerfMark.linkIn(msg.getLink());#g' {} \;
```

Since the deprecated link methods are also `@DoNotCall`, the same
sed calls will need to be used on import.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -201,7 +201,7 @@ subprojects {
             opencensus_impl: "io.opencensus:opencensus-impl:${opencensusVersion}",
             opencensus_impl_lite: "io.opencensus:opencensus-impl-lite:${opencensusVersion}",
             instrumentation_api: 'com.google.instrumentation:instrumentation-api:0.4.3',
-            perfmark: 'io.perfmark:perfmark-api:0.16.0',
+            perfmark: 'io.perfmark:perfmark-api:0.17.0',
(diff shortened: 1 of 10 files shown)


ran-su:
Is this a good time to do a PerfMark import?

carl-mastrangelo:
@ran-su I think you will need to temporarily remove the `@DoNotCall` annotations on `link()`, but yeah.  It makes it much easier to read out the trace data.

PR #5962, opened 2019-07-09
doc: fix version bug in documentation/android-channel-builder.md

dapengzhang0:


dapengzhang0:
@ejona86 I added 1.23 milestone. But I'm OK if it's not.

PR #5965, opened 2019-07-09
Fix mismatched Netty versions in repositories.bzl

aaliddell:
The version of netty-common imported by `repositories.bzl` is out of sync with the rest of the netty deps, leading to `NoSuchMethodError` exceptions due to changes in method signatures between the versions. Presently common is using 4.1.35, whilst the rest of netty is using 4.1.37. This PR updates common to match the others at 4.1.37.


dapengzhang0:
Thank you @aaliddell for catching and fixing this!

PR #5980, opened 2019-07-13
netty: don't share transport tracers across transports

carl-mastrangelo:
Found using TSAN, which shows the number of calls succeeded being incremented from multiple event loops


dapengzhang0:
Would this fix the TSAN issue in net/grpc/testing/interop/lb_integration_tests?

carl-mastrangelo:
No, but it makes it much less flaky.

PR #5987, opened 2019-07-17
alts: provide channel type to HandshakerServiceChannel to suppress warning

creamsoup:
resolves #5987


creamsoup:
background: we change the default channel type to Epoll if available. NettyChannelBuilder expects to have both channel type and event loop group to be provided (or neither to use default).

PR #5988, opened 2019-07-18
interop testing: add a gRPCLB fallback test client

apolcyn:
This is a java analogue of the C++ test added in https://github.com/grpc/grpc/pull/19623. Propose checking it into this repo also mostly to make it easier to run.

Proto changes were synced from grpc/grpc-proto

This client is suitable for testing fallback with any "grpclb" load-balanced service, but is particularly meant to implement a set of test cases described in an internal doc titled "DirectPath Cloud-to-Prod End-to-End Test Cases", section "gRPC DirectPath-to-CFE fallback".


apolcyn:
Windows failure appears to be a flake

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
```diff
@@ -0,0 +1,302 @@
...
+          + "\n      Default: " + c.testCase
+      );
+      System.exit(1);
+    }
+  }
+
+  private ManagedChannel createChannel() {
+    if (!customCredentialsType.equals("compute_engine_channel_creds")) {
+      throw new AssertionError(
+          "This test currently only supports "
```
zhangkun83: Is there a plan to support other types of creds?

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
```diff
@@ -0,0 +1,302 @@
...
+      channel.awaitTermination(1, TimeUnit.SECONDS);
+    } catch (Exception ex) {
+      throw new RuntimeException(ex);
+    }
+  }
+
+  private String streamToString(InputStream inputStream) throws Exception {
+    BufferedReader br = new BufferedReader(
+        new InputStreamReader(inputStream, UTF_8));
+    String out = "";
```
zhangkun83: Can use `CharStreams` from guava.

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
```diff
@@ -0,0 +1,302 @@
...
+    while ((line = br.readLine()) != null) {
+      out += line;
+    }
+    return out;
+  }
+
+  private void runShellCmd(String cmd) throws Exception {
+    logger.info("Run shell command: " + cmd);
+    Process process = Runtime.getRuntime().exec(new String[]{"bash", "-c", cmd});
+    int exitCode = process.waitFor();
```
zhangkun83: `waitFor()` should be put **after** draining the outputs. Otherwise the buffers may be full and that will block the child process. See:
https://wiki.sei.cmu.edu/confluence/display/java/FIO07-J.+Do+not+let+external+processes+block+on+IO+buffers

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
```diff
@@ -0,0 +1,302 @@
...
+        response = blockingStub
+          .withDeadlineAfter(deadlineSeconds, TimeUnit.SECONDS)
+          .unaryCall(request);
+      }
+      result = response.getGrpclbRouteType();
+    } catch (StatusRuntimeException ex) {
+      logger.warning("doRpcAndGetPath failed. Status: " + ex);
+    }
+    if (result == GrpclbRouteType.GRPCLB_ROUTE_TYPE_FALLBACK
+        || result == GrpclbRouteType.GRPCLB_ROUTE_TYPE_BACKEND) {
```
zhangkun83: Should this condition be inverted?

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
```diff
@@ -0,0 +1,302 @@
...
+      throw new AssertionError("Received invalid LB route type. This suggests "
+          + "that the server hasn't implemented this test correctly.");
+    }
+    logger.info("doRpcAndGetPath. GrpclbRouteType result: " + result);
```
zhangkun83: You may want to log this before the if-block above.

interop-testing/src/main/proto/grpc/testing/messages.proto
```diff
@@ -84,7 +102,9 @@ message SimpleResponse {
   string oauth_scope = 3;
   // Server ID. This must be unique among different server instances,
   // but the same across all RPC's made to a particular server instance.
-  string server_id = 4;
+ string server_id = 4;
```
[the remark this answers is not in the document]
apolcyn: Whoops, fixed

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
apolcyn: I think so. I added a TODO into the error message for it. I can add it into this PR if preferred.

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
apolcyn: Yes, good catch.

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
apolcyn: Good point, done

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
apolcyn: Interesting, fixed

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
apolcyn: done

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
```diff
@@ -0,0 +1,296 @@
...
+          .withWaitForReady()
+          .unaryCall(request);
+      } else {
+        response = blockingStub
+          .withDeadlineAfter(deadlineSeconds, TimeUnit.SECONDS)
+          .unaryCall(request);
+      }
+      result = response.getGrpclbRouteType();
+    } catch (StatusRuntimeException ex) {
+      logger.warning("doRpcAndGetPath failed. Status: " + ex);
```
ejona86: I'm surprised this exception isn't propagated. It seems if this triggers `result` will be `UNKNOWN` which will fail in the assertion below and would be a misleading message.

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
```diff
@@ -0,0 +1,296 @@
...
+      String breakLbAndBackendConnsCmd) throws Exception {
+    assertEquals(
+        doRpcAndGetPath(blockingStub, 20, RpcMode.FailFast),
+        GrpclbRouteType.GRPCLB_ROUTE_TYPE_BACKEND);
+    runShellCmd(breakLbAndBackendConnsCmd);
+    for (int i = 0; i < 40; i++) {
+      GrpclbRouteType grpclbRouteType = doRpcAndGetPath(
+          blockingStub, 1, RpcMode.WaitForReady);
+      if (grpclbRouteType != GrpclbRouteType.GRPCLB_ROUTE_TYPE_BACKEND) {
+        throw new AssertionError("Got grpclb route type backend. Backends are "
```
ejona86: You did _not_ get route type backend. Is the condition wrong or the message?

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
```diff
@@ -0,0 +1,296 @@
...
+            + "supposed to be unreachable, so this test is broken");
+      }
+      if (grpclbRouteType == GrpclbRouteType.GRPCLB_ROUTE_TYPE_FALLBACK) {
+        logger.info("Made one successful RPC to a fallback. Now expect the "
```
ejona86: FYI: you can word wrap at 100 columns instead of 80.

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
```diff
@@ -0,0 +1,296 @@
...
+  private void runShellCmd(String cmd) throws Exception {
+    logger.info("Run shell command: " + cmd);
+    ProcessBuilder pb = new ProcessBuilder("bash", "-c", cmd);
+    pb.redirectErrorStream(true);
+    Process process = pb.start();
+    logger.info("Shell command merged stdout and stderr: "
+        + CharStreams.toString(
+            new InputStreamReader(process.getInputStream(), UTF_8)));
+    int exitCode = process.waitFor();
+    logger.info("Shell rommand exit code: " + exitCode);
```
ejona86: s/rommand/command/

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
apolcyn: This was broken, sorry. Changed to return `UNKNOWN` early in case of RPC failure

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
apolcyn: The condition was wrong, fixed.

Thanks for comments, sorry for batch of errors. I pushed a half-done set of changes earlier

interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java
apolcyn: fixed

apolcyn:
Android error seems to be a build infra flake:

```

FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':grpc-context:compileJava'.
> Could not resolve all files for configuration ':grpc-context:errorproneJavac'.
   > Could not download javac.jar (com.google.errorprone:javac:9+181-r4173-1)
      > Could not get resource 'https://maven-central.storage-download.googleapis.com/repos/central/data/com/google/errorprone/javac/9+181-r4173-1/javac-9+181-r4173-1.jar'.
         > Premature end of Content-Length delimited message body (expected: 6885456; received: 4456448

```

dapengzhang0:
OK to merge?

ejona86:
@apolcyn, what's happened on this PR? How have we gone this long without merging? Do we not need this test client?

apolcyn:
@ejona86 sorry this got left behind.

I saw that this worked manually while developing the PR, but never just got around to wiring up this test to run continuously (in any language, not just Java). I think it is still worth having though, mostly as a demo.

I merged with master, and just checked that this still builds manually, with:

```
./gradlew --no-daemon :grpc-interop-testing:installDist -PskipCodegen=true -PskipAndroid=true
```

I've been trying to re-<i>run</i> the test but haven't yet gotten to as I've been wrestling with some infrastructural pre-reqs.

Besides that, if this still looks ok to merge to you, then I think it should still be ready to do.

interop-testing/build.gradle
```diff
@@ -114,6 +115,17 @@ task grpclb_long_lived_affinity_test_client(type: CreateStartScripts) {
     ]
 }
 
+task grpclb_fallback_test_client (type: CreateStartScripts) {
+    mainClassName = "io.grpc.testing.integration.GrpclbFallbackTestClient"
+    applicationName = "grpclb-fallback-test-client"
+    outputDir = new File(project.buildDir, 'tmp')
+    classpath = jar.outputs.files + configurations.runtime
+    defaultJvmOpts = [
+        "-Dio.grpc.internal.DnsNameResolverProvider.enable_grpclb=true",
```
ejona86, line 124: This was not necessary since v1.24.2 and since v1.29.0 it no longer does anything.

interop-testing/build.gradle
```diff
@@ -34,6 +34,7 @@ dependencies {
     compileOnly libraries.javax_annotation
     runtimeOnly libraries.opencensus_impl,
             libraries.netty_tcnative,
+            libraries.netty_epoll,
```
ejona86, line 37: This should not be necessary; ALTS is using grpc-netty-shaded which includes epoll, but also _can't_ use this epoll dep.

ejona86:
It seems we need a link to provide more details about this client. It isn't clear the wider environment it lives in. Maybe it should have a link to the interop test definitions? (IIRC, that is an internal link, but an internal link is better than no link.)

interop-testing/build.gradle
apolcyn: removed

interop-testing/build.gradle
apolcyn: removed

apolcyn:
> It seems we need a link to provide more details about this client. It isn't clear the wider environment it lives in. Maybe it should have a link to the interop test definitions? (IIRC, that is an internal link, but an internal link is better than no link.)

Yes, the details of the spec are in an internal doc.

Because of the names of the paths in a link to that internal doc, I'm not sure it's appropriate to link to from here. I've noted the title and section of the doc though, so it can be easily searched for.

ejona86:
> I've noted the title and section of the doc though, so it can be easily searched for.

That sounds fine, but I don't see the change.

apolcyn:
I had commented about it in this PR's description, but just updated `interop-testing/src/main/java/io/grpc/testing/integration/GrpclbFallbackTestClient.java` to refer to the spec.

ejona86:
Okay, the "rough shape" of things look good. I'll let Kun's review stand, but won't LGTM myself, just because I didn't look at it _that_ closely.

PR #5991, opened 2019-07-19
stub: avoid race in onHalfClose server StreamObserver

carl-mastrangelo:


zhangkun83:
LGTM

stub/src/main/java/io/grpc/stub/ServerCalls.java
```diff
@@ -171,7 +172,7 @@ public void onHalfClose() {
         method.invoke(request, responseObserver);
         request = null;
         responseObserver.freeze();
-        if (call.isReady()) {
+        if (!wasReady) {
```
ejona86: This is inverted. It should be `if (wasReady)`

carl-mastrangelo: Oh, I thought you said onReady should only be called once, and it was which ever method (onHalfClose or onReady) was invoked first.  Is that not correct?

ejona86: Before `onHalfClose()`, `onReadyHandler == null`. So if there was an onReady callback, the code "ignored" it. The onReady() callback will generally happen before onHalfClose(). This check was to "replay" that missed onReady call.

carl-mastrangelo: I see, updated.

PR #5996, opened 2019-07-21
core: Migrate to new OpenCensus method & status tags

### LINKED ISSUE #5593: OpenCensus grpc method is not set

igorbernstein2:
Fixes #5593 and supersedes #5601

Now that https://github.com/census-instrumentation/opencensus-java/pull/1854 has been merged & released as 0.21.0. We can start using the method & status tags.

Background:
Opencensus introduced new tags for status and method (https://github.com/census-instrumentation/opencensus-java/pull/1115). The old views that used those tags were deprecated and new views were created that used the new tags. However grpc-java wasn't updated to use the new tags due to concern of breaking existing metrics. This resulted in the old views being deprecated while the new views were broken (goomics <span>#</span>50).

https://github.com/census-instrumentation/opencensus-java/pull/1854 added a compatibility layer to opencensus that would remap new tags to old tags for old views. This should unblock grpc to switching to the new tags while allowing old views to still be populated. That commit was released as part of opencensus 0.21, which grpc currently uses.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -26,7 +26,8 @@ dependencies {
             project(':grpc-api').sourceSets.test.output,
             project(':grpc-testing'),
             project(':grpc-grpclb'),
-            libraries.guava_testlib
+            libraries.guava_testlib,
+            libraries.opencensus_impl
 
     signature "org.codehaus.mojo.signature:java17:1.0@signature"
     signature "net.sf.androidscents.signature:android-api-level-14:4.0_r4@signature"
--- a/core/src/test/java/io/grpc/internal/CensusModulesTest.java
+++ b/core/src/test/java/io/grpc/internal/CensusModulesTest.java
@@ -17,6 +17,7 @@
 package io.grpc.internal;
 
 import static com.google.common.truth.Truth.assertThat;
+import static com.google.common.truth.Truth.assertWithMessage;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -40,6 +41,7 @@ import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.Mockito.verifyZeroInteractions;
 import static org.mockito.Mockito.when;
 
+import com.google.common.collect.ImmutableList;
 import io.grpc.Attributes;
 import io.grpc.CallOptions;
 import io.grpc.Channel;
@@ -61,8 +63,19 @@ import io.grpc.internal.testing.StatsTestUtils.FakeTagContextBinarySerializer;
 import io.grpc.internal.testing.StatsTestUtils.FakeTagger;
 import io.grpc.internal.testing.StatsTestUtils.MockableSpan;
 import io.grpc.testing.GrpcServerRule;
+import io.opencensus.common.Function;
+import io.opencensus.common.Functions;
 import io.opencensus.contrib.grpc.metrics.RpcMeasureConstants;
+import io.opencensus.contrib.grpc.metrics.RpcViewConstants;
+import io.opencensus.impl.stats.StatsComponentImpl;
+import io.opencensus.stats.AggregationData;
+import io.opencensus.stats.AggregationData.CountData;
+import io.opencensus.stats.AggregationData.LastValueDataDouble;
+import io.opencensus.stats.AggregationData.LastValueDataLong;
+import io.opencensus.stats.AggregationData.SumDataDouble;
 import io.opencensus.stats.Measure;
+import io.opencensus.stats.StatsComponent;
+import io.opencensus.stats.View;
 import io.opencensus.tags.TagContext;
 import io.opencensus.tags.TagValue;
 import io.opencensus.trace.BlankSpan;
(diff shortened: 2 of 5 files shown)


igorbernstein2:
CC @songy23, @sebright, @dapengzhang0

core/src/test/java/io/grpc/internal/CensusModulesTest.java
```diff
@@ -1148,4 +1166,92 @@ private static void assertNoClientContent(StatsTestUtils.MetricsRecord record) {
     assertNull(record.getMetric(DeprecatedCensusConstants.RPC_CLIENT_UNCOMPRESSED_REQUEST_BYTES));
     assertNull(record.getMetric(DeprecatedCensusConstants.RPC_CLIENT_UNCOMPRESSED_RESPONSE_BYTES));
   }
+
+  @SuppressWarnings("deprecation")
```
zhangkun83: Our convention is to annotate the test method as `@Derepcated`. This indicates these tests are to be deleted once the deprecated API under test is deleted.

core/src/test/java/io/grpc/internal/CensusModulesTest.java
```diff
@@ -1148,4 +1166,92 @@ private static void assertNoClientContent(StatsTestUtils.MetricsRecord record) {
...
+          getAggregationValueAsLong(
+              localStats,
+              RpcViewConstants.GRPC_CLIENT_COMPLETED_RPC_VIEW,
+              ImmutableList.of(
+                  TagValue.create(method.getFullMethodName()),
+                  TagValue.create("PERMISSION_DENIED"))))
+        .isEqualTo(1);
+  }
+
+  @SuppressWarnings("deprecation")
```
zhangkun83: ditto

songy23:
> @songy23, can you confirm it's safe to switch to the new tags for both OSS and internal usages?

Yes it should be safe to do the switch. The new tags will be mapped properly to keep backwards-compatibility.

core/src/test/java/io/grpc/internal/CensusModulesTest.java
igorbernstein2: done

core/src/test/java/io/grpc/internal/CensusModulesTest.java
igorbernstein2: done

igorbernstein2:
Thanks for the review, all feedback should be addressed. I'll be OOO for the rest of the week, so if anything else needs to change feel free to update the PR as needed

core/src/test/java/io/grpc/internal/CensusModulesTest.java
```diff
@@ -1148,4 +1166,92 @@ private static void assertNoClientContent(StatsTestUtils.MetricsRecord record) {
     assertNull(record.getMetric(DeprecatedCensusConstants.RPC_CLIENT_UNCOMPRESSED_REQUEST_BYTES));
     assertNull(record.getMetric(DeprecatedCensusConstants.RPC_CLIENT_UNCOMPRESSED_RESPONSE_BYTES));
   }
+
+  @Deprecated
+  @Test
+  public void newTagsPopulateOldViews() throws InterruptedException {
+    StatsComponent localStats = new StatsComponentImpl();
```
songy23: Why do you need this? As an alternative consider:
```java
private static final ViewManager viewManager = Stats.getViewManager();
...
viewManager.registerView(...);
```

core/build.gradle
```diff
@@ -26,7 +26,8 @@ dependencies {
             project(':grpc-api').sourceSets.test.output,
             project(':grpc-testing'),
             project(':grpc-grpclb'),
-            libraries.guava_testlib
+            libraries.guava_testlib,
+            libraries.opencensus_impl
```
songy23: `opencensus_impl` should be `testRuntimeOnly` dependency and shouldn't be needed for compiling.

core/src/test/java/io/grpc/internal/CensusModulesTest.java
igorbernstein2: I didn’t want the unit test to modify global state.

core/build.gradle
igorbernstein2: Will do when I get to a computer

core/src/test/java/io/grpc/internal/CensusModulesTest.java
songy23: OK that makes sense. In this case you don't need to modify the dependencies in build.

core/build.gradle
igorbernstein2: Looks like this can't be done because I need access to StatsComponentImpl to avoid mutating global state

igorbernstein2:
I think this is ready to be merged

songy23:
@zhangkun83 Can you add the label to trigger Kokoro?

igorbernstein2:
Thank you

Issue #5460, opened 2019-03-11
Please, extend support for the GraalVM native JVM target


ludoch:
See some work done on 
https://github.com/hpehl/quarkus-grpc-extension/blob/master/runtime/src/main/java/io/quarkus/grpc/runtime/graal/GrpcNettySubstitutions.javas an example...

If this is backed in in the library, then all users would benefit.
See https://github.com/googleapis/google-cloud-java/issues/4664 for more context.

carl-mastrangelo:
@ludoch Can you clarify what you are asking for?  (also, your first link appears to be to a hidden repo, I can't see it).

ludoch:
link is https://github.com/hpehl/quarkus-grpc-extension/blob/master/runtime/src/main/java/io/quarkus/grpc/runtime/graal/GrpcNettySubstitutions.java

It seems that Graal Vm does not support 100% of the Java APIs, as seen in https://github.com/oracle/graal/blob/master/substratevm/LIMITATIONS.md
The code sample above is the current QuarkIO to make some gprc working with it.

Repo is https://github.com/hpehl/quarkus-grpc-extension
Sample code at: https://github.com/hpehl/quarkus-grpc-quickstart

I see for example:
```
@TargetClass(className = "io.grpc.netty.ProtocolNegotiators")
final class Target_io_grpc_netty_ProtocolNegotiators
```
so this is a way to rewrite some classes so they work with Graal VM.
Not sure if these extensions should be in framewoek libs or handled by core libraries like GRPC.

A troubling line is this one:
```    @Substitute
    private static sun.misc.Unsafe getUnsafe() {
        // since we configured Netty with io.netty.noUnsafe=true, this should be safe
        return null;
    }
```
which may cause trouble down the line for other libraries that depend on this call not being null...
Not sure.

ejona86:
`getUnsafe` isn't surprising and code should be able to handle unsafe being unavailable (although potentially be slower). Note also that that line appears to be for Protobuf's benefit, which I can believe given how it said Graal rewrites index numbers for offsets. This seems like the proper "fix".

The other piece is logSslEngineDetails, which is interesting. The change seems fine, as it basically disables the method, and that doesn't hurt much. It's unclear why it was needed, though.

There doesn't seem to be much to do here; the status-quo looks fine.

I would expect actually running to have a few more issues. For example, Netty uses finalizers, which could leak memory if they aren't being called.

ejona86:
[...]

PR #6006, opened 2019-07-24
netty: bump to 4.1.38

carl-mastrangelo:


viktorklang:
@carl-mastrangelo Hi Carl,
is there any chance for updating 1.22.2+ to Netty 4.1.36+ (Graal support)?
Or will we have to wait for 1.23.0? (Since grpc-java includes a shaded version of netty it isn't that simple to just override the Netty dependency when building)
Thoughts?

Cheers,
√

viktorklang:
@carl-mastrangelo Through adding Netty 4.1.38 to 1.22.2-SNAPSHOT and publishing locally, I'm seeing an issue that needs to be taken care of—the shade plugin which generates the grpc-netty-shaded.jar needs to be augmented to rewrite the FQCNs and package names in the Graal configuration which is in, for instance: https://github.com/netty/netty/tree/4.1/common/src/main/java/io/netty/util/internal/svm and in https://github.com/netty/netty/blob/4.1/transport/src/main/resources/META-INF/native-image/io.netty/transport/reflection-config.json and in https://github.com/netty/netty/tree/4.1/transport/src/main/resources/META-INF/native-image

viktorklang:
All of the above relates to #5460

ejona86:
@viktorklang, we don't generally upgrade the Netty version in a previous release unless there's a significant reason to, because new Netty versions have a significant amount of risk which is generally inappropriate for a patch release.

You are free to use `grpc-netty` instead of `grpc-netty-shaded` and then select the version of Netty (although we really only support the versions of Netty we select for releases).

carl-mastrangelo:
@viktorklang I'm afraid I'm super unfamiliar with Graal and configuring it.   I'm happy to review a PR if you want to send one.

viktorklang:
@carl-mastrangelo You and me both! :) All I know is that there are some META-INF folders which needs to be search-replace renamed, and some strings within text files under META-INF which needs to be search-replaced. Unfortunately I found no obvious way of doing search-replace with the shadow plugin (and I'm no Groovy/Gradle expert by far) so I am unable to be of much help there. But if someone can get the rename/search-replace machinery working then I can help with what needs replacing. :)

viktorklang:
@ejona86 Unfortunately I cannot switch to grpc-netty since I have a 3rd party which compiles against grpc-netty-shaded :(

PR #6022, opened 2019-07-29
compiler: add note about where to download precompiled version of plugin

jadekler:
It is very confusing trying to figure it out yourself.


creamsoup:
merged, thanks @jadekler

PR #6024, opened 2019-07-29
buildscripts: add an optional gradle task for google-java-format

dapengzhang0:
- Add gradle task 
  ```./gradlew googleJavaFormat [-PformatBase=<format base>]```
and an option for build task
  ```./gradlew build -PformatBase=<format base>```
- The following lines of code is just to packet a fat jar `google-java-format-1.7-all-deps.jar` from maven artifacts. In fact, the fat jar can simply be downloaded from github with wget https://github.com/google/google-java-format/releases/download/google-java-format-1.7/google-java-format-1.7-all-deps.jar
But my concern is whether their github link is persistent. 

  ```
  repositories {
      maven { // The google mirror is less flaky than mavenCentral()
          url 'https://maven-central.storage-download.googleapis.com/repos/central/data/' }
  }

  configurations {
      googleJavaFormat {}
  }

  dependencies {
      googleJavaFormat "com.google.googlejavaformat:google-java-format:${googleJavaFormatVer}"
  }

  task googleJavaFormatFatJar(type: Jar) {
      manifest {
          attributes 'Main-Class': 'com.google.googlejavaformat.java.Main'
      }
      destinationDirectory = file("${rootDir}/build/libs")
      archiveFileName = "google-java-format-${googleJavaFormatVer}-all-deps.jar"
      from { configurations.googleJavaFormat.collect { it.isDirectory() ? it : zipTree(it) } }
      googleJavaFormatJar = "${destinationDirectory.get()}/${archiveFileName.get()}"
  }
  ```


creamsoup:
how does work with existing checkStyle{Main, Test}? should we drop and add new tasks using google-java-formatter?

dapengzhang0:
@creamsoup This won't happen in the near future. There seems no clean solution (as @ejona86 suggested) while not losing good usability.

creamsoup:
@dapengzhang0 okay. it would be nice if we have a optional gradle task that performs google-java-format. i'll definitely use it.

dapengzhang0:
@creamsoup If just an optional task, that can be achieved.

dapengzhang0:
@ejona86 PTALA.

google-java-format.gradle
```diff
@@ -0,0 +1,71 @@
...
+            executable = '/usr/bin/env'
+            args 'python2.7'
+        }
+    } catch (Exception e) {
+        print "python2.7 not found, will skip ${name}"
+        enabled = false
+    }
+}
+
+if (googleJavaFormat.enabled) {
```
ejona86: Everything in here seems safe to be unconditional (it won't cause any processing).

google-java-format.gradle
```diff
@@ -0,0 +1,71 @@
...
+    def googleJavaFormatVer = '1.7'
+    def googleJavaFormatDir = "${rootDir}/buildscripts/google-java-format"
+    def googleJavaFormatJar
+
+    repositories {
```
ejona86: I'm concerned about "random" other files changing the repositories. I'd prefer this be in `build.gradle` itself.

google-java-format.gradle
```diff
@@ -0,0 +1,71 @@
+/**
+ * Reformat the diff from a given commit, formatBase. By default formatBase=HEAD.
+ * Requires that bash can run executable python 2.7 files directly.
+ *
+ * Usage:
+ * ./gradlew googleJavaFormat [-PformatBase=<format base>]
+ * 
+ * To enable this task when build, explicitly specify property "formatBase"
```
ejona86: Don't abuse the `formatBase` configuration for this. Use a separate property, like "formatJavaOnBuild".

You might also want to make it clear you are referring to the `build` _task_.

dapengzhang0:
Close because non of the proposed approaches seem to work very well, as the behavior google-java-format tool is not stable and it changes the whole unmodified block that is adjacent to modified lines of code which is not desirable.

Issue #6030, opened 2019-07-31
Tracking issue for Deadline.Ticker and related APIs


zhangkun83:

kluever:
Many of these APIs would change dramatically if `java.time` types could be used. Are there any plans for Java8+ gRPC support?

PR #6031, opened 2019-07-31
context: allow creating Deadline with custom Ticker.

zhangkun83:
This makes `Deadline` more test-friendly.  Next step is to allow
`ServerBuilder` to take a custom `Ticker` and use it for creating
incoming `Deadline`s.  With both changes in, application logic will be
able to verify the Deadlines they set.

For #2531
Tracked by #6030


zhangkun83:
I also renamed `read()` to `nanoTime()`, because I found `read()` is not self-explanatory about what it returns. Other naming suggestions are welcome.

PR #6039, opened 2019-08-06
Remove missing pattern from glob

laurentlb:
There's no third_party directory.
This change makes the repository compatible with
Bazel change --incompatible_disallow_empty_glob.


dapengzhang0:
Thank you @laurentlb for the fix.

PR #6049, opened 2019-08-10
grpclb: shuffle pick first index

dapengzhang0:
Each new instance of GrpcLbBalancer will have a different pick first index.

### DIFF
--- a/grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
+++ b/grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
@@ -104,11 +104,13 @@ final class GrpclbState {
       }
     };
 
-  static enum Mode {
+  enum Mode {
     ROUND_ROBIN,
     PICK_FIRST,
   }
 
+  // backend list for pick_first will be picked from this index
+  private final int pickFirstIndex;
   private final String serviceName;
   private final Helper helper;
   private final SynchronizationContext syncContext;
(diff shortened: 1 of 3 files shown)


grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
```diff
@@ -415,20 +419,24 @@ private void useRoundRobinLists(
...
         for (BackendAddressGroup bag : newBackendAddrList) {
           EquivalentAddressGroup origEag = bag.getAddresses();
           Attributes eagAttrs = origEag.getAttributes();
           if (bag.getToken() != null) {
             eagAttrs = eagAttrs.toBuilder()
                 .set(GrpclbConstants.TOKEN_ATTRIBUTE_KEY, bag.getToken()).build();
           }
-          eagList.add(new EquivalentAddressGroup(origEag.getAddresses(), eagAttrs));
+          eagList.set(
+              offset++ % numOfEags,
```
ejona86: Hmmm... Techincally, `pickFirstIndex` could be `MAX_VALUE` and this would wraparound. Since you already chose to use Math.abs(), you could maybe replace that with something akin to `pickFirstGlobalIndex.getAndIncrement() & 0xFFFF`.

dapengzhang0: Done.

dapengzhang0:
FYI @dfawley

PR #5759, opened 2019-05-17
xds: implement Fallback-at-Startup mode

dapengzhang0:
This is the implementation of the Fallback-at-Startup mode in the [design doc](https://docs.google.com/document/d/1fAfdllHJ3JaaKVIdrrBaMnZSHxGAnlx4BhABMQQOn_4/view?pli=1#heading=h.yoesuc41tq7r).

- The Fallback-After-Startup mode is not implemented.
- Drop related behavior is not implemented.

PR #6070, opened 2019-08-15
xds: restructure XdsLoadBalancer part 1

dapengzhang0:
- Get rid of `XdsLbState` and make config change handling into two layers: `LookasideLb` (handles balancer name) and `LookasideChannelLb` (handles child policy), under `XdsLoadBalanecer` (fallback manager layer)
- Move `XdsComms`/`AdsStream` to a layer under `LookasideChannelLb`. They don't keep the helper, but only `SyncCtx` and `ChannelLogger`
- For each layer, we pass in a `LoadBalancer.Factory` for the next/child layer. In test, 
  + we mock/fake the factory, so we don't care about the implementation details of the child layer.
  + we capture the helper of `factory.newBalancer(factory)`, so we can mimic `updateBalancingState() `from the child layer.

Part 1 contains fallback management logic. There is no change in fallback management logic.

### DIFF
--- a/xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
+++ b/xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
@@ -1,0 +1,267 @@
+/*
+ * Copyright 2019 The gRPC Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.xds;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+import com.google.common.annotations.VisibleForTesting;
+import io.grpc.ChannelLogger.ChannelLogLevel;
+import io.grpc.ConnectivityState;
+import io.grpc.ConnectivityStateInfo;
+import io.grpc.LoadBalancer;
+import io.grpc.Status;
+import io.grpc.SynchronizationContext.ScheduledHandle;
+import io.grpc.util.ForwardingLoadBalancerHelper;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
+
+/**
+ * A {@link LoadBalancer} that uses the XDS protocol.
+ *
+ * <p>This class manages fallback handling. The logic for child policy handling and fallback policy
+ * handling is provided by LookasideLb and FallbackLb.
+ */
+// TODO(zdapeng): migrate name to XdsLoadBlancer
+final class XdsLoadBalancer2 extends LoadBalancer {
+
+  private static final long FALLBACK_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(10); // same as grpclb
+
+  private final Helper helper;
+  private final LoadBalancer lookasideLb;
+  private final LoadBalancer.Factory fallbackLbFactory;
+  private final AdsCallback adsCallback = new AdsCallback() {
+    @Override
+    public void onWorking() {
+      if (childPolicyHasBeenReady) {
+        // cancel Fallback-After-Startup timer if there's any
+        cancelFallbackTimer();
+      }
+
+      childBalancerWorked = true;
+    }
+
+    @Override
+    public void onError() {
+      if (!childBalancerWorked) {
+        // start Fallback-at-Startup immediately
+        useFallbackPolicy();
+      } else if (childPolicyHasBeenReady) {
+        // TODO: schedule a timer for Fallback-After-Startup
+      } // else: the Fallback-at-Startup timer is still pending, noop and wait
+    }
+
+    @Override
+    public void onAllDrop() {
+      cancelFallback();
+    }
+  };
+
+  @Nullable
+  private LoadBalancer fallbackLb;
+  @Nullable
+  private ResolvedAddresses resolvedAddresses;
+  // Scheduled only once.  Never reset.
+  @CheckForNull
+  private ScheduledHandle fallbackTimer;
+  private boolean childBalancerWorked;
+  private boolean childPolicyHasBeenReady;
+
+  // TODO(zdapeng): Add XdsLoadBalancer2(Helper helper) with default factories
+  @VisibleForTesting
+  XdsLoadBalancer2(
+      Helper helper,
+      LookasideLbFactory lookasideLbFactory,
+      LoadBalancer.Factory fallbackLbFactory) {
+    this.helper = helper;
+    this.lookasideLb = lookasideLbFactory.newLoadBalancer(new LookasideLbHelper(), adsCallback);
+    this.fallbackLbFactory = fallbackLbFactory;
+  }
+
+  @Override
+  public boolean canHandleEmptyAddressListFromNameResolution() {
+    return true;
+  }
+
+  @Override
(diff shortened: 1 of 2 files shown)


xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
```diff
@@ -0,0 +1,246 @@
...
+
+  private static final long FALLBACK_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(10); // same as grpclb
+
+  private final Helper helper;
+  private final LoadBalancer lookasideLb;
+  private final LoadBalancer.Factory fallbackLbFactory;
+  private final AdsStreamCallback adsCallback = new AdsStreamCallback() {
+    @Override
+    public void onWorking() {
+      if (childPolicyHasBeenReady) {
```
zhangkun83: It's better to be consistent in terminology. IIUC, this variable is synonyms for "After-Startup", then probably better to name it as `afterStartup`.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
```diff
@@ -0,0 +1,246 @@
...
+        // cancel Fallback-After-Startup timer if there's any
+        cancelFallbackTimer();
+      }
+
+      childBalancerWorked = true;
```
zhangkun83: "Child balancer" looks too close to "child policy". "adsWorked" would stand out better.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
```diff
@@ -0,0 +1,246 @@
...
+    public void onAllDrop() {
+      cancelFallback();
+    }
+  };
+
+  @Nullable
+  private LoadBalancer fallbackLb;
+  @Nullable
+  private ResolvedAddresses resolvedAddresses;
+  // Scheduled only once.  Never reset.
```
zhangkun83: It will be reset in the "Fallback-After-Startup" case, right? Since you already have stubs for "Fallback-After-Startup" in the code, the documentation should be consistent.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
voidzcy: Oh we could maintain a status for XdsLoadBalancer's working state, such as `STARTUP`, `NORMAL`, `FALLBACK`, if that makes things clearer.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
```diff
@@ -0,0 +1,246 @@
...
+  @Override
+  public void handleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
+    this.resolvedAddresses = resolvedAddresses;
+
+    if (isInFallbackMode()) {
+      fallbackLb.handleResolvedAddresses(this.resolvedAddresses);
+    }
+
+    if (fallbackTimer == null) {
+      class FallbackTask implements Runnable {
```
voidzcy: More precisely, `EnterFallbackTask`.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
```diff
@@ -0,0 +1,246 @@
...
+ */
+// TODO(zdapeng): migrate name to XdsLoadBlancer
+final class XdsLoadBalancer2 extends LoadBalancer {
+
+  private static final long FALLBACK_TIMEOUT_MS = TimeUnit.SECONDS.toMillis(10); // same as grpclb
+
+  private final Helper helper;
+  private final LoadBalancer lookasideLb;
+  private final LoadBalancer.Factory fallbackLbFactory;
+  private final AdsStreamCallback adsCallback = new AdsStreamCallback() {
```
voidzcy: This is `FallbackHandler` with `onBalancerWorking()`, `onBalancerError()` and `onBalancerAllDrop()` as Fallback balancer is the implementer and the handler is injected into lookaside balancer. This also makes the naming more semantically meaningful.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
dapengzhang0: For the var names and implementation logic, I tried to keep them the same as the original `XdsLoadBalancer` unless it's obviously necessary, I don't want to fix too much in these restructure PRs.

Actually I was originally using exactly `afterStartup` and changed to `childPolicyHasBeenReady` as suggested. Please see the discussion
https://github.com/grpc/grpc-java/pull/5759#discussion_r286289714

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
dapengzhang0: `AdsStreamCallback` vs `FallbackHandler` was disscussed previously https://github.com/grpc/grpc-java/pull/5356/commits/1483cd356725f75d06b13609e0b3c50beceb943e#r257041208 and offline. Beside this, it will not only be used for fallback, `lrsClient` will also wrap/add listener to it.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
dapengzhang0: Changed to `Never reset to null.`

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
```diff
@@ -0,0 +1,246 @@
...
+      LookasideLbFactory lookasideLbFactory,
+      LoadBalancer.Factory fallbackLbFactory) {
+    this.helper = helper;
+    this.lookasideLb = lookasideLbFactory.newLoadBalancer(new LookasideLbHelper(), adsCallback);
+    this.fallbackLbFactory = fallbackLbFactory;
+  }
+
+  @Override
+  public boolean canHandleEmptyAddressListFromNameResolution() {
+    return true;
```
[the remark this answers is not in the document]
dapengzhang0: Maybe let's keep it as is, and fix later (as well as javadoc).

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
```diff
@@ -0,0 +1,246 @@
...
+    LoadBalancer balancer;
+
+    @Override
+    protected Helper delegate() {
+      return helper;
+    }
+
+    @Override
+    public void updateBalancingState(ConnectivityState newState, SubchannelPicker newPicker) {
+      checkNotNull(balancer, "there is a bug");
```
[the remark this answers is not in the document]
dapengzhang0: Changed.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
dapengzhang0: Done.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
dapengzhang0: Done.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
```diff
@@ -0,0 +1,246 @@
...
+  @Override
+  public boolean canHandleEmptyAddressListFromNameResolution() {
+    return true;
+  }
+
+  @Override
+  public void handleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
+    this.resolvedAddresses = resolvedAddresses;
+
+    if (isInFallbackMode()) {
```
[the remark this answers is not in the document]
voidzcy: You are right. The whole `resolvedAddresses` is saved and to be used later in entering fallback mode.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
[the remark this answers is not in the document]
zhangkun83: This is fine as we don't support fallback at this moment. Please add a TODO here, so that we need to revisit it when we officially support fallback.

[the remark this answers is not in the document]
dapengzhang0: TODO added.

PR #6076, opened 2019-08-16
xds: restructure XdsLoadBalancer Part 2: FallbackLb

dapengzhang0:


xds/src/main/java/io/grpc/xds/FallbackLb.java
```diff
@@ -0,0 +1,106 @@
...
+          backends.add(eag);
+        }
+      }
+      servers = backends.build();
+    }
+
+    // TODO(zhangkun83): FIXME(#5496): this is a temporary hack.
+    if (servers.isEmpty()
+        && !fallbackPolicyLb.canHandleEmptyAddressListFromNameResolution()) {
+      fallbackPolicyLb.handleNameResolutionError(Status.UNAVAILABLE.withDescription(
```
zhangkun83: You don't want to do this because this will trigger #5692. @ejona86, I think this again indicates the `canHandleEmptyAddressListFromNameResolution()` API doesn't work well. In #5803 we got away by saying `AutoConfiguredLoadBalancer` is not a `LoadBalancer`. We can do the same here. WDYT?

dapengzhang0: The problem of #5692 is that name resolver will not refresh, but here it doesn't seem it needs to refresh. Xds policy receiving empty address list is a normal usecase, and fallback policy being round robin is also a normal usecase. As long as there is no resolution error at the top level xds policy, we don't need to refresh the resolver.

zhangkun83: If fallback is round-robin, but the addresses are empty, don't you need a refresh?

dapengzhang0: If the xds policy does not explicitly provide fallback policy, then we use default fallback policy which is round-robin, and in this case the addresses being empty is normal . The fallback at startup timer is only 10 seconds, I think it does not make much sense to refresh resolver when the timer expires.

ejona86: I'd suggest this class's implementation of `canHandleEmptyAddressListFromNameResolution()` should delegate to `fallbackPolicyLb`, but `canHandleEmptyAddressListFromNameResolution()` is called before configuration is provided. That seems like a pretty serious failing. I agree that we should discuss alternatives with the other languages.

dapengzhang0: As discussed offline. We can leave the behavior unchanged.

xds/src/main/java/io/grpc/xds/FallbackLb.java
```diff
@@ -0,0 +1,106 @@
...
+
+    LbConfig fallbackPolicy = xdsConfig.fallbackPolicy;
+    String newFallbackPolicyName = fallbackPolicy.getPolicyName();
+    fallbackPolicyLb.switchTo(lbRegistry.getProvider(newFallbackPolicyName));
+
+    List<EquivalentAddressGroup> servers = resolvedAddresses.getAddresses();
+    // Some addresses in the list may be grpclb-v1 balancer addresses, so if the fallback policy
+    // does not support grpclb-v1 balancer addresses, then we need to exclude them from the list.
+    if (!newFallbackPolicyName.equals("grpclb") && !newFallbackPolicyName.equals(XDS_POLICY_NAME)) {
+      ImmutableList.Builder<EquivalentAddressGroup> backends = ImmutableList.builder();
```
voidzcy: (Personal opinion, you may not need to change.) nit: you don't need to construct an immutable list here. The spec of `ResolvedAddresses.Builder` doesn't requires so.

xds/src/test/java/io/grpc/xds/FallbackLbTest.java
```diff
@@ -0,0 +1,377 @@
...
+            .setAddresses(ImmutableList.of(eag0, eag2))
+            .setAttributes(
+                Attributes.newBuilder()
+                    .set(ATTR_LOAD_BALANCING_CONFIG, ImmutableMap.of("fallback_1_option", "yes"))
+                    .build())
+            .build());
+  }
+
+  @Test
+  public void fallback_onlyGrpclbAddresses_NoBackendAddress() throws Exception {
```
voidzcy: nit: maybe name it "incompatibleWithPureGrpclbAddresses"?

xds/src/test/java/io/grpc/xds/FallbackLbTest.java
```diff
@@ -0,0 +1,377 @@
...
+            .set(ATTR_LOAD_BALANCING_CONFIG, ImmutableMap.of("fallback_1_option", "yes")).build())
+        .build());
+
+    SubchannelPicker picker1 = mock(SubchannelPicker.class);
+    helper1.updateBalancingState(CONNECTING, picker1);
+    verify(helper).updateBalancingState(CONNECTING, picker1);
+
+    EquivalentAddressGroup eag121 = new EquivalentAddressGroup(mock(SocketAddress.class));
+    List<EquivalentAddressGroup> eags12 = ImmutableList.of(eag121);
+    String lbConfigRaw12 = "{"
```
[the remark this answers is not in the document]
dapengzhang0: That's difference level of policy change need be covered. Policy name vs Policy content/sub policy.

xds/src/test/java/io/grpc/xds/FallbackLbTest.java
```diff
@@ -0,0 +1,377 @@
...
+        .setAddresses(eags21)
+        .setAttributes(Attributes.newBuilder().set(ATTR_LOAD_BALANCING_CONFIG, lbConfig21).build())
+        .build());
+
+    verify(balancer1).shutdown();
+    assertThat(helpers2).hasSize(1);
+    assertThat(balancers2).hasSize(1);
+    Helper helper2 = helpers2.get(0);
+    LoadBalancer balancer2 = balancers2.get(0);
+    verify(balancer1, never()).handleResolvedAddresses(ResolvedAddresses.newBuilder()
```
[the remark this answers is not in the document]
dapengzhang0: The behavior here should be the same regardless there's `GracefulSwitchLoadBalancer' in place of hard switch in place.

xds/src/test/java/io/grpc/xds/FallbackLbTest.java
```diff
@@ -0,0 +1,377 @@
...
+    LoadBalancer balancer1 = balancers1.get(0);
+    verify(balancer1).handleResolvedAddresses(ResolvedAddresses.newBuilder()
+        .setAddresses(eags11)
+        .setAttributes(Attributes.newBuilder()
+            .set(ATTR_LOAD_BALANCING_CONFIG, ImmutableMap.of("fallback_1_option", "yes")).build())
+        .build());
+
+    SubchannelPicker picker1 = mock(SubchannelPicker.class);
+    helper1.updateBalancingState(CONNECTING, picker1);
+    verify(helper).updateBalancingState(CONNECTING, picker1);
```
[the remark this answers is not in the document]
dapengzhang0: This is verifying `helper1` is delegating to `helper` before policy name change, in comparison with `helper1` is no longer delegating after policy name change.

xds/src/test/java/io/grpc/xds/FallbackLbTest.java
```diff
@@ -0,0 +1,377 @@
...
+    assertThat(balancers2).hasSize(1);
+
+    verify(balancer2, never()).shutdown();
+    fallbackLb.shutdown();
+    verify(balancer2).shutdown();
+  }
+
+
+  @Test
+  public void fallback_handleBackendsEagsOnly() throws Exception {
```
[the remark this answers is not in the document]
dapengzhang0: Done.

xds/src/test/java/io/grpc/xds/FallbackLbTest.java
dapengzhang0: It's not clear what `incompatible` mean. I think the existing name is more meaningful.

xds/src/test/java/io/grpc/xds/FallbackLbTest.java
```diff
@@ -0,0 +1,377 @@
...
+        .build());
+
+    LoadBalancer balancer1 = balancers1.get(0);
+    ArgumentCaptor<Status> statusCaptor = ArgumentCaptor.forClass(Status.class);
+    verify(balancer1).handleNameResolutionError(statusCaptor.capture());
+    assertThat(statusCaptor.getValue().getCode()).isEqualTo(Code.UNAVAILABLE);
+  }
+
+  @Test
+  public void fallback_handleGrpclbAddresses() throws Exception {
```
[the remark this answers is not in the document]
dapengzhang0: Done.

xds/src/test/java/io/grpc/xds/FallbackLbTest.java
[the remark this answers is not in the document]
voidzcy: Then you don't need to verify anything more than `verify(balancer1).shutdown()` for `balancer1`. Anything more than that is excessive and is a reproduce of implementation. Tests do not aim to cover everything in implementation, but only observable checkpoints. But sometimes this can be vague, I am fine with it..

xds/src/test/java/io/grpc/xds/FallbackLbTest.java
```diff
@@ -0,0 +1,377 @@
...
+            .setAttributes(
+                Attributes.newBuilder()
+                    .set(ATTR_LOAD_BALANCING_CONFIG, ImmutableMap.of("fallback_1_option", "yes"))
+                    .build())
+            .build());
+  }
+
+  @Test
+  public void resolvingWithOnlyGrpclbAddresses_NoBackendAddress() throws Exception {
+
```
voidzcy: Delete this white line.

xds/src/test/java/io/grpc/xds/FallbackLbTest.java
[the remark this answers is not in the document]
voidzcy: This does nothing. Of course `helper1` wraps `helper`. Putting this two lines together does not verify anything other than "`helper1` is a forwarder of `helper`", which is what `ForwardingLoadBalancerHelperTest` should do.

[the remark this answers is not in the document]
dapengzhang0: It may not be a `ForwardingHelper`, even it is, it's all implementation detail. If it is simply a `ForwardingHelper`, how could the later verification
```
 helper1.updateBalancingState(CONNECTING, picker1);
 verify(helper, never()).updateBalancingState(CONNECTING, picker1);
```

pass?

xds/src/test/java/io/grpc/xds/FallbackLbTest.java
dapengzhang0: Done.

xds/src/test/java/io/grpc/xds/FallbackLbTest.java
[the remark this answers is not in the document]
voidzcy: Fine.

PR #6085, opened 2019-08-20
doc: explicitly mention that Deadline might saturate

### LINKED ISSUE #6082: Deadline silently overflows at ~292 years
Inside of `Deadline`'s constructor, there's this snippet of code:

```java
deadlineNanos = baseInstant + offset;
```

A `long` of nanoseconds overflows after ~292 years, so for all intents and purposes, it's _probably_ OK, but it still might be worth checking that an overflow didn't occur (either using a library like Guava's `LongMath.checkedAdd(long, long)` or Java8+'s `Math.addExact(long, long)`, or even just making sure the result isn't negative.

dapengzhang0:
Resolves #6082

### DIFF
--- a/context/src/main/java/io/grpc/Deadline.java
+++ b/context/src/main/java/io/grpc/Deadline.java
@@ -55,6 +55,9 @@ public final class Deadline implements Comparable<Deadline> {
    * Create a deadline that will expire at the specified offset based on the {@link #getSystemTicker
    * system ticker}.
    *
+   * <p>If the given offset is extraordinarily long, say 100 years, the actual deadline created
+   * might saturate.
+   *
    * @param duration A non-negative duration.
    * @param units The time unit for the duration.
    * @return A new deadline.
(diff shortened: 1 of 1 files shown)


context/src/main/java/io/grpc/Deadline.java
```diff
@@ -55,6 +55,9 @@ public static Ticker getSystemTicker() {
    * Create a deadline that will expire at the specified offset based on the {@link #getSystemTicker
    * system ticker}.
    *
+   * <p>If the given offset is extraordinarily long, say 100 years, the actual deadline created
```
ejona86: s/offset/duration/

context/src/main/java/io/grpc/Deadline.java
```diff
@@ -66,6 +69,9 @@ public static Deadline after(long duration, TimeUnit units) {
   /**
    * Create a deadline that will expire at the specified offset based on the given {@link Ticker}.
    *
+   * <p>If the given offset is extraordinarily long, say 100 years, the actual deadline created
```
ejona86: s/offset/duration/ here as well. The last instance is good as "offset"

context/src/main/java/io/grpc/Deadline.java
dapengzhang0: Actually the first sentence mentioned offset 
`* Create a deadline that will expire at the specified offset`

ejona86: Okay. This is fine then.

PR #6087, opened 2019-08-21
netty: Add Bazel target similar to netty-shaded

ejona86:
See census-instrumentation/opencensus-java#1963

CC @songy23 @blt-gw @creamsoup

### DIFF
--- a/netty/BUILD.bazel
+++ b/netty/BUILD.bazel
@@ -27,3 +27,11 @@ java_library(
         "@io_perfmark_perfmark_api//jar",
     ],
 )
+
+# Not actually shaded, but mirrors the shaded dependencies included in the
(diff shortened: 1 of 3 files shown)


netty/BUILD.bazel
```diff
@@ -27,3 +27,11 @@ java_library(
         "@io_perfmark_perfmark_api//jar",
     ],
 )
+
+# Not actually shaded, but mirrors the shaded dependencies included in the
+# artifact on Maven Central for usage with maven_install's override_targets.
+java_library(
```
dapengzhang0: Does this only work on Linux?

ejona86: Yes. Epoll is Linux-only, by definition. It doesn't _break_ on other platforms though.

PR #6089, opened 2019-08-21
bazel: Add helper for maven_install's override_targets

ejona86:
This can be used to prevent duplicate classes in the classpath, one via
Maven and one via Bazel-native.

See census-instrumentation/opencensus-java#1963 and #5359

CC @songy23 @blt-gw @creamsoup

### DIFF
--- a/core/BUILD.bazel
+++ b/core/BUILD.bazel
@@ -63,3 +63,17 @@ java_library(
     ],
 )
 
+# Mirrors the dependencies included in the artifact on Maven Central for usage
+# with maven_install's override_targets. Purposefully does not export any
+# symbols, as it should only be used as a dep for pre-compiled binaries on
+# Maven Central.
+java_library(
+    name = "core_maven",
+    visibility = ["//visibility:public"],
+    runtime_deps = [
+        ":inprocess",
+        ":internal",
+        ":util",
+        "//api",
+    ],
+)
--- a/repositories.bzl
+++ b/repositories.bzl
@@ -3,6 +3,40 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:jvm.bzl", "jvm_maven_import_external")
 
+# For use with maven_install's override_targets.
+# maven_install(
+#     ...
+#     override_targets = IO_GRPC_GRPC_JAVA_OVERRIDE_TARGETS,
+# )
+#
+# If you have your own overrides as well, you can use:
+#     override_targets = dict(
+#         IO_GRPC_GRPC_JAVA_OVERRIDE_TARGETS,
+#         "your.target:artifact": "@//third_party/artifact",
+#     )
+#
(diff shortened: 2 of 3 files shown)


core/BUILD.bazel
```diff
@@ -63,3 +63,17 @@ java_library(
...
 )
 
+# Mirrors the dependencies included in the artifact on Maven Central for usage
+# with maven_install's override_targets. Purposefully does not export any
+# symbols, as it should only be used as a dep for pre-compiled binaries on
+# Maven Central.
+java_library(
+    name = "core_maven",
+    visibility = ["//visibility:public"],
+    runtime_deps = [
```
dapengzhang0: Why is this runtime_deps?

repositories.bzl
```diff
@@ -3,6 +3,40 @@
...
+#         "your.target:artifact": "@//third_party/artifact",
+#     )
+#
+# To combine OVERRIDE_TARGETS from multiple libraries:
+#     override_targets = dict(
+#         IO_GRPC_GRPC_JAVA_OVERRIDE_TARGETS.items() +
+#         OTHER_OVERRIDE_TARGETS.items(),
+#         "your.target:artifact": "@//third_party/artifact",
+#     )
+IO_GRPC_GRPC_JAVA_OVERRIDE_TARGETS = {
```
dapengzhang0: How about IO_GRPC_MAVEN_INSTALL_OVERRIDE_TARGETS?

core/BUILD.bazel
ejona86: Did the documentation not explain it?
> Purposefully does not export any symbols, as it should only be used as a dep for pre-compiled binaries on Maven Central.

I didn't want Bazel users to get lazy and use this "mega" target instead of the more fine-grained targets.

repositories.bzl
ejona86: I was trying to follow the general pattern that mvn_install documents (albeit, for the artifacts field). https://github.com/bazelbuild/rules_jvm_external#exporting-and-consuming-artifacts-from-external-repositories

We are already using io_grpc_grpc_java as the workspace name (`@io_grpc_grpc_java`).

repositories.bzl
```diff
@@ -3,6 +3,40 @@
...
+    "io.grpc:grpc-alts": "@io_grpc_grpc_java//alts",
+    "io.grpc:grpc-api": "@io_grpc_grpc_java//api",
+    "io.grpc:grpc-auth": "@io_grpc_grpc_java//auth",
+    "io.grpc:grpc-context": "@io_grpc_grpc_java//context",
+    "io.grpc:grpc-core": "@io_grpc_grpc_java//core_maven",
```
f355: I believe you have a typo here, should be `@io_grpc_grpc_java//core:core_maven`.

(not sending a PR because of the CLA, sorry)

ejona86: Created https://github.com/grpc/grpc-java/pull/6224 .

Issue #6092, opened 2019-08-23
Unary call can hang forever if uncaught exeception thrown in client interceptor


ghost:
### What version of gRPC are you using?
v1.2.0, but reproduced with v1.23.0

### What did you expect to see?
If a buggy interceptor throws in onClose, gRPC logs the error, but doesn't cancel the call. This can lead to a client hanging forever. Interceptor implementations should take care not to throw, but we think gRPC could also guard against this. A deadline doesn't help, because the deadline future is cleaned up before calling the client observer's onClose method.

Reproduction: https://github.com/lihalite/grpc-error-reproduction

For instance, ClientCallImpl#closeObserver could catch runtime exceptions and cancel the call.

ejona86:
`ClientCallImpl.ClientStreamListenerImpl` does catch exceptions and notify the stub for callbacks. But it does that via `onClose()`! If `onClose()` itself fails, there is no way to notify the stub things are broken. So `onClose()` is the _one method_ that has this hole.

I don't think there is a way to solve this. gRPC should log the exception, but I don't know a way to un-hang the stub.

ejona86:
Note: when `onClose()` is called, `call.cancel()` is a noop because the call is already done. There is no call to cancel.

lihalite:
Thanks @ejona86, that's unfortunate but understandable. We could submit a documentation patch, and perhaps change `closeObserver` to catch and log errors with a message that explicitly indicates client code is misbehaving (the message the executor prints when it catches exceptions is a little vague). Would that be a good resolution to this?

PR #6107, opened 2019-08-29
doc: clarify ClientCall.Listener.onClose()

dapengzhang0:
Addressing #6092 in javadoc. Exception handling implementation will be done separately.

### DIFF
--- a/api/src/main/java/io/grpc/ClientCall.java
+++ b/api/src/main/java/io/grpc/ClientCall.java
@@ -105,6 +105,10 @@ public abstract class ClientCall<ReqT, RespT> {
    *
    * <p>Implementations are free to block for extended periods of time. Implementations are not
    * required to be thread-safe.
+   *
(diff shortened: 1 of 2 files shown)


dapengzhang0:
PTALA

api/src/main/java/io/grpc/ClientCall.java
```diff
@@ -138,6 +138,9 @@ public void onMessage(T message) {}
      * An additional block of trailer metadata may be received at the end of the call from the
      * server. An empty {@link Metadata} object is passed if no trailers are received.
      *
+     * <p>This method should not throw. Implementations should be extra-careful about exceptions
```
ejona86: Maybe something akin to: "If this method throws, there is no way to be notified of the exception. Implementations should therefore be careful of exceptions which can accidentally leak resources."

Not wild about my suggested language, but the "hang the call" is too much of a stub concept.

dapengzhang0:
Close in favor of #7635

PR #6109, opened 2019-08-30
README.md: Swap to Gradle Plugins DSL

ejona86:
The Plugins DSL has matured and is now preferred. It's also more terse,
which is nice.


ejona86:
@dapengzhang0, yeah, I saw that you had already fixed the examples. I did check whether they needed to be updated. The only usages of the old syntax appear to be our own in our build.gradle files, which are fine for now.

PR #6116, opened 2019-09-03
Update to truth 1.0

elharo:


ran-su:
This is not the only place we use truth, if we want to upgrade, we need upgrade them all. You can find older PRs for reference.

dapengzhang0:
@ran-su You are right. @elharo please see 409afe5867b49a71d02047feaf7c0a90c63834f3 for reference

PR #6126, opened 2019-09-05
core: Add Attributes.Key for authority in EquivalentAddressGroup (#4469)

### LINKED ISSUE #4469: Provide a way for NameResolvers to dynamically provide authority
Today, getServiceAuthority() must not do any I/O to determine the authority. That is, it must be computable from the target string.

It's come up before to allow the NameResolver to retrieve the authority from the name resolution system. However, in the past those systems couldn't _securely_ retrieve the authority so it would allow MITM attacks. We've now come across a resolution system that can provide the necessary security guarantee, as it uses TLS to communicate with the resolution system.

Since this is a niche use-case, I think it is fair to leave getServiceAuthority() in place. Also, Channel.authority() is implemented via getServiceAuthority(), so the current restrictions are still probably appropriate. We'll just need to add documentation to Channel.authority() that it may be partially wrong (since each backend may have a different authority). CallCredentials shouldn't have an issue with dynamic authority, although that should be verified.

I think we can use an authority Attribute on the EquivalentAddressGroup ~or PairSocketAddress. This does require us to make a decision for #4302 since it's basically the same problem.~ The design decision has been made; we can safely use EAG here.

CC @zhangkun83

enguerrand:
This enables NameResolvers to dynamically provide authority for each
Subchannel

Fixes #4469

### DIFF
--- a/api/src/main/java/io/grpc/EquivalentAddressGroup.java
+++ b/api/src/main/java/io/grpc/EquivalentAddressGroup.java
@@ -35,6 +35,12 @@ import java.util.List;
 @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1770")
 public final class EquivalentAddressGroup {
 
+  /**
+   * The authority to be used when constructing Subchannels for this EquivalentAddressGroup.
+   */
+  @Attr
+  public static final Attributes.Key<String> ATTR_AUTHORITY_OVERRIDE =
+      Attributes.Key.create("io.grpc.grpclb.authorityOverride");
   private final List<SocketAddress> addrs;
   private final Attributes attrs;
 
--- a/api/src/main/java/io/grpc/EquivalentAddressGroup.java
+++ b/api/src/main/java/io/grpc/EquivalentAddressGroup.java
@@ -39,6 +39,7 @@ public final class EquivalentAddressGroup {
    * The authority to be used when constructing Subchannels for this EquivalentAddressGroup.
    */
   @Attr
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/6138")
   public static final Attributes.Key<String> ATTR_AUTHORITY_OVERRIDE =
       Attributes.Key.create("io.grpc.grpclb.authorityOverride");
   private final List<SocketAddress> addrs;
--- a/api/src/main/java/io/grpc/EquivalentAddressGroup.java
+++ b/api/src/main/java/io/grpc/EquivalentAddressGroup.java
@@ -41,7 +41,7 @@ public final class EquivalentAddressGroup {
   @Attr
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/6138")
   public static final Attributes.Key<String> ATTR_AUTHORITY_OVERRIDE =
-      Attributes.Key.create("io.grpc.grpclb.authorityOverride");
+      Attributes.Key.create("io.grpc.EquivalentAddressGroup.authorityOverride");
   private final List<SocketAddress> addrs;
   private final Attributes attrs;
 
--- a/core/src/main/java/io/grpc/internal/InternalSubchannel.java
+++ b/core/src/main/java/io/grpc/internal/InternalSubchannel.java
@@ -231,10 +231,13 @@ final class InternalSubchannel implements InternalInstrumented<ChannelStats>, Tr
       address = proxiedAddr.getTargetAddress();
     }
 
+    Attributes currentEagAttributes = addressIndex.getCurrentEagAttributes();
+    String eagChannelAuthority = currentEagAttributes
+            .get(EquivalentAddressGroup.ATTR_AUTHORITY_OVERRIDE);
     ClientTransportFactory.ClientTransportOptions options =
         new ClientTransportFactory.ClientTransportOptions()
-          .setAuthority(authority)
-          .setEagAttributes(addressIndex.getCurrentEagAttributes())
+          .setAuthority(eagChannelAuthority != null ? eagChannelAuthority : authority)
+          .setEagAttributes(currentEagAttributes)
           .setUserAgent(userAgent)
           .setHttpConnectProxiedSocketAddress(proxiedAddr);
     TransportLogger transportLogger = new TransportLogger();
--- a/core/src/test/java/io/grpc/internal/InternalSubchannelTest.java
+++ b/core/src/test/java/io/grpc/internal/InternalSubchannelTest.java
@@ -176,6 +176,22 @@ public class InternalSubchannelTest {
         isA(TransportLogger.class));
   }
 
+  @Test public void eagAuthorityOverride_propagatesToTransport() {
+    SocketAddress addr = new SocketAddress() {};
+    String overriddenAuthority = "authority-override";
+    Attributes attr = Attributes.newBuilder()
+            .set(EquivalentAddressGroup.ATTR_AUTHORITY_OVERRIDE, overriddenAuthority).build();
+    createInternalSubchannel(new EquivalentAddressGroup(Arrays.asList(addr), attr));
+
+    // First attempt
+    assertNull(internalSubchannel.obtainActiveTransport());
+    assertEquals(CONNECTING, internalSubchannel.getState());
+    verify(mockTransportFactory).newClientTransport(
+        eq(addr),
+        eq(createClientTransportOptions().setAuthority(overriddenAuthority).setEagAttributes(attr)),
+        isA(TransportLogger.class));
+  }
+
   @Test public void singleAddressReconnect() {
     SocketAddress addr = mock(SocketAddress.class);
     createInternalSubchannel(addr);


core/src/main/java/io/grpc/internal/GrpcAttributes.java
```diff
@@ -50,6 +50,13 @@
   public static final Attributes.Key<Boolean> ATTR_LB_PROVIDED_BACKEND =
       Attributes.Key.create("io.grpc.grpclb.lbProvidedBackend");
 
+  /**
+   * The authority to be used when constructing {@link io.grpc.Channel}s to this this EquivalentAddressGroup.
+   */
+  @EquivalentAddressGroup.Attr
+  public static final Attributes.Key<String> ATTR_CHANNEL_AUTHORITY =
```
ejona86: This should be a public API, not in internal. @zhangkun83, is the only place we have for things like this `io.grpc.Grpc`?

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
```diff
@@ -257,12 +257,16 @@ private void updateBalancingState(ConnectivityState state, RoundRobinPicker pick
...
    * Converts list of {@link EquivalentAddressGroup} to {@link EquivalentAddressGroup} set and
-   * remove all attributes.
+   * remove all attributes except for the channel authority.
    */
   private static Set<EquivalentAddressGroup> stripAttrs(List<EquivalentAddressGroup> groupList) {
     Set<EquivalentAddressGroup> addrs = new HashSet<>(groupList.size());
+
     for (EquivalentAddressGroup group : groupList) {
-      addrs.add(new EquivalentAddressGroup(group.getAddresses()));
+      Attributes whiteListedAttrs = Attributes.newBuilder()
```
ejona86: We do not want to run equals() on Attributes. I added details concerning the approach to take here at https://github.com/grpc/grpc-java/issues/4469#issuecomment-528503518

core/src/main/java/io/grpc/internal/GrpcAttributes.java
sanjaypujare: Yes, the last time I checked `io.grpc.Grpc` is the only place where these Attrs are defined (and they all are prefixed `TRANSPORT_ATTR_`) . I guess this can also be considered a `TRANSPORT_ATTR_` ?

core/src/main/java/io/grpc/internal/GrpcAttributes.java
```diff
@@ -50,6 +50,13 @@
   public static final Attributes.Key<Boolean> ATTR_LB_PROVIDED_BACKEND =
       Attributes.Key.create("io.grpc.grpclb.lbProvidedBackend");
 
+  /**
+   * The authority to be used when constructing {@link io.grpc.Channel}s to this this EquivalentAddressGroup.
```
sanjaypujare: this this -> this

Also did you mean to say Subchannel instead of Channel (which makes sense given the main PR comment.

core/src/main/java/io/grpc/internal/GrpcAttributes.java
zhangkun83: I would suggest we name it as `ATTR_AUTHORITY_OVERRIDE` and put it in `EquivalentAddressGroup`. `TRANSPORT_ATTR_` is for pulling information out of transport, while this attribute is passing information to the transport, which is different. It's better to put the attributes close to where it's used. `TRANSPORT_ATTR_` is in `Grpc` because there is no other public place where we can put them, since transport API is internal.

core/src/main/java/io/grpc/internal/GrpcAttributes.java
```diff
@@ -50,6 +50,13 @@
...
+   */
+  @EquivalentAddressGroup.Attr
+  public static final Attributes.Key<String> ATTR_CHANNEL_AUTHORITY =
+      Attributes.Key.create("io.grpc.grpclb.channelAuthority");
```
zhangkun83: Please rename the key string accordingly.

core/src/main/java/io/grpc/internal/GrpcAttributes.java
sanjaypujare: Makes sense. Agreed on all points.

enguerrand: Renamed the attribute as suggested by @zhangkun83 . I am assuming that 

>  put it in `EquivalentAddressGroup`

means using the `@EquivalentAddressGroup.Attr` annotation as I already did?

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
enguerrand: @ejona86, I opted for the second approach you suggested (of keeping a  a Map<stripped EAG, original EAG>), although I did not fully understand the benefit of having the stripping method generic. 
It is private and it isn't used anywhere else. In my opinion the code would become a lot cleaner and more readable if I just changed the stripping method to actually work on EAGs and "do the right thing", i.e. ignore attributes in the comparison. I am happy to refactor it that way if that's desired.

I will also admit that I do not fully understand the implications of calling updateAddresses() on the reused channels as you suggested. You might want to take a particularly close look at how I am doing this.

core/src/main/java/io/grpc/internal/GrpcAttributes.java
enguerrand: I went for Channel as the Subchannel could not be referenced from this context. I changed this to Subchannel and removed the javadoc link.

core/src/main/java/io/grpc/internal/GrpcAttributes.java
enguerrand: @zhangkun83. Done, although I am not entirely sure about the convention here.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
```diff
@@ -158,6 +168,11 @@ public void onSubchannelState(ConnectivityStateInfo state) {
     for (Subchannel removedSubchannel : removedSubchannels) {
       shutdownSubchannel(removedSubchannel);
     }
+
+    // Update reused subchannels with possibly updated EquivalentAddressGroups
+    for (EquivalentAddressGroup addressGroup : unchangedAddrs) {
+      subchannels.get(addressGroup).updateAddresses(resolvedAddresses.getAddresses());
```
ejona86: No, no. This updates each subchannel to point to all addresses. It should be closer to:
```java
subchannels.get(addressGroup).updateAddresses(addressGroup);
```

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
ejona86: [Attributes.equals() is slow,](https://grpc.github.io/grpc-java/javadoc/io/grpc/Attributes.html#equals-java.lang.Object-) so we are avoiding that.

How are you suggesting to "do the right thing"? Ignore the attributes when comparing within the Map or changing EAG to ignore the attributes in equals/hashCode?

core/src/main/java/io/grpc/internal/GrpcAttributes.java
ejona86: No, I think he meant move the field to `io.grpc.EquivalentAddressGroup`, directly.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
```diff
@@ -87,10 +86,21 @@
...
+    Map<EquivalentAddressGroup, EquivalentAddressGroup> currentAddrs =
+            stripAttrs(subchannels.keySet());
+    Map<EquivalentAddressGroup, EquivalentAddressGroup> latestAddrs = stripAttrs(servers);
+    Set<EquivalentAddressGroup> addedAddrsStripped =
+            setsDifference(latestAddrs.keySet(), currentAddrs.keySet());
+    Set<EquivalentAddressGroup> removedAddrsStripped =
+            setsDifference(currentAddrs.keySet(), latestAddrs.keySet());
+    Set<EquivalentAddressGroup> unchangedAddrsStripped =
+            setsDifference(currentAddrs.keySet(), removedAddrsStripped);
+    Set<EquivalentAddressGroup> addedAddrs =
```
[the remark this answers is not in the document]
enguerrand: I refactored this to get rid of the stripping entirely (as I suggested earlier).
For this to be possible I had to make the "setsDifference" method ungeneric. I personally think this makes the code a lot easier to read but please let me know what you think.
I also added a unit test to validate the new diffing.

core/src/main/java/io/grpc/internal/GrpcAttributes.java
enguerrand: ok. done.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
enguerrand: Done

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
```diff
@@ -88,9 +89,10 @@ public void handleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
     List<EquivalentAddressGroup> servers = resolvedAddresses.getAddresses();
     Attributes attributes = resolvedAddresses.getAttributes();
     Set<EquivalentAddressGroup> currentAddrs = subchannels.keySet();
-    Set<EquivalentAddressGroup> latestAddrs = stripAttrs(servers);
-    Set<EquivalentAddressGroup> addedAddrs = setsDifference(latestAddrs, currentAddrs);
-    Set<EquivalentAddressGroup> removedAddrs = setsDifference(currentAddrs, latestAddrs);
+    Set<EquivalentAddressGroup> latestAddrs = new HashSet<>(servers);
```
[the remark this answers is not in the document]
enguerrand: Ok, thanks for helping out. Should I squash my commits and discard the RR-related changes?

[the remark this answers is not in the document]
ejona86: Yes, please do.

Ideally there would also be a test. I think copying InternalSubchannelTest.eagAttribute_propagatesToTransport would work well. You'd change `attr` to include ATTR_AUTHORITY_OVERRIDE and change the line with `createClientTransportOptions()` to set the authority.

api/src/main/java/io/grpc/EquivalentAddressGroup.java
```diff
@@ -35,6 +35,12 @@
 @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1770")
 public final class EquivalentAddressGroup {
 
+  /**
+   * The authority to be used when constructing Subchannels for this EquivalentAddressGroup.
+   */
+  @Attr
```
ejona86: Add an `@ExperimentalApi` annotation here. Create a new issue to stabilize the API, and include it as the value. You can look at other experimental apis to see the format.

enguerrand: Done

api/src/main/java/io/grpc/EquivalentAddressGroup.java
```diff
@@ -35,6 +35,13 @@
...
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/6138")
+  public static final Attributes.Key<String> ATTR_AUTHORITY_OVERRIDE =
+      Attributes.Key.create("io.grpc.grpclb.authorityOverride");
```
dapengzhang0: I think this is not related to `io.grpc.grpclb`. Maybe `io.grpc.EquivalentAddressGroup.authorityOverride` is a better name.

ejona86: Done. I went ahead and made this change, since it was trivial.

ejona86:
@enguerrand, thank you!

This did not make the branch cut (happened yesterday), which means it would be included in the release-after-next, in 2 months. How much of an issue would that be for you?

enguerrand:
@ejona86, glad to see the changes were merged. If would certainly have helped me to have the changes in the upcoming release but I understand that you have processes to follow...

PR #6136, opened 2019-09-09
core: Propagate EAG Attributes in RoundRobinLoadBalancer

ejona86:
This allows plumbing information to the subchannel via Attributes, like
an authority override. RR still does not support multiple EAGs that only
differ by attributes.

CC @enguerrand

### DIFF
--- a/core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
+++ b/core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
@@ -42,6 +42,7 @@ import io.grpc.internal.GrpcAttributes;
 import io.grpc.internal.ServiceConfigUtil;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -88,9 +89,8 @@ final class RoundRobinLoadBalancer extends LoadBalancer {
     List<EquivalentAddressGroup> servers = resolvedAddresses.getAddresses();
     Attributes attributes = resolvedAddresses.getAttributes();
     Set<EquivalentAddressGroup> currentAddrs = subchannels.keySet();
-    Set<EquivalentAddressGroup> latestAddrs = stripAttrs(servers);
-    Set<EquivalentAddressGroup> addedAddrs = setsDifference(latestAddrs, currentAddrs);
-    Set<EquivalentAddressGroup> removedAddrs = setsDifference(currentAddrs, latestAddrs);
+    Map<EquivalentAddressGroup, EquivalentAddressGroup> latestAddrs = stripAttrs(servers);
+    Set<EquivalentAddressGroup> removedAddrs = setsDifference(currentAddrs, latestAddrs.keySet());
 
     Map<String, ?> serviceConfig = attributes.get(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG);
     if (serviceConfig != null) {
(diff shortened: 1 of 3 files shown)


ejona86:
@zhangkun83, since you are stretched for time, we can have @dapengzhang0 review this if it would help you.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
```diff
@@ -109,8 +109,18 @@ public void handleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
...
-    // Create new subchannels for new addresses.
-    for (EquivalentAddressGroup addressGroup : addedAddrs) {
+    for (Map.Entry<EquivalentAddressGroup, EquivalentAddressGroup> latestEntry :
+        latestAddrs.entrySet()) {
+      EquivalentAddressGroup strippedAddressGroup = latestEntry.getKey();
+      EquivalentAddressGroup originalAddressGroup = latestEntry.getValue();
+      Subchannel existingSubchannel = subchannels.get(strippedAddressGroup);
+      if (existingSubchannel != null) {
+        // EAG's Attributes may have changed.
+        existingSubchannel.updateAddresses(Collections.singletonList(originalAddressGroup));
```
dapengzhang0: `updateAddresses` is for pick first, I don't think this is the right thing to do here.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
```diff
@@ -88,9 +89,8 @@ public void handleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
     List<EquivalentAddressGroup> servers = resolvedAddresses.getAddresses();
     Attributes attributes = resolvedAddresses.getAttributes();
     Set<EquivalentAddressGroup> currentAddrs = subchannels.keySet();
-    Set<EquivalentAddressGroup> latestAddrs = stripAttrs(servers);
-    Set<EquivalentAddressGroup> addedAddrs = setsDifference(latestAddrs, currentAddrs);
-    Set<EquivalentAddressGroup> removedAddrs = setsDifference(currentAddrs, latestAddrs);
+    Map<EquivalentAddressGroup, EquivalentAddressGroup> latestAddrs = stripAttrs(servers);
```
dapengzhang0: `servers` may have repetitions as well as addresses with the same stripped address but different attributes. You may consider
```java
List<EquivalentAddressGroup> strippedServers = stripAttrs(servers);
<T> Set<T> setsDifference(Set<T> a, Collection<T> b);
```

And the identity of a subchannel is always determined by `originalAddressGroup`
```java
subchannels.put(originalAddressGroup, subchannel);
```

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
```diff
@@ -141,7 +151,7 @@ public void onSubchannelState(ConnectivityStateInfo state) {
       if (stickyRef != null) {
         stickyRef.value = subchannel;
       }
-      subchannels.put(addressGroup, subchannel);
+      subchannels.put(strippedAddressGroup, subchannel);
```
dapengzhang0: I think we should use `originalAddressGroup` for the identity of a subchannel.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
dapengzhang0: Actually I think we shouldn't do `stripAttrs` at all.

ejona86:
For a little bit more background on the approach taken here, and its downfalls, https://github.com/grpc/grpc-java/issues/4469#issuecomment-528503518 talks some about the options.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
ejona86: Sorry, I you hadn't seen some of the other details. You are right that nothing changes immediately. However, the next connection that the subchannel makes will use the updated Attributes, which is what we are wanting right now.

core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java
ejona86: `stripAttrs` is to avoid calling Attributes.equals() via EAG.equals(). That's part of the reason the code is a bit wonky.

core/src/test/java/io/grpc/util/RoundRobinLoadBalancerTest.java
```diff
@@ -195,25 +195,22 @@ public void pickAfterResolvedUpdatedHosts() throws Exception {
     Subchannel oldSubchannel = mock(Subchannel.class);
     Subchannel newSubchannel = mock(Subchannel.class);
 
+    Attributes.Key<String> key = Attributes.Key.create("check-that-it-is-propagated");
```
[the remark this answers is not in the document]
ejona86: I really like that idea, but it's not quite that easy at the moment. Equals() is used by the test, both in `subchannels` and the `verify()`. Changing subchannels would need to use a wrapper object, as the List prevents using an IdentityHashMap.

I may do it as a follow-up.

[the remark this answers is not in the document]
dapengzhang0: I tested with the code. `equals` is only used at `Attribute` level, not `AttrType` level.

PR #6154, opened 2019-09-16
Create a new class XdsAttributes in xds to house XDS related attributes and define ATTR_SDS_CONFIG there

sanjaypujare:
Need to define a new attribute so we can plumb SdsSecretConfig from the xDS LB to SSL code so the value can be passed on to the SDS module.

### DIFF
--- a/xds/src/main/java/io/grpc/xds/XdsAttributes.java
+++ b/xds/src/main/java/io/grpc/xds/XdsAttributes.java
@@ -26,7 +26,6 @@ import io.grpc.Grpc;
 final class XdsAttributes {
   /**
    * Attribute key for SdsSecretConfig of a subchannel.
-   * We use String as the value type because SdsSecretConfig is not visible.
    *
    */
   @Grpc.TransportAttr
--- a/xds/src/main/java/io/grpc/xds/XdsAttributes.java
+++ b/xds/src/main/java/io/grpc/xds/XdsAttributes.java
@@ -20,12 +20,10 @@ import io.envoyproxy.envoy.api.v2.auth.SdsSecretConfig;
 import io.grpc.Attributes;
 import io.grpc.Grpc;
 
-import java.util.Map;
-
 /**
  * Special attributes that are only useful to gRPC in the XDS context
  */
-public final class XdsAttributes {
+final class XdsAttributes {
   /**
    * Attribute key for SdsSecretConfig of a subchannel.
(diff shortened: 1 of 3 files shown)


xds/src/main/java/io/grpc/xds/XdsAttributes.java
```diff
@@ -0,0 +1,39 @@
...
+import io.envoyproxy.envoy.api.v2.auth.SdsSecretConfig;
+import io.grpc.Attributes;
+import io.grpc.Grpc;
+
+import java.util.Map;
+
+/**
+ * Special attributes that are only useful to gRPC in the XDS context
+ */
+public final class XdsAttributes {
```
dapengzhang0: Is it going to be used inside `io.grpc.xds` package only? If we can make it package private, otherwise we should annotate `@ExperimentalApi` and add `@since` in javadoc.

xds/src/main/java/io/grpc/xds/XdsAttributes.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.xds;
+
+import io.envoyproxy.envoy.api.v2.auth.SdsSecretConfig;
+import io.grpc.Attributes;
+import io.grpc.Grpc;
+
```
dapengzhang0: We don't add empty line here.

xds/src/main/java/io/grpc/xds/XdsAttributes.java
sanjaypujare: Typically the XDS LB will set this on a subchannel, and that code most probably will go in this package. I'll make it package private (unless the preference is for `@ExperimentalApi`)

xds/src/main/java/io/grpc/xds/XdsAttributes.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Special attributes that are only useful to gRPC in the XDS context
+ */
+final class XdsAttributes {
+  /**
+   * Attribute key for SdsSecretConfig of a subchannel.
+   * We use String as the value type because SdsSecretConfig is not visible.
```
dapengzhang0: Remove this line?

sanjaypujare: Yes :-) Done

PR #6156, opened 2019-09-16
examples: do not call any method that could potentially throw in onClose

dapengzhang0:
Fix our bad example in the sense of issue #6092

### DIFF
--- a/examples/src/main/java/io/grpc/examples/errorhandling/ErrorHandlingClient.java
+++ b/examples/src/main/java/io/grpc/examples/errorhandling/ErrorHandlingClient.java
@@ -52,6 +52,7 @@ public class ErrorHandlingClient {
   }
 
   private ManagedChannel channel;
+  private Status status;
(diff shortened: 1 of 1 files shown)


examples/src/main/java/io/grpc/examples/errorhandling/ErrorHandlingClient.java
```diff
@@ -136,6 +134,9 @@ public void onFailure(Throwable t) {
     if (!Uninterruptibles.awaitUninterruptibly(latch, 1, TimeUnit.SECONDS)) {
       throw new RuntimeException("timeout!");
     }
+    Verify.verify(status.getCode() == Status.Code.INTERNAL);
```
ejona86: We don't want to encourage this approach, since that requires users watching on the latch an having another thread sitting around. We shouldn't depend on the latch. We could log instead, or something like that.

PR #6158, opened 2019-09-18
all: stabilize ManagedChannelBuilder.usePlaintext()

### LINKED ISSUE #1772: Tracking Issue for Plaintext being Experimental.

dapengzhang0:
Resolves #1772

### DIFF
--- a/api/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/api/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -157,39 +157,15 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
    * <p>Should only be used for testing or for APIs where the use of such API or the data
    * exchanged is not sensitive.
    *
-   * @param skipNegotiation @{code true} if there is a priori knowledge that the endpoint supports
-   *                        plaintext, {@code false} if plaintext use must be negotiated.
-   * @deprecated Use {@link #usePlaintext()} instead.
-   *
-   * @throws UnsupportedOperationException if plaintext mode is not supported.
-   * @return this
-   * @since 1.0.0
-   */
-  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1772")
-  @Deprecated
-  public T usePlaintext(boolean skipNegotiation) {
-    throw new UnsupportedOperationException();
-  }
-
-  /**
-   * Use of a plaintext connection to the server. By default a secure connection mechanism
-   * such as TLS will be used.
-   *
-   * <p>Should only be used for testing or for APIs where the use of such API or the data
-   * exchanged is not sensitive.
-   *
    * <p>This assumes prior knowledge that the target of this channel is using plaintext.  It will
    * not perform HTTP/1.1 upgrades.
    *
-   *
-   * @throws UnsupportedOperationException if plaintext mode is not supported.
    * @return this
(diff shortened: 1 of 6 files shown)


api/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -150,27 +150,6 @@
...
-   *                        plaintext, {@code false} if plaintext use must be negotiated.
-   * @deprecated Use {@link #usePlaintext()} instead.
-   *
-   * @throws UnsupportedOperationException if plaintext mode is not supported.
-   * @return this
-   * @since 1.0.0
-   */
-  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1772")
-  @Deprecated
-  public T usePlaintext(boolean skipNegotiation) {
```
ejona86: There's plenty of existing users. We should leave this in-place for a good while (at least a year).

netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java
```diff
@@ -279,23 +279,6 @@ public NettyChannelBuilder maxInboundMetadataSize(int bytes) {
...
-   *
-   * @deprecated use {@link #usePlaintext()} instead.
-   */
-  @Override
-  @Deprecated
-  public NettyChannelBuilder usePlaintext(boolean skipNegotiation) {
-    if (skipNegotiation) {
-      negotiationType(NegotiationType.PLAINTEXT);
-    } else {
-      negotiationType(NegotiationType.PLAINTEXT_UPGRADE);
```
ejona86: I'd _really_ like to kill NegotiationType. We may want to introduce a new method for this before deleting it.

api/src/main/java/io/grpc/ManagedChannelBuilder.java
dapengzhang0: We have already deprecated it for one and half year, still need a year?

dapengzhang0: Sure. Do you mind if I also log a WARNING in the method?

ejona86: Meh. They would already get a warning during compilation. I'd rather us not think about it much and just leave it setting there for a while.

dapengzhang0: Done.

api/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -165,7 +165,6 @@
    * @return this
    * @since 1.0.0
    */
-  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1772")
```
ejona86: Leave the annotation. That way it is clear it can be deleted, vs a deprecated method that was stable so we can't delete it.

dapengzhang0: Done.

api/src/main/java/io/grpc/ManagedChannelBuilder.java
ejona86: Sorry, I misunderstood what you said about a year and a half. Yes, that is plenty. We can remove this. Although we've also found some users that need to be migrated internally. We'll remove it once they are migrated.

PR #6165, opened 2019-09-19
core: Catch uncaught exceptions in ServerImpl

### LINKED ISSUE #6162: Uncaught exceptions in server

1.23.0

Error response after uncaught exception

Refer to the following file/line number:
https://github.com/grpc/grpc-java/blob/ba0fd84d79d66f46b4492043c62ecb0fcf85ead2/core/src/main/java/io/grpc/internal/ServerImpl.java#L558

You can see that `Exception` types that are not subclasses of `Error` or `RuntimeException` (known as "checked exceptions") are not caught here.

Uncaught exceptions result in no response being sent to the client (but the connection remains open) and thus, from the clients perspective, the RPC call hangs indefinitely.

All exceptions should be caught here, not just runtime exceptions and errors.

"Checked" exceptions can be thrown from code that does not declare it in the signature by several means. One example is here:
https://stackoverflow.com/questions/15496/hidden-features-of-java/2131355#2131355

Another example is from any kotlin code that calls a java function that throws a checked exception.

NOTE: it looks like the same problem exists in multiple places in this file. The line number I linked is the specific one that I hit in the wild.

AnarSultanov:
~Fixes #6162~ A cleanup regarding #6162


dapengzhang0:
Ping @ejona86

ejona86:
The discussion for whether we are going to do this is still ongoing in #6162

ejona86:
So I took another look at this. I'm really surprised that Java 8 with source=7 has the type inferencing to notice that "no checked exceptions can be thrown" within the block, so even though we `throw Throwable` it doesn't treat it as Throwable (`throw t` and `throw (Throwable) t` behave differently). I knew that was a thing with later Java versions, but I got mixed up and thought it wasn't available to us. I guess the last time I looked at it we were supporting source=6?

Since all the tests pass, this looks like a general cleanup, and so would be good to merge. #6162 is still about whether we'll go out of our way to handle checked-but-not-checked exceptions, but any place Java isn't a PITA when re-throwing and we can have one code path, that would be good.

PR #6173, opened 2019-09-21
SecretManager implementation for file secrets

sanjaypujare:
Implementation of the SecretManager that will be used eventually for SDS implementation i.e. to fetch certs from the control plane

### DIFF
--- a/xds/src/main/java/io/grpc/xds/sds/SecretManager.java
+++ b/xds/src/main/java/io/grpc/xds/sds/SecretManager.java
@@ -40,7 +40,8 @@ public final class SecretManager {
    */
   public SecretProvider<TlsCertificateStore> findOrCreateTlsCertificateProvider(
       ConfigSource configSource, String name) {
-    checkNotNull(configSource, "configSource was null.");
+    checkNotNull(configSource, "configSource");
+    checkNotNull(name, "name");
     // for now we support only path/volume based secret provider
     if (configSource.getConfigSourceSpecifierCase()
         != ConfigSource.ConfigSourceSpecifierCase.PATH) {
--- a/xds/src/main/java/io/grpc/xds/sds/SecretManager.java
+++ b/xds/src/main/java/io/grpc/xds/sds/SecretManager.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2019 The gRPC Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.xds.sds;
+
+import io.envoyproxy.envoy.api.v2.core.ConfigSource;
+import io.grpc.Internal;
+
+/**
+ * SecretManager to manage secrets. This is used by gRPC-xds to access secrets
+ * and not part of the public API of gRPC
+ */
+@Internal
+public final class SecretManager {
+
+  /**
+   * Finds an existing SecretProvider or creates it if it doesn't exist.
+   *
+   * @param configSource   from the SdsSecretConfig
+   * @param name      from the SdsSecretConfig
+   * @return    the SecretProvider
+   */
+  public SecretProvider<TlsCertificateStore> findOrCreateTlsCertificateProvider(
+      ConfigSource configSource, String name) {
+
+    // for now we support only path/volume based secret provider
+    if (configSource.getConfigSourceSpecifierCase()
+        != ConfigSource.ConfigSourceSpecifierCase.PATH) {
+      throw new UnsupportedOperationException("Only file based secret supported");
+    }
+    return secretVolumeCertificateProviders.findOrCreate(configSource, name);
+  }
+
+  private TlsCertificateSecretProviderMap secretVolumeCertificateProviders =
+      new TlsCertificateSecretProviderMap();
+}
--- a/xds/src/main/java/io/grpc/xds/sds/SecretManager.java
+++ b/xds/src/main/java/io/grpc/xds/sds/SecretManager.java
@@ -25,9 +25,8 @@ import io.grpc.Internal;
  * SecretManager to manage secrets. This is used by gRPC-xds to access secrets
  * and not part of the public API of gRPC
  */
-//  * TODO(sanjaypujare): Lifecycle management for this manager and the {@link SecretProvider}s
-//  * will be based on  {@link io.grpc.internal.SharedResourcePool} or
-//  * {@link io.grpc.internal.SharedResourceHolder}
+//  TODO(sanjaypujare): Lifecycle management for this manager and the {@link SecretProvider}s
+//  to be implemented using an object-pool mechanism
 @Internal
 public final class SecretManager {
 
--- a/xds/src/main/java/io/grpc/xds/sds/SecretManager.java
+++ b/xds/src/main/java/io/grpc/xds/sds/SecretManager.java
@@ -22,6 +22,10 @@ import io.grpc.Internal;
 /**
  * SecretManager to manage secrets. This is used by gRPC-xds to access secrets
  * and not part of the public API of gRPC
+ * TODO: Lifecycle management for this manager and the {@link SecretProvider}s
+ * will be based on  {@link io.grpc.internal.SharedResourcePool} or
+ * {@link io.grpc.internal.SharedResourceHolder}
+ *
  */
 @Internal
 public final class SecretManager {
@@ -44,6 +48,6 @@ public final class SecretManager {
     return secretVolumeCertificateProviders.findOrCreate(configSource, name);
   }
 
-  private TlsCertificateSecretProviderMap secretVolumeCertificateProviders =
+  private final TlsCertificateSecretProviderMap secretVolumeCertificateProviders =
       new TlsCertificateSecretProviderMap();
 }
--- a/xds/src/main/java/io/grpc/xds/sds/SecretManager.java
+++ b/xds/src/main/java/io/grpc/xds/sds/SecretManager.java
@@ -16,17 +16,18 @@
 
 package io.grpc.xds.sds;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import io.envoyproxy.envoy.api.v2.core.ConfigSource;
 import io.grpc.Internal;
 
 /**
  * SecretManager to manage secrets. This is used by gRPC-xds to access secrets
  * and not part of the public API of gRPC
- * TODO: Lifecycle management for this manager and the {@link SecretProvider}s
- * will be based on  {@link io.grpc.internal.SharedResourcePool} or
- * {@link io.grpc.internal.SharedResourceHolder}
- *
  */
+//  * TODO(sanjaypujare): Lifecycle management for this manager and the {@link SecretProvider}s
+//  * will be based on  {@link io.grpc.internal.SharedResourcePool} or
+//  * {@link io.grpc.internal.SharedResourceHolder}
 @Internal
 public final class SecretManager {
 
@@ -39,7 +40,7 @@ public final class SecretManager {
    */
   public SecretProvider<TlsCertificateStore> findOrCreateTlsCertificateProvider(
       ConfigSource configSource, String name) {
-
+    checkNotNull(configSource, "configSource was null.");
     // for now we support only path/volume based secret provider
     if (configSource.getConfigSourceSpecifierCase()
         != ConfigSource.ConfigSourceSpecifierCase.PATH) {
--- a/xds/src/main/java/io/grpc/xds/sds/SecretProvider.java
+++ b/xds/src/main/java/io/grpc/xds/sds/SecretProvider.java
@@ -17,10 +17,7 @@
 package io.grpc.xds.sds;
 
 import io.grpc.Internal;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 
 /**
  * A SecretProvider is a "container" or provider of a secret. This is used by gRPC-xds to access
@@ -32,26 +29,13 @@ import java.util.concurrent.TimeoutException;
 @Internal
 public interface SecretProvider<T> {
 
-  /**
-   * Gets the current secret (waits indefinitely if necessary).
-   */
-  T get() throws InterruptedException, ExecutionException;
+  interface Callback<T> {
+    void updateSecret(T secret);
+  }
 
   /**
-   * Gets the secret (and waits if necessary for the given time and then returns the result, if
-   * available.
+   * Registers a callback on the given executor. The callback will run when secret becomes
+   * available or immediately if the result is already available.
    */
-  T get(long timeout, TimeUnit unit)
-      throws InterruptedException, ExecutionException, TimeoutException;
-
-  /**
-   * Returns {@code true} if secret is available.
-   */
-  boolean isAvailable();
-
-  /**
-   * Registers a listener to be on the given executor. The listener will run when the result is
-   * {@linkplain #isAvailable()} or immediately, if the result is already available.
-   */
-  void addListener(Runnable listener, Executor executor);
+  void addCallback(Callback<T> callback, Executor executor);
 }
--- a/xds/src/main/java/io/grpc/xds/sds/SecretProvider.java
+++ b/xds/src/main/java/io/grpc/xds/sds/SecretProvider.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2019 The gRPC Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.xds.sds;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import io.grpc.Internal;
+
+/**
+ * A SecretProvider is a "container" or provider of a secret.
+ * This is used by gRPC-xds to access secrets and not part of the
+ * public API of gRPC
+ */
+@Internal
+public interface SecretProvider<T> extends ListenableFuture<T> {
+}
--- a/xds/src/main/java/io/grpc/xds/sds/SecretProvider.java
+++ b/xds/src/main/java/io/grpc/xds/sds/SecretProvider.java
@@ -23,6 +23,7 @@ import io.grpc.Internal;
  * A SecretProvider is a "container" or provider of a secret.
  * This is used by gRPC-xds to access secrets and not part of the
  * public API of gRPC
+ * See {@link SecretManager} for a note on lifecycle management
  */
 @Internal
 public interface SecretProvider<T> extends ListenableFuture<T> {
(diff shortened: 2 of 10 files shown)


creamsoup:
is this PR ready for review? can you fix style? (see go/java-practice go/java-sg)

sanjaypujare:
@creamsoup I have fixed the style and all the checks have passed.

dapengzhang0:
@sanjaypujare Do all these APIs need be public? Can they be package private?

sanjaypujare:
@dapengzhang0 good point. Not all of them need to be public. Only the ones from SecretManager are used externally (and possibly SecretProvider). Let me take a look and update accordingly

xds/src/main/java/io/grpc/xds/sds/SecretManager.java
```diff
@@ -0,0 +1,39 @@
...
+
+package io.grpc.xds.sds;
+
+import io.envoyproxy.envoy.api.v2.core.ConfigSource;
+
+/**
+ * SecretManager to manage secrets.
+ *
+ */
+public class SecretManager {
```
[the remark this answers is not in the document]
sanjaypujare: Made final and also marked the class as `@Internal` since this will be used only by the XDS implementation and not meant to be consumed by others

xds/src/main/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProvider.java
```diff
@@ -0,0 +1,109 @@
...
+    } catch (FileNotFoundException e) {
+      throw new ExecutionException(e);
+    }
+  }
+
+  @Override
+  public TlsCertificateStore get(long timeout, TimeUnit unit)
+      throws InterruptedException, ExecutionException, TimeoutException {
+    checkNotNull(unit);
+    return get();
```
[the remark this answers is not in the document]
sanjaypujare: Added comment. Because this is not an async operation so no need to wait for any timeout

xds/src/main/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProvider.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @Override
+  public boolean isDone() {
+    return true;
+  }
+
+  /**
+   * Gets the current contents of the private key and cert file Assume the key has .pem extension
+   * and cert has .crt extension (needs to match how Citadel mounts secrets)
```
creamsoup: nit: .pem and .crt needs to be wrapped by literal.

xds/src/test/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProviderTest.java
```diff
@@ -0,0 +1,115 @@
...
+  @Test
+  public void readMissingFile() throws IOException, ExecutionException, InterruptedException {
+    createATestCertFile(temporaryFolder, "mycert.pem", "pemContents");
+
+    // no crt file
+    File filePath = new File(temporaryFolder.getRoot(), "mycert");
+    TlsCertificateSecretVolumeSecretProvider provider =
+        new TlsCertificateSecretVolumeSecretProvider(filePath.getPath(), "test");
+
+    thrown.expect(ExecutionException.class);
```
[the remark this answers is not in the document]
sanjaypujare: Added the cause check

xds/src/test/java/io/grpc/xds/sds/TlsCertificateSecretProviderMapTest.java
```diff
@@ -0,0 +1,88 @@
...
+  public final ExpectedException thrown = ExpectedException.none();
+
+  /**
+   * Utility function for creation of test files in a temp folder.
+   *
+   * @param temporaryFolder   temporary folder to use
+   * @return a config source representing the file based secret
+   * @throws IOException represents IO exception
+   */
+  public static ConfigSource createFileAndConfigSource(TemporaryFolder temporaryFolder)
```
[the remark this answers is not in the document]
sanjaypujare: Used elsewhere. making package private

xds/src/main/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProvider.java
sanjaypujare: done

xds/src/test/java/io/grpc/xds/sds/SecretManagerTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Test
+  @SuppressWarnings("unused")
+  public void nonFilePathConfigSource() {
+    ConfigSource configSource =
+        ConfigSource.newBuilder()
+            .setApiConfigSource(ApiConfigSource.newBuilder().build())
+            .build();
+
+    thrown.expect(UnsupportedOperationException.class);
```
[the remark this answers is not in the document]
sanjaypujare: I saw thrown.expect used in BootstrapperTest for example. So that's not to be used?

xds/src/test/java/io/grpc/xds/sds/TlsCertificateSecretProviderMapTest.java
```diff
@@ -0,0 +1,88 @@
...
+ * Unit tests for {@link TlsCertificateSecretProviderMap}.
+ */
+@RunWith(JUnit4.class)
+public class TlsCertificateSecretProviderMapTest {
+
+  @Rule
+  public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @Rule
+  public final ExpectedException thrown = ExpectedException.none();
```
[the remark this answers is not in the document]
sanjaypujare: So you are saying ExpectedException is not to be used at all?

xds/src/main/java/io/grpc/xds/sds/TlsCertificateStore.java
```diff
@@ -0,0 +1,89 @@
...
+    this.privateKeyStream = privateKeySteam;
+    this.certChainStream = certChainStream;
+  }
+
+  /**
+   * getter for private key stream.
+   *
+   * @return inputStream representing private key
+   */
+  public InputStream getPrivateKeyStream() {
```
dapengzhang0: When is the input stream supposed to be closed? Who manages its life cycle?

xds/src/main/java/io/grpc/xds/sds/SecretManager.java
```diff
@@ -0,0 +1,49 @@
...
+
+    // for now we support only path/volume based secret provider
+    if (configSource.getConfigSourceSpecifierCase()
+        != ConfigSource.ConfigSourceSpecifierCase.PATH) {
+      throw new UnsupportedOperationException("Only file based secret supported");
+    }
+    return secretVolumeCertificateProviders.findOrCreate(configSource, name);
+  }
+
+  private TlsCertificateSecretProviderMap secretVolumeCertificateProviders =
```
dapengzhang0: Can it be final?

xds/src/main/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProvider.java
```diff
@@ -0,0 +1,119 @@
...
+   */
+  @Override
+  public TlsCertificateStore get(long timeout, TimeUnit unit)
+      throws InterruptedException, ExecutionException, TimeoutException {
+    checkNotNull(unit);
+    return get();
+  }
+
+  private InputStream getFileInputStream(String filePath) throws FileNotFoundException {
+    return new FileInputStream(filePath);
```
dapengzhang0: I think this can be inlined.

xds/src/main/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProvider.java
```diff
@@ -0,0 +1,119 @@
...
+  public boolean isDone() {
+    return true;
+  }
+
+  /**
+   * Gets the current contents of the private key and cert file Assume the key has .pem extension
+   * and cert has .crt extension (needs to match how Citadel mounts secrets)
+   */
+  @Override
+  public TlsCertificateStore get() throws InterruptedException, ExecutionException {
```
dapengzhang0: If you call `get()` twice, it will return different instances, that's not Java Future class should behave.

xds/src/main/java/io/grpc/xds/sds/TlsCertificateStore.java
```diff
@@ -0,0 +1,89 @@
...
+ * TlsCertificate's PrivateKey and Certificate are extracted into InputStream's.
+ * This is used by the gRPC SSLContext/Protocol Negotiator and is internal
+ */
+@Internal
+public final class TlsCertificateStore {
+
+  private final InputStream privateKeyStream;
+  private final InputStream certChainStream;
+
+  private InputStream getInputStreamFromDataSource(DataSource dataSource) {
```
dapengzhang0: Can this be static?

xds/src/main/java/io/grpc/xds/sds/TlsCertificateStore.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+    return dataSourceByteString.newInput();
+  }
+
+  /**
+   * Creates the Store out of the TlsCertificate object of xDS.
+   *
+   * @param tlsCertificate  TlsCertificate Object of xDS
+   */
+  public TlsCertificateStore(TlsCertificate tlsCertificate) {
```
dapengzhang0: Is this constructor only used for testing?

sanjaypujare: It will be used in future (another PR) with actual SDS responses returning TlsCertificate objects. Do you recommend I remove it now if not used?

xds/src/main/java/io/grpc/xds/sds/TlsCertificateStore.java
sanjaypujare: Lifecyle management including refCounting and removal is coming in another PR

xds/src/main/java/io/grpc/xds/sds/TlsCertificateStore.java
dapengzhang0: I would keep it although I prefer to reuse the other constructor by
```
this(
    getInputStreamFromDataSource(checkNotNull(tlsCertificate).getPrivateKey()),
    getInputStreamFromDataSource(tlsCertificate.getCertificateChain());
```

xds/src/main/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProvider.java
sanjaypujare: Looking up the Javadoc for [`Future.get()`](https://docs.oracle.com/javase/6/docs/api/java/util/concurrent/Future.html#get()) it doesn't promise idempotency there. I did see one references about that assumption [here](https://guava.dev/releases/6.0/api/docs/com/google/common/util/concurrent/Futures.html#compose(java.util.concurrent.Future,%20com.google.common.base.Function)) .

I do understand the concern though. The `get()` in this case is supposed to return the current certificate content as loaded from the store so that functionality needs to be there. I can transfer the logic to TlsCertificateStore and make sure the same instance is returned by `get()` if that's acceptable. Or have subclasses for TlsCertificateStore

xds/src/main/java/io/grpc/xds/sds/TlsCertificateStore.java
dapengzhang0: Sure. If you know how the lifecycle management will be can you add some javadoc?

xds/src/main/java/io/grpc/xds/sds/TlsCertificateStore.java
sanjaypujare: I know Java evaluation order implies checkNotNull will happen first although the general recommendation is to not ["crucially rely on it"](https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.7). I will incorporate this suggestion though

xds/src/main/java/io/grpc/xds/sds/TlsCertificateStore.java
sanjaypujare: Done here and main note in SecretManager

xds/src/main/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProvider.java
```diff
@@ -0,0 +1,115 @@
...
+  private static final Logger logger = Logger
+      .getLogger(TlsCertificateSecretVolumeSecretProvider.class.getName());
+  public static final String PEM = ".pem";
+  public static final String CRT = ".crt";
+
+  private String path;
+
+  // for now mark it unused
+  @SuppressWarnings("unused")
+  private String name;
```
dapengzhang0: These fields are final or can be changed?

sanjaypujare: good point. Made final

xds/src/main/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProvider.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+
+  /**
+   * The file based secret provider does not need to wait (reads the current files as per
+   * the contract) so we ignore the timeout.
+   *
+   * @param timeout the maximum time to wait
+   * @param unit the time unit of the timeout argument
+   * @return the computed result
```
creamsoup: fyi. you don't have to have all the params and return if it is obvious (if the description is just repeating the param name).

xds/src/main/java/io/grpc/xds/sds/TlsCertificateStore.java
```diff
@@ -0,0 +1,92 @@
...
+
+  private static InputStream getInputStreamFromDataSource(DataSource dataSource) {
+    Preconditions.checkNotNull(dataSource);
+    ByteString dataSourceByteString = null;
+    if (dataSource.getSpecifierCase() == DataSource.SpecifierCase.INLINE_BYTES) {
+      dataSourceByteString = dataSource.getInlineBytes();
+    } else if (dataSource.getSpecifierCase() == DataSource.SpecifierCase.INLINE_STRING) {
+      dataSourceByteString = dataSource.getInlineStringBytes();
+    } else {
+      throw new IllegalArgumentException(
```
creamsoup: nit: UnsupportedOperationException?

xds/src/main/java/io/grpc/xds/sds/TlsCertificateStore.java
creamsoup: can you explain briefly about how you will handle in other PR? to me this is one of the most concerning part of this PR which may ends up changing some structure depends on how you are planning to handle the lifecycle / ownership of the stateful resource.

sanjaypujare:
> I don't see any non internal public apis, is this intended?

Yes. They are meant to be consumed internally (at least for now)

xds/src/main/java/io/grpc/xds/sds/TlsCertificateStore.java
sanjaypujare: hmmm... debatable. Illegal argument also makes sense since `dataSource.getSpecifierCase()` is checked and based on that this is thrown. But I will change it (unless you are convinced by my statement!)

xds/src/main/java/io/grpc/xds/sds/TlsCertificateStore.java
sanjaypujare: Okay, based on offline discussion - the lifecycle management of the object is what I referred to earlier.  The ownership/state management of the stream is the concern here which I will address now.

xds/src/main/java/io/grpc/xds/sds/TlsCertificateStore.java
creamsoup: i am fine either way. i just noticed that the message mentions unsupported.

xds/src/main/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProvider.java
sanjaypujare: SecretProvider.get is no longer that of Future.get. The ListenableFuture logic now is now inside individual SecretProvider. PTAL

xds/src/main/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProvider.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private final String path;
+
+  // for now mark it unused
+  @SuppressWarnings("unused")
+  private final String name;
+
+  TlsCertificateSecretVolumeSecretProvider(String path, String name) {
+    this.path = path;
+    this.name = name;
```
dapengzhang0: What about init TlsCertificateStore and ExecutionException here?
```java
    TlsCertificateStore tlsCertificateStore = null;
    ExecutionException executionException = null;
    try {
      tlsCertificateStore = new TlsCertificateStore(
          new FileInputStream(path + PEM),
          new FileInputStream(path + CRT));
    } catch (FileNotFoundException e) {
      executionException = new ExecutionException(e);
    }
    this.tlsCertificateStore = tlsCertificateStore;
    this.executionException = executionException;
```

sanjaypujare: No. The point exactly is to read the contents at the time `get()` is called because the certificate(s) on the file system changes (is dynamic). The `get()` is no longer constrained by the semantics of `Future<>.get()`.

xds/src/main/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProvider.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  @Override
+  public void addListener(Runnable listener, Executor executor) {
```
creamsoup: i don't understand this method. is the listener supposed be called when the future (or `get()`) is finished? i know this class is more like immediate future. but how the user use this listener?

let's use the generic async case since this specific one is not generic case.

I imagine when the `get()` has result / failed, it will trigger the listeners (Runnables). Since it is vanilla `Runnable`, the listener has no access to the value of the result, so the runnable must call the `get()` again to access the result? (will this trigger the new refresh?) or the listener doesn't need to use the result?

are those listeners will be triggered everytime the secret is re-evaluated? how do we know when to refresh vs use cache?

let say i only need the first update, how to opt out this notification?

xds/src/main/java/io/grpc/xds/sds/SecretProviderMap.java
```diff
@@ -0,0 +1,63 @@
...
+        provider = create(configSource, name);
+        providers.put(mapKey, provider);
+      }
+    }
+    return provider;
+  }
+
+  abstract SecretProvider<T> create(ConfigSource configSource, String name);
+
+  protected final Map<String, SecretProvider<T>> providers;
```
creamsoup: can you replace to abstract method like `getProviderMap`? Some implementations may want to use other types of map such as `ConcurrentHashMap` or `java.security.Provider`. Using abstract method doesn't restrict which data type to use internally.

xds/src/main/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProvider.java
creamsoup: i think the method name `get()` has strong connection to `Future`, how about using `getSnapshot()` or something similar name to avoid confusion. `getSnapshot` implicitly tells that the result might be different based on time.

xds/src/main/java/io/grpc/xds/sds/SecretProviderMap.java
sanjaypujare: This can also be done later but I can do it now - sounds like a good idea.

xds/src/main/java/io/grpc/xds/sds/SecretProvider.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package io.grpc.xds.sds;
+
+import io.grpc.Internal;
+import java.util.concurrent.Executor;
+
+/**
+ * A SecretProvider is a "container" or provider of a secret. This is used by gRPC-xds to access
```
creamsoup: this doc is no longer true.

sanjaypujare:
Thanks @creamsoup . I will resolve the pending comments in the next PR (coming soon...:-))

xds/src/main/java/io/grpc/xds/sds/TlsCertificateSecretVolumeSecretProvider.java
sanjaypujare: based on offline discussion of the design, I am hoping this is resolved

PR #6177, opened 2019-09-23
xds: usd PGV dependency from maven artifact

dapengzhang0:


xds/build.gradle
```diff
@@ -22,6 +22,9 @@ dependencies {
             project(':grpc-core'),
             project(':grpc-services'),
             project(':grpc-auth')
+    compile (libraries.pgv) {
+        exclude group: 'com.google.protobuf'
```
ejona86: Add a comment saying why it is excluded.

PR #6187, opened 2019-09-24
xds: add server name in EDS request resource_names field

dapengzhang0:


dapengzhang0:
@vishalpowar, I'm moving the GrpcHostname from EDS metadata to the resource_names field, is this the right thing to do for the moment?

vishalpowar:
Yes, that is the right thing to do.

PR #6192, opened 2019-09-25
netty,okhttp,cronet: add option to use get/put when methods are safe/idempotent

carl-mastrangelo:
This change adds two booleans to AbstractManagedChannelBuilderImpl to
allow transports to use get and put.   These are currently defaulted to
on, but unset on the method descriptors.   This change is 1/2 that will
allow the safe / idempotent bits to be set on generated proto code.
Part 2/2 will actually enable it.

The use case for this is for interceptors that implement caching logic.
They need to be able to access the safe/idempotent bits on the MD in
order to decide to how to handle the request, even if gRPC doesn't use
GET / PUT HTTP methods.

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
@@ -144,6 +144,17 @@ public abstract class AbstractManagedChannelImplBuilder
 
   private int maxInboundMessageSize = GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
 
+  /**
+   * If true, indicates that the transport may use the GET method for RPCs, and may include the
+   * request body in the query params.
+   */
+  boolean useGetForSafeMethods;
+
+  /**
+   * If true, indicates that the transport may use the PUT method for RPCs.
+   */
+  boolean usePutForIdempotentMethods;
+
   @Nullable
   BinaryLog binlog;
 
@@ -557,6 +568,14 @@ public abstract class AbstractManagedChannelImplBuilder
    */
   protected int getDefaultPort() {
     return GrpcUtil.DEFAULT_PORT_SSL;
+  }
+
+  protected final boolean useGetForSafeMethods() {
+    return useGetForSafeMethods;
+  }
+
+  protected final boolean usePutForIdempotentMethods() {
+    return usePutForIdempotentMethods;
   }
 
   /**
--- a/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
@@ -144,17 +144,6 @@ public abstract class AbstractManagedChannelImplBuilder
 
   private int maxInboundMessageSize = GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
 
-  /**
-   * If true, indicates that the transport may use the GET method for RPCs, and may include the
-   * request body in the query params.
-   */
-  boolean useGetForSafeMethods;
-
-  /**
-   * If true, indicates that the transport may use the PUT method for RPCs.
-   */
-  boolean usePutForIdempotentMethods;
-
   @Nullable
   BinaryLog binlog;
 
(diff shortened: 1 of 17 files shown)


carl-mastrangelo:
Greetings, @ejona86 !   I believe we talked about this a few months ago, but I now have a concrete use case for it.    The option to use GET and PUT are not currently exposed, and any clients today currently setting these on their MethodDescriptors manually will no longer send the methods after this change.    If this is an issue, we could expose a setting on the MCB, or possibly expose it as a system property.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -144,6 +144,17 @@
 
   private int maxInboundMessageSize = GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE;
 
+  /**
+   * If true, indicates that the transport may use the GET method for RPCs, and may include the
+   * request body in the query params.
+   */
+  boolean useGetForSafeMethods;
```
ejona86: Since these aren't used by this builder, let's not add them here. If we add methods in the future to set these it should probably be per-transport. We can just hard-code false instead of calling methods like `useGetForSafeMethods()`. If you want it for the readability, make a local variable like `boolean useGetForSafeMethods = false;` and then pass useGetForSafeMethods as an argument.

It's fine to have the plumbing like you added to the ClientTransportFactory implementations.

ejona86:
Quick side note: this is possible today if the interceptor digs into the proto reflection information. `((ProtoMethodDescriptorSupplier) methodDescriptor.getSchemaDescriptor()).getMethodDescriptor().getOptions().getIdempotencyLevel()`. Although that assumes you aren't use proto lite.

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
carl-mastrangelo: I moved ti down to the individual transport builders.

carl-mastrangelo:
> Setting safe/idempotent in the generated code will require stabilizing the MethodDescriptor methods.

I can make a proposal, and if possible the team can discuss it in the stabilization meeting.  

> That's probably okay, although it is weird they will be unused by grpc.

There have been several complaints by people at Netflix that gRPC can't really use GET.   Almost all the calls our code makes have data that is very easily cachable.  Also, we have proxies that can retry GETs in case of backend failures.   The proxies can't retry the POST, but they would prefer to.    This is to say I think it would be really helpful if gRPC servers could start accepting PUT/GET.  (assuming things are set up properly).

ejona86:
@carl-mastrangelo, if you could comment on https://github.com/grpc/grpc/issues/12039 ? The reason it stalled was there were no OSS caching proxies that supported HTTP/2 and our internal user didn't need it any more. If there's ways to actually use it, then it is possible to work on it. If there's an OSS caching proxy that would be best, but even if it is internal-only it's probably workable.

Note that it does still have some design holes, like how to enable it on client-side since the client doesn't know if the server supports it.

ejona86:
Hmm... this isn't yet safe to merge.

I see two references to setting `safe(true)` (I see none for idempotent):
 * [TesterActivity](https://github.com/grpc/grpc-java/blob/1ab651073d803e2a6dc77465434eadbca5336c59/android-interop-testing/app/src/main/java/io/grpc/android/integrationtest/TesterActivity.java#L158). This looks like it may be safe to mostly ignore. Although it would probably be good to remove the checkbox
 * [AbstractInteropTest](https://github.com/grpc/grpc-java/blob/1ab651073d803e2a6dc77465434eadbca5336c59/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java#L372). This is more problematic, because it looks like it is run internally. I probably want to just remove it from the interop scenarios we run, but it's unclear whether we should delete the test method. Maybe short-term just nerf the test with a comment saying so. I'll send out a PR to do that.

Appropriately, I don't see any _users_ internally using this feature. So it's just our own goo to deal with.

@voidzcy, would it be easy for you to change the TesterActivity to disable "safe" support?

ejona86:
> Maybe short-term just nerf the test with a comment saying so. I'll send out a PR to do that.

Actually, it turns out weird to do that in a separate PR. I'll add the comment to this one.

ejona86:
cl/271195598 is submitted internally. I created #6195 to track android-interop-test. After the tests finish I'll merge.

carl-mastrangelo:
@ejona86 thanks!   I'll ping that other issue in a moment.

PR #6196, opened 2019-09-25
stub,compiler: generated stub extends Abstract{Async,Future,Blocking}Stub

creamsoup:


creamsoup:
force commited to make it easy to review. all new changes are in [this commit](https://github.com/grpc/grpc-java/pull/6196/commits/1722ecccac113ddc6020b811974fceaa71e9a062)

dapengzhang0:
Make sure android won't complain about method count and/or APK size increase.

Monitor the change https://github.com/grpc/grpc-java/blob/master/buildscripts/kokoro/android.sh 
go/apk_size_check

ejona86:
> Make sure android won't complain about method count and/or APK size increase.

There will be an increase. But I think it may just be the addition of one class per service (no net method count changes in the generated code). We have the android/dex_diff provide the DEX method count change.

stub/build.gradle
```diff
@@ -10,7 +10,8 @@ description = "gRPC: Stub"
 dependencies {
     compile project(':grpc-api')
     testCompile libraries.truth,
-            project(':grpc-testing')
+            project(':grpc-testing'),
+            project(':grpc-testing-proto')
```
ejona86: I don't think we should include this dep.

stub/src/main/java/io/grpc/stub/AbstractAsyncStub.java
```diff
@@ -0,0 +1,61 @@
...
+package io.grpc.stub;
+
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.stub.ClientCalls.StubType;
+import javax.annotation.CheckReturnValue;
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * Stub implementations for async stubs. Stub configuration is immutable; changing the
```
ejona86: I'd prefer not to copy all the text from the base class, since then they will probably diverge. Although maybe the DO NOT MOCK would be good to keep.

stub/src/main/java/io/grpc/stub/AbstractAsyncStub.java
```diff
@@ -0,0 +1,61 @@
...
+ * <p>DO NOT MOCK: Customizing options doesn't work properly in mocks. Use InProcessChannelBuilder
+ * to create a real channel suitable for testing. It is also possible to mock Channel instead.
+ *
+ * @since 1.25.0
+ */
+@ThreadSafe
+@CheckReturnValue
+public abstract class AbstractAsyncStub<S extends AbstractAsyncStub<S>> extends AbstractStub<S> {
+
+  protected AbstractAsyncStub(Channel channel) {
```
ejona86: I guess we could drop this constructor as well?

stub/src/main/java/io/grpc/stub/AbstractAsyncStub.java
```diff
@@ -0,0 +1,61 @@
...
+    this(channel, CallOptions.DEFAULT);
+  }
+
+  protected AbstractAsyncStub(Channel channel, CallOptions callOptions) {
+      super(channel, callOptions);
+  }
+
+  // TODO(jihuncho) javadoc
+  public static <T extends AbstractStub<T>> T newStub(StubFactory<T> factory, Channel channel) {
```
ejona86: s/AbstractStub/AbstractAsyncStub/

creamsoup: removed unused newStub functions from AbstractStub to make it work.

stub/src/test/java/io/grpc/stub/StubTypeTest.java
```diff
@@ -0,0 +1,159 @@
...
+import io.grpc.ManagedChannel;
+import io.grpc.MethodDescriptor;
+import io.grpc.Server;
+import io.grpc.inprocess.InProcessChannelBuilder;
+import io.grpc.inprocess.InProcessServerBuilder;
+import io.grpc.stub.ClientCalls.StubType;
+import io.grpc.testing.GrpcCleanupRule;
+import io.grpc.testing.protobuf.SimpleRequest;
+import io.grpc.testing.protobuf.SimpleResponse;
+import io.grpc.testing.protobuf.SimpleServiceGrpc;
```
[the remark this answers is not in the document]
creamsoup: i actually dislike extending class, but I did it. and if you don't like this, i can change it. PTAL

stub/src/main/java/io/grpc/stub/AbstractStub.java
```diff
@@ -52,16 +52,6 @@
   private final Channel channel;
   private final CallOptions callOptions;
 
-  /**
-   * Constructor for use by subclasses, with the default {@code CallOptions}.
-   *
-   * @since 1.0.0
-   * @param channel the channel that this stub will use to do communications
-   */
-  protected AbstractStub(Channel channel) {
```
ejona86: We can't remove this. Existing generated code is using it.

stub/src/main/java/io/grpc/stub/AbstractAsyncStub.java
ejona86: That doesn't sound right. We should get to the bottom of that.

creamsoup: because they have same method signature after the type erasure, static "override". either we keep `newStub` methods in `AbstractStub` or we use `<T extends AbstractStub<T>>` in extended Stub classes.

ejona86:
We can keep the generics the same for all the `Abstruct*Stub.newStub()` methods. We can then do runtime type checking to make sure that the blocking stub returns a blocking stub and such. We aren't _wild_ about this approach, but it seems much better than any other options we thought up.

creamsoup:
added the runtime check, PTYAL

stub/src/main/java/io/grpc/stub/AbstractStub.java
```diff
@@ -101,6 +102,32 @@ public final CallOptions getCallOptions() {
...
+   *
+   * @since 1.25.0
+   * @param factory the factory to create a stub
+   * @param channel the channel that this stub will use to do communications
+   * @param callOptions the runtime call options to be applied to every call on this stub
+   */
+  public static <T extends AbstractStub<T>> T newStub(
+      StubFactory<T> factory, Channel channel, CallOptions callOptions) {
+    return factory.newStub(
+        channel, callOptions.withOption(ClientCalls.STUB_TYPE_OPTION, StubType.FUTURE));
```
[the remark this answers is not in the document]
creamsoup: right, thanks added a test as well.

PR #6212, opened 2019-09-27
build: add mavenCentral to repositories

creamsoup:


creamsoup:
this helps to resolve the `netty 4.1.42.Final` which is not yet available on the google mirror. It is nice to have fall back anyways.

dapengzhang0:
There may be other `build.gradle` files can be added the same, e.g. `examples/build.gradle`.

creamsoup:
sure, i didn't add the one has `jcenter()` because it always does that. I am not oppose to add to other examples, but seems little redundant.

PR #6215, opened 2019-09-27
alts: Use Conscrypt when available

### LINKED ISSUE #6213: ALTS should use Conscrypt when available
Java 8 has slow AES-GCM
https://github.com/grpc/grpc-java/blob/81da3eb95be37fa0647ce8da2e19de96ab84c600/alts/src/main/java/io/grpc/alts/internal/AesGcmAeadCrypter.java#L43

Something like this would work:
```java
cipher = Cipher.getInstance(AES_GCM, Conscrypt.newProvider());
```

Although it would be better to create the Conscrypt provider with reflection, so that Conscrypt is not a required dependency. There are platforms where Conscrypt is unsupported and making a fat jar with it does require extra configuration. I'm unsure whether we should add a dependency on Conscrypt or let the "user" do so (even if that is gax-java).

ejona86:
We depend on Conscrypt to help ensure Conscrypt 2.1.0 or newer is used.
It's not 100% clear this is the best approach, but it is the simplest at
present. If Conscrypt is not available then we will just use the JDK's
slower implementation of AES-GCM.

Fixes #6213

### DIFF
--- a/alts/build.gradle
+++ b/alts/build.gradle
@@ -19,7 +19,8 @@ dependencies {
             project(':grpc-protobuf'),
             project(':grpc-stub'),
             libraries.lang,
-            libraries.protobuf
+            libraries.protobuf,
+            libraries.conscrypt
     compile (libraries.google_auth_oauth2_http) {
         // prefer 26.0-android from libraries instead of 25.1-android
         exclude group: 'com.google.guava', module: 'guava'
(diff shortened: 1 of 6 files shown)


ejona86:
@creamsoup, PTAL

creamsoup:
still LGTM!

PR #6226, opened 2019-10-01
Update to GSON 2.8.5 and protobuf 3.10.0

elharo:
This turned out to require also updating protobuf

@ejona86 Windows build appears to be stuck but otherwise passes.


ejona86:
@elharo, thank you! I know upgrading Protobuf is a pain, since we have so many references to it!

PR #6228, opened 2019-10-02
xds: plumbing restructured xds loadbalancer

dapengzhang0:
Moved `ManagedChannel initLbChannel(Helper helper, String balancerName)` out of `LookasideChannelLb` because `LoadReportClientImpl` need channel be initialized already. 

Migration will be in an upcoming PR.

### DIFF
--- a/xds/src/main/java/io/grpc/xds/LookasideLb.java
+++ b/xds/src/main/java/io/grpc/xds/LookasideLb.java
@@ -17,18 +17,23 @@
 package io.grpc.xds;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static java.util.logging.Level.FINEST;
 
 import com.google.common.annotations.VisibleForTesting;
 import io.grpc.Attributes;
 import io.grpc.LoadBalancer;
 import io.grpc.LoadBalancerProvider;
 import io.grpc.LoadBalancerRegistry;
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
 import io.grpc.NameResolver.ConfigOrError;
 import io.grpc.util.ForwardingLoadBalancer;
 import io.grpc.util.GracefulSwitchLoadBalancer;
+import io.grpc.xds.LocalityStore.LocalityStoreImpl;
 import io.grpc.xds.XdsComms.AdsStreamCallback;
 import io.grpc.xds.XdsLoadBalancerProvider.XdsConfig;
 import java.util.Map;
+import java.util.logging.Logger;
 
 /** Lookaside load balancer that handles balancer name changes. */
 final class LookasideLb extends ForwardingLoadBalancer {
@@ -40,7 +45,12 @@ final class LookasideLb extends ForwardingLoadBalancer {
 
   private String balancerName;
 
-  // TODO(zdapeng): Add LookasideLb(Helper helper, AdsCallback adsCallback) with default factory
+  LookasideLb(Helper lookasideLbHelper, AdsStreamCallback adsCallback) {
+    this(
+        lookasideLbHelper, adsCallback, new LookasideChannelLbFactoryImpl(),
+        LoadBalancerRegistry.getDefaultRegistry());
+  }
+
   @VisibleForTesting
   LookasideLb(
       Helper lookasideLbHelper,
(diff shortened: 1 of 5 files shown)


xds/src/main/java/io/grpc/xds/LookasideLb.java
```diff
@@ -112,4 +122,38 @@ public LoadBalancer newLoadBalancer(Helper helper) {
   interface LookasideChannelLbFactory {
     LoadBalancer newLoadBalancer(Helper helper, AdsStreamCallback adsCallback, String balancerName);
   }
+
+  private static final class LookasideChannelLbFactoryImpl implements LookasideChannelLbFactory {
```
voidzcy: You do not need both `LoadBalancerProvider` and `LookasideChannelLbFactory`. You `LookasideLb`'s secondary constructor can just take a `LoadBalancerProvider` that creates `LookasideChannelLb` instances. The implementation for how to create `LookasideChannelLb` instances can be private. In your test, you can fake the `LoadBalancerProvider` to capture the `helper`.

```java
LookasideLb(Helper helper, ...) {
   this(..., new LookasideChannelLbProvider(...));
}

LookasideLb(..., LoadBalancerProvider provider) {
   ...
}

private class LookasideChannelLbProvider extends LoadBalancerProvider {
   ...   

   @Override
   public LoadBalancer newLoadBalancer(Helper helper) {
      // Create a LookasideChannelLb instance.
   }
}

```

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
```diff
@@ -245,4 +248,18 @@ public void updateBalancingState(ConnectivityState newState, SubchannelPicker ne
   interface LookasideLbFactory {
     LoadBalancer newLoadBalancer(Helper helper, AdsStreamCallback adsCallback);
   }
+
+  private static final class LookasideLbFactoryImpl implements LookasideLbFactory {
```
voidzcy: You don't need to create a separate `LookasideLbFactory` interface. `LoadBalancer.Factory` suffices. You can make this class non-static and access `AdsCallback` instance directly. You would need a secondary `XdsLoadBalancer2` constructor that takes an `AdsStreamCallback` instance. Make your implementation for `AdsStreamCallback` a standalone class with name (instead of anonymous), and make it package-private with `@VisibleForTesting`. You can access it in test. I feel like this is more elegant.

xds/src/main/java/io/grpc/xds/LookasideLb.java
```diff
@@ -40,7 +45,12 @@
 
   private String balancerName;
 
-  // TODO(zdapeng): Add LookasideLb(Helper helper, AdsCallback adsCallback) with default factory
+  LookasideLb(Helper lookasideLbHelper, AdsStreamCallback adsCallback) {
+    this(
```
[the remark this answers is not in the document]
dapengzhang0: What's the problem with this format? What format is better?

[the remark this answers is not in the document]
voidzcy:
```java
this(lookasideLbHelper, adsCallback, new LookasideChannelLbFactoryImpl(),
      LoadBalancerRegistry.getDefaultRegistry());
```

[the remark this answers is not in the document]
dapengzhang0: I think google-java-format or cider reformat won't do it that way.

xds/src/main/java/io/grpc/xds/LookasideLb.java
dapengzhang0: Then this provider will have a `balancerName` that can be changed. It won't work for graceful switch. Even it can work after some hack, I'm scared of having the `LookasideChannelLbProvider` depending on the state of `LookasideLb`. It should be totally independent of and have no knowledge of the state of `LookasideLb`.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
dapengzhang0: Technically this is verify difficult:

```
XdsLoadBalancer2(Helper helper) {
    this(helper, new LookasideLbFactory(), new FallbackLbFactory());
  }
```

won't compile because can not referece non-static `LookasideLbFactory` in constructor.

Also, I'm not a fan of having `LookasideLbFactory` appear like depending the state of `XdsLoadBalancer2` and appear like it can be changed while the state of `XdsLoadBalancer2` changes.

voidzcy: Oh, I forget that won't compile...

But still, you can pass your `AdsStreamCallback` instance into the constructor of `LookasideLbFactory`, which extends `LoadBalancer.Factory`. We should try to minimize the number of newly interfaces.

dapengzhang0:
```
XdsLoadBalancer2(Helper helper) {
    this(helper, new LookasideLbFactory(adsCallback), new FallbackLbFactory());
  }
```

This or other approaches also won't compile, the technical difficulty is not trivial.

dapengzhang0: If an object you are injecting to the constructor depends on the class implementation, then you should not _only_ use the mocked object in test, because that object depends on implementation.

voidzcy: No, I mean you can make `LookasideLbFactory` static, not depending on `XdsLoadBalancer2`'s state. You just need a `LookasideLbFactory(AdsStreamCallback)` constructor.

xds/src/main/java/io/grpc/xds/LookasideLb.java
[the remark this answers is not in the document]
voidzcy: google-java-format would accept that. This looks much better.

xds/src/main/java/io/grpc/xds/XdsLoadBalancer2.java
dapengzhang0: I knew what you mean. Look at my second snippet, it won't compile.

dapengzhang0: The adsStreamCallback passed in the constructor depends on and changes the state of XdsLoadBalancer2.

xds/src/main/java/io/grpc/xds/LookasideLb.java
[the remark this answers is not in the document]
dapengzhang0: I dare say they _look_ equally good. And ["The prime directive of line-wrapping is: prefer to break at a higher syntactic level."](https://google.github.io/styleguide/javaguide.html#s4.5.1-line-wrapping-where-to-break)

xds/src/test/java/io/grpc/xds/LookasideLbTest.java
```diff
@@ -157,4 +158,27 @@ public void handleChildPolicyChangeThenBalancerNameChangeThenChildPolicyChange()
     lookasideLb.shutdown();
     verify(balancer2).shutdown();
   }
+
+  @Test
+  public void handleAddress_verifyTryToCreateLbChannelToBalancerName()
```
voidzcy: nit: do not add "handleResolvedAddress_createLbChannel".

Issue #6146, opened 2019-09-12
Android kokoro build is very flaky


dapengzhang0:
Android kokoro build became very flaky after #6103

dapengzhang0:
https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod%3Agrpc%2Fjava%2Fmaster%2Fpresubmit%2Fandroid

ejona86:
```
* What went wrong:
Metaspace
```

https://source.cloud.google.com/results/invocations/c66a5a4a-ff2c-4698-b313-a416a8155e0b/targets/grpc%2Fjava%2Fmaster%2Fpresubmit%2Fandroid/log

dapengzhang0:
Removing the highly flaky label as it is not seen again this week. It may also be possible that flakiness occurred after #6103 is by coincidence.

ejona86:
#6103 actually seems a reasonable trigger, because it means the gradle task rebuilds more code. It looks like the failures resolved after #6131 was merged (it is the only PR between the last failure and first success), which is not wholly unreasonable. It isn't completely clear to me why that PR fixed it, but the change impacted gradle enough it does make some sense.

I'm going to mark this as resolved. It could easily rear its head again, but it seems resolved for the while.

dapengzhang0:
I believe #6131 resolves this issue. Actually #6131 itself fails the android build because it was rebuilding head^ with the old build.gradle, but once it's merged no more PR fails.

ejona86:
Uh, oh. This may be resurfacing. I just saw https://source.cloud.google.com/results/invocations/8dcbb963-10c3-43e5-9345-9e0a3a7ed6b6

ejona86:
Actually. That's because it is on the v1.24.x branch. I guess we need a fix for the branch.

dapengzhang0:
An easy way to fix 1.24.x branch is to revert #6103

ejona86:
@dapengzhang0, ah, yeah, that would probably work well. Would you mind sending out a PR?

creamsoup:
I encountered same issue of my backport to 24. this fix is also backported, so reopening the issue. 

https://source.cloud.google.com/results/invocations/f47ddafa-0b41-4c79-97cc-01e050129fa8/targets/grpc%2Fjava%2Fv1.24.x%2Fpresubmit%2Fandroid/log

dapengzhang0:
[...]

PR #6233, opened 2019-10-02
Revert "buildscripts: android.sh to clean the build when building HEAD^"

dapengzhang0:
This reverts commit c240f270773a49eb9c78ad7bcf7ef33c09aa5094.

Fixing #6146 in v1.24.x branch.


dapengzhang0:
Android build failure is expected. There _should_ be no more failure after this PR merged.

ejona86:
Why is the android build failure expected? I don't see any gradle clean in the output, yet it still OOMed.

dapengzhang0:
I was wrong. I mistakenly thought after `checkout HEAD^` it runs the old `android.sh`, but actually it only invoke `build.gradle`. Then the problem is still there. :(

Issue #3703, opened 2017-11-09
DnsNameResolver don't use provided executorPool


Gordiychuk:
### What version of gRPC are you using?
v1.7.0

### What JVM are you using (`java -version`)?
java version "1.8.0_112"
Java(TM) SE Runtime Environment (build 1.8.0_112-b15)
Java HotSpot(TM) 64-Bit Server VM (build 25.112-b15, mixed mode)

### What did you do?
```java
        ExecutorService customExecutor = Executors.newFixedThreadPool(2, new DefaultThreadFactory("grpc-client-test"));
        EventLoopGroup customEventLoop = new EpollEventLoopGroup(1);

        ManagedChannel channel = NettyChannelBuilder.forAddress("localhost", 8080)
                .eventLoopGroup(customEventLoop)
                .executor(customExecutor)
                .build();

        MyServiceGrpc.MyServiceStub stub = MyServiceGrpc.newStub(channel);
        // call any method on stub
        // ...
```

### What did you expect to see?
Will not be created additional thread for rpc call expect configured before.

### What did you see instead?
In my case grpc still use default CachedThreadPool and creates at least 32 thread. 
```
solomon-alert_1  | "grpc-default-executor-31" #124 daemon prio=5 os_prio=0 tid=0x00007f1762851800 nid=0xb3 waiting on condition [0x00007f16ec3cf000]
solomon-alert_1  |    java.lang.Thread.State: TIMED_WAITING (parking)
solomon-alert_1  |      at sun.misc.Unsafe.park(Native Method)
solomon-alert_1  |      - parking to wait for  <0x00000005923d3148> (a java.util.concurrent.SynchronousQueue$TransferStack)
solomon-alert_1  |      at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:215)
solomon-alert_1  |      at java.util.concurrent.SynchronousQueue$TransferStack.awaitFulfill(SynchronousQueue.java:460)
solomon-alert_1  |      at java.util.concurrent.SynchronousQueue$TransferStack.transfer(SynchronousQueue.java:362)
solomon-alert_1  |      at java.util.concurrent.SynchronousQueue.poll(SynchronousQueue.java:941)
solomon-alert_1  |      at java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1066)
solomon-alert_1  |      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1127)
[...]

PR #6238, opened 2019-10-03
api,core: Adds an Executor field to NameResolver.Args.

groakley:
Adds an Executor field to NameResolver.Args, which is optionally set on ManagedChannelBuilder. This allows NameResolver implementations to avoid creating their own thread pools if the application already manages its own pools. 

Addresses #3703.

### DIFF
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -182,7 +182,7 @@ final class DnsNameResolver extends NameResolver {
     this.stopwatch = Preconditions.checkNotNull(stopwatch, "stopwatch");
     this.syncContext =
         Preconditions.checkNotNull(args.getSynchronizationContext(), "syncContext");
-    this.executor = args.getExecutor();
+    this.executor = args.getBlockingExecutor();
     this.usingExecutorResource = executor == null;
   }
 
@@ -194,7 +194,7 @@ final class DnsNameResolver extends NameResolver {
   @Override
   public void start(Listener2 listener) {
     Preconditions.checkState(this.listener == null, "already started");
-    if (executor == null) {
+    if (usingExecutorResource) {
       executor = SharedResourceHolder.get(executorResource);
     }
     this.listener = Preconditions.checkNotNull(listener, "listener");
--- a/core/src/main/java/io/grpc/internal/DnsNameResolver.java
+++ b/core/src/main/java/io/grpc/internal/DnsNameResolver.java
@@ -138,6 +138,8 @@ final class DnsNameResolver extends NameResolver {
   private final String authority;
   private final String host;
   private final int port;
+
+  /** Executor that will be used if an Executor is not provide via {@link NameResolver.Args}. */
   private final Resource<Executor> executorResource;
   private final long cacheTtlNanos;
   private final SynchronizationContext syncContext;
@@ -147,6 +149,10 @@ final class DnsNameResolver extends NameResolver {
   private ResolutionResults cachedResolutionResults;
   private boolean shutdown;
   private Executor executor;
+
+  /** True if using an executor resource that should be released after use. */
+  private final boolean usingExecutorResource;
+
   private boolean resolving;
 
   // The field must be accessed from syncContext, although the methods on an Listener2 can be called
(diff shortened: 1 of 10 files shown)


groakley:
Some additional context: Before this change, there is no way to prevent the Channel returned by the OkHttpChannelBuilder from creating its own thread pools (even if [executor()](https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannelBuilder.html#executor-java.util.concurrent.Executor-) is used to override the default).

From the user's perspective, this means that even though the executor() override was used, "default-grpc-executor*" threads will still be started.

After this change, the NameResolver executor can be overridden using the channelbuilder, e.g.

```
OkHttpChannelBuilder.forAddress("foo", 123)
    .executor(executor1)
    .transportExecutor(executor2)
    .nameResolverExecutor(executor3)
    .build();
```

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -141,7 +141,9 @@
   private final Executor executor;
   private final ObjectPool<? extends Executor> executorPool;
   private final ObjectPool<? extends Executor> balancerRpcExecutorPool;
+  private final ObjectPool<? extends Executor> nameResolverExecutorPool;
```
dapengzhang0: This field is not needed.

ejona86:
I don't know why the discussion on #3703 was ignored.

We don't want an executor just for the name resolver on ManagedChannelBuilder. The reason the NameResolver needs an executor is that it does blocking calls. We have a few cases that need an executor for blocking reasons. Please change it so blockingExecutor() is the exposed API.

Also, in the Args API it should be clear the purpose of the Executor: that it is for blocking operations. The API should be getBlockingExecutor()/setBlockingExecutor().

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -186,7 +194,9 @@ public String getServiceAuthority() {
   @Override
   public void start(Listener2 listener) {
     Preconditions.checkState(this.listener == null, "already started");
-    executor = SharedResourceHolder.get(executorResource);
+    if (executor == null) {
```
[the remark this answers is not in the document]
groakley: Done

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
groakley: Done

groakley:
Renamed NameExecutorResolver->BlockingExecutor.

I use git/github about once a year to contribute to this project. Looks like I screwed up the branch trying to rebase from head. (Looks good in my fork project, totally screwed up in the remote project). Might be something that is easy to resolve, but I can also recreate this PR if that is the easiest path forward.

ejona86:
Shoot. Did it a bit differently than I wanted. One moment.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -565,16 +566,23 @@ ClientStream newSubstream(ClientStreamTracer.Factory tracerFactory, Metadata new
...
+            .setDefaultPort(builder.getDefaultPort())
+            .setProxyDetector(proxyDetector)
+            .setSynchronizationContext(syncContext)
+            .setServiceConfigParser(
+                new ScParser(
+                    retryEnabled,
+                    builder.maxRetryAttempts,
+                    builder.maxHedgedAttempts,
+                    loadBalancerFactory))
+            .setBlockingExecutor(builder.blockingExecutorPool.getObject())
```
ejona86: This getObject() is not returned; the holder is released, but blockingExecutorHolder.getExecutor() wasn't called so it won't return the object. That particular issue can be fixed by just not using ExecutorHolder for the blocking pool. But...

We probably _do_ want to delay allocating the executor until first use. We can make a `ExecutorHolderExecutor` (tehe) with:
```
public void execute(Runnable runnable) {
  holder.getExecutor().execute(runnable);
}
```

Things might be reworked to be less weird, but that's probably good enough for now.

groakley: Whoops, yeah I need to get the instance from the ExecutorHolder.

But wouldn't the proposed solution imbalance the SharedResource refcounting? There would be a `SharedResourceHolder.get()` each time runnable is submitted to the executor, but `SharedResourceHolder.release()` only gets called once when the Channel is terminated.

So iiuc there actually needs to be some additional reference tracking the in the `ExecutorHolderExecutor` to ensure getExecutor() is only called for the first runnable.

ejona86: The ExecutorHolder doesn't need any balancing. It is just lazily grabbing a reference. So the first time getExecutor() is called it will get a reference, but the rest will use what it has.

groakley: Gotcha, fixed.

PR #6245, opened 2019-10-04
COMPILING.md: fix protobuf build steps 

sanjaypujare:
we don't need Linux ldconfig and ld.so.conf stuff because of the --noshared option

### DIFF
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -61,13 +61,6 @@ Protobuf installs to ``/usr/local`` by default.
 For Visual C++, please refer to the [Protobuf README](https://github.com/google/protobuf/blob/master/cmake/README.md)
 for how to compile Protobuf. gRPC-java assumes a Release build.
 
-#### Linux and MinGW
-If ``/usr/local/lib`` is not in your library search path, you can add it by running:
-```
-$ sudo sh -c 'echo /usr/local/lib >> /etc/ld.so.conf'
-$ sudo ldconfig
-```
-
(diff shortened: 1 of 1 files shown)


COMPILING.md
```diff
@@ -45,6 +45,7 @@ For Linux, Mac and MinGW:
 $ git clone https://github.com/google/protobuf.git
 $ cd protobuf
 $ git checkout v3.9.0
+$ git submodule update --init --recursive
```
ejona86: You shouldn't need the submodules. I normally do a build from the .tar.gz, which also doesn't have submodules.

Hmm... we also run "make check"? Meh. And autogen shouldn't be necessary any more.

Let me send out some modifications here.

sanjaypujare: For the "make check" to succeed we need submodules - that's what I ran into

ejona86: Sent out #6249, which I tested on a fresh Debian Buster container.

ejona86: Ah, I see. That makes sense, because that wasn't likely previously necessary. I think it's better to just remove the make check.

ejona86:
I still want this PR for the ldconfig removal.

sanjaypujare:
Ok, I’ll update this to remove sub modules step once I verify the other PR

COMPILING.md
sanjaypujare: removed this update from the PR.

PR #6254, opened 2019-10-07
xds: SdsX509TrustManager implementation for XDS and SDS 

sanjaypujare:
xds: has code to check peer cert SANs against verify_subject_alternate_name from CertificateValidationContext


creamsoup:
Oracle doc has example code:

```java
/* 
The following example reads a file with Base64 encoded certificates, which are each bounded 
at the beginning by -----BEGIN CERTIFICATE-----, and bounded at the end by 
-----END CERTIFICATE-----. We convert the FileInputStream (which does not support mark and 
reset) to a BufferedInputStream (which supports those methods), so that each call to 
generateCertificate consumes only one certificate, and the read position of the input stream is 
positioned to the next certificate in the file:
*/

 FileInputStream fis = new FileInputStream(filename);
 BufferedInputStream bis = new BufferedInputStream(fis);

 CertificateFactory cf = CertificateFactory.getInstance("X.509");

 while (bis.available() > 0) {
    Certificate cert = cf.generateCertificate(bis);
    System.out.println(cert.toString());
 }
```

that example looks almost same as the Util class. can we just use that? `CerficiateFactory` has `generateCertificates` which parses all in one shot.

sanjaypujare:
> Oracle doc has example code:
> ...
> that example looks almost same as the Util class. can we just use that? `CerficiateFactory` has `generateCertificates` which parses all in one shot.

Good point. Will look into it and make changes as appropriate. I got the code from Netty's `io.netty.handler.ssl.SslContext.toX509Certificates(java.io.File)` which I couldn't directly call because of visibility.

xds/src/main/java/io/grpc/xds/sds/trust/CertificateUtils.java
```diff
@@ -0,0 +1,50 @@
...
+
+  private static CertificateFactory factory;
+
+  private static void initInstance() throws CertificateException {
+    if (factory == null) {
+      factory = CertificateFactory.getInstance("X.509");
+    }
+  }
+
+  static synchronized X509Certificate[] toX509Certificates(String fileName)
```
creamsoup: if you are using `synchronized` because some jre's implementation is not thread safe, also add `synchronized` for the `initInstance`. in case, other methods will be added might trigger the race depends on jre.

xds/src/main/java/io/grpc/xds/sds/trust/SdsX509TrustManager.java
```diff
@@ -0,0 +1,295 @@
...
+import javax.net.ssl.X509TrustManager;
+
+/**
+ * Extension of {@link X509ExtendedTrustManager} that implements verification of
+ * SANs (subject-alternate-names) against the list in CertificateValidationContext.
+ */
+final class SdsX509TrustManager extends X509ExtendedTrustManager implements X509TrustManager {
+
+  @SuppressWarnings("unused")
+  private static final Logger logger = Logger.getLogger(SdsX509TrustManager.class.getName());
```
creamsoup: not sure why you want to keep this logger if not used?

xds/src/test/java/io/grpc/xds/sds/trust/SdsX509TrustManagerTest.java
```diff
@@ -0,0 +1,236 @@
...
+
+/**
+ * Unit tests for {@link SdsX509TrustManager}.
+ */
+@RunWith(JUnit4.class)
+public class SdsX509TrustManagerTest {
+  /**
+   * server1 has 4 SANs.
+   */
+  private static final String SERVER_1_PEM_FILE = "../testing/src/main/resources/certs/server1.pem";
```
creamsoup: let's not access other modules file

xds/src/main/java/io/grpc/xds/sds/trust/SdsX509TrustManager.java
```diff
@@ -0,0 +1,295 @@
...
+import javax.annotation.Nullable;
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.X509ExtendedTrustManager;
+import javax.net.ssl.X509TrustManager;
+
+/**
+ * Extension of {@link X509ExtendedTrustManager} that implements verification of
+ * SANs (subject-alternate-names) against the list in CertificateValidationContext.
+ */
+final class SdsX509TrustManager extends X509ExtendedTrustManager implements X509TrustManager {
```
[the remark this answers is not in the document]
sanjaypujare: Used the IDE to generate method definitions and it generates AFTER the calling function. Will fix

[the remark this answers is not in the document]
sanjaypujare: rearranged methods so it's all bottom to top (called methods at the top)

PR #6258, opened 2019-10-09
bazel: Support --incompatible_load_proto_rules_from_bzl

ejona86:
This flag will be enabled by default in Bazel 1.0

Since I changed the grpc-proto repo version for Bazel, I also synced the
protos.


dapengzhang0:
Do we also load this rule in examples?

ejona86:
Ah, yes, we do need to update the examples. Done.

Issue #6202, opened 2019-09-26
INTERNAL: Protocol error Rst Stream when communicating with .NET Core 3 server.


ivan-penchev:
I have compiled a repo with the issue in a reproducible manner : 
https://github.com/ivan-penchev/gRPC-Java-dotnetcore3/tree/master

I have basically download the "Greeter" example from dotnet and java  repos and tried to make them work together. Since we are trying to migrate our services to .NET Core 3.0;

Someone reported the same issue [on stackoverflow for Kotlin](https://stackoverflow.com/questions/57414584/rst-stream-grpc-error-on-android-kotlin-client-when-requesting-net-core-3-0-grp)

I have tried the same .NET Core 3 server with Python, .NET and BloomRPC client and it worked flawlessly, which is why I post the issue here.
### What version of gRPC are you using?
Java uses : 1.23.0
.NET uses: 2.23.2

### What operating system (Linux, Windows,...) and version?
I am running on Windows, using Docker in a Linux container.

### What did you expect to see?
I expected to see a response containing "Hello ~whatever message I sent initially~".

### What did you see instead?

```java
Failed... : 
io.grpc.StatusRuntimeException: INTERNAL: Protocol error
Rst Stream
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:235)
	at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:216)
	at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:141)
	at io.grpc.examples.helloworld.GreeterGrpc$GreeterBlockingStub.sayHello(GreeterGrpc.java:228)
	at io.grpc.helloworldexample.HelloworldActivity$GrpcTask.doInBackground(HelloworldActivity.java:90)
	at io.grpc.helloworldexample.HelloworldActivity$GrpcTask.doInBackground(HelloworldActivity.java:72)
	at android.os.AsyncTask$2.call(AsyncTask.java:333)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:245)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
	at java.lang.Thread.run(Thread.java:764)

```

### Steps to reproduce the bug

[...]

PR #6260, opened 2019-10-09
okhttp: fix header scheme does not match transport type.

ran-su:
Get transport type by check if `sslSocketFactory` is `null`. Set scheme based on the transport type.

fixes: https://github.com/grpc/grpc-java/issues/6202

### DIFF
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
@@ -415,7 +415,7 @@ class OkHttpClientStream extends AbstractClientStream {
               authority,
               userAgent,
               useGet,
-              transport.usePlaintext());
+              transport.isUsingPlaintext());
       transport.streamReadyToStart(OkHttpClientStream.this);
     }
 
--- a/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
+++ b/okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
@@ -408,7 +408,14 @@ class OkHttpClientStream extends AbstractClientStream {
 
     @GuardedBy("lock")
     private void streamReady(Metadata metadata, String path) {
-      requestHeaders = Headers.createRequestHeaders(metadata, path, authority, userAgent, useGet);
+      requestHeaders =
+          Headers.createRequestHeaders(
+              metadata,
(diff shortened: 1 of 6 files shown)


okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -60,6 +60,7 @@
   private final Attributes attributes;
 
   private boolean useGet = false;
+  private boolean usePlaintext;
```
creamsoup: can it be final?

ran-su: Yeah, it should be final, fixed.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -75,7 +76,8 @@
       StatsTraceContext statsTraceCtx,
       TransportTracer transportTracer,
       CallOptions callOptions,
-      boolean useGetForSafeMethods) {
+      boolean useGetForSafeMethods,
+      boolean usePlaintext) {
```
sanjaypujare: I wonder if we need to pass this if this can be deduced from the `transport` argument (of type `OkHttpClientTransport`). For plaintext is it possible that `sslSocketFactory` is null, so all we need is to expose that in `OkHttpClientTransport`. Trying to see if we can avoid redundancy.

ran-su: What we want is just to know if we are using plaintext. In my opinion, passing a boolean is much safer than expose the `sslSocketFactory`.

sanjaypujare: Not exposing `sslSocketFactory`; just a boolean method returning true if `sslSocketFactory == null` (assuming that flow works)

ran-su: That's a great point, changed my implementation.

sanjaypujare: Great, the test `io.grpc.okhttp.OkHttpChannelBuilderTest#usePlaintextCreatesNullSocketFactory' confirms the assumption made here. LGTM

PR #6261, opened 2019-10-09
xds: first installment of SdsProtocolNegotiators with Client and Server builders

sanjaypujare:


xds/src/main/java/io/grpc/xds/sds/XdsServerBuilder.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  @Override
+  public XdsServerBuilder fallbackHandlerRegistry(@Nullable HandlerRegistry fallbackRegistry) {
+    delegate.fallbackHandlerRegistry(fallbackRegistry);
+    return this;
+  }
+
+  @Override
+  public XdsServerBuilder useTransportSecurity(File certChain, File privateKey) {
```
creamsoup: how this will work with the SDS?

xds/src/main/java/io/grpc/xds/sds/internal/SdsProtocolNegotiators.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Override
+    public AsciiString scheme() {
+      return SCHEME;
+    }
+
+    @Override
+    public ChannelHandler newHandler(GrpcHttp2ConnectionHandler grpcHandler) {
+      //TODO(sanjaypujare): once implemented return ServerSdsHandler
+      return null;
```
creamsoup: throwing exception might be better here.

xds/src/main/java/io/grpc/xds/sds/XdsChannelBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+import io.grpc.netty.NettyChannelBuilder;
+import io.grpc.xds.sds.internal.SdsProtocolNegotiators;
+import java.net.SocketAddress;
+import javax.annotation.CheckReturnValue;
+
+/**
+ * A wrapper around NettyChannelBuilder that sets up the appropriate ProtocolNegotiator
+ * and ChannelHandlers to use SDS for TLS.
```
creamsoup: this doc just describes the implementation detail, should describe the function of class like AltsChannelBuilder. also, applies to the server builder. also the doc should mention that this is not ready to use, etc.

xds/src/main/java/io/grpc/xds/sds/internal/SdsProtocolNegotiators.java
```diff
@@ -0,0 +1,119 @@
...
+ * An internal accessor for {@link ProtocolNegotiator}s  similar to
+ * {@link io.grpc.netty.InternalProtocolNegotiators}. A gRPC {@link ProtocolNegotiator} is used
+ * to provide SSL context using certs/keys provided by SDS.
+ */
+public class SdsProtocolNegotiators {
+
+  private static final AsciiString SCHEME = AsciiString.of("https");
+
+  /**
+   * Factory for Netty clients protocol negotiation.
```
creamsoup: should mention sds negotiation? this doc is less descriptive than the class name itself.

xds/src/main/java/io/grpc/xds/sds/internal/SdsProtocolNegotiators.java
```diff
@@ -0,0 +1,119 @@
...
+import io.grpc.netty.InternalProtocolNegotiator;
+import io.grpc.netty.InternalProtocolNegotiator.ProtocolNegotiator;
+import io.grpc.netty.NettyChannelBuilder;
+import io.netty.channel.ChannelHandler;
+import io.netty.util.AsciiString;
+
+/**
+ * An internal accessor for {@link ProtocolNegotiator}s  similar to
+ * {@link io.grpc.netty.InternalProtocolNegotiators}. A gRPC {@link ProtocolNegotiator} is used
+ * to provide SSL context using certs/keys provided by SDS.
```
creamsoup: this is not internal accessor. it is actual implementation of ProtocolNegotiators for SDS, can you update this doc?

xds/src/main/java/io/grpc/xds/sds/XdsChannelBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+  private final NettyChannelBuilder delegate;
+
+  private XdsChannelBuilder(NettyChannelBuilder delegate) {
+    this.delegate = delegate;
+    SdsProtocolNegotiators.setProtocolNegotiatorFactory(delegate);
+  }
+
+  /**
+   * Creates a new builder with the given server address.
+   * See {@link NettyChannelBuilder#forAddress(SocketAddress)} for more info.
```
creamsoup: FYI. it is unclear why this is 2nd line. if you want the javadoc look like this, consider using `<p>`. if not please ignore.

```java
/** 
 * Create....
 *
 * <p>See ...
 */
```

xds/src/main/java/io/grpc/xds/sds/XdsServerBuilder.java
sanjaypujare: Good point and ideally this should not be called when using SDS. Eventually there could be more complex cases I was thinking of - like SDS does not return any cert info so the user wants their own cert/key to be used. But for now I can change this to throw an exception or leave it as it is for now and change it to the "right thing" (whatever that happens to be) later. Comments?

xds/src/main/java/io/grpc/xds/sds/XdsChannelBuilder.java
sanjaypujare: No the 2nd line is just for convenience (to avoid breaking at some other place)

xds/src/main/java/io/grpc/xds/sds/internal/SdsProtocolNegotiators.java
sanjaypujare: removing the comment - private class

xds/src/main/java/io/grpc/xds/sds/XdsChannelBuilder.java
sanjaypujare: the reformat tool took care of this

xds/src/main/java/io/grpc/xds/sds/internal/SdsProtocolNegotiators.java
```diff
@@ -0,0 +1,110 @@
...
+import io.grpc.netty.InternalProtocolNegotiator.ProtocolNegotiator;
+import io.grpc.netty.NettyChannelBuilder;
+import io.netty.channel.ChannelHandler;
+import io.netty.util.AsciiString;
+
+/**
+ * Provides client and server side gRPC {@link ProtocolNegotiator}s that use SDS to provide the SSL
+ * context.
+ */
+@Internal
```
creamsoup: FYI.you don't need to mark `@Internal` if package is internal package.

PR #6267, opened 2019-10-10
xds: SdsTrustManagerFactory implementation - uses SdsX509TrustManager

sanjaypujare:


xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
```diff
@@ -0,0 +1,98 @@
...
+    X509ExtendedTrustManager myDelegate = null;
+    if (tms != null) {
+      for (TrustManager tm : tms) {
+        if (tm instanceof X509ExtendedTrustManager) {
+          myDelegate = (X509ExtendedTrustManager) tm;
+          break;
+        }
+      }
+    }
+    sdsX509TrustManager = new SdsX509TrustManager(certContext, myDelegate);
```
creamsoup: from the code `myDelegate` can be null but the SdsX509TrustManager doesn't accept null. it should check the condition and throw exception or handle etc.

xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  @Override
+  protected void engineInit(KeyStore keyStore) throws Exception {}
+
+  @Override
+  protected void engineInit(ManagerFactoryParameters managerFactoryParameters) throws Exception {}
```
creamsoup: this can mislead, throwing exception might be better.

xds/src/test/java/io/grpc/xds/sds/trust/SdsTrustManagerFactoryTest.java
```diff
@@ -0,0 +1,116 @@
...
+      Assert.fail("no exception thrown");
+    } catch (CertificateException expected) {
+      assertThat(expected)
+          .hasMessageThat()
+          .contains("unable to find valid certification path to requested target");
+    }
+  }
+
+  @Test
+  public void badCertCheckClientTrusted_ExpectException() throws CertificateException, IOException {
```
creamsoup: _expectedException. or it should be `checkClientTrusted_badCert()` or `checkClientTrusted_badCert_throwsException()`

xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
+import javax.net.ssl.ManagerFactoryParameters;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509ExtendedTrustManager;
+
+@Internal
+public final class SdsTrustManagerFactory extends SimpleTrustManagerFactory {
```
creamsoup: we still want javadoc for internal class. it is unclear why it implements netty's class, who will consume this? by how?

xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
```diff
@@ -0,0 +1,98 @@
...
+      ks.load(/* stream= */ null, /* password= */ null);
+      int i = 1;
+      for (X509Certificate cert : certs) {
+        // note: alias lookup uses toLowerCase(Locale.ENGLISH)
+        // so our alias needs to be all lower-case and unique
+        ks.setCertificateEntry("alias" + Integer.toString(i), cert);
+        i++;
+      }
+      tmf.init(ks);
+    } catch (NoSuchAlgorithmException | KeyStoreException | IOException | CertificateException e) {
```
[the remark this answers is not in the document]
creamsoup: discussed with @dapengzhang0, since xds is not going to be used in android. it is fine to use it here.

xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
sanjaypujare: My thinking was `checkNotNull` in `SdsX509TrustManager` constructor will throw an exception. But I guess you want a more informative exception that an appropriate TrustManager was not found?

xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
sanjaypujare: The SslContextSecretVolumeSecretProvider (in the other PR) will consume this. The Netty class makes it easy to provide a custom trust manager. Hope there is no issue with that.

xds/src/test/java/io/grpc/xds/sds/trust/SdsTrustManagerFactoryTest.java
sanjaypujare: I am trying to follow a certain naming convention: 

action/condition  _  expected result

Let me know if you think it's not consistent,

sanjaypujare: resolving for now without changing names

sanjaypujare: renamed as per the unit testing naming guideline

xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
sanjaypujare: added code to catch that exception

xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
sanjaypujare: added throw exceptions in both

xds/src/test/certs/badclient.pem
```diff
@@ -0,0 +1,17 @@
+-----BEGIN CERTIFICATE-----
```
dapengzhang0: Can we reuse https://github.com/grpc/grpc-java/tree/master/testing/src/main/resources/certs directly?

sanjaypujare: That's what I had before. @creamsoup  didn't want to depend on external files/resources and I tend to agree to minimize chances of breakage

creamsoup: i am a bit against using other module's file. 
we could do that if we change the interface to accept input stream and modify the build to properly copy the resources.

dapengzhang0: It's just adding a test scope dependency of grpc-testing library and calling `getResourceAsStream()`. 
```
testCompile project('grpc-testing')
```

The grpc-testing library (available in Maven `io.grpc:grpc-testing`) is designed for testing purpose, even for external users. I don't think there is problem for sds test.

sanjaypujare: Also I would like to be able to change the files without breaking others. As I understand these files under "resources" are used by examples so let's not depend on them.

xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
```diff
@@ -0,0 +1,110 @@
...
+import java.security.cert.X509Certificate;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
+import javax.net.ssl.ManagerFactoryParameters;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509ExtendedTrustManager;
+
+@Internal
```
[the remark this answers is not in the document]
sanjaypujare: It was already done in the latest commit. PTAL

xds/src/test/certs/badclient.pem
dapengzhang0: If you need a custom cert file then you can not modify and use the resources from grpc-testing, otherwise it may break the users.

sanjaypujare: @dapengzhang0 yes, we are saying the same thing. That is one reason I don't want to use the files from grpc-testing.

Issue #6274, opened 2019-10-11
Disabled tests in io.grpc.xds.sds.trust.SdsX509TrustManagerTest need to be re-enabled


sanjaypujare:
Many tests in io.grpc.xds.sds.trust.SdsX509TrustManagerTest read from files that fail with blaze (internal failure). Need to fix by using resources

sanjaypujare:
#6273 disables the tests

creamsoup:
resolved by #6281

PR #6275, opened 2019-10-11
xds: disable test that breaks on blaze.

ran-su:
This test is included in the most recent attempt of import, and it also broken with blaze. Disabling it.


creamsoup:
related to #6274

PR #6278, opened 2019-10-14
protobuf: StatusProto#fromStatusAndTrailers fall-back use status

creamsoup:
the original intention was only generating Status proto from trailers. In actual use cases requires to handle null and derive the Status proto from status. The main reason is there are few cases when the server doesn't send trailers such as server UNREACHABLE, UNAVAILABLE or DEADLINE_EXCEEDED.

### DIFF
--- a/protobuf/src/main/java/io/grpc/protobuf/StatusProto.java
+++ b/protobuf/src/main/java/io/grpc/protobuf/StatusProto.java
@@ -148,7 +148,8 @@ public final class StatusProto {
 
   /**
    * Extracts the google.rpc.Status from trailers, and makes sure they match the gRPC
-   * {@code status}.
+   * {@code status}. If the trailers doesn't contain status entry, it uses {@code status} param to
+   * generate status.
    *
    * @return the embedded google.rpc.Status or {@code null} if it is not present.
    * @since 1.11.0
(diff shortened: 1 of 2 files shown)


protobuf/src/main/java/io/grpc/protobuf/StatusProto.java
```diff
@@ -164,6 +165,15 @@ private static Metadata toMetadata(com.google.rpc.Status statusProto, Metadata m
         return statusProto;
       }
     }
+    // fall-back to status, this is useful if the error is local. e.g. Server is unavailable.
+    if (status != null) {
```
ericgribkoff: Optional: It seems like `status` should not be null,  might be cleaner to use a precondition rather than having to put an if-block here. Although we already do a check for null trailers...should the `Metadata` param have a `@Nullable` annotation? Looking [here](https://github.com/grpc/grpc-java/blob/master/api/src/main/java/io/grpc/StatusException.java#L51) is seems like the general "convention" is that status is non-null but trailers can be null, although that doesn't have to match the convention here.

protobuf/src/main/java/io/grpc/protobuf/StatusProto.java
```diff
@@ -148,7 +148,8 @@ private static Metadata toMetadata(com.google.rpc.Status statusProto, Metadata m
 
   /**
    * Extracts the google.rpc.Status from trailers, and makes sure they match the gRPC
-   * {@code status}.
+   * {@code status}. If the trailers doesn't contain status entry, it uses {@code status} param to
```
ericgribkoff: Nit: s/doesn't/do not/

s/status entry/a google.rpc.Status/

protobuf/src/main/java/io/grpc/protobuf/StatusProto.java
```diff
@@ -148,7 +148,8 @@ private static Metadata toMetadata(com.google.rpc.Status statusProto, Metadata m
...
+   * generate status.
```
ericgribkoff: s/status/a google.rpc.Status/

protobuf/src/main/java/io/grpc/protobuf/StatusProto.java
creamsoup: I didn't do it because it is api change, but since it is experimental and you have same impression, might as well fix it. also removed `@Nullable` from return type.

PR #6285, opened 2019-10-16
api: Safe implies idempotent

ejona86:
CC @carl-mastrangelo

### DIFF
--- a/api/src/main/java/io/grpc/MethodDescriptor.java
+++ b/api/src/main/java/io/grpc/MethodDescriptor.java
@@ -224,7 +224,7 @@ public final class MethodDescriptor<ReqT, RespT> {
       boolean idempotent,
       boolean safe,
       boolean sampledToLocalTracing) {
-
+    assert !safe || idempotent : "safe should imply idempotent";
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -536,25 +536,31 @@ private Builder() {}
...
      * Sets whether this method is safe.  If true, calling this method any number of times doesn't
-     * have side effects.
+     * have side effects. If {@code true}, method is also idempotent.
      *
      * @since 1.1.0
      */
     @ExperimentalApi("https://github.com/grpc/grpc-java/issues/1775")
     public Builder<ReqT, RespT> setSafe(boolean safe) {
       this.safe = safe;
+      if (safe) {
```
carl-mastrangelo: This is a little errorprone.  Setting the method to be safe, follow by setting it to be not safe has a side effect of leaving it idempotent.  This behavior should be called out in the docs.

ejona86: Created https://github.com/grpc/grpc-java/pull/6290

Issue #5968, opened 2019-07-09
async client stub: need to ignore unary response msg if status is not OK


ejona86:
The future and blocking APIs wait until the close() before delivering any received message. If the status is an error they throw away the request.

https://github.com/grpc/grpc-java/blob/1c9345987f6501c43c748ad15b410f695a75bef3/stub/src/main/java/io/grpc/stub/ClientCalls.java#L481-L491

We need to do the same for async stub. See also https://github.com/grpc/grpc/issues/12824

ejona86:
Note that Go and C both throw away the message here. So Java async stub is the _only_ client that would see such a message.

ejona86:
This was reverted in #6342

PR #6288, opened 2019-10-16
stub: ignore unary response msg if status is not OK

ran-su:
fixes: https://github.com/grpc/grpc-java/issues/5968

### DIFF
--- a/stub/src/main/java/io/grpc/stub/ClientCalls.java
+++ b/stub/src/main/java/io/grpc/stub/ClientCalls.java
@@ -436,6 +436,7 @@ public final class ClientCalls {
       if (streamingResponse) {
         observer.onNext(message);
       } else {
+        // will send message in onClose() for unary calls.
         unaryMessage = message;
       }
 
@@ -448,8 +449,14 @@ public final class ClientCalls {
     @Override
     public void onClose(Status status, Metadata trailers) {
       if (status.isOk()) {
-        if (!streamingResponse && unaryMessage != null) {
-          observer.onNext(unaryMessage);
+        if (!streamingResponse) {
+          if (unaryMessage != null) {
+            observer.onNext(unaryMessage);
+          } else {
+            throw Status.INTERNAL
+                .withDescription("Status is OK but message is null for unary call")
+                .asRuntimeException();
+          }
         }
         observer.onCompleted();
       } else {
--- a/stub/src/main/java/io/grpc/stub/ClientCalls.java
+++ b/stub/src/main/java/io/grpc/stub/ClientCalls.java
@@ -450,13 +450,10 @@ public final class ClientCalls {
     public void onClose(Status status, Metadata trailers) {
       Throwable error = null;
       if (status.isOk()) {
-        boolean shouldCallOnCompleted = true;
         if (!streamingResponse) {
-          shouldCallOnCompleted = false;
           if (unaryMessage != null) {
             try {
               observer.onNext(unaryMessage);
-              shouldCallOnCompleted = true;
             } catch (Throwable t) {
               error = t;
             }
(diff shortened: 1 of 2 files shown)


stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -431,7 +432,12 @@ public void onMessage(RespT message) {
...
       firstResponseReceived = true;
-      observer.onNext(message);
+
+      if (streamingResponse) {
+        observer.onNext(message);
+      } else {
+        unaryMessage = message;
+      }
 
       if (streamingResponse && adapter.autoFlowControlEnabled) {
```
ejona86: Nit: could move this into the new if block

stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -442,6 +448,9 @@ public void onMessage(RespT message) {
     @Override
     public void onClose(Status status, Metadata trailers) {
       if (status.isOk()) {
+        if (!streamingResponse && unaryMessage != null) {
```
ejona86: If there is no message, it would probably be good to generate an error.

stub/src/main/java/io/grpc/stub/ClientCalls.java
ran-su: I added a comment here, I think it is more helpful than move it into a new if block.

stub/src/main/java/io/grpc/stub/ClientCalls.java
ran-su: Now it throws exception when status is OK but msg is null. A test for that is also added.

stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -442,7 +449,17 @@ public void onMessage(RespT message) {
     @Override
     public void onClose(Status status, Metadata trailers) {
       if (status.isOk()) {
-        observer.onCompleted();
+        if (!streamingResponse) {
+          if (unaryMessage != null) {
+            observer.onNext(unaryMessage);
```
ejona86: This flow doesn't call onCompleted(), but it should.

stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -442,7 +449,18 @@ public void onMessage(RespT message) {
...
```
ejona86: I only just noticed, because it is a bit subtle, but we now need to handle exceptions from `onNext()`.

Previously if onNext() threw an exception, onMessage() would propagate the exception which would cancel the call and so onClose() would be called, which would call `onError()`.

And the code will need a bit of an ugly dance. Maybe something like:
```java
boolean noException = false;
try {
  observer.onNext(unaryMessage);
  noException = true;
} catch (Throwable t) {
  observer.onError(t);
}
if (noException) {
  observer.onCompleted();
}
```

stub/src/main/java/io/grpc/stub/ClientCalls.java
```diff
@@ -441,10 +448,33 @@ public void onMessage(RespT message) {
...
+              error = t;
+            }
+          } else {
+            error = Status.INTERNAL.withDescription("Response message is null for unary call")
+                .asRuntimeException();
+          }
+        }
+
+        if (shouldCallOnCompleted) {
+          observer.onCompleted();
```
ejona86: You can move this to an else at the end of the method, which makes it clear one of onError/onCompleted will be called. You also don't need `shouldCallOnCompleted` because you can use `error` instead.

stub/src/test/java/io/grpc/stub/ClientCallsTest.java
```diff
@@ -99,6 +99,124 @@ public void tearDown() {
...
+
+      @Override
+      public void onCompleted() {
+        fail("Should not reach here");
+      }
+    };
+
+    ClientCalls.asyncUnaryCall(call, req, responseObserver);
+    assertThat(expected.size()).isEqualTo(1);
+    assertThat(expected.get(0).getMessage())
```
ejona86: Use `assertThat(expected.get(0)).hasMessageThat()` to produce a better error message when it fails. Ditto below.

PR #6290, opened 2019-10-17
api: Clarify setting safe/idempotent may change other value

ejona86:
CC @carl-mastrangelo


carl-mastrangelo:
One other thought I just had:  you may want to make setting either value to false to clear the other.   That would probably be the safest option.   Either way tho.

ejona86:
@carl-mastrangelo, that would be an option. I'm optimizing for the getters() so that idempotent and safe are consistent. Most MethodDescriptors are created from scratch, so I don't think flipping it back and forth and getting surprising results will be much of a problem.

Issue #5969, opened 2019-07-09
async server: need to ignore unary response msg if status is not OK


ejona86:
The expectation for unary responses is that the server will respond with an message or an error. If the server responds with a message and an error, the message should be thrown away. The async API currently would send the message. For unary responses, we should delay sending the response until the onComplete()/onError(). If onError() is called we should not send the message.

https://github.com/grpc/grpc-java/blob/b22017851560197a41015acd90f443f7b9519984/stub/src/main/java/io/grpc/stub/ServerCalls.java#L336-L349

See #5968 and grpc/grpc#12824

Alam-Saran:
I working on this issue and analysis is progress.

Alam-Saran:
@kannanjgithub I started my initial investigation to this defect and found there is no steps how to reproduce it at my local,Can you please suggest me the reprodicing steps of this defect.

PR #6296, opened 2019-10-17
stub: ignore unary response msg if status is not OK for async server

ran-su:
fixes: https://github.com/grpc/grpc-java/issues/5969

### DIFF
--- a/stub/src/main/java/io/grpc/stub/ServerCalls.java
+++ b/stub/src/main/java/io/grpc/stub/ServerCalls.java
@@ -314,6 +314,7 @@ public final class ServerCalls {
     private Runnable onCancelHandler;
     private boolean aborted = false;
     private boolean completed = false;
+    private RespT unaryResponse;
 
     // Non private to avoid synthetic class
     ServerCallStreamObserverImpl(ServerCall<ReqT, RespT> call) {
@@ -348,7 +349,13 @@ public final class ServerCalls {
         call.sendHeaders(new Metadata());
         sentHeaders = true;
       }
-      call.sendMessage(response);
+
+      if (!call.getMethodDescriptor().getType().serverSendsOneMessage()) {
+        call.sendMessage(response);
+      } else {
+        // delay the sendMessage() to onComplete()/onError()
+        unaryResponse = response;
+      }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


stub/src/main/java/io/grpc/stub/ServerCalls.java
```diff
@@ -368,8 +375,28 @@ public void onCompleted() {
...
+              call.sendMessage(unaryResponse);
+            } catch (Throwable t) {
+              error = t;
+            }
+          } else {
+            error = Status.INTERNAL.withDescription("Response message is null for unary call")
+                .asRuntimeException();
+          }
+          if (error != null) {
+            onError(error);
```
ejona86: This error is sent to the client. We should not send either of the two possible `error` values, as we don't want to send exceptions to clients for the application (it's bad practice that we don't want to encourage) and the other message doesn't make sense to the client.

What we should do instead is slightly unclear, but we can talk about it offline.

stub/src/main/java/io/grpc/stub/ServerCalls.java
```diff
@@ -348,7 +349,13 @@ public void onNext(RespT response) {
         call.sendHeaders(new Metadata());
         sentHeaders = true;
       }
-      call.sendMessage(response);
+
+      if (!call.getMethodDescriptor().getType().serverSendsOneMessage()) {
+        call.sendMessage(response);
+      } else {
+        // delay the sendMessage() to onComplete()/onError()
+        unaryResponse = response;
```
ST-DDT: IMO this should throw an exception if the server sends multiple responses/overwrites the previous obe.

stub/src/main/java/io/grpc/stub/ServerCalls.java
```diff
@@ -368,8 +375,28 @@ public void onCompleted() {
           throw Status.CANCELLED.withDescription("call already cancelled").asRuntimeException();
         }
       } else {
-        call.close(Status.OK, new Metadata());
-        completed = true;
+
+        if (call.getMethodDescriptor().getType().serverSendsOneMessage()) {
+          Throwable error = null;
```
ST-DDT: This error should be logged to the server log.

PR #6297, opened 2019-10-17
xds: TlsContextManager refactoring to match its usage; add tests for TlsContextManager

sanjaypujare:


xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
```diff
@@ -47,19 +46,17 @@
...
   }
 
-  /** Constructor that takes array of certs. */
-  public SdsTrustManagerFactory(
-      X509Certificate[] certs, @Nullable CertificateValidationContext certContext)
-      throws CertStoreException {
-    checkNotNull(certs, "certs");
-    createSdsX509TrustManager(certs, certContext);
+  private static String getTrustedCaFromCertContext(CertificateValidationContext certContext) {
+    return certContext != null ? certContext.getTrustedCa().getFilename() : null;
```
creamsoup: we don't handle other 2 types? (inline bytes or string)

xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
```diff
@@ -47,19 +46,17 @@
...
 
-  /** Constructor that loads certs from a file. */
-  public SdsTrustManagerFactory(
-      String certsFile, @Nullable CertificateValidationContext certContext)
+  /** Constructor constructs from a {@link CertificateValidationContext}. */
+  public SdsTrustManagerFactory(CertificateValidationContext certContext)
       throws CertificateException, IOException, CertStoreException {
-    this(CertificateUtils.toX509Certificates(new File(certsFile)), certContext);
+    checkNotNull(certContext, "certContext");
+    String certsFile = getTrustedCaFromCertContext(certContext);
```
creamsoup: this certsFile can be null but nothing is handling null case

xds/src/main/java/io/grpc/xds/sds/SslContextSecretVolumeSecretProvider.java
```diff
@@ -99,28 +105,41 @@ static TlsCertificate validateTlsCertificate(
...
+      DownstreamTlsContext downstreamTlsContext) {
+    checkNotNull(downstreamTlsContext, "downstreamTlsContext");
+    CommonTlsContext commonTlsContext = downstreamTlsContext.getCommonTlsContext();
+    TlsCertificate tlsCertificate = null;
+    if (commonTlsContext.getTlsCertificatesCount() > 0) {
+      tlsCertificate = commonTlsContext.getTlsCertificates(0);
+    }
     // first validate
     validateTlsCertificate(tlsCertificate, /* optional= */ false);
+    CertificateValidationContext certContext = commonTlsContext.getValidationContext();
```
creamsoup: this is also oneof field, how we make sure this exists? do we need to handle other 2 types?
also, the variable name is misleading, certification *validation* context is not certification context.

xds/src/test/java/io/grpc/xds/sds/TlsContextManagerTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+/** Unit tests for {@link TlsContextManager}. */
+@RunWith(JUnit4.class)
+public class TlsContextManagerTest {
+
+  private static final String SERVER_1_PEM_FILE = "server1.pem";
+  private static final String SERVER_1_KEY_FILE = "server1.key";
+  private static final String CA_PEM_FILE = "ca.pem";
+
+  @Rule public TemporaryFolder temporaryFolder = new TemporaryFolder();
```
creamsoup: this is not used, remove it?

xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
sanjaypujare: No. The checks are already there in `SslContextSecretVolumeSecretProvider.validate*` [methods](https://github.com/grpc/grpc-java/blob/master/xds/src/main/java/io/grpc/xds/sds/SslContextSecretVolumeSecretProvider.java#L66) and this is not yet used by the control plane AFAIK

xds/src/main/java/io/grpc/xds/sds/SslContextSecretVolumeSecretProvider.java
sanjaypujare: Good points. I will address both of them. Regarding oneof I thought the control plane will not send inconsistent data (one type in `TlsCertitificate` and another type for `validationContext`) but it is good to put the checks in.

xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
sanjaypujare: If you see how `SdsTrustManagerFactory(CertificateValidationContext)` is called from `SslContextSecretVolumeSecretProvider.buildSslContextFromSecrets` after going thru various `validate*` methods you will see that if `certContext` is non-null it has to have the file-name (non-null and file-name type data-source). That's why I didn't add that check but it is easy to add one if you insist.

xds/src/main/java/io/grpc/xds/sds/SslContextSecretVolumeSecretProvider.java
sanjaypujare: Fixed.

xds/src/main/java/io/grpc/xds/sds/trust/SdsTrustManagerFactory.java
creamsoup: this method is called from internal public api, you can't assume that it will be only called from `SslContextSecretVolumeSecretProvider`. also why it checks null if it cannot be null? it needs to have `@Nullable`. 
those 2 cases are completely opposite approach (one is not handling possible case with the class, the other is handling impossible case) which made me very confused.

sanjaypujare: Fixed the logic to make it consistent and clear for this class alone. Also fixed format/style

PR #6298, opened 2019-10-17
core,grpclb: DnsNameResolver will use srv record by default if grpclb

creamsoup:
The DNS name resolver will look up SRV records by default if the grpclb artifact is in the classpath.


core/src/main/java/io/grpc/internal/BaseDnsNameResolverProvider.java
```diff
@@ -0,0 +1,72 @@
...
+import io.grpc.Internal;
+import io.grpc.InternalServiceProviders;
+import io.grpc.NameResolver;
+import io.grpc.NameResolverProvider;
+import java.net.URI;
+
+/**
+ * Base provider of name resolvers for name agnostic consumption.
+ */
+@Internal
```
zhangkun83: You don't need the annotation if the class is already in the `internal` package.

core/src/main/java/io/grpc/internal/DnsNameResolverProvider.java
```diff
@@ -38,37 +31,11 @@
...
-  public String getDefaultScheme() {
-    return SCHEME;
-  }
+public final class DnsNameResolverProvider extends BaseDnsNameResolverProvider {
 
   @Override
-  protected boolean isAvailable() {
-    return true;
+  protected boolean isSrvEnabled() {
+    return Boolean.parseBoolean(System.getProperty(ENABLE_GRPCLB_PROPERTY_NAME, "false"));
```
zhangkun83: The original code parses this boolean in a static context and assign it to a static field. We should keep doing this to prevent the original behavior.

grpclb/src/main/java/io/grpc/grpclb/GrpclbNameResolverProvider.java
```diff
@@ -0,0 +1,53 @@
...
+ *   <li>{@code "dns://8.8.8.8/foo.googleapis.com:8080"} (using alternative DNS (not implemented
+ *   yet))</li>
+ *   <li>{@code "dns:///foo.googleapis.com"} (without port)</li>
+ * </ul>
+ *
+ * <p>Note: the main difference between {@code io.grpc.DnsNameResolver} is service record is enabled
+ * by default.
+ */
+@Internal
+public class GrpclbNameResolverProvider extends BaseDnsNameResolverProvider {
```
zhangkun83: There is a trick to make the provider package-private while still accessible by service loader. See https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/util/SecretRoundRobinLoadBalancerProvider.java

core/src/test/java/io/grpc/internal/DnsNameResolverProviderTest.java
```diff
@@ -46,15 +51,52 @@ public void uncaughtException(Thread t, Throwable e) {
...
+
+    assertThat(new DnsNameResolverProvider().isSrvEnabled()).isTrue();
+  }
+
+  @Test
+  public void enableSrv_defaultDisabled() {
+    System.setProperty(ENABLE_GRPCLB_PROPERTY_NAME, "false");
+
+    assertThat(new DnsNameResolverProvider().isSrvEnabled()).isFalse();
+  }
```
zhangkun83: Such test requires a behavior change (mutating the property at runtime would change the behavior of DnsNameResolvers created later, while it was not the case previously). Since the original DnsNameResolver didn't test with mutating the system property, you can probably leave it out too.

grpclb/src/main/java/io/grpc/grpclb/GrpclbNameResolverProvider.java
```diff
@@ -0,0 +1,53 @@
...
+
+  @Override
+  protected boolean isSrvEnabled() {
+    return Boolean.parseBoolean(System.getProperty(ENABLE_GRPCLB_PROPERTY_NAME, "true"));
```
zhangkun83: According to the design, GRPCLB will have SRV enabled "unconditionally". Just return "true" here, then the test can be largely simplified.

grpclb/src/test/java/io/grpc/grpclb/GrpclbNameResolverProviderTest.java
```diff
@@ -0,0 +1,69 @@
...
+import javax.annotation.Nullable;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class GrpclbNameResolverProviderTest {
+
```
zhangkun83: We should verify that this provider's priority is higher than the core provider's.

grpclb/src/main/java/io/grpc/grpclb/GrpclbNameResolverProvider.java
creamsoup: i discussed with @ejona86, we decided to make that conditional for debugging purpose etc. 
i made a local change to make it testable, but it is pretty nasty (using class loader and reflection to reevaluate class... not many lines of code but yup....) i am not sure we want to use this approach or not.

grpclb/src/test/java/io/grpc/grpclb/GrpclbNameResolverProviderTest.java
creamsoup: without using reflection it is not possible to test since the priority is protected and NameResolverProvider is hiding everything.

creamsoup:
added tests for default value, PTAL

PR #6302, opened 2019-10-18
core,grpclb: DnsNameResolver will use srv record by default if grpclb (backport of #6298)

creamsoup:


creamsoup:
manually verified (e2e) by @apolcyn

PR #6314, opened 2019-10-22
examples: adjust comment

steverao:
adjust comment in examples


creamsoup:
merged, thanks @steverao

PR #6328, opened 2019-10-23
core: delay sending cancel request on client-side when deadline expires

ran-su:
When deadline expires, both the client and the server try to cancel the stream. There is a race between server receiving the cancellation from the client and the server cancelling the stream, which changes the final status of the stream from the server's perspective. If the former wins, server sees CANCELLED. If the latter wins, server sees DEADLINE_EXCEEDED.

Because `ServerCall.Listener` doesn't pass the final status to the server-side application, this ambiguity isn't a problem in most cases. However, the status is passed to `ServerStreamTracer`, and thus will cause confusion in stats.

This PR delays the client-side cancellation in case of deadline exceeded by a small margin, thus makes DEADLINE_EXCEEDED be reported most of the time.

b/118879795


core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -68,6 +68,8 @@
   private static final Logger log = Logger.getLogger(ClientCallImpl.class.getName());
   private static final byte[] FULL_STREAM_DECOMPRESSION_ENCODINGS
       = "gzip".getBytes(Charset.forName("US-ASCII"));
+  // visible for testing
+  static final long EXTRA_WAIT_TIME_BEFORE_SEND_CANCEL_IN_NS = 1_000_000;
```
creamsoup: use `@VisibleForTesting`, can you use `TimeUnit.MILLIS.toNano(1)`? also, this deserves some explanation (can use the comment in the test).

core/src/test/java/io/grpc/internal/ClientCallImplTest.java
```diff
@@ -816,14 +817,36 @@ public void expiredDeadlineCancelsStream_CallOptions() {
...
+    // The deadline needs to be a number large enough to get encompass the call to start, otherwise
+    // the scheduled cancellation won't be created, and the call will fail early.
+    ClientCallImpl<Void, Void> call = new ClientCallImpl<>(
+        method,
+        MoreExecutors.directExecutor(),
+        baseCallOptions.withDeadline(Deadline.after(1000, TimeUnit.NANOSECONDS)),
+        provider,
+        deadlineCancellationExecutor,
+        channelCallTracer,
+        false /* retryEnabled */);
```
creamsoup: `/* retryEnabled= */ false`. fyi: linter actually checks this.

core/src/test/java/io/grpc/internal/ClientCallImplTest.java
```diff
@@ -816,14 +817,36 @@ public void expiredDeadlineCancelsStream_CallOptions() {
...
+
+    call.start(callListener, new Metadata());
+
+    fakeClock.forwardNanos(950L + EXTRA_WAIT_TIME_BEFORE_SEND_CANCEL_IN_NS /* nanoseconds */);
```
creamsoup: comment seems unnecessary

core/src/test/java/io/grpc/internal/ClientCallImplTest.java
```diff
@@ -816,14 +817,36 @@ public void expiredDeadlineCancelsStream_CallOptions() {
...
+
+    verify(stream, times(0)).cancel(statusCaptor.capture());
```
creamsoup: nit: `times(0)` -> `never()`

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -88,6 +95,8 @@
   private boolean fullStreamDecompression;
   private DecompressorRegistry decompressorRegistry = DecompressorRegistry.getDefaultInstance();
   private CompressorRegistry compressorRegistry = CompressorRegistry.getDefaultInstance();
+  private ScheduledFuture<?> deadlineCancellationNotifyApplicationFuture;
```
zhangkun83: This needs to be volatile too.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -287,6 +296,7 @@ public void runInContext() {
     }
     stream.setDecompressorRegistry(decompressorRegistry);
     channelCallsTracer.reportCallStarted();
+    this.observer = observer;
```
zhangkun83: You could instead pass the observer to `startDeadlineNotifyApplicationTimer` so that the timer keeps a reference to the observer. Passing the object as argument is better than saving it as a mutable field, because it's fewer states.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -333,37 +346,55 @@ private static void logIfContextNarrowedTimeout(
...
     if (f != null) {
       f.cancel(false);
     }
   }
 
-  private class DeadlineTimer implements Runnable {
+  private ScheduledFuture<?> startDeadlineSendCancelToServerTimer(
+      long remainingNanos, InsightBuilder insight) {
+    return deadlineCancellationExecutor.schedule(
+        new LogExceptionRunnable(new DeadlineSendCancelToServerTimer(remainingNanos, insight)),
```
zhangkun83: To make the code easier to follow and reduce the fragmentation, consider moving the whole `DeadlineSendCancelToServerTimer` into this method body (probably also eliminating the arguments to the timer constructor since it's now a closure).

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -333,37 +346,55 @@ private static void logIfContextNarrowedTimeout(
...
+        DEADLINE_EXPIRATION_CANCEL_DELAY_NANOS,
+        TimeUnit.NANOSECONDS);
+  }
+
+  private ScheduledFuture<?> startDeadlineNotifyApplicationTimer(Deadline deadline) {
+    long remainingNanos = deadline.timeRemaining(TimeUnit.NANOSECONDS);
+    return deadlineCancellationExecutor.schedule(
```
zhangkun83: ditto

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -506,8 +537,25 @@ public Attributes getAttributes() {
...
+    public void run() {
+      InsightBuilder insight = new InsightBuilder();
+      stream.appendTimeoutInsight(insight);
+      deadlineCancellationSendToServerFuture =
+          startDeadlineSendCancelToServerTimer(remainingNanos, insight);
+      // DelayedStream.cancel() is safe to call from a thread that is different from where the
+      // stream is created.
+      closeObserver(observer, DEADLINE_EXCEEDED.augmentDescription(
+          "deadline exceeded after " + remainingNanos + "ns. " + insight), new Metadata());
+      observerClosed = true;
```
zhangkun83: This line can be moved into `closeObserver()`, so that it becomes self-guarded.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -506,8 +537,25 @@ public Attributes getAttributes() {
...
+
+    @Override
+    public void run() {
+      InsightBuilder insight = new InsightBuilder();
+      stream.appendTimeoutInsight(insight);
+      deadlineCancellationSendToServerFuture =
+          startDeadlineSendCancelToServerTimer(remainingNanos, insight);
+      // DelayedStream.cancel() is safe to call from a thread that is different from where the
+      // stream is created.
+      closeObserver(observer, DEADLINE_EXCEEDED.augmentDescription(
```
zhangkun83: `closeObserver()` must be called from `callExecutor` and inside a `ContextRunnable`. You need something similar to `class StreamClosed()` in this file.

core/src/test/java/io/grpc/internal/ClientCallImplTest.java
```diff
@@ -812,18 +817,52 @@ public void expiredDeadlineCancelsStream_CallOptions() {
...
+
+    // verify cancel send to server is delayed with DEADLINE_EXPIRATION_CANCEL_DELAY
     verify(stream, times(1)).cancel(statusCaptor.capture());
     assertEquals(Status.Code.DEADLINE_EXCEEDED, statusCaptor.getValue().getCode());
     assertThat(statusCaptor.getValue().getDescription())
         .matches("deadline exceeded after [0-9]+ns. \\[remote_addr=127\\.0\\.0\\.1:443\\]");
   }
 
+  @Test
+  public void expiredDeadlineWithoutExtraWaitingDoNotReceiveCancel() {
```
zhangkun83: What's covered here could be covered in `expiredDeadlineCancelsStream_CallOptions()`, where you forward the clock by `DEADLINE_EXPIRATION_CANCEL_DELAY_NANOS - 1` and verify that the stream is not cancelled, then forward by 1 and verify that the stream is now cancelled.

core/src/test/java/io/grpc/internal/ClientCallImplTest.java
```diff
@@ -812,18 +817,52 @@ public void expiredDeadlineCancelsStream_CallOptions() {
...
+
+    call.start(callListener, new Metadata());
+
+    fakeClock.forwardNanos(950L + DEADLINE_EXPIRATION_CANCEL_DELAY_NANOS);
+
+    verify(stream, never()).cancel(statusCaptor.capture());
+  }
+
   @Test
   public void expiredDeadlineCancelsStream_Context() {
```
zhangkun83: This case should see the delay too. To implement that, you need to change `ContextCancellationListener` and check if `deadline.isExpired()`. If true, schedule deadlineCancellationSendToServerFuture; otherwise, cancel the stream immediately.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -118,8 +127,27 @@
...
-    public void cancelled(Context context) {
-      stream.cancel(statusFromCancelled(context));
+    public void cancelled(final Context context) {
+      if (streamCancelledByDeadline) {
+        return;
+      }
+      if (context.getDeadline() == null || !context.getDeadline().isExpired()) {
+        stream.cancel(statusFromCancelled(context));
+      } else {
+        if (deadlineCancellationSendToServerFuture != null) {
```
zhangkun83: I think we can just leave the `deadlineCancellationSendToServerFuture` and let it do the job.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -88,6 +93,10 @@
   private boolean fullStreamDecompression;
   private DecompressorRegistry decompressorRegistry = DecompressorRegistry.getDefaultInstance();
   private CompressorRegistry compressorRegistry = CompressorRegistry.getDefaultInstance();
+  private volatile ScheduledFuture<?> deadlineCancellationNotifyApplicationFuture;
+  private volatile ScheduledFuture<?> deadlineCancellationSendToServerFuture;
+  private boolean observerClosed = false;
+  private volatile boolean streamCancelledByDeadline = false;
```
zhangkun83: `stream.cancel()` is safe to be called more than once. This boolean is unnecessary.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -3488,6 +3488,10 @@ protected ClientTransportFactory buildTransportFactory() {
       assertThat(Throwables.getStackTraceAsString(e.getCause())).contains("deadline");
     }
 
+    // run the delayed cancel request task for deadline exceeded.
+    timer.forwardTime(5, TimeUnit.SECONDS);
```
zhangkun83: This test method is not concerned about the delayed cancellation of deadline. These two added lines are not needed.

core/src/test/java/io/grpc/internal/ClientCallImplTest.java
```diff
@@ -808,17 +813,28 @@ public void expiredDeadlineCancelsStream_CallOptions() {
...
 
-    verify(stream, times(1)).cancel(statusCaptor.capture());
+    // Verify cancel sent to application when deadline just past
+    verify(callListener).onClose(statusCaptor.capture(), metadataArgumentCaptor.capture());
+    assertThat(statusCaptor.getValue().getDescription())
+        .matches("deadline exceeded after [0-9]+ns. \\[remote_addr=127\\.0\\.0\\.1:443\\]");
+    assertThat(statusCaptor.getValue().getCode()).isEqualTo(Code.DEADLINE_EXCEEDED);
+    verify(stream, never()).cancel(statusCaptor.capture());
+
+    fakeClock.forwardNanos(DEADLINE_EXPIRATION_CANCEL_DELAY_NANOS);
```
zhangkun83: To make sure that the delay is not shorter than `DEADLINE_EXPIRATION_CANCEL_DELAY_NANOS`, we'd need to
```java
fakeClock.forwardNanos(DEADLINE_EXPIRATION_CANCEL_DELAY_NANOS - 1);
verify(stream, never()).cancel(any(Status.class));
fakeClock.forwardNanos(1);
verify(stream).cancel(statusCaptor.capture());
```

core/src/test/java/io/grpc/internal/ClientCallImplTest.java
```diff
@@ -839,14 +855,16 @@ public void expiredDeadlineCancelsStream_Context() {
...
 
     context.detach(origContext);
 
     call.start(callListener, new Metadata());
 
-    fakeClock.forwardNanos(TimeUnit.SECONDS.toNanos(1) + 1);
+    fakeClock.forwardTime(1000, TimeUnit.MILLISECONDS);
+    verify(stream, never()).cancel(statusCaptor.capture());
 
+    fakeClock.forwardNanos(DEADLINE_EXPIRATION_CANCEL_DELAY_NANOS);
```
zhangkun83: ditto

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
ran-su: In the `@After`, it actually verifies if `timer.getDueTasks()` is empty. Since we are adding a new delayed task(send cancel to server), we need to let the task run.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -118,8 +126,22 @@
...
   private final class ContextCancellationListener implements CancellationListener {
     @Override
-    public void cancelled(Context context) {
-      stream.cancel(statusFromCancelled(context));
+    public void cancelled(final Context context) {
+      if (context.getDeadline() == null || !context.getDeadline().isExpired()) {
+        stream.cancel(statusFromCancelled(context));
+      } else {
+        if (deadlineCancellationSendToServerFuture == null) {
+          deadlineCancellationSendToServerFuture = deadlineCancellationExecutor.schedule(
```
ejona86: I don't see anything here that triggers the application to be called immediately when the deadline is exceeded. Won't this delay all deadline notifications to the application by 1 second?

ran-su: All notifications to the application are sent in `startDeadlineNotifyApplicationTimer()`, which follows the initial deadline.

zhangkun83: In the case calloption's deadline equals to the context deadline, `startDeadlineNotifyApplicationTimer()` is not called when the ClientCall is started.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -298,8 +322,15 @@ public void runInContext() {
...
-        && deadlineCancellationExecutor != null) {
-      deadlineCancellationFuture = startDeadlineTimer(effectiveDeadline);
+        && deadlineCancellationExecutor != null
+        // if already expired deadline let failing stream handle
+        && !(stream instanceof FailingClientStream)) {
+      deadlineCancellationNotifyApplicationFuture =
+          startDeadlineNotifyApplicationTimer(effectiveDeadline, observer);
+    } else {
+      if (cancellationListener instanceof ClientCallImpl.ContextCancellationListener) {
+        ((ContextCancellationListener) cancellationListener).setObserver(observer);
```
zhangkun83: You probably need to move line 320 (`context.addListener()`) to after this else-block, so that if the deadline is exceeded immediately, observer can still be notified.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -117,9 +126,23 @@
   }
 
   private final class ContextCancellationListener implements CancellationListener {
+    private Listener<RespT> observer;
```
zhangkun83: Annotate it as nullable, and add a comment saying that this observer is non-null only if we want to delay the cancellation of stream in case of deadline-exceeded.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -117,9 +126,23 @@
...
-    public void cancelled(Context context) {
-      stream.cancel(statusFromCancelled(context));
+    public void cancelled(final Context context) {
+      if (context.getDeadline() == null || !context.getDeadline().isExpired()) {
+        stream.cancel(statusFromCancelled(context));
+      } else {
+        if (observer != null) {
+          Status status = statusFromCancelled(context);
+          deadlineCancellationSendToServerFuture = startDeadlineSendCancelToServerTimer(status);
+          executeCloseObserverInContext(observer, status);
```
zhangkun83: These two lines are repeated in `DeadlineExceededNotifyApplicationTimer`. Since this is the core logic of delayed cancellation, it should be in its own method (`delayedCancelOnDeadlineExceeded(status, observer)`), and the two cases just call it.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -377,45 +382,51 @@ private void removeContextListenerAndCancelDeadlineFuture() {
...
-          @Override
-          public void runInContext() {
-            closeObserver(observer, status, new Metadata());
-          }
-        }
+    return DEADLINE_EXCEEDED.augmentDescription(buf.toString());
+  }
 
-        callExecutor.execute(new CloseInContext());
+  private void executeCloseObserverInContext(final Listener<RespT> observer, final Status status) {
```
zhangkun83: There are other two call sites of `closeObserver()` at the start of `startInternal()` that can be simplified by calling this method.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -333,46 +340,97 @@ private static void logIfContextNarrowedTimeout(
...
-      buf.append("deadline exceeded after ");
-      if (remainingNanos < 0) {
-        buf.append('-');
+    return deadlineCancellationExecutor.schedule(
+        new LogExceptionRunnable(new DeadlineExceededNotifyApplicationTimer()),
+        remainingNanos,
+        TimeUnit.NANOSECONDS);
+  }
+
+  private Status buildDeadlineExceededStatusWithRemainingNanos(long remainingNanos) {
```
zhangkun83: nit: this method is used only once, and the only caller is very short. It may be more readable as an inline codeblock at the call site.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -333,46 +340,97 @@ private static void logIfContextNarrowedTimeout(
...
-  private ScheduledFuture<?> startDeadlineTimer(Deadline deadline) {
-    long remainingNanos = deadline.timeRemaining(TimeUnit.NANOSECONDS);
+  private void closeObserver(Listener<RespT> observer, Status status, Metadata trailers) {
+    if (!observerClosed) {
+      observerClosed = true;
+      observer.onClose(status, trailers);
+    }
+  }
+
+  private ScheduledFuture<?> startDeadlineSendCancelToServerTimer(final Status status) {
```
zhangkun83: nit: now that there is only one caller of this method, it would be more readable to merge it into its caller.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -193,6 +218,7 @@ public void start(Listener<RespT> observer, Metadata headers) {
     }
   }
 
+  @SuppressWarnings("unchecked")
```
ejona86: Why is this necessary?

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -117,9 +127,24 @@
...
+    }
+
     @Override
-    public void cancelled(Context context) {
-      stream.cancel(statusFromCancelled(context));
+    public void cancelled(final Context context) {
+      if (context.getDeadline() == null || !context.getDeadline().isExpired()) {
+        stream.cancel(statusFromCancelled(context));
+      } else {
+        if (observer != null) {
```
ejona86: I'm not very comfortable with this doing nothing if observer is null. It looks like an optimization, but it requires multiple pieces to behave in a particular way. I think it would be better to add the optimization to `delayedCancelOnDeadlineExceeded()`: if `deadlineCancellationSendToServerFuture` is already non-null (the method has already been called), return early. Note there's a race there, but there's a race there already (and so I'll have another comment about it elsewhere).

Changing this would also mean we could pass the `Listener` in the constructor and get rid of `setObserver()`.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -333,46 +340,92 @@ private static void logIfContextNarrowedTimeout(
...
+    class DeadlineExceededSendCancelToServerTimer implements Runnable {
+      @Override
+      public void run() {
+        // DelayedStream.cancel() is safe to call from a thread that is different from where the
+        // stream is created.
+        stream.cancel(status);
+      }
+    }
+
+    deadlineCancellationSendToServerFuture =  deadlineCancellationExecutor.schedule(
```
ejona86: Setting this races with `removeContextListenerAndCancelDeadlineFuture()`. We have two options: 1) let it race. Since we control `DEADLINE_EXPIRATION_CANCEL_DELAY_NANOS` we know the event will fire relatively soon (which is not the case for normal deadlines). 2) Make it thread-safe.

I'm fine with (1), but we should have a comment mentioning the race and that we are okay with it.

core/src/main/java/io/grpc/internal/ClientCallImpl.java
```diff
@@ -293,14 +289,18 @@ public void runInContext() {
...
         // If the channel has been terminated, we don't need to schedule an extra task.
-        && deadlineCancellationExecutor != null) {
-      deadlineCancellationFuture = startDeadlineTimer(effectiveDeadline);
+        && deadlineCancellationExecutor != null
+        // if already expired deadline let failing stream handle
+        && !(stream instanceof FailingClientStream)) {
+      deadlineCancellationNotifyApplicationFuture =
+          startDeadlineNotifyApplicationTimer(effectiveDeadline, observer);
     }
+    cancellationListener = new ContextCancellationListener(observer);
```
ejona86: I don't think this is thread-safe, since this field is read in `removeContextListenerAndCancelDeadlineFuture()`. I think you need to just move this initialization up higher before stream.start().

Issue #6300, opened 2019-10-18
Race condition causing NPE with LocalChannelProvider and GrpcCallContext with timeout


olavloite:
There seems to be a race condition causing a `NullPointerException` when using an `InProcessServer`, a `LocalChannelProvider` and a `GrpcCallContext` with a very short timeout.

The following example call will (often) cause the problem:

```java
GrpcCallContext context = GrpcCallContext.createDefault().withTimeout(Duration.ofMillis(1L));
client
    .getOperationCallable()
    .futureCall(GetOperationRequest.getDefaultInstance(), context)
    .get();
```

Assuming that the (mock) server will take more than 1 millisecond to respond, the expected outcome should be a `DeadlineExceededException` wrapped in an `ExecutionException`. Instead, a `NullPointerException` is logged and the call hangs indefinitely.

The exception that is logged is:

```
Oct 18, 2019 3:07:03 PM io.grpc.Context$ExecutableListener deliver
INFO: Exception notifying context listener
java.lang.NullPointerException
	at io.grpc.inprocess.InProcessTransport$InProcessStream$InProcessClientStream.internalCancel(InProcessTransport.java:746)
	at io.grpc.inprocess.InProcessTransport$InProcessStream$InProcessClientStream.serverClosed(InProcessTransport.java:683)
	at io.grpc.inprocess.InProcessTransport$InProcessStream$InProcessClientStream.access$1700(InProcessTransport.java:619)
	at io.grpc.inprocess.InProcessTransport$InProcessStream$InProcessServerStream.cancel(InProcessTransport.java:563)
	at io.grpc.internal.ServerImpl$ServerTransportListenerImpl$1ServerStreamCancellationListener.cancelled(ServerImpl.java:596)
	at io.grpc.Context$ExecutableListener.run(Context.java:1005)
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:30)
	at io.grpc.Context$ExecutableListener.deliver(Context.java:997)
	at io.grpc.Context.addListener(Context.java:479)
	at io.grpc.internal.ServerImpl$ServerTransportListenerImpl.createContext(ServerImpl.java:601)
	at io.grpc.internal.ServerImpl$ServerTransportListenerImpl.streamCreatedInternal(ServerImpl.java:501)
	at io.grpc.internal.ServerImpl$ServerTransportListenerImpl.streamCreated(ServerImpl.java:476)
[...]

PR #6333, opened 2019-10-24
core: delay attaching cancellation listener to context on server-side.

ran-su:
fixes: https://github.com/grpc/grpc-java/issues/6300

With these change, it passes the reproducer in that issue.

### DIFF
--- a/core/src/main/java/io/grpc/internal/ServerImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerImpl.java
@@ -554,9 +554,19 @@ public final class ServerImpl extends io.grpc.Server implements InternalInstrume
               return;
             }
             listener = startCall(stream, methodName, method, headers, context, statsTraceCtx, tag);
-
-            // Delay cancellation to after startCall().
-            // related issue: https://github.com/grpc/grpc-java/issues/6300
+          } catch (RuntimeException e) {
+            stream.close(Status.fromThrowable(e), new Metadata());
+            context.cancel(null);
+            throw e;
+          } catch (Error e) {
+            stream.close(Status.fromThrowable(e), new Metadata());
+            context.cancel(null);
+            throw e;
+          } finally {
+            jumpListener.setListener(listener);
+            // An extremely short deadline may expire before the stream listener is set. This causes
+            // NPE as in issue: https://github.com/grpc/grpc-java/issues/6300
+            // Delay the setting of listener will fix the NPE.
             final class ServerStreamCancellationListener implements Context.CancellationListener {
               @Override
               public void cancelled(Context context) {
@@ -570,16 +580,6 @@ public final class ServerImpl extends io.grpc.Server implements InternalInstrume
             }
 
             context.addListener(new ServerStreamCancellationListener(), directExecutor());
-          } catch (RuntimeException e) {
-            stream.close(Status.fromThrowable(e), new Metadata());
-            context.cancel(null);
-            throw e;
-          } catch (Error e) {
-            stream.close(Status.fromThrowable(e), new Metadata());
-            context.cancel(null);
-            throw e;
-          } finally {
-            jumpListener.setListener(listener);
           }
         }
       }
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -553,8 +553,23 @@ private void runInternal() {
               context.cancel(null);
               return;
             }
-            listener =
-                startCall(stream, methodName, method, headers, context, statsTraceCtx, tag);
+            listener = startCall(stream, methodName, method, headers, context, statsTraceCtx, tag);
+
+            // Delay cancellation to after startCall().
```
ejona86: I don't think this is right. It just needs to be after the `stream.setListener()` (which means also after jumpListener.setListener() which is described in the comment above).

I will say this comment isn't all that helpful as-is because it doesn't explain _why_ delaying is important.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -565,6 +564,22 @@ private void runInternal() {
             throw e;
           } finally {
             jumpListener.setListener(listener);
+            // An extremely short deadline may expire before the stream listener is set. This causes
```
ejona86: Note: this can be _anywhere_ in this method, although after the `try {` makes it easier to reason about.

It's a bit weird to be in the `finally`, since it isn't essential it is run. Could we put it after the finally block or within the try (e.g., after `startCall()`)?

core/src/test/java/io/grpc/internal/ServerImplTest.java
```diff
@@ -1030,6 +1032,28 @@ public void cancelled(Context context) {
...
+  @Test
+  public void testContextExpiredBeforeStreamCreate_StreamCancelNotCalledBeforeSetListener()
+      throws Exception {
+    AtomicBoolean contextCancelled = new AtomicBoolean(false);
+    AtomicReference<Context> context = new AtomicReference<>();
+    AtomicReference<ServerCall<String, Integer>> callReference
+        = new AtomicReference<>();
+
+    testStreamClose_setup(callReference,
+        context, contextCancelled, 0L);
```
ejona86: unwrap line.

core/src/test/java/io/grpc/internal/ServerImplTest.java
```diff
@@ -1030,6 +1032,28 @@ public void cancelled(Context context) {
...
+
+    // This assert that stream.setListener() is called before stream.cancel(), which avoid every
+    // short deadlines causing NPEs.
+    InOrder inOrder = inOrder(stream);
+    inOrder.verify(stream).setListener(any(ServerStreamListener.class));
+    inOrder.verify(stream).cancel(statusCaptor.capture());
+
+    assertThat(
+        statusCaptor.getValue().asException()).hasMessageThat().contains("context timed out");
```
ejona86: Please wrap before the hasMessageThat(), since then it keeps the argument to assertThat together.

core/src/test/java/io/grpc/internal/ServerImplTest.java
```diff
@@ -1030,6 +1032,28 @@ public void cancelled(Context context) {
     return streamListener;
   }
 
+  @Test
+  public void testContextExpiredBeforeStreamCreate_StreamCancelNotCalledBeforeSetListener()
+      throws Exception {
+    AtomicBoolean contextCancelled = new AtomicBoolean(false);
+    AtomicReference<Context> context = new AtomicReference<>();
+    AtomicReference<ServerCall<String, Integer>> callReference
+        = new AtomicReference<>();
```
ejona86: Unwrap line.

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -566,14 +565,31 @@ private void runInternal() {
           } finally {
             jumpListener.setListener(listener);
           }
+
+          // An extremely short deadline may expire before the stream listener is set. This causes
```
ejona86: This comment could be seen as misleading. Try to make it clear we are talking about `stream.setListener(jumpListener)` and not `jumpListener.setListener(listener)`

PR #6338, opened 2019-10-25
xds: reference counting map implementation for TlsContextManager and SslProviders

sanjaypujare:


creamsoup:
Discussed offline. We agreed that `ReferenceCountingMap`(RCM) is very similar to `SharedResourceHolder`(SRH) except 

- RCM wants to use hash based look up of the value in the `ResourceDefinition` (factory class) to resolve key whereas the SRH is using reference based lookup of the `Resource`(factory/key class).

There are few ways to achieve this. 

- Just using `Map` and take an instance of map in private constructor. Factory method for SRH can decide which map implementation to use. 
- Could have a base class of `SharedObjectHolder` and SRH and RCM can extends the `SharedObjectHolder`.
- SRH can simply use `HashMap` which shouldn't change the behavior because all the implementation of `Resource`s are not implementing the `equals` and `hashCode`. 
- Have two very similar implementations

the main purpose of this discussion is DRY. @ejona86,  WDYT?

sanjaypujare:
I agree we should avoid code duplication. I prefer to go with option 1:

> * Just using `Map` and take an instance of map in private constructor. Factory method for SRH can decide which map implementation to use.

we can modify `SharedResourceHolder` without affecting the current clients and the new client can use the new constructor (or another factory method) to get another instance of the SRH.

sanjaypujare:
@ejona86 @creamsoup once we agree on the way forward I will close/update this PR. Thanks

ejona86:
Do not use SharedResourceHolder for this problem. SharedResourceHolder's primary goal is to _delay_ releasing a resource when it is no longer used, in case it will be used immediately again.

That does not apply here, as we are going to be having the LB keep a reference over a long period of time. So I would just trash the entire "wait a second" behavior. That simplifies things considerably.

sanjaypujare:
That's what I initially thought (about not reusing `SharedResourceHolder`) but @creamsoup's suggestion was to refactor `SharedResourceHolder` to reuse the code for this PR's use-case. I am assuming you are not in favor of that (even if I can disable "wait a second" behavior with refactored `SharedResourceHolder` class) - just checking.

ejona86:
I agree that given what this code is trying to do, SharedResourceHolder looks helpful. But I disagreed with what the code was trying to do; I don't think the code should include a grace period after releasing. Resources can just be destroyed immediately.

ejona86:
If you disable the "wait for a second" with SharedResourceHolder, then it would become ~5 LOC which defeats the point.

We already have an abstraction if you need one: `ObjectPool`. But I honestly don't see any reason to try to reuse code here; the other usages are different enough to be unhelpful given the small amount of code you would need to share.

sanjaypujare:
@ejona86 got it. So I will update this PR to remove the "wait for a second" behavior so the review can  continue. @creamsoup ?

creamsoup:
yes as discussed offline. we want to make a `ReferenceCountingSecretProviderMap` which is not aiming general purpose reference counting map.

sanjaypujare:
closing in favor of #6350

PR #6341, opened 2019-10-27
core: use seconds in deadline exceeded string

carl-mastrangelo:
This matches the format in `Deadline.toString`


carl-mastrangelo:
@dapengzhang0 @creamsoup friendly ping.

PR #6342, opened 2019-10-28
Revert "stub: ignore unary response msg if status is not OK (#6288)"

ran-su:
This reverts commit fe46edac
The previous PR broke some internal tests. Revert this to un-block import.


ejona86:
To be clear, this broke tests that were testing unsupported cases. So the verification worked. But we're going to try harder to find broken tests/code and then re-enable.

ejona86:
#5968 was not properly re-opened when this was reverted, so now we've gone a year without this being addressed.

PR #6350, opened 2019-10-29
xds: reference counting map for SslContextProvider needed for TlsContextManager

sanjaypujare:
closes #6338

### DIFF
--- a/xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
+++ b/xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
@@ -16,10 +16,11 @@
 
 package io.grpc.xds.sds;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.common.base.Preconditions;
 import java.util.HashMap;
+import java.util.Map;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
@@ -35,12 +36,11 @@ import javax.annotation.concurrent.ThreadSafe;
 @ThreadSafe
 final class ReferenceCountingSslContextProviderMap<K> {
 
-  private final HashMap<K, Instance> instances;
+  private final Map<K, Instance> instances = new HashMap<>();
   private final SslContextProviderFactory<K> sslContextProviderFactory;
 
   ReferenceCountingSslContextProviderMap(SslContextProviderFactory<K> sslContextProviderFactory) {
     checkNotNull(sslContextProviderFactory, "sslContextProviderFactory");
-    instances = new HashMap<>();
     this.sslContextProviderFactory = sslContextProviderFactory;
   }
 
@@ -78,20 +78,16 @@ final class ReferenceCountingSslContextProviderMap<K> {
       instance = new Instance(sslContextProviderFactory.createSslContextProvider(key));
       instances.put(key, instance);
     }
-    instance.refcount++;
-    return instance.sslContextProvider;
+    return instance.acquire();
   }
 
   private synchronized SslContextProvider releaseInternal(
       final K key, final SslContextProvider instance) {
     final Instance cached = instances.get(key);
-    if (cached == null) {
-      throw new IllegalArgumentException("No cached instance found for " + key);
-    }
-    Preconditions.checkArgument(
+    checkArgument(cached != null, "No cached instance found for " + key);
+    checkArgument(
         instance == cached.sslContextProvider, "Releasing the wrong instance");
-    cached.refcount--;
-    if (cached.refcount == 0) {
+    if (cached.release()) {
       try {
         cached.sslContextProvider.close();
       } finally {
@@ -109,11 +105,22 @@ final class ReferenceCountingSslContextProviderMap<K> {
 
   private static class Instance {
     final SslContextProvider sslContextProvider;
-    int refcount;
+    private int refCount;
+
+    /** Increment refCount and acquire a reference to sslContextProvider. */
+    SslContextProvider acquire() {
+      refCount++;
+      return sslContextProvider;
+    }
+
+    /** Decrement refCount and return true if it has reached 0. */
+    boolean release() {
+      return --refCount == 0;
+    }
 
     Instance(SslContextProvider sslContextProvider) {
       this.sslContextProvider = sslContextProvider;
-      this.refcount = 0;
+      this.refCount = 0;
     }
   }
 }
--- a/xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
+++ b/xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright 2019 The gRPC Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.xds.sds;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.base.Preconditions;
+import java.util.HashMap;
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * A map for managing {@link SslContextProvider}s as reference-counted shared resources.
+ *
+ * <p>A key (of generic type K) identifies a {@link SslContextProvider}. The map also depends on a
+ * factory {@link SslContextProviderFactory} to create a new instance of {@link SslContextProvider}
+ * as needed. {@link SslContextProvider}s are ref-counted and closed by calling {@link
+ * SslContextProvider#close()} when ref-count reaches zero.
+ *
+ * @param <K> Key type for the map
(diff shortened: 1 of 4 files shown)


xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
```diff
@@ -0,0 +1,119 @@
...
+  private synchronized SslContextProvider releaseInternal(
+      final K key, final SslContextProvider instance) {
+    final Instance cached = instances.get(key);
+    if (cached == null) {
+      throw new IllegalArgumentException("No cached instance found for " + key);
+    }
+    Preconditions.checkArgument(
+        instance == cached.sslContextProvider, "Releasing the wrong instance");
+    cached.refcount--;
+    if (cached.refcount == 0) {
```
creamsoup: this logic can be in the Instance. and the map just need to call release which returns boolean. if it is true, remove from map. (similar to netty approach)

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
```diff
@@ -0,0 +1,119 @@
...
+      try {
+        cached.sslContextProvider.close();
+      } finally {
+        instances.remove(key);
+      }
+    }
+    // Always return null
+    return null;
```
creamsoup: if it always returns null, why not make it void?

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /** A factory to create an SslContextProvider from the given key. */
+  public interface SslContextProviderFactory<K> {
+    SslContextProvider createSslContextProvider(K key);
+  }
+
+  private static class Instance {
+    final SslContextProvider sslContextProvider;
+    int refcount;
```
creamsoup: this is not thread safe

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+    instance.refcount++;
+    return instance.sslContextProvider;
+  }
+
+  private synchronized SslContextProvider releaseInternal(
+      final K key, final SslContextProvider instance) {
+    final Instance cached = instances.get(key);
+    if (cached == null) {
+      throw new IllegalArgumentException("No cached instance found for " + key);
```
creamsoup: it should be IllegalStateException, you can simplify it by
`checkState(cache != null, "msg");`

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+    Preconditions.checkArgument(
+        instance == cached.sslContextProvider, "Releasing the wrong instance");
+    cached.refcount--;
+    if (cached.refcount == 0) {
+      try {
+        cached.sslContextProvider.close();
```
creamsoup: i am not sure why we want to call close?

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
```diff
@@ -0,0 +1,119 @@
...
+    SslContextProvider createSslContextProvider(K key);
+  }
+
+  private static class Instance {
+    final SslContextProvider sslContextProvider;
+    int refcount;
+
+    Instance(SslContextProvider sslContextProvider) {
+      this.sslContextProvider = sslContextProvider;
+      this.refcount = 0;
```
creamsoup: this should set to 1. we should treat the refcount==0 as impossible condition. also, `refCount` instead of `refcount`.

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
creamsoup: this is wrong because it is guarded by synchronized block.

sanjaypujare: what is wrong? The comment that "this is not thread safe"?

creamsoup: yes, i meant my comment.

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
```diff
@@ -0,0 +1,119 @@
...
+ */
+@ThreadSafe
+final class ReferenceCountingSslContextProviderMap<K> {
+
+  private final HashMap<K, Instance> instances;
+  private final SslContextProviderFactory<K> sslContextProviderFactory;
+
+  ReferenceCountingSslContextProviderMap(SslContextProviderFactory<K> sslContextProviderFactory) {
+    checkNotNull(sslContextProviderFactory, "sslContextProviderFactory");
+    instances = new HashMap<>();
```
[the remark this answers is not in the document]
sanjaypujare: Any advantage of declaring it as `Map`? I can move the initialization to L38 but note there is only 1 constructor so it's kind of moot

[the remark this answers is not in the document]
creamsoup: it is recommended by effective java. Refer to objects by their interfaces (go/ej3e-64)

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
sanjaypujare: If you use the pattern 
```
instance = map.release(key, instance);
```
it will assign `null` to `instance` so there is no chance of accidentally reusing instance when its refcount has reached 0. But I see that it is always returning null (which probably also exists in `SharedResourceHolder`?). I think it's a bug (@ejona86 ?)

sanjaypujare: resolving it based on my response...

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
sanjaypujare: made the change to `checkArgument` but still treat it as a "bad argument" error

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
sanjaypujare: PTAL and see if you still want to see refCount initialized to 1.

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
creamsoup: either one(IAE or ISE) is fine and arguable. iirc checkArgument you can use errorMessage template. 
`checkArgument(cached != null, "No cached instance found for %s", key);`

sanjaypujare: done

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
creamsoup: object being created implies that it has refCount 1. if it is initially refCount = 0, it means it should be released. overall the layering (or separate of concern) is not clear (what the instant supposed to do is in the outside such as mutating refCount).

sanjaypujare: couple of points:
- the refCount manipulation is now completely internal to the `Instance` class as per your suggestion so that part is addressed.
- the refCount logic is copied from `SharedResourceHolder` class and looked okay to me. Of course I can try to address your concern and see if I can initialize it to 1 by eliminating the increment in the first `get`.

sanjaypujare: So refCount starts at 1 now.

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
```diff
@@ -0,0 +1,128 @@
...
+
+    /** Increment refCount and acquire a reference to sslContextProvider. */
+    SslContextProvider acquire() {
+      refCount++;
+      return sslContextProvider;
+    }
+
+    /** Decrement refCount and return true if it has reached 0. */
+    boolean release() {
+      return --refCount == 0;
```
creamsoup: nit: assert if refcount < 0 after this.

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
```diff
@@ -0,0 +1,128 @@
...
+  ReferenceCountingSslContextProviderMap(SslContextProviderFactory<K> sslContextProviderFactory) {
+    checkNotNull(sslContextProviderFactory, "sslContextProviderFactory");
+    this.sslContextProviderFactory = sslContextProviderFactory;
+  }
+
+  /**
+   * Gets an existing instance of {@link SslContextProvider}. If it doesn't exist, creates a new one
+   * using the provided {@link SslContextProviderFactory&lt;K&gt;}
+   */
+  public SslContextProvider get(K key) {
```
creamsoup: nit: would be nice to annotate `@CheckReturnValue`

PR #6358, opened 2019-10-30
core: Set jndi connection timeout and retires to default.

jhaals:
The current DirContext properties are for LDAP and does not affect the
actual DNS settings. This surfaces the actual properties for tuning
connection timeout and retries. Values are set to the default in
[openJDK](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/jdk.naming.dns/share/classes/com/sun/jndi/dns/DnsContext.java#L70)

### DIFF
--- a/core/src/main/java/io/grpc/internal/JndiResourceResolverFactory.java
+++ b/core/src/main/java/io/grpc/internal/JndiResourceResolverFactory.java
@@ -249,12 +249,7 @@ final class JndiResourceResolverFactory implements DnsNameResolver.ResourceResol
       checkAvailable();
       String[] rrType = new String[]{recordType};
       List<String> records = new ArrayList<>();
-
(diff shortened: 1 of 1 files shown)


jhaals:
I can find them in [openJDK7](http://www.docjar.com/html/api/com/sun/jndi/dns/DnsContext.java.html). @dapengzhang0 do you want me to remove them entirely since the project rely on the defaults?

dapengzhang0:
@jhaals What I was talking about is the JNDI-DNS properties you added have default. The JNDI-LDAP properties do not have default timeout. I don't think you should remove JNDI-LDAP properties. The JNDI may use LDAP providers.

jhaals:
Sounds like ldap settings in the dns provider should stay then.

Issue #6367, opened 2019-10-31
ComputeEngineChannelBuilder has wrong signature


dapengzhang0:
The generic type in `ComputeEngineChannelBuilder` class signature
```java
public final class ComputeEngineChannelBuilder
    extends ForwardingChannelBuilder<GoogleDefaultChannelBuilder> 
```
is wrong, it should be `ForwardingChannelBuilder<ComputeEngineChannelBuilder> `

However, fixing the bug might be ABI breaking for existing users.

cc @zhangkun83

dapengzhang0:
Did the following tests:

1. With `io.grpc:grpc-alts:1.24.1`, the following code can run:
```java
ManagedChannel channel = ComputeEngineChannelBuilder.forTarget(serverAddress)
        .build();
```

2. With  `io.grpc:grpc-alts:1.24.1`, the following code fails to run:
```java
ManagedChannel channel = ComputeEngineChannelBuilder.forTarget(serverAddress)
       .keepAliveTime(10, TimeUnit.MINUTES).build();
```

```
Exception in thread "main" java.lang.ClassCastException: io.grpc.alts.ComputeEngineChannelBuilder cannot be cast to io.grpc.alts.GoogleDefaultChannelBuilder
        at io.grpc.examples.alts.HelloWorldAltsClient.run(HelloWorldAltsClient.java:87)
        at io.grpc.examples.alts.HelloWorldAltsClient.main(HelloWorldAltsClient.java:39)
```
**So either no existing users are writing code like above or they have broken code like this but never tested it.**

3. Compile with  `io.grpc:grpc-alts:1.24.1`, and run with ` io.grpc:grpc-alts:1.26.0-SNAPSHOT` (with fixed signature) in the classpath:

```java
ManagedChannel channel = ComputeEngineChannelBuilder.forTarget(serverAddress)
        .build();
```
as well as 

```java
ManagedChannelBuilder<?> builder = ComputeEngineChannelBuilder.forTarget(serverAddress);
ManagedChannel channel = builder.keepAliveTime(10, TimeUnit.MINUTES).build();

```
Both can run.

So we can fix the signature directly without breaking ABI for existing users.

PR #6368, opened 2019-10-31
alts: deprecate ComputeEngineChannelBuilder and add GoogleComputeEngineChannelBuilder

dapengzhang0:
Resolving #6367


ejona86:
Did you confirm the fix was ABI-breaking?

I might just want to leave it as-is until it is naturally replaced by ChannelCredentials. There's not really any methods that matter and the main user of this is the google cloud libraries, so the bug may not impact many people.

dapengzhang0:
Yes, with tests described in https://github.com/grpc/grpc-java/issues/6367#issuecomment-548901382, we concluded that it can be fixed directly without breaking ABI for existing users.

PR #6372, opened 2019-11-01
alts: Check Cipher before declaring Conscrypt working

ejona86:


alts/src/main/java/io/grpc/alts/internal/AesGcmAeadCrypter.java
```diff
@@ -121,11 +121,36 @@ private static Provider getConscrypt() {
...
+      // This is because Conscrypt uses a newer (more secure) signing CA than the earlier Java
+      // supported. https://www.oracle.com/technetwork/java/javase/8u111-relnotes-3124969.html
+      // https://www.oracle.com/technetwork/java/javase/documentation/javase7supportreleasenotes-1601161.html#R170_121
+      //
+      // Use WARNING instead of INFO in this case because it is unlikely to be a supported
+      // environment. In the other cases we might be on Java 9+; it seems unlikely in this case.
+      // Note that on Java 7, we're likely to crash later because GCM is unsupported.
+      logger.log(
+          Level.WARNING,
+          "Could not load Conscrypt. Will try slower JDK implementation. This is may be because "
```
sanjaypujare: This is may be => This may be

PR #6385, opened 2019-11-02
xds: use ReferenceCountingSslContextProviderMap in TlsContextManager for clients and  servers

sanjaypujare:


xds/src/main/java/io/grpc/xds/sds/TlsContextManager.java
```diff
@@ -45,12 +63,42 @@ public static synchronized TlsContextManager getInstance() {
...
 
   /** Creates a SslContextProvider. Used for retrieving a client-side SslContext. */
   public SslContextProvider findOrCreateClientSslContextProvider(
       UpstreamTlsContext upstreamTlsContext) {
-    return SecretVolumeSslContextProvider.getProviderForClient(upstreamTlsContext);
+    checkNotNull(upstreamTlsContext, "upstreamTlsContext");
+    return mapForClients.get(upstreamTlsContext);
+  }
+
+  /** Releases an instance of the given client-side {@link SslContextProvider}.
```
creamsoup: nit: first line should be empty for multiline javadoc

xds/src/main/java/io/grpc/xds/sds/TlsContextManager.java
```diff
@@ -45,12 +63,42 @@ public static synchronized TlsContextManager getInstance() {
...
+   *  <p>Caller must not release a reference more than once. It's advised that you clear the
+   *  reference to the instance with the null returned by this method.
+   */
+  public SslContextProvider releaseClientSslContextProvider(SslContextProvider sslContextProvider) {
+    checkNotNull(sslContextProvider, "sslContextProvider");
+    return mapForClients.release(
+        (UpstreamTlsContext) sslContextProvider.getKey(), sslContextProvider);
+  }
+
+  /** Releases an instance of the given server-side {@link SslContextProvider}.
```
creamsoup: ditto

xds/src/main/java/io/grpc/xds/sds/TlsContextManager.java
```diff
@@ -45,12 +63,42 @@ public static synchronized TlsContextManager getInstance() {
...
+   *
+   *  <p>The instance must have been obtained from {@link #findOrCreateServerSslContextProvider}.
+   *  Otherwise will throw IllegalArgumentException.
+   *
+   *  <p>Caller must not release a reference more than once. It's advised that you clear the
+   *  reference to the instance with the null returned by this method.
+   */
+  public SslContextProvider releaseServerSslContextProvider(SslContextProvider sslContextProvider) {
+    checkNotNull(sslContextProvider, "sslContextProvider");
+    return mapForServers.release(
```
creamsoup: for this design, we can make the map takes sslContextProvider which can get the Key. although this seems a bit after thought, BiMap could be useful here but less depends on guava is better. any idea improve this api? Also, if we can avoid object type to ensure type safety, it would be really nice.

xds/src/test/java/io/grpc/xds/sds/ClientSslContextProviderFactoryTest.java
```diff
@@ -0,0 +1,159 @@
...
+                                    .addGrpcServices(
+                                        GrpcService.newBuilder()
+                                            .setGoogleGrpc(
+                                                GoogleGrpc.newBuilder()
+                                                    .setTargetUri(targetUri)
+                                                    .build())
+                                            .build())
+                                    .build())
+                            .build())
+                    .build());
```
creamsoup: nit: this is my personal preference, so you can ignore. but i think this is abit ugly and you can flatten it.

xds/src/main/java/io/grpc/xds/sds/TlsContextManager.java
sanjaypujare: I did give it a lot of thought and this was the best (according to me) of all the alternatives.I wasn't happy about the `Object` to `DownstreamTlsContext` cast either.

The RefCountingMap actually verifies the key+value (SslContextProvider) passed by the caller are consistent so instead of that check we just have the Map get the key from the value and perform release - I am okay with that. But the cast will still remain (from Object to K which is the key-type in the Map).

xds/src/test/java/io/grpc/xds/sds/ClientSslContextProviderFactoryTest.java
sanjaypujare: ok

xds/src/main/java/io/grpc/xds/sds/TlsContextManager.java
sanjaypujare: moved getKey into RefCountingMap.release

creamsoup: if `SslContextProvider` accepts a generic type, it can be solved. I am not sure it is the best way to fix the type safety issue. it is probably better than type unsafe code. also, the field name `key` is very `ReferenceCountingSslContextProviderMap` oriented name.

sanjaypujare: Regarding `SslContextProvider` accepting a generic type, that doesn't work as well because of type erasure. I will have to subclass `SslContextProvider` just based on the type of the `key` that's a member variable. So something like:

```
class SslContextProvider<K> {
K key;
}

class SslContextProviderForServer extends SslContextProvider<DownstreamTlsContext> {
}

class SslContextProviderForClient extends SslContextProvider<UpstreamTlsContext> {
}
```

That will provide the type-awareness for `key` that we are looking for. I think that's an overkill.

I agree with naming of `key` and did consider it.

creamsoup: `ReferenceCountingSslContextProviderMap` is already taking generic K, so the release just need to be `public SslContextProvider release(final SslContextProvider<K> value)`. during the compile time the type will be checked. maybe i am misunderstanding or missing something?

sanjaypujare: That might work. In my thinking `SslContextProvider` is not a concrete class but the subclass `SecretVolumeSslContextProvider` is and that will also need to be made generic with the `K` being propagated only to the superclass `SslContextProvider`. So this will be:
```
final class SecretVolumeSslContextProvider<K> extends SslContextProvider<K> {
}
```
And I can then just use `SecretVolumeSslContextProvider<DownstreamTlsContext>` vs `SecretVolumeSslContextProvider<UpstreamTlsContext>`

Okay, let me try that!

sanjaypujare: fixed.

xds/src/main/java/io/grpc/xds/sds/ReferenceCountingSslContextProviderMap.java
```diff
@@ -65,33 +65,32 @@ public SslContextProvider get(K key) {
...
-   * @param key the key that identifies the shared resource
    * @param value the instance to be released
    * @return a null which the caller can use to clear the reference to that instance.
    */
-  public SslContextProvider release(final K key, final SslContextProvider value) {
-    checkNotNull(key, "key");
+  public SslContextProvider<K> release(final SslContextProvider<K> value) {
     checkNotNull(value, "value");
+    K key = value.getKey();
+    checkNotNull(key, "key");
```
creamsoup: nit: we can skip this check if key is not null field. because NPE for "key" is not that meaningful for users.

xds/src/main/java/io/grpc/xds/sds/SslContextProvider.java
```diff
@@ -27,22 +29,33 @@
...
-  interface Callback {
+  public interface Callback {
     /** Informs callee of new/updated SslContext. */
     void updateSecret(SslContext sslContext);
 
     /** Informs callee of an exception that was generated. */
     void onException(Throwable throwable);
   }
 
+  protected SslContextProvider(K key) {
```
creamsoup: nit: key -> source? (I commented in other, but i am starting new to not dilute the other thread) 
key is one use of this field, but this class is not tied to the ReferenceCountingMap.

PR #6387, opened 2019-11-04
netty: remove redundant variable references

steverao:
Remove redundant variable references


dapengzhang0:
Thanks @steverao for the fix.

Issue #6279, opened 2019-10-15
offloadExecutor is experimental API


ejona86:
`ManagedChannelBuilder.offloadExecutor()`
`NameResolver.Args` `setOffloadExecutor()`/`getOffloadExecutor()`

Was originally added in #6238 (was renamed from blockingExecutor)

ejona86:
API review:
 * Rename to `offloadExecutor`

CC @groakley, FYI

ejona86:
CallCredentials needs to use offload executor. We should fix CallCredentials in ManagedChannelImpl and wait a release before stabilizing offload exeuctor.

larry-safran:
Ready to stabilize per meeting.

PR #6389, opened 2019-11-04
api: Rename blockingExecutor to offloadExecutor

ejona86:
The API review for #6279 came up with a more meaningful name that
better explains the intent. A setter for the old name was left in
ManagedChannelBuilder to ease migration to the new name by current
users.

CC @groakley 

I want to backport this before the release, since this is a brand-new API
in this release.

### DIFF
--- a/api/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/api/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -104,7 +104,7 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
   public abstract T executor(Executor executor);
 
   /**
-   * Provides a custom executor that will be used for operations that block.
+   * Provides a custom executor that will be used for operations that block or are expensive.
    *
    * <p>It's an optional parameter. If the user has not provided an executor when the channel is
    * built, the builder will use a static cached thread pool.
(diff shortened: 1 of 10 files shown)


ejona86:
CC @zhangkun83

api/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -117,10 +117,16 @@
...
    */
   @ExperimentalApi("https://github.com/grpc/grpc-java/issues/6279")
-  public T blockingExecutor(Executor executor) {
+  public T offloadExecutor(Executor executor) {
     throw new UnsupportedOperationException();
   }
 
+  @Deprecated
+  @ExperimentalApi("https://github.com/grpc/grpc-java/issues/6279")
+  public T blockingExecutor(Executor executor) {
```
dapengzhang0: Was `blockingExecutor` introduced in 1.25? Is `offloadExecutor()` going to be added into 1.25.0? If so we can just remove. @zhangkun83

zhangkun83: It has not made into a release, but there are internal users, so we can't simply delete it.

ejona86: There's not been a release with it, but we do have some internal users (3 apps/refs) using it. So I just left the old method to allow migrating them separately. I mention that in the PR description.

Issue #6322, opened 2019-10-23
Localized exceptions are being logged at level INFO (see also #5872)


teemann:
This issue is related to #5872.

### What version of gRPC are you using?
1.24.0
### What operating system (Linux, Windows,...) and version?
Windows 10
### What did you expect to see?
Useless exceptions being logged at a lower level than `INFO`.
### What did you see instead?
Useless exceptions being logged at the level `INFO`.
### Steps to reproduce the bug
Have your Windows set to german and have a client forcefully close the connection to the grpc server.

The following exception is then logged at `INFO` level:
```
Okt 23, 2019 10:04:15 AM io.grpc.netty.NettyServerTransport notifyTerminated
INFORMATION: Transport failed
java.io.IOException: Eine vorhandene Verbindung wurde vom Remotehost geschlossen
	at sun.nio.ch.SocketDispatcher.read0(Native Method)
	at sun.nio.ch.SocketDispatcher.read(SocketDispatcher.java:43)
	at sun.nio.ch.IOUtil.readIntoNativeBuffer(IOUtil.java:223)
	at sun.nio.ch.IOUtil.read(IOUtil.java:192)
	at sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:380)
	at io.netty.buffer.PooledByteBuf.setBytes(PooledByteBuf.java:247)
	at io.netty.buffer.AbstractByteBuf.writeBytes(AbstractByteBuf.java:1147)
	at io.netty.channel.socket.nio.NioSocketChannel.doReadBytes(NioSocketChannel.java:347)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:148)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:700)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:635)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:552)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:514)
	at io.netty.util.concurrent.SingleThreadEventExecutor$6.run(SingleThreadEventExecutor.java:1044)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
[...]

PR #6394, opened 2019-11-05
netty,okhttp: log FINE level for pure IOExceptions when transport terminated (v1.25.x backport)

creamsoup:
backport of #6334 related to #6322


creamsoup:
this was sync'd about 5 days ago, I think this is still safe. @ejona86, let me know if this is a deal breaker.

creamsoup:
cc @zhangkun83

PR #6414, opened 2019-11-11
core: service config error handling

creamsoup:
The APIs for the two-phase parse-apply service config error handling is already in place, while it's not being used. With this PR, we are switching to the new code path.

Spec: https://github.com/grpc/proposal/blob/master/A21-service-config-error-handling.md

### DIFF
--- a/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
+++ b/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
@@ -106,9 +106,12 @@ public final class AutoConfiguredLoadBalancerFactory {
     }
 
     /**
-     * Returns non-OK status if resolvedAddresses is rejected and should be considered as a
-     * name-resolution error.
+     * Returns non-OK status if resolvedAddresses is empty and delegate lb requires address ({@link
+     * LoadBalancer#canHandleEmptyAddressListFromNameResolution()} returns {@code false}).
+     * {@code AutoConfiguredLoadBalancer} doesn't expose {@code
+     * canHandleEmptyAddressListFromNameResolution} because delegate LB may not be deterministic.
      */
+    // TODO(jihuncho) migrate to use parsed config
     Status tryHandleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
       List<EquivalentAddressGroup> servers = resolvedAddresses.getAddresses();
       Attributes attributes = resolvedAddresses.getAttributes();
@@ -142,11 +145,14 @@ public final class AutoConfiguredLoadBalancerFactory {
             old.getClass().getSimpleName(), delegate.getClass().getSimpleName());
       }
 
+      ConfigOrError lbPolicyConfig = null;
       if (selection.config != null) {
         helper.getChannelLogger().log(
             ChannelLogLevel.DEBUG, "Load-balancing config: {0}", selection.config);
         attributes =
-            attributes.toBuilder().set(ATTR_LOAD_BALANCING_CONFIG, selection.config).build();
+            attributes.toBuilder().set(ATTR_LOAD_BALANCING_CONFIG, selection.rawConfig).build();
+        lbPolicyConfig = delegateProvider
+            .parseLoadBalancingPolicyConfig(selection.rawConfig);
       }
 
       LoadBalancer delegate = getDelegate();
@@ -159,6 +165,7 @@ public final class AutoConfiguredLoadBalancerFactory {
             ResolvedAddresses.newBuilder()
                 .setAddresses(selection.serverList)
                 .setAttributes(attributes)
+                .setLoadBalancingPolicyConfig(lbPolicyConfig)
                 .build());
         return Status.OK;
       }
@@ -251,6 +258,7 @@ public final class AutoConfiguredLoadBalancerFactory {
             return new PolicySelection(
                 provider,
                 policy.equals(GRPCLB_POLICY_NAME) ? servers : backendAddrs,
+                provider.parseLoadBalancingPolicyConfig(lbConfig.getRawConfigValue()),
                 lbConfig.getRawConfigValue());
           }
         }
@@ -280,9 +288,12 @@ public final class AutoConfiguredLoadBalancerFactory {
           return new PolicySelection(
               getProviderOrThrow(
                   "round_robin", "received balancer addresses but grpclb runtime is missing"),
-              backendAddrs, null);
+              backendAddrs,
+              /* config= */ null,
+              /* rawConfig= */ null);
         }
-        return new PolicySelection(grpclbProvider, servers, null);
+        return
+            new PolicySelection(grpclbProvider, servers, /* config= */ null, /* rawConfig= */ null);
       }
       // No balancer address this time.  If balancer address shows up later, we want to make sure
       // the warning is logged one more time.
@@ -290,7 +301,10 @@ public final class AutoConfiguredLoadBalancerFactory {
 
       // No config nor balancer address. Use default.
       return new PolicySelection(
-          getProviderOrThrow(defaultPolicy, "using default policy"), servers, null);
+          getProviderOrThrow(defaultPolicy, "using default policy"),
+          servers,
+          /* config= */ null,
+          /* rawConfig= */ null);
     }
   }
 
@@ -329,7 +343,8 @@ public final class AutoConfiguredLoadBalancerFactory {
           } else {
             return ConfigOrError.fromConfig(new PolicySelection(
                 provider,
-                /* serverList= */ null,
+                /* serverList= */ Collections.<EquivalentAddressGroup>emptyList(),
+                provider.parseLoadBalancingPolicyConfig(lbConfig.getRawConfigValue()),
                 lbConfig.getRawConfigValue()));
           }
         }
@@ -357,23 +372,19 @@ public final class AutoConfiguredLoadBalancerFactory {
   static final class PolicySelection {
     final LoadBalancerProvider provider;
     @Nullable final List<EquivalentAddressGroup> serverList;
-    // TODO(carl-mastrangelo): make this the non-raw service config object.
-    @Nullable final Map<String, ?> config;
-
-    PolicySelection(
-        LoadBalancerProvider provider, List<EquivalentAddressGroup> serverList,
-        @Nullable Map<String, ?> config) {
-      this.provider = checkNotNull(provider, "provider");
-      this.serverList = Collections.unmodifiableList(checkNotNull(serverList, "serverList"));
-      this.config = config;
-    }
+    @Nullable final ConfigOrError config;
+    // TODO(jihuncho) remove when AutoConfiguredLoadBalancerFactory doesn't depend on the Attributes
+    @Nullable final Map<String, ?> rawConfig;
 
     PolicySelection(
         LoadBalancerProvider provider,
-        @Nullable Map<String, ?> config) {
+        List<EquivalentAddressGroup> serverList,
+        @Nullable ConfigOrError config,
+        @Nullable Map<String, ?> rawConfig) {
       this.provider = checkNotNull(provider, "provider");
-      this.serverList = null;
+      this.serverList = Collections.unmodifiableList(checkNotNull(serverList, "serverList"));
       this.config = config;
+      this.rawConfig = rawConfig;
     }
   }
 
--- a/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
+++ b/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
@@ -20,6 +20,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
+import com.google.common.base.Objects;
 import io.grpc.Attributes;
 import io.grpc.ChannelLogger;
 import io.grpc.ChannelLogger.ChannelLogLevel;
@@ -115,14 +116,12 @@ public final class AutoConfiguredLoadBalancerFactory {
       List<EquivalentAddressGroup> servers = resolvedAddresses.getAddresses();
       Attributes attributes = resolvedAddresses.getAttributes();
 
-      Object lbPolicyConfig = resolvedAddresses.getLoadBalancingPolicyConfig();
-      ManagedChannelServiceConfig mcsc = (ManagedChannelServiceConfig) lbPolicyConfig;
+      PolicySelection policySelection =
+          (PolicySelection) resolvedAddresses.getLoadBalancingPolicyConfig();
       ResolvedPolicySelection resolvedSelection;
 
       try {
-        PolicySelection policySelection =
-            mcsc != null ? (PolicySelection) mcsc.getLoadBalancingConfig() : null;
-        resolvedSelection = pickLoadBalancerProvider(servers, policySelection);
+        resolvedSelection = resolveLoadBalancerProvider(servers, policySelection);
       } catch (PolicyException e) {
         Status s = Status.INTERNAL.withDescription(e.getMessage());
         helper.updateBalancingState(ConnectivityState.TRANSIENT_FAILURE, new FailingPicker(s));
@@ -200,23 +199,16 @@ public final class AutoConfiguredLoadBalancerFactory {
     }
 
     /**
-     * Picks a load balancer based on given criteria.  In order of preference:
-     *
-     * <ol>
-     *   <li>User provided lb on the channel.  This is a degenerate case and not handled here.
-     *       This options is deprecated.</li>
-     *   <li>Policy from "loadBalancingConfig" if present.  This is not covered here.</li>
-     *   <li>The policy from "loadBalancingPolicy" if present</li>
-     *   <li>"grpclb" if any gRPC LB balancer addresses are present</li>
-     *   <li>"pick_first" if the service config choice does not specify</li>
-     * </ol>
+     * Resolves a load balancer based on given criteria.  If policySelection is {@code null} and
+     * given servers contains any gRPC LB addresses, it will fall back to "grpclb". If no gRPC LB
+     * addresses are not present, it will fall back to {@link #defaultPolicy}.
      *
      * @param servers The list of servers reported
      * @param policySelection the selected policy from raw service config
-     * @return the new load balancer factory, never null
+     * @return the resolved policy selection
      */
     @VisibleForTesting
-    ResolvedPolicySelection pickLoadBalancerProvider(
+    ResolvedPolicySelection resolveLoadBalancerProvider(
         List<EquivalentAddressGroup> servers, @Nullable PolicySelection policySelection)
         throws PolicyException {
       // Check for balancer addresses
@@ -257,11 +249,13 @@ public final class AutoConfiguredLoadBalancerFactory {
               new PolicySelection(
                   getProviderOrThrow(
                       "round_robin", "received balancer addresses but grpclb runtime is missing"),
+                  /* rawConfig = */ null,
                   /* config= */ null),
               backendAddrs);
         }
         return new ResolvedPolicySelection(
-            new PolicySelection(grpclbProvider, /* config= */ null), servers);
+            new PolicySelection(
+                grpclbProvider, /* rawConfig= */ null, /* config= */ null), servers);
       }
       // No balancer address this time.  If balancer address shows up later, we want to make sure
       // the warning is logged one more time.
@@ -270,7 +264,9 @@ public final class AutoConfiguredLoadBalancerFactory {
       // No config nor balancer address. Use default.
       return new ResolvedPolicySelection(
           new PolicySelection(
-              getProviderOrThrow(defaultPolicy, "using default policy"), /* config= */ null),
+              getProviderOrThrow(defaultPolicy, "using default policy"),
+              /* rawConfig= */ null,
+              /* config= */ null),
           servers);
     }
   }
@@ -292,10 +288,18 @@ public final class AutoConfiguredLoadBalancerFactory {
    * means, it ignores LoadBalancer policies after the first available one even if any of them are
    * invalid.
    *
+   * <p>Order of policy preference:
+   *
+   * <ol>
+   *    <li>Policy from "loadBalancingConfig" if present</li>
+   *    <li>The policy from deprecated "loadBalancingPolicy" if present</li>
+   * </ol>
+   * </p>
+   *
    * <p>Unlike a normal {@link LoadBalancer.Factory}, this accepts a full service config rather than
    * the LoadBalancingConfig.
    *
-   * @return null if no selection could be made.
+   * @return the parsed {@link PolicySelection}, or {@code null} if no selection could be made.
    */
   @Nullable
   ConfigOrError parseLoadBalancerPolicy(Map<String, ?> serviceConfig, ChannelLogger channelLogger) {
@@ -321,6 +325,7 @@ public final class AutoConfiguredLoadBalancerFactory {
             }
             return ConfigOrError.fromConfig(new PolicySelection(
                 provider,
+                serviceConfig,
                 provider.parseLoadBalancingPolicyConfig(lbConfig.getRawConfigValue())));
           }
         }
@@ -347,19 +352,42 @@ public final class AutoConfiguredLoadBalancerFactory {
   @VisibleForTesting
   static final class PolicySelection {
     final LoadBalancerProvider provider;
+    @Nullable final Map<String, ?> rawConfig;
     @Nullable final ConfigOrError config;
 
     PolicySelection(
         LoadBalancerProvider provider,
+        @Nullable Map<String, ?> rawConfig,
         @Nullable ConfigOrError config) {
       this.provider = checkNotNull(provider, "provider");
+      this.rawConfig = rawConfig;
       this.config = config;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+      PolicySelection that = (PolicySelection) o;
+      return Objects.equal(provider, that.provider)
+          && Objects.equal(rawConfig, that.rawConfig)
+          && Objects.equal(config, that.config);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hashCode(provider, rawConfig, config);
     }
 
     @Override
     public String toString() {
       return MoreObjects.toStringHelper(this)
           .add("provider", provider)
+          .add("rawConfig", rawConfig)
           .add("config", config)
           .toString();
     }
@@ -370,7 +398,7 @@ public final class AutoConfiguredLoadBalancerFactory {
     final PolicySelection policySelection;
     final List<EquivalentAddressGroup> serverList;
 
-    public ResolvedPolicySelection(
+    ResolvedPolicySelection(
         PolicySelection policySelection, List<EquivalentAddressGroup> serverList) {
       this.policySelection = checkNotNull(policySelection, "policySelection");
       this.serverList = Collections.unmodifiableList(checkNotNull(serverList, "serverList"));
--- a/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
+++ b/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
@@ -17,10 +17,12 @@
 package io.grpc.internal;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static io.grpc.LoadBalancer.ATTR_LOAD_BALANCING_CONFIG;
+import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.MoreObjects;
 import io.grpc.Attributes;
+import io.grpc.ChannelLogger;
 import io.grpc.ChannelLogger.ChannelLogLevel;
 import io.grpc.ConnectivityState;
 import io.grpc.ConnectivityStateInfo;
@@ -39,7 +41,6 @@ import io.grpc.Status;
 import io.grpc.internal.ServiceConfigUtil.LbConfig;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Logger;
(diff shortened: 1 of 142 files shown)


core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -303,26 +303,29 @@ public void run() {
...
 
-      Attributes.Builder attrs = Attributes.newBuilder();
+      ResolutionResult.Builder resultBuilder = ResolutionResult.newBuilder().setAddresses(servers);
       if (!resolutionResults.txtRecords.isEmpty()) {
         ConfigOrError serviceConfig =
             parseServiceConfig(resolutionResults.txtRecords, random, getLocalHostname());
         if (serviceConfig != null) {
           if (serviceConfig.getError() != null) {
-            savedListener.onError(serviceConfig.getError());
+            savedListener.handleError(resultBuilder.setServiceConfig(serviceConfig).build());
```
zhangkun83: This is the "invalid config" case. Default config may be used in such case. You should still send the `ResolutionResult` to `savedListener.onResult()`, and let the error be handled by `ManagedChannelImpl` who has access to the default config. `savedListener.onError()` is intended for name resolution error, which is not the case here.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1332,19 +1334,24 @@ public void run() {
...
               effectiveServiceConfig = serviceConfig;
             } else {
-              effectiveServiceConfig = defaultServiceConfig;
+              effectiveServiceConfig = defaultServiceConfig != null
+                  ? defaultServiceConfig : EMPTY_SERVICE_CONFIG;
               if (defaultServiceConfig != null) {
                 channelLogger.log(
                     ChannelLogLevel.INFO,
                     "Received no service config, using default service config");
+              } else {
```
zhangkun83: Since you need this if-else branch anyway, it would be more readable to integrate the assignment of `effectiveServiceConfig` into this struct, rather than keeping it separate.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
creamsoup: thanks! you are right, onError should be the name resolution error. 
onResult calls onError if the invalid serviceConfig ends up cause the channel transient failure. PTAL

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
creamsoup: done.

dapengzhang0:
Anything need be done for the TODOs https://github.com/grpc/grpc-java/pull/6414/files#diff-354509714620f5c493a6810aab9419f2R252 and https://github.com/grpc/grpc-java/pull/6414/files#diff-354509714620f5c493a6810aab9419f2R259 ?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1300,24 +1302,34 @@ public NameResolverRegistry getNameResolverRegistry() {
...
 
           nameResolverBackoffPolicy = null;
+          ConfigOrError configOrError = resolutionResult.getServiceConfig();
+          Map<String, ?> serviceConfig = null;
+          if (configOrError != null) {
+            if (configOrError.getConfig() != null) {
+              serviceConfig = (Map<String, ?>) configOrError.getConfig();
+            } else if (nrReplied || defaultServiceConfig == null) {
+              // First DNS lookup has invalid service config, and cannot fall back to default(=null)
+              onError(configOrError.getError());
```
zhangkun83: This should be down in the else-branch of `if (!lookUpServiceConfig)`. If `lookUpServiceConfig == false`, the service config doesn't matter thus invalid config shouldn't be turned into an error. See my comment below.

creamsoup:
@dapengzhang0 thanks for finding those TODOs. I can use 2nd TODO instead of `nrReplied`.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1331,20 +1343,23 @@ public void run() {
...
-              if (defaultServiceConfig != null) {
-                channelLogger.log(
-                    ChannelLogLevel.INFO,
-                    "Received no service config, using default service config");
-              }
+              channelLogger.log(
+                  ChannelLogLevel.INFO,
+                  "Received no service config, using default service config");
+            } else {
+              effectiveServiceConfig = EMPTY_SERVICE_CONFIG;
```
zhangkun83: `onError()` should be put in here. Something like:
```java
if (firstResolutionResponse && configOrError.getError() != null) {
  onError(configOrError.getError());
}
```

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1300,24 +1302,34 @@ public NameResolverRegistry getNameResolverRegistry() {
...
-        @SuppressWarnings("ReferenceEquality")
+        @SuppressWarnings({"ReferenceEquality", "unchecked"})
         @Override
         public void run() {
           List<EquivalentAddressGroup> servers = resolutionResult.getAddresses();
           Attributes attrs = resolutionResult.getAttributes();
           channelLogger.log(
               ChannelLogLevel.DEBUG, "Resolved address: {0}, config={1}", servers, attrs);
 
+          boolean nrReplied = haveBackends == null;
```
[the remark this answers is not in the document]
creamsoup: renamed the `haveBackends`, change to use `waitingForServiceConfig` instead of using the old `haveBackends`.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1300,24 +1302,34 @@ public NameResolverRegistry getNameResolverRegistry() {
...
             haveBackends = true;
           }
 
           nameResolverBackoffPolicy = null;
+          ConfigOrError configOrError = resolutionResult.getServiceConfig();
+          Map<String, ?> serviceConfig = null;
+          if (configOrError != null) {
+            if (configOrError.getConfig() != null) {
+              serviceConfig = (Map<String, ?>) configOrError.getConfig();
+            } else if (nrReplied || defaultServiceConfig == null) {
```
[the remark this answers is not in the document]
creamsoup: oops, thanks. added a test!

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
creamsoup: done.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1399,13 +1414,20 @@ private void handleErrorInSyncContext(Status error) {
           new Object[] {getLogId(), error});
       if (haveBackends == null || haveBackends) {
         channelLogger.log(ChannelLogLevel.WARNING, "Failed to resolve name: {0}", error);
-        haveBackends = false;
       }
       // Call LB only if it's not shutdown.  If LB is shutdown, lbHelper won't match.
       if (NameResolverListener.this.helper != ManagedChannelImpl.this.lbHelper) {
         return;
       }
+      boolean nameResolverReplied = haveBackends == null;
```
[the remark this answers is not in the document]
creamsoup: done.

core/src/test/java/io/grpc/internal/ServiceConfigErrorHandlingTest.java
```diff
@@ -0,0 +1,627 @@
...
+    }
+  }
+
+  @After
+  public void cleanUp() {
+    LoadBalancerRegistry.getDefaultRegistry().deregister(mockLoadBalancerProvider);
+  }
+
+  @Test
+  public void emptyAddressesWithValidConfig_lbNeedsAddress() throws Exception {
```
zhangkun83: This is actually `emptyAddressesWithValidConfig_lbNeedsAddress_firstResolution()`, and we also need a `emptyAddressesWithValidConfig_lbNeedsAddress_secondResolution()` where the config is ignored.

Please make sure to cover all the cases listed under "Service Config Handling" in the spreadsheet.

It would be better to standardize the naming of the test cases according to the columns from the spec spreadsheet, so that we can easily check the test methods against the spec. For example, this would be `emptyAddresses_validConfig_firstResolution_lbNeedsAddress()`.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1397,15 +1408,21 @@ public void run() {
...
+      if (lastResolutionResultWasSuccess == null || lastResolutionResultWasSuccess) {
         channelLogger.log(ChannelLogLevel.WARNING, "Failed to resolve name: {0}", error);
-        haveBackends = false;
       }
       // Call LB only if it's not shutdown.  If LB is shutdown, lbHelper won't match.
       if (NameResolverListener.this.helper != ManagedChannelImpl.this.lbHelper) {
         return;
       }
+      if (waitingForServiceConfig) {
+        lastServiceConfig = defaultServiceConfig != null
```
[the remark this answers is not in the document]
creamsoup: yes. tests also supports that. thanks

core/src/test/java/io/grpc/internal/ServiceConfigErrorHandlingTest.java
creamsoup: tests are added. PTALA

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1331,20 +1333,29 @@ public void run() {
...
+                  "Received no service config and default service config is not provided. "
+                      + "Falling back to empty service config");
             }
 
             // FIXME(notcarl): reference equality is not right (although not harmful) right now.
             //                 Name resolver should return the same config if txt record is the same
             if (effectiveServiceConfig != lastServiceConfig) {
-              channelLogger.log(ChannelLogLevel.INFO,
-                  "Service config changed{0}", effectiveServiceConfig == null ? " to null" : "");
+              channelLogger.log(ChannelLogLevel.INFO, "Service config changed");
```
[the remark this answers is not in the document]
zhangkun83: For better introspection/debugging support, we can generate a string in each of the decision branch, and log it here.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1331,20 +1336,33 @@ public void run() {
...
+            } else if (waitingForServiceConfig
+                && configOrError != null
+                && configOrError.getError() != null) {
+              // First DNS lookup has invalid service config, and cannot fall back to default(=null)
+              onError(configOrError.getError());
+              return;
+            } else if (!waitingForServiceConfig
+                && configOrError != null
+                && configOrError.getError() != null) {
+              effectiveServiceConfig = lastServiceConfig;
```
zhangkun83: Maybe combine the two branches:
```java
else if (configOrError != null
￼       && configOrError.getError() != null) {
  if (waitingForServiceConfig) {
    onError(configOrError.getError());
    return;
  } else {
    effectiveServiceConfig = lastServiceConfig;
  }
}
```

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1300,24 +1298,31 @@ public NameResolverRegistry getNameResolverRegistry() {
...
-            haveBackends = true;
+            lastResolutionResultWasSuccess = true;
           }
 
           nameResolverBackoffPolicy = null;
+          ConfigOrError configOrError = resolutionResult.getServiceConfig();
+          Map<String, ?> serviceConfig = null;
+          if (configOrError != null && configOrError.getConfig() != null) {
+            serviceConfig = (Map<String, ?>) configOrError.getConfig();
+          }
```
zhangkun83: How about:
```java
Map<String, ?> serviceConfig = null;
Status serviceConfigError = null;
if (configOrError != null) {
  serviceConfig = (Map<String, ?>) configOrError.getConfig();
  serviceConfigError = configOrError.getError();
}
```

Then the code below on line 1344 can just use `serviceConfigError` and that may be more readable.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1300,24 +1298,31 @@ public NameResolverRegistry getNameResolverRegistry() {
     public void onResult(final ResolutionResult resolutionResult) {
       final class NamesResolved implements Runnable {
 
-        @SuppressWarnings("ReferenceEquality")
+        @SuppressWarnings({"ReferenceEquality", "unchecked"})
         @Override
         public void run() {
           List<EquivalentAddressGroup> servers = resolutionResult.getAddresses();
+          if (servers.isEmpty() && lastResolutionResultWasSuccess == Boolean.TRUE) {
```
zhangkun83: According to the "Address handling" section from the spreadsheet, empty address should be applied to LB if LB doesn't need address. So, config should be ignored but the empty addresses should be passed to `tryHandleResolvedAddresses()`. To make the decision we need to know whether LB needs addresses, thus it has to be done inside `tryHandleResolvedAddresses()` with `waitingForServiceConfig` passed to it too.

Above is about the addresses. For config, it'd be better to make the decision in the main if-tree below. It's easier to follow if all decisions are made in the same place.

zhangkun83:
Sorry, I was focusing on the decision-making tree too much and missed one important part -- the applying of the config. You need to cast the object from `ConfigOrError.getConfig()` to `ManagedChannelServiceConfig`, and do the following things with its content:

1. Apply the parameters inside `ManagedChannelServiceConfig` to the channel.
2. Pass `loadBalancingPolicySelection` to `AutoConfiguredLoadBalancerFactory.tryHandleResolvedAddresses()`, which would then apply the new config, including LB policy selection, and pass the LB config to the policy via `LoadBalancer.ResolvedAddresses.getLoadBalancingPolicyConfig()`.

This would be a much larger change. We might be able to submit this PR without other changes, meaning we only switch the parsing path, but keep the applying path. Then we can switch the applying path in a separate PR. Let's talk tomorrow.

core/src/test/java/io/grpc/internal/ServiceConfigErrorHandlingTest.java
```diff
@@ -0,0 +1,655 @@
...
+    nameResolverFactory.nextConfigOrError.set(ConfigOrError.fromConfig(serviceConfig));
+
+    createChannel();
+
+    ArgumentCaptor<ResolvedAddresses> resultCaptor =
+        ArgumentCaptor.forClass(ResolvedAddresses.class);
+    verify(mockLoadBalancer).handleResolvedAddresses(resultCaptor.capture());
+    assertThat(resultCaptor.getValue().getAddresses()).containsExactly(addressGroup);
+    Attributes actualAttrs = resultCaptor.getValue().getAttributes();
+    assertThat(actualAttrs.get(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG))
```
zhangkun83: This is not what LBs are looking for. You should check `LoadBalancer.ATTR_LOAD_BALANCING_CONFIG` instead, and the expected value should be the value of the `mock_lb` key in the service config, which is
```
          {
            "param1": "value1",
            "param2": "value2",
          }
```
from my comment above.

core/src/test/java/io/grpc/internal/ServiceConfigErrorHandlingTest.java
```diff
@@ -0,0 +1,655 @@
...
+  @Test
+  public void validConfig_lbDoesNotNeedAddress() throws Exception {
+    FakeNameResolverFactory nameResolverFactory =
+        new FakeNameResolverFactory.Builder(expectedUri)
+            .setServers(ImmutableList.of(addressGroup))
+            .build();
+    channelBuilder.nameResolverFactory(nameResolverFactory);
+
+    ImmutableMap<String, Object> serviceConfig =
+        ImmutableMap.<String, Object>of("loadBalancingPolicy", "mock_lb");
```
zhangkun83: `loadBalancingPolicy` is the deprecated format, as it doesn't carry the LB config. The new format should be:
```
{
      "loadBalancingConfig": [
        {"mock_lb" :
          {
            "param1": "value1",
            "param2": "value2",
          }
        }
    ]
}
```

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
creamsoup: done.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
creamsoup: resolved PTAL

core/src/test/java/io/grpc/internal/ServiceConfigErrorHandlingTest.java
creamsoup: done.

core/src/test/java/io/grpc/internal/ServiceConfigErrorHandlingTest.java
creamsoup: added TODO. iirc we discussed in offline that we want to make this change after this PR because it requires changes in LB to reduce the scope of this pr.

creamsoup:
AutoconfiguredLoadBalancerFactory is using ManagedChannelServiceConfig instead of parsed json. PTAL

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -141,24 +148,26 @@ Status tryHandleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
             ChannelLogLevel.INFO, "Load balancer changed from {0} to {1}",
             old.getClass().getSimpleName(), delegate.getClass().getSimpleName());
       }
-
-      if (selection.config != null) {
+      PolicySelection policySelection =
+          mcsc != null ? (PolicySelection) mcsc.getLoadBalancingConfig() : null;
+      ConfigOrError lbConfig = null;
+      if (policySelection != null) {
```
zhangkun83: `selection` is non-null.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -204,18 +213,18 @@ LoadBalancerProvider getDelegateProvider() {
...
      *   <li>Policy from "loadBalancingConfig" if present.  This is not covered here.</li>
-     *   <li>"grpclb" if any gRPC LB balancer addresses are present</li>
      *   <li>The policy from "loadBalancingPolicy" if present</li>
+     *   <li>"grpclb" if any gRPC LB balancer addresses are present</li>
      *   <li>"pick_first" if the service config choice does not specify</li>
      * </ol>
      *
      * @param servers The list of servers reported
-     * @param config the service config object
+     * @param mcsc the parsed {@link ManagedChannelServiceConfig}
```
zhangkun83: This method only uses `mcsc.getLoadBalancingConfig()` which is a `PolicySelection`. We should only pass it to this method.

This method is focused on handling the GRPCLB special case, the whole javadoc needs to be rewritten to reflect that. The terminology is currently around the raw config, and that needs to be changed too.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -228,35 +237,12 @@ PolicySelection decideLoadBalancerProvider(
...
-                policy.equals(GRPCLB_POLICY_NAME) ? servers : backendAddrs,
-                lbConfig.getRawConfigValue());
-          }
-        }
-        if (!haveBalancerAddress) {
-          throw new PolicyException(
-            "None of " + policiesTried + " specified by Service Config are available.");
+      if (mcsc != null && mcsc.getLoadBalancingConfig() != null) {
+        PolicySelection policy = (PolicySelection) mcsc.getLoadBalancingConfig();
+        if (policy.provider != null) {
```
zhangkun83: `PolicySelection.provider` is non-null.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -303,31 +304,31 @@ public void run() {
...
-          } else {
+          resultBuilder.setServiceConfig(serviceConfig);
+          if (serviceConfig.getError() == null) {
             @SuppressWarnings("unchecked")
             Map<String, ?> config = (Map<String, ?>) serviceConfig.getConfig();
-            attrs.set(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG, config);
+            resultBuilder
+                .setAttributes(
+                    Attributes.newBuilder()
+                        .set(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG, config)
```
zhangkun83: This attribute should be marked as deprecated.

core/src/main/java/io/grpc/internal/DnsNameResolver.java
```diff
@@ -303,31 +304,31 @@ public void run() {
...
             parseServiceConfig(resolutionResults.txtRecords, random, getLocalHostname());
         if (serviceConfig != null) {
-          if (serviceConfig.getError() != null) {
-            savedListener.onError(serviceConfig.getError());
-            return;
-          } else {
+          resultBuilder.setServiceConfig(serviceConfig);
+          if (serviceConfig.getError() == null) {
             @SuppressWarnings("unchecked")
             Map<String, ?> config = (Map<String, ?>) serviceConfig.getConfig();
```
zhangkun83: This doesn't seem right... You have changed `parseServiceConfig` to return the `ConfigOrError` returned from `serviceConfigParser`, which is no longer the raw format. To keep the old behavior that `GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG` is the raw config, you probably want to keep `parseServiceConfig()` as is, and instead call `serviceConfigParser` from here.

core/src/main/java/io/grpc/internal/ManagedChannelServiceConfig.java
```diff
@@ -65,7 +73,7 @@ static ManagedChannelServiceConfig fromServiceConfig(
       boolean retryEnabled,
       int maxRetryAttemptsLimit,
       int maxHedgedAttemptsLimit,
-      @Nullable Object loadBalancingConfig) {
+      Object loadBalancingConfig) {
```
zhangkun83: Why remove `@Nullable`?

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -1308,72 +1322,116 @@ public void run() {
...
+                && !effectiveServiceConfig.rawServiceConfig.isEmpty()) {
               effectiveAttrs = attrs.toBuilder()
-                  .set(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG, effectiveServiceConfig)
+                  .set(
+                      GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG,
+                      effectiveServiceConfig.rawServiceConfig)
                   .build();
             }
+
+            Object loadBalancingPolicyConfig = effectiveServiceConfig.parsed.getConfig();
```
zhangkun83: You should extract `PolicySelection` out of this `ManagedChannelServiceConfig` and pass it to `.setLoadBalancingPolicyConfig()`, because `lb` returned a `PolicySelection` from `parseLoadBalancingPolicy()`.

core/src/test/java/io/grpc/internal/AutoConfiguredLoadBalancerFactoryTest.java
```diff
@@ -580,102 +593,23 @@ public void decideLoadBalancerProvider_grpclbProviderNotFound_noBackendAddress()
...
-    }
-  }
-
-  @Test
-  public void decideLoadBalancerProvider_serviceConfigLbConfigFailsOnUnknown() throws Exception {
-    AutoConfiguredLoadBalancer lb = lbf.newLoadBalancer(new TestHelper());
-    Map<String, ?> serviceConfig =
-        parseConfig("{\"loadBalancingConfig\": [ {\"magic_balancer\": {} } ] }");
+    Map<String, ?> rawServiceConfig =
+        parseConfig("{\"loadBalancingPolicy\": \"round_robin\"}");
```
zhangkun83: This old format json `loadBalancingPolicy` should be the one removed instead of the new json format `loadBalancingConfig`. But then you would need to add a case for `loadBalancingPolicy` in the set of `parseLoadBalancerPolicy_*()`.

core/src/test/java/io/grpc/internal/AutoConfiguredLoadBalancerFactoryTest.java
```diff
@@ -884,13 +902,18 @@ public void updateAddresses(List<EquivalentAddressGroup> addrs) {
     }
   }
 
-  private static final class FakeLoadBalancerProvider extends LoadBalancerProvider {
+  private static class FakeLoadBalancerProvider extends LoadBalancerProvider {
```
zhangkun83: Why remove `final`?

core/src/test/java/io/grpc/internal/DnsNameResolverTest.java
```diff
@@ -1072,7 +1075,8 @@ public void shouldUseJndi_trueIfItMightPossiblyBeValid() {
 
   @Test
   public void parseServiceConfig_capturesParseError() {
-    ConfigOrError result = DnsNameResolver.parseServiceConfig(
+    DnsNameResolver resolver = newResolver("localhost", DEFAULT_PORT);
```
zhangkun83: `parseServiceConfig()` is still static so this change is not necessary.

core/src/test/java/io/grpc/internal/DnsNameResolverTest.java
```diff
@@ -1092,15 +1097,20 @@ public void parseServiceConfig_capturesChoiceError() {
...
 
     assertThat(result).isNull();
   }
 
   @Test
+  @SuppressWarnings("unchecked")
   public void parseServiceConfig_matches() {
-    ConfigOrError result = DnsNameResolver.parseServiceConfig(
+    DnsNameResolver resolver = newResolver("localhost", DEFAULT_PORT);
+    when(mockServiceConfigParser.parseServiceConfig(any(Map.class)))
```
zhangkun83: This could be gone too.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -1101,6 +1041,7 @@ public void firstResolvedServerFailedToConnect() throws Exception {
     inOrder.verify(mockLoadBalancer).handleResolvedAddresses(
         ResolvedAddresses.newBuilder()
             .setAddresses(Arrays.asList(addressGroup))
+            .setAttributes(Attributes.EMPTY)
```
zhangkun83: Attributes are by default empty, so this shouldn't be necessary.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -2804,6 +2750,7 @@ public void channelTracing_serviceConfigChange() throws Exception {
             new EquivalentAddressGroup(
                 Arrays.asList(new SocketAddress() {}, new SocketAddress() {}))))
         .setAttributes(attributes)
+        .setServiceConfig(ConfigOrError.fromConfig(managedChannelServiceConfig))
```
zhangkun83: This could just use `ManagedChannelServiceConfig.empty()`.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -2820,9 +2767,10 @@ public void channelTracing_serviceConfigChange() throws Exception {
...
         .build();
     nameResolverFactory.resolvers.get(0).listener.onResult(resolutionResult2);
-    assertThat(getStats(channel).channelTrace.events).hasSize(prevSize);
+    assertThat(getStats(channel).channelTrace.events).hasSize(prevSize + 1);
```
zhangkun83: You should probably assert that additional event.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -3186,16 +3135,21 @@ public void retryBackoffThenChannelShutdown_retryShouldStillHappen_newCallShould
...
-        .newBuilder().set(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG, serviceConfig).build();
+        .newBuilder().set(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG, rawServiceConfig).build();
 
     FakeNameResolverFactory nameResolverFactory =
         new FakeNameResolverFactory.Builder(expectedUri)
             .setServers(Collections.singletonList(new EquivalentAddressGroup(socketAddress)))
             .build();
-    nameResolverFactory.nextResolvedAttributes.set(attributesWithRetryPolicy);
+    ManagedChannelServiceConfig managedChannelServiceConfig =
+        ManagedChannelServiceConfig.fromServiceConfig(rawServiceConfig, true, 3, 3, null);
```
zhangkun83: Probably better to define the limits (3, 3) as constants and refer them throughout the test.

core/src/test/java/io/grpc/internal/ServiceConfigErrorHandlingTest.java
```diff
@@ -0,0 +1,736 @@
...
+  }
+
+  private static final class FakeNameResolverFactory extends NameResolver.Factory {
+    final URI expectedUri;
+    final List<EquivalentAddressGroup> servers;
+    final boolean resolvedAtStart;
+    final Status error;
+    final ArrayList<FakeNameResolver> resolvers = new ArrayList<>();
+    final AtomicReference<ConfigOrError> nextConfigOrError = new AtomicReference<>();
+    final AtomicReference<Map<String, ?>> nextRawServiceConfig = new AtomicReference<>();
```
zhangkun83: `nextConfigOrError` should go away. Instead you will call `ServiceConfigParser.parseServiceConfig()` with `nextRawServiceConfig`, and use the returned `ConfigOrError` to call `listener.onResult()`. In each individual test you should not call `mockLoadBalancerProvider.parseLoadBalancingConfig()` because it's called by `AutoConfiguredLoadBalancerFactory`.

core/src/test/java/io/grpc/internal/ServiceConfigInterceptorTest.java
```diff
@@ -92,30 +90,22 @@ public void withWaitForReady() {
     JsonObj name = new JsonObj("service", "service");
     JsonObj methodConfig = new JsonObj("name", new JsonList(name), "waitForReady", true);
     JsonObj serviceConfig = new JsonObj("methodConfig", new JsonList(methodConfig));
+    ManagedChannelServiceConfig parsedServiceConfig =
```
zhangkun83: Better to add a factory method for this object.

core/src/main/java/io/grpc/internal/ManagedChannelImpl.java
```diff
@@ -582,18 +587,20 @@ ClientStream newSubstream(ClientStreamTracer.Factory tracerFactory, Metadata new
         new ExecutorHolder(
             checkNotNull(builder.offloadExecutorPool, "offloadExecutorPool"));
     this.nameResolverRegistry = builder.nameResolverRegistry;
+    ScParser serviceConfigParser =
```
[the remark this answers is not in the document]
creamsoup: oops i was wrong, it was used to parse default service config.

core/src/test/java/io/grpc/internal/AutoConfiguredLoadBalancerFactoryTest.java
creamsoup: it is used in mock to verify and to capture the argument in `handleResolvedAddressGroups_propagateLbConfigToDelegate()`.

core/src/test/java/io/grpc/internal/ManagedChannelImplTest.java
```diff
@@ -962,67 +961,6 @@ public void nameResolutionFailed_delayedTransportShutdownCancelsBackoff() {
     assertTrue(nameResolverBackoff.isCancelled());
   }
 
-  @Test
-  public void nameResolverReturnsEmptySubLists_becomeErrorByDefault() throws Exception {
```
zhangkun83: What are the replacement for these two deleted tests?

core/src/test/java/io/grpc/internal/AutoConfiguredLoadBalancerFactoryTest.java
```diff
@@ -17,7 +17,6 @@
 package io.grpc.internal;
 
 import static com.google.common.truth.Truth.assertThat;
-import static io.grpc.LoadBalancer.ATTR_LOAD_BALANCING_CONFIG;
```
zhangkun83: Checks for `ATTR_LOAD_BALANCING_CONFIG` should be preserved, because it's still there for back-ward compatibility.

zhangkun83:
LGTM. Please make a separate PR to do the fork.

creamsoup:
closing it. see #6564

PR #6424, opened 2019-11-13
Upgrade to protobuf lite 3.10

### LINKED ISSUE #6405: Have grpc-protobuf-lite depend on protobuf-javalite, not protobuf-lite
### What version of gRPC-Java are you using?
1.25.0, 1.24.2

My understanding is that with more recent protobuf versions the Java Lite code generation requires a different approach and is compatible with com.google.protobuf:protobuf-javalite artifacts as opposed to the old com.google.protobuf:protobuf-lite:3.0.1 artifacts. The current grpc-protobuf-lite still depends on the old com.google.protobuf:protobuf-lite:3.0.1, yet the current grpc artifacts generally use recent protobuf versions. I expected to see a dependency on the new protobuf-javalite artifact instead. Or would a new grpc-protobuf-javalite artifact be more appropriate?

ejona86:
This finally brings lite into version number sync with full proto.

-----

This was a change I made in September for proto 3.9. We have since upgraded to 3.10 and the commit message was no longer correct, but it seems the code didn't need to change at all. (You can see my branch name is lite-3.9 :-) )

Fixes #6405 

CC @liusheng, @kiwik, @rafi-kamal, @bubenheimer

See also protocolbuffers/protobuf#6867

### DIFF
--- a/android-interop-testing/app/build.gradle
+++ b/android-interop-testing/app/build.gradle
@@ -41,14 +41,15 @@ android {
 protobuf {
     protoc { artifact = 'com.google.protobuf:protoc:3.10.0' }
     plugins {
-        javalite { artifact = "com.google.protobuf:protoc-gen-javalite:3.0.0" }
         grpc { artifact = 'io.grpc:protoc-gen-grpc-java:1.26.0-SNAPSHOT' // CURRENT_GRPC_VERSION
         }
     }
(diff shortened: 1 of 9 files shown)


ejona86:
@creamsoup, PTAL. I needed a reasonable amount more changes for Android

PR #6433, opened 2019-11-15
xds: refactor to eliminate LookasideChannelLb

dapengzhang0:
This is a followup of #6425 to further refactor/cleanup. Now that `LookasideChannelLb` does very little thing, it does not need to be a class anymore.

After this refactor, the `XdsLoadBalancer2` will have simpler architecture:
```txt
             XdsLoadBalancer2
                /           \
        fallbackLb         lookasdieLb
```


dapengzhang0:
PTAL

xds/src/main/java/io/grpc/xds/LookasideLb.java
```diff
@@ -119,8 +135,14 @@ public void handleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
     lookasideChannelLb.handleResolvedAddresses(resolvedAddresses);
   }
 
+  /**
+   * Creates a balancer that is associated to a balancer name. The is to handle the legacy usecase
+   * that requires balancerName from xds config.
+   */
   private LoadBalancerProvider newLookasideChannelLbProvider(
       final String balancerName, final Node node, final List<ChannelCreds> channelCredsList) {
+    checkNotNull(balancerName, "balancerName is required");
```
voidzcy: Is this really required? IMO, I would not bother have this.

xds/src/test/java/io/grpc/xds/LookasideLbTest.java
```diff
@@ -179,11 +363,304 @@ public void handleResolvedAddress_createLbChannel()
...
+        .build();
+    serverResponseWriter.onNext(
+        DiscoveryResponse.newBuilder()
+            .addResources(Any.pack(clusterLoadAssignment))
+            .setTypeUrl("type.googleapis.com/envoy.api.v2.ClusterLoadAssignment")
+            .build());
+
+    verify(lookasideChannelCallback, never()).onAllDrop();
+    verify(localityStore).updateDropPercentage(ImmutableList.of(
+        new DropOverload("cat_1", 300_00),
```
voidzcy: FYI: These numbers do not look elegant. Please be consistent for all numbers you write, either put underscores for all or none (in general I don't). You could choose to not add the underscore, but if you do, please do it elegantly. 

I know this is copied from the old code and these tests would eventually go away, or need to be re-written in another form (with `XdsClient`, you do not need a RPC service involved, just need to build `EndpointUpdate`s and test your load balancing behaviors). So you could choose to not change it now.

xds/src/main/java/io/grpc/xds/LookasideLb.java
dapengzhang0: It was use to make sure the method is used only for the old usecase that needs balancer name.

xds/src/test/java/io/grpc/xds/LookasideLbTest.java
dapengzhang0: It's just an arbitrary string in test I don't think it necessary to keep that pattern consistent. It dose not affect readability.

xds/src/main/java/io/grpc/xds/LookasideLb.java
```diff
@@ -119,8 +135,13 @@ public void handleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
     lookasideChannelLb.handleResolvedAddresses(resolvedAddresses);
   }
 
+  /**
+   * Creates a balancer that is associated to a balancer name. The is to handle the legacy usecase
+   * that requires balancerName from xds config.
+   */
   private LoadBalancerProvider newLookasideChannelLbProvider(
```
voidzcy: Why this needs to be a method? Basically it only creates a new object and return it? Why wrap this into a method? I would say you should clean things up as you move them around. Otherwise, this sort of things will spread everywhere.

dapengzhang0: Cleaned up the method.

xds/src/main/java/io/grpc/xds/LookasideLb.java
```diff
@@ -142,72 +164,164 @@ public String getPolicyName() {
...
+    }
+    return channel;
   }
 
-  private static final class LookasideChannelLbFactoryImpl implements LookasideChannelLbFactory {
+  /**
+   * Callback on ADS stream events. The callback methods should be called in a proper {@link
+   * io.grpc.SynchronizationContext}.
+   */
+  interface LookasideChannelCallback {
```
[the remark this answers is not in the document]
dapengzhang0: Renamed and changed/deleted doc.

xds/src/main/java/io/grpc/xds/LookasideLb.java
```diff
@@ -142,72 +164,164 @@ public String getPolicyName() {
...
+     */
+    void onAllDrop();
+  }
+
+  private final class EndpointWatcherImpl implements EndpointWatcher {
+
+    final LoadReportClient lrsClient;
+    final LoadReportCallback lrsCallback;
+    final LocalityStore localityStore;
+    boolean firstEdsResponseReceived;
```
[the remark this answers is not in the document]
dapengzhang0: Renamed.

xds/src/main/java/io/grpc/xds/LookasideLb.java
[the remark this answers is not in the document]
voidzcy: Did you push?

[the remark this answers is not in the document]
dapengzhang0: I forgot :(

Issue #6364, opened 2019-10-31
Propose to officially support ARM64 platform


liusheng:
### Is your feature request related to a problem?

Yes, recently, I am working on promoting Hadoop building on ARM64 server, while an error occured due to missing the `io.grpc:protoc-gen-grpc-java:exe:linux-aarch_64:1.15.1` artifact for aarch platform, see error messages:
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  03:13 min
[INFO] Finished at: 2019-05-17T09:28:55+08:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.xolstice.maven.plugins:protobuf-maven-plugin:0.5.1:compile-custom (default) on project hadoop-yarn-csi: Missing:
[ERROR] ----------
[ERROR] 1) io.grpc:protoc-gen-grpc-java:exe:linux-aarch_64:1.15.1
[ERROR]
[ERROR]   Try downloading the file manually from the project website.
[ERROR]
[ERROR]   Then, install it using the command:
[ERROR]       mvn install:install-file -DgroupId=io.grpc -DartifactId=protoc-gen-grpc-java -Dversion=1.15.1 -Dclassifier=linux-aarch_64 -Dpackaging=exe -Dfile=/path/to/file
[ERROR]
[ERROR]   Alternatively, if you host your own repository you can deploy the file there:
[ERROR]       mvn deploy:deploy-file -DgroupId=io.grpc -DartifactId=protoc-gen-grpc-java -Dversion=1.15.1 -Dclassifier=linux-aarch_64 -Dpackaging=exe -Dfile=/path/to/file -Durl=[url] -DrepositoryId=[id]
[ERROR]
[ERROR]   Path to dependency:
[ERROR]         1) org.apache.hadoop:hadoop-yarn-csi:jar:3.3.0-SNAPSHOT
[ERROR]         2) io.grpc:protoc-gen-grpc-java:exe:linux-aarch_64:1.15.1
[ERROR]
```
In the Maven central repo, there are different version of  `protoc-gen-grpc-java` artifacts for linux-x86, osx-x86, windows-x86, but no `aarch64` specific package. see:
https://search.maven.org/search?q=g:io.grpc%20AND%20a:protoc-gen-grpc-java&core=gav

[...]

PR #6441, opened 2019-11-19
Add support for cross-compiling for aarch64 platform

liusheng:
This change modified the building scripts to support cross compile
grpc-java for aarch64 platform

For grpc/grpc-java#6364

### DIFF
--- a/buildscripts/kokoro/linux_artifacts.sh
+++ b/buildscripts/kokoro/linux_artifacts.sh
@@ -36,4 +36,4 @@ cp -r "$LOCAL_MVN_TEMP"/* "$MVN_ARTIFACT_DIR"/
 
 # for aarch64 platform
 sudo apt-get install -y g++-aarch64-linux-gnu
-SKIP_TESTS=true ARCH=aarch_64 ./buildscripts/kokoro/unix.sh
+SKIP_TESTS=true ARCH=aarch_64 "$GRPC_JAVA_DIR"/buildscripts/kokoro/unix.sh
--- a/buildscripts/kokoro/linux_artifacts.sh
+++ b/buildscripts/kokoro/linux_artifacts.sh
@@ -35,12 +35,5 @@ mkdir -p "$MVN_ARTIFACT_DIR"
 cp -r "$LOCAL_MVN_TEMP"/* "$MVN_ARTIFACT_DIR"/
 
 # for aarch64 platform
-apt-get install -y wget autoconf automake libtool g++-aarch64-linux-gnu g++ make
-
-apt-get install software-properties-common -y
-add-apt-repository ppa:openjdk-r/ppa -y
-apt-get update
-apt install openjdk-8-jdk -y
-export JAVA_HOME=`dirname $(dirname $(update-alternatives --list javac |grep java-8))`
-update-ca-certificates -f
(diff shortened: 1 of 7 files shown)


liusheng:
Hi @ejona86,

I have updated this PR and now the `buildscripts/kokoro/linux_artifacts.sh` can be used to build the compiler for `aarch64` platform. cloud you please take a look ?

Thanks a lot!

buildscripts/kokoro/linux_artifacts.sh
```diff
@@ -33,3 +33,14 @@ popd
 readonly MVN_ARTIFACT_DIR="${MVN_ARTIFACT_DIR:-$GRPC_JAVA_DIR/mvn-artifacts}"
 mkdir -p "$MVN_ARTIFACT_DIR"
 cp -r "$LOCAL_MVN_TEMP"/* "$MVN_ARTIFACT_DIR"/
+
+# for aarch64 platform
+apt-get install -y wget autoconf automake libtool g++-aarch64-linux-gnu g++ make
```
ejona86: wget, autoconf, automake, libtool, g++, and make should all already be available by the image supplied by Kokoro. Similarly openjdk and JAVA_HOME and really even update-ca-certificates seem like they shouldn't be necessary (we just ran java on the earlier lines of this file).

So I'd expect this entire thing to become something more like:
```sh
apt-get install -y g++-aarch64-linux-gnu
SKIP_TESTS=true ARCH=aarch_64 ./buildscripts/kokoro/unix.sh
```

buildscripts/kokoro/unix.sh
```diff
@@ -70,6 +72,11 @@ fi
 LOCAL_MVN_TEMP=$(mktemp -d)
 # Note that this disables parallel=true from GRADLE_FLAGS
 if [[ -z "${ALL_ARTIFACTS:-}" ]]; then
+  if [[ $ARCH == "aarch_64" ]]; then
+    GRADLE_FLAGS+=" -x grpc-compiler:generateTestProto -x grpc-compiler:generateTestLiteProto"
```
ejona86: This is probably now unnecessary, since #6424 is now merged.

buildscripts/kokoro/linux_artifacts.sh
ejona86: Looks like this needs a 'sudo' before it.

ejona86: Looks like this needs a 'sudo' before it. I'll tweak some of this and push an update.

buildscripts/kokoro/unix.sh
ejona86: Nevermind. This is because we can't run the compiled code since it is for a different architecture.

ejona86:
@liusheng, thank you!

liusheng:
@ejona86 thanks a lot for changing and proposing this change :)
Is there plan to release and publish the artifacts including aarch64 support to maven central ? I will be very appreciated if you can help to do that.

ejona86:
@liusheng, with #6496 it will go out in the 1.26 release, which is scheduled for Dec 17th.

liusheng:
@ejona86 got it, thank  you!

PR #6460, opened 2019-11-23
Add Executor When Creating SslContext

ZhenLian:
This PR is to add one more Executor parameter when creating the SslContext.
In Netty, we already have this implementation for passing Executor when creating SslContext: https://github.com/netty/netty/pull/8847. This extra Executor is used to take some time-consuming tasks when doing SSL handshake. However, in current gRPC implementation, we are not using this API. 

In this PR, the relevant changes are:

- get the executorPool from ChannelBuilder or ServerBuilder

- pass the executorPool all the way down to ClientTlsHandler

- fill executorPool in when creating SslHandler

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
@@ -627,4 +627,11 @@ public abstract class AbstractManagedChannelImplBuilder
     T thisT = (T) this;
     return thisT;
   }
+
+  /**
+   * Returns the internal offload executor pool for offloading tasks.
+   */
+  protected ObjectPool<? extends Executor> getOffloadExecutorPool() {
+    return this.offloadExecutorPool;
+  }
 }
--- a/core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
+++ b/core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
@@ -297,4 +297,11 @@ public abstract class AbstractServerImplBuilder<T extends AbstractServerImplBuil
       return null;
     }
   }
+
+  /**
+   * Returns the internal ExecutorPool for offloading tasks.
+   */
+  protected ObjectPool<? extends Executor> getExecutorPool() {
+    return this.executorPool;
+  }
 }
--- a/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
+++ b/netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
@@ -33,6 +33,7 @@ import io.grpc.SecurityLevel;
 import io.grpc.Status;
 import io.grpc.internal.GrpcAttributes;
 import io.grpc.internal.GrpcUtil;
+import io.grpc.internal.ObjectPool;
 import io.netty.channel.ChannelDuplexHandler;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandler;
@@ -58,6 +59,7 @@ import io.netty.util.AttributeMap;
 import java.net.SocketAddress;
 import java.net.URI;
 import java.util.Arrays;
+import java.util.concurrent.Executor;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
@@ -106,20 +108,35 @@ final class ProtocolNegotiators {
 
   /**
    * Create a server TLS handler for HTTP/2 capable of using ALPN/NPN.
+   * @param executorPool a dedicated {@link Executor} pool for time-consuming TLS tasks
    */
-  public static ProtocolNegotiator serverTls(final SslContext sslContext) {
+  public static ProtocolNegotiator serverTls(final SslContext sslContext,
+      final ObjectPool<? extends Executor> executorPool) {
     Preconditions.checkNotNull(sslContext, "sslContext");
+    final Executor executor;
+    if (executorPool != null) {
+      // The handlers here can out-live the {@link ProtocolNegotiator}.
+      // To keep their own reference to executor from executorPool, we use an extra (unused)
+      // reference here forces the executor to stay alive, which prevents it from being re-created
+      // for every connection.
+      executor = executorPool.getObject();
+    } else {
+      executor = null;
+    }
     return new ProtocolNegotiator() {
       @Override
       public ChannelHandler newHandler(GrpcHttp2ConnectionHandler handler) {
         ChannelHandler gnh = new GrpcNegotiationHandler(handler);
-        ChannelHandler sth = new ServerTlsHandler(gnh, sslContext);
+        ChannelHandler sth = new ServerTlsHandler(gnh, sslContext, executorPool);
         return new WaitUntilActiveHandler(sth);
       }
 
       @Override
-      public void close() {}
-
+      public void close() {
+        if (executorPool != null && executor != null) {
+          executorPool.returnObject(executor);
+        }
+      }
 
       @Override
       public AsciiString scheme() {
@@ -128,22 +145,37 @@ final class ProtocolNegotiators {
     };
   }
 
+  /**
+   * Create a server TLS handler for HTTP/2 capable of using ALPN/NPN.
+   */
+  public static ProtocolNegotiator serverTls(final SslContext sslContext) {
+    return serverTls(sslContext, null);
+  }
+
   static final class ServerTlsHandler extends ChannelInboundHandlerAdapter {
+    private Executor executor;
     private final ChannelHandler next;
     private final SslContext sslContext;
 
     private ProtocolNegotiationEvent pne = ProtocolNegotiationEvent.DEFAULT;
 
-    ServerTlsHandler(ChannelHandler next, SslContext sslContext) {
+    ServerTlsHandler(ChannelHandler next,
+        SslContext sslContext,
+        final ObjectPool<? extends Executor> executorPool) {
       this.sslContext = checkNotNull(sslContext, "sslContext");
       this.next = checkNotNull(next, "next");
+      if (executorPool != null) {
+        this.executor = executorPool.getObject();
+      }
     }
 
     @Override
     public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
       super.handlerAdded(ctx);
       SSLEngine sslEngine = sslContext.newEngine(ctx.alloc());
-      ctx.pipeline().addBefore(ctx.name(), null, new SslHandler(sslEngine, false));
+      ctx.pipeline().addBefore(ctx.name(), /* name= */ null, this.executor != null
+          ? new SslHandler(sslEngine, false, this.executor)
+          : new SslHandler(sslEngine, false));
     }
 
     @Override
@@ -259,11 +291,18 @@ final class ProtocolNegotiators {
 
   static final class ClientTlsProtocolNegotiator implements ProtocolNegotiator {
 
-    public ClientTlsProtocolNegotiator(SslContext sslContext) {
+    public ClientTlsProtocolNegotiator(SslContext sslContext,
+        ObjectPool<? extends Executor> executorPool) {
       this.sslContext = checkNotNull(sslContext, "sslContext");
+      this.executorPool = executorPool;
+      if (this.executorPool != null) {
+        this.executor = this.executorPool.getObject();
+      }
     }
 
     private final SslContext sslContext;
+    private final ObjectPool<? extends Executor> executorPool;
+    private Executor executor;
 
     @Override
     public AsciiString scheme() {
@@ -273,12 +312,17 @@ final class ProtocolNegotiators {
     @Override
     public ChannelHandler newHandler(GrpcHttp2ConnectionHandler grpcHandler) {
       ChannelHandler gnh = new GrpcNegotiationHandler(grpcHandler);
-      ChannelHandler cth = new ClientTlsHandler(gnh, sslContext, grpcHandler.getAuthority());
+      ChannelHandler cth = new ClientTlsHandler(gnh, sslContext, grpcHandler.getAuthority(),
+          this.executor);
       return new WaitUntilActiveHandler(cth);
     }
 
     @Override
-    public void close() {}
+    public void close() {
+      if (this.executorPool != null && this.executor != null) {
+        this.executorPool.returnObject(this.executor);
+      }
+    }
   }
 
   static final class ClientTlsHandler extends ProtocolNegotiationHandler {
@@ -286,13 +330,16 @@ final class ProtocolNegotiators {
     private final SslContext sslContext;
     private final String host;
     private final int port;
+    private Executor executor;
 
-    ClientTlsHandler(ChannelHandler next, SslContext sslContext, String authority) {
+    ClientTlsHandler(ChannelHandler next, SslContext sslContext, String authority,
+        Executor executor) {
       super(next);
       this.sslContext = checkNotNull(sslContext, "sslContext");
       HostPort hostPort = parseAuthority(authority);
       this.host = hostPort.host;
       this.port = hostPort.port;
+      this.executor = executor;
     }
 
     @Override
@@ -301,7 +348,9 @@ final class ProtocolNegotiators {
       SSLParameters sslParams = sslEngine.getSSLParameters();
       sslParams.setEndpointIdentificationAlgorithm("HTTPS");
       sslEngine.setSSLParameters(sslParams);
-      ctx.pipeline().addBefore(ctx.name(), /* name= */ null, new SslHandler(sslEngine, false));
+      ctx.pipeline().addBefore(ctx.name(), /* name= */ null, this.executor != null
+          ? new SslHandler(sslEngine, false, this.executor)
+          : new SslHandler(sslEngine, false));
     }
 
     @Override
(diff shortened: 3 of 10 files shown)


netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -128,29 +133,52 @@ public AsciiString scheme() {
...
     }
 
     @Override
     public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
       super.handlerAdded(ctx);
       SSLEngine sslEngine = sslContext.newEngine(ctx.alloc());
-      ctx.pipeline().addBefore(ctx.name(), null, new SslHandler(sslEngine, false));
+      if (this.executorPool != null
+          && SystemPropertyUtil.getBoolean("io.netty.handler.ssl.openssl.useTasks", false)
+          == true) {
```
sanjaypujare: why compare with `true`? Shouldn't `&& SystemPropertyUtil.getBoolean("io.netty.handler.ssl.openssl.useTasks", false))` be enough?

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -128,29 +133,52 @@ public AsciiString scheme() {
     };
   }
 
+  public static ProtocolNegotiator serverTls(final SslContext sslContext) {
```
sanjaypujare: Copy the Javadoc from lines 110-112?

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -368,8 +419,13 @@ static HostPort parseAuthority(String authority) {
    * be negotiated, the {@code handler} is added and writes to the {@link io.netty.channel.Channel}
    * may happen immediately, even before the TLS Handshake is complete.
    */
+  public static ProtocolNegotiator tls(SslContext sslContext,
+      ObjectPool<? extends Executor> executorPool) {
+    return new ClientTlsProtocolNegotiator(sslContext, executorPool);
+  }
+
   public static ProtocolNegotiator tls(SslContext sslContext) {
```
sanjaypujare: Copy the Javadoc from lines 417-421?

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -627,4 +627,11 @@ private T thisT() {
     T thisT = (T) this;
     return thisT;
   }
+
+  /**
+   * Returns the internal ExecutorPool for offloading tasks.
+   */
+  protected ObjectPool<? extends Executor> getExecutor() {
```
sanjaypujare: `getExecutor` => `getExecutorPool` (to be accurate)

core/src/main/java/io/grpc/internal/AbstractServerImplBuilder.java
```diff
@@ -297,4 +297,11 @@ private T thisT() {
       return null;
     }
   }
+
+  /**
+   * Returns the internal ExecutorPool for offloading tasks.
+   */
+  protected ObjectPool<? extends Executor> getExecutor() {
```
sanjaypujare: `getExecutor` => `getExecutorPool` (to be accurate)

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -627,4 +627,11 @@ private T thisT() {
...
+    return this.executorPool;
```
sanjaypujare: Should you be using `this.offloadExecutorPool` if the purpose it to use it for "some time-consuming tasks"? I think the `offloadExecutorPool` is used for resolver tasks but seems like could be used here as well. But both are initialized to `DEFAULT_EXECUTOR_POOL` BTW so for default case there's no difference.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -128,29 +133,52 @@ public AsciiString scheme() {
...
+      }
     }
 
     @Override
     public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
       if (evt instanceof ProtocolNegotiationEvent) {
         pne = (ProtocolNegotiationEvent) evt;
       } else if (evt instanceof SslHandshakeCompletionEvent) {
+        if (this.executor != null && this.executorPool != null) {
+          this.executorPool.returnObject(this.executor);
```
sanjaypujare: So the `executor` is returned to the pool only if `SslHandshakeCompletionEvent` is received (success or failure). Is it possible for the `ServerTlsHandler` (or `ClientTlsHandler`) to be removed without ever receiving this event? In that case there is a resource leak?

ZhenLian: Thanks for the comments Sanjay! 
Yeah my previous assumption was SslHandshakeCompletionEvent will always be passed through and if that's not the case, we might end up with a resource leak. A better way might be to make use of the close() function in ClientTlsProtocolNegotiator and the equivalent anonymous part in serverTls. I will update the code accordingly, and add some tests.

sanjaypujare: Reading the Javadoc for `ProtocolNegotiator.close` : "Called when the Channel transitions to terminated. Is currently only supported on client-side; server-side protocol negotiators will not see this method called."

So this gets called only on channel termination and not for server-side protocol negotiator.

ZhenLian: Yeah I think this needs to be further discussed before I could move on.
@ejona86 I remembered we talked about this in our last meeting and you suggested putting the cleaning up functions into the close() method. But since it is not guaranteed on server side that this function will be called when TLS handshake ends, could you please provide more insights on how to handle this for server side? 
Any suggestions would be really appreciated :-)

ejona86: I would suggest putting any cleanup code in handlerRemoved().

ZhenLian:
@ejona86 @sanjaypujare  Sorry for the delay of this PR. Now it is ready to be reviewed. Could you please take a look at it when free? Thanks in advance!

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -128,22 +132,44 @@ public AsciiString scheme() {
...
+    ServerTlsHandler(ChannelHandler next, SslContext sslContext) {
+      this(next, sslContext, null);
     }
 
     @Override
     public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
       super.handlerAdded(ctx);
       SSLEngine sslEngine = sslContext.newEngine(ctx.alloc());
-      ctx.pipeline().addBefore(ctx.name(), null, new SslHandler(sslEngine, false));
+      if (this.executorPool != null) {
```
sanjaypujare: You removed the check for `"io.netty.handler.ssl.openssl.useTasks"` system property. May be it is okay - but just curious why it was there and why you removed it now.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -107,13 +109,15 @@ public static ProtocolNegotiator serverPlaintext() {
   /**
    * Create a server TLS handler for HTTP/2 capable of using ALPN/NPN.
    */
-  public static ProtocolNegotiator serverTls(final SslContext sslContext) {
+  public static ProtocolNegotiator serverTls(final SslContext sslContext,
+      final ObjectPool<? extends Executor> executorPool) {
     Preconditions.checkNotNull(sslContext, "sslContext");
+
```
ejona86: We don't want to create an `executor` every connection, so we should have the protocol negotiator keep a reference alive. 

The `executor` should be pulled out of the `executorPool` before returning and returned to the pool in `close()`. But it actually looks like server-side never calls close(). @jiangtaoli2016, is it possible your [alts protocol negotiator](https://github.com/grpc/grpc-java/blob/2de5b4e254a09f69555251151c2a8b950310bc9b/alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java#L116) hasn't been cleaning up the channel? On server-side I'd sort of expect close() to be called while some handlers may still be functioning, so that means you would continue passing the pool to the handler and let it _also_ get the executor from the pool.

Ditto for client-side, although on client-side we do guarantee that no connections exist when the negotiator is closed, so it is safe to pass Executor directly to the handler and only deal with the pool in the negotiator.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
ZhenLian: According to this PR, https://github.com/netty/netty/pull/8847, we need to set this system property in order to make the extra Executor to take effect. It is used for offloading time-consuming TLS operations.
Not setting this property would leave the Executor unused, which falls back to the original implementation, though the overhead of getting executor might be a concern. 
My earlier intention is that, if users didn't specify this property, we'd better use the old API to avoid the overhead. But after a while I realized that checking this property at handler adding time has nothing to do with whether it is set when actually needed ... This system property is a global configuration and is subject to change at any time. Adding an additional check here doesn't help that much, IIUC.
My current plan is, we will finally have a user documentation, which will ask users to set this property if using any time-consuming TLS operations. @ejona86 please correct me if I misunderstand something or overthinking this because I remembered we also talked about this before :-)

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
ZhenLian: IIUC, regardless of how alts protocol negotiator behaves, the current solution is:
1. on client side, we have the protocol negotiator with a reference to a `Executor` and clean it up when calling `close()`
2. on server side, since `close()` is not guaranteed to be called every time, we have to keep the current approach, which is to  have a reference to a `ExecutorPool` and create an `Executor` every connection

ejona86: You should rely on close(). If close is not functioning that is a bug that should be fixed, not worked around. You should not _just_ get a reference for each connection on server-side. That can lead to _horrible_ behavior with frequent destruction/creation of threads.

ejona86: #6508 fixes the NettyServer behavior to close the negotiator.

ZhenLian:
@ejona86 @sanjaypujare 
Thanks for the first round of feedback!
I resolved all the comments in the implementations, and adjusted the test structure to cover more test behaviors. I added a new class CountingExecutorPool in NettyTestUtil.java for the purpose of keeping track of the Executors inside the pool.
Please let me know if this looks better, and if there are any other things to be improved. 
Thanks again for the reviewing!

core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java
```diff
@@ -627,4 +627,11 @@ private T thisT() {
     T thisT = (T) this;
     return thisT;
   }
+
+  /**
+   * Returns the internal offload executor pool for offloading tasks.
+   */
+  protected ObjectPool<? extends Executor> getOffloadExecutorPool() {
```
sanjaypujare: Now that we have this getter can we mark the data member `this.offloadExecutorPool` private and have all the sites (tests) use the getter?

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -363,13 +415,23 @@ static HostPort parseAuthority(String authority) {
     return new HostPort(host, port);
   }
 
+  /**
+   * Returns a {@link ProtocolNegotiator} that ensures the pipeline is set up so that TLS will
+   * be negotiated, the {@code handler} is added and writes to the {@link io.netty.channel.Channel}
+   * may happen immediately, even before the TLS Handshake is complete.
```
sanjaypujare: Do you want to mention the additional param? Otherwise comment is identical to the next one.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -107,19 +109,29 @@ public static ProtocolNegotiator serverPlaintext() {
...
+      executor = executorPool.getObject();
+    } else {
+      executor = null;
+    }
     return new ProtocolNegotiator() {
       @Override
       public ChannelHandler newHandler(GrpcHttp2ConnectionHandler handler) {
         ChannelHandler gnh = new GrpcNegotiationHandler(handler);
-        ChannelHandler sth = new ServerTlsHandler(gnh, sslContext);
+        ChannelHandler sth = new ServerTlsHandler(gnh, sslContext, executor);
```
ejona86: The handlers here can out-live the `ProtocolNegotiator`. So they need to get their own reference to `executor` from `executorPool`. I suggest passing in `executorPool` here and calling `getObject()` within the constructor.

You would still keep the rest of this protocol negotiator the same; you'd still call `getObject()` within `serverTls()` and `returnObject()` in `close()`. Keeping the extra (unused) reference here forces the executor to stay alive, which prevents it from being re-created for every connection.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -301,7 +345,15 @@ protected void handlerAdded0(ChannelHandlerContext ctx) {
...
-      ctx.pipeline().addBefore(ctx.name(), /* name= */ null, new SslHandler(sslEngine, false));
+
+      SslHandler sslHandler;
+      if (this.executor != null) {
+        sslHandler = new SslHandler(sslEngine, false, this.executor);
+      } else {
+        sslHandler = new SslHandler(sslEngine, false);
+      }
+      checkNotNull(sslHandler, "sslHandler");
+      ctx.pipeline().addBefore(ctx.name(), null, sslHandler);
```
sanjaypujare: The original code had the param `/* name= */` which is missing here for the 2nd arg. In any case do we need 7 additional lines or can you have the following:
```
ctx.pipeline().addBefore(ctx.name(), /* name= */ null, this.executor != null ?
      new SslHandler(sslEngine, false, this.executor) :
      new SslHandler(sslEngine, false));
```
The `checkNotNull` is gone but why is it required? A `new` call will never return null (the only 2 paths used here).

netty/src/test/java/io/grpc/netty/ProtocolNegotiatorsTest.java
```diff
@@ -246,7 +246,7 @@ public void tlsHandler_failsOnNullEngine() throws Exception {
 
   @Test
   public void tlsHandler_handlerAddedAddsSslHandler() throws Exception {
-    ChannelHandler handler = new ServerTlsHandler(grpcHandler, sslContext);
+    ChannelHandler handler = new ServerTlsHandler(grpcHandler, sslContext, null);
```
sanjaypujare: For all `null` or literal values passed as arguments, it is recommended to use a comment like `executorPool` as the parameter name. For all call sites (lines 258, 279, ...)

ejona86: I think for tests we generally don't care.

netty/src/test/java/io/grpc/netty/NettyChannelBuilderTest.java
```diff
@@ -143,26 +145,42 @@ public void failIfSslContextIsNotClient() {
...
     assertTrue(negotiator instanceof ProtocolNegotiators.PlaintextProtocolNegotiator);
+    negotiator.close();
   }
 
   @Test
   public void createProtocolNegotiatorByType_plaintextUpgrade() {
     ProtocolNegotiator negotiator = NettyChannelBuilder.createProtocolNegotiatorByType(
         NegotiationType.PLAINTEXT_UPGRADE,
-        noSslContext);
+        noSslContext,null);
```
sanjaypujare: Add space and param name `executorPool` (in all such calls)

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
sanjaypujare: may be it is obvious but I thought I should mention this: with this additional call to `getObject()` in the `ServerTlsHandler` constructor you will also need a `returnObject()` call within `ServerTlsHandler` somewhere - you may need to implement `handlerRemoved` just for this.

I didn't completely understand the comment "Keeping the extra (unused) reference here forces the executor to stay alive, which prevents it from being re-created for every connection."

ejona86: We expect the Pool here to generally be a SharedResourcePool. It releases a resource when it is no longer used, so if we don't hold a reference to the executor it may be created for a connection, shutdown after the connection handshake, created again for the next connection, shutdown after the connection handshake, etc.

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
ZhenLian: Thank you so much for the suggestion!

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
ZhenLian: Thanks a lot for the valuable discussion here!
I have updated as suggested but am not 100% confident that I understand correctly.
Previously, I thought the Pool would be a collection of objects, and every time a getObject() call will return a completely new Object. This requires us to call returnObject() for every object got from getObject().
It looks like for a SharedResourcePool, the new object will only be created when the old one is released(by GC, etc). The number of calling getObject() and returnObject() doesn't necessarily to be the same.
Please correct me if I am wrong. Thanks again!

ZhenLian:
@ejona86 @sanjaypujare 
Thanks for the review. All comments are addressed. Could you please take another look? Thanks!

netty/src/main/java/io/grpc/netty/ProtocolNegotiators.java
```diff
@@ -106,20 +108,35 @@ public static ProtocolNegotiator serverPlaintext() {
...
+    } else {
+      executor = null;
+    }
     return new ProtocolNegotiator() {
       @Override
       public ChannelHandler newHandler(GrpcHttp2ConnectionHandler handler) {
         ChannelHandler gnh = new GrpcNegotiationHandler(handler);
-        ChannelHandler sth = new ServerTlsHandler(gnh, sslContext);
+        // The handlers here can out-live the {@link ProtocolNegotiator}.
+        // To keep their own reference to executor from executorPool, we use an extra (unused)
```
ejona86: This comment looks like it was intended for line 118 instead of here. There isn't actually an executor reference here, and it seems to imply the one in the ServerTlsHandler constructor, but that one is not unused.

ZhenLian: Thanks for the catch! The comments are updated. Since I don't have the write access, could you also merge this PR if everything looks fine? Thanks in advance!

PR #6466, opened 2019-11-26
Support for lazily serialized values in Metadata.

markb74:
Add a new a Metadata.BinaryStreamMarshaller interface which serializes to/from instances of InputStream, and a corresponding Key.of() factory method.
    
Values set with this type of key will be kept unserialized internally, alongside a reference to the Marshaller. A new method InternalMetadata.serializePartial(), returns values which are either
byte[] or InputStream, allowing transport-specific handling of lazily-serialized values.

For the regular serialize() method, stream-marshalled values will be converted to byte[] via InputStreams.

Transport-internal code can also create metadata with pre-parsed values.

### DIFF
--- a/api/src/main/java/io/grpc/InternalMetadata.java
+++ b/api/src/main/java/io/grpc/InternalMetadata.java
@@ -106,7 +106,7 @@ public final class InternalMetadata {
    */
   @Internal
   public static <T> Object parsedValue(BinaryStreamMarshaller<T> marshaller, T value) {
-    return new Metadata.LazyValue(marshaller, value);
+    return new Metadata.LazyValue<>(marshaller, value);
   }
 
   /**
--- a/api/src/main/java/io/grpc/InternalMetadata.java
+++ b/api/src/main/java/io/grpc/InternalMetadata.java
@@ -83,16 +83,43 @@ public final class InternalMetadata {
     return md.headerCount();
   }
 
+  /**
+   * Serializes all metadata entries, leaving some values as {@link InputStream}s.
+   *
+   * <p>Produces serialized names and values interleaved. result[i*2] are names, while
+   * result[i*2+1] are values.
+   *
+   * <p>Names are byte arrays as described according to the {@link Metadata#serialize}
+   * method. Values are either byte arrays or {@link InputStream}s.
+   */
   @Internal
   public static Object[] serializePartial(Metadata md) {
     return md.serializePartial();
   }
 
+  /**
+   * Creates a holder for a pre-parsed value read by the transport.
+   *
+   * @param key The key associated with this value.
+   * @param value The value to store.
+   * @return an object holding the pre-parsed value for this key.
+   */
   @Internal
   public static <T> Object parsedValue(Key<T> key, T value) {
     return Metadata.LazyValue.create(key, value);
   }
 
+  /**
+   * Creates a new {@link Metadata} instance from serialized data,
+   * with some values pre-parsed. Metadata will mutate the passed in array.
+   *
+   * @param usedNames The number of names used.
+   * @param namesAndValues An array of iterleaved names and values,
+   * with each name (at even indices) represented as a byte array,
+   * and each value (at odd indices) represented as either a byte
+   * array or an object returned by the {@link #parsedValue}
+   * method.
+   */
   @Internal
   public static Metadata newMetadataWithParsedValues(int usedNames, Object[] namesAndValues) {
     return new Metadata(usedNames, namesAndValues);
--- a/api/src/main/java/io/grpc/InternalMetadata.java
+++ b/api/src/main/java/io/grpc/InternalMetadata.java
@@ -19,7 +19,6 @@ package io.grpc;
 import com.google.common.io.BaseEncoding;
 import io.grpc.Metadata.AsciiMarshaller;
 import io.grpc.Metadata.Key;
-import java.io.InputStream;
 import java.nio.charset.Charset;
 
 /**
--- a/api/src/main/java/io/grpc/InternalMetadata.java
+++ b/api/src/main/java/io/grpc/InternalMetadata.java
@@ -89,7 +89,7 @@ public final class InternalMetadata {
   }
 
   @Internal
-  public static <T> Metadata.LazyValue<T> parsedValue(Key<T> key, T value) {
+  public static <T> Object parsedValue(Key<T> key, T value) {
     return Metadata.LazyValue.create(key, value);
   }
 
--- a/api/src/main/java/io/grpc/InternalMetadata.java
+++ b/api/src/main/java/io/grpc/InternalMetadata.java
@@ -82,4 +82,19 @@ public final class InternalMetadata {
   public static int headerCount(Metadata md) {
     return md.headerCount();
   }
+
+  @Internal
+  public static Object[] serializePartial(Metadata md) {
+    return md.serializePartial();
+  }
+
+  @Internal
+  public static <T> Metadata.LazyValue<T> parsedValue(Key<T> key, T value) {
+    return Metadata.LazyValue.create(key, value);
+  }
+
+  @Internal
+  public static Metadata newMetadataWithStreamValues(int usedNames, Object[] namesAndValues) {
+    return new Metadata(usedNames, namesAndValues);
+  }
 }
--- a/api/src/main/java/io/grpc/InternalMetadata.java
+++ b/api/src/main/java/io/grpc/InternalMetadata.java
@@ -115,10 +115,10 @@ public final class InternalMetadata {
    *
    * @param usedNames The number of names used.
    * @param namesAndValues An array of iterleaved names and values,
-   * with each name (at even indices) represented as a byte array,
-   * and each value (at odd indices) represented as either a byte
-   * array or an object returned by the {@link #parsedValue}
-   * method.
+   *     with each name (at even indices) represented as a byte array,
+   *     and each value (at odd indices) represented as either a byte
+   *     array or an object returned by the {@link #parsedValue}
+   *     method.
    */
   @Internal
   public static Metadata newMetadataWithParsedValues(int usedNames, Object[] namesAndValues) {
--- a/api/src/main/java/io/grpc/InternalMetadata.java
+++ b/api/src/main/java/io/grpc/InternalMetadata.java
@@ -100,13 +100,13 @@ public final class InternalMetadata {
   /**
    * Creates a holder for a pre-parsed value read by the transport.
    *
-   * @param key The key associated with this value.
+   * @param marshaller The {@link Metadata#BinaryStreamMarshaller} associated with this value.
    * @param value The value to store.
    * @return an object holding the pre-parsed value for this key.
    */
   @Internal
-  public static <T> Object parsedValue(Key<T> key, T value) {
-    return Metadata.LazyValue.create(key, value);
+  public static <T> Object parsedValue(BinaryStreamMarshaller<T> marshaller, T value) {
+    return new Metadata.LazyValue(marshaller, value);
   }
 
   /**
--- a/api/src/main/java/io/grpc/InternalMetadata.java
+++ b/api/src/main/java/io/grpc/InternalMetadata.java
@@ -114,7 +114,7 @@ public final class InternalMetadata {
    * with some values pre-parsed. Metadata will mutate the passed in array.
    *
    * @param usedNames The number of names used.
-   * @param namesAndValues An array of iterleaved names and values,
+   * @param namesAndValues An array of interleaved names and values,
    *     with each name (at even indices) represented as a byte array,
    *     and each value (at odd indices) represented as either a byte
    *     array or an object returned by the {@link #parsedValue}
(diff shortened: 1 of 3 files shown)


api/src/main/java/io/grpc/InternalMetadata.java
```diff
@@ -82,4 +82,19 @@ public static Metadata newMetadata(int usedNames, byte[]... binaryValues) {
...
+    return md.serializePartial();
+  }
+
+  @Internal
+  public static <T> Metadata.LazyValue<T> parsedValue(Key<T> key, T value) {
+    return Metadata.LazyValue.create(key, value);
+  }
+
+  @Internal
+  public static Metadata newMetadataWithStreamValues(int usedNames, Object[] namesAndValues) {
```
markb74: I should change this to "*WithParsedValues"

api/src/main/java/io/grpc/InternalMetadata.java
```diff
@@ -82,4 +82,19 @@ public static Metadata newMetadata(int usedNames, byte[]... binaryValues) {
...
     return md.headerCount();
   }
+
+  @Internal
+  public static Object[] serializePartial(Metadata md) {
+    return md.serializePartial();
+  }
+
+  @Internal
+  public static <T> Metadata.LazyValue<T> parsedValue(Key<T> key, T value) {
```
ejona86: I guess we should make this return code be `Object`, since LazyValue is hidden? Although we don't really care about ABI here, it also means ABI won't break if the internals change.

api/src/test/java/io/grpc/MetadataTest.java
```diff
@@ -334,6 +390,96 @@ public void invalidKeyName() {
     }
   }
 
+  @Test
+  public void getSameKeyReturnsSameInstnce() {
```
ejona86: s/Instnce/Instance/ ditto below.

api/src/test/java/io/grpc/MetadataTest.java
```diff
@@ -334,6 +390,96 @@ public void invalidKeyName() {
...
+    Fish salmon = new Fish("salmon");
+    Metadata h = new Metadata();
+    h.put(KEY_STREAMED, salmon);
+    assertSame(salmon, h.get(KEY_STREAMED));
```
ejona86: Is this a guarantee we actually want to provide? I was thinking we were saying we'd let the InputStream hacks take care of this, so objects would be duplicated by default and mutable objects would be "safe."

I think you were making use of this in your example consuming code internally. Getting rid of this would require using the same InputStream trick that's used for outbound messages?

markb74: Good point, this was effectively testing the implementation, changed to assertEquals (also above).

api/src/test/java/io/grpc/MetadataTest.java
markb74: Done.

api/src/main/java/io/grpc/InternalMetadata.java
markb74: Done.

api/src/main/java/io/grpc/Metadata.java
```diff
@@ -406,11 +459,36 @@ private void expand(int newCapacity) {
...
+      for (int i = 0; i < size; i++) {
+        serialized[i * 2] = name(i);
+        serialized[i * 2 + 1] = valueAsBytes(i);
+      }
+    }
+    return serialized;
+  }
+
+  /**
+   * Serialize all metadata entries, leaving some values as {@link InputStream}s.
```
dapengzhang0: s/Serialize/Serializes and ditto for verbs somewhere else.

api/src/main/java/io/grpc/Metadata.java
```diff
@@ -706,14 +815,36 @@ public String toString() {
      * @return a parsed instance of type T
      */
     abstract T parseBytes(byte[] serialized);
+
+    /**
+     * @return whether this key should be serialized to bytes lazily.
+     */
+    boolean serializeToStreams() {
```
dapengzhang0: What about `isSerializationLazy()` or `serializableToStreams()` or `serializesToStreams()`? The current name looks like "serialize to streams now and return true if successful".

api/src/main/java/io/grpc/Metadata.java
dapengzhang0: Also add javadoc for `InternalMetadata.serializePartial()`?

api/src/main/java/io/grpc/Metadata.java
```diff
@@ -706,14 +815,36 @@ public String toString() {
...
+      return false;
+    }
+
+    /**
+     * Get this keys (implementation-specific) marshaller, or null if the
+     * marshaller is not of the given time.
```
dapengzhang0: I don't understand `the given time` here, as  `marshaller` is a final field.

api/src/main/java/io/grpc/Metadata.java
```diff
@@ -118,34 +122,80 @@ public String parseAsciiString(String serialized) {
...
+   * @param usedNames the number of names
    */
   Metadata(int usedNames, byte[]... binaryValues) {
     assert (binaryValues.length & 1) == 0 : "Odd number of key-value pairs " + binaryValues.length;
     size = usedNames;
     namesAndValues = binaryValues;
   }
 
-  private byte[][] namesAndValues;
+  /**
```
dapengzhang0: This above constructor can reuse this constructor.

```java
  Metadata(int usedNames, byte[]... binaryValues) {
    this(usedNames, (Object[]) binaryValues);
  }
```

api/src/main/java/io/grpc/InternalMetadata.java
```diff
@@ -82,4 +82,19 @@ public static Metadata newMetadata(int usedNames, byte[]... binaryValues) {
...
+    return md.serializePartial();
+  }
+
+  @Internal
+  public static <T> Object parsedValue(Key<T> key, T value) {
+    return Metadata.LazyValue.create(key, value);
+  }
+
+  @Internal
+  public static Metadata newMetadataWithParsedValues(int usedNames, Object[] namesAndValues) {
```
dapengzhang0: Note that if the run time type of the passed array is `byte[]`, it will invoke the varargs constructor 
```java
Metadata(int usedNames, byte[]... binaryValues)
```

and fail.

dapengzhang0: As these methods are going to be used in other packages than `grpc-api`, better to have a javadoc, e.g. document the restriction on the argument `Object[] namesAndValues`.

api/src/main/java/io/grpc/Metadata.java
markb74: Done.

api/src/main/java/io/grpc/Metadata.java
markb74: Makeses sense. I used "serializesToStreams", though I was considering "shouldSerializeToStreams". lmk if you have a preference.

api/src/main/java/io/grpc/Metadata.java
markb74: Oops, that's a typo. time -> type.

api/src/main/java/io/grpc/Metadata.java
markb74: Done.

markb74: Done.

api/src/main/java/io/grpc/InternalMetadata.java
markb74: Done.

api/src/main/java/io/grpc/Metadata.java
```diff
@@ -706,14 +813,36 @@ public String toString() {
...
+
+    /**
+     * Gets this keys (implementation-specific) marshaller, or null if the
+     * marshaller is not of the given type.
+     *
+     * @param marshallerClass The type we expect the marshaller to be.
+     * @return the marshaller object for this key, or null.
+     */
+    @Nullable
+    final <M> M getMarshaller(Class<M> marshallerClass) {
```
dapengzhang0: Seems this method is only for binary stream marshaller, why not just
```java
    @Nullable
    final BinaryStreamMarshaller<T> getBinaryStreamMarshaller() {
      if (BinaryStreamMarshaller.class.isInstance(marshaller)) {
        return (BinaryStreamMarshaller<T>) marshaller;
      }
      return null;
    }
```

api/src/main/java/io/grpc/Metadata.java
```diff
@@ -734,12 +863,97 @@ T parseBytes(byte[] serialized) {
...
+        }
+      }
+      return serialized;
+    }
+
+    @SuppressWarnings({"unchecked", "ReferenceEquality"})
+    <T2> T2 toObject(Key<T2> key) {
+      if (key.serializesToStreams()) {
+        BinaryStreamMarshaller<T2> marshaller = getBinaryStreamMarshaller(key);
+        if (marshaller == this.marshaller) {
```
ejona86: I brought this up in the last review, but I think we were saying we weren't going to return back the same object by default, since if the object is mutable it may cause confusion. So I thought we would be removing this optimization here and always use the marshaller.parseStream(toStream()). Your particular Marshaller would do the protobuf trick to pull out the original object (or copy) during parseStream().

markb74: Oh whoops, yeah I'd misunderstood your comment. Removed.

api/src/main/java/io/grpc/InternalMetadata.java
```diff
@@ -82,4 +82,46 @@ public static Metadata newMetadata(int usedNames, byte[]... binaryValues) {
...
+  public static <T> Object parsedValue(Key<T> key, T value) {
+    return Metadata.LazyValue.create(key, value);
+  }
+
+  /**
+   * Creates a new {@link Metadata} instance from serialized data,
+   * with some values pre-parsed. Metadata will mutate the passed in array.
+   *
+   * @param usedNames The number of names used.
+   * @param namesAndValues An array of iterleaved names and values,
```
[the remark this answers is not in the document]
markb74: Done.

api/src/main/java/io/grpc/Metadata.java
```diff
@@ -118,34 +122,78 @@ public String parseAsciiString(String serialized) {
...
   Metadata(int usedNames, byte[]... binaryValues) {
-    assert (binaryValues.length & 1) == 0 : "Odd number of key-value pairs " + binaryValues.length;
+    this(usedNames, (Object[]) binaryValues);
+  }
+
+  /**
+   * Constructor called by the transport layer when it receives partially-parsed metadata.
+   * Metadata will mutate the passed in array.
+   */
+  Metadata(int usedNames, Object[] namesAndValues) {
```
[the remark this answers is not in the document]
markb74: Done.

api/src/main/java/io/grpc/Metadata.java
```diff
@@ -118,34 +122,78 @@ public String parseAsciiString(String serialized) {
...
+      return value;
+    } else {
+      return ((LazyValue<?>) value).toStream();
+    }
+  }
+
+  private <T> T valueAsT(int i, Key<T> key) {
+    Object value = value(i);
+    if (value instanceof byte[]) {
+      return key.parseBytes((byte[]) value(i));
```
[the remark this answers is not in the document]
markb74: good catch. Done.

api/src/main/java/io/grpc/Metadata.java
markb74: I wanted to avoid having the Key class directly depend on a single Marshaller type really. This approach let me avoid that, while supporting other marshaller types as well. If figured this approach was worth it since we'd discussed possibly adding a TextStreamMarshaller later on.

I'm happy either way though, so I'll defer to your preference.

ejona86: I saw the method and thought it was a bit odd, but it didn't see an obvious simple alternative that everyone would agree is superior. It is a bit weirder to be BinaryStreamMarshaller-specific as long as it is on the Key class. I'd be prone to moving the method to LazyStreamBinaryKey. But it just seemed beside the point and what is here is fine. We can always change it later.

api/src/main/java/io/grpc/Metadata.java
```diff
@@ -931,10 +936,7 @@ InputStream toStream() {
     <T2> T2 toObject(Key<T2> key) {
       if (key.serializesToStreams()) {
         BinaryStreamMarshaller<T2> marshaller = getBinaryStreamMarshaller(key);
-        if (marshaller == this.marshaller) {
-          // The marshallers match so T2 == T.
-          return (T2) value;
-        } else if (marshaller != null) {
+        if (marshaller != null) {
```
dapengzhang0: Can remove `@SuppressWarnings`?

markb74: Right, done.

ejona86:
@markb74, thank you! Thank you for working with us on the back-and-forth!

PR #6475, opened 2019-12-02
alts: fix lazy_channel close

jiangtaoli2016:
This was caught by security review.


alts/src/main/java/io/grpc/alts/internal/AltsProtocolNegotiator.java
```diff
@@ -298,6 +298,7 @@ synchronized Channel get() {
     synchronized void close() {
       if (channel != null) {
         channelPool.returnObject(channel);
+        channel = null;
```
sanjaypujare: You can also do:
```
channel = channelPool.returnObject(channel);
```
which is the recommended idiom.

jiangtaoli2016: Thanks for advice. Done.

jiangtaoli2016:
Thanks much for quick review @ejona86

PR #6499, opened 2019-12-08
api: Fix javadoc for ManagedChannelBuilder#forTarget(String)

ST-DDT:
The dns scheme is only the default scheme with grpc-java. Other
libraries could add more NameResolvers and thus change the default. For
compatibility reasons, the schema should therefore be specified
explicitly.

The current javadocs conflict with those of `NameResolverProdiver#priority()`, which allow setting a priority higher than DNS.

Users of grpc-spring-boot-starter who want to use a service discovery have the problem that many libraries don't provide a schema for DNS addresses and therefore don't work anymore without additional configuration.

https://github.com/yidongnan/grpc-spring-boot-starter/issues/268
https://github.com/yidongnan/grpc-spring-boot-starter/issues/304
https://github.com/yidongnan/grpc-spring-boot-starter/issues/300

### DIFF
--- a/api/src/main/java/io/grpc/ManagedChannelBuilder.java
+++ b/api/src/main/java/io/grpc/ManagedChannelBuilder.java
@@ -55,8 +55,11 @@ public abstract class ManagedChannelBuilder<T extends ManagedChannelBuilder<T>> 
    * </ul>
    *
    * <p>An authority string will be converted to a {@code NameResolver}-compliant URI, which has
-   * {@code "dns"} as the scheme, no authority, and the original authority string as its path after
-   * properly escaped. Example authority strings:
+   * the scheme from the name resolver with the highest priority (e.g. {@code "dns"}),
+   * no authority, and the original authority string as its path after properly escaped.
+   * We recommend specifying the schema explicitly if it is known, to avoid problems with other
+   * libraries that might contain additional NameResolvers.
+   * Example authority strings:
    * <ul>
    *   <li>{@code "localhost"}</li>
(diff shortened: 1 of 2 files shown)


dapengzhang0:
> Users of grpc-spring-boot-starter who want to use a service discovery have the problem that many libraries don't provide a schema for DNS addresses and therefore don't work anymore without additional configuration.
yidongnan/grpc-spring-boot-starter#268
yidongnan/grpc-spring-boot-starter#304
yidongnan/grpc-spring-boot-starter#300

@ejona86

api/src/main/java/io/grpc/ManagedChannelBuilder.java
```diff
@@ -55,8 +55,11 @@
    * </ul>
    *
    * <p>An authority string will be converted to a {@code NameResolver}-compliant URI, which has
-   * {@code "dns"} as the scheme, no authority, and the original authority string as its path after
-   * properly escaped. Example authority strings:
+   * the scheme from the name resolver with the highest priority (e.g. {@code "dns"}),
+   * no authority, and the original authority string as its path after properly escaped.
+   * We recommend specifying the schema explicitly if it is known, to avoid problems with other
```
ejona86: This is really working as designed. So I'm not wild about saying "problem." But I do agree it takes some coordination.

How about something like:

> We recommend libraries specify the schema explicitly if it is known, since libraries cannot know which NameResolver will be default during runtime.

ejona86:
Thank you!

PR #6500, opened 2019-12-08
xds: set UpstreamTlsContext in Attributes

sanjaypujare:
@dapengzhang0 pls check as per our discussion.

### DIFF
--- a/xds/src/main/java/io/grpc/xds/CdsLoadBalancer.java
+++ b/xds/src/main/java/io/grpc/xds/CdsLoadBalancer.java
@@ -41,6 +41,7 @@ import io.grpc.xds.XdsLoadBalancerProvider.XdsConfig;
 import io.grpc.xds.XdsSubchannelPickers.ErrorPicker;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.Nullable;
 
 /**
@@ -233,22 +234,27 @@ public final class CdsLoadBalancer extends LoadBalancer {
   @VisibleForTesting
   static class EdsLoadBalancingHelper extends ForwardingLoadBalancerHelper {
     private final Helper delegate;
-    private final UpstreamTlsContext upstreamTlsContext;
+    private final AtomicReference<UpstreamTlsContext> upstreamTlsContext;
 
-    EdsLoadBalancingHelper(Helper helper, UpstreamTlsContext upstreamTlsContext) {
+    EdsLoadBalancingHelper(Helper helper, AtomicReference<UpstreamTlsContext> upstreamTlsContext) {
       this.delegate = helper;
       this.upstreamTlsContext = upstreamTlsContext;
     }
 
     @Override
     public Subchannel createSubchannel(CreateSubchannelArgs createSubchannelArgs) {
-      createSubchannelArgs =
-          createSubchannelArgs.toBuilder()
-              .setAttributes(
-                  createSubchannelArgs.getAttributes().toBuilder()
-                      .set(XdsAttributes.ATTR_UPSTREAM_TLS_CONTEXT, upstreamTlsContext)
-                      .build())
-              .build();
+      if (upstreamTlsContext.get() != null) {
+        createSubchannelArgs =
+            createSubchannelArgs
+                .toBuilder()
+                .setAttributes(
+                    createSubchannelArgs
+                        .getAttributes()
+                        .toBuilder()
+                        .set(XdsAttributes.ATTR_UPSTREAM_TLS_CONTEXT, upstreamTlsContext.get())
+                        .build())
+                .build();
+      }
       return delegate.createSubchannel(createSubchannelArgs);
     }
 
@@ -260,7 +266,7 @@ public final class CdsLoadBalancer extends LoadBalancer {
 
   private final class ClusterWatcherImpl implements ClusterWatcher {
 
-    final Helper helper;
+    final EdsLoadBalancingHelper helper;
     final ResolvedAddresses resolvedAddresses;
 
     // EDS balancer for the cluster.
@@ -270,7 +276,7 @@ public final class CdsLoadBalancer extends LoadBalancer {
     LoadBalancer edsBalancer;
 
     ClusterWatcherImpl(Helper helper, ResolvedAddresses resolvedAddresses) {
-      this.helper = helper;
+      this.helper = new EdsLoadBalancingHelper(helper, new AtomicReference<UpstreamTlsContext>());
       this.resolvedAddresses = resolvedAddresses;
     }
 
@@ -288,13 +294,12 @@ public final class CdsLoadBalancer extends LoadBalancer {
           /* fallbackPolicy = */ null,
           /* edsServiceName = */ newUpdate.getEdsServiceName(),
           /* lrsServerName = */ newUpdate.getLrsServerName());
+      UpstreamTlsContext upstreamTlsContext = newUpdate.getUpstreamTlsContext();
+      if (upstreamTlsContext != null) {
+        helper.upstreamTlsContext.set(upstreamTlsContext);
+      }
       if (edsBalancer == null) {
-        Helper localHelper = helper;
-        UpstreamTlsContext upstreamTlsContext = newUpdate.getUpstreamTlsContext();
-        if (upstreamTlsContext != null) {
-          localHelper = new EdsLoadBalancingHelper(helper, upstreamTlsContext);
-        }
-        edsBalancer = lbRegistry.getProvider(XDS_POLICY_NAME).newLoadBalancer(localHelper);
+        edsBalancer = lbRegistry.getProvider(XDS_POLICY_NAME).newLoadBalancer(helper);
       }
       edsBalancer.handleResolvedAddresses(
           resolvedAddresses.toBuilder()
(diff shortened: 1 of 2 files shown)


xds/src/test/java/io/grpc/xds/CdsLoadBalancerTest.java
```diff
@@ -325,6 +330,88 @@ public void handleCdsConfigs() throws Exception {
...
+            .setEdsServiceName("edsServiceFoo.googleapis.com")
+            .setLbPolicy("round_robin")
+            .setEnableLrs(false)
+            .setUpstreamTlsContext(upstreamTlsContext)
+            .build());
+
+    assertThat(edsLbHelpers).hasSize(1);
+    assertThat(edsLoadBalancers).hasSize(1);
+    Helper edsLbHelper1 = edsLbHelpers.poll();
+    assertThat(edsLbHelper1).isInstanceOf(EdsLoadBalancingHelper.class);
```
dapengzhang0: We shouldn't assert this because this is an implementation detail that relies on the fact that `EdsLoadBalancingHelper` is the outermost forwarding helper, which may no longer be true if we were adding new features (with new forwarding helpers). As you are verifying the behavior of `createSubchannel` later, this line is not needed.

xds/src/test/java/io/grpc/xds/CdsLoadBalancerTest.java
```diff
@@ -325,6 +330,88 @@ public void handleCdsConfigs() throws Exception {
...
+  private static class FakeSocketAddress extends SocketAddress {
+    final String name;
+
+    FakeSocketAddress(String name) {
+      this.name = name;
+    }
+
+    @Override
+    public String toString() {
+      return "FakeSocketAddress-" + name;
```
dapengzhang0: Where is this used? Why create `FakeSocketAddress` class?

xds/src/test/java/io/grpc/xds/CdsLoadBalancerTest.java
```diff
@@ -325,6 +330,88 @@ public void handleCdsConfigs() throws Exception {
...
+    assertThat(edsLbHelper1).isInstanceOf(EdsLoadBalancingHelper.class);
+
+    LoadBalancer.CreateSubchannelArgs createSubchannelArgs =
+        LoadBalancer.CreateSubchannelArgs.newBuilder()
+            .setAddresses(new EquivalentAddressGroup(new FakeSocketAddress("fake")))
+            .build();
+    ArgumentCaptor<LoadBalancer.CreateSubchannelArgs> createSubchannelArgsCaptor1 =
+        ArgumentCaptor.forClass(null);
+    verify(helper, never())
+        .createSubchannel(any(LoadBalancer.CreateSubchannelArgs.class)); // remove
```
dapengzhang0: `// remove` You wanted to remove the comment?

xds/src/main/java/io/grpc/xds/CdsLoadBalancer.java
```diff
@@ -230,6 +232,34 @@ public void handleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
...
   }
 
+  @VisibleForTesting
+  static class EdsLoadBalancingHelper extends ForwardingLoadBalancerHelper {
+    private final Helper delegate;
+    private final UpstreamTlsContext upstreamTlsContext;
+
+    EdsLoadBalancingHelper(Helper helper, UpstreamTlsContext upstreamTlsContext) {
+      this.delegate = helper;
+      this.upstreamTlsContext = upstreamTlsContext;
```
dapengzhang0: Should this really be final?

dapengzhang0: Maybe an `AtomicReference<UpstreamTlsContext>` so that it can be updated?

xds/src/test/java/io/grpc/xds/CdsLoadBalancerTest.java
sanjaypujare: This is used on line 392. `CreateSubchannelArgs` has to have non-null `List<EquivalentAddressGroup> addrs` and the `EquivalentAddressGroup` has to have a non-null `SocketAddress` :-( 

I'll be happy to get rid of this if there is a better way to test `createSubchannel`.

xds/src/test/java/io/grpc/xds/CdsLoadBalancerTest.java
sanjaypujare: But this is unit-testing which implies I am testing this unit as a white-box and the test is aware of the internal implementation detail and taking advantage of it to test various conditions. Once new features/implementation are added the test will also need to be updated.

xds/src/main/java/io/grpc/xds/CdsLoadBalancer.java
sanjaypujare: If you see where this is called (line 297), a new instance of `EdsLoadBalancingHelper` is created whenever a new LoadBalancer is created and takes the delegate helper and `UpstreamTlsContext` as arguments to the constructor. There are multiple instances as needed. Okay to make them final.

dapengzhang0: `onClusterChanged()` can be called multiple times. Each time it may contain a different `UpstreamTlsContext`, but your helper is only created for the first time. So I suggested using `AtomicReference<UpstreamTlsContext>`.

xds/src/test/java/io/grpc/xds/CdsLoadBalancerTest.java
dapengzhang0: Is it possible to simply use `new InetSocketAddress("foo.example.com", 8080)`?

xds/src/test/java/io/grpc/xds/CdsLoadBalancerTest.java
dapengzhang0: My point is this does not really test any behavior, just verifying the implementation, and if implementation changes, this will also need to be updated. So the line is not needed. While the behavior verification below does not need any change as long as the spec is unchanged.

xds/src/test/java/io/grpc/xds/CdsLoadBalancerTest.java
sanjaypujare: Sure. `FakeSocketAddress` is defined and used quite a lot in various places in the gRPC test code, so I thought that's the more desirable pattern. May be we should have a single `FakeSocketAddress` defined that can be used in all such places.

Problem with `new InetSocketAddress("foo.example.com", 8080)` is an additional dependency on `java.net.InetSocketAddress` (not a big deal) and an unnecessary `getHostByName` call to resolve `foo.example.com` which I can avoid by using `InetAddress.anyLocalAddress()` instead of `foo.example.com` .

xds/src/main/java/io/grpc/xds/CdsLoadBalancer.java
sanjaypujare: Oh okay, good point. But then this needs a bigger change than just using  `AtomicReference<UpstreamTlsContext>`. Specifically when `edsBalancer != null` (line 292) I need to access the `Helper` inside the existing `edsBalancer`. Isn't that right?

dapengzhang0: One way is to declare the helper in `ClusterWatcherImpl` and init it in the constructor

```java
final EdsLoadBalancingHelper helper;
ClusterWatcherImpl(Helper helper, ResolvedAddresses resolvedAddresses) {
      this.helper = new EdsLoadBalancingHelper(helper, new AtomicReference<>());
      this.resolvedAddresses = resolvedAddresses;
}
```

sanjaypujare: @dapengzhang0 after talking to @ejona86 I looked at the code and I need to set the `Attributes` on each EAG in the list inside `CreateSubchannelArgs`. I made that change and also took care of null as we discussed.

xds/src/main/java/io/grpc/xds/CdsLoadBalancer.java
```diff
@@ -228,9 +234,60 @@ public void handleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
...
+    }
+
+    @Override
+    public Subchannel createSubchannel(CreateSubchannelArgs createSubchannelArgs) {
+      if (upstreamTlsContext.get() != null) {
+        createSubchannelArgs =
+            createSubchannelArgs
+                .toBuilder()
+                .setAttributes(createSubchannelArgs.getAttributes())
```
dapengzhang0: This is not needed now.

xds/src/main/java/io/grpc/xds/CdsLoadBalancer.java
```diff
@@ -228,9 +234,60 @@ public void handleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
     }
   }
 
+  @VisibleForTesting
```
dapengzhang0: It can be `private static final` now.

xds/src/main/java/io/grpc/xds/CdsLoadBalancer.java
sanjaypujare: I need to copy attributes from the original `createSubchannelArgs` to the new one I am building - otherwise I could be breaking somebody else's code. Note I am not adding anything.

dapengzhang0: It's already copied by default I believe.

sanjaypujare: you are right - I missed that!

PR #6502, opened 2019-12-09
xds: update deprecated charset import

ericgribkoff:


ericgribkoff:
All the references new to the current import are removed; lots of others exist in older but I'll take care of those in a separate PR (so they will not be included with the current import).

ejona86:
Note that we have generally just let these be because the import isn't available on many versions of Android we support. Although this code is fine to use the new import, it can be hard to deal with differences between different folders and so have generally left everything mostly supporting older Android.

So that is to say, these were fine as-is. These changes are fine, but I'm not sure we're ready to "upgrade" the rest.

sanjaypujare:
Just confirming my understanding: so for older Android code the Guava import is better than the `java.nio` import for the `UTF_8` charset?

ejona86:
If you try to use the `java.nio` import for `UTF_8` in code that runs on Android, you should get an animalsniffer error saying it isn't available. In that case, yes, Guava is clearly better.

PR #6508, opened 2019-12-10
netty: NettyServer should close ProtocolNegotiator

ejona86:
CC @jiangtaoli2016 @ZhenLian. Note that without this ALTS Server leaked its Channel


jiangtaoli2016:
Thanks @ejona86  for the fix.

ZhenLian:
Thanks for the quick fix @ejona86 ! Since now we also clean up the resource on server side, maybe the comments here : https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/ProtocolNegotiator.java#L40-L44 could also be updated?

ejona86:
@creamsoup, PTAL

Issue #6511, opened 2019-12-11
Example HelloWorldServer does not gracefully shutdown


dounan:
### What version of gRPC-Java are you using?
On a fork up at commit https://github.com/grpc/grpc-java/commit/d168632f8229e387c19e820c26cf1d0f25bb0c84

### What is your environment?
Mac OSX 10.14.6
Java 11.0.2

### What did you expect to see?
The server should allow in-flight RPCs to finish

### What did you see instead?
The client receives error `WARNING: RPC 1 failed: Status{code=UNAVAILABLE, description=Network closed for unknown reason, cause=null}`

### Steps to reproduce the bug
- Add a call to `Thread.sleep(15000)` in `io.grpc.examples.helloworld.HelloWorldServer.GreeterImpl.sayHello`
- Build and run the server
- Run the client
- Ctrl+C the server once it gets the request

It seems like `server.shutdown()` does not wait for in-flight RPCs to finish before returning. The documentation makes it seem like that it should block. 

I am able to "fix" the issue by adding a call to `server.awaitTermination()` in the `stop` method.

sanjaypujare:
This is behaving as designed/expected. The doc for `server.shutdown()` does say "..preexisting calls continue..." but does not imply that `shutdown()` blocks until the calls end. So `shutdown` will return after initiating orderly shutdown but you need to wait using `server.awaitTermination()` for the inflight calls to finish. If the doc seems unclear/ambiguous suggest new wording.

dounan:
Would it make sense to update all the example code servers to reflect the fact that `awaitTermination()` needs to be called? Both that and the wording on the documentation is a bit misleading. I can try putting together a PR for that change

sanjaypujare:
> Would it make sense to update all the example code servers to reflect the fact that `awaitTermination()` needs to be called? 

Seems like a good idea since `awaitTermination()` is currently not called in case of control-C termination (shutdown-hook).

> Both that and the wording on the documentation is a bit misleading. I can try putting together a PR for that change

Even if not misleading, more clarification cannot hurt.

dounan:
Created a PR here: https://github.com/grpc/grpc-java/pull/6512

ejona86:
Taken from #6512:
Hmmm... something else is going on. The blockUntilShutdown() in the main() should be enough...

[...]

PR #6512, opened 2019-12-11
Gracefully shutdown example servers

dounan:
The example servers do not properly wait until preexisting calls complete before shutting down (https://github.com/grpc/grpc-java/issues/6511).

This PR
- Adds a call to `awaitTermination` to wait for preexisting calls to complete (with a 30s timeout)
- Updates the API documentation for `Server.shutdown` to clarify that the call does not wait for preexisting calls to complete before returning.

### DIFF
--- a/api/src/main/java/io/grpc/Server.java
+++ b/api/src/main/java/io/grpc/Server.java
@@ -107,7 +107,7 @@ public abstract class Server {
    * After this call returns, this server has released the listening socket(s) and may be reused by
    * another server.
    *
-   * Note that this method will not wait for preexisting calls to finish before returning.
+   * <p>Note that this method will not wait for preexisting calls to finish before returning.
    * {@link #awaitTermination()} or {@link #awaitTermination(long, TimeUnit)} needs to be called to
    * wait for existing calls to finish.
    *
--- a/api/src/main/java/io/grpc/Server.java
+++ b/api/src/main/java/io/grpc/Server.java
@@ -104,8 +104,12 @@ public abstract class Server {
 
   /**
    * Initiates an orderly shutdown in which preexisting calls continue but new calls are rejected.
-   * This call will not wait for preexisting calls finish before returning. After this call returns,
-   * this server has released the listening socket(s) and may be reused by another server.
+   * After this call returns, this server has released the listening socket(s) and may be reused by
+   * another server.
+   *
+   * Note that this method will not wait for preexisting calls to finish before returning.
+   * {@link #awaitTermination()} or {@link #awaitTermination(long, TimeUnit)} needs to be called to
+   * wait for existing calls to finish.
    *
    * @return {@code this} object
    * @since 1.0.0
--- a/api/src/main/java/io/grpc/Server.java
+++ b/api/src/main/java/io/grpc/Server.java
@@ -104,8 +104,8 @@ public abstract class Server {
 
   /**
    * Initiates an orderly shutdown in which preexisting calls continue but new calls are rejected.
-   * After this call returns, this server has released the listening socket(s) and may be reused by
-   * another server.
+   * This call will not wait for preexisting calls finish before returning. After this call returns,
+   * this server has released the listening socket(s) and may be reused by another server.
    *
    * @return {@code this} object
    * @since 1.0.0
--- a/examples/src/main/java/io/grpc/examples/advanced/HelloJsonServer.java
+++ b/examples/src/main/java/io/grpc/examples/advanced/HelloJsonServer.java
@@ -60,19 +60,19 @@ public class HelloJsonServer {
       public void run() {
         // Use stderr here since the logger may have been reset by its JVM shutdown hook.
         System.err.println("*** shutting down gRPC server since JVM is shutting down");
-        HelloJsonServer.this.stop();
+        try {
+          HelloJsonServer.this.stop();
+        } catch (InterruptedException e) {
+          e.printStackTrace();
+        }
         System.err.println("*** server shut down");
       }
     });
   }
 
-  private void stop() {
+  private void stop() throws InterruptedException {
     if (server != null) {
-      try {
-        server.shutdown().awaitTermination(30, TimeUnit.SECONDS);
-      } catch (InterruptedException e) {
-        logger.warning(e.getMessage());
-      }
+      server.shutdown().awaitTermination(30, TimeUnit.SECONDS);
     }
   }
 
--- a/examples/src/main/java/io/grpc/examples/advanced/HelloJsonServer.java
+++ b/examples/src/main/java/io/grpc/examples/advanced/HelloJsonServer.java
@@ -28,6 +28,7 @@ import io.grpc.examples.helloworld.HelloRequest;
 import io.grpc.stub.ServerCalls.UnaryMethod;
 import io.grpc.stub.StreamObserver;
 import java.io.IOException;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 
 /**
@@ -67,7 +68,11 @@ public class HelloJsonServer {
 
   private void stop() {
     if (server != null) {
-      server.shutdown();
+      try {
+        server.shutdown().awaitTermination(30, TimeUnit.SECONDS);
+      } catch (InterruptedException e) {
+        logger.warning(e.getMessage());
+      }
     }
   }
 
--- a/examples/src/main/java/io/grpc/examples/advanced/HelloJsonServer.java
+++ b/examples/src/main/java/io/grpc/examples/advanced/HelloJsonServer.java
@@ -63,7 +63,7 @@ public class HelloJsonServer {
         try {
           HelloJsonServer.this.stop();
         } catch (InterruptedException e) {
-          e.printStackTrace();
+          e.printStackTrace(System.err);
         }
         System.err.println("*** server shut down");
       }
--- a/examples/src/main/java/io/grpc/examples/experimental/CompressingHelloWorldServerAllMethods.java
+++ b/examples/src/main/java/io/grpc/examples/experimental/CompressingHelloWorldServerAllMethods.java
@@ -17,6 +17,7 @@
 package io.grpc.examples.experimental;
 
 import java.io.IOException;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 
 import io.grpc.Metadata;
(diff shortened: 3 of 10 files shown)


api/src/main/java/io/grpc/Server.java
```diff
@@ -104,8 +104,8 @@ public int getPort() {
 
   /**
    * Initiates an orderly shutdown in which preexisting calls continue but new calls are rejected.
```
ejona86: I think "initiates" is the key word here.

I think this would probably go better as a second paragraph that is merely informative. Something like:

> Note that this method does not wait for the server to stop, so existing calls may continue after returning. To stop existing calls use {`@link` #shutdownNow()} and to wait for the server to fully stop use {`@link` #awaitTermination()}.

api/src/main/java/io/grpc/Server.java
```diff
@@ -104,8 +104,8 @@ public int getPort() {
...
-   * After this call returns, this server has released the listening socket(s) and may be reused by
-   * another server.
+   * This call will not wait for preexisting calls finish before returning. After this call returns,
+   * this server has released the listening socket(s) and may be reused by another server.
```
sanjaypujare: I suggest the following:

>  Initiates an orderly shutdown to allow preexisting calls to continue but new calls are
rejected. {@link #awaitTermination()} or {@link #awaitTermination(long, TimeUnit)} needs 
to be called to wait for existing calls to finish. After this call returns, this server
has released the listening socket(s) and may be reused by another server.

dounan:
Thanks for taking a look @ejona86!

> Hmmm... something else is going on. The blockUntilShutdown() in the main() should be enough...

From my understanding, the `blockUntilShutdown()` is useful to keep the Java application running so that the gRPC can serve requests, which is a bit different than handling the JVM's shutdown process.

When the JVM is instructed to shutdown (such as via a SIGINT), it will call all registered shutdown handlers that are registered via [`addShutdownHook`](https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#addShutdownHook-java.lang.Thread-):

```
Runtime.getRuntime().addShutdownHook(new Thread() {
    @Override
    public void run() {
        ...
    }
});
```

Once all those `run` methods return, the JVM will shutdown.

For this reason, it is important for the shutdown hook's `run()` method to not return until you are ok with the JVM shutting down, hence the additional call to `server.awaitTermination()`

sanjaypujare:
> Hmmm... something else is going on. The blockUntilShutdown() in the main() should be enough...

When you press control-C, it invokes the shutdown-hook. So `blockUntilShutdown()` in the `main()` is skipped right? In the original issue @dounan mentioned pressing control-C.

ejona86:
> Once all those run methods return, the JVM will shutdown.

That's not true. On ctrl-c, the JVM simply calls `start()` on the threads registered, and nothing really more. What causes JVM shutdown is when there are no non-daemon threads. The main thread is non-daemon, so it should keep the JVM alive.

ejona86:
I'm going to move the conversation back to issue #6511.

api/src/main/java/io/grpc/Server.java
dounan: Combined your and @sanjaypujare's suggestions: https://github.com/grpc/grpc-java/pull/6512/commits/af9a157931771751dacddc334f578c557e6afaac

api/src/main/java/io/grpc/Server.java
dounan: Combined your and @ejona86's suggestions: https://github.com/grpc/grpc-java/pull/6512/commits/af9a157931771751dacddc334f578c557e6afaac

api/src/main/java/io/grpc/Server.java
```diff
@@ -104,8 +104,12 @@ public int getPort() {
 
   /**
    * Initiates an orderly shutdown in which preexisting calls continue but new calls are rejected.
-   * This call will not wait for preexisting calls finish before returning. After this call returns,
-   * this server has released the listening socket(s) and may be reused by another server.
+   * After this call returns, this server has released the listening socket(s) and may be reused by
+   * another server.
+   *
+   * Note that this method will not wait for preexisting calls finish before returning.
```
sanjaypujare: preexisting calls finish => preexisting calls to finish

dounan: Doh! Good catch

examples/build.gradle
```diff
@@ -22,7 +22,7 @@ targetCompatibility = 1.7
 
 // Feel free to delete the comment at the next line. It is just for safely
 // updating the version in our release process.
-def grpcVersion = '1.27.0-SNAPSHOT' // CURRENT_GRPC_VERSION
+def grpcVersion = '1.24.2' // CURRENT_GRPC_VERSION
```
ejona86: This will need to be reverted.

examples/src/main/java/io/grpc/examples/advanced/HelloJsonServer.java
```diff
@@ -67,7 +68,11 @@ public void run() {
 
   private void stop() {
     if (server != null) {
-      server.shutdown();
+      try {
+        server.shutdown().awaitTermination(30, TimeUnit.SECONDS);
+      } catch (InterruptedException e) {
+        logger.warning(e.getMessage());
```
ejona86: So logging here is useless if called from a shutdown hook, because the logging system itself is shut down via a shutdown hook (and so this log goes nowhere).

It is probably best to do the awaitTermination()/blockUntilShutdown() from the shutdown hook and handle InterruptedException via `ex.printStackTrace()`. The other option would be to do awaitTermination() within stop() but have it throw InterruptedException, and then handle the InterruptedException within the hook via `ex.printStackTrace()`.

I really want to rewrite/reorganize the examples, since the start()/stop() methods just make things more complicated without a major benefit.

examples/src/main/java/io/grpc/examples/helloworld/HelloWorldServer.java
```diff
@@ -33,30 +34,33 @@
   private void start() throws IOException {
     /* The port on which the server should run */
     int port = 50051;
-    server = ServerBuilder.forPort(port)
-        .addService(new GreeterImpl())
-        .build()
-        .start();
+    server = ServerBuilder.forPort(port).addService(new GreeterImpl()).build().start();
```
ejona86: revert

examples/src/main/java/io/grpc/examples/helloworld/HelloWorldServer.java
```diff
@@ -33,30 +34,33 @@
...
     logger.info("Server started, listening on " + port);
     Runtime.getRuntime().addShutdownHook(new Thread() {
       @Override
       public void run() {
-        // Use stderr here since the logger may have been reset by its JVM shutdown hook.
+        // Use stderr here since the logger may have been reset by its JVM shutdown
```
ejona86: Revert. We use 100 column lines.

examples/src/main/java/io/grpc/examples/helloworld/HelloWorldServer.java
```diff
@@ -77,6 +81,13 @@ public static void main(String[] args) throws IOException, InterruptedException
 
     @Override
     public void sayHello(HelloRequest req, StreamObserver<HelloReply> responseObserver) {
+      logger.info("!!! Received call");
```
ejona86: Revert.

examples/build.gradle
dounan: Ah sry I accidentally pushed a bunch of random testing code In that last commit.  I'll get that cleaned up

dounan: Done

examples/src/main/java/io/grpc/examples/advanced/HelloJsonServer.java
dounan: Done in https://github.com/grpc/grpc-java/pull/6512/commits/8e75b2dd4ebe7db8fa0cb951d0868853c81b0350

examples/src/main/java/io/grpc/examples/helloworld/HelloWorldServer.java
dounan: Done

examples/src/main/java/io/grpc/examples/helloworld/HelloWorldServer.java
dounan: Done

examples/src/main/java/io/grpc/examples/helloworld/HelloWorldServer.java
dounan: Done

examples/src/test/java/io/grpc/examples/routeguide/RouteGuideServerTest.java
```diff
@@ -86,7 +86,11 @@ public void setUp() throws Exception {
 
   @After
   public void tearDown() {
-    server.stop();
+    try {
+		  server.stop();
```
ejona86: Indentation needs to be fixed.

dounan: How did those tabs get there...fixed! https://github.com/grpc/grpc-java/pull/6512/commits/91391d2017a0481fbf29f7ecebf510d1e4abfe04

examples/src/main/java/io/grpc/examples/experimental/CompressingHelloWorldServerAllMethods.java
```diff
@@ -62,15 +63,19 @@ private void start() throws IOException {
       public void run() {
         // Use stderr here since the logger may have been reset by its JVM shutdown hook.
         System.err.println("*** shutting down gRPC server since JVM is shutting down");
-        CompressingHelloWorldServerAllMethods.this.stop();
+        try {
+          CompressingHelloWorldServerAllMethods.this.stop();
+        } catch (InterruptedException e) {
+          e.printStackTrace();
```
sanjaypujare: Couple of comments:
- do you want to print to stderr in keeping with lines 65 and 71?
`e.printStackTrace(System.err)`

- do you want to mention/comment saying this is because the logger is closed as part of shutdown?

and of course make this change everywhere.

dounan: > do you want to print to stderr in keeping with lines 65 and 71?
e.printStackTrace(System.err)

Done in https://github.com/grpc/grpc-java/pull/6512/commits/d997e5be9504f61d4f9fe23998cb400899a95274

I also went ahead and fixed the call to `logger` in `ManualFlowControlServer`

> do you want to mention/comment saying this is because the logger is closed as part of shutdown?

The call to `System.err.println` above already has this comment 😄

PR #6521, opened 2019-12-14
examples: Throw exception in test instead of printStackTrace

ejona86:
Throwing makes cleaner code and also is more helpful if the exception is ever
thrown, as the error will be more clear.

-----

I noticed this after #6512 was already merged. Minor, but good to fix in an example.

CC @dounan

### DIFF
--- a/examples/src/test/java/io/grpc/examples/routeguide/RouteGuideServerTest.java
+++ b/examples/src/test/java/io/grpc/examples/routeguide/RouteGuideServerTest.java
@@ -85,12 +85,8 @@ public class RouteGuideServerTest {
   }
 
   @After
-  public void tearDown() {
-    try {
-      server.stop();
-    } catch (InterruptedException e) {
-      e.printStackTrace();
-    }
+  public void tearDown() throws Exception {
+    server.stop();
   }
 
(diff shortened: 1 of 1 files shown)


examples/src/test/java/io/grpc/examples/routeguide/RouteGuideServerTest.java
```diff
@@ -85,12 +85,8 @@ public void setUp() throws Exception {
   }
 
   @After
-  public void tearDown() {
-    try {
-      server.stop();
-    } catch (InterruptedException e) {
-      e.printStackTrace();
-    }
+  public void tearDown() throws Exception {
```
sanjaypujare: Couple of points:
- when Exception does happen, it will fail the test (before it would have passed). I think that's better but making sure that's what you want.

- `throws` is wider than it needs to be (`InterruptedException`).

ejona86: Yes, I want the test to fail. Yes, I purposefully used Exception, since that is pretty normal in tests since most exceptions should just cause the test to fail.

PR #6529, opened 2019-12-17
alts: add client authorization util library

jiangtaoli2016:
Add a simple authorization utility library. 

@ejona86

### DIFF
--- a/alts/src/main/java/io/grpc/alts/AuthorizationUtil.java
+++ b/alts/src/main/java/io/grpc/alts/AuthorizationUtil.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2019 The gRPC Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.grpc.alts;
+
+import io.grpc.ServerCall;
+import io.grpc.Status;
+import io.grpc.alts.internal.AltsAuthContext;
+import io.grpc.alts.internal.AltsProtocolNegotiator;
+import java.util.List;
+
+/** Utility class for ALTS client authorization. */
+public final class AuthorizationUtil {
+
+  private AuthorizationUtil() {}
+
+  /**
+   * Given a server call, performs client authorization check, i.e., checks if the client service
+   * account matches one of the expected service accounts. It returns OK if client is authorized and
+   * an error otherwise.
+   */
+  public static Status clientAuthorizationCheck(
+      ServerCall<?, ?> call, List<String> expectedServiceAccounts) {
+    AltsAuthContext altsContext =
+        (AltsAuthContext) call.getAttributes().get(AltsProtocolNegotiator.AUTH_CONTEXT_KEY);
+    if (altsContext == null) {
+      return Status.NOT_FOUND.withDescription("Peer ALTS AuthContext not found");
+    }
+    if (expectedServiceAccounts.contains(altsContext.getPeerServiceAccount())) {
+      return Status.OK;
+    }
+    return Status.PERMISSION_DENIED.withDescription(
+        "Client " + altsContext.getPeerServiceAccount() + " is not authorized");
+  }
+}
--- a/alts/src/main/java/io/grpc/alts/AuthorizationUtil.java
+++ b/alts/src/main/java/io/grpc/alts/AuthorizationUtil.java
@@ -20,7 +20,7 @@ import io.grpc.ServerCall;
 import io.grpc.Status;
 import io.grpc.alts.internal.AltsAuthContext;
 import io.grpc.alts.internal.AltsProtocolNegotiator;
-import java.util.List;
+import java.util.Collection;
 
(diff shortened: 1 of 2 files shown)


alts/src/main/java/io/grpc/alts/AuthorizationUtil.java
```diff
@@ -0,0 +1,49 @@
...
+public final class AuthorizationUtil {
+
+  private AuthorizationUtil() {}
+
+  /**
+   * Given a server call, performs client authorization check, i.e., checks if the client service
+   * account matches one of the expected service accounts.
+   */
+  public static Status clientAuthorizationCheck(
+      ServerCall<?, ?> call, ImmutableList<String> expectedServiceAccounts) {
```
ejona86: Don't expose ImmutableList in an API surface, especially since the list isn't stored anywhere. You probably just want List or Collection, or even Iterable.

alts/src/main/java/io/grpc/alts/AuthorizationUtil.java
```diff
@@ -0,0 +1,49 @@
...
+    AltsAuthContext altsCtx =
+        (AltsAuthContext) call.getAttributes().get(AltsProtocolNegotiator.AUTH_CONTEXT_KEY);
+    if (altsCtx == null) {
+      return Status.NOT_FOUND.withDescription("Peer ALTS AuthContext not found");
+    }
+    for (String serviceAccount : expectedServiceAccounts) {
```
ejona86:
```java
if (expectedServiceAccounts.contains(altsCtx.getPeerServiceAccount())) {
  return Status.OK;
}
```

alts/src/main/java/io/grpc/alts/AuthorizationUtil.java
```diff
@@ -0,0 +1,49 @@
...
+  private AuthorizationUtil() {}
+
+  /**
+   * Given a server call, performs client authorization check, i.e., checks if the client service
+   * account matches one of the expected service accounts.
+   */
+  public static Status clientAuthorizationCheck(
+      ServerCall<?, ?> call, ImmutableList<String> expectedServiceAccounts) {
+    AltsAuthContext altsCtx =
+        (AltsAuthContext) call.getAttributes().get(AltsProtocolNegotiator.AUTH_CONTEXT_KEY);
```
ejona86: Why isn't `AUTH_CONTEXT_KEY` a `Attributes.Key<AltsAuthContext>`?

alts/src/main/java/io/grpc/alts/AuthorizationUtil.java
```diff
@@ -0,0 +1,49 @@
...
+import io.grpc.alts.internal.AltsAuthContext;
+import io.grpc.alts.internal.AltsProtocolNegotiator;
+
+/** Utility class for ALTS client authorization. */
+public final class AuthorizationUtil {
+
+  private AuthorizationUtil() {}
+
+  /**
+   * Given a server call, performs client authorization check, i.e., checks if the client service
```
ejona86: Document what it returns (namely, that it returns OK if client is authorized and an error otherwise).

alts/src/test/java/io/grpc/alts/AuthorizationUtilTest.java
```diff
@@ -0,0 +1,114 @@
...
+                .build();
+        AltsAuthContext altsAuthContext = new AltsAuthContext(handshakerResult);
+        attrsBuilder.set(AltsProtocolNegotiator.AUTH_CONTEXT_KEY, altsAuthContext);
+      }
+      attrs = attrsBuilder.build();
+    }
+
+    @Override
+    public void request(int numMessages) {
+      throw new AssertionError("Should not be called");
```
ejona86: A trick you can do for these is to extend ForwardingServerCall and then implement `delegate()` to throw like you're doing here. This is fine, though.

alts/src/main/java/io/grpc/alts/AuthorizationUtil.java
```diff
@@ -0,0 +1,49 @@
...
+   * account matches one of the expected service accounts.
+   */
+  public static Status clientAuthorizationCheck(
+      ServerCall<?, ?> call, ImmutableList<String> expectedServiceAccounts) {
+    AltsAuthContext altsCtx =
```
elharo: altsContext because no abbreviations in variable names per Google java style

alts/src/main/java/io/grpc/alts/AuthorizationUtil.java
jiangtaoli2016: Done.

alts/src/main/java/io/grpc/alts/AuthorizationUtil.java
jiangtaoli2016: changed to List

alts/src/main/java/io/grpc/alts/AuthorizationUtil.java
jiangtaoli2016: Done.

alts/src/test/java/io/grpc/alts/AuthorizationUtilTest.java
jiangtaoli2016: Ack on the trick. Let's keep the current FakeServerCall if you don't mind.

alts/src/main/java/io/grpc/alts/AuthorizationUtil.java
jiangtaoli2016: This because `extractPeerObject()` returns an Object, rather than AltAuthContext. Internally, we have a different AuthContext class. We need to use share the same API for returning the peer object. Let me know if you have better ideas.

jiangtaoli2016:
thanks @ejona86 @elharo for comments.

@ejona86 Please take a look again.

jiangtaoli2016:
Changed to Collection instead of List.

PR #6531, opened 2019-12-17
xds: allow publishing grpc-xds to maven local

dapengzhang0:


xds/build.gradle
```diff
@@ -95,4 +95,4 @@ publishing {
         }
     }
 }
-[publishMavenPublicationToMavenRepository, publishMavenPublicationToMavenLocal]*.onlyIf { false }
+[publishMavenPublicationToMavenRepository]*.onlyIf { false }
```
sanjaypujare: Is this change enough? I looked at `../build.gradle` and it has on lines 435-437:
```
    [publishMavenPublicationToMavenRepository, publishMavenPublicationToMavenLocal]*.onlyIf {
        !name.contains("grpc-gae-interop-testing") && !name.contains("grpc-xds")
    }
```

That needs fixing too, right? There could be other changes needed as well in the parent gradle file.

dapengzhang0:
Only need change this subproject directory, because only this subproject was blacklisted.

xds/build.gradle
dapengzhang0: I don't find these lines in `../build.gradle`. Are you looking at master branch?

sanjaypujare: You are right - I was looking at an older branch and it looks like these lines have been removed. Sorry.

PR #6533, opened 2019-12-17
xds: use UpstreamTlsContext from attributes if available

sanjaypujare:


sanjaypujare:
> LGTM
> An integration test need be added later. It's very hard to see whether it will work.

Agreed. Now that the XdsClientImpl + XdsResolver integration is done I will create a test to verify the change made here.

PR #6543, opened 2019-12-18
bazel: Use https to download from Maven Central

### LINKED ISSUE #6536: repositories.bzl uses http (not https)
I see a lot of lines like this:

        server_urls = ["http://central.maven.org/maven2"],

If we're actually resolving these they're going to stop working on January 15, 2020. 

I'm not 100% sure this matters, but just in case I wanted to raise the issue.

ejona86:
central.maven.org shouldn't have been used in the first place as it isn't one
of the canonical URLs to Maven Central, but even more importantly we want to
use https. The previous URL will probably stop working January 15, 2020[1][2].

Fixes #6536

1. https://central.sonatype.org/articles/2019/Apr/30/http-access-to-repo1mavenorg-and-repomavenapacheorg-is-being-deprecated/
2. https://central.sonatype.org/articles/2019/Nov/15/non-canonical-urls-will-be-redirected-today/

### DIFF
--- a/repositories.bzl
+++ b/repositories.bzl
@@ -165,7 +165,7 @@ def com_google_android_annotations():
     jvm_maven_import_external(
         name = "com_google_android_annotations",
         artifact = "com.google.android:annotations:4.1.1.4",
-        server_urls = ["http://central.maven.org/maven2"],
+        server_urls = ["https://repo.maven.apache.org/maven2/"],
         artifact_sha256 = "ba734e1e84c09d615af6a09d33034b4f0442f8772dec120efb376d86a565ae15",
         licenses = ["notice"],  # Apache 2.0
     )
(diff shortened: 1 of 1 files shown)


repositories.bzl
```diff
@@ -165,7 +165,7 @@ def com_google_android_annotations():
     jvm_maven_import_external(
         name = "com_google_android_annotations",
         artifact = "com.google.android:annotations:4.1.1.4",
-        server_urls = ["http://central.maven.org/maven2"],
+        server_urls = ["https://repo.maven.apache.org/maven2/"],
```
sanjaypujare: Is "https://repo.maven.apache.org/maven2/" preferred over "https://repo1.maven.org/maven2" ? Wondering why not use a "maven.org" URL.

dapengzhang0: Seems they are resolved to different ip addresses, can we add both?

ejona86: https://central.sonatype.org/articles/2018/Dec/17/ssl-endpoints/ talks about the two canonical URLs. I learned of repo.maven.apache.org as part of https://github.com/gradle/gradle/issues/3463 . It's hard to tell exactly, but it seems _Maven_ recommends repo.maven.apache.org (maybe because they control it) but Sonatype (Maven Central itself) tends to recommend repo1.maven.org (maybe because they control it, or they have out-dated documentation).

ejona86: To be more clear: yes, maven.org is not controlled by the Maven project, but instead a commercial entity that grew out of the Maven project.

sanjaypujare: Adding both seems to be a good idea if they are actually different servers (I thought one of them was just a redirect).

ejona86: Maven and Gradle both only use a single of the two URLs. I don't see much advantage in using both. Both should be behind a CDN, so each probably have many IPs.

ejona86:
I've backported this as far as 1.23, since this will probably blind-side users. I can backport further back if necessary.

Issue #6507, opened 2019-12-10
Allow creating context with more than 4 keys


nathanhleung:
### Is your feature request related to a problem?
I'd like to create a context with more than 4 keys (limit seems to be 4 according to: https://github.com/grpc/grpc-java/blob/master/context/src/main/java/io/grpc/Context.java#L359)

### Describe the solution you'd like
Ideally, an overload to `withValues` that potentially could take a `Map<Key, V>` of Context keys and values, enabling an arbitrary number of context keys to be used in a single context if a consumer of this library wanted more than 4

### Describe alternatives you've considered
We could just use a single Context key and store all of our context in a single map associated with that key, but before I do that I wanted to understand if there was a reason for limiting `withValues` to taking 4 key-value pairs

ejona86:
You can just chain together multiple withValues.

```java
context.withValues(K1, V1, K2, V2);
// is same as
context.withValues(K1, V1).withValues(K2, V2);
```

The only difference is that the first produces a bit less garbage and in the initial implementation of Context was slightly more efficient to do lookups (since keys were stored as a linked list).

I will note that Context should not be treated like a general purpose map, throwing lots of stuff in it. That is to say, combine multiple related items together into a single key instead of having them separate. But if there are two unrelated things, have two keys for them. `Key`s should virtually always be constants.

nathanhleung:
Ok, thanks! I'd be open to creating a PR to add overloads for `withValues` for more key-value pairs (up to 4, 5, etc) if that would help, or updating the Javadoc of the existing `withValues` method with the example you just mentioned. What would you prefer?

ejona86:
A PR to update the `withValues` Javadoc to say it is equivalent of the separate calls sounds good.

nathanhleung:
Opened #6545

ran-su:
#6545 added detailed documents for `withValues`. So this issue is resolved.

PR #6545, opened 2019-12-18
Update javadoc for withValue(s) to indicate withValue can be chained

nathanhleung:
See #6507 for discussion

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -326,6 +326,19 @@ public class Context {
    *   });
    * </pre>
    *
+   * <p>Note that multiple calls to {@link #withValue} can be chained together.
+   * That is,
+   *
+   * <pre>
+   * context.withValues(K1, V1, K2, V2);
+   * // is the same as
+   * context.withValue(K1, V1).withValue(K2, V2);
+   * </pre>
+   *
+   * <p>Nonetheless, {@link Context} should not be treated like a general purpose
+   * map with a large number of keys and values — combine multiple related items
+   * together into a single key instead of separating them. But if the items
+   * are unrelated, have separate keys for them.
    */
   public <V> Context withValue(Key<V> k1, V v1) {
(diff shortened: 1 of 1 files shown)


context/src/main/java/io/grpc/Context.java
```diff
@@ -326,6 +326,19 @@ public CancellableContext withDeadline(Deadline newDeadline, ScheduledExecutorSe
    *   });
    * </pre>
    *
+   * Note that multiple calls to {@link #withValue} can be chained together.
```
ejona86: Needs `<p>`. Ditto for "Nonetheless". See Travis failure.

nathanhleung:
Oops, thanks -- fixed and rebased

ran-su:
@nathanhleung Merged. Thanks.

context/src/main/java/io/grpc/Context.java
```diff
@@ -326,6 +326,19 @@ public CancellableContext withDeadline(Deadline newDeadline, ScheduledExecutorSe
...
    * </pre>
    *
+   * <p>Note that multiple calls to {@link #withValue} can be chained together.
+   * That is,
+   *
+   * <pre>
+   * context.withValues(K1, V1, K2, V2);
+   * // is the same as
+   * context.withValue(K1, V1).withValue(K2, V2);
+   * </pre>
```
bogdandrutu: @ejona86 I thought that the first is more performant, if that is the case we should add that to the comment.

ejona86: The memory/performance difference isn't as much now that we have the HashArrayMappedTrie. We had different code paths when we used the linked list. With the new implementation values are packed together automatically, and the double-value version just does two adds internally:

https://github.com/grpc/grpc-java/blob/a8bdf5cb1abf7a857e263a5d1500f1bba2c940e3/context/src/main/java/io/grpc/Context.java#L354

The only difference between the two would be an additional Context allocation. Most users shouldn't concern themselves with that difference.

PR #6549, opened 2019-12-18
grpclb: support explicit fallback from LB

creamsoup:
FYI, the [proto](https://github.com/grpc/grpc-proto/blob/master/grpc/lb/v1/load_balancer.proto) is not updated yet.

### DIFF
--- a/grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
+++ b/grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
@@ -582,6 +582,7 @@ final class GrpclbState {
       }
 
       if (typeCase == LoadBalanceResponseTypeCase.FALLBACK_RESPONSE) {
+        cancelFallbackTimer();
         useFallbackBackends();
         maybeUpdatePicker();
         return;
--- a/grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
+++ b/grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
@@ -581,7 +581,11 @@ final class GrpclbState {
         return;
       }
 
-      if (typeCase != LoadBalanceResponseTypeCase.SERVER_LIST) {
+      if (typeCase == LoadBalanceResponseTypeCase.FALLBACK_RESPONSE) {
+        useFallbackBackends();
+        maybeUpdatePicker();
+        return;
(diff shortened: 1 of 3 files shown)


grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
```diff
@@ -581,7 +581,11 @@ private void handleResponse(LoadBalanceResponse response) {
         return;
       }
 
-      if (typeCase != LoadBalanceResponseTypeCase.SERVER_LIST) {
+      if (typeCase == LoadBalanceResponseTypeCase.FALLBACK_RESPONSE) {
+        useFallbackBackends();
+        maybeUpdatePicker();
```
zhangkun83: Do you want to call `cancelFallbackTimer()`?

grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancerTest.java
```diff
@@ -2084,6 +2084,207 @@ public void retrieveModeFromLbConfig_badConfigDefaultToRoundRobin() throws Excep
...
+    // enter fallback mode
+    lbResponseObserver.onNext(buildLbFallbackResponse());
+
+    // existing subchannels must be returned immediately to gracefully shutdown.
+    verify(subchannelPool)
+        .returnSubchannel(eq(subchannel1), eq(ConnectivityStateInfo.forNonError(READY)));
+    verify(subchannelPool)
+        .returnSubchannel(eq(subchannel2), eq(ConnectivityStateInfo.forNonError(READY)));
+
+    // verify fallback (no backends)
```
zhangkun83: The more common case is that there are fallback backends.

grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
creamsoup: done.

grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancerTest.java
creamsoup: done.

creamsoup:
although this change is backward compatible, waiting for the proto to be published.

PR #6553, opened 2019-12-19
bazel: Support maven_install

### LINKED ISSUE #5359: bazel: migrate to deps of maven_install

ejona86:
maven_install is strongly superior to previous forms of grabbing dependencies
from Maven as it computes the appropriate versions to use from the full
transitive dependencies of all libraries used by an application. It also has
much less boilerplate and includes dependencies with generated targets.

In the future we will drop the jvm_maven_import_external usages and require
maven_install, at which point we can swap to using the `@maven' repository and
no longer depend on compat_repositories.

Fixes #5359

-----

There are three commits. They are best reviewed separately. The http_repository change is non-essential, but I noticed it when working on this and didn't bother splitting it out to a separate PR.


examples/WORKSPACE
```diff
@@ -1,35 +1,56 @@
...
+load("@io_grpc_grpc_java//:repositories.bzl", "IO_GRPC_GRPC_JAVA_ARTIFACTS")
+load("@io_grpc_grpc_java//:repositories.bzl", "IO_GRPC_GRPC_JAVA_OVERRIDE_TARGETS")
+
+maven_install(
+    artifacts = [
+        "com.google.api.grpc:grpc-google-cloud-pubsub-v1:0.1.24",
+        "com.google.api.grpc:proto-google-cloud-pubsub-v1:0.1.24",
+    ] + IO_GRPC_GRPC_JAVA_ARTIFACTS,
+    generate_compat_repositories = True,
+    maven_install_json = "//:maven_install.json",
```
dapengzhang0: Could this be too much for examples? It may add difficulty for users to copy the example and make changes. Even the examples of `rules_jvm_external` itself are not always pinning the maven artifacts. https://github.com/bazelbuild/rules_jvm_external/tree/master/examples

repositories.bzl
```diff
@@ -37,115 +82,78 @@ IO_GRPC_GRPC_JAVA_OVERRIDE_TARGETS = {
...
-        omit_org_apache_commons_lang3 = False,
-        omit_org_codehaus_mojo_animal_sniffer_annotations = False):
+def grpc_java_repositories():
     """Imports dependencies for grpc-java."""
-    if not omit_com_google_android_annotations:
+    if not native.existing_rule("com_google_protobuf"):
+        com_google_protobuf()
+    if not native.existing_rule("com_google_protobuf_javalite"):
+        com_google_protobuf_javalite()
+    if not native.existing_rule("io_grpc_grpc_proto"):
```
dapengzhang0: Not in dictionary order?

repositories.bzl
```diff
@@ -37,115 +82,78 @@ IO_GRPC_GRPC_JAVA_OVERRIDE_TARGETS = {
...
+        com_squareup_okhttp_okhttp()
+    if not native.existing_rule("com_squareup_okio_okio"):
+        com_squareup_okio_okio()
+    if not native.existing_rule("io_netty_netty_buffer"):
+        io_netty_netty_buffer()
+    if not native.existing_rule("io_netty_netty_common"):
+        io_netty_netty_common()
+    if not native.existing_rule("io_netty_netty_transport"):
+        io_netty_netty_transport()
+    if not native.existing_rule("io_netty_netty_transport_native_epoll_linux_x86_64"):
```
dapengzhang0: Not in dictionary order?

repositories.bzl
ejona86: I don't understand. I don't see anything wrong here and below.

examples/WORKSPACE
ejona86: Hmm... Good point. I'd rather have the examples be as "real" as possible, and I expect most Bazel users will want to pin. I've added a comment to mention how to regenerate the pinned versions.

I'm fine with dropping this if it becomes a problem, but I think I'd prefer to "do it real" first and remove it if it becomes a problem.

ejona86:
Reminder: coordinate with #6574

repositories.bzl
dapengzhang0: I mean the artifact names `"com_google_protobuf_javalite"`, `"io_grpc_grpc_proto"`, `"com_google_android_annotations"`... are not sorted in alphabetic order. It will be easier to look up if listing them in alphabetic order.

ejona86: Oh, you are comparing with annotations. These three were pulled from the rest because they are different (and see the extra newline after io_grpc_grpc_proto). These three are Bazel-aware repositories, the rest are Maven repositories. Once we require maven_install we will delete the rest, but leave these.

repositories.bzl
ejona86: I see now that codec is going backward. I had seen buffer, common, transport, transport_native_epoll and thought they were in order.

I didn't rearrange these from the previous version, but since the diff is already poor here but I'm also not doing anything special I'll go ahead and put the netty artifacts in sorted order. But I'm not going to rearrange them below; I don't want to create the diff noise and these will be deleted in a few months anyway.

ejona86:
So the Bazel failure has shown me that the version pinning here will actually be pretty annoying, because it includes all transitive dependencies and a `__AUTOGENERATED_FILE_DO_NOT_MODIFY_THIS_FILE_MANUALLY` hash of some sort. That will prevent cherry-picking and cause merge conflicts. I'm going to drop the version pinning.

ejona86:
@dapengzhang0, PTAL, just to confirm you are fine with me removing the pinning.

I'm going to need to pull in the version change from #6572 as well.

PR #6561, opened 2019-12-20
Clarify CallStreamObserver's Javadoc

### LINKED ISSUE #6552: Clarify the Javadoc for CallStreamObserver
The relevant part of the current Javadoc:
https://github.com/grpc/grpc-java/blob/6b04fc18f30deb3519c389fe457b99cf5a8d99d0/stub/src/main/java/io/grpc/stub/CallStreamObserver.java#L25-L35

As far as I can tell, for any call, there are three implementations of `StreamObserver` involved:
 1) Implemented by the application, on the RPC client side ('inbound')
 2) Implemented by the framework, on the RPC client side ('outbound'), `ClientCallStreamObserver`
 3) Implemented by the framework, on the RPC server side ('outbound'), `ServerCallStreamObserver`

I think it would be nice to clarify that. I've spent significant time trying to understand where does `ClientCallStreamObserver` appear, being confused by the premise that there are only two implementations of `StreamObserver`.

leventov:
Fixes #6552.

Apart the class-level comment, there are a few unclear places in the documentation to the methods which I don't know how to improve--please make your suggestions.

1. `... without requiring excessive buffering internally ...` in `isReady()` suggests that there may be some internal buffer, in-process, which may have spare space, and then `isReady()` can return true. However, the documentation `setOnReadyHandler()` has phrase `when peer is ready to receive more messages` which doesn't consider the above option.

2. Phrase `On server-side it may only be called during the initial call to the application, before the service returns its {@code StreamObserver}.` -- cannot make sense of this phrase, because service implementations don't normally return `StreamObserver` instances. Does it have something to do with `ClientStreamingMethod` and `BidiStreamingMethod`, the `StreamObserver`s returned by them?

3. In the documentation for `disableAutoInboundFlowControl()`, in the first phrase `Disables automatic flow control where a token is returned to the peer after a call to the 'inbound' {@link io.grpc.stub.StreamObserver#onNext(Object)} has completed.` it's unclear what does the phrase `a token is returned to the peer` mean, what is the "token" there.

### DIFF
--- a/stub/src/main/java/io/grpc/stub/CallStreamObserver.java
+++ b/stub/src/main/java/io/grpc/stub/CallStreamObserver.java
@@ -19,20 +19,25 @@ package io.grpc.stub;
 import io.grpc.ExperimentalApi;
 
 /**
- * A refinement of StreamObserver provided by the GRPC runtime to the application that allows for
- * more complex interactions with call behavior.
+ * A refinement of StreamObserver provided by the GRPC runtime to the application (the client or
+ * the server) that allows for more complex interactions with call behavior.
  *
- * <p>In any call there are logically two {@link StreamObserver} implementations:
+ * <p>In any call there are logically three {@link StreamObserver} implementations:
  * <ul>
- *   <li>'inbound' - which the GRPC runtime calls when it receives messages from the
- *   remote peer. This is implemented by the application.
+ *   <li>'inbound' - which the GRPC runtime calls when it receives messages from the server. This
+ *   is implemented by the client application.
  *   </li>
- *   <li>'outbound' - which the GRPC runtime provides to the application which it uses to
- *   send messages to the remote peer.
+ *   <li>'outbound', client-side - which the GRPC runtime provides to the client application and the
+ *   client uses this {@code StreamObserver} to send messages to the server.
+ *   </li>
+ *   <li>'outbound', server-side - which the GRPC runtime provides to the server application and
+ *   the server uses this {@code StreamObserver} to send messages (responses) to the client.
  *   </li>
  * </ul>
  *
- * <p>Implementations of this class represent the 'outbound' message stream.
+ * <p>Implementations of this class represent the 'outbound' message streams. The client-side
+ * one is {@link ClientCallStreamObserver} and the service-side one is
+ * {@link ServerCallStreamObserver}.
  *
  * <p>Like {@code StreamObserver}, implementations are not required to be thread-safe; if multiple
  * threads will be writing to an instance concurrently, the application must synchronize its calls.
--- a/stub/src/main/java/io/grpc/stub/CallStreamObserver.java
+++ b/stub/src/main/java/io/grpc/stub/CallStreamObserver.java
@@ -22,13 +22,18 @@ import io.grpc.ExperimentalApi;
  * A refinement of StreamObserver provided by the GRPC runtime to the application (the client or
  * the server) that allows for more complex interactions with call behavior.
  *
- * <p>In any call there are logically three {@link StreamObserver} implementations:
+ * <p>In any call there are logically four {@link StreamObserver} implementations:
  * <ul>
- *   <li>'inbound' - which the GRPC runtime calls when it receives messages from the server. This
- *   is implemented by the client application.
+ *   <li>'inbound', client-side - which the GRPC runtime calls when it receives messages from
+ *   the server. This is implemented by the client application and passed into a service method
+ *   on a stub object.
  *   </li>
  *   <li>'outbound', client-side - which the GRPC runtime provides to the client application and the
  *   client uses this {@code StreamObserver} to send messages to the server.
+ *   </li>
+ *   <li>'inbound', server-side - which the GRPC runtime calls when it receives messages from
+ *   the client. This is implemented by the server application and returned from service
+ *   implementations of client-side streaming and bidirectional streaming methods.
  *   </li>
  *   <li>'outbound', server-side - which the GRPC runtime provides to the server application and
  *   the server uses this {@code StreamObserver} to send messages (responses) to the client.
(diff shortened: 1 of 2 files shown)


stub/src/main/java/io/grpc/stub/CallStreamObserver.java
```diff
@@ -19,20 +19,25 @@
...
- * more complex interactions with call behavior.
+ * A refinement of StreamObserver provided by the GRPC runtime to the application (the client or
+ * the server) that allows for more complex interactions with call behavior.
  *
- * <p>In any call there are logically two {@link StreamObserver} implementations:
+ * <p>In any call there are logically three {@link StreamObserver} implementations:
  * <ul>
- *   <li>'inbound' - which the GRPC runtime calls when it receives messages from the
- *   remote peer. This is implemented by the application.
+ *   <li>'inbound' - which the GRPC runtime calls when it receives messages from the server. This
```
ran-su: This 'inbound' `StreamObserver` concept also exists in server-side. As an example, you can check [here](https://github.com/grpc/grpc-java/blob/6b04fc18f30deb3519c389fe457b99cf5a8d99d0/examples/src/main/java/io/grpc/examples/manualflowcontrol/ManualFlowControlServer.java#L38).

leventov: Thanks @ran-su for pointing to this example. A comment in this example suggests that isReady may turn to false not only when the remote peer (the client) is not able to receive more messages, but also when they have sent us a lot of messages to overflow the `receive buffer`. Should it be reflected in the Javadoc for `CallStreamObserver.isReady()` or `setOnReadyHandler()`? Or was it actually "send buffer", not "receive buffer" which should be mentioned in this comment:
https://github.com/grpc/grpc-java/blob/9e02cf089ee8b3ff635c8531039cfe87da417779/examples/src/main/java/io/grpc/examples/manualflowcontrol/ManualFlowControlServer.java#L92-L98

ran-su: In here, server is only response for manage its own ability to receive messages. If client-side's send buffer fills up there are other system mechanisms to handle it. 

This example is manual flow control example, which may be considered advanced usage of gRPC. IMO, since this is not for all gRPC users, keep all related information in a more detailed example may be better for both user who don't need this feature and for user who actually want this feature.

leventov: How a series of onNext() -> request(1) -> onNext() -> request(1) calls may lead to overflow of server's receive buffer?

ran-su: In the code I linked, `disableAutoInboundFlowControl()` is called above which disables the ` onNext() -> request(1) -> onNext() -> request(1)` cycle. In the case of `disableAutoInboundFlowControl`, it up to the user of gRPC to call `request()`.

leventov: In the comment that I excerpted, I think it talks about manual calls already. `serverCallStreamObserver.request(1);`, the very next line after that comment. So yes, user calls `serverCallStreamObserver.request(1)` (strictly once, as ensured in this line:
https://github.com/grpc/grpc-java/blob/9e02cf089ee8b3ff635c8531039cfe87da417779/examples/src/main/java/io/grpc/examples/manualflowcontrol/ManualFlowControlServer.java#L61

Then gRPC runtime calls incoming StreamObserver's `onNext()`, and so on, in a cycle (as far as I understand). How the incoming server's receive buffer may steadily grow and become overflown in this case?

The comment I excerpted above would make more sense to me if it was talking about server's send buffer, e. g. that gRPC runtime makes the server to throttle by means of setting `isReady` to false if the server's send buffer is full, e. g. client doesn't keep up with responses.

ran-su: In the example we are demonstrating a correct way of using those features. If the user is not careful, they may do something equal to `request(some_very_big_number)`, in that case the server receive buffer may get full.

leventov: @ran-su thanks. I've updated a comment in ManualFlowControlServer.java hopefully to reflect that less ambiguously. I've also removed the usage of `AtomicBoolean` there because it took me quite some time of attempts to understand why a CAS operation was needed there in onReadyHandler before I realized that was just a coding convenience.

I've also updated the Javadoc for `CallStreamObserver`, please review.

Questions 1. and 3. from [the message above](https://github.com/grpc/grpc-java/pull/6561#issue-355749164) also remain.

creamsoup:
@ran-su, is this pr ready to be merged?

creamsoup:
merged, thanks @leventov!

leventov:
Thanks for review and merge.

Could you please also kindly review questions 1. and 3. in the [top message](https://github.com/grpc/grpc-java/pull/6561#issue-355749164) in this PR?

creamsoup:
> 1. `... without requiring excessive buffering internally ...` in `isReady()` suggests that there may be some internal buffer, in-process, which may have spare space, and then `isReady()` can return true. However, the documentation `setOnReadyHandler()` has phrase `when peer is ready to receive more messages` which doesn't consider the above option.

very good catch! in general the doc in `ClientCall` or `ServerCall` is more accurate. although, they should agree each other =p So, the explanation in `isReady()` is correct. the return statement in the `setOnReadyHandler()` is incorrect and it should be fixed.

> 3. In the documentation for `disableAutoInboundFlowControl()`, in the first phrase `Disables automatic flow control where a token is returned to the peer after a call to the 'inbound' {@link io.grpc.stub.StreamObserver#onNext(Object)} has completed.` it's unclear what does the phrase `a token is returned to the peer` mean, what is the "token" there.

yes. this is indeed confusing. the `disableAutoInboundFlowControl()` is disabling per message "token" e.g. `request(1)`. not necessarily means http2 flow control. we are (or more like were) trying to change to `disableAutoRequest()`. see discussion in https://github.com/grpc/grpc-java/issues/1788. anyways, basically the "token" is number of messages.

PR #6563, opened 2019-12-20
grpclb: add child_policy_config_target_field_name to grpclb config

creamsoup:


grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -124,25 +124,27 @@ public void requestConnection() {
...
       if (rawChildPolicies == null) {
-        return DEFAULT_MODE;
+        return GrpclbConfig.create(DEFAULT_MODE);
       }
       List<LbConfig> childPolicies =
           ServiceConfigUtil.unwrapLoadBalancingConfigList(checkObjectList(rawChildPolicies));
+
       for (LbConfig childPolicy : childPolicies) {
         String childPolicyName = childPolicy.getPolicyName();
+        String target = (String) childPolicy.getRawConfigValue().get("targetName");
```
zhangkun83: Define a constant for "targetName".

grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancerTest.java
```diff
@@ -2017,71 +2017,90 @@ private static Attributes eagAttrsWithToken(String token) {
...
     assertThat(logs).containsExactly("DEBUG: grpclb ignoring unsupported child policy nono");
-    assertThat(mode).isEqualTo(Mode.PICK_FIRST);
+    assertThat(config.getMode()).isEqualTo(Mode.PICK_FIRST);
+    assertThat(config.getTarget()).isNull();
+  }
+
+  @Test
+  public void retrieveModeFromLbConfig_skipUnsupportedChildPolicyWithTarget() throws Exception {
+    String lbConfig = "{\"childPolicy\" : "
+        + "[ {\"nono\" : {}}, {\"pick_first\" : {\"targetName\": \"foo.google.com\"} } ]}";
```
zhangkun83: "targetName" should be one the same level of "childPolicy".

creamsoup: done.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancerProvider.java
```diff
@@ -77,33 +79,38 @@ public ConfigOrError parseLoadBalancingPolicyConfig(
   ConfigOrError parseLoadBalancingConfigPolicyInternal(
       Map<String, ?> rawLoadBalancingPolicyConfig) {
     if (rawLoadBalancingPolicyConfig == null) {
-      return ConfigOrError.fromConfig(DEFAULT_MODE);
+      return ConfigOrError.fromConfig(GrpclbConfig.create(DEFAULT_MODE));
+    }
+    Object rawTarget =
+        rawLoadBalancingPolicyConfig.get(GrpclbLoadBalancerProvider.SERVICE_CONFIG_TARGET_NAME);
+    String target = null;
+    if (rawTarget instanceof String) {
```
zhangkun83: If `rawTarget` is not a string, an error should be returned instead of silently ignored. You could either cast without checking, or add an else-branch to return the error.

creamsoup: `GrpclbLoadBalancer#retrieveGrpcLbConfig` always expects valid config even if it is invalid. this was trying matching the same behavior. just like childPolicy mode falls back to DEFAULT_MODE if unknown mode is passed which should be error? I can see we at least should have same type, but it is still weird.

zhangkun83: The current behavior you described was implemented before Service Config Error Handling was designed. We may want to revisit it now. @ejona86 may have some opinions.

ejona86: Failing if wrong type sounds correct, but I'm not familiar with the two different code paths here (why there are two code paths; I can tell one is old and one is new, but I'm still surprised). So I don't know if this change change short-term or long-term.

creamsoup: both are old code ways, the only difference i can tell is one logs and the other doesn't.
in long term, we should throw and handle service config error instead of falling back to default. given grpclb's status i am not sure it will happen or not.

zhangkun83: No, this file is the new code path, and the one in `GrpclbLoadBalancer` is the old one. The new one is not used until the new `ManagedChannelImpl2` is used (#6564). Once the old `ManagedChannelImpl` is replaced by the new one, the old path can be deleted. I expected this to happen soon as we just need to gain confidence from internal testing and integration testing.

creamsoup: i migrated the parsing in LB, and made this parse **very strict**. following suggestions involves behavior changes, and i want to get some feedback. @ejona86 

- return `defaultChildPolicy` (round robin?) if no child policy specified. we still should return Error if the `childPolicy` contains only unknown policies.
- previously, it doesn't ignore unknown policy. it falls back to default but it didn't fall back to next specified policy in the `childPolicy`. Since service config error handling design skips unknown lb policy, it should follow the same logic. It should skip unknown until it sees supported child policies (pick first and round robin)

grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancerProviderTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+    assertThat(configOrError.getConfig()).isNotNull();
+    GrpclbConfig config = (GrpclbConfig) configOrError.getConfig();
+    assertThat(config.getMode()).isEqualTo(Mode.ROUND_ROBIN);
+    assertThat(config.getTarget()).isNull();
+  }
+
+  @Test
+  public void retrieveModeFromLbConfig_emptyChildPolicyUseRoundRobin() throws Exception {
+    String lbConfig = "{\"childPolicy\" : []}";
```
ejona86: This one may be best to use default policy. Because of the proto conversation, we should treat "missing", "null", and "default value" for a key the same. (Although I know we don't do that very many places.)

grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancerProviderTest.java
```diff
@@ -0,0 +1,159 @@
...
+  @Test
+  public void retrieveModeFromLbConfig_skipUnsupportedChildPolicy() throws Exception {
+    String lbConfig = "{\"childPolicy\" : [ {\"nono\" : {}}, {\"pick_first\" : {} } ]}";
+
+    ConfigOrError configOrError =
+        provider.parseLoadBalancingPolicyConfig(parseJsonObject(lbConfig));
+
+    assertThat(configOrError.getError()).isNotNull();
+    assertThat(configOrError.getError().getDescription())
+        .contains("Unknown grpclb childPolicy: nono");
```
ejona86: This should use pick_first. TODO: make sure there is cross-language agreement on this behavior.

Ditto below.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancerProvider.java
creamsoup: i'll summarize this change in other comment.

creamsoup:
config parsing behavior changes:

- if `childPolicy` has only unknown policies, it will return Error (previously `ROUND_ROBIN`).
- instead of only using the 1st child policy, it looks until it sees known policy.

see `GrpclbLoadBalancerProviderTest`as examples.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancerProvider.java
```diff
@@ -38,7 +39,9 @@
  */
 @Internal
 public final class GrpclbLoadBalancerProvider extends LoadBalancerProvider {
+
   private static final Mode DEFAULT_MODE = Mode.ROUND_ROBIN;
+  private static final String SERVICE_CONFIG_SERVICE_NAME = "serviceName";
```
ejona86: Inline this literal?

grpclb/src/test/java/io/grpc/grpclb/GrpclbLoadBalancerProviderTest.java
creamsoup: talked to @markdroth, c core is already doing what we want to do (skip unknown policy).

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -114,11 +105,10 @@ public void handleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
     newLbAddressGroups = Collections.unmodifiableList(newLbAddressGroups);
     List<EquivalentAddressGroup> newBackendServers =
         Collections.unmodifiableList(resolvedAddresses.getAddresses());
-    Map<String, ?> rawLbConfigValue = attributes.get(ATTR_LOAD_BALANCING_CONFIG);
-    Mode newMode = retrieveModeFromLbConfig(rawLbConfigValue, helper.getChannelLogger());
-    if (!mode.equals(newMode)) {
-      mode = newMode;
-      helper.getChannelLogger().log(ChannelLogLevel.INFO, "Mode: " + newMode);
+    GrpclbConfig newConfig = (GrpclbConfig) resolvedAddresses.getLoadBalancingPolicyConfig();
+    if (config == null || !config.equals(newConfig)) {
```
ejona86: If `config == null`, don't you want to use `GrpclbConfig.create(Mode.ROUND_ROBIN)`? This was previously handled in the retrieveModeFromLbConfig method. It doesn't look like `this.config` should ever be `null`, since `GrpclbState` can't handle a null config.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancerProvider.java
ejona86: Oh, I see now that Kun asked for the literal. I think we probably want to get away from having those literals for strings only used one place.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancerProvider.java
```diff
@@ -78,31 +80,42 @@ public ConfigOrError parseLoadBalancingPolicyConfig(
   ConfigOrError parseLoadBalancingConfigPolicyInternal(
       Map<String, ?> rawLoadBalancingPolicyConfig) {
     if (rawLoadBalancingPolicyConfig == null) {
-      return ConfigOrError.fromConfig(DEFAULT_MODE);
+      return ConfigOrError.fromConfig(GrpclbConfig.create(DEFAULT_MODE));
     }
+    Object rawServiceName = rawLoadBalancingPolicyConfig.get("serviceName");
```
ejona86: I'd have expected this here:
```java
String serviceName = JsonUtil.getString(rawLoadBalancingPolicyConfig, "serviceName");
```

It produces a better error message when the value is not the correct type.

But it also looks like this code was avoiding the JsonUtil (e.g., the `getList()` just below). I think that may have been because it was ServiceConfigUtil before...? But it uses `ServiceConfigUtil.unwrapLoadBalancingConfigList`...

creamsoup: done. i am not sure why it was not using `JsonUtil`.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
creamsoup: thanks, done.

PR #6580, opened 2019-12-31
core: Simplify ServiceConfigUtil via utility methods

ejona86:
This should be functionally equivalent. More can be done, but wanted all the
changes to be "obvious" when reading them.

-----

I expect to have a followup where I inline the one-line methods into their callers, which should greatly reduce the size of ServiceConfigUtil. We can also move methods only used one place to their consumer. I think then we'd be left with just a handful of methods left, which is fine.

### DIFF
--- a/core/src/main/java/io/grpc/internal/JsonUtil.java
+++ b/core/src/main/java/io/grpc/internal/JsonUtil.java
@@ -16,19 +16,23 @@
 
 package io.grpc.internal;
 
+import static com.google.common.math.LongMath.checkedAdd;
+
+import java.text.ParseException;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 
 /**
- * Helper utility to work with JSON values in Java types.
+ * Helper utility to work with JSON values in Java types. Includes the JSON dialect used by
+ * Protocol Buffers.
  */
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/grpc/internal/JsonUtil.java
```diff
@@ -146,4 +212,138 @@ public static Boolean getBoolean(Map<String, ?> obj, String key) {
...
+    }
+    return true;
+  }
+
+  /**
+   * Returns the sum of {@code a} and {@code b} unless it would overflow or underflow in which case
+   * {@code Long.MAX_VALUE} or {@code Long.MIN_VALUE} is returned, respectively.
+   *
+   * <p>Copy of {@link com.google.common.math.LongMath#saturatedAdd}.</p>
+   *
```
creamsoup: nit: remove empty line.

ejona86: Hmmm... that's preexisting. I'd rather leave it as-is in this PR simply because any changes in this block of code are impossible to spot. I'd be happy to do it as a followup if we care.

ejona86:
This has low chance of causing conflict with xds work, so merging. Inlining the methods will need to be a bit more coordinated to avoid causing merge conflicts with inflight PRs.

PR #6585, opened 2020-01-06
interop-testing: Add XdsTestClient and XdsTestServer

ericgribkoff:


interop-testing/build.gradle
```diff
@@ -24,6 +24,7 @@ dependencies {
             project(':grpc-protobuf'),
             project(':grpc-stub'),
             project(':grpc-testing'),
+	    project(':grpc-xds'),
```
dapengzhang0: Not aligned.

interop-testing/src/main/proto/grpc/testing/messages.proto
```diff
@@ -160,3 +162,13 @@ message ReconnectInfo {
...
 }
+
+message LoadBalancerStatsRequest {
+  int32 num_rpcs = 1;
+  int32 timeout_sec = 2;
+}
+
+message LoadBalancerStatsResponse {
+  map<string, int32> rpcs_by_peer = 1;
+  int32 num_without_peer = 2;
```
dapengzhang0: What does this field mean?

interop-testing/src/main/java/io/grpc/testing/integration/XdsTestServer.java
```diff
@@ -0,0 +1,142 @@
...
+
+  private class TestServiceImpl extends TestServiceGrpc.TestServiceImplBase {
+    private final String host;
+
+    private TestServiceImpl() {
+      String tmpHost;
+      try {
+        tmpHost = InetAddress.getLocalHost().getHostName();
+      } catch (UnknownHostException e) {
+        tmpHost = "failed to get host " + e;
```
dapengzhang0: Why not fail the entire test immediately? Or just log the error and let host name be empty.

interop-testing/src/main/java/io/grpc/testing/integration/XdsTestClient.java
```diff
@@ -0,0 +1,336 @@
...
+        synchronized (lock) {
+          currentRequestId += 1;
+          requestId = currentRequestId;
+          savedWatchers.addAll(watchers);
+        }
+
+        SimpleRequest request = SimpleRequest.newBuilder().setFillServerId(true).build();
+        ManagedChannel channel = chans.get((int) (requestId % chans.size()));
+        final ClientCall<SimpleRequest, SimpleResponse> call =
+            channel.newCall(
```
dapengzhang0: Why not just call the rpc method with stub?

ericgribkoff: To be able to invoke `call.getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR)`. If there's a simpler way to get access to the call attributes I'd be happy to change this.

dapengzhang0: An alternative is using interceptor, it may not be much simpler though.

```java
AtomicReference<ClientCall> clientCallRef = new AtomicReference<>();
stub = stub.withDeadlineAfter().withInterceptors(
  new ClientInterceptor() {
      @Override
        ClientCall<ReqT, RespT> call = next.newCall(method,callOptions);
        clientCallRef.set(call);
        return call;
  });

try {
  SimpleResponse response = stub.unaryCall(request);
  serverId = response.getServerId();
  if (printResponse) {
    System.out.println(
       "Greeting: Hello world, this is "
          + response.getHostname()
          + ", from "
          + clientCallRef.get().getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR));
  }
} catch(RuntimeException e) {
} finally {
  for (XdsStatsWatcher watcher : savedWatchers) {
    watcher.rpcCompleted(requestId, serverId);
  }
}
```

ericgribkoff: Yeah I'm not sure that's simpler either :) I think it's "equally" valid to directly interact with the ClientCall API as opposed to using an interceptor (there are at least examples of how do so in the ClientCall javadoc)?

interop-testing/build.gradle
ericgribkoff: Fixed

interop-testing/src/main/java/io/grpc/testing/integration/XdsTestServer.java
ericgribkoff: Done

interop-testing/src/main/proto/grpc/testing/messages.proto
ericgribkoff: Renamed to `num_failures` (it doesn't exactly correspond with failed RPCs, but for the purpose of the load balancer stats request these are rpcs which failed to produce stats)

dapengzhang0: I don't care too much about choosing the name of the field. As there is always subtleties, it's worth a protodoc to clarify any subtleties that the user should be aware of.

ericgribkoff: Ok, done.

interop-testing/src/main/java/io/grpc/testing/integration/XdsTestClient.java
```diff
@@ -0,0 +1,339 @@
...
+      for (int i = 0; i < numChannels; i++) {
+        chans.add(NettyChannelBuilder.forTarget(server).usePlaintext().build());
+      }
+      exec = MoreExecutors.listeningDecorator(Executors.newSingleThreadScheduledExecutor());
+      try {
+        runQps(chans, exec);
+        success = true;
+      } finally {
+        for (ManagedChannel chan : chans) {
+          chan.shutdownNow();
```
dapengzhang0: nit: This should also be done in ShutdownHook, otherwise it may not be trigger if you press Ctrl-C.

ericgribkoff: done

interop-testing/src/main/java/io/grpc/testing/integration/XdsTestClient.java
```diff
@@ -0,0 +1,335 @@
...
+      for (int i = 0; i < numChannels; i++) {
+        channels.add(NettyChannelBuilder.forTarget(server).usePlaintext().build());
+      }
+      exec = MoreExecutors.listeningDecorator(Executors.newSingleThreadScheduledExecutor());
+      runQps();
+      success = true;
+    } catch (Throwable t) {
+      logger.log(Level.SEVERE, "Error running client", t);
+    }
+    System.exit(success ? 0 : 1);
```
dapengzhang0: Should we always `System.exit(1)`?

ericgribkoff: yep, done

PR #6589, opened 2020-01-08
xds: suppress hostname check but keep trust check in the delegated X509TrustManagerImpl

sanjaypujare:


xds/src/main/java/io/grpc/xds/sds/trust/SdsX509TrustManager.java
```diff
@@ -256,13 +258,26 @@ public void checkClientTrusted(X509Certificate[] chain, String authType)
   @Override
   public void checkServerTrusted(X509Certificate[] chain, String authType, Socket socket)
       throws CertificateException {
+    if (socket instanceof SSLSocket) {
+      SSLSocket sslSocket = (SSLSocket)socket;
```
creamsoup:
```suggestion
      SSLSocket sslSocket = (SSLSocket) socket;
```

xds/src/test/java/io/grpc/xds/sds/trust/SdsX509TrustManagerTest.java
```diff
@@ -221,14 +244,346 @@ public void oneIpAddressInPeerCertsMismatch() throws CertificateException, IOExc
...
+    };
+    sslEngine.setSSLParameters(sslParams);
+    return sslEngine;
+  }
+
+  private SSLSocket buildTrustManagerAndGetSslSocket()
+      throws CertificateException, IOException, CertStoreException {
+    SSLParameters sslParams = buildTrustManagerAndGetSslParameters();
+
+    SSLSocket sslSocket = new SSLSocket() {
```
creamsoup: same for this.

xds/src/test/java/io/grpc/xds/sds/trust/SdsX509TrustManagerTest.java
```diff
@@ -221,14 +244,346 @@ public void oneIpAddressInPeerCertsMismatch() throws CertificateException, IOExc
...
+    } catch (ValidatorException expected) {
+      assertThat(expected).hasMessageThat()
+          .endsWith("unable to find valid certification path to requested target");
+    }
+  }
+
+  private SSLEngine buildTrustManagerAndGetSslEngine()
+      throws CertificateException, IOException, CertStoreException {
+    SSLParameters sslParams = buildTrustManagerAndGetSslParameters();
+    SSLEngine sslEngine =  new SSLEngine() {
```
creamsoup: can you use mock to get rid of boiler plate? mock should be avoided, but in this case it seems like a better idea.

sanjaypujare: I would have used a mock if it worked. In the actual (new) code it is setting values in the object:
```
    if (sslParams != null) {
      sslParams.setEndpointIdentificationAlgorithm(null);
      sslEngine.setSSLParameters(sslParams);
    }
```
If `sslEngine` and `sslParams` were mocks, the setters will be no-op and the corresponding getters (called in Java/Sun library code) won't return the correct values. Isn't that right?

xds/src/test/java/io/grpc/xds/sds/trust/SdsX509TrustManagerTest.java
sanjaypujare: Mocks won't work here (AFAIK) as stated before.

xds/src/test/java/io/grpc/xds/sds/trust/SdsX509TrustManagerTest.java
creamsoup: i think you can create an abstract class with the set/get implemented and use mock with Mockito.CALLS_REAL_METHODS to achieve exact same behavior as current code.

sanjaypujare: good suggestion. done.

xds/src/test/java/io/grpc/xds/sds/trust/SdsX509TrustManagerTest.java
```diff
@@ -221,14 +244,132 @@ public void oneIpAddressInPeerCertsMismatch() throws CertificateException, IOExc
...
   }
+
+  @Test
+  public void checkServerTrustedSslEngine()
+      throws CertificateException, IOException, CertStoreException {
+    TestSslEngine sslEngine = buildTrustManagerAndGetSslEngine();
+    X509Certificate[] serverCerts =
+        CertificateUtils.toX509Certificates(TestUtils.loadCert(SERVER_1_PEM_FILE));
+    trustManager.checkServerTrusted(serverCerts, "ECDHE_ECDSA", sslEngine);
+    verify(sslEngine, times(1)).getHandshakeSession();
```
creamsoup: fyi, times(1) can be omitted unless very good reason to emphasize times(1). i am just pointing out because all verify has it.

PR #6600, opened 2020-01-14
core: fix bugs in service config error handling

creamsoup:
~1. Allow legacy way to pass service config via attributes~
2. fix AutoConfiguredLBFactory may populate ConfigOrError instead of
actual config.

### DIFF
--- a/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory2.java
+++ b/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory2.java
@@ -167,7 +167,7 @@ public final class AutoConfiguredLoadBalancerFactory2 {
             ResolvedAddresses.newBuilder()
                 .setAddresses(resolvedSelection.serverList)
                 .setAttributes(attributes)
-                .setLoadBalancingPolicyConfig(lbConfig)
+                .setLoadBalancingPolicyConfig(lbConfig != null ? lbConfig.getConfig() : null)
                 .build());
         return Status.OK;
       }
--- a/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory2.java
+++ b/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory2.java
@@ -149,7 +149,7 @@ public final class AutoConfiguredLoadBalancerFactory2 {
             ChannelLogLevel.INFO, "Load balancer changed from {0} to {1}",
             old.getClass().getSimpleName(), delegate.getClass().getSimpleName());
       }
-      ConfigOrError lbConfig = selection.config;
+      Object lbConfig = selection.config;
       if (lbConfig != null) {
         helper.getChannelLogger().log(
             ChannelLogLevel.DEBUG, "Load-balancing config: {0}", selection.config);
@@ -167,7 +167,7 @@ public final class AutoConfiguredLoadBalancerFactory2 {
             ResolvedAddresses.newBuilder()
                 .setAddresses(resolvedSelection.serverList)
                 .setAttributes(attributes)
-                .setLoadBalancingPolicyConfig(lbConfig != null ? lbConfig.getConfig() : null)
+                .setLoadBalancingPolicyConfig(lbConfig)
                 .build());
         return Status.OK;
       }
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory2.java
```diff
@@ -167,7 +167,7 @@ Status tryHandleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
             ResolvedAddresses.newBuilder()
                 .setAddresses(resolvedSelection.serverList)
                 .setAttributes(attributes)
-                .setLoadBalancingPolicyConfig(lbConfig)
+                .setLoadBalancingPolicyConfig(lbConfig != null ? lbConfig.getConfig() : null)
```
zhangkun83: Looks like there isn't a case where `selection.config` is an error. We may just change `PolicySelection.config` from `ConfigOrError` to `Object`.

core/src/main/java/io/grpc/internal/ManagedChannelImpl2.java
```diff
@@ -1332,9 +1333,13 @@ public void run() {
           ConfigOrError configOrError = resolutionResult.getServiceConfig();
           ServiceConfigHolder validServiceConfig = null;
           Status serviceConfigError = null;
+          Map<String, ?> rawServiceConfig =
+              resolutionResult.getAttributes().get(GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG);
+          if (configOrError == null && rawServiceConfig != null) {
+            // use service config json in deprecated attributes
+            configOrError = serviceConfigParser.parseServiceConfig(rawServiceConfig);
+          }
           if (configOrError != null) {
```
zhangkun83: I think the original code is correct.

For a LoadBalancer that's still on the old service config path, it won't override `LoadBalancerProvider.parseLoadBalancingPolicyConfig`, thus `configOrError` would be `LoadBalancerProvider.UNKNOWN_CONFIG`, which is non-null, thus `NAME_RESOLVER_SERVICE_CONFIG` will get populated. We shouldn't call `parseServiceConfig` again.

creamsoup: Some 1st/3rd party name resolver might not use the `serviceConfig` field which can be ignored in current implementation even if the deprecated `Attribute` is set. 
`LoadBalancerProvider.UNKNOWN_CONFIG` is still a valid config from the name resolver point of view, thus those aren't consumed.

zhangkun83: Now I see this is for "old" `NameResolver`s that don't call `parseServiceConfig()` at all but emit the raw config. Granted the raw config attribute is defined in the internal `GrpcAttributes` thus no third-party NameResolvers are expected to do that. It's fine to call parser here if that makes the migration for first-party NameResolver easier. It's going to be removed along with `GrpcAttributes.NAME_RESOLVER_SERVICE_CONFIG` anyway so not a big deal.

core/src/test/java/io/grpc/internal/ServiceConfigErrorHandlingTest.java
```diff
@@ -541,7 +571,17 @@ public NameResolver newNameResolver(final URI targetUri, NameResolver.Args args)
         return null;
       }
       assertEquals(DEFAULT_PORT, args.getDefaultPort());
-      FakeNameResolver resolver = new FakeNameResolver(args.getServiceConfigParser());
+      FakeNameResolver resolver;
+      if (enableServiceConfigParsing) {
```
zhangkun83: An "old" NameResolver that doesn't do service config parsing should call `listener.onAddresses()`, not `onResult()`. This makes sure the compatibility logic in `Listener2#onAddresses()` is exercised.

core/src/main/java/io/grpc/internal/ManagedChannelImpl2.java
creamsoup: it turns out that internal migration was easy. Thus, as pointed out the attributes shouldn't be used out side of gRPC. removing the service config parsing in `ManagedChannelImpl2`.

PR #6603, opened 2020-01-15
netty: fix server and client handlers to check the correct alpn list

sanjaypujare:


sanjaypujare:
@ejona86 have you heard from C/Go? Are we okay to go ahead with this PR?

ejona86:
I talked to @dfawley and @menghanl. The subject brought up bad memories for them because they have some unresolved things of their own. They need to investigate their own code/issues a bit. On C-side I talked to @yashykt, and things are mostly fine.

sanjaypujare:
> ...On C-side I talked to @yashykt, and things are mostly fine.

Thanks. Will wait for the Go investigation to finish.

menghanl:
To confirm, with this change, the client will allow the negotiated protocol to be anything from NextProtos, right? We would be OK with this change.

We have an opposite problem in Go: we never checked the negotiated protocols. So if we add the check, existing users might be broken.

sanjaypujare:
That's correct. So if both client and server have "managed-tls" (ahead of "h2") so they agree on "managed-tls" as the "next protocol" as the negotiated protocol. Looks like now both C and Go are okay with this change.

PR #6610, opened 2020-01-15
netty: fix a race for channelz at server transport creation

### LINKED ISSUE #6601: Deadlock on start gRPC server
### What version of gRPC-Java are you using?
1.26.0
### What is your environment?
Linux, Alpine
openjdk version "1.8.0_171"
OpenJDK Runtime Environment (IcedTea 3.8.0) (Alpine 8.171.11-r0)
OpenJDK 64-Bit Server VM (build 25.171-b11, mixed mode)
Netty - 4.1.44.Final
Vertx & Vertx-gRPC - 3.8.4
Proper start of gRPC Server
### What did you see instead?
Start sometimes hangs with deadlock
### Steps to reproduce the bug
[int_threaddump.txt](https://github.com/grpc/grpc-java/files/4058905/int_threaddump.txt)

I suppose it's a race condition related to synchronization in gRPC (ServerImpl), await in NetServer.start and vertx/nettty event loops (probably single threaded). Probably it could happened at any time if someone start gRPC server and concurrently open new client connection to that server.
In my case I stopped and started the gGPR server but I'm not sure if it is somehow related.

### Analysis
What I see in the thread dump is the following 2 threads that stays in that state, seems, forever:
>
"vert.x-eventloop-thread-0" #39 prio=10 os_prio=0 tid=0x000055711e379000 nid=0x2d waiting for monitor entry [0x00007fb72abc8000]
java.lang.Thread.State: BLOCKED (on object monitor)
  at io.grpc.internal.ServerImpl$ServerListenerImpl.transportCreated(ServerImpl.java:379)
  \- waiting to lock <0x00000000c559f1a0> (a java.lang.Object)
  at io.grpc.netty.NettyServer$1.initChannel(NettyServer.java:224)
  \- locked <0x00000000c559bfd8> (a io.grpc.netty.NettyServer)
  at io.netty.channel.ChannelInitializer.initChannel(ChannelInitializer.java:129)

"vert.x-worker-thread-12" #214 prio=10 os_prio=0 tid=0x000055711f2d1800 nid=0x418 in Object.wait() [0x00007fb720656000]
java.lang.Thread.State: WAITING (on object monitor)
  at java.lang.Object.wait(Native Method)
  at java.lang.Object.wait(Object.java:502)
  at io.netty.util.concurrent.DefaultPromise.await(DefaultPromise.java:252)
  \- locked <0x00000000c589baa0> (a io.netty.util.concurrent.PromiseTask)
  at io.netty.util.concurrent.DefaultPromise.await(DefaultPromise.java:35)
  at io.grpc.netty.NettyServer.start(NettyServer.java:269)
  at io.grpc.internal.ServerImpl.start(ServerImpl.java:184)
  \- locked <0x00000000c559f1a0> (a java.lang.Object)
  at io.grpc.internal.ServerImpl.start(ServerImpl.java:90)
>
From what I see in these thread dumps and the code I think that this could be the problem (deadlock):
1. Vertx grpc starts server (ServerImpl.start) in vertx blocking thread
2. ServerImpl synchronize on lock and then try (keeping lock) to start server (NetServer.start)
3. NetServer.start opens a channel, binds to it, and since that moment it, I assume, may receive connections from remote clients
4. It seems, at this time a remote client opens connection to this server (already bound)
5. Then in channel's event loop (probably single threaded) is received initChannel which try to get ServerImpl.lock in ServerListenerImpl.transportCreated (coudln't because got by ServerImpl.start)
6. NetServer.start then schedules runnable in channel's event loop and blocks with channelzFuture.await() 
7. Now, channelzFuture.await() waits for a runnable to be executed in channel's event loop (probably single threaded)
8. At this point channelzFuture.await keeps ServerImpl.lock lock, while the ServerListenerImpl.transportCreated occupies/blocks (this is what I suppose) the single threaded channel's event loop thus making impossible to process further

I'm attaching file with thread dumps of the whole JVM

dapengzhang0:
A race condition was reported by user in #6601:

`ServerImpl.start()` calls `NettyServer.start()` while holding `ServerImpl.lock`. `NettyServer.start()` awaits a submitted runnable in eventloop. However, this pending runnable may never be executed because the eventloop might be executing some other task, like `ServerListenerImpl.transportCreated()`, that is trying to acquire `ServerImpl.lock` causing a deadlock.

This PR resolves the particular issue reported in #6601 for server with a single port, but `NettyServer` (https://github.com/grpc/grpc-java/blob/v1.26.0/netty/src/main/java/io/grpc/netty/NettyServer.java#L251) and `ServerImpl` (https://github.com/grpc/grpc-java/blob/v1.26.0/core/src/main/java/io/grpc/internal/ServerImpl.java#L184) in general still have the same potential risk of deadlock, which need further fix.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServer.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServer.java
@@ -251,19 +251,10 @@ class NettyServer implements InternalServer, InternalWithLogId {
       throw new IOException("Failed to bind", future.cause());
     }
     channel = future.channel();
-    Future<?> channelzFuture = channel.eventLoop().submit(new Runnable() {
-      @Override
-      public void run() {
-        InternalInstrumented<SocketStats> listenSocket = new ListenSocket(channel);
-        listenSocketStats.set(listenSocket);
-        channelz.addListenSocket(listenSocket);
-      }
-    });
-    try {
-      channelzFuture.await();
-    } catch (InterruptedException ex) {
-      throw new RuntimeException("Interrupted while registering listen socket to channelz", ex);
-    }
+
+    InternalInstrumented<SocketStats> listenSocket = new ListenSocket(channel);
+    listenSocketStats.set(listenSocket);
+    channelz.addListenSocket(listenSocket);
   }
 
   @Override
--- a/netty/src/main/java/io/grpc/netty/NettyServer.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServer.java
@@ -56,7 +56,6 @@ import java.net.SocketAddress;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -93,9 +92,8 @@ class NettyServer implements InternalServer, InternalWithLogId {
   private final List<? extends ServerStreamTracer.Factory> streamTracerFactories;
   private final TransportTracer.Factory transportTracerFactory;
   private final InternalChannelz channelz;
-  // Only modified in event loop but safe to read any time. Set at startup and unset at shutdown.
-  private final AtomicReference<InternalInstrumented<SocketStats>> listenSocketStats =
-      new AtomicReference<>();
+  // Only modified in event loop but safe to read any time.
+  private volatile InternalInstrumented<SocketStats> listenSocketStats;
 
   NettyServer(
       SocketAddress address, ChannelFactory<? extends ServerChannel> channelFactory,
@@ -149,7 +147,7 @@ class NettyServer implements InternalServer, InternalWithLogId {
 
   @Override
   public InternalInstrumented<SocketStats> getListenSocketStats() {
-    return listenSocketStats.get();
+    return listenSocketStats;
   }
 
   @Override
@@ -251,10 +249,13 @@ class NettyServer implements InternalServer, InternalWithLogId {
       throw new IOException("Failed to bind", future.cause());
     }
     channel = future.channel();
-
-    InternalInstrumented<SocketStats> listenSocket = new ListenSocket(channel);
-    listenSocketStats.set(listenSocket);
-    channelz.addListenSocket(listenSocket);
+    channel.eventLoop().execute(new Runnable() {
+      @Override
+      public void run() {
+        listenSocketStats = new ListenSocket(channel);
+        channelz.addListenSocket(listenSocketStats);
+      }
+    });
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -251,19 +251,10 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
-        channelz.addListenSocket(listenSocket);
-      }
-    });
-    try {
-      channelzFuture.await();
-    } catch (InterruptedException ex) {
-      throw new RuntimeException("Interrupted while registering listen socket to channelz", ex);
-    }
+
+    InternalInstrumented<SocketStats> listenSocket = new ListenSocket(channel);
```
ejona86: Make a comment here that this _requires_ the channel to have completed bind()ing.

ejona86:
Okay. I'm seeing now there is a loop with ServerImpl.lock from start() and transportCreated(). I think the lock in start() is to avoid callbacks from being seen too early. Hmm... and it is also used for things like getPort and shutdown(), although honestly we don't really support calling shutdown/getPort before start returns, so we could potentially call ts.start() after releasing the lock within ServerImpl.start()

netty/src/test/java/io/grpc/netty/NettyServerTest.java
```diff
@@ -239,8 +241,13 @@ public void serverShutdown() {
         shutdownCompleted.set(null);
       }
     });
+
     assertThat(((InetSocketAddress) ns.getListenSocketAddress()).getPort()).isGreaterThan(0);
 
+    Socket socket = new Socket();
+    socket.connect(ns.getListenSocketAddress(), /* timeout= */ 8000);
+    socket.close();
+    countDownLatch.await(); // SocketStats won't be available until binding complete.
```
ejona86: Binding is complete when start() returns, so I don't think this comment is accurate.

netty/src/test/java/io/grpc/netty/NettyServerTest.java
```diff
@@ -239,8 +241,13 @@ public void serverShutdown() {
         shutdownCompleted.set(null);
       }
     });
+
     assertThat(((InetSocketAddress) ns.getListenSocketAddress()).getPort()).isGreaterThan(0);
 
+    Socket socket = new Socket();
```
ejona86: This works, but is a roundabout way of making sure the event loop has completed the runnable. Would it be better to Create a single EventLoop and pass that as the boss+worker EventLoopGroup to the NettyServer constructor? Then you can submit() your own Runnable and await its completion. We could use that same EventLoop for all the tests, really, if you'd prefer a `@BeforeClass`/`@AfterClass` to manage its lifetime.

dapengzhang0: This roundabout is just copied from the test above it. Actually `ns.getListenSocketStats()` is not required to be called in the event loop. I'm not sure submit() is the most correct way to do it.

ejona86: The one above is actually checking the worker channel options. So creating a connection appropriate and necessary there.

I agree `ns.getListenSocketStats()` isn't required to be done in the event loop, but I wasn't suggesting to call it from the event loop. I was suggesting:
```
eventLoop.submit(new Runnable() {
    @Override public void run() {}
}).await();
InternalInstrumented<SocketStats> listenSocket = ns.getListenSocketStats();
```

It is just a synchronization barrier.

dapengzhang0: Done with fixed event loop pool.

PR #6614, opened 2020-01-16
core: fix a race in ServerImpl for multi-transport-server

dapengzhang0:
As discovered in #6601 and its partial fix #6610,  `ServerImpl` (https://github.com/grpc/grpc-java/blob/v1.26.0/core/src/main/java/io/grpc/internal/ServerImpl.java#L184) and  `NettyServer` (https://github.com/grpc/grpc-java/blob/v1.26.0/netty/src/main/java/io/grpc/netty/NettyServer.java#L251) have risk of deadlock.

Targeting the issue #6641.

There might be two things need be done to make the code healthy:
1. move `ts.start(listener)` out of synchronization block
2. move multiple port support from `ServerImpl` to `NettyServer`.

This PR is working on (1).

### DIFF
--- a/core/src/main/java/io/grpc/internal/ServerImpl.java
+++ b/core/src/main/java/io/grpc/internal/ServerImpl.java
@@ -179,21 +179,50 @@ public final class ServerImpl extends io.grpc.Server implements InternalInstrume
     synchronized (lock) {
       checkState(!started, "Already started");
       checkState(!shutdown, "Shutting down");
+      activeTransportServers = transportServers.size();
       executor = Preconditions.checkNotNull(executorPool.getObject(), "executor");
       started = true;
     }
 
     // Start and wait for any ports to actually be bound.
     ServerListenerImpl listener = new ServerListenerImpl();
+    int unstarted = transportServers.size();
     try {
       for (InternalServer ts : transportServers) {
         ts.start(listener);
-        synchronized (lock) {
-          activeTransportServers++;
-        }
+        unstarted--;
       }
     } finally {
       startComplete.countDown();
+
+      // Adjust number of active transport servers if there is any unstarted servers.
+      List<ServerTransport> transportsToCleanUp = null;
+      Status shutdownNowStatusCopy = null;
+      synchronized (lock) {
+        if (unstarted != 0) {
+          activeTransportServers -= unstarted;
+          if (activeTransportServers == 0) {
+            // transports collection can be modified during shutdown(), even if we hold the lock,
+            // due to reentrancy.
+            transportsToCleanUp = new ArrayList<>(transports);
+            shutdownNowStatusCopy = shutdownNowStatus;
+            serverShutdownCallbackInvoked = true;
+          }
+        }
+      }
+      if (transportsToCleanUp != null) {
+        for (ServerTransport transport : transportsToCleanUp) {
+          if (shutdownNowStatusCopy == null) {
+            transport.shutdown();
+          } else {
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/grpc/internal/ServerImplTest.java
```diff
@@ -420,7 +420,6 @@ public void start(ServerListener listener) throws IOException {
     } catch (IOException e) {
       assertSame(ex, e);
     }
-    verifyNoMoreInteractions(executorPool);
```
ejona86: Why was this removed?

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -388,6 +409,7 @@ public ServerTransportListener transportCreated(ServerTransport transport) {
     public void serverShutdown() {
       ArrayList<ServerTransport> copiedTransports;
       Status shutdownNowStatusCopy;
+      awaitStartComplete();
```
ejona86: This looks dangerous.

I'd much rather we set `activeTransportServers = transportServers.size()` within the initial synchronization block of start() to prevent needing to block here.

dapengzhang0: set `activeTransportServers = transportServers.size()` is broken if one of `ts.start()` throws because not enough times of `listern.serverShutdown()` will be called .

core/src/main/java/io/grpc/internal/ServerImpl.java
```diff
@@ -269,13 +277,26 @@ public ServerImpl shutdown() {
...
         ts.shutdown();
       }
     }
     return this;
   }
 
+  private void awaitStartComplete() {
+    boolean startCompleted = false;
+    try {
+      startCompleted = startComplete.await(30, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
dapengzhang0: We can assert that is true.

core/src/test/java/io/grpc/internal/ServerImplTest.java
dapengzhang0: Because `executorPool.getObject()` is called after the change.

core/src/main/java/io/grpc/internal/ServerImpl.java
dapengzhang0: Seems we need redesign multiple transportServers support in future. It's just too inconvenient now. For now the fix seems working. Can you elaborate what is the danger?

dapengzhang0:
> Hmmm... It is really awkward that we can have bind succeed for one transport and have it fail for another and then start() with throw and leave things in a half-started state. Previously, when there was only one transport, ts.start() throwing would leave started == false.

Yes previously it was broken in multiple ways.

core/src/main/java/io/grpc/internal/ServerImpl.java
dapengzhang0: Now I see the danger, because `serverShutdown()` is running in the event loop, if another `ts.start()` submits something into the event loop while `awaitStartComplete()` in  `serverShutdown()`, then it will be deadlock again.

ejona86:
We gave up on trying to fix Server for multiple ports. The API is just wrong to support partial failures. The multiple-port handling will be moved into the Netty transport, where partial failures can be handled.

dapengzhang0:
Close in favor of #7674 and subsequent PRs

PR #6615, opened 2020-01-16
core,netty: let ServerListener.serverShutdown() only be called when InternalServer.start() returns

dapengzhang0:
`ServerImpl.activeTransportServers` is broken, because `listener.serverShutdown()` may or may not be called when `start()` throws. Now change the API spec of `ServerListener.serverShutdown()` so that it won't be called if `start()` throws. 

This spec change is reasonable because `server.shutdown()` is never called if `Server.start()` throws in a lot of users' code, e.g.:

```
Server server = null;
try {
  server = ServerBuilder.forPort(8080).build().start();
} catch (IOException e) {
  // server is null, so shutdown() is never called for the half-baked server.
}

if (server != null) {
  server.shutdown();
} 
```

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServer.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServer.java
@@ -250,8 +250,8 @@ class NettyServer implements InternalServer, InternalWithLogId {
     }
     channel = future.channel();
 
-    // We should never throw any Exception after this point to abide by the contract of
-    // ServerListener.serverShutdown()
+    // We should never throw any Exception below while the channel is open to abide by the contract
+    // of ServerListener.serverShutdown() method.
 
     try {
       channel.eventLoop().execute(new Runnable() {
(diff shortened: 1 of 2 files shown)


netty/src/main/java/io/grpc/netty/NettyServer.java
```diff
@@ -249,13 +249,21 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
+    try {
+      channel.eventLoop().execute(new Runnable() {
+        @Override
+        public void run() {
+          listenSocketStats = new ListenSocket(channel);
+          channelz.addListenSocket(listenSocketStats);
+        }
+      });
+    } catch (RuntimeException e) {
+      log.log(Level.WARNING, "Error while starting server", e);
```
ejona86: This does not seem good. I don't think we should go in a direction that logs exceptions instead of propagating them.

dapengzhang0: When the eventloop.execute() is broken, there is not much we can handle. What about 
```java
try {
      channel.eventLoop().execute(new Runnable() {
        @Override
        public void run() {
          listenSocketStats = new ListenSocket(channel);
          channelz.addListenSocket(listenSocketStats);
        }
      });
      try {
        channel.close();
      } finally {
        channel = null;
        throw e;
      }
    }
```

ejona86:
@dapengzhang0, we've resolved this in other ways, right? Can this be closed, or is it a helpful reminder of anything?

PR #6616, opened 2020-01-17
Add a util function to check if a transport is secure enough to transfer CallCredentials 

yihuazhang:
This PR adds a util function that checks if the security level of transport is higher than or equal to that of CallCredentials. The util function is called `GoogleAuthLibraryCallCredentials` and `FileBasedPluginCredential` for which I believe privacy+integrity should be required for the transport. After this PR gets merged, I will update the internal CallCredentials implementation to call the util function (for the auditing purpose). 

gRFC for security level negotiation: https://github.com/grpc/proposal/pull/167


xds/src/main/java/io/grpc/xds/sds/FileBasedPluginCredential.java
```diff
@@ -90,6 +91,12 @@ private static DataSource buildDataSourceFromConfigStruct(Struct secretValueStru
   @Override
   public void applyRequestMetadata(
       @Nullable RequestInfo requestInfo, Executor appExecutor, final MetadataApplier applier) {
+    if (!checkSecurityLevel(requestInfo, SecurityLevel.PRIVACY_AND_INTEGRITY)) {
```
sanjaypujare: So this is currently used for SDS which is over an UDS channel. Will this check now break the SDS communication or is UDS considered at `SecurityLevel.PRIVACY_AND_INTEGRITY` ?

sanjaypujare:
One test failed:
```
io.grpc.xds.sds.SdsClientUdsFileBasedMetadataTest > testSecretWatcher_tlsCertificate FAILED
    expected to be true
        at io.grpc.xds.sds.SdsClientUdsFileBasedMetadataTest.testSecretWatcher_tlsCertificate(SdsClientUdsFileBasedMetadataTest.java:105)
```

This indicates the code change will also break prod behavior. Did the test pass in your testing?

xds/src/main/java/io/grpc/xds/sds/FileBasedPluginCredential.java
yihuazhang: UDS should be considered secure (i.e., `SecurityLevel.PRIVACY_AND_INTEGRITY`). Do you know if the security level is correctly set for the UDS channel?

sanjaypujare: Don't know. If not, we will have to do that before merging this change to ensure this feature doesn't break. Let me know if you want to discuss

api/src/main/java/io/grpc/CallCredentials.java
```diff
@@ -62,6 +62,22 @@ public abstract void applyRequestMetadata(
...
+   * level required by {@link CallCredentials} in order to transfer it.
+   *
+   * <p>It should be called in {@link applyRequestMetadata} before sending any
+   * individual RPC. The RPC should not be sent if the API returns false.
+   * More details can be found in https://github.com/grpc/proposal/pull/167.
+   *
+   * @param requestInfo request-related information
+   * @param minSecurity minimum security level required by {@link CallCredentials}
+   */
+  public final boolean checkSecurityLevel(RequestInfo requestInfo, SecurityLevel minSecurity) {
```
ejona86: `check` makes it seem like this will throw. Maybe name it `allowedSecurityLevel` or `fulfullsSecurityLevel` or similar that makes it more obvious it returns a boolean?

CC @jiangtaoli2016

ejona86: `static`?

yihuazhang:
@ejona86 @sanjaypujare 

Is it POR for using a plaintext channel for UDS? If that's the case, I have a following plan to associate the security level `PRIVACY_AND_INTEGRITY` to a UDS channel. PLMK what you think. 

- Add a new `NegotiationType` - UDS. 
-  In `usePlaintext()`, check if the server's `SocketAddress` is an instance of `DomainSocketAddress` and if it is, set the negotiation type to UDS. We do not allow users to specify if the channel is UDS or not and by doing that, we can avoid their intentional (or unintentional) mistakes.
-  Create a new API - uds() in `ProtocolNegotiator` and add a new constructor for `PlaintextProtocolNegotiator` - `PlaintextProtocolNegotiator(SecurityLevel)`. For UDS, the security level will be PRIVACY_AND_INTEGRITY. 
- Update security level attribute accordingly.

sanjaypujare:
> @ejona86 @sanjaypujare
> 
> Is it POR for using a plaintext channel for UDS? 

This is dictated by the other end - the SDS server (Citadel agent) only supports UDS based plaintext channel. See https://istio.io/docs/tasks/security/citadel-config/auth-sds/#securing-sds-with-pod-security-policies (instead of Envoy sidecar, read it as proxyless gRPC client).

> If that's the case, I have a following plan to associate the security level `PRIVACY_AND_INTEGRITY` to a UDS channel. PLMK what you think.

Comments below:

> 
> * Add a new `NegotiationType` - UDS.

`NegotiationType` is meant to indicate 'negotiation used for starting up HTTP/2' so it is not appropriate to include transport characteristics here.

> * In `usePlaintext()`, check if the server's `SocketAddress` is an instance of `DomainSocketAddress` and if it is, set the negotiation type to UDS. We do not allow users to specify if the channel is UDS or not and by doing that, we can avoid their intentional (or unintentional) mistakes.

What if user called `forTarget` so there is no `directServerAddress` (`SocketAddress`) set? You might be able to check `unix:` prefix in the target string but the bigger problem is using `negotiationType` to convey transport.

> * Create a new API - uds() in `ProtocolNegotiator` and add a new constructor for `PlaintextProtocolNegotiator` - `PlaintextProtocolNegotiator(SecurityLevel)`. For UDS, the security level will be PRIVACY_AND_INTEGRITY.
> * Update security level attribute accordingly.

Looking at this attribute code it might be better to use the `ChannelHandlerContext` that is passed in `fireProtocolNegotiationEvent` (for example) and you call `ChannelHandlerContext.channel()` and on that channel get `remoteAddress()` to see what kind of server address it is. This is better than overloading `NegotiationType` .

ejona86:
> Add a new NegotiationType - UDS.

No, I don't think we want to do that. It would be strictly better to check for DomainSocketAddress at [NettyChannelBuilder.createProtocolNegotiatorByType()](https://github.com/grpc/grpc-java/blob/44db31d147141ad9e066eb8ed28077dc2ea71999/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java#L467). We can either pass in the SocketAddress to the function or make it non-static.

That said, I'd really prefer a solution that works with a future UnixNameResolver. We can move the DomainSocketAddress check to PlaintextProtocolNegotiator. Unfortunately, it wouldn't be guaranteed to be authenticated in that case, since maybe a MyCustomNameResolver returns unauthenticated DomainSocketAddresses. We could require the filename also matches the authority (and use INSECURE if it is not).

xds/src/main/java/io/grpc/xds/sds/FileBasedPluginCredential.java
```diff
@@ -90,6 +91,12 @@ private static DataSource buildDataSourceFromConfigStruct(Struct secretValueStru
   @Override
   public void applyRequestMetadata(
       @Nullable RequestInfo requestInfo, Executor appExecutor, final MetadataApplier applier) {
+    if (!allowedSecurityLevel(requestInfo, SecurityLevel.PRIVACY_AND_INTEGRITY)) {
```
ejona86, line 94: Actually, this may not be correct. @sanjaypujare, can you comment whether we may be required to use bearer tokens on an unauthenticated connection?

sanjaypujare: If you are asking about "unauthenticated", yes - with at least 2 examples :

- in case of SDS over UDS the connection does not authenticate the client but we do send bearer tokens (as in this code) over that channel

- in TLS  (non-mTLS) the client is not authenticated to the server but the client can send bearer tokens.

IIRC both UDS and TLS are considered at `SecurityLevel.PRIVACY_AND_INTEGRITY` in this PR to support the above cases. We probably should have more levels to accurately reflect some of the characteristics.

ejona86: For plain TLS it is fine that the client is unauthenticated; the client doesn't need to authenticate itself, just the server. If FileBasedPluginCredential is used for SDS, then this would break it currently. Right now UDS is not considered PRIVACY_AND_INTEGRITY; there's no special code for it so it is the same as plain-text.

sanjaypujare: Above in https://github.com/grpc/grpc-java/pull/6616#issuecomment-575827581 @yihuazhang did say this "...I have a following plan to associate the security level PRIVACY_AND_INTEGRITY to a UDS channel...". Haven't heard anything since then... @yihuazhang can you throw some light on this?

ejona86: But that isn't implemented. It seems we should not add the check here, to let this PR be merged.

sanjaypujare: Yes, I agree. Unless he decides to implement that other thing in this PR

jiangtaoli2016: we agreed in go/grpc-security-level-negotiation that UDS is treated as privacy-and-integrity.

jiangtaoli2016:
@yihuazhang Could you please resolve the conflict

xds/src/main/java/io/grpc/xds/sds/FileBasedPluginCredential.java
sanjaypujare: So that should be implemented here, right?

ejona86: @jiangtaoli2016, I'm not arguing that UDS _shouldn't_ be privacy-and-integrity. I'm arguing that it _isn't_ today. So adding this code would completely break the SDS implementation. Adding this code is also unrelated to the goal of this PR: to add a utility function for credential implementations. If that is the goal, then we should drop the change in this file and merge the utility.

yihuazhang: Thanks all for the clarification. I will drop the changes in this function and will re-visit it in a separate PR (maybe after UDS is supported and a correct security level is set for it).

sanjaypujare: The checks are already in in some places like https://github.com/grpc/grpc-java/blob/master/auth/src/main/java/io/grpc/auth/GoogleAuthLibraryCallCredentials.java#L93 . Need to make sure those won't break anything.

ejona86: If the check is already there, then using the utility is safe. They produce the same results.

yihuazhang:
@ejona86 @sanjaypujare PTALA

api/src/main/java/io/grpc/CallCredentials.java
```diff
@@ -62,6 +62,23 @@ public abstract void applyRequestMetadata(
...
+   * security level required by {@link CallCredentials} in order to transfer it.
+   *
+   * <p>It should be called in {@link applyRequestMetadata} before sending any
+   * individual RPC. The RPC should not be sent if the API returns false.
+   * More details can be found in https://github.com/grpc/proposal/pull/167.
+   *
+   * @param requestInfo request-related information
+   * @param minSecurity minimum security level required by {@link CallCredentials}
+   */
+  public static final boolean allowedSecurityLevel(
```
ejona86, line 76: Thinking about it more, let's make this protected. Only extending classes will need to call this. It wouldn't really hurt anything by having it public, but pollutes what consumers of CallCredentials would see.

larry-safran:
@yihuazhang Would you please sign the EasyCLA so that this can be merged?

PR #6622, opened 2020-01-18
xds: enable keepalive for XDS channel

dapengzhang0:
XDS clients will use a keepalive time of 5 minutes ~~, and a keepalive timeout of 20 seconds~~.

Unfortunately it's not possible to be tested in unit test.

### DIFF
--- a/xds/src/main/java/io/grpc/xds/XdsClient.java
+++ b/xds/src/main/java/io/grpc/xds/XdsClient.java
@@ -40,6 +40,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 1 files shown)


xds/src/main/java/io/grpc/xds/XdsClient.java
```diff
@@ -491,19 +492,23 @@ ManagedChannel createChannel(List<ServerInfo> servers) {
...
-        if (ch == null) {
-          ch = ManagedChannelBuilder.forTarget(serverUri).build();
+        if (channelBuilder == null) {
+          channelBuilder = ManagedChannelBuilder.forTarget(serverUri);
         }
-        return ch;
+
+        return channelBuilder
+            .keepAliveTime(5, TimeUnit.MINUTES)
+            .keepAliveTimeout(20, TimeUnit.SECONDS)
```
sanjaypujare: DO we really want to hardcode these values? WHat will be ideal is if we can supply these values from the Bootstrap file as `XdsNameResolver.start` -> `XdsClientImpl` ctor -> `XdsClient.createChannel()` so these are configurable at runtime. The defaults could be these hardcoded values.

ejona86: While we could, most value is provided by having _some_ value here. Anything via bootstrap would be an optimization. It seems there is a wide range of "acceptable" values, so this will probably be "good enough" for everyone. We can maybe add a configuration option in the future, but there are _much_ more important things to do and we have limited resources.

ejona86: I was thinking we'd just leave Timeout as its default. Yes, the default is 20 seconds, but no need to specify that here.

sanjaypujare: > Anything via bootstrap would be an optimization.

I think it is more than just optimization. Because of network conditions if someone wants to change these at runtime instead of having to recompile gRPC library they will be able to do that. But I agree this might be sufficient for now and there are more important things to do.

dapengzhang0: Our internal prod channel builder has a different timeout. What's the specification? Use channel's default keepalive timeout or use 20 seconds?

ejona86: It is "just an optimization" because xds LB should be able to function for 5 minutes without XDS server. That is to say it is not a correctness issue.

ejona86: Each language handles internal vs external differently. I think Java and C have some options here, since the keepalive setting is in globalish state. I think Go may not have many options, since it probably wouldn't get the internal configuration naturally.

I don't think 10 seconds vs 20 seconds makes much of a difference internally, so in some ways I don't really care. More importantly we will be overriding the default internal keepalive time of 20 seconds and making it 5 minutes, which has a much bigger effect.

Options that come to mind:
 1. Don't care. We said 5 minutes will be fine, use 5 minutes. Could be surprising since it has different configuration from everything else
 2. Add configuration. Internally we can set different values in the bootstrap file. Could be surprising since it has different configuration from everything else
 3. Figure out a hack to use the default keepalive internally but set it to 5 minutes externally.
 4. Change the default keepalive for channels to something reasonable, and then not specify anything here. Too risky short-term. We know we already want to change the default keepalive for channels

Right now I think I'm favoring (1), with potential to do (2) and/or (4) in the future. We would leave timeout to be its system-specified default (so not specify it here).

CC @markdroth

PR #6628, opened 2020-01-22
xds: better error handling to avoid RPC hangup

dapengzhang0:
This change will fail application RPC immediately if XdsClient encounters any error instead of retrying or getting to fallback silently. 

There could be optimization if the channel is currently READY while XdsClient stream just closed due to connection error, in which case we could still be using the current available subchannels while retrying, but this requires the LB knows the semantics of error status from the XdsClient. This optimization is not worth the effort for now.


dapengzhang0:
@srini100 Can you test with patching this PR to see if it has better error handling and provides better user experience.

PR #6630, opened 2020-01-22
examples: Allow passing target and simplify lifecycle

ejona86:
The target can be provided on the command line to avoid needing to recompile
the example just to change where the server is located. We use a target instead
of addresses as that is the approach we have wanted to encourage for a while
since it allows choosing alternative name resolvers.

We typically encourage injecting Channels, not ManagedChannels, which has the
added benefit of simplifying the example. Less indirection makes for a better
example.

Swapping to target string could be done to examples-tls and examples-gauth as
well, but it would be much more invasive to the tls example and the gauth
example would need proper testing after the change.

------

These are things that I've wanted for a while, but now the target change is
necessary for the xds example and the rest sort of falls out of that.

### DIFF
--- a/examples/src/test/java/io/grpc/examples/routeguide/RouteGuideClientTest.java
+++ b/examples/src/test/java/io/grpc/examples/routeguide/RouteGuideClientTest.java
@@ -40,7 +40,6 @@ import java.util.Random;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
-import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
@@ -96,14 +95,9 @@ public class RouteGuideClientTest {
     // Use a mutable service registry for later registering the service impl for each test case.
     grpcCleanup.register(InProcessServerBuilder.forName(serverName)
         .fallbackHandlerRegistry(serviceRegistry).directExecutor().build().start());
-    client =
-        new RouteGuideClient(InProcessChannelBuilder.forName(serverName).directExecutor());
+    client = new RouteGuideClient(grpcCleanup.register(
+        InProcessChannelBuilder.forName(serverName).directExecutor().build()));
     client.setTestHelper(testHelper);
-  }
-
-  @After
-  public void tearDown() throws Exception {
-    client.shutdown();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


ejona86:
CC @donnadionne, @menghanl, this is a general cleanup of the helloworld example, but it does add more argument parsing. See the README added in #6631 to see how this will look to users in Java. C++'s plan of just adding a flag to set the target string sounds fine. Go may need to copy the main.go from the helloworld example, since it can't as easily do the trick I did to reuse the main() function.

sanjaypujare:
For the example to work with `"xds:///..."` (or  `"xds-experimental:///..."` currently) target string the dependencies will need to include grpc-xds?

ejona86:
@sanjaypujare, yes. See #6631

dapengzhang0:
Seems the README.md must be changed.

ejona86:
@dapengzhang0, why must the readme be changed?

dapengzhang0:
Never mind, I though the hostname and port are present in the command line arguments in README, they are not.

ejona86:
Note: After the next release this will require a smallish change to [the tutorial](https://grpc.io/docs/tutorials/basic/java/) since it has snippets of the routeguide that changed here. The quickstart is not impacted.

examples/src/test/java/io/grpc/examples/routeguide/RouteGuideClientTest.java
```diff
@@ -96,16 +95,11 @@ public void setUp() throws Exception {
     // Use a mutable service registry for later registering the service impl for each test case.
     grpcCleanup.register(InProcessServerBuilder.forName(serverName)
         .fallbackHandlerRegistry(serviceRegistry).directExecutor().build().start());
-    client =
-        new RouteGuideClient(InProcessChannelBuilder.forName(serverName).directExecutor());
+    client = new RouteGuideClient(grpcCleanup.register(
```
dapengzhang0: nit: I would declare a local variable for channel as this expression is somewhat complex.

ejona86: I considered that as well, but when I tried it it didn't really seem to help much. I think the best approach might be to make the channel/server fields which would remove most of the code from this `@Before`. But even that I'm not 100% sold on. I'll leave this as-is for now.

PR #6631, opened 2020-01-22
examples: Add XDS client example

ejona86:
~~This depends on #6630, and includes a commit from it.~~

### DIFF
--- a/examples/example-xds/README.md
+++ b/examples/example-xds/README.md
@@ -10,19 +10,43 @@ will be very hard to produce a working enviornment just by this example. Please
 refer to documentation specific for your XDS management server and
 environment.__
 
-The example requires grpc-java to already be built. You are strongly encouraged
-to check out a git release tag, since there will already be a build of grpc
-available. Otherwise you must follow [COMPILING](../../COMPILING.md).
+The example requires grpc-xds, but grpc-xds is not currently being published.
+You will thus need to build it yourself. This should guide you, but if you
+encounter issues please consult [COMPILING.md](../../COMPILING.md).
 
 ### Build the example
 
-1. Build the hello-world example server or the hostname example server. See
+1. The server does not use XDS, so recent releases work fine. Building using
+recent releases is much easier, so check out the most recent release tag:
+```
+$ git checkout v1.27.0
+```
+
+2. Build the hello-world example server or the hostname example server. See
    [the examples README](../README.md) or the
    [hostname example README](../example-hostname/README.md).
 
-2. Build this client. From the `grpc-java/examples/examples-xds` directory:
+3. Since XDS is still developing rapidly, XDS-using code should be built from
+master:
 ```
-$ ../gradlew installDist
+$ git checkout master
+```
+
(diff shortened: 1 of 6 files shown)


examples/example-xds/README.md
```diff
@@ -0,0 +1,53 @@
...
+XDS management protocol. Out-of-the-box it behaves the same as hello world
+client.
+
+__XDS support is incomplete and experimental, with limited compatibility. It
+will be very hard to produce a working enviornment just by this example. Please
+refer to documentation specific for your XDS management server and
+environment.__
+
+The example requires grpc-java to already be built. You are strongly encouraged
+to check out a git release tag, since there will already be a build of grpc
```
dapengzhang0: Release tag may not work until grpc-xds is published to maven.

examples/example-xds/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+And in a different terminal window run this client:
+
+```
+$ ./build/install/example-xds/bin/xds-hello-world-client
+```
+
+However, that didn't use XDS! To use XDS we assume you have deployed the server
+in your deployment environment and know its name. You need to set the
+`GRPC_XDS_BOOTSTRAP` environment variable to point to a gRPC XDS bootstrap
```
dapengzhang0: There is no example of bootstrap. How about referencing https://github.com/grpc/proposal/pull/170 and in particular https://github.com/grpc/proposal/pull/170/files#diff-05ea4a5894abbc0261b006741220598cR100 when it's merged?

ejona86:
@dapengzhang0, PTAL

examples/example-xds/README.md
```diff
@@ -10,19 +10,43 @@ will be very hard to produce a working enviornment just by this example. Please
...
+You will thus need to build it yourself. This should guide you, but if you
+encounter issues please consult [COMPILING.md](../../COMPILING.md).
 
 ### Build the example
 
-1. Build the hello-world example server or the hostname example server. See
+1. The server does not use XDS, so recent releases work fine. Building using
+recent releases is much easier, so check out the most recent release tag:
+```
+$ git checkout v1.27.0
```
dapengzhang0: This number and below will be changed after every release. Add an line of instruction at https://github.com/grpc/grpc-java/pull/6631/files#diff-60cd2f42437b5cb3c2c09391a6f4c54aR125 ?

dapengzhang0: If you look at the history of https://github.com/grpc/grpc-java/commits/master/documentation/android-channel-builder.md, we miss a commit "Update README etc to reference 1.27.0", but RELEASING.md doesn't document to add this commit in the master branch, while similar commits were added for the past releases. I'm confused.

ejona86: As far as I'm concerned, I'm okay if we don't update this. Two reasons: 1) it'll work fine, no big deal. 2) grpc-xds should be published the next release or the release after, so it will only be "wrong" for a maximum of one release. Once grpc-xds is published this will be rewritten to look like the other examples and just say "use the latest tag"

PR #6633, opened 2020-01-23
core: promote ServiceConfigErrorHandling

creamsoup:


creamsoup:
manual changes:
- {Hedging, Retry}PolicyTest use new method signature when using `ServiceConfigInterceptor`.
- removed opt-in mechanism. this is technically breaking api change, but it is never introduced. thus, the actual change is not really exposed from 3rd party point of view since the service config attribute is internal.

checked newer changes, the only change actually fixed both.

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -39,13 +42,12 @@
...
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
 
-@SuppressWarnings("deprecation") // migrate to AutoConfiguredLoadBalancerFactory2 is required
+@SuppressWarnings("deprecation") // after migrated to 2, we can deprecate it
```
zhangkun83: What deprecated usages are suppressed?

creamsoup: `LoadBalancer.ATTR_LOAD_BALANCING_CONFIG`. those files look like changed in this PR, but those files are just swapped from 2 to 1. i'll update the comment since the 2 doesn't make sense anymore

PR #6639, opened 2020-01-24
interop-testing: use server hostname instead of id for xds test

ericgribkoff:
Hostname is more informative for tests running on GCP, and unlike server ID, doesn't require the client to have knowledge about the server's command-line arguments

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/XdsTestClient.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/XdsTestClient.java
@@ -208,17 +208,17 @@ public final class XdsTestClient {
                 CallOptions.DEFAULT.withDeadlineAfter(rpcTimeoutSec, TimeUnit.SECONDS));
         call.start(
             new ClientCall.Listener<SimpleResponse>() {
-              private String serverId;
+              private String hostname;
 
               @Override
               public void onMessage(SimpleResponse response) {
-                serverId = response.getServerId();
+                hostname = response.getHostname();
                 // TODO(ericgribkoff) Currently some test environments cannot access the stats RPC
                 // service and rely on parsing stdout.
                 if (printResponse) {
                   System.out.println(
                       "Greeting: Hello world, this is "
-                          + response.getHostname()
+                          + hostname
                           + ", from "
                           + call.getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR));
                 }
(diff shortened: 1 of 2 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/XdsTestClient.java
```diff
@@ -295,14 +295,14 @@ private XdsStatsWatcher(long startId, long endId) {
...
-    void rpcCompleted(long requestId, @Nullable String serverId) {
+    void rpcCompleted(long requestId, @Nullable String hostname) {
       synchronized (lock) {
         if (startId <= requestId && requestId < endId) {
-          if (serverId != null) {
-            if (rpcsByPeer.containsKey(serverId)) {
-              rpcsByPeer.put(serverId, rpcsByPeer.get(serverId) + 1);
+          if (hostname != null) {
+            if (rpcsByPeer.containsKey(hostname)) {
+              rpcsByPeer.put(hostname, rpcsByPeer.get(hostname) + 1);
```
dapengzhang0: You might need use a generated random hostname as a fallback hostname in `XdsTestServer`.  As in https://github.com/grpc/grpc-java/commit/e7d7c5bde4caf606f02b2575d5f62e80ad5135b0#diff-2140bcb433e6510183658de7492bfb57R58

ericgribkoff: I think it's WAI as-is: the test driver script will need to assert on the hostnames recorded by the client, and so a randomly generated host won't be useful - if hostname is missing, the test will (and should) be failed.

dapengzhang0: If the hostname is missing, how will test fail? Is it easy to debug then?

ericgribkoff: The test spec is still in progress, but a sample scenario is something like: there are four backends, and we expect the client to round-robin over all four. If one of the hostnames is missing, the test will see that it is round-robining over three backends and fail.

ericgribkoff: Good point about debugging. Modified server to throw (fail) if it can't get hostname.

interop-testing/src/main/java/io/grpc/testing/integration/XdsTestServer.java
```diff
@@ -130,7 +130,8 @@ private TestServiceImpl() {
       try {
         host = InetAddress.getLocalHost().getHostName();
```
dapengzhang0: `host` can be final now.  But not a big deal.

ericgribkoff: done

PR #6645, opened 2020-01-27
xds: use resolvedAddresses.getLoadBalancingPolicyConfig() to get LB config

dapengzhang0:
`CdsLoadBalancer` was using `ATTR_LOAD_BALANCING_CONFIG` attributes to get service config. However, it did not work after 751faa6faa3fbccf46af74c6597218438f39c2ca and e9882ec78db581106f709ab5130bfe30af660e79 were merge (The cause should be fixed in #6648).


xds/src/main/java/io/grpc/xds/CdsLoadBalancer.java
```diff
@@ -328,7 +314,6 @@ public void onClusterChanged(ClusterUpdate newUpdate) {
           resolvedAddresses.toBuilder()
               .setAttributes(
                   resolvedAddresses.getAttributes().toBuilder()
```
creamsoup: now you can skip setAttributes.

dapengzhang0: Done.

PR #6646, opened 2020-01-27
interop-testing: log RPC failures for XdsTestClient

dapengzhang0:


interop-testing/src/main/java/io/grpc/testing/integration/XdsTestClient.java
```diff
@@ -226,6 +223,9 @@ public void onMessage(SimpleResponse response) {
 
               @Override
               public void onClose(Status status, Metadata trailers) {
+                if (!status.isOk()) {
+                  logger.log(Level.WARNING, "Greeting RPC failed with status {0}", status);
```
ericgribkoff: This seems good. I may need to add a wait-for-channel-ready mechanism in the client to avoid log spam when the server and client are brought up together, as the first n rpcs will fail before the server is ready. Will see if that becomes necessary.

PR #6649, opened 2020-01-27
all: log picker when updating balancing state

dapengzhang0:
Let `ManagedChannelImpl` log picker update when updating balancing state. (https://github.com/grpc/grpc-java/compare/master...dapengzhang0:log-picker?expand=1#diff-354509714620f5c493a6810aab9419f2R1156)

### DIFF
--- a/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
+++ b/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
@@ -431,6 +431,11 @@ public final class AutoConfiguredLoadBalancerFactory {
     public PickResult pickSubchannel(PickSubchannelArgs args) {
       return PickResult.withNoResult();
     }
+
+    @Override
+    public String toString() {
+      return "Empty picker";
+    }
   }
 
   private static final class FailingPicker extends SubchannelPicker {
--- a/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
+++ b/core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
@@ -434,7 +434,7 @@ public final class AutoConfiguredLoadBalancerFactory {
 
     @Override
     public String toString() {
-      return "Empty picker";
+      return MoreObjects.toStringHelper(EmptyPicker.class).toString();
     }
   }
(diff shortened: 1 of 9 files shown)


core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
```diff
@@ -431,6 +431,11 @@ public String toString() {
     public PickResult pickSubchannel(PickSubchannelArgs args) {
       return PickResult.withNoResult();
     }
+
+    @Override
+    public String toString() {
```
zhangkun83: The default toString() should work fine, or toStringHelper.

core/src/main/java/io/grpc/internal/OobChannel.java
```diff
@@ -180,6 +180,11 @@ public Object getInternalSubchannel() {
         public PickResult pickSubchannel(PickSubchannelArgs args) {
           return result;
         }
+
+        @Override
+        public String toString() {
+          return "Subchannel picker with pick result: " + result;
```
zhangkun83: Better to use toStringHelper in all cases to produce standardized outputs.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java
```diff
@@ -63,6 +63,11 @@ public void handleNameResolutionError(final Status error) {
             public PickResult pickSubchannel(PickSubchannelArgs args) {
               return PickResult.withError(error);
             }
+
+            @Override
+            public String toString() {
+              return "Error picker with status: " + error;
```
zhangkun83: ditto

core/src/main/java/io/grpc/internal/AutoConfiguredLoadBalancerFactory.java
dapengzhang0: Done.

core/src/main/java/io/grpc/internal/OobChannel.java
dapengzhang0: Done.

core/src/main/java/io/grpc/util/GracefulSwitchLoadBalancer.java

Issue #6268, opened 2019-10-10
Tracking Issue for Xds/SDS Channel/Server Builders being experimental


sanjaypujare:
### Additional context
For now (in the absence of XdsCredential) we need to use XdsChannelBuilder and XdsServerBuilder to be able to use XDS provided TlsContext information.

dapengzhang0:
These APIs are moved to internal. Closing it and may reopen or create a new issue once new public APIs need be added.

PR #6699, opened 2020-02-11
xds: refactor io.grpc.xds.sds to io.grpc.xds.internal.sds

dapengzhang0:


sanjaypujare:
will do this ASAP (today)

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -48,7 +45,6 @@
  * Provides client and server side gRPC {@link ProtocolNegotiator}s that use SDS to provide the SSL
  * context.
  */
-@Internal
 public final class SdsProtocolNegotiators {
```
sanjaypujare: This one doesn't need to be public any more, right?

xds/src/main/java/io/grpc/xds/internal/sds/SslContextProvider.java
```diff
@@ -33,7 +32,6 @@
  * stream that is receiving the requested secret(s) or it could represent file-system based
  * secret(s) that are dynamic.
  */
-@Internal
 public abstract class SslContextProvider<K> {
```
sanjaypujare: This too doesn't need to be public any more?

xds/src/main/java/io/grpc/xds/internal/sds/XdsChannelBuilder.java
```diff
@@ -32,7 +30,6 @@
  * A version of {@link ManagedChannelBuilder} to create xDS managed channels that will use SDS to
  * set up SSL with peers. Note, this is not ready to use yet.
  */
-@ExperimentalApi("https://github.com/grpc/grpc-java/issues/6268")
```
sanjaypujare: Why was this removed? Issue 6268 would still be open until we have implemented XdsCredentials API

xds/src/main/java/io/grpc/xds/internal/sds/XdsServerBuilder.java
```diff
@@ -40,7 +38,6 @@
  * A version of {@link ServerBuilder} to create xDS managed servers that will use SDS to set up SSL
  * with peers. Note, this is not ready to use yet.
  */
-@ExperimentalApi("https://github.com/grpc/grpc-java/issues/6268")
```
sanjaypujare: Same thing here. Until XdsCreds is implemented we would still treat this as an experimental API. Unless the fact that this is now in "internal" makes Experimental unnecessary but I think internal and experimental are 2 different things

xds/src/main/java/io/grpc/xds/internal/sds/XdsChannelBuilder.java
dapengzhang0: Because it's moved to internal. `@ExperimentalApi` is for APIs that are not internal.

dapengzhang0: We can close #6268 and open another once we have to add a public API in `io.grpc.xds`.

xds/src/main/java/io/grpc/xds/internal/sds/SslContextProvider.java
dapengzhang0: It is used by `io.grpc.xds.CdsLoadBalancer`.

xds/src/main/java/io/grpc/xds/internal/sds/XdsServerBuilder.java
dapengzhang0: We can close #6268 and open another once we have to add `XdsServerBuilder` in `io.grpc.xds`.
`@Experimental` means the public API is not stable. `@Interanl` is not "public" API and it's inherently unstable.

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
dapengzhang0: That's right. Thanks for the catch. Done.

PR #6713, opened 2020-02-14
interop-testing: sync SimpleResponse definition with core and go

ericgribkoff:
The `grpclb_route_type` field already exists in the core and Go repos.

### DIFF
--- a/interop-testing/src/main/proto/grpc/testing/messages.proto
+++ b/interop-testing/src/main/proto/grpc/testing/messages.proto
@@ -41,6 +41,21 @@ message EchoStatus {
   string message = 2;
 }
 
+// The type of route that a client took to reach a server w.r.t. gRPCLB.
+// The server must fill in "fallback" if it detects that the RPC reached
+// the server via the "gRPCLB fallback" path, and "backend" if it detects
+// that the RPC reached the server via "gRPCLB backend" path (i.e. if it got
+// the address of this server from the gRPCLB server BalanceLoad RPC). Exactly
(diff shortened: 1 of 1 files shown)


creamsoup:
this looks like a braking change. i am surprised that it worked before. probably, our tests don't set this value?

anyways, the proper way of fixing this is 
modify [sync-protos script](https://github.com/grpc/grpc-java/blob/master/buildscripts/sync-protos.sh) by adding `interop-testing` to the projects. and run the script. (for this time, we need to make sure deleted things are correct or not). we may need to upstream part of deleted (deleted is modified in grpc-java only) to the `grpc-proto` (e.g. add java_package or LoadBalancerStatsRequest/Response?).

i am not sure how urgent this change need to be merged, if we have time let's fix it correct way.

creamsoup:
> this looks like a braking change. i am surprised that it worked before. probably, our tests don't set this value?
> 
> anyways, the proper way of fixing this is
> modify [sync-protos script](https://github.com/grpc/grpc-java/blob/master/buildscripts/sync-protos.sh) by adding `interop-testing` to the projects. and run the script. (for this time, we need to make sure deleted things are correct or not). we may need to upstream part of deleted (deleted is modified in grpc-java only) to the `grpc-proto` (e.g. add java_package or LoadBalancerStatsRequest/Response?).
> 
> i am not sure how urgent this change need to be merged, if we have time let's fix it correct way.

discussed offline. it looks like the proto is fairly diverged between different grpc langauges. created #6719

PR #6718, opened 2020-02-14
xds: support for xDS data types needed for server side SDS support

sanjaypujare:
first part that includes xDS data definitions corresponding to protos


xds/src/test/java/io/grpc/xds/internal/sds/CommonTlsContextTestsUtil.java
```diff
@@ -134,4 +115,27 @@ static CommonTlsContext buildCommonTlsContextWithAdditionalValues(
     }
     return builder.build();
   }
+
+  /**
+   * Helper method to build DownstreamTlsContext for above tests. Called from other classes as well.
```
creamsoup: above tests? you mean tests in SdsSslContextProviderTest/ScretVolumnSslContextProviderTest?

sanjaypujare:
> the title of PR doesn't match what I reviewed, can you update the title?

done

xds/src/main/java/io/grpc/xds/EnvoyProtoData.java
```diff
@@ -339,4 +339,270 @@ public String toString() {
...
+  static final class CidrRange {
+    private final String addressPrefix;
+    private final int prefixLen;
+
+    public CidrRange(String addressPrefix, int prefixLen) {
+      this.addressPrefix = addressPrefix;
+      this.prefixLen = prefixLen;
+    }
+
+    static CidrRange fromEnvoyProtoCidrRange(
```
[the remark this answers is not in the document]
sanjaypujare: First, I used the existing pattern in `EnvoyProtoData` which seems workable.

Separating the conversion into a Converter class seems like a good idea and can be implemented when v3 protos are introduced. Also I am not sure how different/similar v3 protos will be to existing ones. There is a distinct possibility that v3 protos would force us to also update our internal xDS data types (because of new concepts) so it might not be worth our while to develop a complex conversion structure.

xds/src/main/java/io/grpc/xds/EnvoyProtoData.java
```diff
@@ -339,4 +339,270 @@ public String toString() {
...
+          + ", downstreamTlsContext=" + downstreamTlsContext
+          + '}';
+    }
+  }
+
+  /**
+   * Corresponds to Envoy proto message {@link io.envoyproxy.envoy.api.v2.Listener} & related
+   * classes.
+   */
+  static final class Listener {
```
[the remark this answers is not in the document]
sanjaypujare: I can certainly rename it but the idea was to match the name used in Envoy protos (so reader can easily map one to the other) and there is no possibility of name collisions if all (most) of xDS code will only use this `Listener` class and not the Envoy one. BTW, `Listener` may sound too generic but has a specific meaning in the xDS world

xds/src/main/java/io/grpc/xds/EnvoyProtoData.java
```diff
@@ -339,4 +339,270 @@ public String toString() {
           .toString();
     }
   }
+
+  static final class CidrRange {
```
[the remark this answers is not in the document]
sanjaypujare: This is not SDS specific: only that SDS will be the first one to use it. This is server side Listener stuff that is going to be needed for more server side features. We can split client vs server. @voidzcy what do you think?

xds/src/test/java/io/grpc/xds/internal/sds/CommonTlsContextTestsUtil.java
sanjaypujare: sorry  - let me clarify

xds/src/main/java/io/grpc/xds/EnvoyProtoData.java
```diff
@@ -339,4 +339,271 @@ public String toString() {
...
+    private final List<FilterChain> filterChains;
+
+    private Listener(String name, String address,
+        List<FilterChain> filterChains) {
+      this.name = name;
+      this.address = address;
+      this.filterChains = filterChains;
+    }
+
+    static String fromEnvoyProtoAddress(io.envoyproxy.envoy.api.v2.core.Address proto) {
```
voidzcy: This seems to be a helper method only. Should be private and name it to something different than `fromEnvoyProtoXXX`. Here `fromEnvoyProtoXXX` methods are used as factory methods to create converted data objects. 

Btw, since this helper method is only used once and it's not complex, it can be inlined instead of being a standalone method.

xds/src/main/java/io/grpc/xds/EnvoyProtoData.java
```diff
@@ -339,4 +339,271 @@ public String toString() {
...
+    }
+  }
+
+  /**
+   * Corresponds to Envoy proto message {@link io.envoyproxy.envoy.api.v2.Listener} & related
+   * classes.
+   */
+  static final class Listener {
+    private final String name;
+    private final String address;
```
voidzcy: Is this nullable?

xds/src/main/java/io/grpc/xds/EnvoyProtoData.java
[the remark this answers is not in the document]
voidzcy: Hmm... It is subtle. There is no clear distinction of which protos are used for client side and which are for server side. Some of them are used both client and server sides. Yes, I agree we probably do not want to dump everything here. Do we have a better solution? @creamsoup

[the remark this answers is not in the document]
creamsoup: i like to have bunch of smaller classes. so i would split into xdsProtoData like EdsProtoData, CdsProtoData, etc. eventually something similar to envoy proto's package structure (more like flattened way). for this PR, possibly `SdsProtoData` will have SDS specific protos. this can help readers to understand which xds service it interacts with. 

i used this converter / data / proto separation in bazel, it works very well. but, not so well in other build system imo.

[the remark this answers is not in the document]
sanjaypujare: @creamsoup can you clarify your comment? Preferring "bunch of smaller classes" is actually opposite of "flattened", so I am a bit confused. Also talking to @ejona86 not having smaller classes (on the xDS side) seems to be the preferred way. So on the xDS side we should create definitions that match our usage instead of just mirroring those on the Envoy proto side.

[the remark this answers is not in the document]
creamsoup: oops yes.
flattened as package level. the data classes from various xds protos (has many sub-packages) can be in a single java package. but they can have their own classes such as Cds, Eds, Lds, Rds, Srds, Discovery, Common (common for listener, endpoint etc) etc. 

what do you mean not having smaller classes seems to be preferred? i meant the smaller as each DS (sorry if this was not clear), current code is somewhat in that structure (Cds, Eds, Sds, etc).  
i am referring envoy proto structure, because it is organized in the usage not having everything in one class. seems like you are interpreting it in too extreme case (all in one class vs as tiny as message?)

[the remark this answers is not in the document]
sanjaypujare: @creamsoup okay, thanks for the clarification. So regarding not preferring smaller classes, let's consider an example: `Listener` contains (list of) `FilterChain` and a `FilterChain` contains `DownstreamTlsContext` in the original Envoy protos. We don't have to replicate this structure on our side but just define the top level `Listener` with all fields expanded "inline" unless the types `FilterChain` and `DownstreamTlsContext` are required elsewhere. Note this is just an example to illustrate the point.

On the Envoy side, they are not defined by us and so have less control over how types/protos are defined - but we don't have to follow the same structure.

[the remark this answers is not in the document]
creamsoup: yes, that's why i specifically says "similar structure". if it is same, no reason to have data classes. also, this is why I suggested to rename the Listener to something else.
your response still sounds like you are thinking too micro level, my comments was for overall structure. I completely agree that Listener can have its own fields as long as it can represent current Listener structure. What I was saying is this `EnvoyProtoData` is too big. a single class represents the whole envoy protos sounds very extreme, so we may need to split into smaller chunks.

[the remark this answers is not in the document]
sanjaypujare: Created a separate EnvoyServerProtoData class

xds/src/main/java/io/grpc/xds/EnvoyProtoData.java
sanjaypujare: good point. fixed

xds/src/main/java/io/grpc/xds/EnvoyProtoData.java
sanjaypujare: Made changes (except inlining because in future we may have an internal Address class)

xds/src/main/java/io/grpc/xds/EnvoyServerProtoData.java
```diff
@@ -0,0 +1,303 @@
...
+      for (io.envoyproxy.envoy.api.v2.core.CidrRange range : proto.getPrefixRangesList()) {
+        prefixRanges.add(CidrRange.fromEnvoyProtoCidrRange(range));
+      }
+      List<String> applicationProtocols = new ArrayList<>();
+      for (String appProtocol  : proto.getApplicationProtocolsList()) {
+        applicationProtocols.add(appProtocol);
+      }
+      return new FilterChainMatch(
+          proto.getDestinationPort().getValue(),
+          Collections.unmodifiableList(prefixRanges),
```
creamsoup: i expected to have it (unmodifiableList) in the constructor, just in case we have another factory method it can be missed.

xds/src/test/java/io/grpc/xds/EnvoyServerProtoDataTest.java
```diff
@@ -0,0 +1,134 @@
...
+    EnvoyServerProtoData.FilterChainMatch inFilterChainMatch = inFilter.getFilterChainMatch();
+    assertThat(inFilterChainMatch).isNotNull();
+    assertThat(inFilterChainMatch.getDestinationPort()).isEqualTo(8000);
+    assertThat(inFilterChainMatch.getApplicationProtocols()).containsExactly("managed-mtls");
+    assertThat(inFilterChainMatch.getPrefixRanges()).containsExactly(
+        new EnvoyServerProtoData.CidrRange("10.20.0.15", 32));
+    DownstreamTlsContext inFilterTlsContext = inFilter.getDownstreamTlsContext();
+    assertThat(inFilterTlsContext).isNotNull();
+    CommonTlsContext commonTlsContext = inFilterTlsContext.getCommonTlsContext();
+    assertThat(commonTlsContext).isNotNull();
```
voidzcy: Proto fields will never be null.

Anyway, it's not necessary to write tests for this data conversion class (unless there is some complex logic other than straightly populating data from proto messages).

sanjaypujare: Tests are meant to catch potential regressions in the future

creamsoup: i agree the statement, but it doesn't apply to this case. just like we don't unittest standard library. we will probably have a fix from next import.

sanjaypujare: My comment was about "not needing a test" or "test not being useful". I agree that the "NotNull" checks can be (should be?) removed based on linter issues.

creamsoup: oops, yes i believe the conversion tests are useful.
other test will verify behavior using converted data class. without proto to converted verification, we can't trust other tests.

Issue #6725, opened 2020-02-19
Update okhttp version from 2.5.0 to 2.7.4


sourabhsparkala:
This update of `com.squareup.okhttp:okhttp:jar` version from `2.5.0` to `2.7.4` would help resolve a reported vulnerability [CVE-2016-2402](https://nvd.nist.gov/vuln/detail/CVE-2016-2402)

It would help us a great deal, if you can update the version
As shown in the mvn tree
```
| | +- io.grpc:grpc-okhttp:jar:1.17.1:compile
| | | \- com.squareup.okhttp:okhttp:jar:2.5.0:compile
```
Also, we would appreciate it if you can let us know, whether it possible to do a quick release as early as possible.

Requesting you to downport the `com.squareup.okhttp:okhttp:jar` version fix update to **`io.grpc:grpc-okhttp:jar:1.17.1`** as well

ericgribkoff:
gRPC Java uses very little of the internals of okhttp; in particular, we do not use OkHttp's certificate pinning so this vulnerability does not apply to gRPC Java users. Please see the discussion from https://github.com/grpc/grpc-java/issues/6119#issuecomment-541794420 for more details around what (minimal) parts of OkHttp we actually use.

sourabhsparkala:
> gRPC Java uses very little of the internals of okhttp; in particular, we do not use OkHttp's certificate pinning so this vulnerability does not apply to gRPC Java users. Please see the discussion from [#6119 (comment)](https://github.com/grpc/grpc-java/issues/6119#issuecomment-541794420) for more details around what (minimal) parts of OkHttp we actually use.

Thank you for the clarification 👍

creamsoup:
upgraded is done vis #6726, as discussed above and in the PR. we don't need to patch release this fix because grpc-okhttp is not affected by the linked CVE.

PR #6726, opened 2020-02-19
Updating okhttp version from 2.5.0 to 2.7.4 to fix [CVE-2016-2402]

sourabhsparkala:
Addressing Issue https://github.com/grpc/grpc-java/issues/6725

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -125,7 +125,7 @@ subprojects {
             google_api_protos: 'com.google.api.grpc:proto-google-common-protos:1.17.0',
             google_auth_credentials: "com.google.auth:google-auth-library-credentials:${googleauthVersion}",
             google_auth_oauth2_http: "com.google.auth:google-auth-library-oauth2-http:${googleauthVersion}",
-            okhttp: 'com.squareup.okhttp:okhttp:2.5.0',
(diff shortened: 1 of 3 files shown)


creamsoup:
@ejona86 i can't imagine how this cannot already be detected/fixed since it was reported long time ago, do you have any idea? the internal fork has fix or not affected by this?
anyways, users can use this transient dependency even if this is fixed, so we should fix this.

creamsoup:
seems like we are not using the affected code since we are using `SSLSocket`, so this is not really affecting user unless your's newer version of okhttp is downgraded by maven. but, i would wait @ejona86 's comfirmation.

creamsoup:
yes we should upgrade, but we may need to do a patch release if we are not affected.

creamsoup:
merged, thanks @sourabhsparkala!

sourabhsparkala:
> merged, thanks @sourabhsparkala!

👍

ejona86:
@sourabhsparkala, it looks like this made it in before the branch cut, so it is in [the v1.28.x branch](https://github.com/grpc/grpc-java/commits/v1.28.x). It will be in v1.28.0 in a bit over a week.

sourabhsparkala:
@ejona86  thank you for the update.

PR #6731, opened 2020-02-20
xds: implement WeightedTargetLoadBalancer

dapengzhang0:
- Implement weighted_target lb part of go/grpc-xds-traffic-splitting-and-path-matching 
- Refactor `InterLocalityPicker` to reuse the `RandomWeightedPicker`.


xds/src/main/java/io/grpc/xds/WeightedTargetLoadBalancer.java
```diff
@@ -0,0 +1,216 @@
...
+      WeightedChildLbConfig weightedChildLbConfig = newTargets.get(targetName);
+      if (!targets.containsKey(targetName)) {
+        // Create child balancers for new names.
+        ChildHelper childHelper = new ChildHelper();
+        GracefulSwitchLoadBalancer childBalancer = new GracefulSwitchLoadBalancer(childHelper);
+        childBalancer.switchTo(lbRegistry.getProvider(weightedChildLbConfig.policyName));
+        childHelpers.put(targetName, childHelper);
+        childBalancers.put(targetName, childBalancer);
+      } else if (!weightedChildLbConfig.policyName.equals(targets.get(targetName).policyName)) {
+        // Policy name change for the same target is not a typical usecase, but we should support
```
voidzcy: nit: commenting is redundant. Same for above and below. Code is already clearly saying what it is doing.

xds/src/main/java/io/grpc/xds/WeightedTargetLoadBalancer.java
```diff
@@ -0,0 +1,216 @@
...
+      picker = weightedPickerFactory.picker(childPickers);
+    }
+
+    if (overallState != null) {
+      helper.updateBalancingState(overallState, picker);
+    }
+  }
+
+  @Nullable
+  private ConnectivityState aggregateState(
```
voidzcy: We may want to make a utility method for this, as it is used in a couple of places (for any load balancers that contain a bunch of child balancers).

xds/src/main/java/io/grpc/xds/RandomWeightedPicker.java
```diff
@@ -27,7 +27,7 @@
 import io.grpc.LoadBalancer.SubchannelPicker;
 import java.util.List;
 
-final class InterLocalityPicker extends SubchannelPicker {
+final class RandomWeightedPicker extends SubchannelPicker {
```
voidzcy: `WeightedRandomPicker` sounds better.

xds/src/main/java/io/grpc/xds/RandomWeightedPicker.java
```diff
@@ -116,4 +116,17 @@ public String toString() {
         .add("totalWeight", totalWeight)
         .toString();
   }
+
+  /** Factory that creates a SubchannelPicker for a given list of weighted child pickers. */
+  interface WeightedPickerFactory {
```
voidzcy: Maybe we do not need both `RandomWeightedPicker` and `WeightedPickerFactory` (I feel like we are introducing many classes/interfaces that do not seem to be necessary). 

```java
final class WeightedRamdomPicker {  // or some better class name
   private final Random rand;
   WeightedRamdonPicker(Random rand) {
      this.rand = rand;
   }
   SubchannelPicker picker(List<WeightedChildPicker> childPickers) {  // or some better method name
      // compute total weight for the list of child pickers, randomly choose one.
   }
}
```

xds/src/main/java/io/grpc/xds/WeightedTargetLoadBalancer.java
```diff
@@ -0,0 +1,216 @@
...
+
+  private final class ChildHelper extends ForwardingLoadBalancerHelper {
+    ConnectivityState currentState = CONNECTING;
+    SubchannelPicker currentPicker = BUFFER_PICKER;
+
+    @Override
+    public void updateBalancingState(ConnectivityState newState, SubchannelPicker newPicker) {
+      currentState = newState;
+      currentPicker = newPicker;
+      WeightedTargetLoadBalancer.this.updateBalancingState();
```
voidzcy: nit: you don't need `WeightedTargetLoadBalancer.this` (may need to change the method name `updateBalancingState()`).

xds/src/main/java/io/grpc/xds/WeightedTargetLoadBalancer.java
dapengzhang0: This is related to #6650. Might do it after the design in general is clear.

xds/src/main/java/io/grpc/xds/WeightedTargetLoadBalancerProvider.java
```diff
@@ -0,0 +1,236 @@
...
+
+/**
+ * The provider for the weighted_target balancing policy.  This class should not be
+ * directly referenced in code.  The policy should be accessed through {@link
+ * LoadBalancerRegistry#getProvider} with the name "weighted_target_experimental".
+ */
+@Internal
+public final class WeightedTargetLoadBalancerProvider extends LoadBalancerProvider {
+
+  static final String WEIGHTED_TARGET_POLICY_NAME = "weighted_target_experimental";
```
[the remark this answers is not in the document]
dapengzhang0: This is like EDS_POLICY_NAME, which is used by cds lb. CDS_POLICY_NAME will be used by weighted-target-lb. WEIGHTED_TARGET_POLICY_NAME will be used by xds-routing lb. Now only XDS_POLICY_NAME won't be used by other lb.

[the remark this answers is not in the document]
voidzcy: I don't think usages like [`CdsLoadBalancer`](https://github.com/grpc/grpc-java/blob/4a2c5d6e9c0f90de961bc60f2549582e5da1e803/xds/src/main/java/io/grpc/xds/CdsLoadBalancer.java#L295)) is a good practice. `LoadBalancerRegistry` is used for looking up some LB policy given the policy name, the caller treats the policy name opaquely (it just blindly as the registry to give the corresponding LB provider). In CDS LB, it's the design/implementation's choice to load EDS policy. CDS policy explicitly knows what it wants and what that is. It should call `new EdsLoadBalancerProvider()` directly (or even `EdsLoadBalancer(...)` constructor). Existing way of instantiating child policy is an abuse of registry.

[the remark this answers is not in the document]
dapengzhang0: There are two reasons using register:

- The EDS balancer is a plugin, the built-in implementation can be deregistered and re-registered with a custom implementation plugin with EDS_POLICY_NAME. 

- In testing, we can use a fake EDS balancer for testing CDS.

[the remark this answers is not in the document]
dapengzhang0: If you look at the javadoc 

```java
/**
   * Register a provider.
   *
   * <p>If the provider's {@link LoadBalancerProvider#isAvailable isAvailable()} returns
   * {@code false}, this method will throw {@link IllegalArgumentException}.
   *
   * <p>If more than one provider with the same {@link LoadBalancerProvider#getPolicyName policy
   * name} are registered, the one with the highest {@link LoadBalancerProvider#getPriority
   * priority} will be effective.  If there are more than one name-sake providers rank the highest
   * priority, the one registered first will be effective.
   */
  public synchronized void register(LoadBalancerProvider provider)
```

Even "round_robin" can be re-registered with different implementations, if multiple round robin lbs are registered, it will pick the one with highest priority.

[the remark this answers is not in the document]
voidzcy: Alright, I accept the current way of looking up registry for creating the downstream EDS policy.

xds/src/main/java/io/grpc/xds/WeightedRandomPicker.java
```diff
@@ -62,12 +62,12 @@ public String toString() {
     }
   }
 
-  InterLocalityPicker(List<WeightedChildPicker> weightedChildPickers) {
+  WeightedRandomPicker(List<WeightedChildPicker> weightedChildPickers) {
```
voidzcy: This class structure does not help with testing at all, as in order to provide your own random number generator in test, you have to implement a fake `WeightedPickerFactory` instance whose `picker(...)` method creates a fake `SubchannelPicker`. So everything about picking is fake. This tests nothing. 

Instead, a better strategy is to not have this constructor at all. You implementation code should always call the constructor with injectable random number generator. In that way, your test code is  still executing the real `WeightedRandomPicker` but only with the random number generator being faked out.

xds/src/main/java/io/grpc/xds/WeightedTargetLoadBalancer.java
```diff
@@ -0,0 +1,205 @@
...
+  private final Map<String, ChildHelper> childHelpers = new HashMap<>();
+  private final Helper helper;
+  private final WeightedPickerFactory weightedPickerFactory;
+
+  private Map<String, WeightedPolicySelection> targets = ImmutableMap.of();
+
+  WeightedTargetLoadBalancer(Helper helper) {
+    this(
+        checkNotNull(helper, "helper"),
+        WeightedRandomPicker.RANDOM_PICKER_FACTORY);
```
voidzcy: This looks bad... You should rather call constructor to create a factory or call the factory class to get the instance. At least, not something that is not the factory itself.

xds/src/main/java/io/grpc/xds/WeightedRandomPicker.java
dapengzhang0: `WeightedPickerFactory` does not claim to test anything that `WeightedRandomPicker` does. What `WeightedRandomPicker` does is well-tested in its own test with a fake random. Using `WeightedPickerFactory` in `LocalityStoreTest`/`WeightedTargetLoadBalancerTest` is to avoid re-testing `WeightedRandomPicker`. Passing a fake random instead of the factory to`LocalityStore`/`WeightedTargetLoadBalancer` in test, you need recalculate the relation between the weights and the fake random number, that's basically following the implementation details of `WeightedRandomPicker`.

We have argued about this a lot of times. When class `A` uses class `B` where `B` is well-tested with its own `BTest`. Then `ATest` will just use `BFactory`. `BFactory` does not claim to test anything of `B`. `ATest` does not claim to test any implementation detail encapsulated in `B` Using `B`'s constructor arg instead of `BFactory` in `ATest`, you are retesting `B` again.

Now I revised the class not to use a factory, but I'm still not going to retest the well-encapsulated and well-tested class before `EdsLoadBalancer` is refactored to use `WeightedTargetLoadBalancer`.

xds/src/main/java/io/grpc/xds/RandomWeightedPicker.java
voidzcy: This comment is outdated.

xds/src/main/java/io/grpc/xds/WeightedRandomPicker.java
voidzcy: This test strategy is poor. It should be used only when the subcomponent `B` is heavy (i.e., small interface with large implementation). `WeightedRandomPicker` is completely an internal implementation of `WeightedTargetLoadBalancer`/`LocalityStore`, even though you write it as a modularized subcomponent. It's better even if `WeightedRandomPicker` is inlined in `WeightedTargetLoadBalancer`/`LocalityStore`. But in terms of code reuse, put it into a modularized class. Its implementation is not complex (even somewhat trivial), it's just using a random number generator to choose something in an input list. It is more like a utility. We are not keenly interested in testing the isolated behaviors of this sub-component. Instead, we are more interested in the real output of `WeigtedTargetLoadBalancer`/`LocalityStore`, which is the content of picker that eventually propagated to Channel (i.e., `helper.updateBalancingState(State, Picker)`).

The above is just my opinion. I am not going to block you on tests, as long as the class structure is obviously inappropriate. But really, existing tests test barely little real things...

PR #6746, opened 2020-02-24
compiler: add std:: qualifications to all references to std::string

creamsoup:


creamsoup:
cc @ransu we can remove `using namespace std;` in the `compiler.diff`. (when import add me as a reviewer)

ejona86:
`compiler.diff` _deletes_ `using namespace std;`. That needs to stay in the diff.

But we could potentially have a followup PR that removes `using namespace std;` from github, which would mean it wouldn't be needed in the diff any more. But all that should be separate.

PR #6771, opened 2020-02-27
Enable use of InProcessTransport outside of InProcessServer.

markb74:
This allows an InProcessTransport instance to be created via a new internal accessor class InternalInProcess. We effectively just expose a method to create an InProcessTransport with a existing ServerListener instance.

Will be used for in-process channels to an under-development on-device server.


markb74:
Re-sending this from a fresh branch, without the metadata commit history.

PR #6772, opened 2020-02-27
deps: update guava to 28.2

elharo:
@creamsoup


creamsoup:
merged, thanks again @elharo

Issue #6765, opened 2020-02-27
MethodDescriptor.setRawMethodName breaks Immutability if multiple types of transport are used


dapengzhang0:
```
java.lang.ClassCastException: io.netty.util.AsciiString cannot be cast to io.grpc.netty.shaded.io.netty.util.AsciiString
	at io.grpc.netty.shaded.io.grpc.netty.NettyClientStream$Sink.writeHeadersInternal(NettyClientStream.java:130)
	at io.grpc.netty.shaded.io.grpc.netty.NettyClientStream$Sink.writeHeaders(NettyClientStream.java:122)
	at io.grpc.internal.AbstractClientStream.start(AbstractClientStream.java:166)
	at io.grpc.internal.ForwardingClientStream.start(ForwardingClientStream.java:87)
	at io.grpc.internal.InternalSubchannel$CallTracingTransport$1.start(InternalSubchannel.java:645)
	at io.grpc.internal.DelayedStream$5.run(DelayedStream.java:229)
	at io.grpc.internal.DelayedStream.drainPendingCalls(DelayedStream.java:163)
	at io.grpc.internal.DelayedStream.setStream(DelayedStream.java:132)
	at io.grpc.internal.DelayedClientTransport$PendingStream.createRealStream(DelayedClientTransport.java:358)
	at io.grpc.internal.DelayedClientTransport$PendingStream.access$300(DelayedClientTransport.java:341)
	at io.grpc.internal.DelayedClientTransport$5.run(DelayedClientTransport.java:300)
	at io.grpc.stub.ClientCalls$ThreadlessExecutor.waitAndDrain(ClientCalls.java:691)
	at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:133)
	at io.grpc.testing.integration.TestServiceGrpc$TestServiceBlockingStub.unaryCall(TestServiceGrpc.java:637)
	at io.grpc.testing.integration.CompressionTest.compression(CompressionTest.java:202)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #6774, opened 2020-02-27
api,netty: fix MethodDescriptor and InternalKnownTransport for netty-shaded

dapengzhang0:
Continue to use enum `InternalKnownTransport` and extend it for ABI backward compatibility (hopefully).

### DIFF
--- a/api/src/main/java/io/grpc/MethodDescriptor.java
+++ b/api/src/main/java/io/grpc/MethodDescriptor.java
@@ -51,9 +51,7 @@ public final class MethodDescriptor<ReqT, RespT> {
 
   // Must be set to InternalKnownTransport.values().length
   // Not referenced to break the dependency.
-  private final AtomicReferenceArray<Object> rawMethodNames =
-      new AtomicReferenceArray<>(InternalKnownTransport.values().length);
-
+  private final AtomicReferenceArray<Object> rawMethodNames = new AtomicReferenceArray<>(2);
 
   /**
    * Gets the cached "raw" method name for this Method Descriptor.  The raw name is transport
--- a/api/src/main/java/io/grpc/MethodDescriptor.java
+++ b/api/src/main/java/io/grpc/MethodDescriptor.java
@@ -51,7 +51,8 @@ public final class MethodDescriptor<ReqT, RespT> {
 
   // Must be set to InternalKnownTransport.values().length
   // Not referenced to break the dependency.
-  private final AtomicReferenceArray<Object> rawMethodNames = new AtomicReferenceArray<>(1);
+  private final AtomicReferenceArray<Object> rawMethodNames =
+      new AtomicReferenceArray<>(InternalKnownTransport.values().length);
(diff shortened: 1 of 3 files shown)


netty/src/main/java/io/grpc/netty/NettyClientStream.java
```diff
@@ -54,7 +54,9 @@
  */
 class NettyClientStream extends AbstractClientStream {
   private static final InternalMethodDescriptor methodDescriptorAccessor =
-      new InternalMethodDescriptor(InternalKnownTransport.NETTY);
+      new InternalMethodDescriptor(
+          NettyClientTransport.class.getName().contains("grpc.netty.shaded")
+              ? InternalKnownTransport.NETTY_SHADED : InternalKnownTransport.NETTY);
```
dapengzhang0: I'm not sure if it's possible that the compiler can optimize this `contains()?:` expression and replace it with `InternalKnownTransport.NETTY` directly before shadow plugin does its job. My local interop test did not show it happened.

dapengzhang0:
Resolves #6765

api/src/main/java/io/grpc/MethodDescriptor.java
```diff
@@ -51,7 +51,8 @@
 
   // Must be set to InternalKnownTransport.values().length
   // Not referenced to break the dependency.
-  private final AtomicReferenceArray<Object> rawMethodNames = new AtomicReferenceArray<>(1);
+  private final AtomicReferenceArray<Object> rawMethodNames =
+      new AtomicReferenceArray<>(InternalKnownTransport.values().length);
```
ejona86: This will break Bazel; use a literal `2`.

The comment alludes to why this was done, but even if you thought that no longer applied the comment should have been removed.

netty/src/main/java/io/grpc/netty/NettyClientStream.java
ejona86: No, it won't be able to rework this string. You could also just do `"shaded"` as another option. But all those variations should be fine.

api/src/main/java/io/grpc/MethodDescriptor.java
dapengzhang0: `InternalKnownTransport` is part of grpc-api so Bazel seems not impacted. It breaks blaze though.

dapengzhang0: Just noticed there were already comments on this field
```java
// Must be set to InternalKnownTransport.values().length
// Not referenced to break the dependency.
private final AtomicReferenceArray<Object> rawMethodNames =
```

netty/src/main/java/io/grpc/netty/NettyClientStream.java
dapengzhang0: Just "shaded" will produce false positive if grpc as a whole is shaded by the application.

PR #6780, opened 2020-02-27
all: let interop test use shaded dependency correctly

dapengzhang0:


netty/shaded/build.gradle
```diff
@@ -18,7 +18,6 @@ dependencies {
             project(':grpc-testing-proto'),
             project(':grpc-testing'),
             libraries.truth
-    shadow project(':grpc-core')
```
ejona86: I figured out why this is here: it is necessary for the pom to include a dependency on grpc-core.

xds/build.gradle
```diff
@@ -39,11 +38,16 @@ dependencies {
 
     testCompile project(':grpc-core').sourceSets.test.output
 
-    compileOnly libraries.javax_annotation
+    compileOnly libraries.javax_annotation,
+            // The unshaded netty_epoll should not be a transitive dependency.
```
ejona86: This comment sort of just reexplains what a compileOnly dependency is. Instead how about:
```
      // At runtime use the epoll included in grpc-netty-shaded
```

netty/shaded/build.gradle
dapengzhang0: I see. Changed back.

dapengzhang0:
gae-interop-testing was failing https://source.cloud.google.com/results/invocations/24432bbe-0f09-4149-89dd-7190c766cf73/targets
because the test uses ManagedChannelBuilder and then NettyChannelBuilder
```java
assertTrue(builder instanceof NettyChannelBuilder);
((NettyChannelBuilder) builder).flowControlWindow(65 * 1024);
```

ejona86:
This PR caused a regression in the ALTS tests. https://source.cloud.google.com/results/invocations/691d9965-fea1-487d-b606-352a5234039e/targets/grpc%2Fcore%2Fpull_request%2Flinux%2Fgrpc_interop_toprod/log

```
2020-03-01 20:02:12,491 Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/commons/lang3/SystemUtils
	at io.grpc.alts.CheckGcpEnvironment.isRunningOnGcp(CheckGcpEnvironment.java:69)
	at io.grpc.alts.CheckGcpEnvironment.isOnGcp(CheckGcpEnvironment.java:44)
	at io.grpc.alts.ComputeEngineChannelBuilder.<init>(ComputeEngineChannelBuilder.java:62)
	at io.grpc.alts.ComputeEngineChannelBuilder.forTarget(ComputeEngineChannelBuilder.java:72)
	at io.grpc.alts.ComputeEngineChannelBuilder.forAddress(ComputeEngineChannelBuilder.java:77)
	at io.grpc.testing.integration.TestServiceClient$Tester.createChannel(TestServiceClient.java:399)
	at io.grpc.testing.integration.AbstractInteropTest.setUp(AbstractInteropTest.java:309)
	at io.grpc.testing.integration.TestServiceClient.setUp(TestServiceClient.java:198)
	at io.grpc.testing.integration.TestServiceClient.main(TestServiceClient.java:56)
Caused by: java.lang.ClassNotFoundException: org.apache.commons.lang3.SystemUtils
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 9 more
```

It can be trivially reproduced by running on your machine (no need for any other setup):
```
$ ./build/install/grpc-interop-testing/bin/test-client --use_alts=true
```

@dapengzhang0, we should probably revert this change while we try to figure out the solution.

vjpai:
Cc: @jtattermusch

dapengzhang0:
Trying to fix the issue of missing transitive dependencies for shadow configuration in #6792.

PR #6791, opened 2020-03-02
Revert "all: let interop test use shaded dependency correctly (#6780)"

dapengzhang0:
This reverts commit c5f48b8e384ca1ad91b67886f8cbff27c9bcfd8d.

Revert because caused a regression in the ALTS tests. https://source.cloud.google.com/results/invocations/691d9965-fea1-487d-b606-352a5234039e/targets/grpc%2Fcore%2Fpull_request%2Flinux%2Fgrpc_interop_toprod/log

2020-03-01 20:02:12,491 Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/commons/lang3/SystemUtils
	at io.grpc.alts.CheckGcpEnvironment.isRunningOnGcp(CheckGcpEnvironment.java:69)
	at io.grpc.alts.CheckGcpEnvironment.isOnGcp(CheckGcpEnvironment.java:44)
	at io.grpc.alts.ComputeEngineChannelBuilder.<init>(ComputeEngineChannelBuilder.java:62)
	at io.grpc.alts.ComputeEngineChannelBuilder.forTarget(ComputeEngineChannelBuilder.java:72)
	at io.grpc.alts.ComputeEngineChannelBuilder.forAddress(ComputeEngineChannelBuilder.java:77)
	at io.grpc.testing.integration.TestServiceClient$Tester.createChannel(TestServiceClient.java:399)
	at io.grpc.testing.integration.AbstractInteropTest.setUp(AbstractInteropTest.java:309)
	at io.grpc.testing.integration.TestServiceClient.setUp(TestServiceClient.java:198)
	at io.grpc.testing.integration.TestServiceClient.main(TestServiceClient.java:56)
Caused by: java.lang.ClassNotFoundException: org.apache.commons.lang3.SystemUtils
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 9 more

PR #6792, opened 2020-03-02
all: let interop test use shaded dependency correctly take 2

dapengzhang0:
This include two commits: 

- first rolls forward #6791 
- then fixes the issue of missing transitive dependency for the shadow configuration.

### DIFF
--- a/alts/build.gradle
+++ b/alts/build.gradle
@@ -38,6 +38,7 @@ dependencies {
             libraries.truth
     testRuntime libraries.netty_tcnative,
             libraries.netty_epoll
+    shadow project(path: ':grpc-netty-shaded', configuration: 'shadow')
     signature 'org.codehaus.mojo.signature:java17:1.0@signature'
 }
 
--- a/alts/build.gradle
+++ b/alts/build.gradle
@@ -21,6 +21,7 @@ dependencies {
             libraries.lang,
             libraries.protobuf,
             libraries.conscrypt
+    def nettyDependency = compile project(':grpc-netty')
     compile (libraries.google_auth_oauth2_http) {
         // prefer our own versions instead of google-auth-oauth2-http's dependency
         exclude group: 'com.google.guava', module: 'guava'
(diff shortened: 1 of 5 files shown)


alts/build.gradle
```diff
@@ -29,7 +28,14 @@ dependencies {
         exclude group: 'io.grpc', module: 'grpc-context'
     }
     compileOnly libraries.javax_annotation
-    testCompile project(':grpc-testing'),
+
+    // Use grpc-netty-shaded for transitive dependency.
+    compileOnly project(':grpc-netty')
+    shadow project(path: ':grpc-netty-shaded', configuration: 'shadow')
+    shadow configurations.compile.getDependencies()
```
ejona86: This is doubly suspicious. It is strange to need to copy the dependencies, but wouldn't this also include compile-only dependencies?

dapengzhang0: My understanding is:

 * `DependencyHandler.project(path : 'grpc-alts', configuration: 'shadow')` returns a `Dependency` that includes the output of `:grpc-alts:shadowJar` plus all dependencies added (manually) into `shadow` configuration.

However,

* The output of `:grpc-alts:shadowJar` does not depend on anything in the shadow configuration.

* The output pom.xml of maven publication does not include anything in the shadow configuration unless you modify `pom.withXml {}` manually.

xds/build.gradle
```diff
@@ -47,7 +47,11 @@ dependencies {
             libraries.guava_testlib,
             libraries.netty_epoll
 
+    // Use grpc-netty-shaded for transitive dependency.
+    compileOnly project(':grpc-netty')
     shadow project(path: ':grpc-netty-shaded', configuration: 'shadow')
+    shadow configurations.compile.getDependencies()
```
ejona86: This brings in library.pgv. It needs to be excluded. (maybe with minus() like used in grpc-all?)

I sort of think excluding may be better than using compileOnly for grpc-netty, since then both 'compile' configuration and 'shadow' configuration are functional outside of this project. It also seems to be a bit less likely to have surprises. But I can accept what's been done here.

xds/build.gradle
```diff
@@ -39,11 +38,18 @@ dependencies {
 
     testCompile project(':grpc-core').sourceSets.test.output
 
-    compileOnly libraries.javax_annotation
+    compileOnly libraries.javax_annotation,
+            // At runtime use the epoll included in grpc-netty-shaded
```
ejona86: Technically it would probably be best to do the same `minus()` thing here, but that can be a follow-up.

Issue #1710, opened 2016-04-21
Server.REMOTE_ADDR_KEY and SSL_SESSION_KEY should be checked to see if still experimental.


carl-mastrangelo:
Other languages propagate these via a context, maybe they should be done that way?  In any case, these should be looked at to see if they are part of 1.0.

carl-mastrangelo:
Specific usages:

* `ServerCall.REMOTE_ADDR_KEY`
* `ServerCall.SSL_SESSION_KEY`

louiscryan:
No need to resolve for 1.0, current usage is low.

PR #6805, opened 2020-03-05
Add a New ConfigurableX509TrustManager Class

ZhenLian:
This PR contains the following three classes:

1. TlsOptions.java: an option set by users to specify some advanced features when performing TLS handshake, both on client side and on server side. These features include:

    - choose different levels of peer verification
    - provide custom peer verification check
    - change the trust CA certificate bundle

2. ConfigurableX509TrustManager.java: a class that extends X509ExtendedTrustManager and takes TlsOptions as parameter. Users need to feed ConfigurableX509TrustManager when creating GrpcSslContexts, in order to use the advanced features.

3. AdvancedTlsTest.java: integration tests.


netty/src/main/java/io/grpc/netty/TlsOptions.java
```diff
@@ -0,0 +1,62 @@
...
+    return this.verificationType;
+  }
+
+  // used to perform custom peer authorization checking
+  abstract void verifyPeerCertificate(X509Certificate[] peerCertChain, String authType,
+      SSLEngine engine) throws Exception;
+
+  // used to perform trust CA certificates reloading
+  abstract KeyStore getTrustedCerts() throws Exception;
+}
```
sanjaypujare: missing newline at end of file

netty/src/main/java/io/grpc/netty/TlsOptions.java
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+
+package io.grpc.netty;
+
+import java.security.KeyStore;
+import java.security.cert.X509Certificate;
+import javax.net.ssl.SSLEngine;
+
+// TlsOptions contains different options users could choose. In a nutshell, it provides three main
```
sanjaypujare: How about making these comments Javadoc comments (within /** */) ? Also all the other // comments in this class.

netty/src/main/java/io/grpc/netty/ConfigurableX509TrustManager.java
```diff
@@ -0,0 +1,134 @@
...
+import java.security.KeyStore;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509ExtendedTrustManager;
+
+public class ConfigurableX509TrustManager extends X509ExtendedTrustManager {
```
sanjaypujare: 1. Javadoc comments for  public class? 
2. Can it be marked `final`?

netty/src/main/java/io/grpc/netty/ConfigurableX509TrustManager.java
```diff
@@ -0,0 +1,134 @@
...
+
+  private TlsOptions tlsOptions;
+
+  public ConfigurableX509TrustManager(TlsOptions tlsOptions) {
+    this.tlsOptions = tlsOptions;
```
sanjaypujare: `checkNotNull` for `tlsOptions`?

sanjaypujare:
I can see configurable trust manager being useful but this PR has no gRPC specific code nor netty specific code - but just an implementation of `X509ExtendedTrustManager` that can work for anyone wanting to customize Java's trust manager. Should this code be here?

netty/src/main/java/io/grpc/netty/ConfigurableX509TrustManager.java
ZhenLian: I would expect some other users to further extend this class. In that case, it is better to not make it `final`?

ZhenLian:
@sanjaypujare 
Thanks for bringing this up! I also considered this problem before but this place is the best place I could find.
To my understanding, configurable trust manager is a class dealing with the trust behaviors when performing TLS handshake, and it might be most relevant to the Netty layer, if we put it somewhere in the grpc-java repo.
As far as I know, it might not easy to add this implementation directly into some "standard" Java libraries. Putting in Netty repo might be another potential alternative, and I could ask in their community. 
@ejona86 Could you please make some suggestions on this? Thanks in advance!

sanjaypujare:
There is nothing netty specific in this PR either. I think the best place might be an example in grpc where we include these classes with a concrete example of configurable trust manager.

ZhenLian:
Are you suggesting somewhere [here](https://github.com/grpc/grpc-java/tree/master/examples)? I am OK with the place, as long as users can actually use these classes. 
And when saying putting it in Netty repo, I meant [here](https://github.com/netty/netty/tree/4.1/handler/src/main/java/io/netty/handler/ssl/util). Similar as Netty, We could also have a new util directory somewhere in grpc-java.

sanjaypujare:
I looked at the Netty [link](https://github.com/netty/netty/tree/4.1/handler/src/main/java/io/netty/handler/ssl/util) you posted and it does have util classes so this could fit right in there (and will have more audience than just gRPC).

For [grpc examples](https://github.com/grpc/grpc-java/tree/master/examples) you will need to create a proper example which includes client and server. The added advantage is that you can then show `ConfigurableX509TrustManager` use on both sides.

I talked to @ejona86 offline and he'll add his comments too.

ZhenLian:
@ejona86 I couldn't add you as the reviewer but feel free to comment on this PR. 
@jiangtaoli2016 FYI.

netty/src/main/java/io/grpc/netty/ConfigurableX509TrustManager.java
```diff
@@ -0,0 +1,148 @@
...
+        if (authType == VerificationAuthType.CertificateAndHostNameVerification
+            && (sslEngine == null || sslEngine.getSSLParameters() == null)) {
+          throw new CertificateException(
+              "SSLEngine or SSLParameters is null. Couldn't check host name");
+        }
+        if (sslEngine != null && sslEngine.getSSLParameters() != null) {
+          String algorithm = authType == VerificationAuthType.CertificateAndHostNameVerification
+              ? "HTTPS" : "";
+          SSLParameters sslParams = sslEngine.getSSLParameters();
+          sslParams.setEndpointIdentificationAlgorithm(algorithm);
```
ejona86, line 133: It is surprising this is being set, since these variables are normally set elsewhere are _read_ within the trust manager as configuration. This looks suspicious. It will also permanently change the parameters, which seems concerning. Have you verified that it is safe to disable endpoint identification here and it not impact any other code?

netty/src/main/java/io/grpc/netty/ConfigurableX509TrustManager.java
```diff
@@ -0,0 +1,148 @@
...
+          throw new CertificateException(
+              "Instance delegateX509TrustManager is null. Failed to initialize");
+        }
+      } catch (Exception e) {
+        throw new CertificateException("Failed to initialize delegateX509TrustManager, error: "
+            + e.getMessage());
+      }
+      if (isClient) {
+        if (authType == VerificationAuthType.CertificateAndHostNameVerification
+            && (sslEngine == null || sslEngine.getSSLParameters() == null)) {
```
ejona86, line 125: I don't think `getSSLParameters()` can be `null`.

netty/src/main/java/io/grpc/netty/ConfigurableX509TrustManager.java
```diff
@@ -0,0 +1,148 @@
...
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509ExtendedTrustManager;
+
+/**
+ * ConfigurableX509TrustManager is a highly configurable class that allows users choose different
+ * level of peer checking mechanisms, as well as some customized check. It could also be used to
+ * reload trust certificate bundle client/server uses.
+ */
+public class ConfigurableX509TrustManager extends X509ExtendedTrustManager {
```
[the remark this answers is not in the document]
ZhenLian: Thanks for the suggestions! Shall I also create an issue to keep track of it, like this one: https://github.com/grpc/grpc-java/issues/1710?

netty/src/main/java/io/grpc/netty/TlsOptions.java
```diff
@@ -0,0 +1,84 @@
...
+import javax.net.ssl.SSLEngine;
+
+/**
+ * TlsOptions contains different options users could choose. In a nutshell, it provides three main
+ * features users could customize:
+ * 1. choose different levels of peer verification by specifying |VerificationAuthType|
+ * 2. provide custom peer verification check by inheriting |verifyPeerCertificate|
+ * 3. change the trust CA certificate bundle by inheriting |getTrustedCerts|
+ */
+public abstract class TlsOptions {
```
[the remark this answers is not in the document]
ZhenLian: Sorry, I'd like to discuss a bit more about this here. Making it an interface seems a reasonable common practice, but the implication becomes slightly different. Before, I want users to set a `VerificationAuthType` at the construction time, and this can be expressed by making it a construction function of an abstract class; If using interface, users may need to read the comments to know they need to set `VerificationAuthType`. The method might be more like `setVerificationAuthType` instead of `getVerificationAuthType` if using interface. May I know your opinions?
Also, does `TlsConfig` sound like a better name for you?

netty/src/main/java/io/grpc/netty/ConfigurableX509TrustManager.java
```diff
@@ -0,0 +1,148 @@
...
+
+  @Override
+  public void checkServerTrusted(X509Certificate[] x509Certificates, String s)
+      throws CertificateException {
+
+  }
+
+  @Override
+  public X509Certificate[] getAcceptedIssuers() {
+    return new X509Certificate[0];
```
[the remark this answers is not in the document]
ZhenLian: Implementing it will cause one test to fail, and I think the cause is probably https://github.com/netty/netty/blob/netty-4.1.48.Final/handler/src/main/java/io/netty/handler/ssl/ReferenceCountedOpenSslServerContext.java#L153-L164. It looks like if setting this. Netty internally will use it to do some additional checks, which is not the behavior I want. I might want `checkClientTrusted` and `checkServerTrusted` to cover all the trust behaviors we might have during the authentication.

netty/src/main/java/io/grpc/netty/ConfigurableX509TrustManager.java
ZhenLian: Yes, thanks for the reminding! I have added some integration tests(`basicClientSideIntegrationTest` and `basicServerSideIntegrationTest`), and it looks like this `EndpointIdentificationAlgorithm` behaves as expected, without impacting other parts of the code. Please let me know if you find anything suspicious, thanks!

ZhenLian:
@ejona86 
Hi Eric, thank you so much for your comments! Apologize for the late responses since I was busy with other stuffs. I left a few more comments on your review. Please take a look when you have time and let me know your opinions. 
By the way, this PR keeps failing on the codecov, and from the report it is complaining about the coverage of some override functions(https://codecov.io/gh/grpc/grpc-java/compare/e29561fbca3eb7a536b38d0e0b91c95caef94700...624edfdae8c0feee597cf3399579b798519ff5cf/diff#D2-47...82). I added a unit test(`basicConfigurableX509TrustManagerTest`) and tried to explicit call these uncovered functions, but the codecov is still complaining about it. Locally, it looks like these functions are indeed invoked and can be referenced to the correct function I want them to be through IDE. I have no idea what's going on. Can you provide any guidance on this please?
Thanks again for taking a look into this ;-)

netty/src/main/java/io/grpc/netty/TlsOptions.java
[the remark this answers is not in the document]
ejona86: I was not suggesting to make it an interface; keep it as an abstract class. We use abstract classes so we can add methods to them later.

No, even if the user is implementing it, `getVerificationAuthType()` is a proper name for it. I don't think that would cause a problem for Java programmers, especially those that would be implementing this.

The other option is to guarantee that the verification type does not change over time and make `getVerificationAuthType()` final. The problem now is that it can be overridden but it is also required for the constructor.

netty/src/main/java/io/grpc/netty/ConfigurableX509TrustManager.java
```diff
@@ -0,0 +1,149 @@
...
+              "Instance delegateX509TrustManager is null. Failed to initialize");
+        }
+      } catch (Exception e) {
+        throw new CertificateException("Failed to initialize delegateX509TrustManager", e);
+      }
+      if (checkingServer) {
+        if (authType == VerificationAuthType.CertificateAndHostNameVerification
+            && sslEngine == null) {
+          throw new CertificateException(
+              "SSLEngine or SSLParameters is null. Couldn't check host name");
```
ejona86, line 128: This error message is no longer accurate.

netty/src/main/java/io/grpc/netty/ConfigurableX509TrustManager.java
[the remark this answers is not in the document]
ejona86: Hmm... That's sort of worrisome. Go ahead and put a comment here saying why it is empty.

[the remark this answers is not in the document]
ejona86: Actually... this looks like it is only for certificate chain verification. What does it break?

netty/src/main/java/io/grpc/netty/ConfigurableX509TrustManager.java
ejona86: "I tested it once" isn't what I really meant by verify. But since this on the engine, it is probably okay.

netty/src/main/java/io/grpc/netty/TlsOptions.java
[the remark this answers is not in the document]
ZhenLian: SG. Updated accordingly. 
Regarding the name of this class, would `TlsConfig` sound better?

netty/src/main/java/io/grpc/netty/ConfigurableX509TrustManager.java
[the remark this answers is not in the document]
ZhenLian: I've summarized my findings in this Gist: https://gist.github.com/ZhenLian/50c73f146493ac2fe9746d196082679b.
If I changed `getAcceptedIssuers` to the one named `change.java` in my Gist, one of the tests I wrote would fail, with the error message named `failure.log`.
That is actually an expected behavior under my current implementation. The failing test is basically to test under the scenario that server chooses to `SkipAllVerification`, and client sends a bad certificate, in mTLS. This should be OK because server chooses to skip all verification(in real situations, we want application to provide additional custom checks, but since this is test, we just test the simple behavior). But this test won't pass if we have additional check in `getAcceptedIssuers`.

ZhenLian:
I've updated the tests and now the coverage issue is gone. I've also added some more comments since your last review @ejona86  Could you please take a look when you have time please? Thanks!

ZhenLian:
@ejona86  Friendly ping :-)

ejona86:
Is #8175 intended to replace this PR?

ZhenLian:
@ejona86 
Ah yes, let me just close this PR in favor of #8175

PR #6809, opened 2020-03-06
xds: parse RDS response to ConfigUpdate for path matching

dapengzhang0:


xds/src/main/java/io/grpc/xds/XdsClientImpl.java
```diff
@@ -624,18 +632,26 @@ private void handleLdsResponseForConfigUpdate(DiscoveryResponse ldsResponse) {
     adsStream.sendAckRequest(ADS_TYPE_URL_LDS, ImmutableList.of(ldsResourceName),
         ldsResponse.getVersionInfo());
 
-    if (clusterName != null || rdsRouteConfigName != null) {
+    if (routes != null || rdsRouteConfigName != null) {
```
voidzcy: `routes` would never be `null`. Ditto below.

xds/src/main/java/io/grpc/xds/XdsClientImpl.java
```diff
@@ -842,24 +869,98 @@ static String findClusterNameInRouteConfig(RouteConfiguration config, String hos
...
+      }
+    }
+    return routes;
+  }
+
+  /**
+   * Validates the given list of routes and return error details if there's any error.
+   */
+  // We only validate the default route unless path matching is enabled.
+  // We do more validation if path matching is enabled, but whether every single route should be
```
voidzcy: nit: the sentence has grammatical error, rephrase.

xds/src/main/java/io/grpc/xds/XdsClientImpl.java
```diff
@@ -842,24 +869,98 @@ static String findClusterNameInRouteConfig(RouteConfiguration config, String hos
...
+      List<Route> routesProto = targetVirtualHost.getRoutesList();
+      for (Route route : routesProto) {
+        routes.add(EnvoyProtoData.Route.fromEnvoyProtoRoute(route));
+      }
+    }
+    return routes;
+  }
+
+  /**
+   * Validates the given list of routes and return error details if there's any error.
```
voidzcy: nit: returns

xds/src/main/java/io/grpc/xds/XdsClientImpl.java
```diff
@@ -842,24 +869,98 @@ static String findClusterNameInRouteConfig(RouteConfiguration config, String hos
...
+      EnvoyProtoData.Route route = routes.get(i);
+
+      if (!route.routeAction.isPresent()) {
+        return "Route action is not specified for one of the routes";
+      }
+
+      RouteMatch routeMatch = route.routeMatch;
+      String prefix = routeMatch.prefix;
+      String path = routeMatch.path;
+      // Whether some of the errors can be skipped is TBD.
```
voidzcy: nit: if it is something TBD. Mark it as `TODO`. Also I would suggest rephrasing the sentence .

xds/src/main/java/io/grpc/xds/XdsClientImpl.java
```diff
@@ -786,25 +804,34 @@ private void handleRdsResponse(DiscoveryResponse rdsResponse) {
...
-      ConfigUpdate configUpdate = ConfigUpdate.newBuilder().setClusterName(clusterName).build();
+
+      // Found clusterName in the in-lined RouteConfiguration.
+      String clusterName = routes.get(routes.size() - 1).routeAction.get().cluster;
+      if (!enablePathMatching) {
+        logger.log(XdsLogLevel.INFO, "Found cluster name: {0}", clusterName);
+      } else {
+        logger.log(
+            XdsLogLevel.INFO,
+            "Found routes: {0}", routes);
```
voidzcy: This INFO log will print a list of complex structures, which is not appropriate for INFO level logs. Use INFO level logging to print concise descriptions for those routes or just print out the number of routes found. INFO level log messages are intended to be concise and informative.

xds/src/main/java/io/grpc/xds/XdsClientImpl.java
```diff
@@ -786,25 +804,34 @@ private void handleRdsResponse(DiscoveryResponse rdsResponse) {
...
         rdsRespTimer.cancel();
         rdsRespTimer = null;
       }
-      logger.log(XdsLogLevel.INFO, "Found cluster name: {0}", clusterName);
-      ConfigUpdate configUpdate = ConfigUpdate.newBuilder().setClusterName(clusterName).build();
+
+      // Found clusterName in the in-lined RouteConfiguration.
+      String clusterName = routes.get(routes.size() - 1).routeAction.get().cluster;
+      if (!enablePathMatching) {
+        logger.log(XdsLogLevel.INFO, "Found cluster name: {0}", clusterName);
```
voidzcy: cluster name is always queried, why only logged when path matching is enabled?

xds/src/main/java/io/grpc/xds/XdsClientImpl.java
```diff
@@ -767,16 +783,18 @@ private void handleRdsResponse(DiscoveryResponse rdsResponse) {
...
+      routes = findRoutesInRouteConfig(requestedRouteConfig, ldsResourceName);
+      String errorDetail = validateRoutes(routes);
+      if (errorDetail != null) {
         adsStream.sendNackRequest(
             ADS_TYPE_URL_RDS, ImmutableList.of(adsStream.rdsResourceName),
             rdsResponse.getVersionInfo(),
             "RouteConfiguration " + requestedRouteConfig.getName() + ": cannot find a "
                 + "valid cluster name in any virtual hosts with domains matching: "
-                + ldsResourceName);
+                + ldsResourceName
```
voidzcy: Would to rewrite the entire error message instead of just appending an error detail. Also, this entire message becomes the `error_details` field in the message sent to TD, you should avoid having a nested "Error detail:".

xds/src/main/java/io/grpc/xds/XdsClientImpl.java
dapengzhang0: `routes` initially is null. It's assigned conditionally. Why would route never be null?

xds/src/main/java/io/grpc/xds/XdsClientImpl.java
dapengzhang0: cluster name is logged when path matching is _NOT_ enabled, which keeps the existing behavior.

When path matching is enabled, it logs the whole list of routes. In this case, the default route (last in the list) may not be a cluster, it could be weighted clusters.

xds/src/main/java/io/grpc/xds/XdsClientImpl.java
dapengzhang0: Changed to DEBUG level. INFO level only logs count now.

xds/src/main/java/io/grpc/xds/XdsClientImpl.java
dapengzhang0: Rephrased.

xds/src/main/java/io/grpc/xds/EnvoyProtoData.java
```diff
@@ -339,4 +340,194 @@ public String toString() {
           .toString();
     }
   }
+
+  /** See corresponding Envoy proto message {@link io.envoyproxy.envoy.api.v2.route.Route}. */
+  static final class Route {
+    final RouteMatch routeMatch;
+    final Optional<RouteAction> routeAction;
```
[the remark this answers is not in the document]
dapengzhang0: The benefit: 
https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained#whats-the-point

That might be opinionated as well as the SO.

xds/src/main/java/io/grpc/xds/EnvoyProtoData.java
```diff
@@ -339,4 +340,194 @@ public String toString() {
           .toString();
     }
   }
+
+  /** See corresponding Envoy proto message {@link io.envoyproxy.envoy.api.v2.route.Route}. */
+  static final class Route {
+    final RouteMatch routeMatch;
```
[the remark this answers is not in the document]
dapengzhang0: This is just style preference. I'll do so for consistency in this class.

xds/src/main/java/io/grpc/xds/EnvoyProtoData.java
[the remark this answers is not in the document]
voidzcy: According to its [Javadoc](https://guava.dev/releases/19.0/api/docs/com/google/common/base/Optional.html), the most common usages for `Optional` are:
- __As a method return type__, as an alternative to returning null to indicate that no value was available
- To __distinguish between "unknown" (for example, not present in a map) and "known to have no value" (present in the map, with value Optional.absent())__
- To __wrap nullable references for storage in a collection that does not support null__ (though there are several other approaches to this that should be considered first)

It's not for usage of avoiding nullness just for convenience. This usage is completely an abuse, even your IDE will warn you "Optional used as type for field". For details, read the first item of [Optional Anti-Patterns](https://dzone.com/articles/optional-anti-patterns).

xds/src/main/java/io/grpc/xds/XdsClientImpl.java
voidzcy: You are right. It's conditionally assigned.

xds/src/main/java/io/grpc/xds/XdsClientImpl.java
```diff
@@ -842,24 +868,99 @@ static String findClusterNameInRouteConfig(RouteConfiguration config, String hos
...
+        routes.add(EnvoyProtoData.Route.fromEnvoyProtoRoute(route));
+      }
+    }
+    return routes;
+  }
+
+  /**
+   * Validates the given list of routes and returns error details if there's any error.
+   */
+  // We only validate the default route unless path matching is enabled.
```
voidzcy: nit: be formal. If it is part of method specification, put it inside the head comment of this method. If it is about implementation details, put it at where the code logic is present. If it is your personal note, should just delete it.

xds/src/main/java/io/grpc/xds/EnvoyProtoData.java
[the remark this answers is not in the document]
dapengzhang0: For the IDE waring and Optional Anti-Patterns, it's only a problem when you want to serialize the class. The usage here is the first one in the javadoc, though the field can be just nullable to be exactly the case.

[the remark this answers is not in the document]
voidzcy: "Anti-pattern" indicates it's a rule-of-thumb convention that you should avoid, even though it may not cause problem now. It's a bad practice to do that and it's more likely to cause problem in the future. I do not see any benefit than a nullable field (which is more elegant and efficient). Why using `Optional` as a class field type is the first category of usage (i.e., as return type)? The first category of usage is referring to the following pattern:

```java
class Foo {
   // nullable
   private Bar bar;

  Optional<Bar> getBar() {
     // ...
  }
}
```

xds/src/main/java/io/grpc/xds/EnvoyProtoData.java
```diff
@@ -339,4 +341,239 @@ public String toString() {
...
+    Route(RouteMatch routeMatch, @Nullable RouteAction routeAction) {
+      this.routeMatch = routeMatch;
+      this.routeAction = routeAction;
+    }
+
+    RouteMatch getRouteMatch() {
+      return routeMatch;
+    }
+
+    Optional<RouteAction> getRouteAction() {
```
voidzcy: I would just prefer using nullable return value and all the other places used that (it doesn't give us much benefit for using `Optional`). But I am fine with this.

xds/src/main/java/io/grpc/xds/XdsClientImpl.java
voidzcy: But cluster name (`routes.get(routes.size() - 1).getRouteAction().get().getCluster()`) is always queried and `clusterName` in `ConfigUpdate` is always set. Is that information used when path matching is enabled? If it is, it should be logged.

dapengzhang0: When path matching is enabled:
In the weighted cluster case (), `getCluster()` returns an empty string, it does not make sense to log the empty string. The xds-routing lb will only use the list of routes in configUpdate.

When path matching is not enabled (existing behavior):
The `getCluster()` of the default (last) route should always return a valid cluster name, in that case we log it. The cds lb will only use the clusterName in configUpdate.

PR #6816, opened 2020-03-10
buildscripts: update xds test server arg

ericgribkoff:
Please excuse the numerous commits: the test results on `d1ccbb1` are useful as the Android build log shows a successful run of the gRPC xDS job (probably should figure out a better way to test this on PRs, but for now the job only runs on master so this hack of borrowing another Kokoro job seems to work for PR testing).

This depends on https://github.com/grpc/grpc/pull/22297.

I also removed the Python package installation bits from our `xds.sh` script as all prereqs are installed by the invocation of `prep_xds.sh` from the main repo.


buildscripts/kokoro/xds.sh
```diff
@@ -22,4 +19,4 @@ python3 grpc/tools/run_tests/run_xds_tests.py \
     --project_id=grpc-testing \
     --gcp_suffix=$(date '+%s') \
     --verbose \
-    --client_cmd='grpc-java/interop-testing/build/install/grpc-interop-testing/bin/xds-test-client --server=xds-experimental:///{service_host}:{service_port} --stats_port={stats_port} --qps={qps}'
+    --client_cmd='grpc-java/interop-testing/build/install/grpc-interop-testing/bin/xds-test-client --server=xds-experimental:///{server_uri} --stats_port={stats_port} --qps={qps}'
```
dapengzhang0: nit: break the line at `--server=` etc

ericgribkoff: Done

dapengzhang0: Oh, I didn't notice that it's inside a quote. Then it's not necessary. My bad.

ericgribkoff: Still looks nicer with the line breaks :)

PR #6831, opened 2020-03-13
grpclb: migrate create subchannel (minimal change)

creamsoup:


grpclb/src/main/java/io/grpc/grpclb/CachedSubchannelPool.java
```diff
@@ -34,32 +35,46 @@
  * A {@link SubchannelPool} that keeps returned {@link Subchannel}s for a given time before it's
  * shut down by the pool.
  */
+// TODO(creamsoup) address to subchannel is not 1:1 mapping because subchannel can update its
+//  address. Use Multimap and ForwardingSubchannel (override updateAddresses) to maintain most up to
```
ejona86: There should be no need for a multimap. When a subchannel is added to the pool it should no longer be used, which means updateAddresses shouldn't be called. When a subchannel is added this code can just call `subchannel.getAddresses()`/`getAllAddresses()` to add the subchannel to the `cache` map.

creamsoup: i think it is a bit weird that the pool has no idea about the subchannels it created. i agree that the updateAddresses shouldn't be called externally, but nothing prevent it from doing that. And it seems like blocking it by throwing exception is not an option due to the hierarchical lb policy.
My point is the pool should know the status of subchannels it owns (ownership of the subchannel should be on the pool not the caller). currently, user of the pool can return a subchannel created elsewhere, and the pool won't complain. also, `clear` just shutdowns already returned subchannels and doesn't care about subchannels currently used.
if it was a SubchannelFactory it might look okay.

ejona86: It is very normal for pools dealing with exclusive objects have no tracking of the objects they hand out. "Exclusive" meaning that there can only be one user of the object at a time, like DB connections and the like. If the pool does keep track of the objects handed out, it is generally for resource limiting reasons (like a DB pool maxing out at 5 connections).

> And it seems like blocking it by throwing exception is not an option due to the hierarchical lb policy.

I don't understand this. The pool could wrap the returned subchannels and override getAddresses(). When the subchannel is returned to the pool the pool sets a `throwOnAddresses = true` variable in the subchannel that `getAddresses()` checks. When the subchannel is taken from the pool, `throwOnAddresses = false` is set before the pool returns the subchannel.

> My point is the pool should know the status of subchannels it owns (ownership of the subchannel should be on the pool not the caller)

I don't think so. And it clearly makes things harder. What benefit is there to having the pool track the subchannels it has handed out? If there is no benefit, why wouldn't we do it "the easy way?"

> currently, user of the pool can return a subchannel created elsewhere, and the pool won't complain

Fair. Although that is technically true of a couple different things related to subchannels (like which ones are returned from picks). This would be trivial to check in the future because we are guaranteed to wrap subchannel to handle `shutdown()`, so we could do an `instanceof` or (if we are really paranoid) a nested class instance check (`CachedSubchannelPool.this` within the subchannel).

> also, clear just shutdowns already returned subchannels and doesn't care about subchannels currently used.

That sounds perfect. Policies should shutdown/return the subchannels they created.

creamsoup: okay thanks for the explanation. i think i went side ways by trying to get rid of the handleSubchannelState.

grpclb/src/main/java/io/grpc/grpclb/CachedSubchannelPool.java
```diff
@@ -38,28 +39,39 @@
...
+      subchannel =
+          helper.createSubchannel(
+              CreateSubchannelArgs.newBuilder()
+                  .setAddresses(eag)
+                  .setAttributes(defaultAttributes)
+                  .build());
+      subchannel.start(new SubchannelStateListener() {
+        @Override
+        public void onSubchannelState(ConnectivityStateInfo newState) {
+          listener.onSubchannelState(subchannel, newState);
```
dapengzhang0: It seems you make `CachedSubchannelPool#handleSubchannelState()` private  and call  `handleSubchannelState()` here directly. Then you don't need the API `SubchannelPool#handleSubchannelState(Subchannel subchannel, ConnectivityStateInfo newStateInfo)` in the interface anymore.  Otherwise, `PooledSubchannelStateListener#onSubchannelState()` calls `GrpclbState#handleSubchannelState`, which calls `CachedSubchannelPool#handleSubchannelState` back and forth.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -68,19 +68,10 @@
     this.stopwatch = checkNotNull(stopwatch, "stopwatch");
     this.backoffPolicyProvider = checkNotNull(backoffPolicyProvider, "backoffPolicyProvider");
     this.subchannelPool = checkNotNull(subchannelPool, "subchannelPool");
-    this.subchannelPool.init(helper, this);
```
dapengzhang0: Because we removed `init`, we don't need the `subchannelPool` field in this class. We can make `subchannelPool` a final field of `GrpclbState`.

creamsoup: good catch, i didn't do it because it makes very hard to convert existing tests because grpclb tests mocks the subchannelpool and passed it through `GrpclbLoadBalancer`. alternatively, I could create a provider or factory class which i thought little too much, wdyt?

grpclb/src/main/java/io/grpc/grpclb/CachedSubchannelPool.java
creamsoup: i am little bit confused. first method mean `Grpclb#handleSubchannelState(Subchannel, ConnectivityStateInfo)`?

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
```diff
@@ -54,8 +53,9 @@
   private GrpclbConfig config = DEFAULT_CONFIG;
 
   // All mutable states in this class are mutated ONLY from Channel Executor
+  @VisibleForTesting
```
dapengzhang0: There is a getter visible for testing already.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
dapengzhang0: The current test approach seems terribly wrong:
```java
      doAnswer(new Answer<Subchannel>() {
        @Override
        public Subchannel answer(InvocationOnMock invocation) throws Throwable {
          Subchannel subchannel = mock(Subchannel.class);
          EquivalentAddressGroup eag = (EquivalentAddressGroup) invocation.getArguments()[0];
          Attributes attrs = (Attributes) invocation.getArguments()[1];
          when(subchannel.getAllAddresses()).thenReturn(Arrays.asList(eag));
          when(subchannel.getAttributes()).thenReturn(attrs);
          mockSubchannels.add(subchannel);
          pooledSubchannelTracker.add(subchannel);
          return subchannel;
        }
      }).when(subchannelPool).takeOrCreateSubchannel(
          any(EquivalentAddressGroup.class), any(Attributes.class));
```

I think a more correct way to test is not mocking subchannelPool, but just mocking `helper.createSubchannel()` and keeping track of the EAGs and subchannels. But then a lot of current tests need be rewritten. 

I'm OK with using a factory in the short term if it's too hard to fix tests, but it worth rewriting some of the tests in the future.

grpclb/src/main/java/io/grpc/grpclb/CachedSubchannelPool.java
dapengzhang0: I mean you can make `CachedSubchannelPool#handleSubchannelState(Subchannel subchannel, ConnectivityStateInfo newStateInfo)` a private method. Remove this method from interface. And call this private method directly here.

```java
listener.onSubchannelState(subchannel, newState);
handleSubchannelState(subchannel, newState);
```

Do not call this private method from the other class `GrpclbState`.

grpclb/src/main/java/io/grpc/grpclb/CachedSubchannelPool.java
```diff
@@ -87,8 +88,7 @@ public void run() {
     return subchannel;
```
dapengzhang0: Do you need to call `handleSubchannelState()` in `run()` as well?

grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
```diff
@@ -196,10 +196,7 @@ void handleSubchannelState(Subchannel subchannel, ConnectivityStateInfo newState
     if (newState.getState() == SHUTDOWN) {
       return;
     }
-    if (!subchannels.values().contains(subchannel)) {
-      if (subchannelPool != null ) {
-        subchannelPool.handleSubchannelState(subchannel, newState);
-      }
+    if (subchannelPool == null && !subchannels.values().contains(subchannel)) {
```
dapengzhang0: Why need check `subchannelPool == null`?

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
creamsoup: oops done.

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
creamsoup: i do agree that current test is not really testing anything. at best, it is testing it's own implementation via mock. fixing this (by making fake classes or using real classes instead of mock) is essentially rewriting the entire test.
i'll use the factory approach, i am not sure how much time we want to invest to fix the test though given the state of the grpclb.

grpclb/src/main/java/io/grpc/grpclb/CachedSubchannelPool.java
creamsoup: we don't need to do that because the state remains same.

grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
creamsoup: removed, thanks!

grpclb/src/main/java/io/grpc/grpclb/CachedSubchannelPool.java
creamsoup: oh yes. done PTAL

grpclb/src/main/java/io/grpc/grpclb/GrpclbLoadBalancer.java
creamsoup: actually i like the factory better, because the pool is ignored/unused when it is using pick_first. using factory prevents it to create unnecessary pool. sounds like win-win.

grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
```diff
@@ -453,9 +460,18 @@ private void useRoundRobinLists(
...
+          subchannel =
+              helper.createSubchannel(
+                  CreateSubchannelArgs.newBuilder()
+                      .setAddresses(eagList)
+                      .setAttributes(createSubchannelAttrs())
+                      .build());
+          subchannel.start(new SubchannelStateListener() {
+            @Override
+            public void onSubchannelState(ConnectivityStateInfo newState) {
+              // No-op
```
dapengzhang0: I think still need to call `GrpclbState.handleSubchannelState()` otherwise it's never called for pick_first. And the test calling `GrpclbState.handleSubchannelState()` directly for pick_first seems cheating. It might require capturing/storing the listeners and delivering subchannel state with the listeners, could be messy.

grpclb/src/main/java/io/grpc/grpclb/CachedSubchannelPool.java
dapengzhang0: Then we don't even need the private method, we can just inline it here
```java
CacheEntry cached = cache.get(subchannel.getAddresses());
if (cached != null && cached.subchannel != subchannel) {
  cached.state = newState;
}
```

creamsoup: i don't see why having a small method is bad. even though it is short, method name is helpful to read the code.

grpclb/src/main/java/io/grpc/grpclb/GrpclbState.java
creamsoup: okay. i use that shortcut because it requires non trivial changes. converted mocked classes to fake/real classes, PTAL.

grpclb/src/main/java/io/grpc/grpclb/CachedSubchannelPool.java
creamsoup: okay. I believe smaller method is almost always preferred, so problem here is the method name. how about `updateCachedSubchannelState`?

PR #6837, opened 2020-03-17
xds: generate xds-routing config from XdsNameResolver

dapengzhang0:


xds/src/main/java/io/grpc/xds/XdsNameResolver.java
```diff
@@ -131,23 +137,129 @@ XdsClient createXdsClient() {
...
+
+    ConfigWatcherImpl(Listener2 listener) {
+      this.listener = listener;
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public void onConfigChanged(ConfigUpdate update) {
+      Map<String, ?> config;
+      if (update.getRoutes().size() > 1) {
```
voidzcy: The conditions of this if-elseif-else is verbose and unclear. Use `null` value of fields in `ConfigUpdate` to make this cleaner:

```java
if (update.clusterName != null) {
   // CDS
} else if (update.getRoutes.size() == 1) {
   // weighted-target
} else {
   // xds-routing
}
```

xds/src/main/java/io/grpc/xds/XdsNameResolver.java
```diff
@@ -131,23 +137,129 @@ XdsClient createXdsClient() {
...
+            XdsLogLevel.INFO,
+            "Received config update with {0} routes from xDS client {1}",
+            update.getRoutes().size(),
+            xdsClient);
+        logger.log(
+            XdsLogLevel.DEBUG,
+            "Received config update from xDS client {0}: {1}",
+            xdsClient,
+            update);
+        List<Object> routes = new ArrayList<>(update.getRoutes().size());
```
voidzcy: This whole logic of generating LB configs is complicated, given that each LB config can have recursively nested child policies. Same for config generation in EDS policy (TODO).

Previously generating service config with CDS LB config very simple (a raw JSON string, only ~10 lines). But now, it's bad to write all of this here (config generation in EDS policy is probably to be even more complicated). This is unreadable, with three levels of child policy nesting (maps with values of maps with values of maps). We would need to seek for a better approach:

Similar to how LB config is parsed (each LB config parser recursively calls child policy's own parser to parse child policy), we can implement a LB config generator for each LB policy so that each config generator generates its own config (recursively use child policy's generator to generate child policies) and put the output together.

dapengzhang0:
I changed the `ConfigUpdate` data,  it has only `routes` field now. When path matching is not enabled, xdsClient will generate an configUpdate with only a single cluster route, the default route.

xds/src/main/java/io/grpc/xds/XdsNameResolver.java
dapengzhang0: Reorganized code. Seems a little better.

xds/src/main/java/io/grpc/xds/XdsClient.java
```diff
@@ -17,13 +17,15 @@
...
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 // TODO(sanjaypujare): remove dependency on envoy data types.
+import com.google.common.collect.Iterables;
```
voidzcy, line 28: nit: this line should be put one line above.

xds/src/main/java/io/grpc/xds/XdsClient.java
```diff
@@ -63,19 +65,18 @@
...
-    private ConfigUpdate(String clusterName, List<Route> routes) {
-      this.clusterName = clusterName;
+    private ConfigUpdate(List<Route> routes) {
       this.routes = routes;
     }
 
+    /** Gets the cluster name for the default route. */
     String getClusterName() {
-      return clusterName;
+      return Iterables.getLast(routes).getRouteAction().getCluster();
```
voidzcy, line 76: Ah... I was wondering how existing test cases in `XdsClientImplTest` are not affected, this is why. `ConfigUpdate` intends to be POJO, it's a data class. Its getter should not involve any business logic, it's just for getting a property value. This method doesn't make sense when `routes` contains more than one items and it should not be called in that case.

So I would still leaning towards `ConfigUpdate` containing a String and a List, with "oneof" semantics. XdsClient will set `clusterName` (String) if path matching is not enabled and leave `routes` (List) being `null`, and vice versa for path matching enabled.

dapengzhang0: I agree that getter should not involve business logic, so I removed `getClusterName()`. It was just  used as a helper method, mostly for making test easy. I'm not a fan of oneof with two nullalbe fields either, because it will add more awkward business logic for `XdsNameResolver`.  Instead, I define a helper method in the test.

voidzcy: The logic for not enabling path matching is wired, as you can see in XdsClientImpl, it goes into `Route` in `VirtualHost` to populate `clusterName` but it puts `Route` into `ConfigUpdate` and the resolver goes into the `Route` to populate `clusterName` again.

dapengzhang0: > as you can see in XdsClientImpl, it goes into Route in VirtualHost to populate clusterName but it puts Route into ConfigUpdate and the resolver goes into the Route to populate clusterName again.

The XdsClientImple goes into Route in VirtualHost NOT to populate clusterName, it populates the whole route. (Yeah, it reads clusterName, but that's only for validation and INFO level logging.)

```java
    if (!enablePathMatching) {
        EnvoyProtoData.Route defaultRoute = Iterables.getLast(routes);
        configUpdate =
            ConfigUpdate.newBuilder()
                .addRoutes(ImmutableList.of(defaultRoute))
                .build();
        logger.log(
            XdsLogLevel.INFO,
            "Found cluster name (inlined in route config): {0}",
            defaultRoute.getRouteAction().getCluster());
      }
```

dapengzhang0: The con of this approach is:
- For path matching not enabled case, `XdsNameResolver` needs to dig one level deeper to get `clusterName`.

The con of oneof approach is:
- The data structure of `ConfigUpdate` is ugly, and it is so only for the purpose of a usecase that might be migrated away.
- `XdsNameResolver` need to check null for the oneof fields.

voidzcy: You have to populate `Route` message when path matching is disabled. It needs to validate

> whose match field must contain a prefix field whose value is the empty
string and whose route field must be set.

https://github.com/markdroth/proposal/commit/242375246285666fff622a60a988822c2466c09a#diff-05ea4a5894abbc0261b006741220598cR342-R344

This piece of logic is deleted in your last change, you really need to. revert it back.


Enabling/disabling patch matching runs in separate code paths, having them receiving different content of `ConfigUpdate` is a desired approach. IMO, using "oneof" semantics seems to be the desired way.

voidzcy: > The data structure of `ConfigUpdate` is ugly, and it is so only for the purpose of a usecase that might be migrated away.

It's not "ugly". The two code paths use different fields in `ConfigUpdate`, separate code uses separate data, that's elegant. When eliminating one code path, just delete the field it uses.

> * `XdsNameResolver` need to check null for the oneof fields.

That's the beauty of null here. Elegant and concise meaning of something unset and do not use.

dapengzhang0: > This piece of logic is deleted in your last change, you really need to. revert it back.

I still have the logic in `validateRoutes()`.

voidzcy: I missed that one. But, still, I think using "oneof" different code paths makes implementation cleaner.

dapengzhang0: I think we don't have to over-engineer for something that's temporary: the code path split for the enable path matching flag.
For me the code path split only in XdsClientImpl is easiest approach.

xds/src/main/java/io/grpc/xds/XdsNameResolver.java
```diff
@@ -131,62 +136,160 @@ XdsClient createXdsClient() {
...
-        //  a temporary solution. More design discussion needs to be done.
-        if (error.getCode().equals(Code.NOT_FOUND)) {
-          logger.log(
-              XdsLogLevel.WARNING,
-              "Received error from xDS client {0}: {1}", xdsClient, error.getDescription());
-          listener.onResult(ResolutionResult.newBuilder().build());
-          return;
+      Map<String, ?> serviceConfig =
+          ImmutableMap.of("loadBalancingConfig", ImmutableList.of(rawLbConfig));
+      logger.log(XdsLogLevel.INFO, "Generated service config:\n{0}", serviceConfig);
```
voidzcy, line 183: This message would be almost unreadable, as `serviceConfig` is a deeply nested map. We may need a utility that coverts `Map<String, ?>` to `String`, but seems would only be useful for visualization. It's nontrivial to implement that. 

So maybe, we have to delete this log message (or change it to `DEBUG`?) to avoid printing out tons of hard-to-read data.

xds/src/main/java/io/grpc/xds/XdsClient.java
voidzcy: Err... I don't think that's over-engineering, it's about implementation integrity and code style. Alright, I failed to convince you. I am not going to block you on this.

xds/src/main/java/io/grpc/xds/XdsNameResolver.java
dapengzhang0: Service config is in general a very large, as it contains not only lb config.  I think service config is important logging information and it might worth INFO level. But we need to de-dup ConfigUpdate to avoid logging spam.

voidzcy: I mean, printing out the `Map<String, ?>` is bad, it's unreadable when it's large. If we have utility to convert `Map<String, ?>` to a nicely formatted JSON string, that would be the best. But, we don't have such a utility.

dapengzhang0: Use gson reverse parser now.

xds/src/test/java/io/grpc/xds/XdsNameResolverTest.java
```diff
@@ -175,6 +185,7 @@ public BootstrapInfo readBootstrap() {
   @After
   public void tearDown() {
     xdsNameResolver.shutdown();
+    XdsClientImpl.enablePathMatching = false;
```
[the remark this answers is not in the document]
dapengzhang0: Reset in test method may not be executed if the test method fails unless the reset is in `finally{}` after surrounding the whole test with `try {}`.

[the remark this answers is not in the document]
voidzcy: Setting it back to `false` is completely redundant.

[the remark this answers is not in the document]
dapengzhang0: Alright. But when I remove the reset in this and the other test class, some tests just fail.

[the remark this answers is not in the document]
voidzcy: Ah, this is static, so test methods do affect each other... Alright, leave it as is.

xds/src/test/java/io/grpc/xds/XdsNameResolverTest.java
```diff
@@ -330,6 +341,182 @@ public void resolve_resourceUpdated() {
...
+    assertThat(action.keySet()).containsExactly("childPolicy");
+    Map<String, ?> lbConfig =
+        Iterables.getOnlyElement((List<Map<String, ?>>) action.get("childPolicy"));
+    assertThat(lbConfig.keySet()).containsExactly(CDS_POLICY_NAME);
+    Map<String, ?> rawConfigValues = (Map<String, ?>) lbConfig.get(CDS_POLICY_NAME);
+    assertThat(rawConfigValues).containsExactly("cluster", clusterName);
+  }
+
+  @SuppressWarnings("unchecked")
+  private static void assertRouteActionIsWeightedTargetPolicy(
```
voidzcy, line 492: nit: same for the change above. Name this to `assertWeightedTargetPolicy`.

dapengzhang0:
The problem is I don't agree the most of these. The oneof approach spreads the two code paths across `XdsClientImpl`, `ConfigUdate` and `XdsNameResolver`, if you were adding a new orthogonal flag, you would have four code paths across each of these three classes, that's very messy, not elegant. The current approach constraints the two code paths within `XdsClientImpl`, and it's easier to migrate and change. I'm not hoping my argument can convince you, but we do have different opinions.

You seem like tending to impose some opinionated dogmatic coding preferences that's not well-accepted/applied in google or necessary to other people. That's very counter-productive. The experience of  XdsLoadBalancer, EdsLoadBalancer, FallbackL is whenever I tried to make a change, a lot of them are imposed to me, and a lot of time was spent to argue about or apply them, but those historical changes are totally gone today, we could have done the changes sooner. Similarly LoadReportClient was also making a lot of changes, and those historical changes, no matter how perfect to you or imperfect to me, are gone today. 

Unlike other projects, for the XDS project, productivity is more important than "perfect" code style.

>It's not "ugly". The two code paths use different fields in ConfigUpdate, separate code uses separate data, that's elegant. When eliminating one code path, just delete the field it uses. 

>Due to my recent experience making changes in XdsLoadBalancer, EdsLoadBalancer, FallbackLb, and their tests etc, I am gonna be stricter on code style and tests. 😢

>Again, it rises my concern about ConfigUpdate's usage for two code paths. I am worrying things like this are gonna spread and we become addictive to easy but harmful (to code health/readability/style) helpers such as assertConfigUpdateContainsSingleClusterRoute.

PR #6838, opened 2020-03-18
xds: integration of XdsClientImpl with XdsServerBuilder to deliver Listener updates

sanjaypujare:


sanjaypujare:
@voidzcy I am assuming you are okay to review this or I could ask @creamsoup ....

voidzcy:
> @voidzcy I am assuming you are okay to review this or I could ask @creamsoup ....

I can look at the code and see if the workflow is reasonable. But I do not have the entire picture for the design. So it would be great if both of us take a look.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,299 @@
...
+                serverList,
+                XdsClient.XdsChannelFactory.getInstance(),
+                node,
+                syncContext,
+                timeService,
+                new ExponentialBackoffPolicy.Provider(),
+                GrpcUtil.STOPWATCH_SUPPLIER);
+          }
+        };
+    xdsClientPool = new XdsClient.RefCountedXdsClientObjectPool(xdsClientFactory);
```
voidzcy: Why you use an object pool? You are just creating an `XdsClient` instance in-place without sharing with other classes.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,299 @@
...
+ * io.grpc.xds.internal.sds.XdsServerBuilder}.
+ */
+@Internal
+public final class XdsClientWrapperForServerSds {
+  private static final Logger logger =
+      Logger.getLogger(XdsClientWrapperForServerSds.class.getName());
+
+  private static final EventLoopGroupResource eventLoopGroupResource =
+      Epoll.isAvailable() ? new EventLoopGroupResource("GrpcServerXdsClient") : null;
+  private final InternalLogId logId;
```
voidzcy: I don't see any value of using a logId for this class.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,299 @@
...
+  @Nullable private XdsClient xdsClient;
+
+  /** Wraps the XdsClientImpl for use by SdsProtocolNegotiators. */
+  public XdsClientWrapperForServerSds(int port, Bootstrapper bootstrapper) {
+    this.logId = InternalLogId.allocate("GrpcServer", Integer.toString(port));
+    Bootstrapper.BootstrapInfo bootstrapInfo = null;
+    try {
+      bootstrapInfo = bootstrapper.readBootstrap();
+    } catch (IOException e) {
+      logger.log(Level.SEVERE, "Error from readBootstrap", e);
```
voidzcy: The exception goes into nowhere except a log message. And this is a constructor. This pattern is bad.

Same for other places, error handling/propagation in this class is in appropriate.

xds/src/test/java/io/grpc/xds/XdsClientWrapperForServerSdsTest.java
```diff
@@ -0,0 +1,213 @@
...
+    XdsClient.ListenerUpdate listenerUpdate =
+        XdsClient.ListenerUpdate.newBuilder().setListener(listener).build();
+    registeredWatcher.onListenerChanged(listenerUpdate);
+    DownstreamTlsContext downstreamTlsContext =
+        xdsClientWrapperForServerSds.getDownstreamTlsContext(channel);
+    assertThat(downstreamTlsContext).isSameInstanceAs(expectedTlsContext);
+  }
+
+  @After
+  public void tearDown() {}
```
voidzcy: If nothing to do, just don't implement it.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,299 @@
...
+public final class XdsClientWrapperForServerSds {
+  private static final Logger logger =
+      Logger.getLogger(XdsClientWrapperForServerSds.class.getName());
+
+  private static final EventLoopGroupResource eventLoopGroupResource =
+      Epoll.isAvailable() ? new EventLoopGroupResource("GrpcServerXdsClient") : null;
+  private final InternalLogId logId;
+  // Must be accessed from the syncContext
+  private boolean panicMode;
+  final SynchronizationContext syncContext =
```
voidzcy: Not sure if creating a sync context here is appropriate. The whole control plane workflow should use a single sync context, it should not be something to be created just because you need it. You need to consider the whole server side flow.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,299 @@
...
+            panic(e);
+          }
+        });
+
+  private EnvoyServerProtoData.Listener curListener;
+  @Nullable private ObjectPool<XdsClient> xdsClientPool;
+  @Nullable private XdsClient xdsClient;
+
+  /** Wraps the XdsClientImpl for use by SdsProtocolNegotiators. */
+  public XdsClientWrapperForServerSds(int port, Bootstrapper bootstrapper) {
```
voidzcy: It's usually a bad practice to do so much in constructor.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: You are right and my original prototype implementation didn't have the pool. When I adapted to the new XdsClientImpl I ended up using the existing pattern. I don't think there will be other classes needing to use a "shared" XdsClient on the server side so I will change this to the non-pool implementation.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: Okay, this was really used in the `syncContext` definition and I used the pattern from elsewhere. I notice you have some comments on `SynchronizationContext` as well so once that is addressed this one hopefully will be addressed to.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: I'll see if this can be refactored into a separate `init` method but it will have to be in this class because this is the bridge between `XdsClientImpl` and `SdsProtocolNegotiators`. Do you agree?

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: Good point. At this point there are no "listeners" like you have the `NameResolver.Listener2`  to propagate this error to. Will wrap this into a runtime exception.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: I looked at the client side and over there a sync context is [created](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/ManagedChannelImpl.java#L152) by a `ManagedChannelImpl`  and used everywhere starting from the name-resolver, load-balancers, and the XdsClient as well. So you have one sync context per managed channel. In this case (server side), there is currently no sync context being created because one wasn't needed on the server side so far. So I ended up creating it in this class which associates one sync context with one server (listener) and that cardinality seems correct. But may be the correct place to create this sync context is in the server builder and plumb the value all the way down here instead of creating it here. Let me know if you agree.

voidzcy: Would need suggestion from @creamsoup

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -63,12 +65,12 @@ public static ProtocolNegotiatorFactory clientProtocolNegotiatorFactory(
 
   /**
    * Creates an SDS based {@link ProtocolNegotiator} for a {@link io.grpc.netty.NettyServerBuilder}.
```
creamsoup: can you add doc for the new behavior (fall back to plaintext)?

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -294,11 +293,29 @@ public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
...
     @Override
     protected void handlerAdded0(final ChannelHandlerContext ctx) {
+      DownstreamTlsContext downstreamTlsContext =
+          xdsClientWrapperForServerSds == null
+              ? null
+              : xdsClientWrapperForServerSds.getDownstreamTlsContext(ctx.channel());
+      if (isTlsContextEmpty(downstreamTlsContext)) {
+        downstreamTlsContext = downstreamTlsContextFromBuilder;
+      }
+      if (isTlsContextEmpty(downstreamTlsContext)) {
```
creamsoup: can you add logging if it falls back to plaintext

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -294,11 +293,29 @@ public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
...
+      this.downstreamTlsContextFromBuilder = downstreamTlsContext;
+      this.xdsClientWrapperForServerSds = xdsClientWrapperForServerSds;
+    }
+
+    private static boolean isTlsContextEmpty(DownstreamTlsContext downstreamTlsContext) {
+      return downstreamTlsContext == null || !downstreamTlsContext.hasCommonTlsContext();
     }
 
     @Override
     protected void handlerAdded0(final ChannelHandlerContext ctx) {
```
creamsoup: we would love to make ProtocolNegotiationHandler as small as possible. can you make a new handler that decides which negotiator to use (plaintext/sds)?

xds/src/test/java/io/grpc/xds/XdsClientWrapperForServerSdsTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+  @Test
+  public void verifyListenerWatcherRegistered() {
+    verify(xdsClient, times(1)).watchListenerData(eq(PORT), any(XdsClient.ListenerWatcher.class));
+  }
+
+  @Test
+  public void listener_nullFilterChainMatch1() throws UnknownHostException {
+    // filterChainMatch missing for filter1, so it matches and returns tlsContext1
+    commonFilterChainMatchTest(-1, null, null, "192.168.10.1", "192.168.10.2", tlsContext1);
```
creamsoup: this is harder to read. consider parameterized test.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: Prefer factory than putting it in the caller.

voidzcy: > can you move some of the prep work to the caller side or factory?

Yes, this is what I was thinking about.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,299 @@
...
+   *
+   * <p>-1 => mismatch (of port or IP address etc)
+   *
+   * <p>0 => unimplemented (types or logic etc)
+   *
+   * <p>1 => filterChainMatch is null, so nothing to match.
+   *
+   * <p>value > 2 indicates some kind of IP address match.
+   */
+  private static int getFilterChainMatchScore(FilterChainMatch filterChainMatch, Channel channel) {
```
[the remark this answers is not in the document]
sanjaypujare: In the original filterChainMatch (from proto) there are other fields like the remote address,  which will ideally need to be matched with the remote address, remote port in the Channel. Trying to implement all the logic now seems like an overkill now since most filterChainMatch don't use it but it makes sense to pass Channel instead of separate fields from the Channel. Agree?

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,299 @@
...
+   * Locates the best matching FilterChain to the channel from the current listener and returns the
+   * DownstreamTlsContext from that FilterChain.
+   */
+  public DownstreamTlsContext getDownstreamTlsContext(Channel channel) {
+    if (curListener != null && channel != null) {
+      List<FilterChain> filterChains = curListener.getFilterChains();
+      int highestScore = -1;
+      FilterChain bestMatch = null;
+      for (FilterChain filterChain : filterChains) {
+        int curScore = getFilterChainMatchScore(filterChain.getFilterChainMatch(), channel);
```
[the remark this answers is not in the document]
sanjaypujare: Can you give me some idea why this approach is more error prone? Errors are possible with computation of the score (or heuristics involved in assigning a score) but I don't see why `Collections.max` is better (except in cases where you have multiple max candidates and there is some logic about which one to pick). Also we don't need elaborate sort/select algorithm as I don't see the control plane sending more than one or two with complex selection criterion.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
creamsoup: it is a private method so we can always change if we need more fields. also, if you use comparator it may look very different.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
creamsoup: it is maybe not as error prone as it is (although it is harder to read the code). when you try to modify, it can be very hard to keep everything aligned (how many methods here need to be modified to add new logic?). thus, those score has no meaning and it only makes reader/maintainer confused. if you implement it as comparator, it is easy to add new logic. this it doesn't have any performance penalty in terms of big O.
i tried to do the same thing before. and, i got same comment which I now agree.

xds/src/test/java/io/grpc/xds/XdsClientWrapperForServerSdsTest.java
sanjaypujare: You mean `org.junit.runners.Parameterized` ? I prefer adding more comments to the existing code to make it more readable.

creamsoup: requires comment means it is not readable. the code structure is already ready to be converted to parameterized. anyways, i'll leave it up to you as long as it is readable.

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
sanjaypujare: Sounds like a good idea. How about if I create a TODO task for a separate PR? So this PR doesn't have too many changes.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: @creamsoup lmk if you have comments here

xds/src/test/java/io/grpc/xds/XdsClientWrapperForServerSdsTest.java
sanjaypujare: With parameterized, it is helpful if parameter values tell you what the test is. In this case one needs to know more. e.g. passing null for certain params doesn't convey anything. But I'll give it some more thought.

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
creamsoup: this should be relatively simple change (okay maybe some boiler plates, but this way will be helpful when you need to debug), i don't see why we want to delay.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
sanjaypujare: Okay, just to clarify: you are saying to ditch the whole idea of scores and just compare 2 filterChains for a given channel? So initialize the `Comparator` with the channel and then given 2 filterChains pick one? That's interesting and let me think about it. In any case I think is probably overkill for the number of filterChains we will be expected to compare (in most cases just 1?) But I'll definitely consider it

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
sanjaypujare: Yes, with comparator it will use localAddress now and Channel later.

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
sanjaypujare: Yes, it is actually not that difficult. And with your other suggestion about handling exception(s) in the caller for `xdsClientWrapperForServerSds`, this change goes well with that. So I'll do it in this one

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
sanjaypujare: `Collections.max` with comparator doesn't work here. There may not be a matching filter at all in which case we need to return null and that's not handled by `max`.

sanjaypujare:
@creamsoup @voidzcy all comments addressed. PTAL

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,259 @@
...
+  @Nullable private XdsClient xdsClient;
+  private final int port;
+
+  /** Creates a new instance. */
+  public static XdsClientWrapperForServerSds newInstance(
+      int port, Bootstrapper bootstrapper, SynchronizationContext syncContext) throws IOException {
+    Bootstrapper.BootstrapInfo bootstrapInfo = bootstrapper.readBootstrap();
+    final List<Bootstrapper.ServerInfo> serverList = bootstrapInfo.getServers();
+    if (serverList.isEmpty()) {
+      throw new NoSuchElementException("No management server provided by bootstrap");
```
voidzcy: I doubt the usage of `NoSuchElementException` here and `IllegalStateException` below. Error handling still looks a bit abrupt, basically it would panic immediately when encountering those cases. @creamsoup Does server side have any error propagation APIs?

sanjaypujare: No these are caught by the caller where it will switch over to plaintext. As mentioned elsewhere there is no error propagation/error listener similar to what you have on the client side.

creamsoup: netty installs `WriteBufferingAndExceptionHandler` which handles exception. but, this specific exception happens on `XdsServerBuilder#build`.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+    final Node node = bootstrapInfo.getNode();
+    if (eventLoopGroupResource == null) {
+      throw new IllegalStateException(Epoll.unavailabilityCause());
```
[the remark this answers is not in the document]
sanjaypujare: Good point! If `Epoll` is not available would it be a good idea to use `DefaultEventLoopGroup` ? Or `ScheduledThreadPoolExecutor` ? @voidzcy for the `XdsClientImpl` what will be better?

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,259 @@
...
+      InetSocketAddress localInetAddr = (InetSocketAddress) localAddress;
+      if (port != localInetAddr.getPort()) {
+        throw new IllegalStateException(
+            "Channel localAddress port does not match requested listener port");
+      }
+      List<FilterChain> filterChains = curListener.getFilterChains();
+      int highestScore = -1;
+      FilterChain bestMatch = null;
+      for (FilterChain filterChain : filterChains) {
+        int curScore = getFilterChainMatchScore(filterChain.getFilterChainMatch(), localInetAddr);
```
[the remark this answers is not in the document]
sanjaypujare: Did you see my other comment https://github.com/grpc/grpc-java/pull/6838#discussion_r396115410 ?

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
creamsoup: post check is what i was expecting. this is relatively minor issue compare to the score based approach.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
creamsoup: if we can reuse one from `XdsClientImpl` that would be the best. i think `ScheduledThreadPoolExecutor` is better than `DefaultEventLoopGroup`.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
creamsoup: commented on the other one.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
sanjaypujare: @creamsoup Honestly I think the Comparator here is an overkill assuming the control plane will mostly give us a single filterChain matching our requirements. The multiple filterChains happens with Envoy (since it's a proxy) but most likely won't happen with proxyless gRPC. I still need to confirm with the control plane team and if that is confirmed I can even remove the current logic and error in case of multiple filters and verify the single filter returned matches what I need.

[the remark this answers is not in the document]
creamsoup: i know it can be tedious to covert but it is not hard nor overkill (compare to score based) while it makes the code more maintainable/readable.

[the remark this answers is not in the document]
sanjaypujare: I spent a whole afternoon on it and I realized I still have to compute scores for doing a match :-) We can discuss offline to discuss the details. If the control plane sends only one filterChain then there is no need to implement a matcher and we can just error out when multiple are received. Do you agree?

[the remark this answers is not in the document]
creamsoup: okay if we don't need to compare, problem solved. 
I am not sure about it is desired behavior or not. how confident are we that the control plain will only send 1 filterChain (including in near future)?

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
voidzcy: > No these are caught by the caller where it will switch over to plaintext. As mentioned elsewhere there is no error propagation/error listener similar to what you have on the client side.

I still do not follow where these runtime exceptions are handled.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
sanjaypujare: reuse which one from `XdsClientImpl`? `XdsClientImpl ` doesn't have one of its own and I am passing one to it while creating a `XdsClientImpl` and there is no other `XdsClientImpl` instance to reuse it from unless I am missing something. I will use `ScheduledThreadPoolExecutor`.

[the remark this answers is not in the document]
creamsoup: it was more generic comment. yes please use the `ScheduledThreadPoolExecutor`.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: https://github.com/grpc/grpc-java/pull/6838/files#diff-fe0356a9c9642d0f91961b377c24e5d3R265 line 265 in SdsProtocolNegotiators.java (in this commit). Just follow calls to `XdsClientWrapperForServerSds.newInstance`.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
voidzcy: I mean `XdsClientImpl` just takes in a `ScheduledExecutorService`, with whatever its concrete implementation is. It doesn't matter from `XdsClientImpl`'s perspective.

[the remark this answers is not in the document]
sanjaypujare: @voidzcy I think there is some miscommunication here. I know `XdsClientImpl` takes a `ScheduledExecutorService` and I (as the creator of that object) pass a concrete implementation. When @creamsoup mentioned "we can reuse one from XdsClientImpl that would be the best" there isn't anything to reuse so that part threw me off. The plan now is to  use `ScheduledThreadPoolExecutor`. I hope we are all on the same page.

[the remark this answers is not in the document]
sanjaypujare: Oh  BTW, if @creamsoup meant, do the same thing the current client side user of `XdsClientImpl` does then that is what I did (after tracing laboriously where it came from via XdsNameResolver etc). I don't have the reference handy and I cannot see what it does when epoll is not available.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
voidzcy: That's catching the IOException. I am talking about `NoSuchElementException` here and `IllegalStateException` exception below. They are not gracefully handled. It panic the server indefinitely once getting into those situations.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
voidzcy: I am following. I thought you were gonna change `XdsClientImpl` and @creamsoup suggested we are able to "reuse" the existing code as is.  So nvm.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: Good point, thanks for catching that. The `IllegalStateException` will go away since I will use `ScheduledThreadPoolExecutor` as we discussed in the other thread. And I will handle `NoSuchElementException` (or whatever is suitable type of exception for that case).

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
sanjaypujare: since it accepts a `ScheduledExecutorService` I never intended to change `XdsClientImpl`. Anyway looks like this is all settled now

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
sanjaypujare: So I am talking to the control plane team to find out how many filterChains I can expect to get. If it is just 1 (or max 2) I don't see why spend so much time on the Comparator. But I have opened this again (marked it not resolved) and will get back to this after addressing all the other comments. By that time I would have some answers and hopefully a clearer idea about which way to go and we can discuss offline if nedeed.

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -245,12 +253,21 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause)
...
-    ServerSdsProtocolNegotiator(DownstreamTlsContext downstreamTlsContext) {
+    ServerSdsProtocolNegotiator(
+        DownstreamTlsContext downstreamTlsContext, int port, SynchronizationContext syncContext) {
       this.downstreamTlsContext = downstreamTlsContext;
+      XdsClientWrapperForServerSds localXdsClientWrapperForServerSds;
+      try {
+        localXdsClientWrapperForServerSds =
+            XdsClientWrapperForServerSds.newInstance(port, Bootstrapper.getInstance(), syncContext);
+      } catch (Exception e) {
+        logger.log(Level.SEVERE, "XdsClientWrapperForServerSds constructor", e);
```
voidzcy: nit: rephrase the error message. This is above-DEBUG-level logging, so it should be semantically meaningful to readers. It should be something like "Failed to ...". Also, "SEVER" means something is very broken and the system is probably panic. But it turns out we are still able to proceed by falling back to plaintext. I think if something required is missing and we have nothing to do, we just let it panic and halt. But if we do have alternative to recover, it may just be a warning.

Same thing may apply for other log messages.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,281 @@
...
+        logger.log(Level.SEVERE, "cidrRange address parsing", e);
+      }
+    }
+    // TODO(sanjaypujare): implement CIDR logic to match prefixes if needed
+    return 0;
+  }
+
+  /** Shutdown this instance and release resources. */
+  public void shutdown() {
+    logger.log(Level.INFO, "Shutdown");
```
voidzcy: nit: this will be printed as just "Shutdown", without anything else. It doesn't give any useful information.

Also, I doubt there should be any log messages above INFO level as anything above (or equal) INFO level will be printed out by default.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,281 @@
...
+    public void close(ScheduledExecutorService instance) {
+      try {
+        if (instance instanceof EpollEventLoopGroup) {
+          ((EpollEventLoopGroup)instance).shutdownGracefully(0, 0, TimeUnit.SECONDS).sync();
+        } else {
+          instance.shutdown();
+        }
+      } catch (InterruptedException e) {
+        logger.log(Level.SEVERE, "from EventLoopGroup.shutdownGracefully", e);
+        Thread.currentThread().interrupt(); // to not "swallow" the exception...
```
voidzcy: nit: delete your personal note.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,281 @@
...
+      new TimeServiceResource("GrpcServerXdsClient");
+
+  private EnvoyServerProtoData.Listener curListener;
+  // TODO(sanjaypujare): implement shutting down XdsServer which will need xdsClient reference
+  @SuppressWarnings("unused")
+  @Nullable private XdsClient xdsClient;
+  private final int port;
+  private final ScheduledExecutorService timeService;
+
+  /** Creates a new instance. */
```
voidzcy: nit: redundant comment, should delete.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,281 @@
...
+    logger.log(Level.INFO, "Shutdown");
+    if (xdsClient != null) {
+      xdsClient.shutdown();
+    }
+    if (timeService != null) {
+      timeServiceResource.close(timeService);
+    }
+  }
+
+  private static final class TimeServiceResource
```
voidzcy: IMO, if `syncContext` comes from the caller side, this should also come from the caller side.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
sanjaypujare: So it looks like the planned interface will take some time to be implemented on the control plane side so we still need to be prepared to handle multiple filterChains on our side. But the bottom line is that there will be exactly one filterChain that will be a match for the localAddress where destPort, localIPaddress will match and there is no need to implement either a score or sorting based selection of filterChain. So that is my plan for now: select the filterChain that perfectly matches my localAddress or else report error.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: Forced by checkstyle - for a public method this is needed,

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: But the "caller" is this class as far as XdsClientImpl is concerned. If you go up the call chain (all the way up to `XdsServerBuilder`) no other component either needs or creates a `ScheduledExecutorService` so this seems to be the right place.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
voidzcy: Then write something semantically meaningful.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: The `AbstractServerImplBuilder` defines/uses `DEFAULT_EXECUTOR_POOL` which cannot be used here because it's not a `ScheduledExecutorService` (and probably not accessible). There is no other "caller-provided" `ScheduledExecutorService` we can use unless this wrapper class creates it for `XdsClientImpl`.

voidzcy: Same thing apply to syncContext. It would be very weird that they are created separately.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: So all the *dsLoadBalancer.shutdown() (e.g. `CdsLoadBalancer.shutdown`) have this:
```
  public void shutdown() {
    logger.log(XdsLogLevel.INFO, "Shutdown");
...
}
```
Agreed `XdsLogLevel.INFO` is not the same as `Level.INFO`  but they all say "Shutdown" so being consistent. I am changing from `INFO` to `FINER` .

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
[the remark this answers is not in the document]
sanjaypujare: So I ended up implementing the `Comparator` as per the suggestion here. PTAL

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
voidzcy: xDS policies (e.g., CdsLoadBalancer) uses `XdsLogger`, which prepends a unique `LogId` to the log message. So logging "Shutdown" really would print out "[cds-lb<ID>] Shutdown". But here, when seeing a "Shutdown" log message, we have no clue (unless the class name is shown) what is being shut down. Not a big issue.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -0,0 +1,296 @@
...
+          }
+        });
+  }
+
+  /**
+   * Locates the best matching FilterChain to the channel from the current listener and if found
+   * returns the DownstreamTlsContext from that FilterChain, else null.
+   */
+  @Nullable
+  public DownstreamTlsContext getDownstreamTlsContext(Channel channel) {
```
voidzcy: A better API would be to take in an `InetSocketAddress` as the argument instead of a `Channel`, which makes the method specification stronger (i.e., weaker precondition, the method requires less information).

sanjaypujare: You probably missed the comment exchange where @creamsoup talked about this for internal functions.

I do agree that a type like `InetSocketAddress` is better than `Channel` in an API but the filters to be matched have various things like remoteAddress, protocol-type (not just TCP sockets) even if we haven't implemented all of the matching for those fields. And this an internal class (API) and is used by `SdsProtocolNegotiators` which is Netty specific so using `Channel` seems to be the better option.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: each log line output has the classname and method as the prefix so users know where this message is coming from.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
voidzcy: Alright, that't fine. I was just based on what I saw in this method.

sanjaypujare:
> LGTM for XdsClientWrapper related classes, while I am concerning the usages of logging levels. In general you should have nothing printed out by default unless the system is panic and halted.

Good point about logging levels. For a new feature that is going to be extensively debugged and tested in the early phases I think it will be good to have log messages come out by default. I will do a clean sweep of all log messages in this component when this code is production ready.

PR #6840, opened 2020-03-18
Implemented Frame Size Negotiation in ALTS for gRPC Java

ashithasantosh:
gRPC Java uses zero copy frame protectors, and hence are capable of handling larger frame size. Frame Size Negotiation support in ALTS, extends send frame size from 16KB to 128KB, and thereby improves performance.

Design Document: go/seal-frame-size-negotiation-alts

### DIFF
--- a/alts/src/main/java/io/grpc/alts/internal/AltsHandshakerClient.java
+++ b/alts/src/main/java/io/grpc/alts/internal/AltsHandshakerClient.java
@@ -82,6 +82,7 @@ class AltsHandshakerClient {
         startClientReq.addTargetIdentitiesBuilder().setServiceAccount(serviceAccount);
       }
     }
+    startClientReq.setMaxFrameSize(AltsTsiFrameProtector.getMaxFrameSize());
     req.setClientStart(startClientReq);
   }
 
@@ -97,6 +98,7 @@ class AltsHandshakerClient {
     if (handshakerOptions.getRpcProtocolVersions() != null) {
       startServerReq.setRpcVersions(handshakerOptions.getRpcProtocolVersions());
     }
+    startServerReq.setMaxFrameSize(AltsTsiFrameProtector.getMaxFrameSize());
     req.setServerStart(startServerReq);
   }
 
--- a/alts/src/main/java/io/grpc/alts/internal/AltsTsiHandshaker.java
+++ b/alts/src/main/java/io/grpc/alts/internal/AltsTsiHandshaker.java
@@ -26,11 +26,15 @@ import java.nio.ByteBuffer;
 import java.security.GeneralSecurityException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Negotiates a grpc channel key to be used by the TsiFrameProtector, using ALTs handshaker service.
  */
 public final class AltsTsiHandshaker implements TsiHandshaker {
+  private static final Logger logger = Logger.getLogger(AltsTsiHandshaker.class.getName());
+
   public static final String TSI_SERVICE_ACCOUNT_PEER_PROPERTY = "service_account";
 
   private final boolean isClient;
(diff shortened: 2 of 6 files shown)


alts/src/main/java/io/grpc/alts/internal/AltsHandshakerClient.java
```diff
@@ -35,6 +35,9 @@
   private static final String APPLICATION_PROTOCOL = "grpc";
   private static final String RECORD_PROTOCOL = "ALTSRP_GCM_AES128_REKEY";
   private static final int KEY_LENGTH = AltsChannelCrypter.getKeyLength();
+  // Frame size negotiation extends frame size range to [MIN_FRAME_SIZE, MAX_FRAME_SIZE].
+  private static final int MIN_FRAME_SIZE = 16 * 1024;
```
jiangtaoli2016: Currently max_frame_size is defined in AltsTsiFrameProtector.java.
Please define min/max in AltsTsiFrameProtector.java.

alts/src/main/java/io/grpc/alts/internal/AltsTsiHandshaker.java
```diff
@@ -178,6 +182,17 @@ public TsiFrameProtector createFrameProtector(int maxFrameSize, ByteBufAllocator
     byte[] key = handshaker.getKey();
     Preconditions.checkState(key.length == AltsChannelCrypter.getKeyLength(), "Bad key length.");
 
+    // In case the peer does not send max frame size (e.g. peer is gRPC Go or peer uses an old
+    // binary), the negotiated frame size is set to MIN_FRAME_SIZE. Otherwise, the negotiated
+    // value is based on peer frame size. maxFrameSize parameter is ignored.
+    maxFrameSize = AltsHandshakerClient.getMinFrameSize();
```
jiangtaoli2016: As in my other comment above, don't set maxFrameSize here. 
If there is no frame size negotiation, use the value from parameter -- which will be the default frame size

alts/src/main/java/io/grpc/alts/internal/AltsHandshakerClient.java
jiangtaoli2016: In AltsTsiFrameProtector.java, we could have 4 constants

MIN_FRAME_SIZE, 16KB
MAX_FRAME_SIZE, 128KB
DEFAULT_FRAME_SIZE, 16KB
LIMIT_MAX_ALLOWED_FRAME_SIZE, 1MB

We may need to change internal code to sync up.

ashithasantosh: Do we need both MIN_FRAME_SIZE and DEFAULT_FRAME_SIZE?
I think for consistency sake, MIN_FRAME_SIZE name would be preferred. But that would involve updating MAX_ALLOWED_FRAME_BYTES to MIN_FRAME_SIZE, and wherever the getter is called. If the variable name modification seems fine, we could have just 3 constants.

jiangtaoli2016: 3 constants are fine. 
We need 4 getter functions to be backward compatible.

ashithasantosh: I am using just 3 getters. 

For Loas2TsiHandshaker.java, i will cherrypick this CL, and include the update there.

jiangtaoli2016:
@ashithasantosh , please work with gRPC folks (e.g., @ejona86 ) on the internal patch, since you removed the function getMaxAllowedFrameBytes.

ashithasantosh:
I created a CL that imports this PR. 
https://critique.corp.google.com/#review/305926248
This CL includes update to Loas2TsiHandshaker.java

After that CL is reviewed, we can merge this PR and then I will submit that CL.

ashithasantosh:
Thanks for the review, @jiangtaoli2016!

PR #6844, opened 2020-03-20
RouteGuideClient example: use E7 format for latitude & longitude

chalin:
As indicated in the proto file, "Points are represented as latitude-longitude pairs in the E7 representation".

@srini100

### DIFF
--- a/examples/src/main/java/io/grpc/examples/routeguide/RouteGuideClient.java
+++ b/examples/src/main/java/io/grpc/examples/routeguide/RouteGuideClient.java
@@ -218,8 +218,8 @@ public class RouteGuideClient {
 
     try {
       RouteNote[] requests =
-          {newNote("First message", 0, 0), newNote("Second message", 0, 1),
-              newNote("Third message", 1, 0), newNote("Fourth message", 1, 1)};
+          {newNote("First message", 0, 0), newNote("Second message", 0, 10000000),
+              newNote("Third message", 10000000, 0), newNote("Fourth message", 10000000, 10000000)};
 
       for (RouteNote request : requests) {
         info("Sending message \"{0}\" at {1}, {2}", request.getMessage(), request.getLocation()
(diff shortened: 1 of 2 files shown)


srini100:
The change looks fine to me but I see that other languages like C++ and Python are also using 0 and 1. Not sure what the history is on this. Assigning to @ejona86 to decide.

chalin:
Yeah, probably isn't essential, since this part of the example code is independent of the rest. This issue arose because we print normalized degrees in the Kotlin version of the code, and it seemed a bit strange to see 1e-7 as a degree (FYI).

examples/src/main/java/io/grpc/examples/routeguide/RouteGuideClient.java
```diff
@@ -218,8 +218,8 @@ public void onCompleted() {
 
     try {
       RouteNote[] requests =
-          {newNote("First message", 0, 0), newNote("Second message", 0, 1),
-              newNote("Third message", 1, 0), newNote("Fourth message", 1, 1)};
+          {newNote("First message", 0, 0), newNote("Second message", 0, 10000000),
+              newNote("Third message", 10000000, 0), newNote("Fourth message", 10000000, 10000000)};
```
dapengzhang0, line 222: nit: it's hard to count how many zeros are there. `10_000_000` might be better.

chalin: Done

chalin:
Test file updated too.

chalin:
All checks are passing now.

chalin:
Anything else to be done before this is merged?

dapengzhang0:
Ping @ejona86 

>The change looks fine to me but I see that other languages like C++ and Python are also using 0 and 1. Not sure what the history is on this. Assigning to @ejona86 to decide.

ejona86:
The other languages should be fixed as well, but I don't want organizing that to hold up this PR longer.

PR #6851, opened 2020-03-23
core: copy the SchemaDescriptor when rebuilding descriptor

herbyderby:
useMarshalledMessages works by duplicating a ServerServiceDefinition while replacing just the marshallers. It currently does not copy over the SchemaDescriptors, which breaks at least the ProtoReflectionService.

### DIFF
--- a/api/src/main/java/io/grpc/ServerInterceptors.java
+++ b/api/src/main/java/io/grpc/ServerInterceptors.java
@@ -187,13 +187,15 @@ public final class ServerInterceptors {
       wrappedMethods.add(wrapMethod(definition, wrappedMethodDescriptor));
     }
     // Build the new service descriptor
-    final ServerServiceDefinition.Builder serviceBuilder = ServerServiceDefinition
-        .builder(new ServiceDescriptor(serviceDef.getServiceDescriptor().getName(),
-            wrappedDescriptors));
-    // Create the new service definiton.
-    for (ServerMethodDefinition<?, ?> definition : wrappedMethods) {
-      serviceBuilder.addMethod(definition);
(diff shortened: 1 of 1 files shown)


zhangkun83:
LGTM

api/src/main/java/io/grpc/ServerInterceptors.java
```diff
@@ -187,13 +187,15 @@ public InputStream parse(final InputStream stream) {
       wrappedMethods.add(wrapMethod(definition, wrappedMethodDescriptor));
```
zhangkun83: `wrappedMethods` is never used, and this is causing the CI error.

ejona86: Look like the earlier for loop should not have been deleted and should occur after `serviceBuilder` is created (just like before).

herbyderby: Sorry, not sure what happened--some patching problem on my end. Should look something like this:

```
    // Build the new service descriptor
    final ServiceDescriptor.Builder serviceDescriptorBuilder =
        ServiceDescriptor.newBuilder(serviceDef.getServiceDescriptor().getName())
            .setSchemaDescriptor(serviceDef.getServiceDescriptor().getSchemaDescriptor());
    // Create the new service definition.
    final ServerServiceDefinition.Builder serviceBuilder =
        ServerServiceDefinition.builder(serviceDescriptorBuilder.build());
    for (ServerMethodDefinition<?, ?> definition : wrappedMethods) {
      serviceBuilder.addMethod(definition);
    }
    return serviceBuilder.build();
```

zhangkun83: Applied fix.

PR #6852, opened 2020-03-23
core: preserve KnownLength when wrapping InputStream

herbyderby:
useInputStreamMessages ensures that the InputStream supports marking by wrapping the stream in a BufferedInputStream if markSupported() returns false. This change uses a new subclass of BufferedInputStream that also implements KnownLength, when the original stream also implements KnownLength.


ejona86:
In places we check for KnownLength we try to check for ByteArrayInputStream as well, explicitly. It seems that is missing in ProtoLiteUtils. We should just fix it to be `inputStream instanceof KnownLength || inputStream instanceof ByteArrayInputStream`.

(Alternatively we could have a `static` method that did the check for us. But we'd need to find a home for the method.)

ejona86:
Oh, I see. This is BufferedInputStream

PR #6865, opened 2020-03-27
xds: rollback PGV dependency from using maven artifact to importing proto source

dapengzhang0:
PGV is adding new fields (`well_known_regex` and `strict`) to proto and some of our new features require the most current envoy API which in turn requires the new fields in latest PGV proto (https://github.com/envoyproxy/envoy/blob/master/api/envoy/api/v2/core/base.proto#L251), but they are not available in the latest PGV maven release.

### DIFF
--- a/xds/build.gradle
+++ b/xds/build.gradle
@@ -25,11 +25,7 @@ dependencies {
             project(':grpc-services'),
             project(path: ':grpc-alts', configuration: 'shadow')
     def nettyDependency = compile project(':grpc-netty')
-    def pgvDependency = compile (libraries.pgv) {
-        // PGV depends on com.google.protobuf:protobuf-java 3.6.1 conflicting with :grpc-protobuf
-        exclude group: 'com.google.protobuf'
-    }
-
+    
     compile (libraries.protobuf_util) {
         // prefer our own versions instead of protobuf-util's dependency
         exclude group: 'com.google.guava', module: 'guava'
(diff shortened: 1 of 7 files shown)


xds/build.gradle
```diff
@@ -77,7 +74,6 @@ shadowJar {
     classifier = null
     dependencies {
         include(project(':grpc-xds'))
-        include(dependency('io.envoyproxy.protoc-gen-validate:'))
     }
     relocate 'com.github.udpa', 'io.grpc.xds.shaded.com.github.udpa'
     relocate 'io.envoyproxy', 'io.grpc.xds.shaded.io.envoyproxy'
```
ejona86: Seems like we would need a new relocation.

dapengzhang0: Are you referring to shadowing PGV?
https://github.com/grpc/grpc-java/blob/ad5701815c6659968a5b47fb3c3cbe17bf5d3ed8/xds/third_party/protoc-gen-validate/src/main/proto/validate/validate.proto#L5

The parent package `io.envoyproxy` is relocated.

ejona86: Yeah. Okay. Sounds good.

xds/third_party/protoc-gen-validate/import.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+# Copyright 2018 The gRPC Authors
```
sanjaypujare: should change year to 2020?

dapengzhang0: Good question. I don't know the answer, will keep it as is.

Issue #4327, opened 2018-04-10
Reconnect wait time? 


Necropaw:
Please answer these questions before submitting your issue.

### What version of gRPC are you using?
1.11.0

### What did you expect to see?
Some way to configure a delay before a reconnect may happen.

### Details

We are currently running our applications in a kubernetes cluster and using services to point to the gRPC server:

`client-pod -> service (IP: 192.168.0.1) -> server-pod (IP: 192.168.0.10)`

When the `server-pod` is terminated, it sends a `GO AWAY` to the client and kubernetes spawns a new pod.

`client-pod -> service (IP: 192.168.0.1) -> server-pod2 (IP: 192.168.0.11)`

If in between the `GO AWAY` and the start of `server-pod2` the client-pod tries to connect, the kubernetes service forwards the request to the old pod (IP: 192.168.0.10) and the client-pod has to wait for a socket timeout around 2min before it can reconnect. 

Do you know of a way to configure the client to wait after a `GO AWAY` before trying to reconnect? The waiting time would be very short (like 5s).

Log: 
[grpc-call.log](https://github.com/grpc/grpc-java/files/1893916/grpc-call.log)

zpencer:
I do see some code related to tuneable reconnection policies in ExponentialBackoffPolicy but unfortunately I can't find any public APIs for configuring this. I'll double check with my team to see if this is the appropriate to expose.

ejona86:
No, we don't allow the backoff policy to be configured. In this case, it seems there's a network black hole, which really should be fixed.

#1943 would help reduce the impact here, as it would limit the initial (failed) reconnect to 20 seconds.

Necropaw:
And this is more a kubernetes issue as an grpc issue: https://github.com/kubernetes/kubernetes/issues/48719
https://github.com/kubernetes/kubernetes/issues/38456

ejona86:
@Necropaw, thanks for the k8s links. We'll take this also as an implicit vote to fix #1943 (as the point of that issue is to avoid huge slowdowns when the network is broken, such as you were experiencing).

Issue #1943, opened 2016-06-16
Implement connection attempt exponential timeout


ejona86:
This is not for exponential backoff, but for exponential timeout. Currently the timeouts involved aren't tuned at all between attempts. We want to have a single timeout applying to all parts of the connection attempt, as defined in the [connection backoff spec](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md). This generally prevents the need for users to specify connect timeouts.

PR #6873, opened 2020-03-28
backOffPolicyProvider() for ManagedChannelBuilder

antonbashir:
In this PR added method in ManagedChannelBuilder for changing backoff policy.


antonbashir:
> Could you provide some context about this PR? It would help a lot if you could describe what you're trying to accomplish that would require pasasing in a backoff policy provider to `ManagedChannel`. Thanks!

Of course!

Imagine that you stop your GRPC Server but do not stop GRPC client.
Then GRPC client starts backoff process and exponentially increments interval.
After start of GRPC server next request from client will be waiting for successful reconnect after backoff interval. 
Because of it in our project we have situation when first request processing during (5-10 seconds).
So, i want to set some minimal value (one second or less) to max backoff interval parameter.
May be i mistake in this reason of long first request processing, but anyway i think that so parameters like backoff policy must be configurable through builder.

ejona86:
> Imagine that you stop your GRPC Server but do not stop GRPC client.
Then GRPC client starts backoff process and exponentially increments interval.
After start of GRPC server next request from client will be waiting for successful reconnect after backoff interval.

This is essential behavior. Without the backoff, clients will reconnect rapidly to the server and overload it. The backoff _protects_ the server.

So I don't think reducing the backoff time is a good approach. Why does the server restart? And what type of server is it such that there is only one instance?

mizantrop2397:
Same issue about my case:
https://github.com/grpc/grpc-java/issues/4327

> Why does the server restart? - deployment process of new service version. 

There is no any Network or Application error, just graceful server shutdown.

> And what type of server is it such that there is only one instance?

No, no, not one instance. There are two OpenShift's pods. As balancer we use OpenShift/Kubernetes Service-Name Load balancing and "canary" redeployment policy. Anyway, i don't sure that GRPC users always have clustered environment, despite of that single-instance mode is not recommended.

> Without the backoff, clients will reconnect rapidly to the server and overload it. The backoff protects the server.

I am absolutely agree with this fact. But in this PR i don't try to reduce backoff default interval or changing some backoff logic. I just offered to make this parameter configurable same as load balancing policy or deadline timeout parameters. As default policy use current Exponential policy with default parameters. Also, as you could see in last version of source code, there is comment: 

> // TODO(carl-mastrangelo): Allow clients to pass this in

I saw this comment and thought that not only me want to make this parameter configurable. 
Also, i want to try some other policy, for example, periodic backoff policy with constant interval.

As you offer in other issue, solution for my case could be "connection attempt exponential timeout", but i think that this solution costs more developing time than configuring of backoff policy.
https://github.com/grpc/grpc-java/issues/1943

As you could be seen this is my first PR in this project, so if you see some reasons to refuse this feature, please tell me. I want to understand, what negative aspects of configuring backoff policy.

antonbashir:
Same issue about my case:
https://github.com/grpc/grpc-java/issues/4327

> Why does the server restart? - deployment process of new service version. 

There is no any Network or Application error, just graceful server shutdown.

> And what type of server is it such that there is only one instance?

No, no, not one instance. There are two OpenShift's pods. As balancer we use OpenShift/Kubernetes Service-Name Load balancing and "canary" redeployment policy. Anyway, i don't sure that GRPC users always have clustered environment, despite of that single-instance mode is not recommended.

> Without the backoff, clients will reconnect rapidly to the server and overload it. The backoff protects the server.

I am absolutely agree with this fact. But in this PR i don't try to reduce backoff default interval or changing some backoff logic. I just offered to make this parameter configurable same as load balancing policy or deadline timeout parameters. As default policy use current Exponential policy with default parameters. Also, as you could see in last version of source code, there is comment: 

> // TODO(carl-mastrangelo): Allow clients to pass this in

I saw this comment and thought that not only me want to make this parameter configurable. 
Also, i want to try some other policy, for example, periodic backoff policy with constant interval.

As you offer in other issue, solution for my case could be "connection attempt exponential timeout", but i think that this solution costs more developing time than configuring of backoff policy.
https://github.com/grpc/grpc-java/issues/1943

As you could be seen this is my first PR in this project, so if you see some reasons to refuse this feature, please tell me. I want to understand, what negative aspects of configuring backoff policy.

ejona86:
> Same issue about my case: #4327

In that issue it says:
> the kubernetes service forwards the request to the old pod (IP: 192.168.0.10) and the client-pod has to wait for a socket timeout around 2min before it can reconnect

Are you needing to wait 2 minutes? That 2 minutes is unrelated to the backoff policy. It sounded like you were waiting single-digit seconds, which sounds more like the backoff policy.

With the backoff policy, you should never wait more than ⅔ the amount of time the server was down. So if the server was down 30 seconds, the _maximum_ wait time would be 20 seconds after the server is started. That normally works amazingly well, but I can understand how waiting those 20 seconds would be annoying to the development cycle.

I'll note that there are APIs that let you react to events that gRPC wouldn't know about, like `ManagedChannel.resetConnectBackoff()` for mobile use-cases. If the client was "pinged" when the server was started that API could be used. But that doesn't seem like a good fit in this case.

Maybe we should have a "dev mode" that uses different limits. We'd mainly need to discourage it being used in production. Limiting the solution to development makes it a lot easier to solve, as there are fewer risks during development.

> Anyway, i don't sure that GRPC users always have clustered environment

I wasn't criticizing. I was just trying to understand your environment/use-case better. Often with this sort of thing there is something "different" about the environment.

>  TODO(carl-mastrangelo): Allow clients to pass this in

Yeah... Carl and I disagreed about this. I wouldn't be surprised to have this _eventually_, but it needs a strong use-case in my mind. I've seen people wanting to be able to control the backoff behavior to workaround other issues in production, and most people don't have the experience to know "a max backoff of 10ms" is a bad idea. I'll also note that TODO is from at 2015, and even Carl may no longer agree as his viewpoint shifted from "as an advanced application developer I want..." to something more like "as a library developer with users that can't be trusted...".

I'll also note that the code that calls the provider is assuming it is well-behaved, which is no longer the case if we expose it. We'd need to improve exception-catching and audit the threading model (which thread it is executed on). It is more likely, in my mind, to let a user configure the _parameters_, to avoid the user being able to inject code.

>  As you offer in other issue, solution for my case could be "connection attempt exponential timeout"... #1943

That's actually something different. The important part of that is the _timeout_. It causes _errors_ to occur sooner. If that would fix your issue, then this PR would _not_; they fix separate issues.

ejona86:
@antonbashir, closing the PR is fine, but did you want to open an issue so that we could discuss ways to resolve your use-case?

PR #6876, opened 2020-03-31
xds: import opencensus proto and revert earlier change that used dependency

sanjaypujare:
this reverts the earlier change in the PR https://github.com/grpc/grpc-java/pull/6871 that used dependency `io.opencensus:opencensus-proto:0.2.0`. The problem with this dependency is it downloads too many files (we need only one proto file) and mainly that it introduces conflicts. This artifact itself has dependencies back on grpc artifacts (`grpc-protobuf` and `grpc-stub`) and although this seems to be okay on command line builds, it shows up in IDE as: 
```
Could not resolve all dependencies for configuration ':grpc-xds:compile'.
Conflict(s) found for the following module(s):
  - io.grpc:grpc-protobuf between versions 1.29.0-SNAPSHOT and 1.14.0  
  - io.grpc:grpc-stub between versions 1.29.0-SNAPSHOT and 1.14.0
```
So it is best to just import the needed proto file similar to other 3rd party protos.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -149,7 +149,6 @@ subprojects {
             opencensus_contrib_grpc_metrics: "io.opencensus:opencensus-contrib-grpc-metrics:${opencensusVersion}",
             opencensus_impl: "io.opencensus:opencensus-impl:${opencensusVersion}",
             opencensus_impl_lite: "io.opencensus:opencensus-impl-lite:${opencensusVersion}",
-            opencensus_proto: "io.opencensus:opencensus-proto:0.2.0",
             instrumentation_api: 'com.google.instrumentation:instrumentation-api:0.4.3',
             perfmark: 'io.perfmark:perfmark-api:0.19.0',
             protobuf: "com.google.protobuf:protobuf-java:${protobufVersion}",
--- a/xds/build.gradle
+++ b/xds/build.gradle
@@ -23,8 +23,7 @@ dependencies {
             project(':grpc-stub'),
             project(':grpc-core'),
             project(':grpc-services'),
-            project(path: ':grpc-alts', configuration: 'shadow'),
-            libraries.opencensus_proto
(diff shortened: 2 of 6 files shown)


dapengzhang0:
Is the conflict only an issue of the IDE?

dapengzhang0:
@ejona86 any though on what is the best solution ?

sanjaypujare:
> Is the conflict only an issue of the IDE?

So far it looks like. The command line builds all succeeded but my IDE projects on 2 different systems failed (in one case silently/mysteriously) and IDE was unusable. With this PR the IDEs are back to behaving normally so they are usable.

In any case I think this is the better thing to do because of this weird circular dependency and the fact that we just need one proto file. I suspect more people are going to complain with the dependency. Also if Envoy proto is fixed (issue https://github.com/envoyproxy/envoy/issues/10576) we can remove this import as well.

sanjaypujare:
Talked to @ejona86 offline. Instead of doing the import (as in this PR) the better thing to do is to fix the dependency issues in `opencensus-proto` by excluding them (similar to how we have done in `grpc-census` or elsewhere).

I might fix this PR or close this and open a brand new one.

ejona86:
That error is normal. We [turn it on explicitly](https://github.com/grpc/grpc-java/blob/68391e4d1b0d9ae5ed7277eb8b2920e99f2dc1b0/build.gradle#L200). It just tells us that our dependencies may be misinterpreted by Maven, since Maven is poor at transitive dependencies. It just needs the excludes like elsewhere; virtually all the excludes we have today are because of this checker.

The weird thing is that the build doesn't fail on the command line. I've poked around at this and can't figure out why it isn't failing during normal builds. `gradle dependencies` clearly shows the conflict. Whatever.

sanjaypujare:
Closing this and creating a new PR with "excludes" added to resolve the dependencies conflict issue.

PR #6883, opened 2020-04-01
rls: lb policy config object & ChildPolicyReportingHelper

creamsoup:
some of them are public because it needs to be accessed from Lb/LbProvider.


rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
```diff
@@ -0,0 +1,372 @@
...
+  }
+
+  /**
+   * ChildPolicyWrapper is a wrapper class for child load balancing policy with associated helper /
+   * utility classes to manage the child policy.
+   */
+  static final class ChildPolicyWrapper implements Closeable {
+
+    @VisibleForTesting
+    static final Map<String /* target */, RefCountedChildPolicyWrapper> childPolicyMap =
```
dapengzhang0, line 220: This map is shared for the whole JVM for all channels to totally different servers. This worries me.  Also the map needs synchronization.

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
```diff
@@ -0,0 +1,347 @@
...
+          effectiveLbProvider = provider;
+          effectiveChildPolicy = Collections.unmodifiableMap(childPolicy);
+          break;
+        }
+        policyTried.add(policyName);
+      }
+      checkState(
+          effectiveChildPolicy != null,
+          "no valid childPolicy found, policy tried: %s", policyTried);
+      checkArgument(
```
dapengzhang0, line 137: Seems you can just implement `LrsLoadBalancerProvider. parseLoadBalancingPolicyConfig ()` by merging master.  Return ConfigOrError instead of throw.

creamsoup:
somehow github doesn't allow me to reply
>Seems you can just implement LrsLoadBalancerProvider.parseLoadBalancingPolicyConfig() by merging master. Return ConfigOrError instead of throw.

i am not following, can you explain little more detail?

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
```diff
@@ -0,0 +1,372 @@
...
+   */
+  static final class ChildPolicyWrapper implements Closeable {
+
+    @VisibleForTesting
+    static final Map<String /* target */, RefCountedChildPolicyWrapper> childPolicyMap =
+        new HashMap<>();
+
+    private final String target;
+    @Nullable
+    private ChildLoadBalancingPolicy childPolicy;
```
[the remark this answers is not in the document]
dapengzhang0: Yes, it would be definitely better to make it final unless there is restriction such as the helper and childPolicy are unknown when it has to be instantiated. Would that be difficault?

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
```diff
@@ -344,4 +348,14 @@ public String toString() {
...
     }
   }
+
+  /** Exception thrown when attempting to parse child policy encountered parsing issue. */
+  public static final class InvalidChildPolicyConfigException extends Exception {
+
+    private static final long serialVersionUID = 0L;
+
+    public InvalidChildPolicyConfigException(String message) {
+      super(message);
```
dapengzhang0, line 365: Maybe you can turn off writableStackTrace.

creamsoup: done.

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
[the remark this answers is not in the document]
creamsoup: i had to bring in the other helper due to circular dependency if it just accept helper in actual lb impl. PTAL

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    @Override
+    @Nullable
+    public ChildPolicyWrapper returnObject(Object object) {
+      checkState(
+          childPolicyWrapper == object,
+          "returned object doesn't match the pooled childPolicyWrapper");
+      long newCnt = refCnt.decrementAndGet();
+      if (newCnt == 0) {
```
dapengzhang0, line 335: What's the expected usecase here? `getObject()` once and `returnObject()` twice()?

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
```diff
@@ -0,0 +1,369 @@
...
+        ConnectivityStateInfo.forNonError(ConnectivityState.IDLE);
+    private SubchannelPicker picker;
+
+    public ChildPolicyWrapper(
+        String target,
+        ChildLoadBalancerHelperProvider childLbHelperProvider,
+        @Nullable  ChildLbStatusListener childLbStatusListener) {
+      this.target = target;
+      this.helper =
+          new ChildPolicyReportingHelper(childLbHelperProvider, this, childLbStatusListener);
```
dapengzhang0, line 250: This circular dependency looks weird, because `ChildPolicyWrapper` is consumed by `new ChildPolicyReportingHelper()` before instantiation is complete.

Making `ChildPolicyReportingHelperImpl` an inner (non-static) class of `ChildPolicyWrapper` might be more natural.

rls/src/main/java/io/grpc/rls/internal/ChildPolicyReportingHelper.java
```diff
@@ -0,0 +1,105 @@
...
+ * The governing picker will be reported back to the parent load-balancer.
+ */
+final class ChildPolicyReportingHelper extends ForwardingLoadBalancerHelper {
+
+  private final ChildLoadBalancerHelper delegate;
+  private final ChildPolicyWrapper childPolicyWrapper;
+  @Nullable
+  private final ChildLbStatusListener listener;
+
+  ChildPolicyReportingHelper(
```
dapengzhang0, line 47: Not used?

creamsoup: not used yet. this will be used by fallback channel.

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
creamsoup: yes i was wondering same thing.

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
creamsoup: this handles when the object is just released, free up the reference.

dapengzhang0: If you call x times `getObject()`, you need call x+1 times of `returnObject()` to get here, is that the contract?

creamsoup: `create` itself is counted as `getObject()` since the object is passed in.

dapengzhang0: > create itself is counted as getObject() 
But you can not use the object unless you call `getObject()` explicitly.  The API contract looks strange.  If you just need to you the object once, you have to explicitly call  `getObject()` once, and you need `returnObject()` twice to get here.

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
```diff
@@ -0,0 +1,446 @@
...
+    private final String target;
+    private final ChildPolicyReportingHelper helper;
+    private ConnectivityStateInfo connectivityStateInfo =
+        ConnectivityStateInfo.forNonError(ConnectivityState.IDLE);
+    private SubchannelPicker picker;
+
+    public ChildPolicyWrapper(
+        String target,
+        ChildLoadBalancerHelperProvider childLbHelperProvider,
+        @Nullable  ChildLbStatusListener childLbStatusListener) {
```
dapengzhang0, line 253: There is extra space.

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
```diff
@@ -0,0 +1,446 @@
...
+import io.grpc.LoadBalancer.CreateSubchannelArgs;
+import io.grpc.LoadBalancer.Helper;
+import io.grpc.LoadBalancer.Subchannel;
+import io.grpc.LoadBalancer.SubchannelPicker;
+import io.grpc.LoadBalancer.SubchannelStateListener;
+import io.grpc.LoadBalancerProvider;
+import io.grpc.LoadBalancerRegistry;
+import io.grpc.internal.ObjectPool;
+import io.grpc.rls.internal.ChildLoadBalancerHelper.ChildLoadBalancerHelperProvider;
+import io.grpc.rls.internal.LbPolicyConfiguration.ChildPolicyWrapper.ChildLbStatusListener;
```
dapengzhang0, line 36: nit: The listener interface is nested too deep. Can be one level up.

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
```diff
@@ -0,0 +1,446 @@
...
+          childPolicyWrapper != null,
+          "cannot return already released ChildPolicyWrapper, this is possibly a bug.");
+      checkState(
+          childPolicyWrapper == object,
+          "returned object doesn't match the pooled childPolicyWrapper");
+      long newCnt = refCnt.decrementAndGet();
+      if (newCnt == 0) {
+        childPolicyWrapper = null;
+      }
+      return childPolicyWrapper;
```
dapengzhang0, line 415: This API should always return null.
https://github.com/grpc/grpc-java/blob/v1.28.1/core/src/main/java/io/grpc/internal/ObjectPool.java#L32-L36

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
creamsoup: why it cannot use the object before getObject()? anyways, okay it depends on what is the focus. focused more on the "RefCounted" it makes more sense to have it as 1, if you focused more on "ObjectPool" part starting with 0 makes more sense. refCount = 0 is released state (ref counted bytebuf starts with 1 for example), but for the sake of consistency i can follow the other convention, also it implements ObjectPool makes me leaning toward that.

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
creamsoup: good catch!

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
dapengzhang0: I see, it is used directly for the first time without calling `getObject()`
```java
RefCountedChildPolicyWrapper wrapper = RefCountedChildPolicyWrapper.of(childPolicyWrapper);
childPolicyMap.put(target, wrapper);
return childPolicyWrapper;
```

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
creamsoup: done.

rls/src/main/java/io/grpc/rls/internal/LbPolicyConfiguration.java
dapengzhang0: Might need introduce something like `isReleased()`

creamsoup: tried to avoid additional check, but this sounds better.

PR #2494, opened 2016-12-07
all: Add more projects to the all super-project

ejona86:
This adds to the published JavaDoc and reported code coverage.

PR #6888, opened 2020-04-01
all: Add grpc-xds

ejona86:
This will fix reporting of xDS code coverage.

CC @voidzcy

### DIFF
--- a/all/build.gradle
+++ b/all/build.gradle
@@ -18,6 +18,7 @@ def subprojects = [
     project(':grpc-protobuf-lite'),
     project(':grpc-stub'),
     project(':grpc-testing'),
+    project(':grpc-xds'),
 ]
 
 for (subproject in rootProject.subprojects) {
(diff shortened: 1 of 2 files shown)


all/build.gradle
```diff
@@ -18,6 +18,7 @@ def subprojects = [
     project(':grpc-protobuf-lite'),
     project(':grpc-stub'),
     project(':grpc-testing'),
+    project(':grpc-xds'),
```
dapengzhang0, line 21: Is this collection used for both javadoc and jacoco?

ejona86: Yes. It also adds a dependency in the pom to grpc-all.

ejona86:
Those are part of #2494, which still needs some love fixing up the jacoco report. (xds was a bit easier since generated code was in non-grpc packages.)

ejona86:
Err.. I guess I hadn't planned to for grpc-testing-proto, but you're right, we should for the javadoc.

PR #6889, opened 2020-04-01
buildscripts: use xds-test-server image

ericgribkoff:
Added in https://github.com/grpc/grpc/pull/22538

### DIFF
--- a/buildscripts/kokoro/xds.sh
+++ b/buildscripts/kokoro/xds.sh
@@ -24,6 +24,8 @@ JAVA_OPTS=-Djava.util.logging.config.file=grpc-java/buildscripts/xds_logging.pro
   python3 grpc/tools/run_tests/run_xds_tests.py \
     --test_case=all \
     --project_id=grpc-testing \
+    --source_image=projects/grpc-testing/global/images/xds-test-server \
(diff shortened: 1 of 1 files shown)


buildscripts/kokoro/xds.sh
```diff
@@ -24,6 +24,8 @@ JAVA_OPTS=-Djava.util.logging.config.file=grpc-java/buildscripts/xds_logging.pro
   python3 grpc/tools/run_tests/run_xds_tests.py \
     --test_case=all \
     --project_id=grpc-testing \
+    --source_image=projects/grpc-testing/global/images/xds-test-server \
```
dapengzhang0, line 27: A question at line 20: "${branch}" does not work well for manual invocation if grpc-java is a custom branch or commit hash and grpc is master. Why not just use 'master'?

ericgribkoff: Just always using master would mean that changes to the test framework at head on grpc/grpc could break tests running on the release branches, e.g. v1.28.x. This is undesirable. There was some discussion around obtaining the corresponding branch at https://github.com/grpc/grpc-go/pull/3456 (and referenced from https://github.com/grpc/grpc-java/pull/6828).

Since this is in the kokoro directory I didn't put much effort into making this robust to manual runs.

PR #6890, opened 2020-04-01
xds: ensure we shutdown XdsClientImpl when XDS Server is shut down

sanjaypujare:


xds/src/main/java/io/grpc/xds/internal/sds/XdsServer.java
```diff
@@ -0,0 +1,116 @@
...
+ * Wraps a {@link Server} delegate and {@link XdsClientWrapperForServerSds} and intercepts
+ * {@link Server#shutdown()} etc to properly shut down the {@link XdsClientWrapperForServerSds}
+ * object.
+ */
+@VisibleForTesting
+public final class XdsServer extends Server {
+  private final Server delegate;
+  @Nullable private final XdsClientWrapperForServerSds xdsClientWrapperForServerSds;
+
+  XdsServer(Server delegate, XdsClientWrapperForServerSds xdsClientWrapperForServerSds) {
```
creamsoup, line 42: nit: add `@Nullable` to the param xdsClientWrapperForServerSds. but, i don't see why it wants null? if this is null the caller can use delegate Server. factory method returns Server type can easily achieve that if caller doesn't want to check null.

xds/src/main/java/io/grpc/xds/internal/sds/XdsServer.java
```diff
@@ -0,0 +1,116 @@
...
+      xdsClientWrapperForServerSds.shutdown();
+    }
+    delegate.shutdown();
+    return this;
+  }
+
+  @Override
+  public Server shutdownNow() {
+    if (xdsClientWrapperForServerSds != null) {
+      xdsClientWrapperForServerSds.shutdown();
```
creamsoup, line 66: should this call `shutdownNow`?

xds/src/main/java/io/grpc/xds/internal/sds/XdsServer.java
```diff
@@ -0,0 +1,116 @@
...
+    return delegate.isShutdown();
+  }
+
+  @Override
+  public boolean isTerminated() {
+    return delegate.isTerminated();
+  }
+
+  @Override
+  public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
```
creamsoup, line 83: those should do awaitTermination on sds andThen server.

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -264,12 +266,24 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause)
...
       } catch (Exception e) {
-        logger.log(Level.WARNING, "Exception while creating the xDS client", e);
+        logger.log(Level.FINE, "Fallback to plaintext due to exception", e);
         localXdsClientWrapperForServerSds = null;
       }
       this.xdsClientWrapperForServerSds = localXdsClientWrapperForServerSds;
     }
 
+    @VisibleForTesting
+    public ServerSdsProtocolNegotiator(
```
creamsoup, line 276: nit: it would be nice if the other constructor is calling this one (or could be a factory method).

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -264,12 +266,24 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause)
...
+        DownstreamTlsContext downstreamTlsContext,
+        XdsClientWrapperForServerSds xdsClientWrapperForServerSds) {
+      this.downstreamTlsContext = downstreamTlsContext;
```
creamsoup, line 279: nit: this field can be final

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -264,12 +266,24 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause)
...
+      this.xdsClientWrapperForServerSds = xdsClientWrapperForServerSds;
```
creamsoup, line 280: nit: null check here seems reasonable

xds/src/test/java/io/grpc/xds/XdsServerBuilderTest.java
```diff
@@ -14,17 +14,23 @@
  * limitations under the License.
  */
 
-package io.grpc.xds.internal.sds;
```
creamsoup, line 17: if you keep it in the io.grpc.xds.internal.sds, the ServerSdsProtocolNegotiator  doesn't need to expose it as public. also, the Builder is anyways in the sds package it seems better place to be.

sanjaypujare: Yes and I would have liked that. But `XdsClient`, `XdsClientImpl`, and `XdsClient.shutdown` are package private which causes problem for this test unless I make them all public. So making those things public just for a test in SDS didn't sound appealing to me.

xds/src/main/java/io/grpc/xds/internal/sds/XdsServer.java
sanjaypujare: There is no `awaitTermination` in `XdsClient` (`XdsClientImpl`) which is I think what you are suggesting. It can be plumbed and added here but it doesn't seem necessary. What does the user gain by waiting for `XdsClient` to shut down?

xds/src/main/java/io/grpc/xds/internal/sds/XdsServer.java
sanjaypujare: good  catch.

creamsoup: oops actually XdsClient only has shutdown ignore my comment.

xds/src/main/java/io/grpc/xds/internal/sds/XdsServer.java
creamsoup: same here. XdsClient doesn't have it.

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
sanjaypujare: with the refactoring both params can be null (plaintext case) so no null checks

xds/src/test/java/io/grpc/xds/XdsServerBuilderTest.java
sanjaypujare: No change based on my comment.

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
creamsoup: then add `@Nullable`, if both are null isn't it plaintext?

xds/src/main/java/io/grpc/xds/internal/sds/XdsServer.java
```diff
@@ -0,0 +1,121 @@
...
+ * Wraps a {@link Server} delegate and {@link XdsClientWrapperForServerSds} and intercepts
+ * {@link Server#shutdown()} etc to properly shut down the {@link XdsClientWrapperForServerSds}
+ * object.
+ */
+final class XdsServer extends Server {
+  private final Server delegate;
+  private final XdsClientWrapperForServerSds xdsClientWrapperForServerSds;
+
+  static Server newInstance(
+      Server delegate, XdsClientWrapperForServerSds xdsClientWrapperForServerSds) {
```
creamsoup, line 40: add @Nullable for 2nd param. overall methods returns null should also be marked with the annotation. can you check other methods/params/instance variables for the classes in this PR?

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -251,23 +252,38 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause)
...
+        int port, SynchronizationContext syncContext) {
       try {
-        localXdsClientWrapperForServerSds =
-            XdsClientWrapperForServerSds.newInstance(port, Bootstrapper.getInstance(), syncContext);
+        return XdsClientWrapperForServerSds.newInstance(
+            port, Bootstrapper.getInstance(), syncContext);
       } catch (Exception e) {
-        logger.log(Level.WARNING, "Exception while creating the xDS client", e);
-        localXdsClientWrapperForServerSds = null;
+        logger.log(Level.FINE, "Fallback to plaintext due to exception", e);
```
creamsoup, line 272: why we do this instead of using plaintext when creating the protocol negotiator?

sanjaypujare: So for the logic, it does end up creating the plaintext protocol negotiator if you follow this flow: `null` for `XdsClientWrapperForServerSds` causes `SdsProtocolNegotiators.HandlerPickerHandler` on line 335 to replace with a plaintext negotiator. Is that what you were asking or something else?

creamsoup: if we know it will be plaintext, the factory method `SdsProtocolNegotiators#serverProtocolNegotiator` can just return the simple plaintext PN not in the protocol negotiation handler if possible. we want to minimize conditional logic in each protocol to easy to maintain and debug. it can be very hard due to the async nature of handler.

sanjaypujare: I think it can go both ways. For the `HandlerPickerHandler` you wanted (in a previous PR) proper handling of `ProtocolNegotiationEvent` to pick the actual Handler (plaintext or `ServerSdsHandler`)  based on `downstreamTlsContext` being empty. So that logic is centralized in one place even if the decision is delayed in the case you point out.

What you are saying can be done but will be a non-trivial change and I will have to create tests for it. Since this PR is really about server shutdown I am thinking of creating a TODO for me to do in another PR. What do you think?

xds/src/main/java/io/grpc/xds/internal/sds/XdsServerBuilder.java
```diff
@@ -184,9 +185,15 @@ void panic(final Throwable t) {
    */
   @VisibleForTesting
   public Server createXdsServer(
-      SdsProtocolNegotiators.ServerSdsProtocolNegotiator serverSdsProtocolNegotiator) {
-    delegate.protocolNegotiator(serverSdsProtocolNegotiator);
-    return XdsServer.newInstance(
-        delegate.build(), serverSdsProtocolNegotiator.getXdsClientWrapperForServerSds());
+          InternalProtocolNegotiator.ProtocolNegotiator serverProtocolNegotiator) {
```
creamsoup, line 188: indentation is wrong.

xds/src/main/java/io/grpc/xds/internal/sds/XdsServer.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+
+/**
+ * Wraps a {@link Server} delegate and {@link XdsClientWrapperForServerSds} and intercepts
+ * {@link Server#shutdown()} etc to properly shut down the {@link XdsClientWrapperForServerSds}
+ * object.
+ */
+final class XdsServer extends Server {
```
creamsoup, line 36: what makes it XdsServer?

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -77,7 +79,8 @@ public static ProtocolNegotiatorFactory clientProtocolNegotiatorFactory(
   public static ProtocolNegotiator serverProtocolNegotiator(
       @Nullable DownstreamTlsContext downstreamTlsContext, int port,
       SynchronizationContext syncContext) {
-    return new ServerSdsProtocolNegotiator(downstreamTlsContext, port, syncContext);
+    return ServerSdsProtocolNegotiator.createServerNegotiator(
+            downstreamTlsContext, port, syncContext);
```
creamsoup, line 83: wrong indentation

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -251,23 +254,47 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause)
     }
   }
 
-  private static final class ServerSdsProtocolNegotiator implements ProtocolNegotiator {
+  @VisibleForTesting
+  public static final class ServerSdsProtocolNegotiator implements ProtocolNegotiator {
 
-    private DownstreamTlsContext downstreamTlsContext;
+    private final DownstreamTlsContext downstreamTlsContext;
```
creamsoup, line 260: looks like those 2 can be null.

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -251,23 +254,47 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause)
...
-        DownstreamTlsContext downstreamTlsContext, int port, SynchronizationContext syncContext) {
-      this.downstreamTlsContext = downstreamTlsContext;
-      XdsClientWrapperForServerSds localXdsClientWrapperForServerSds;
+    static ProtocolNegotiator createServerNegotiator(
+        @Nullable DownstreamTlsContext downstreamTlsContext,
+        int port,
+        SynchronizationContext syncContext) {
+      XdsClientWrapperForServerSds xdsClientWrapperForServerSds =
+          getXdsClientWrapperForServerSds(port, syncContext);
+      return (xdsClientWrapperForServerSds == null && downstreamTlsContext == null)
```
creamsoup, line 269: can we log when it downgrades to plaintext?

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -251,23 +254,47 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause)
...
+    }
+
+    private static XdsClientWrapperForServerSds getXdsClientWrapperForServerSds(
+        int port, SynchronizationContext syncContext) {
       try {
-        localXdsClientWrapperForServerSds =
-            XdsClientWrapperForServerSds.newInstance(port, Bootstrapper.getInstance(), syncContext);
+        return XdsClientWrapperForServerSds.newInstance(
+            port, Bootstrapper.getInstance(), syncContext);
       } catch (Exception e) {
```
creamsoup, line 279: (i know we discussed before) this exception seems too aggressive. a bit worried that this can make it very hard to debug

xds/src/main/java/io/grpc/xds/internal/sds/XdsServer.java
```diff
@@ -0,0 +1,122 @@
...
+  private final Server delegate;
+  private final XdsClientWrapperForServerSds xdsClientWrapperForServerSds;
+
+  static Server newInstance(
```
creamsoup, line 40: newInstance is misleading especially for server. it sounds like it will spin up a new server.

xds/src/main/java/io/grpc/xds/internal/sds/XdsServer.java
```diff
@@ -0,0 +1,122 @@
...
+
+import java.io.IOException;
+import java.net.SocketAddress;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+
+/**
+ * Wraps a {@link Server} delegate and {@link XdsClientWrapperForServerSds} and intercepts
+ * {@link Server#shutdown()} etc to properly shut down the {@link XdsClientWrapperForServerSds}
```
creamsoup, line 33: when do we need to have  `XdsClientWrapperForServerSds`? it looks like it is only required during ProtocolNegotiation. if this is the case, we can rely on the PN#close.

xds/src/test/java/io/grpc/xds/internal/sds/SdsProtocolNegotiatorsTest.java
```diff
@@ -257,6 +258,14 @@ public void clientSdsProtocolNegotiatorNewHandler_fireProtocolNegotiationEvent()
     assertTrue(channel.isOpen());
   }
 
+  @Test
+  public void serverSdsProtocolNegotiator_passNulls_expectPlaintext() {
+    InternalProtocolNegotiator.ProtocolNegotiator protocolNegotiator =
+        SdsProtocolNegotiators.ServerSdsProtocolNegotiator.createServerNegotiator(null, 7000,
+           null);
```
creamsoup, line 265:
```suggestion
            null);
```

xds/src/main/java/io/grpc/xds/internal/sds/XdsServerBuilder.java
```diff
@@ -171,9 +173,27 @@ void panic(final Throwable t) {
               panicMode = true;
             }
           });
-    delegate.protocolNegotiator(
+    InternalProtocolNegotiator.ProtocolNegotiator serverProtocolNegotiator =
         SdsProtocolNegotiators.serverProtocolNegotiator(
-            this.downstreamTlsContext, port, syncContext));
-    return delegate.build();
+            this.downstreamTlsContext, port, syncContext);
+    return createXdsServer(serverProtocolNegotiator);
```
creamsoup, line 179: it is weird that this build() doesn't return XdsServer even though it is XdsServerBuilder. Although it is questionable that we need that XdsServer class or not.

xds/src/main/java/io/grpc/xds/internal/sds/XdsServer.java
sanjaypujare: Okay the name is misleading but the builder was named XdsServerBuilder so this ended up as XdsServer. It is basically a wrapper for an actual `io.grpc.Server` with a reference to the accompanying `XdsClient` that watches for LDS updates. It is package private so not sure if it's worth renaming.

xds/src/main/java/io/grpc/xds/internal/sds/XdsServer.java
sanjaypujare: The `XdsClientWrapperForServerSds` is needed for the life of the server. Note that every incoming connection to the server results into TLS handshake and protocol neg and should use the latest cert/key material pointed by the tlsContext. Why close is on PN#close?

xds/src/main/java/io/grpc/xds/internal/sds/XdsServerBuilder.java
sanjaypujare: As mentioned above, we need to keep the `XdsClientWrapperForServerSds` instance thru the life of the server and shut down when the server shuts down. It can always return XdsServer even if not needed just to avoid the "weirdness"?

xds/src/main/java/io/grpc/xds/internal/sds/XdsServer.java
creamsoup: okay. that's why i am confused. the ownership of XdsClientWrapperForServerSds is weird. XdsServer should own the XdsClientWrapperForServerSds, and PN should borrow it from the XdsServer. because PN only exists during protocol negotiation. 
but, this code create it in PN. and the ownership is transferred to XdsServer. if we change PN to accept XdsClientWrapperForServerSds from server side the code will be much easier to follow. also, PN doesn't need to create/handle exception of the XdsClientWrapperForServerSds.

sanjaypujare: Just realized the PN exists for the whole lifetime of the server: you set it on the server-builder with `NettyServerBuilder.protocolNegotiator()` and the server calls `newHandler` every time it needs a handshaker. So you were right: I can keep the `XdsClientWrapperForServerSds` with the PN and inside PN#close I can call `XdsClientWrapperForServerSds.shutdown` and I probably don't even need `XdsServer` then.

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -77,7 +79,14 @@ public static ProtocolNegotiatorFactory clientProtocolNegotiatorFactory(
   public static ProtocolNegotiator serverProtocolNegotiator(
       @Nullable DownstreamTlsContext downstreamTlsContext, int port,
       SynchronizationContext syncContext) {
-    return new ServerSdsProtocolNegotiator(downstreamTlsContext, port, syncContext);
+    XdsClientWrapperForServerSds xdsClientWrapperForServerSds =
+        ServerSdsProtocolNegotiator.getXdsClientWrapperForServerSds(port, syncContext);
+    if (xdsClientWrapperForServerSds == null && downstreamTlsContext == null) {
+      logger.log(Level.INFO, "Fallback to plaintext for server at port {0}", port);
```
creamsoup, line 85: the port is not used if the xdsClientWrapperForServerSds is null, right?

xds/src/test/java/io/grpc/xds/XdsServerBuilderTest.java
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2019 The gRPC Authors
```
creamsoup, line 2:
```suggestion
 * Copyright 2020 The gRPC Authors
```

sanjaypujare: The file was moved from under sds to this new location and a test was added. Not sure if the year needs to be updated

creamsoup:
Oh my bad if it is just moved should be 2019. Git says it is added hmm

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
sanjaypujare: But the server is bound to a port e.g. 8080. As a log message it will be useful to point out to the user that server bound to port 8080 fell back to plaintext for whatever reason. If an application creates, say 5 servers out of which 3 fell back to plaintext because of xDS issues but 2 kept TLS/mTLS because they didn't use xDS it will be useful to log the port in the log.

sanjaypujare:
> Oh my bad if it is just moved should be 2019. Git says it is added hmm

It was moved *and* modified. git probably can't detect that

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
creamsoup: this message is very confusing. there are multiple servers, what is the server here?

sanjaypujare: So not saying it is a "grpc xds-based" server is confusing? 

Given
```
Server serverA = XdsServerBuilder.forPort(80).build();
Server serverB = XdsServerBuilder.forPort(81).build();
Server serverC = NettyServerBuilder.forPort(82).build();
```
The port uniquely identifies the server. So port number 80 or 81 will be mentioned in the log message to point to serverA or serverB resp. I don't understand the confusion.

sanjaypujare: Can you suggest what the message should be?

PR #6894, opened 2020-04-02
api: Fix javadoc reference to deprecated method

ST-DDT:
Updates the `ManagedChannelBuilder#nameResolverFactory()`'s javadocs to refer to the new `NameResolverRegistry` instead of the deprecated `NameResolverProvider#providers()` method.

This is also closer to the truth for the default implementations:

https://github.com/grpc/grpc-java/blob/84dd812db86cbfa77b0dda7839ae0ea876135057/core/src/main/java/io/grpc/internal/AbstractManagedChannelImplBuilder.java#L251-L255


creamsoup:
merged, thanks @ST-DDT

PR #6900, opened 2020-04-03
stub: add Blocking StubType to blocking ClientCalls methods.

ran-su:
b/34358463

### DIFF
--- a/stub/src/test/java/io/grpc/stub/ClientCallsTest.java
+++ b/stub/src/test/java/io/grpc/stub/ClientCallsTest.java
@@ -24,7 +24,6 @@ import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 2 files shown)


stub/src/test/java/io/grpc/stub/ClientCallsTest.java
```diff
@@ -203,6 +220,48 @@ public void blockingUnaryCall2_interruptedWaitsForOnClose() throws Exception {
     assertTrue("context not cancelled", methodImpl.observer.isCancelled());
   }
 
+  @Test
+  public void blockingUnaryCall_HasBlockingStubType() {
+    when(mockChannel.newCall(
+        ArgumentMatchers.<MethodDescriptor<Integer, Integer>>any(), any(CallOptions.class)))
+        .thenReturn(mockClientCall);
+    // we only interested in the call options, stop the call by thrown.
+    doThrow(new RuntimeException("intended")).when(mockClientCall).sendMessage(any(Integer.class));
```
zhangkun83, line 229: `unaryBlockingCallFailed()` uses a `NoopClientCall` that calls `listener.onClose()` in its `start()`. That is closer to the reality. We should also keep the same style throughout a file.

ran-su: Done.

PR #6901, opened 2020-04-05
xds: eliminate downstreamTlsContext from XdsServerBuilder

sanjaypujare:


xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,224 @@
+/*
```
creamsoup, line 1: can you make it branched? it is hard to tell what is modified.

xds/src/test/java/io/grpc/xds/internal/sds/SdsProtocolNegotiatorsTest.java
```diff
@@ -217,7 +231,7 @@ public void serverSdsHandler_addLast() throws IOException {
   @Test
   public void serverSdsHandler_nullTlsContext_expectPlaintext() throws IOException {
     SdsProtocolNegotiators.HandlerPickerHandler handlerPickerHandler =
-            new SdsProtocolNegotiators.HandlerPickerHandler(grpcHandler, null, null);
+            new SdsProtocolNegotiators.HandlerPickerHandler(grpcHandler, null);
```
creamsoup, line 234: seems like wrong indentation

xds/src/test/java/io/grpc/xds/internal/sds/SecretVolumeSslContextProviderTest.java
```diff
@@ -43,11 +43,11 @@
 @RunWith(JUnit4.class)
 public class SecretVolumeSslContextProviderTest {
 
-  private static final String SERVER_1_PEM_FILE = "server1.pem";
-  private static final String SERVER_1_KEY_FILE = "server1.key";
-  private static final String CLIENT_PEM_FILE = "client.pem";
-  private static final String CLIENT_KEY_FILE = "client.key";
-  private static final String CA_PEM_FILE = "ca.pem";
+  public static final String SERVER_1_PEM_FILE = "server1.pem";
```
creamsoup, line 46: this is getting disorganized, can you make a utility class including other public methods used in multiple tests?

xds/src/test/java/io/grpc/xds/internal/sds/SecretVolumeSslContextProviderTest.java
```diff
@@ -430,6 +434,19 @@ static DownstreamTlsContext buildDownstreamTlsContextFromFilenames(
    */
   public static UpstreamTlsContext buildUpstreamTlsContextFromFilenames(
       String privateKey, String certChain, String trustCa) {
+    try {
+      if (certChain != null) {
```
creamsoup, line 438: `@Nullable` to params?

xds/src/test/java/io/grpc/xds/internal/sds/SdsProtocolNegotiatorsTest.java
```diff
@@ -261,8 +275,7 @@ public void clientSdsProtocolNegotiatorNewHandler_fireProtocolNegotiationEvent()
   @Test
   public void serverSdsProtocolNegotiator_passNulls_expectPlaintext() {
```
creamsoup, line 276: passNull? what is null?

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -106,29 +107,28 @@ public static XdsClientWrapperForServerSds newInstance(
...
-                "Setting myListener from ConfigUpdate listener :{0}",
-                update.getListener().toString());
-            curListener = update.getListener();
-          }
+    this.listenerWatcher = new XdsClient.ListenerWatcher() {
+      @Override
+      public void onListenerChanged(XdsClient.ListenerUpdate update) {
+        logger.log(
+            Level.INFO,
+            "Setting myListener from ConfigUpdate listener :{0}",
```
creamsoup, line 115:
```suggestion
            "Setting myListener from ConfigUpdate listener: {0}",
```

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -106,29 +107,28 @@ public static XdsClientWrapperForServerSds newInstance(
...
-        });
+      @Override
+      public void onError(Status error) {
+        // In order to distinguish between IO error and resource not found, set curListener
+        // to null in case of NOT_FOUND
+        if (error.getCode().equals(Status.Code.NOT_FOUND)) {
+          curListener = null;
+        }
+        // TODO(sanjaypujare): Implement logic for other cases based on final design.
+        logger.log(Level.SEVERE, "ListenerWatcher in XdsClientWrapperForServerSds:{0}", error);
```
creamsoup, line 128:
```suggestion
        logger.log(Level.SEVERE, "ListenerWatcher in XdsClientWrapperForServerSds: {0}", error);
```

xds/src/main/java/io/grpc/xds/internal/sds/XdsServerBuilder.java
```diff
@@ -174,7 +161,7 @@ void panic(final Throwable t) {
           });
     InternalProtocolNegotiator.ProtocolNegotiator serverProtocolNegotiator =
         SdsProtocolNegotiators.serverProtocolNegotiator(
-            this.downstreamTlsContext, port, syncContext);
+            port, syncContext);
```
creamsoup, line 164: this looks like it can be 1 line.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,224 @@
...
+    EnvoyServerProtoData.Listener listener =
+        new EnvoyServerProtoData.Listener(name, address, Arrays.asList(filterChain1));
+    return listener;
+  }
+
+  private void buildClientAndTest(
+      UpstreamTlsContext upstreamTlsContext,
+      String overrideAuthority,
+      String requestMessage,
+      int serverPort) {
```
creamsoup, line 191: why passing port if it can be from instance variable?

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,224 @@
...
+        new EnvoyServerProtoData.FilterChain(filterChainMatch, tlsContext);
+    EnvoyServerProtoData.Listener listener =
+        new EnvoyServerProtoData.Listener(name, address, Arrays.asList(filterChain1));
+    return listener;
+  }
+
+  private void buildClientAndTest(
+      UpstreamTlsContext upstreamTlsContext,
+      String overrideAuthority,
+      String requestMessage,
```
creamsoup, line 190: for this test, this param is meaningless.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,224 @@
...
+  /** Bad/untrusted server certs. */
+  private static final String BAD_SERVER_PEM_FILE = "badserver.pem";
+  private static final String BAD_SERVER_KEY_FILE = "badserver.key";
+
+  @Rule public final GrpcCleanupRule cleanupRule = new GrpcCleanupRule();
+  private Server server;
+
+  @Test
+  public void plaintextClientServer() throws IOException {
+    getXdsServer(/* downstreamTlsContext= */ null);
```
creamsoup, line 62: why get server is not assigned?

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,224 @@
...
+ */
+@RunWith(JUnit4.class)
+public class XdsSdsClientServerTest {
+
+  /** Bad/untrusted server certs. */
+  private static final String BAD_SERVER_PEM_FILE = "badserver.pem";
+  private static final String BAD_SERVER_KEY_FILE = "badserver.key";
+
+  @Rule public final GrpcCleanupRule cleanupRule = new GrpcCleanupRule();
+  private Server server;
```
creamsoup, line 58: this instance variable can be local variable.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+  @Test
+  public void plaintextClientServer() throws IOException {
+    getXdsServer(/* downstreamTlsContext= */ null);
+    buildClientAndTest(
```
creamsoup, line 63: can you split into 2 methods?

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+  /** mTLS - client auth enabled. */
+  @Test
+  public void mtlsClientServer_withClientAuthentication() throws IOException {
+    UpstreamTlsContext upstreamTlsContext =
+        SecretVolumeSslContextProviderTest.buildUpstreamTlsContextFromFilenames(
+            SecretVolumeSslContextProviderTest.CLIENT_KEY_FILE,
+            SecretVolumeSslContextProviderTest.CLIENT_PEM_FILE,
+            SecretVolumeSslContextProviderTest.CA_PEM_FILE);
+    mtlsCommonTest(upstreamTlsContext);
```
creamsoup, line 93: is this testing test utility method?

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,224 @@
...
+            SecretVolumeSslContextProviderTest.SERVER_1_KEY_FILE,
+            SecretVolumeSslContextProviderTest.SERVER_1_PEM_FILE,
+            SecretVolumeSslContextProviderTest.CA_PEM_FILE);
+
+    XdsClient.ListenerWatcher listenerWatcher = getXdsServer(downstreamTlsContext);
+    buildClientAndTest(upstreamTlsContext, "foo.test.google.fr", "buddy", server.getPort());
+    return listenerWatcher;
+  }
+
+  private XdsClient.ListenerWatcher getXdsServer(DownstreamTlsContext downstreamTlsContext)
```
creamsoup, line 133: ListenerWatcher is not XdsServer, is it?

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  private static int findFreePort() throws IOException {
+    try (ServerSocket socket = new ServerSocket(0)) {
+      socket.setReuseAddress(true);
+      return socket.getLocalPort();
+    }
+  }
+
+  private static void createListenerUpdate(
```
creamsoup, line 164: the method name doesn't match to what it actually does. it makes the reader very confusing.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+    XdsChannelBuilder builder =
+        XdsChannelBuilder.forTarget("localhost:" + serverPort).tlsContext(upstreamTlsContext);
+    if (overrideAuthority != null) {
+      builder = builder.overrideAuthority(overrideAuthority);
+    }
+    SimpleServiceGrpc.SimpleServiceBlockingStub blockingStub =
+        SimpleServiceGrpc.newBlockingStub(cleanupRule.register(builder.build()));
+    String resp = unaryRpc(requestMessage, blockingStub);
+    assertThat(resp).isEqualTo("Hello " + requestMessage);
```
creamsoup, line 201: this doesn't verify any security, it just verifies it is broken or not. is there way to verify?

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,224 @@
...
+    try {
+      buildClientAndTest(upstreamTlsContext, "foo.test.google.fr", "buddy", server.getPort());
+      fail("exception expected");
+    } catch (StatusRuntimeException sre) {
+      assertThat(sre).hasCauseThat().isInstanceOf(SSLHandshakeException.class);
+      assertThat(sre).hasCauseThat().hasMessageThat().isEqualTo("General OpenSslEngine problem");
+    }
+  }
+
+  private XdsClient.ListenerWatcher mtlsCommonTest(UpstreamTlsContext upstreamTlsContext)
```
creamsoup, line 120: why test returns something unrelated?

creamsoup, line 120: it is hard to expect the behavior of this test, can you change to better name?

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,224 @@
...
+    final XdsClientWrapperForServerSds xdsClientWrapperForServerSds =
+        createXdsClientWrapperForServerSds(port, downstreamTlsContext);
+    SdsProtocolNegotiators.ServerSdsProtocolNegotiator serverSdsProtocolNegotiator =
+        new SdsProtocolNegotiators.ServerSdsProtocolNegotiator(xdsClientWrapperForServerSds);
+    server = cleanupRule.register(builder.buildServer(serverSdsProtocolNegotiator)).start();
+    return xdsClientWrapperForServerSds.getListenerWatcher();
+  }
+
+  /** Creates XdsClientWrapperForServerSds: also used by other classes. */
+  public static XdsClientWrapperForServerSds createXdsClientWrapperForServerSds(
```
creamsoup, line 147: should probably moved to util class.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: What is "branched" ? 

I agree it is hard - git knows the file has moved but can't show the diff in this UI

xds/src/test/java/io/grpc/xds/internal/sds/SecretVolumeSslContextProviderTest.java
sanjaypujare: I agree there are too many places defining the same thing. If I do that in this PR it will pull in too many files in this PR. See my comments elsewhere.

xds/src/main/java/io/grpc/xds/internal/sds/XdsServerBuilder.java
sanjaypujare: formatted by my IDE :-) but will make the change

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: No, this is testing a complete mTLS scenario but used more than once to test the listener update scenario - hence the util function `mtlsCommonTest`

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: The intent is not to verify the actual TLS handshake but that an RPC was successful after an xDS/SDS managed mTLS channel was set up: so it is more of an E2E test without actually setting up xDS or even SDS server. So this part of the test hasn't changed

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: because it is used in `mtlsClientServer_changeServerContext_expectException`. I can change it from a return value to a class instance variable or rename it to address this comment

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: ok

xds/src/test/java/io/grpc/xds/internal/sds/SecretVolumeSslContextProviderTest.java
sanjaypujare: Assuming you are okay with adding other changes to this PR I will move this stuff to the utility class.

xds/src/test/java/io/grpc/xds/internal/sds/SdsProtocolNegotiatorsTest.java
sanjaypujare: renamed to say "nullSyncContext"

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: If they always go together, I prefer to keep it as one method.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: BTW, security is verified by using a "bad" cert which is part of `mtlsClientServer_changeServerContext_expectException`. That also tests updating the server context.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: This is a plain mTLS test that is similar to `mtlsClientServer_withClientAuthentication`. For code reuse with `mtlsClientServer_changeServerContext_expectException` I made this a common method called from both and I think `mtlsCommonTest` is a concise and appropriate name. `mtls` implies "withClientAuthentication" so not sure what else to add.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: Moved to `XdsClientWrapperForServerSdsTest` since it cannot be in `CommonTlsContextTestsUtil` due to different package and I didn't see the need to create a  new util class here just for this

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: If you have a suggestion let me know

sanjaypujare: renamed the method which also addresses the other comment about method name

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: still waiting for the "branched" explanation..

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
creamsoup: what i am worried is go/unit-test-practices#assertions-examples

most of the test looks like the bad example in that link.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,204 @@
...
+    getXdsServer(/* downstreamTlsContext= */ null);
+    buildClientAndTest(
+        /* upstreamTlsContext= */ null, /* overrideAuthority= */ null, "buddy", server.getPort());
+  }
+
+  /** TLS channel - no mTLS. */
+  @Test
+  public void tlsClientServer_noClientAuthentication() throws IOException {
+    DownstreamTlsContext downstreamTlsContext =
+        SecretVolumeSslContextProviderTest.buildDownstreamTlsContextFromFilenames(
```
creamsoup, line 67: this method should be in the util class too

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,204 @@
...
+            CommonTlsContextTestsUtil.SERVER_1_KEY_FILE,
+            CommonTlsContextTestsUtil.SERVER_1_PEM_FILE,
+            null);
+
+    getXdsServer(downstreamTlsContext);
+
+    // for TLS, client only needs trustCa
+    UpstreamTlsContext upstreamTlsContext =
+        SecretVolumeSslContextProviderTest.buildUpstreamTlsContextFromFilenames(
```
creamsoup, line 76: same here. the method should be in the util class.

xds/src/test/java/io/grpc/xds/internal/sds/SdsProtocolNegotiatorsTest.java
```diff
@@ -259,10 +271,9 @@ public void clientSdsProtocolNegotiatorNewHandler_fireProtocolNegotiationEvent()
   }
 
   @Test
-  public void serverSdsProtocolNegotiator_passNulls_expectPlaintext() {
+  public void serverSdsProtocolNegotiator_nullSyncContext_expectPlaintext() {
     InternalProtocolNegotiator.ProtocolNegotiator protocolNegotiator =
-        SdsProtocolNegotiators.serverProtocolNegotiator(null, 7000,
-            null);
+        SdsProtocolNegotiators.serverProtocolNegotiator(7000, null);
```
creamsoup, line 276: can you use? `/* paramName= */`, like i mentioned before this is actually checked by linter which is less error prone. applies to other places.

xds/src/test/java/io/grpc/xds/internal/sds/SecretVolumeSslContextProviderTest.java
```diff
@@ -419,8 +404,22 @@ static void doChecksOnSslContext(boolean server, SslContext sslContext,
   /**
    * Helper method to build DownstreamTlsContext for above tests. Called from other classes as well.
    */
-  static DownstreamTlsContext buildDownstreamTlsContextFromFilenames(
-      String privateKey, String certChain, String trustCa) {
+  public static DownstreamTlsContext buildDownstreamTlsContextFromFilenames(
```
creamsoup, line 407: should be moved to utility class if shared in many tests.

xds/src/test/java/io/grpc/xds/internal/sds/SecretVolumeSslContextProviderTest.java
```diff
@@ -429,7 +428,20 @@ static DownstreamTlsContext buildDownstreamTlsContextFromFilenames(
    * Helper method to build UpstreamTlsContext for above tests. Called from other classes as well.
    */
   public static UpstreamTlsContext buildUpstreamTlsContextFromFilenames(
```
creamsoup, line 430: same here, moved to utility class.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,204 @@
...
+    buildClientAndTest(
+        /* upstreamTlsContext= */ null, /* overrideAuthority= */ null, "buddy", server.getPort());
+  }
+
+  /** TLS channel - no mTLS. */
+  @Test
+  public void tlsClientServer_noClientAuthentication() throws IOException {
+    DownstreamTlsContext downstreamTlsContext =
+        SecretVolumeSslContextProviderTest.buildDownstreamTlsContextFromFilenames(
+            CommonTlsContextTestsUtil.SERVER_1_KEY_FILE,
```
creamsoup, line 68: FYI/nit: those could be static imported, because it is not vague.

sanjaypujare: Yes, I used IDE's "Find and Replace Code Duplicates" and "Inline Field" and this was the result

xds/src/test/java/io/grpc/xds/internal/sds/SdsProtocolNegotiatorsTest.java
sanjaypujare: I was talking to @ejona86 some time ago and he said for test code this is completely unnecessary (IIRC). It will be good to have consensus

ejona86: I don't think we have an exception for this for test code, but I also don't see any failure in the build so I don't think lint complains about this.

Looking at ErrorProne these are the checks I could find that are impacted by argument comments (none of them apply):
https://errorprone.info/bugpattern/ArgumentSelectionDefectChecker
https://errorprone.info/bugpattern/BooleanParameter
https://errorprone.info/bugpattern/ParameterComment
https://errorprone.info/bugpattern/ParameterName

It isn't at all clear to me why `syncContext` would be `null`, and why that would imply plain-text. That seems super-strange to me. The `port` also seems unnecessary since the negotiator will have access to that information from io.netty.channel.Channel.localAddress().

I don't think we have a rule that you are required to put the parameter names, but you should put the parameter names when it could be confusing otherwise. In this case the `null` is really important to the test, but there's also only one null. The previous version is more confusing. In any case for a test, instead of having the comment I would be more prone to doing:
```
SynchronizationContext syncContext = null; // or syncContext = null; if it is a class field
SdsProtocolNegotiators.serverProtocolNegotiator(7000, syncContext);
```

But really, I'd hope we can drop the 7000 and always specify a syncContext.

xds/src/test/java/io/grpc/xds/internal/sds/SecretVolumeSslContextProviderTest.java
sanjaypujare: For file-based credentials (as the name suggests), this can be considered as the utility class. Are you thinking of one utility class to have all these methods?

xds/src/test/java/io/grpc/xds/internal/sds/SdsProtocolNegotiatorsTest.java
ejona86: Hmm... I guess the 7000 is specified here since there will be a single watch for the entire server. This is a problem though, because that is not normal initialization for a protocol negotiator; this approach won't be available in a ServerCredentials API. There may need to be more hole-punching to allow it.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: For some reason I cannot comment on your other comment: at the time of first checkin git knows the file has been moved as it shows this:
```
git status
On branch xds-server-test
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	renamed:    xds/src/test/java/io/grpc/xds/internal/sds/XdsSdsClientServerTest.java -> xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
...
```
But if you do a git log for that commit it shows delete and add of 2 different files:
```
xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java                          | 224 +++++++++++++++++++++++++++++++++++++++++++++
xds/src/test/java/io/grpc/xds/internal/sds/XdsSdsClientServerTest.java             | 193 --------------------------------------
```
Unless you have an idea how to fix it, I am going to resolve this comment

creamsoup: don't worry about it. i reviewed the file anyways.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: No, unless I declare a POJO just to return and use 2 values: the `Server` and `listenerWatcher`, since a function cannot return 2 values. Look at the flow in `mtlsClientServer_changeServerContext_expectException` calling `performMtlsTestAndGetListenerWatcher` and `buildClientAndTest`

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: because I don't need it. I will assign it to `unused` to address this comment

xds/src/test/java/io/grpc/xds/internal/sds/SdsProtocolNegotiatorsTest.java
sanjaypujare: Yes the port is provided for setting the LDS watch for the server (not sure what you mean by "entire server"). ServerCredentials is attached to a server and a server is attached to a creds (since you can't reuse creds?) In that case you can get the server port from the creds, right?

sanjaypujare: `syncContext` can never be null since it is created by the server builder before invoking the SdsProtocolNegotiator.  So the test is a bit misleading but is there to simulate XdsClientImpl throwing exception (because syncContext is null).

sanjaypujare: added the param comments after doing a quick search

ejona86: The user will create any ServerCredentials before they are passed to the ServerBuilder, so [passing in port like this simply won't work](https://github.com/grpc/grpc-java/blob/54f9e29cdb205062158e19d9dc1a567afd61a318/xds/src/main/java/io/grpc/xds/internal/sds/XdsServerBuilder.java#L177). And yes, the user could use the same server credentials with multiple servers. It also looks like you are having "creation imply starting", as the credential appears to start the watch when it is created. The user will create the credential well before the port is even determined, so that is also not going to work. These things are important to call out because they mean additional plumbing would be required.

This test is horrible. Abusing `syncContext` in that way would deserve a big comment or _some_ sort of pointer to something weird going on. Naming it something like "serverSdsProtocolNegotiator_xdsClientCreationThrows_expectPlaintext" would be a good starter. But that is even wrong. `syncContext` is a red herring because [the only catch in SdsProtocolNegotiators is for IOException](https://github.com/grpc/grpc-java/blob/54f9e29cdb205062158e19d9dc1a567afd61a318/xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java#L289). That means it wouldn't catch NullPointerException and it also wouldn't catch [NoSuchElementException](https://github.com/grpc/grpc-java/blob/54f9e29cdb205062158e19d9dc1a567afd61a318/xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java#L86). So at the very least I've determined the exception handling is broken (XdsClientWrapperForServerSds should probably not be throwing a RuntimeException based on bad input from a user). The reason the test passes is because it triggers "java.io.IOException: Environment variable GRPC_XDS_BOOTSTRAP not defined.".

What is clearly missing is calling `serverProtocolNegotiator()` and it returning an actual ServerSdsProtocolNegotiator. That said, even ServerSdsProtocolNegotiator should have a protocol of `http` as the xds server may not return security information. (Actually, it seems it should delegate to a fallback protocol negotiator, e.g., serverPlaintext(), so the protocol wouldn't end up changing.)

sanjaypujare: > The user will create the credential well before the port is even determined, so that is also not going to work. These things are important to call out because they mean additional plumbing would be required.

But the port is determined or known when the server is built (`XdsServerBuilder.build`) and that is when the watch is set: although I agree it could be delayed until `start`. Doing that will require me to reintroduce `XdsServer` I had in a previous PR for the shutdown case.  Can be done in a separate PR.

> But that is even wrong. syncContext is a red herring because the only catch in SdsProtocolNegotiators is for IOException. That means it wouldn't catch NullPointerException and it also wouldn't catch NoSuchElementException. So at the very least I've determined the exception handling is broken (XdsClientWrapperForServerSds should probably not be throwing a RuntimeException based on bad input from a user).

Interesting - I did have the broader exception earlier but made it narrower based on PR comments. Again I can fix this in a separate PR after we agree on the approach.

> What is clearly missing is calling serverProtocolNegotiator() and it returning an actual ServerSdsProtocolNegotiator. 

So you think returning `InternalProtocolNegotiators.serverPlaintext()` is not correct for `serverProtocolNegotiator()` and should always be delegated to the `ServerSdsProtocolNegotiator`? I think that was the old approach.

> That said, even ServerSdsProtocolNegotiator should have a protocol of http as the xds server may not return security information.

Interesting. The `ServerSdsProtocolNegotiator` does the right thing based on the current state of the security information from the xDS server (whether DownstreamTlsContext is present or not) just that SCHEME is hardcoded to https. If you are saying it should always be "http" that should be easy to change. Changing it to "https" or "http" based on the current state is also not that hard (I think).

> (Actually, it seems it should delegate to a fallback protocol negotiator, e.g., serverPlaintext(), so the protocol wouldn't end up changing.)

It does delegate to `serverPlaintext()` at the time of creating the handler if you see `HandlerPickerHandler.newHandler`.

ejona86: > But the port is determined or known when the server is built (XdsServerBuilder.build) and that is when the watch is set: although I agree it could be delayed until start.

Even now it must be delayed until start, because that's when I/O is permitted to start. Technically, the user could create a server and never start it. Yes, I'd expect fixing this sort of thing to be done in a separate PR.

> Interesting - I did have the broader exception earlier but made it narrower based on PR comments. Again I can fix this in a separate PR after we agree on the approach.

The problem was with the code throwing a RuntimeException. The reviewer was right to ask for more narrow exception handling. But code randomly throwing RuntimeExceptions is asking for bugs, like failure to notice changing Exception to IOException in a catch would leave some cases unhandled. In our code, RuntimeExceptions should generally be reserved for programmer errors.

> So you think returning InternalProtocolNegotiators.serverPlaintext() is not correct for serverProtocolNegotiator() and should always be delegated to the ServerSdsProtocolNegotiator? I think that was the old approach.

If the bootstrap is unavailable, returning serverPlaintext is fine. But that's "not the end of it." I thought we had decided we were going to have the user specify a fallback credential in case xds has no tls configuration, so it seems some credential should still be provided to ServerSdsProtocolNegotiator as a fallback and that should behave the same from the user's perspective independent of when it happens. (i.e., the user should see the same behavior if GRPC_XDS_BOOTSTRAP is unset as if the xds server has no tls configuration)

> Interesting. The ServerSdsProtocolNegotiator does the right thing based on the current state of the security information from the xDS server (whether DownstreamTlsContext is present or not) just that SCHEME is hardcoded to https. If you are saying it should always be "http" that should be easy to change. Changing it to "https" or "http" based on the current state is also not that hard (I think).

Does it guarantee it is going to use a secure connection? If not, it should be http. So yes, it should be http.

Changing it on-the-fly won't work, because this is called separately from an actual connection being handled and the setting can change at any point based on xds. There's no form of synchronization available to make sure that the right value is used for the right connections.

(In truth, this value is unused on server-side today. So it doesn't really matter. Except this is security stuff. So it totally matters.)

> It does delegate to serverPlaintext() at the time of creating the handler if you see HandlerPickerHandler.newHandler.

It should have a "fallback" protocol negotiator passed in. That will mirror the final design where the user gets to decide. Hard-coding this stuff now means that we're less likely to notice issues with the long-term design.

sanjaypujare: @ejona86 for all of your comments requiring code change I'll create a separate PR instead of complicating this PR (which was meant to remove DownstreamTlsContext from the XdsServerBuilder). For "fallback" too the XdsServerBuilder could define something like say `XdsServerBuilder.fallbackProtocolNegotiator(ProtocolNegotiator protocolNegotiator)` and plumb it down. Again let's do it in a separate PR or at the time of XdsServerCredentials.

I can create an issue to keep track of all items needing that separate PR.

sanjaypujare: And there is one case of `NoSuchElementException` that I can convert to IOException or log message and return `null` (in another PR).

xds/src/test/java/io/grpc/xds/XdsClientWrapperForServerSdsTest.java
```diff
@@ -139,6 +140,26 @@
   private XdsClientWrapperForServerSds xdsClientWrapperForServerSds;
   private final DownstreamTlsContext[] tlsContexts = new DownstreamTlsContext[3];
 
+  /** Creates XdsClientWrapperForServerSds: also used by other classes. */
+  public static XdsClientWrapperForServerSds createXdsClientWrapperForServerSds(
```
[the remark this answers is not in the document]
sanjaypujare: For now I am leaving it here as the "container" for `XdsClientWrapperForServerSds` related util methods instead of creating a new util class in this package.

sanjaypujare:
rebased to resolve the merge conflict.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -0,0 +1,213 @@
...
+ */
+@RunWith(JUnit4.class)
+public class XdsSdsClientServerTest {
+
+  @Rule public final GrpcCleanupRule cleanupRule = new GrpcCleanupRule();
+  private int port;
+
+  @Before
+  public void setUp() throws IOException {
+    port = findFreePort();
```
creamsoup, line 65: nit: you can directly assign findFreePort() in L61.

PR #6904, opened 2020-04-07
rls: delegating helper for rls child policies

creamsoup:


rls/src/main/java/io/grpc/rls/internal/ChildLoadBalancerHelper.java
```diff
@@ -0,0 +1,84 @@
...
+    private final SubchannelPicker picker;
+
+    ChildLoadBalancerHelperProvider(
+        Helper helper, SubchannelStateManager subchannelStateManager, SubchannelPicker picker) {
+      this.helper = checkNotNull(helper, "helper");
+      this.subchannelStateManager = checkNotNull(subchannelStateManager, "subchannelStateManager");
+      this.picker = checkNotNull(picker, "picker");
+    }
+
+    ChildLoadBalancerHelper forTarget(String target) {
```
dapengzhang0, line 80: Is just `Helper` as return value type sufficient?

rls/src/test/java/io/grpc/rls/internal/ChildLoadBalancerHelperTest.java
```diff
@@ -0,0 +1,60 @@
...
+public class ChildLoadBalancerHelperTest {
+
+  private final Helper helper = mock(Helper.class);
+  private final SubchannelStateManager subchannelStateManager = new SubchannelStateManagerImpl();
+  private final SubchannelPicker picker = mock(SubchannelPicker.class);
+  private final ChildLoadBalancerHelperProvider provider =
+      new ChildLoadBalancerHelperProvider(helper, subchannelStateManager, picker);
+
+  @Test
+  public void childLoadBalancerHelper_shouldReportsSubchannelState() {
```
dapengzhang0, line 41: no assertion for `subchannelStateManager.getState(target)` ?

creamsoup: hmm i was trying to limit the scope but it sounds fair to add them.

rls/src/main/java/io/grpc/rls/internal/ChildLoadBalancerHelper.java
creamsoup: yes, `Helper` is sufficient and it is usually better choice. I am worried about passing/using wrong helper. There will be another helper class (total 3 helpers - original from channel, this one and the other one not sent yet). Using it as `Helper` class confused me, so probably it will confuse other people.

PR #6918, opened 2020-04-10
netty: Reduce race window size between GOAWAY and new streams

ejona86:
The race between new streams and transport shutdown is #2562, but it is still
far from being generally solved. This reduces the race window of new streams
from (transport selection → stream created on network thread) to (transport
selection → stream enqueued on network thread). Since only a single thread now
needs to do work in the stream creation race window, the window should be
dramatically smaller.

This only reduces GOAWAY races when the server performs a graceful shutdown
(using two GOAWAYs), as that is the only non-racy way on-the-wire to shutdown a
connection in HTTP/2.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/WriteQueue.java
+++ b/netty/src/main/java/io/grpc/netty/WriteQueue.java
@@ -102,9 +102,9 @@ class WriteQueue {
   }
 
   /**
-   * Execute enqueued work directly on the current thread. This can be used to trigger writes before
-   * performing additional reads. Must be called from the event loop. This method makes no guarantee
-   * that the work queue is empty when it returns.
+   * Executes enqueued work directly on the current thread. This can be used to trigger writes
+   * before performing additional reads. Must be called from the event loop. This method makes no
+   * guarantee that the work queue is empty when it returns.
    */
   void drainNow() {
     Preconditions.checkState(channel.eventLoop().inEventLoop(), "must be on the event loop");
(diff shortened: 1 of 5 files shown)


netty/src/main/java/io/grpc/netty/WriteQueue.java
```diff
@@ -101,6 +101,19 @@ void enqueue(Runnable runnable, boolean flush) {
     }
   }
 
+  /**
+   * Execute enqueued work directly on the current thread. This can be used to trigger writes before
```
dapengzhang0, line 105: s/Execute/Executes/

netty/src/test/java/io/grpc/netty/NettyHandlerTestBase.java
```diff
@@ -203,6 +203,10 @@ public int compareTo(Delayed o) {
     }
   }
 
+  protected final WriteQueue writeQueue() {
```
dapengzhang0, line 206: This method should be final because `initChannel()` is final, also `enqueue()` is final.

ejona86: This method is final...

netty/src/main/java/io/grpc/netty/WriteQueue.java
ejona86: Done. Hmm... the other methods are wrong as well. I'll leave them as-is.

PR #6920, opened 2020-04-11
Change logging level from INFO to FINE.

ashithasantosh:
In AltsTsiHandshaker.java, we log maximum frame size value. Changing the log level from INFO to FINE.


alts/src/main/java/io/grpc/alts/internal/AltsTsiHandshaker.java
```diff
@@ -189,7 +189,7 @@ public TsiFrameProtector createFrameProtector(int maxFrameSize, ByteBufAllocator
       maxFrameSize = Math.min(peerMaxFrameSize, AltsTsiFrameProtector.getMaxFrameSize());
       maxFrameSize = Math.max(AltsTsiFrameProtector.getMinFrameSize(), maxFrameSize);
     }
-    logger.log(Level.INFO, "Maximum frame size value is " + maxFrameSize);
+    logger.log(Level.FINE, String.format("Maximum frame size value is %s.", maxFrameSize));
```
dapengzhang0, line 192: Did you mean `logger.log(Level.FINE, "Maximum frame size value is {0}.", maxFrameSize);`?

PR #6927, opened 2020-04-14
deps: update to latest Guava and protobuf-java

elharo:
@creamsoup


creamsoup:
there are many references to the protobuf version in the repo.
`git grep '3\.11\.0'` returns 33 references. can you update thsoe? it seems like just replacing them all should work.

dapengzhang0:
Guava was updated by #7079 
Protobuf was updated by #7052

PR #6935, opened 2020-04-16
xds: add tests & misc fixes based on outstanding items

sanjaypujare:


xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -84,7 +84,7 @@ public static XdsClientWrapperForServerSds newInstance(
     Bootstrapper.BootstrapInfo bootstrapInfo = bootstrapper.readBootstrap();
     final List<Bootstrapper.ServerInfo> serverList = bootstrapInfo.getServers();
     if (serverList.isEmpty()) {
-      throw new NoSuchElementException("No management server provided by bootstrap");
+      throw new FileNotFoundException("No management server provided by bootstrap");
```
creamsoup, line 87: this does not look correct. it is not even an IO exception (this is why it is originally not an IO exception), consider using custom checked exception if there isn't any good exception to use.

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
```diff
@@ -60,7 +60,7 @@ private SdsProtocolNegotiators() {
 
   private static final Logger logger = Logger.getLogger(SdsProtocolNegotiators.class.getName());
 
-  private static final AsciiString SCHEME = AsciiString.of("https");
+  private static final AsciiString SCHEME = AsciiString.of("http");
```
creamsoup, line 63: IIRC there was a test checking scheme http if plaintext. the test should be fixed.

sanjaypujare: There was/is no test - otherwise the build would have failed. Do you want me to add a test - not sure if we achieve anything with the test.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: Yeah, I can create a custom checked exception or just return with a log message. I think the latter is preferred but let me know your preferrence.

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
creamsoup: https://github.com/grpc/grpc-java/blob/master/xds/src/test/java/io/grpc/xds/internal/sds/SdsProtocolNegotiatorsTest.java#L288

that test, it was checking "http" vs "https". now everything is "http". this test is no longer relevant and should be fixed.

sanjaypujare: Yes, you are right, may bad! I will have to modify the test to detect plaintext vs Sds PN and let me think about it.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
creamsoup: it is definitely preferred to throw exception especially if the caller need to do something.

Looked at the code i think the `getXdsClientWrapperForServerSds` shouldn't return `null`. it should just throw and  `serverProtocolNegotiator` can handle the exception (or just remove the `getXdsClientWrapperForServerSds` and inline it since it does not really belong to ServerSdsProtocolNegotiator); it sees exception so it can fallback seems more natural than hmm somehow we have null, lets fallback (not very idiomatic java, it is c style - no offense to c =p).

xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java
sanjaypujare: With scheme being the same the only other thing to check is the class-name of the returned `protocolNegotiator` it should match `...ServerPlaintextNegotiator` or should have `plaintext` in it. Do you have a better suggestion?

creamsoup: yeah there isn't a good way to test this other than using name of the class. i think it is fine to check the class name (`ServerPlaintextNegotiator`).

sanjaypujare:
thanks @creamsoup

PR #6942, opened 2020-04-17
Avoid null checks for internal methods in Context

bogdandrutu:


bogdandrutu:
Failing test in travis is transient (download issue for one dep). Please restart that :)

PR #6948, opened 2020-04-20
xds: implement requireClientCertificate semantics

sanjaypujare:


xds/src/test/java/io/grpc/xds/internal/sds/CommonTlsContextTestsUtil.java
```diff
@@ -170,7 +174,10 @@ static String getTempFileNameForResourcesFile(String resFile) throws IOException
    * Helper method to build DownstreamTlsContext for above tests. Called from other classes as well.
    */
   public static DownstreamTlsContext buildDownstreamTlsContextFromFilenames(
-      @Nullable String privateKey, @Nullable String certChain, @Nullable String trustCa) {
+      @Nullable String privateKey,
+      @Nullable String certChain,
+      @Nullable String trustCa,
+      boolean requireClientCert) {
```
creamsoup, line 180: consider make a 2 methods sets `requirClientCert` and make this method private. so existing test doesn't need to be modified. although name is already long, this can be more descriptive.

sanjaypujare: So you are suggesting 2 (public) methods like
```
buildDownstreamTlsContextFromFilenames
buildDownstreamTlsContextFromFilenamesRequireClientCert
```
calling this method with requireClientCert=false and requireClientCert=true respectively (and make this one private)?

xds/src/main/java/io/grpc/xds/internal/sds/SslContextProvider.java
```diff
@@ -66,6 +73,21 @@ CommonTlsContext getCommonTlsContext() {
     return null;
   }
 
+  protected void setClientAuthValues(
+      SslContextBuilder sslContextBuilder, CertificateValidationContext localCertValidationContext)
+      throws CertificateException, IOException, CertStoreException {
+    if (localCertValidationContext != null) {
+      sslContextBuilder.trustManager(new SdsTrustManagerFactory(localCertValidationContext));
+      DownstreamTlsContext downstreamTlsContext = (DownstreamTlsContext) getSource();
```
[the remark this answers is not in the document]
sanjaypujare: added checks in SslContextProvider constructor and also in this method

xds/src/test/java/io/grpc/xds/internal/sds/CommonTlsContextTestsUtil.java
sanjaypujare: changed made.

creamsoup: yes

xds/src/test/java/io/grpc/xds/internal/sds/SecretVolumeSslContextProviderTest.java
```diff
@@ -274,7 +274,8 @@ public void getProviderForServer_defaultTlsCertificate_throwsException() {
     try {
       SecretVolumeSslContextProvider.getProviderForServer(
           CommonTlsContextTestsUtil.buildDownstreamTlsContext(
-              CommonTlsContextTestsUtil.getCommonTlsContext(tlsCert, /* certContext= */ null)));
+              CommonTlsContextTestsUtil.getCommonTlsContext(tlsCert, /* certContext= */ null),
+              /* requireClientCert= */ false));
```
creamsoup, line 278: we don't need those changes anymore.

sanjaypujare: this is a different one. This one doesn't take filenames but just a CommonTlsContext and the new bool. I could have created the same 2 methods here too but this is not so widely used

creamsoup: oops okay, ideally passing boolean param is not recommended if can avoid. but this one is not that bad.

xds/src/main/java/io/grpc/xds/internal/sds/SslContextProvider.java
```diff
@@ -66,6 +79,22 @@ CommonTlsContext getCommonTlsContext() {
     return null;
   }
 
+  protected void setClientAuthValues(
+      SslContextBuilder sslContextBuilder, CertificateValidationContext localCertValidationContext)
+      throws CertificateException, IOException, CertStoreException {
+    checkState(server);
```
creamsoup, line 85: nit: should have a meaningful message.
btw, this class is rather confusing (and errorprone). it should be split into 2 different classes (server and client). can it be done? how hard to change?

sanjaypujare: We had this discussion before. UpstreamTlsContext and DownstreamTlsContext are quite similar but there is no common abstraction to use. So I ended up using the generic K to eliminate code duplication. Splitting into server & client classes is definitely possible but I suspect there will be code duplication. Do you have better ideas?

creamsoup: Yeah i thought we discussed this before.
i think code dup is better than error prone / confusing nature of current structure. we can still  share some code as static or protected in parent method.

sanjaypujare: Okay I will add a TODO for this suggestion and in the list of pending PRs

PR #6957, opened 2020-04-21
rls: fake direct scheduled executor supports schedule

creamsoup:
FYI. the schedule method is not used in this test, but will be used in other test.

### DIFF
--- a/rls/src/main/java/io/grpc/rls/internal/LinkedHashLruCache.java
+++ b/rls/src/main/java/io/grpc/rls/internal/LinkedHashLruCache.java
@@ -192,22 +192,6 @@ abstract class LinkedHashLruCache<K, V> implements LruCache<K, V> {
   }
 
   @Override
-  public final boolean invalidate(K key, V value) {
-    checkNotNull(key, "key");
-    checkNotNull(value, "value");
-    synchronized (lock) {
-      SizedValue existing = delegate.get(key);
-      if (existing != null && existing.value.equals(value)) {
-        delegate.remove(key);
-        evictionListener.onEviction(key, existing, EvictionType.EXPLICIT);
-        return true;
-      } else {
-        return false;
-      }
-    }
-  }
-
-  @Override
   public final void invalidateAll(Iterable<K> keys) {
     checkNotNull(keys, "keys");
     synchronized (lock) {
--- a/rls/src/main/java/io/grpc/rls/internal/LinkedHashLruCache.java
+++ b/rls/src/main/java/io/grpc/rls/internal/LinkedHashLruCache.java
@@ -192,6 +192,22 @@ abstract class LinkedHashLruCache<K, V> implements LruCache<K, V> {
   }
 
   @Override
+  public final boolean invalidate(K key, V value) {
+    checkNotNull(key, "key");
+    checkNotNull(value, "value");
+    synchronized (lock) {
+      SizedValue existing = delegate.get(key);
+      if (existing != null && existing.value.equals(value)) {
+        delegate.remove(key);
+        evictionListener.onEviction(key, existing, EvictionType.EXPLICIT);
+        return true;
+      } else {
+        return false;
+      }
+    }
+  }
+
+  @Override
   public final void invalidateAll(Iterable<K> keys) {
     checkNotNull(keys, "keys");
     synchronized (lock) {
--- a/rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
+++ b/rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
@@ -107,7 +107,7 @@ abstract class DoNotUseDirectScheduledExecutorService implements ScheduledExecut
     return scheduledRunnable.scheduledFuture;
   }
 
-  FakeTimeProvider getFakeTimeProvider() {
+  final FakeTimeProvider getFakeTimeProvider() {
     maybeInit();
     return new FakeTimeProvider();
   }
@@ -172,7 +172,7 @@ abstract class DoNotUseDirectScheduledExecutorService implements ScheduledExecut
           .toString();
     }
 
-    private class FakeScheduledFuture implements ScheduledFuture<Object> {
+    private final class FakeScheduledFuture implements ScheduledFuture<Object> {
       boolean cancelled = false;
 
       @Override
@@ -225,7 +225,7 @@ abstract class DoNotUseDirectScheduledExecutorService implements ScheduledExecut
     }
   }
 
-  class FakeTimeProvider implements TimeProvider {
+  final class FakeTimeProvider implements TimeProvider {
 
     @Override
     public long currentTimeNanos() {
--- a/rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
+++ b/rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
@@ -60,6 +60,10 @@ abstract class DoNotUseDirectScheduledExecutorService implements ScheduledExecut
   private PriorityQueue<ScheduledRunnable> scheduledCommands;
   private boolean initialized;
 
+  private DoNotUseDirectScheduledExecutorService() {
+    throw new UnsupportedOperationException("this class is for mock only");
+  }
+
   /**
    * Note: CALLS_REAL_METHODS doesn't initialize instance variables, all the methods need to call
    * maybeInit if they access instance variables.
--- a/rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
+++ b/rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
@@ -68,7 +68,7 @@ abstract class DoNotUseDirectScheduledExecutorService implements ScheduledExecut
    * Note: CALLS_REAL_METHODS doesn't initialize instance variables, all the methods need to call
    * maybeInit if they access instance variables.
    */
-  private synchronized void maybeInit() {
+  private void maybeInit() {
     if (initialized) {
       return;
     }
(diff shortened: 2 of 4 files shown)


rls/src/main/java/io/grpc/rls/internal/LinkedHashLruCache.java
```diff
@@ -191,6 +191,22 @@ private V invalidate(K key, EvictionType cause) {
     }
   }
 
+  @Override
+  public final boolean invalidate(K key, V value) {
+    checkNotNull(key, "key");
+    checkNotNull(value, "value");
+    synchronized (lock) {
+      SizedValue existing = delegate.get(key);
```
dapengzhang0, line 199: `get()` will trigger LinkedHashMap's access ordering. If the existing value is not the one to be deleted, it will be re-inserted to the tail of the linked list. Maybe 
```
boolean replaced = replace(key, value, null);
if (replaced) {
  delegate.remove(key);
  evictionListener.onEviction(key, existing, EvictionType.EXPLICIT);
  return true;
}
```

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
```diff
@@ -0,0 +1,235 @@
...
+  private boolean initialized = false;
+
+  private synchronized void maybeInit() {
+    if (initialized) {
+      return;
+    }
+
+    initialized = true;
+    repeatedCommand = new AtomicReference<>();
+    scheduledCommands = new PriorityQueue<>(11, new ScheduledRunnableComparator());
```
dapengzhang0, line 70: Seems every method of `DoNotUseDirectScheduledExecutorService` calls `maybeInit()`, why not just init these fields directly at their declaration and remove `maybeInit()`.

rls/src/main/java/io/grpc/rls/internal/LinkedHashLruCache.java
creamsoup: yes, i didn't really consider the access ordering. thanks

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
creamsoup: this was due to the CALLS_REAL_METHODS doesn't initialize instance variables.

rls/src/main/java/io/grpc/rls/internal/LinkedHashLruCache.java
creamsoup: unfortunately the replace is java8 api. doesn't seem like there is a good option for java7

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
dapengzhang0: Maybe also add a non-constructor
```
private DoNotUseDirectScheduledExecutorService() {
  throw new UnsupportedOperationException("this class is for mock only");
}
```

Btw, did you ever consider implementing `FakeClock.ScheduledExecutorImpl.scheduleAtFixedRate()`? Will there be any problem?

rls/src/main/java/io/grpc/rls/internal/LinkedHashLruCache.java
dapengzhang0: Maybe leave here a TODO with rationale on access ordering and java7 restriction.

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
```diff
@@ -0,0 +1,239 @@
...
+ * implementation is not thread-safe. Not safe to use elsewhere.
+ */
+abstract class DoNotUseDirectScheduledExecutorService implements ScheduledExecutorService {
+
+  private long currTimeNanos;
+  private long period;
+  private long nextRun;
+  private AtomicReference<Runnable> repeatedCommand;
+  private PriorityQueue<ScheduledRunnable> scheduledCommands;
+  private boolean initialized = false;
```
dapengzhang0, line 61: Don't set initial value here.

rls/src/main/java/io/grpc/rls/internal/LinkedHashLruCache.java
creamsoup: i am trying to remove this method. and make the caller work around this problem.

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
creamsoup: the only reason i didn't implement scheduleAtFixedRate is i don't need it. it can be added if we need. also supporting multiple scheduled tasks is not hard. just don't need it at this point.

rls/src/main/java/io/grpc/rls/internal/LinkedHashLruCache.java
creamsoup: okay removed.

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
```diff
@@ -0,0 +1,243 @@
...
+      public int compareTo(Delayed unused) {
+        throw new UnsupportedOperationException();
+      }
+
+      @Override
+      public synchronized boolean cancel(boolean mayInterruptIfRunning) {
+        if (running) {
+          return false;
+        }
+        running = true;
```
dapengzhang0, line 196: Why set running = true?

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
```diff
@@ -0,0 +1,243 @@
...
+        return unit.convert(scheduledTimeNanos - currTimeNanos, TimeUnit.NANOSECONDS);
+      }
+
+      @Override
+      public int compareTo(Delayed unused) {
+        throw new UnsupportedOperationException();
+      }
+
+      @Override
+      public synchronized boolean cancel(boolean mayInterruptIfRunning) {
```
dapengzhang0, line 192: If you really want ThreadSafe, isCancelled() and isDone() also need `synchronized`.

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
```diff
@@ -0,0 +1,243 @@
...
+
+  private DoNotUseDirectScheduledExecutorService() {
+    throw new UnsupportedOperationException("this class is for mock only");
+  }
+
+  /**
+   * Note: CALLS_REAL_METHODS doesn't initialize instance variables, all the methods need to call
+   * maybeInit if they access instance variables.
+   */
+  private synchronized void maybeInit() {
```
dapengzhang0, line 71: If you really want to support callers from multiple threads without caller's synchronization, other methods also need `synchronized`. As javadoc says implementation is not thread-safe,  so `synchronized` is not necessary.

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
```diff
@@ -0,0 +1,243 @@
...
+      command.run();
+    }
+    this.repeatedCommand.set(checkNotNull(command, "command"));
+    this.nextRun = checkNotNull(unit, "unit").toNanos(initialDelay) + currTimeNanos;
+    this.period = unit.toNanos(period);
+    return mock(ScheduledFuture.class);
+  }
+
+  @Override
+  public ScheduledFuture<?> schedule(Runnable command, long delay, TimeUnit unit) {
```
dapengzhang0, line 99: nit: final?

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
```diff
@@ -0,0 +1,243 @@
...
+    scheduledCommands.add(scheduledRunnable);
+    return scheduledRunnable.scheduledFuture;
+  }
+
+  FakeTimeProvider getFakeTimeProvider() {
+    maybeInit();
+    return new FakeTimeProvider();
+  }
+
+  void forwardTime(long delta, TimeUnit unit) {
```
dapengzhang0, line 115: nit: private?

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
creamsoup: yeah removing it makes sense. making it thread safe requires little more work, but it is really meant to be "direct".

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
creamsoup: done.

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
```diff
@@ -0,0 +1,239 @@
...
+
+  private static final class ScheduledRunnableComparator
+      implements Comparator<ScheduledRunnable> {
+    @Override
+    public int compare(ScheduledRunnable o1, ScheduledRunnable o2) {
+      return Long.compare(o1.scheduledTimeNanos, o2.scheduledTimeNanos);
+    }
+  }
+
+  class FakeTimeProvider implements TimeProvider {
```
dapengzhang0, line 228: nit: final

rls/src/test/java/io/grpc/rls/internal/DoNotUseDirectScheduledExecutorService.java
```diff
@@ -0,0 +1,239 @@
...
+    checkNotNull(command, "command");
+    checkNotNull(unit, "unit");
+    checkArgument(delay > 0, "delay must be positive");
+    ScheduledRunnable scheduledRunnable =
+        new ScheduledRunnable(currTimeNanos + TimeUnit.NANOSECONDS.convert(delay, unit), command);
+    scheduledCommands.add(scheduledRunnable);
+    return scheduledRunnable.scheduledFuture;
+  }
+
+  FakeTimeProvider getFakeTimeProvider() {
```
dapengzhang0, line 110: nit: finale

PR #6965, opened 2020-04-22
travis: add spell check for PRs

sf7h:
**What**
Add spelling check for PRs on newly added files

**Why**
See [issue 5609](https://github.com/grpc/grpc-java/issues/5609)

**How**
- Added a script that downloads `misspell` and runs it on newly added files
- Added a check in travis.xml to run the spelling script 
- Added a helper script for `misspell` downloading


creamsoup:
thanks for the PR. 
for spell checking, I think those 3 requirements are must. other maintainers may have more.

* it seems like the spellcheck is for the entire repo. is there way to limit the scope? we would like to check the specific change.
* it would be nice to check this locally (probably gradle task), instead of discovering it on travis which can be PITA. also, we would like to have an option to disable it temporarily like `-PskipSpellcheck=true`.
* Is there way to whitelist words? e.g. loadbalancer is usually not in the dictionary, but we use it very often. I am pretty sure we have plentiful of those. comment based might not be enough because of how frequently we use certain terms.

do you have rough idea, how many spell errors it discovered?

fathalls:
Thanks for this @creamsoup

> it seems like the spellcheck is for the entire repo. is there way to limit the scope? we would like to check the specific change.

It doesn't run on the whole codebase, I am applying it just for the modified files:
[Applying on git ls-files -m ](https://github.com/grpc/grpc-java/pull/6965/files#diff-eab80c8155794a243a1eba38db664ec7R11)

>it would be nice to check this locally (probably gradle task), instead of discovering it on travis which can be PITA. also, we would like to have an option to disable it temporarily like -PskipSpellcheck=true.

Good point, I initially thought about a pre-commit hook. I realised that it's not a very secure choice.
Can put something together as a Gradle task. Do we want to keep the additional check in Travis? 

> Is there way to whitelist words? e.g. loadbalancer is usually not in the dictionary, but we use it very often. I am pretty sure we have plentiful of those. comment based might not be enough because of how frequently we use certain terms.

Yes, `misspell` comes with this option, we could maintain txt file with whitelisted words. Thoughts?

fathalls:
About this
> do you have rough idea, how many spell errors it discovered?
There are 50 misspelling :)

creamsoup:
>It doesn't run on the whole codebase, I am applying it just for the modified files:
> [Applying on git ls-files -m ](https://github.com/grpc/grpc-java/pull/6965/files#diff-eab80c8155794a243a1eba38db664ec7R11)

yes, this sounds good.
 
> Can put something together as a Gradle task. Do we want to keep the additional check in Travis?

Travis will build using gradle. if the spellcheck is enabled by default, it will spellcheck without additional configuration.

> Yes, `misspell` comes with this option, we could maintain txt file with whitelisted words. Thoughts?

yes, this is what i was hoping.

creamsoup:
> About this
> 
> > do you have rough idea, how many spell errors it discovered?
> 
> There are 50 misspellings :)

cool, that's within what I was expecting.

larry-safran:
Since travis has been removed, this PR is no longer relevant.

PR #6966, opened 2020-04-22
rls: caching rls client

creamsoup:
the ASYNC_LOOKUP_DEFAULT_TARGET_ON_MISS will be removed via https://github.com/grpc/grpc-proto/pull/77. that is reflected here.

### DIFF
--- a/rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
+++ b/rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
@@ -107,7 +107,7 @@ public final class CachingRlsLbClient {
   private final ResolvedAddressFactory childLbResolvedAddressFactory;
   private final RefCountedChildPolicyWrapperFactory refCountedChildPolicyWrapperFactory;
 
-  CachingRlsLbClient(Builder builder) {
+  private CachingRlsLbClient(Builder builder) {
     helper = checkNotNull(builder.helper, "helper");
     scheduledExecutorService = helper.getScheduledExecutorService();
     synchronizationContext = helper.getSynchronizationContext();
@@ -427,7 +427,7 @@ public final class CachingRlsLbClient {
 
     abstract int getSizeBytes();
 
-    boolean isExpired() {
+    final boolean isExpired() {
       return isExpired(timeProvider.currentTimeNanos());
     }
 
@@ -571,7 +571,7 @@ public final class CachingRlsLbClient {
     private final Status status;
     private final ScheduledFuture<?> scheduledFuture;
     private final BackoffPolicy backoffPolicy;
-    private final long expireMills;
+    private final long expireNanos;
     private boolean shutdown = false;
 
     BackoffCacheEntry(RouteLookupRequest request, Status status, BackoffPolicy backoffPolicy) {
@@ -579,7 +579,7 @@ public final class CachingRlsLbClient {
       this.status = checkNotNull(status, "status");
       this.backoffPolicy = checkNotNull(backoffPolicy, "backoffPolicy");
       long delayNanos = backoffPolicy.nextBackoffNanos();
-      this.expireMills = timeProvider.currentTimeNanos() + delayNanos;
+      this.expireNanos = timeProvider.currentTimeNanos() + delayNanos;
       this.scheduledFuture =
           scheduledExecutorService.schedule(
               new Runnable() {
@@ -634,19 +634,17 @@ public final class CachingRlsLbClient {
 
     @Override
     boolean isExpired(long now) {
-      return expireMills <= now;
+      return expireNanos <= now;
     }
 
     @Override
     void cleanup() {
-      synchronized (lock) {
-        if (shutdown) {
-          return;
-        }
-        shutdown = true;
-        if (!scheduledFuture.isCancelled()) {
-          scheduledFuture.cancel(true);
-        }
+      if (shutdown) {
+        return;
+      }
+      shutdown = true;
+      if (!scheduledFuture.isCancelled()) {
+        scheduledFuture.cancel(true);
       }
     }
 
--- a/rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
+++ b/rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
@@ -449,14 +449,10 @@ public final class CachingRlsLbClient {
         updateLbState();
       } else {
         createChildLbPolicy();
-        linkedHashLruCache.updateEntrySize(request);
       }
     }
 
     private void updateLbState() {
-      checkState(
-          childPolicyWrapper.getHelper() != null,
-          "incomplete childPolicyWrapper found, this is a bug");
       childPolicyWrapper
           .getHelper()
           .updateBalancingState(
--- a/rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
+++ b/rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
@@ -896,6 +896,7 @@ public final class CachingRlsLbClient {
       String defaultTarget = lbPolicyConfig.getRouteLookupConfig().getDefaultTarget();
       if (fallbackChildPolicyWrapper == null
           || !fallbackChildPolicyWrapper.getTarget().equals(defaultTarget)) {
+        // TODO(creamsoup) wait until lb is ready
         startFallbackChildPolicy();
       }
       switch (fallbackChildPolicyWrapper.getConnectivityStateInfo().getState()) {
--- a/rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
+++ b/rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
@@ -59,7 +59,6 @@ import io.grpc.rls.internal.Throttler.ThrottledException;
 import io.grpc.stub.StreamObserver;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.CheckReturnValue;
@@ -897,19 +896,8 @@ public final class CachingRlsLbClient {
       String defaultTarget = lbPolicyConfig.getRouteLookupConfig().getDefaultTarget();
       if (fallbackChildPolicyWrapper == null
           || !fallbackChildPolicyWrapper.getTarget().equals(defaultTarget)) {
-        try {
-          startFallbackChildPolicy()
-              .await(
-                  lbPolicyConfig.getRouteLookupConfig().getLookupServiceTimeoutInMillis(),
-                  TimeUnit.MILLISECONDS);
-        } catch (InterruptedException e) {
-          Thread.currentThread().interrupt();
-          return PickResult.withError(Status.ABORTED.withDescription("interrupted"));
-        } catch (Exception e) {
-          return PickResult.withError(Status.fromThrowable(e));
-        }
+        startFallbackChildPolicy();
       }
-      SubchannelPicker picker = fallbackChildPolicyWrapper.getPicker();
       switch (fallbackChildPolicyWrapper.getConnectivityStateInfo().getState()) {
         case CONNECTING:
           return PickResult.withNoResult();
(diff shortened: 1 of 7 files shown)


rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
```diff
@@ -0,0 +1,979 @@
...
+      this.delegate = delegate;
+    }
+
+    @Override
+    public void onEviction(RouteLookupRequest key, CacheEntry value, EvictionType cause) {
+      if (delegate != null) {
+        delegate.onEviction(key, value, cause);
+      }
+      // performs cleanup after delegation
+      value.cleanup();
```
dapengzhang0, line 746: `LinkedHashLruCache` is guarded by `CachingRlsLbClient.lock`.
`LinkedHashLruCache.evictionListener.onEviction()` is guarded by `LinkedHashLruCache.lock`.
There could be deadlock:

`LinkedHashLruCache.cache() -> evictionListener.onEviction() -> value.cleanup()`

creamsoup: oh yes good catch. the lock in cleanup is not required any more. it was added when I was using invalidate with key value. i am glad  your previous finding of the invalidate(Key, Value), it fixed two issues =p.

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
```diff
@@ -0,0 +1,977 @@
...
+      this.pendingCall = pendingCall;
+      this.backoffPolicy = backoffPolicy == null ? backoffProvider.get() : backoffPolicy;
+      pendingCall.addListener(
+          new Runnable() {
+            @Override
+            public void run() {
+              handleDoneFuture();
+            }
+          },
+          synchronizationContext);
```
dapengzhang0, line 375: Why only this place is using `synchronizationContext`, whereas all other places are using `scheduledExecutorService`? Is there any guidance when to use which?

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
```diff
@@ -0,0 +1,977 @@
...
+     *
+     * <pre>
+     * Timeline                       | async refresh
+     *                                V put new cache (entry2)
+     * entry1: Pending | hasValue | staled |
+     * entry2:                        | OV | pending | hasValue | staled |
+     * </pre>
+     */
+    void maybeRefresh() {
+      synchronized (lock) {
```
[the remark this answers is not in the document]
creamsoup: hmm thats a good idea. but i am not sure why it doesn't work.

```
/home/jihuncho/git/worktree/rls_client/rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java:195: error: [GuardedBy] Access should be guarded by 'lock' in enclosing instance 'io.grpc.rls.internal.CachingRlsLbClient' of 'dataEntry', which is not accessible in this scope; instead found: 'this.lock'
```

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
```diff
@@ -0,0 +1,977 @@
...
+      long now = timeProvider.currentTimeNanos();
+      expireTime = now + maxAgeNanos;
+      staleTime = now + staleAgeNanos;
+
+      if (childPolicyWrapper.getPicker() != null) {
+        // using cached childPolicyWrapper
+        updateLbState();
+      } else {
+        createChildLbPolicy();
+        linkedHashLruCache.updateEntrySize(request);
```
dapengzhang0, line 452: Why `updateEntrySize` in constructor? It looks wired the constructor can modify linkedHashLruCache's state.

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
creamsoup: good catch! both places need to use synchronizationContext especially if it can transition to DataCacheEntry.

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
```diff
@@ -0,0 +1,977 @@
...
+        RouteLookupRequest request,
+        DataCacheEntry dataCacheEntry,
+        PendingCacheEntry pendingCacheEntry,
+        BackoffCacheEntry backoffCacheEntry) {
+      this.request = checkNotNull(request, "request");
+      this.dataCacheEntry = dataCacheEntry;
+      this.pendingCacheEntry = pendingCacheEntry;
+      this.backoffCacheEntry = backoffCacheEntry;
+      checkState((dataCacheEntry != null ^ pendingCacheEntry != null ^ backoffCacheEntry != null)
+          && !(dataCacheEntry != null && pendingCacheEntry != null && backoffCacheEntry != null),
```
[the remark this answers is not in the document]
dapengzhang0: "not all has value" is logically the same as "at least one is no value", but for the latter the expression does not need double negation, sort of a little "simpler".

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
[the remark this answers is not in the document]
dapengzhang0: Seems this usage is not supported by the current version of errorprone. Maybe keep as is.

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
creamsoup: yes it should be called outside. but for current impl it is constant, so removed it.

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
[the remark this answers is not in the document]
creamsoup: i am not saying it is logically different (i said they are same). 
you can express 2 ways to say at least one value exists using xor. a^b^c except 1 case where all has value. or, you can say a^b^c works when at least one is no value. i found out the 1st is easier to understand because maybe how i memorize the xor truth table. if you says the later one is equally easy to understand i am fine changing it, but just trying to avoid double negation defeat the purpose why it suggest that; it *usually* makes the logic harder to understand.

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
```diff
@@ -0,0 +1,970 @@
...
+
+    /** Uses Subchannel connected to default target. */
+    private PickResult useFallback(PickSubchannelArgs args) {
+      String defaultTarget = lbPolicyConfig.getRouteLookupConfig().getDefaultTarget();
+      if (fallbackChildPolicyWrapper == null
+          || !fallbackChildPolicyWrapper.getTarget().equals(defaultTarget)) {
+        try {
+          startFallbackChildPolicy()
+              .await(
+                  lbPolicyConfig.getRouteLookupConfig().getLookupServiceTimeoutInMillis(),
```
dapengzhang0, line 903: Why block for fallback policy for `lookupServiceTimeoutInMillis`, which is the timeout value for lookup service requests?

Actually this latch normally is release immediately unless there is some hanging task in the queue of sychronizationConxtext.

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
```diff
@@ -0,0 +1,962 @@
...
+    linkedHashLruCache =
+        new RlsAsyncLruCache(
+            rlsConfig.getCacheSizeBytes(),
+            builder.evictionListener,
+            scheduledExecutorService,
+            timeProvider);
+    RlsRequestFactory requestFactory = new RlsRequestFactory(lbPolicyConfig.getRouteLookupConfig());
+    rlsPicker = new RlsPicker(requestFactory);
+    rlsChannel = helper.createResolvingOobChannel(rlsConfig.getLookupService());
+    helper.updateBalancingState(ConnectivityState.CONNECTING, rlsPicker);
```
dapengzhang0, line 129: Is this run in SynchronizationContext?

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
```diff
@@ -0,0 +1,962 @@
...
+        updateLbState();
+      } else {
+        createChildLbPolicy();
+      }
+    }
+
+    private void updateLbState() {
+      childPolicyWrapper
+          .getHelper()
+          .updateBalancingState(
```
dapengzhang0, line 457: Is this run in SynchronizationContext?

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
```diff
@@ -0,0 +1,962 @@
...
+    private void createChildLbPolicy() {
+      ChildLoadBalancingPolicy childPolicy = lbPolicyConfig.getLoadBalancingPolicy();
+      LoadBalancerProvider lbProvider = childPolicy.getEffectiveLbProvider();
+      ConfigOrError lbConfig =
+          lbProvider
+              .parseLoadBalancingPolicyConfig(
+                  childPolicy.getEffectiveChildPolicy(childPolicyWrapper.getTarget()));
+
+      LoadBalancer lb = lbProvider.newLoadBalancer(childPolicyWrapper.getHelper());
+      lb.handleResolvedAddresses(childLbResolvedAddressFactory.create(lbConfig.getConfig()));
```
dapengzhang0, line 471: Is this run in SynchronizationContext?

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
creamsoup: yes this is created in handleResolvedAddresses.

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
creamsoup: i believe everything here is called under sync context except the offloaded tasks (only the PeriodicCleaner is not using sync context). verified with local integration test.

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
creamsoup: yes. see comment above.

rls/src/test/java/io/grpc/rls/internal/CachingRlsLbClientTest.java
```diff
@@ -0,0 +1,520 @@
...
+      mock(DoNotUseDirectScheduledExecutorService.class, CALLS_REAL_METHODS);
+  private final FakeTimeProvider fakeTimeProvider =
+      fakeScheduledExecutorService.getFakeTimeProvider();
+  private final StaticFixedDelayRlsServerImpl rlsServerImpl =
+      new StaticFixedDelayRlsServerImpl(
+          TimeUnit.MILLISECONDS.toNanos(SERVER_LATENCY_MILLIS), fakeScheduledExecutorService);
+  private final ChildLoadBalancingPolicy childLbPolicy =
+      new ChildLoadBalancingPolicy("target", Collections.<String, Object>emptyMap(), lbProvider);
+  private final Helper helper =
+      mock(FakeHelper.class, AdditionalAnswers.delegatesTo(new FakeHelper()));
```
dapengzhang0, line 128: `mock(Helper.class, AdditionalAnswers.delegatesTo(new FakeHelper()));` and `FakeHelper` and some other inner classes can be final.

rls/src/main/java/io/grpc/rls/internal/CachingRlsLbClient.java
```diff
@@ -0,0 +1,962 @@
...
+  }
+
+  /**
+   * Returns async response of the {@code request}. The returned value can be in 3 different states;
+   * cached, pending and backed-off due to error. The result remains same even if the status is
+   * changed after the return.
+   */
+  @CheckReturnValue
+  public final CachedRouteLookupResponse get(final RouteLookupRequest request) {
+    synchronized (lock) {
```
dapengzhang0, line 184: Can you and `synchronizationContext.throwIfNotInThisSynchronizationContext()` here?

PR #6968, opened 2020-04-23
5439: InProcessTransport optionally adds causal information to status

### LINKED ISSUE #5439: InProcessChannel with optional stack trace
1.19.0

Error Stack trace.

In reference to this stack overflow question: [here](https://stackoverflow.com/questions/55003239/grpc-java-logging-in-tests)

Is it possible to have an option to show full stack trace for InProcessChannel, since I use this in integration tests and I need to see full reason of exception?

reggiemcdonald:
Closes #5439 
Feature request to be able to add causal information to the status in `InProcessTransport` 

As per @ejona86 suggested:
- [x] Added causal information in `ServerImpl`
- [x] Optionally copy that information in `InProcessTransport`

I also 
- [x] Updated tests
- [x] Updated factories

This is my first open source PR, so any feedback is valued!


core/src/test/java/io/grpc/inprocess/InProcessTransportTestWithCause.java
```diff
@@ -0,0 +1,44 @@
...
+import static org.junit.Assert.assertEquals;
+
+import io.grpc.Status;
+import io.grpc.internal.InternalServer;
+import io.grpc.internal.ManagedClientTransport;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class InProcessTransportTestWithCause extends InProcessTransportTest {
```
ejona86, line 28: Hmm... Extending creates a lot of unnecessary tests.

Let's create a new test from scratch. I'd basically start with [HelloWorldServerTest](https://github.com/grpc/grpc-java/blob/master/examples/src/test/java/io/grpc/examples/helloworld/HelloWorldServerTest.java), with some replacements. We don't want to deal with generated code here, so we can use [voidMethod()](https://grpc.github.io/grpc-java/javadoc/io/grpc/testing/TestMethodDescriptors.html#voidMethod--) and ClientCalls and implement the stub stuff ourselves.

```java
@Rule
public final GrpcCleanupRule grpcCleanup = new GrpcCleanupRule();

...
final Throwable t = new Throwable();
ServerServiceDefinition def = ServerServiceDefinition.builder("Failing")
    .addMethod(TestMethodDescriptors.voidMethod(), new ServerCallHandler<Void, Void>() {
      @Override public ServerCall.Listener<Void> startCall(ServerCall<Void, Void> call, Metadata headers) {
        call.close(Status.UNAVAILABLE.withDescription("best error message").withCause(t), new Metadata());
        return new ServerCall.Listener() {};
      }
    })
    .build();
grpcCleanup.register(InProcessServerBuilder.forName("failing-server").directExecutor().addService(def).build().start());
Channel channel = grpcCleanup.register(InProcessChannelBuilder.forName("failing-server").directExecutor().build());
try {
  ClientCalls.blockingUnaryCall(channel, TestMethodDescriptors.voidMethod(), CallOptions.DEFAULT);
  fail("expected exception");
} catch (StatusRuntimeException ex) {
  assertEquals(t, ex.getCause());
}
```

I've just "dumped" that out there; feel free to change formatting and organization. I also don't claim it is 100% correct :smile:. But it should be close.

I know there are a lot of pieces there, but they are all virtually boilerplate. The only "interesting" things are creating the Throwable, call.close(), and catching StatusRuntimeException (which is how the blocking stub reports errors).

core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
```diff
@@ -157,11 +158,22 @@ public InProcessChannelBuilder maxInboundMetadataSize(int bytes) {
     return this;
   }
 
+  /**
+   * Sets whether to override the default behaviour of InProcessTransport to include
+   * the cause of the status.
```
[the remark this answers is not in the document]
reggiemcdonald: Good point 🙂
I added some documentation.

core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
```diff
@@ -157,11 +158,22 @@ public InProcessChannelBuilder maxInboundMetadataSize(int bytes) {
...
+   * @param enable whether to include cause in status
+   * @return this
+   */
+  public InProcessChannelBuilder transportIncludeStatusCause(boolean enable) {
```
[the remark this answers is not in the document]
reggiemcdonald: I struggled with this also. I really think `propagateCauseWithStatus` is the most descriptive! 
Also it follows nicely with the field name

core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
```diff
@@ -157,11 +158,22 @@ public InProcessChannelBuilder maxInboundMetadataSize(int bytes) {
     return this;
   }
 
+  /**
+   * Sets whether to override the default behaviour of InProcessTransport to include
+   * the cause of the status.
+   * @param enable whether to include cause in status
```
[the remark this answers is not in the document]
reggiemcdonald: done

core/src/main/java/io/grpc/inprocess/InProcessTransport.java
```diff
@@ -129,26 +131,39 @@ private InProcessTransport(String name, int maxInboundMetadataSize, String autho
         .build();
     this.optionalServerListener = optionalServerListener;
     logId = InternalLogId.allocate(getClass(), name);
+    this.includeCauseWithStatus = includeCauseWithStatus;
   }
 
   public InProcessTransport(
```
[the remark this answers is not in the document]
reggiemcdonald: done

core/src/main/java/io/grpc/inprocess/InProcessTransport.java
```diff
@@ -323,6 +338,10 @@ public ScheduledExecutorService getScheduledExecutorService() {
     return ret;
   }
 
+  public boolean getIncludeCauseWithStatus() {
```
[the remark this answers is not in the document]
reggiemcdonald: done

core/src/main/java/io/grpc/inprocess/InProcessTransport.java
```diff
@@ -83,6 +83,7 @@
   private final String userAgent;
   private final Optional<ServerListener> optionalServerListener;
   private int serverMaxInboundMetadataSize;
+  private boolean includeCauseWithStatus;
```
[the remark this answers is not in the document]
reggiemcdonald: done

core/src/main/java/io/grpc/inprocess/InProcessChannelBuilder.java
```diff
@@ -70,6 +70,7 @@ public static InProcessChannelBuilder forAddress(String name, int port) {
   private final String name;
   private ScheduledExecutorService scheduledExecutorService;
   private int maxInboundMetadataSize = Integer.MAX_VALUE;
+  private boolean transportIncludeStatusCause = false;
```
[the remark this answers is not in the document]
reggiemcdonald: Will avoid this in the future 🙂

core/src/test/java/io/grpc/inprocess/InProcessTransportTestWithCause.java
reggiemcdonald: I followed your advice with the test case. I made some small changes to it, but by and large it worked great! 

Organization-wise, I moved the new test case into the `InProcessTransportTest` class because I thought that it belonged there. But for this to work, I had to expose the field `server` in `AbstractTransportTest`: 
https://github.com/grpc/grpc-java/blob/2efad75c7b0895419d8465a1a410acb8c0b28569/core/src/test/java/io/grpc/internal/AbstractTransportTest.java#L158-L161
so that I could set `server=null`. I had to do this because the after each hook in `AbstractTransportTest` would try to call `server.shutdown` if I didnt (which would fail since I dont start that server in the test case).
https://github.com/grpc/grpc-java/blob/2efad75c7b0895419d8465a1a410acb8c0b28569/core/src/test/java/io/grpc/internal/AbstractTransportTest.java#L230-L242

So I'm not sure whether this was the right move. That being said I'm happy to change it! 

And thank you so much for all the help and pointers you've been giving me 🙂

core/src/test/java/io/grpc/internal/AbstractTransportTest.java
```diff
@@ -1219,9 +1210,6 @@ public void clientCancel() throws Exception {
     assertNotNull(clientStreamListener.trailers.get(TIMEOUT_MS, TimeUnit.MILLISECONDS));
     Status serverStatus = serverStreamListener.status.get(TIMEOUT_MS, TimeUnit.MILLISECONDS);
     assertNotEquals(Status.Code.OK, serverStatus.getCode());
-    // Cause should not be transmitted between client and server
```
ejona86, line 1222: Deleting these lines concerns me. Is this necessary?

reggiemcdonald: You're right, not necessary! I will add the assertion back in

core/src/main/java/io/grpc/inprocess/InProcessTransport.java
```diff
@@ -849,13 +853,17 @@ public void appendTimeoutInsight(InsightBuilder insight) {
    * <p>This is, so that the InProcess transport behaves in the same way as the other transports,
    * when exchanging statuses between client and server and vice versa.
    */
-  private static Status stripCause(Status status) {
+  private static Status stripCause(Status status, boolean includeCauseWithStatus) {
```
dapengzhang0, line 856: The method name, the argument, and the javadoc does not match.

reggiemcdonald: Not quite sure what you mean. Where is the discrepancy?

ejona86: I think he's referring to where it says "but stripped of any other information (i.e. cause)"

dapengzhang0: > Where is the discrepancy?

The method name says "strip cause"
The javadoc says "strip something, i.e. cause"
The new argument says "do not strip cause if true".

reggiemcdonald: Ok gotcha. I'll update the javadoc

ejona86:
@reggiemcdonald, thank you!

FYI: for the commits in response to review feedback, (at least in grpc) you don't tend to need to write much of a commit description. We can tend to figure out what is going on, and may not even read the commit description of cleanups.

In the grpc/grpc repo, those cleanup commits are kept in the final history. But in grpc/grpc-java we generally "squash" them all together into one "pristine" commit.

PR #6973, opened 2020-04-24
interop-testing: add --fail_on_failed_rpc flag support

ericgribkoff:
Some of the xDS test cases should not have any failed RPCs. This adds a flag that will cause the client to fail with an error if any expected RPCs fail. Support for setting the flag when appropriate is being added in https://github.com/grpc/grpc/pull/22763.

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/XdsTestClient.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/XdsTestClient.java
@@ -59,10 +59,11 @@ public final class XdsTestClient {
   private final Object lock = new Object();
   private final List<ManagedChannel> channels = new ArrayList<>();
 
+  private boolean failOnFailedRpc = false;
   private int numChannels = 1;
   private boolean printResponse = false;
   private int qps = 1;
-  private int rpcTimeoutSec = 2;
+  private int rpcTimeoutSec = 30;
   private String server = "localhost:8080";
   private int statsPort = 8081;
   private Server statsServer;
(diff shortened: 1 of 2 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/XdsTestClient.java
```diff
@@ -59,10 +59,11 @@
   private final Object lock = new Object();
   private final List<ManagedChannel> channels = new ArrayList<>();
 
+  private boolean failOnFailedRpc = false;
   private int numChannels = 1;
   private boolean printResponse = false;
   private int qps = 1;
-  private int rpcTimeoutSec = 2;
+  private int rpcTimeoutSec = 30;
```
dapengzhang0, line 66: Why making this change?

ericgribkoff: xDS name resolution via TD can take 10+ seconds. So a timeout of 2 seconds leads to lots of failed RPCs when the client is starting, which combined with the new flag means (incorrectly) failing tests.

dapengzhang0: xDS name resolution via TD taking 10+ seconds should be considered as a bug, seems we should not allow test result be green with it. @srini100

srini100: Is it because of the amount of time it takes between creating the resources and TD getting the configuration and pushing to the client?

ericgribkoff: @srini100 No, I don't think so. My original comment about 10+ seconds was just anecdotal as I believe I have fairly frequently seen this type of delay occasionally when manually invoking gRPC clients. Will get better data for this.

But see https://g3c.corp.google.com/results/invocations/c6d2b729-244f-4afb-bc45-7966f9b0218f/targets/github%2Fgrpc%2Freports%2Fping_pong/log for a failure with Go where xDS resolution takes > 10 seconds and the client fails (Go is currently setting the timeout to only 10s). The previous test that ran for these logs was `new_instance_group_receives_traffic`, which has just removed one (of two) instance groups from the backend service as part of its cleanup stage, but no new resources have recently been created.

larry-safran:
@ericgribkoff Can you please rebase and make sure that github recognizes the EasyCLA?

PR #6978, opened 2020-04-24
xds: create XdsServer wrapper for correct start and shutdown semantics

sanjaypujare:
@ejona86 PR for the server-wrapper we discussed the other day

### DIFF
--- a/xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
+++ b/xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
@@ -159,6 +159,7 @@ public final class XdsClientWrapperForServerSds {
   @VisibleForTesting
   public void start(XdsClient xdsClient) {
     checkState(this.xdsClient == null, "start() called more than once");
+    checkNotNull(xdsClient, "xdsClient");
     this.xdsClient = xdsClient;
     this.listenerWatcher =
         new XdsClient.ListenerWatcher() {
@@ -314,7 +315,7 @@ public final class XdsClientWrapperForServerSds {
       xdsClient = null;
     }
     if (timeService != null) {
-      timeServiceResource.close(timeService);
+      timeService = SharedResourceHolder.release(timeServiceResource, timeService);
     }
   }
 
--- a/xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
+++ b/xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
@@ -23,6 +23,7 @@ import com.google.common.annotations.VisibleForTesting;
 import io.envoyproxy.envoy.api.v2.auth.DownstreamTlsContext;
 import io.envoyproxy.envoy.api.v2.core.Node;
 import io.grpc.Internal;
+import io.grpc.InternalLogId;
 import io.grpc.Status;
 import io.grpc.SynchronizationContext;
 import io.grpc.internal.ExponentialBackoffPolicy;
@@ -64,12 +65,11 @@ public final class XdsClientWrapperForServerSds {
       new TimeServiceResource("GrpcServerXdsClient");
 
   private EnvoyServerProtoData.Listener curListener;
-  // TODO(sanjaypujare): implement shutting down XdsServer which will need xdsClient reference
   @SuppressWarnings("unused")
-  @Nullable private final XdsClient xdsClient;
+  @Nullable private XdsClient xdsClient;
   private final int port;
-  private final ScheduledExecutorService timeService;
-  private final XdsClient.ListenerWatcher listenerWatcher;
+  private ScheduledExecutorService timeService;
+  private XdsClient.ListenerWatcher listenerWatcher;
 
   /**
    * Thrown when no suitable management server was found in the bootstrap file.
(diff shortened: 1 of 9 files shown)


sanjaypujare:
Friendly ping...

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -84,41 +84,76 @@ public ManagementServerNotFoundException(String msg) {
...
-      throws IOException, ManagementServerNotFoundException {
-    Bootstrapper.BootstrapInfo bootstrapInfo = bootstrapper.readBootstrap();
-    final List<Bootstrapper.ServerInfo> serverList = bootstrapInfo.getServers();
-    if (serverList.isEmpty()) {
-      throw new ManagementServerNotFoundException("No management server provided by bootstrap");
+  public XdsClientWrapperForServerSds(int port) {
+    this.port = port;
+  }
+
+  private SynchronizationContext createSynchronizationContext(int port) {
```
creamsoup, line 95: you can just use instance variable `port`. 

```suggestion
  private SynchronizationContext createSynchronizationContext() {
```

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -84,41 +84,76 @@ public ManagementServerNotFoundException(String msg) {
...
+            panicMode = true;
+            shutdown();
+          }
+        });
+  }
+
+  /** Creates an XdsClient and starts a watch. */
+  public void start() {
+    Bootstrapper.BootstrapInfo bootstrapInfo;
+    final List<Bootstrapper.ServerInfo> serverList;
```
creamsoup, line 126: `final` is not necessary.  we discourage to use local scoped final if not required. since final should be default behavior.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -84,41 +84,76 @@ public ManagementServerNotFoundException(String msg) {
...
+      serverList = bootstrapInfo.getServers();
+      if (serverList.isEmpty()) {
+        throw new ManagementServerNotFoundException("No management server provided by bootstrap");
+      }
+    } catch (IOException | ManagementServerNotFoundException e) {
+      logger.log(Level.FINE, "Exception reading bootstrap", e);
+      logger.log(Level.INFO, "Fallback to plaintext for server at port {0}", port);
+      return;
     }
     final Node node = bootstrapInfo.getNode();
```
creamsoup, line 138: same here, no final.

xds/src/main/java/io/grpc/xds/internal/sds/ServerWrapperForXds.java
```diff
@@ -0,0 +1,109 @@
...
+  ServerWrapperForXds(Server delegate, XdsClientWrapperForServerSds xdsClientWrapperForServerSds) {
+    this.delegate = checkNotNull(delegate, "delegate");
+    this.xdsClientWrapperForServerSds =
+        checkNotNull(xdsClientWrapperForServerSds, "xdsClientWrapperForServerSds");
+  }
+
+  @Override
+  public Server start() throws IOException {
+    delegate.start();
+    xdsClientWrapperForServerSds.start();
```
creamsoup, line 47: should those be in reverse order? since server is using the xdsClientWrapperForServerSds?

xds/src/test/java/io/grpc/xds/XdsClientWrapperForServerSdsTest.java
```diff
@@ -163,7 +164,8 @@ static void generateListenerUpdateToWatcher(
   @Before
   public void setUp() throws IOException {
     MockitoAnnotations.initMocks(this);
-    xdsClientWrapperForServerSds = new XdsClientWrapperForServerSds(PORT, xdsClient, null);
+    xdsClientWrapperForServerSds = new XdsClientWrapperForServerSds(PORT);
+    xdsClientWrapperForServerSds.start(xdsClient);
```
creamsoup, line 168: can you cleanup in @After.

xds/src/test/java/io/grpc/xds/XdsClientWrapperForServerSdsTestMisc.java
```diff
@@ -55,7 +54,8 @@
   @Before
   public void setUp() throws IOException {
     MockitoAnnotations.initMocks(this);
-    xdsClientWrapperForServerSds = new XdsClientWrapperForServerSds(PORT, xdsClient, null);
+    xdsClientWrapperForServerSds = new XdsClientWrapperForServerSds(PORT);
+    xdsClientWrapperForServerSds.start(xdsClient);
```
creamsoup, line 58: same here. cleanup the resource.

xds/src/main/java/io/grpc/xds/internal/sds/ServerWrapperForXds.java
sanjaypujare: `delegate.start()` can throw an `IOException` in which case we avoid starting the `xdsClientWrapperForServerSds` unnecessarily.  In terms of server depending on xdsClientWrapperForServerSds, having the reverse order doesn't guarantee anything: only when the xDS response arrives the server is ready to use the correct SslContext.

xds/src/test/java/io/grpc/xds/XdsClientWrapperForServerSdsTest.java
sanjaypujare: `start(xdsClient)` just sets up a watch but in this case `xdsClient' is supposed to be a mock so there is nothing to clean up here. Agree?

xds/src/test/java/io/grpc/xds/XdsClientWrapperForServerSdsTestMisc.java
sanjaypujare: Same as above. There is no real client or a real connection to anything - so nothing to clean up AFAIS

xds/src/main/java/io/grpc/xds/internal/sds/ServerWrapperForXds.java
sanjaypujare: Not changing the order based on my previous comment

creamsoup: as discussed offline this raises concern about how we handle fallback in general. as the design is finalized, we will revisit.

xds/src/test/java/io/grpc/xds/XdsClientWrapperForServerSdsTest.java
creamsoup: i think `timeServiceResource` still need to be cleaned.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -84,41 +84,76 @@ public ManagementServerNotFoundException(String msg) {
...
+              return;
+            }
+            panicMode = true;
+            shutdown();
+          }
+        });
+  }
+
+  /** Creates an XdsClient and starts a watch. */
+  public void start() {
```
creamsoup, line 124: make sure the start can be called at most once. although it is programmer error, it will be very hard to debug resource leaks and random behavior due to internal reference changes.

xds/src/test/java/io/grpc/xds/XdsClientWrapperForServerSdsTest.java
sanjaypujare: `timeServiceResource` is not touched in the tests which call only `start(xdsClient)` . Mainline code calls `start()` which uses `timeServiceResource`. Pls verify

creamsoup: okay it is not used. but i don't understand why you don't want to add cleanup? what's the downside? i can't think any downsides other than few positives.

sanjaypujare: There is no downside. What is there to call in cleanup? Just call `xdsClientWrapperForServerSds.shutdown()` knowing that it will have no effect? I am asking because having superfluous code can also look bad

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
sanjaypujare: added the check in `start(XdsClient xdsClient)` which this one calls

xds/src/test/java/io/grpc/xds/XdsClientWrapperForServerSdsTest.java
creamsoup: maybe it is just for me, but calling start and not calling shutdown look bad. it almost feel like you don't call close for closeable because you maybe know the internal at that time.

sanjaypujare: I don't mind adding the shutdown calls and I will add them. But I can see someone commenting that this is superfluous code. And I remember seeing those comments before.

creamsoup: if your api design is calling shutdown only when it must be, i am fine with what you are saying.

sanjaypujare: added

xds/src/test/java/io/grpc/xds/XdsClientWrapperForServerSdsTestMisc.java
sanjaypujare: done

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
creamsoup: this is not correct because internal state (timeService) is already mutated.

sanjaypujare: you are right. I will move it to the real "start" then

sanjaypujare: Done. I had to create another `startForTestXdsClient()` method just for the tests.

xds/src/main/java/io/grpc/xds/internal/sds/ServerWrapperForXds.java
```diff
@@ -48,6 +50,12 @@ public Server start() throws IOException {
     return this;
   }
 
+  @VisibleForTesting
+  public Server startForTestXdsClient() throws IOException {
```
creamsoup, line 54: why we need this? test seems fine with using the start

sanjaypujare: This is needed for tests in `XdsSdsClientServerTest`. To avoid an actual XDS server, I use a mock XDS client that I inject using `XdsClientWrapperForServerSds.start(XdsClient xdsClient)`. So when a regular `ServerWrapperForXds.start` is called I get the "start() called more than once" exception. Under the circumstances I thought this was the best solution.

sanjaypujare: The other option is to remove the check for "start() called more than once" and instead silently return from `XdsClientWrapperForServerSds.start()` if xdsClient is set. Since end users never use `XdsClientWrapperForServerSds` directly this may not be too bad.

xds/src/main/java/io/grpc/xds/XdsClientWrapperForServerSds.java
```diff
@@ -84,41 +84,82 @@ public ManagementServerNotFoundException(String msg) {
...
   }
 
+  /** Accepts an XdsClient and starts a watch. */
   @VisibleForTesting
-  XdsClientWrapperForServerSds(int port, XdsClient xdsClient,
-      ScheduledExecutorService timeService) {
-    this.port = port;
+  public void start(XdsClient xdsClient) {
+    checkState(this.xdsClient == null, "start() called more than once");
     this.xdsClient = xdsClient;
```
creamsoup, line 162: can you add null check here?

sanjaypujare: done

PR #6987, opened 2020-04-28
core: support default method config in service config

dapengzhang0:
References:
service config spec change: https://github.com/grpc/grpc-proto/pull/75
c-core implementation https://github.com/grpc/grpc/pull/22232

### DIFF
--- a/core/src/test/java/io/grpc/internal/ServiceConfigInterceptorTest.java
+++ b/core/src/test/java/io/grpc/internal/ServiceConfigInterceptorTest.java
@@ -304,31 +304,6 @@ public class ServiceConfigInterceptorTest {
   }
 
   @Test
-  public void createManagedChannelServiceConfig_failsOnMissingServiceName() {
-    JsonObj name = new JsonObj("method", "method");
-    JsonObj methodConfig = new JsonObj("name", new JsonList(name));
-    JsonObj serviceConfig = new JsonObj("methodConfig", new JsonList(methodConfig));
-
-    thrown.expect(IllegalArgumentException.class);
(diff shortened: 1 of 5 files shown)


core/src/test/java/io/grpc/internal/ServiceConfigInterceptorTest.java
```diff
@@ -312,10 +312,7 @@ public void handleUpdate_failsOnMissingServiceName() {
     thrown.expect(IllegalArgumentException.class);
     thrown.expectMessage("missing service");
 
-    ManagedChannelServiceConfig parsedServiceConfig =
-        createManagedChannelServiceConfig(serviceConfig);
-
-    interceptor.handleUpdate(parsedServiceConfig);
+    createManagedChannelServiceConfig(serviceConfig);
```
creamsoup, line 315: rename test since it is no longer testing handleUpdate. there are few more tests need to renamed too. those should be moved to `ManagedChannelServiceConfigTest` because it is only testing ManagedChannelServiceConfig.

dapengzhang0: Good catch. Done.

dapengzhang0: Oh, you want to move them to another class. Sure, although some of the tests are out of the scope of current PR.

creamsoup: yes, moving makes more sense (only tests you just renamed). should be fairly simple to move, if not i am fine as it is.

PR #6993, opened 2020-04-30
stub: Clarify thread-safety mention in StreamObserver

ejona86:
This is basically s/write/call/ and s/application/caller/ to make the text more
precise and accurate.

### DIFF
--- a/stub/src/main/java/io/grpc/stub/StreamObserver.java
+++ b/stub/src/main/java/io/grpc/stub/StreamObserver.java
@@ -29,8 +29,8 @@ package io.grpc.stub;
  * <a href="http://www.ibm.com/developerworks/library/j-jtp09263/">thread-compatible</a>).
  * Separate {@code StreamObserver}s do
  * not need to be synchronized together; incoming and outgoing directions are independent.
- * Since individual {@code StreamObserver}s are not thread-safe, if multiple threads will be
- * writing to a {@code StreamObserver} concurrently, the application must synchronize calls.
+ * Since individual {@code StreamObserver}s are not thread-safe, if multiple threads may call
+ * a {@code StreamObserver} concurrently, the caller must synchronize the calls.
  */
 public interface StreamObserver<V>  {
   /**


stub/src/main/java/io/grpc/stub/StreamObserver.java
```diff
@@ -29,8 +29,8 @@
  * <a href="http://www.ibm.com/developerworks/library/j-jtp09263/">thread-compatible</a>).
  * Separate {@code StreamObserver}s do
  * not need to be synchronized together; incoming and outgoing directions are independent.
- * Since individual {@code StreamObserver}s are not thread-safe, if multiple threads will be
- * writing to a {@code StreamObserver} concurrently, the application must synchronize calls.
+ * Since individual {@code StreamObserver}s are not thread-safe, if multiple threads may call
+ * a {@code StreamObserver} concurrently, the caller must synchronize the calls.
  */
 public interface StreamObserver<V>  {
```
sanjaypujare, line 35: So I looked at the Javadoc for this and related classes and this is what I am thinking:

1. when gRPC provides an implementation of `StreamObserver<V>` for the "outbound" side, it is always via `CallStreamObserver<V>`, right? The Javadoc of  `CallStreamObserver<V>` ([here](https://github.com/grpc/grpc-java/blob/master/stub/src/main/java/io/grpc/stub/CallStreamObserver.java#L25)) in fact has more details of `StreamObserver<V>` which I think belong here instead of there (and we can put a reference there). 

2. `CallStreamObserver<V>` doc [here](https://github.com/grpc/grpc-java/blob/master/stub/src/main/java/io/grpc/stub/CallStreamObserver.java#L47) also has similar wording and that also needs to be fixed or enhanced since the application will be calling into it?

3. That leaves us with this class `StreamObserver<V>`. It would be nice to have an intermediate interface/class exclusively for application side to implement (similar to `CallStreamObserver<V>` on the gRPC side) but in the absence of that, we can probably write the doc here exclusively for the consumption of app developers who will implement `StreamObserver<V>`. In that case should we even say anything about "if multiple threads may call..." because gRPC is the caller and it offers these concurrency guarantees so we can just say that.

4. I remember some user asking for "happens before" guarantees on the inbound side (which is probably the reason for this PR). Are we able to say something about that based on the `SerializingExecutor` ?  "happens before" is a stronger guarantee than just synchronization.

ejona86: (1) seems unrelated to this change. I really don't want to boil the ocean in this PR. And all that language is to be able to explain 
> Implementations of this class represent the 'outbound' message streams. The client-side
one is {@ link ClientCallStreamObserver} and the service-side one is
{@ link ServerCallStreamObserver}.

"application" in (2) is actually valid, since gRPC will be the only implementer of class. It could be changed, but it also is fine as-is.

> we can probably write the doc here exclusively for the consumption of app developers who will implement StreamObserver

No, I don't think so. And I have a pretty strong reaction against "It would be nice to have an intermediate interface/class exclusively for application side to implement." We _return this interface_ to the user. So we can't say gRPC is the only caller, because it isn't. Also, just because you make a subinterface doesn't mean the superinterface no longer matters; everything in the superinterface still applies.

FWIW, this interface is [mirroring RxJava 1's Observer](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Observer.java) (note that links to reactivex.io are likely "wrong", since RxJava is on 3.x now). Note that it says nothing about thread-safety. That's because it isn't thread-safe! This paragraph was only to make it super-clear it isn't thread-safe. If we continue arguing about the paragraph it may be better to just delete the paragraph.

Oh, gosh. Not (4) again. I'm going to just close this PR.

PR #6998, opened 2020-05-01
rls: add rls-exprimnetal LoadBalancer

creamsoup:


dapengzhang0:
Want to add some tests?

creamsoup:
> Want to add some tests?

done, PTAL

rls/src/main/java/io/grpc/rls/RlsLoadBalancer.java
```diff
@@ -0,0 +1,111 @@
...
+
+  RlsLoadBalancer(Helper helper) {
+    this.helper = checkNotNull(helper, "helper");
+  }
+
+  @Override
+  public void handleResolvedAddresses(ResolvedAddresses resolvedAddresses) {
+    LbPolicyConfiguration lbPolicyConfiguration =
+        (LbPolicyConfiguration) resolvedAddresses.getLoadBalancingPolicyConfig();
+    if (lbPolicyConfiguration != null
```
dapengzhang0, line 51: Is it possible that `lbPolicyConfiguration  == null`? If possible, is no-op the right behavior in that case?

rls/src/main/java/io/grpc/rls/RlsLoadBalancer.java
```diff
@@ -0,0 +1,111 @@
...
+  public void handleNameResolutionError(final Status error) {
+    class ErrorPicker extends SubchannelPicker {
+      @Override
+      public PickResult pickSubchannel(PickSubchannelArgs args) {
+        return PickResult.withError(error);
+      }
+
+      @Override
+      public String toString() {
+        return MoreObjects.toStringHelper(this).toString();
```
dapengzhang0, line 92: Include `error`.

rls/src/test/java/io/grpc/rls/internal/RlsLoadBalancerTest.java
```diff
@@ -0,0 +1,578 @@
...
+      new SynchronizationContext(new UncaughtExceptionHandler() {
+        @Override
+        public void uncaughtException(Thread t, Throwable e) {
+          throw new RuntimeException(e);
+        }
+      });
+  private final Helper helper =
+      mock(Helper.class, AdditionalAnswers.delegatesTo(new FakeHelper()));
+  private final FakeRlsServerImpl fakeRlsServerImpl = new FakeRlsServerImpl();
+  private final Deque<FakeSubchannel> subchannels = new LinkedList<>();
```
dapengzhang0, line 115: Probably will see lint warnings for using LinkedList rather than ArrayDequeue when import.

rls/src/main/java/io/grpc/rls/RlsLoadBalancer.java
creamsoup: hmm i think it is possible if using no service config and default lb policy is rls. throwing NPE instead of ignoring it.

rls/src/test/java/io/grpc/rls/internal/RlsLoadBalancerTest.java
creamsoup: yeah, it is pretty bad they added that check. we should generally ignore it unless it is indeed right, because the warning is you are **maybe** doing wrong.

PR #7000, opened 2020-05-01
all: introduce gradle util functions to manage guava dependency

dapengzhang0:
Define util function to exclude guava's transitive dependencies jsr305 and animal-sniffer-annotations, and always manually add them as runtimeOnly dependency. error_prone_annotations is an exception: It is also excluded but manually added not as runtimeOnly. It must always compile with guava, otherwise users will see warning spams if guava is in the compile classpath but error_prone_annotations is not.

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -15,12 +15,7 @@ dependencies {
             libraries.errorprone,
             libraries.jsr305,
             libraries.animalsniffer_annotations
-    compile (libraries.guava) {
-        // prefer our own versions from libraries instead of Guava's dependency
-        exclude group: 'com.google.errorprone', module: 'error_prone_annotations'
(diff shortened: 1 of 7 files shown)


dapengzhang0:
>I'm a bit concerned that the explicit dependencies on the libraries were not cleaned up, but I understand that will be a follow-up PR. I'll just re-iterate that this change was fine in the earlier PR.

erroprone dep in grpc-api is cleaned up in the followup PR #7001. errorpone in grpc-core is kept there because it is still an implementation dependency for grpc-core although without it the build can still pass. Were there anything else need be cleaned up?

ejona86:
I think with those, the only one left would be animalsniffer_annotations in grpc-api looks unused.

dapengzhang0:
Yes, animalsniffer_annotations in grpc-api is also cleaned up in the followup PR.

PR #7004, opened 2020-05-05
rls: generate rls coverage

creamsoup:


rls/build.gradle
```diff
@@ -19,7 +20,26 @@ dependencies {
...
+    }
+}
+
 configureProtoCompilation()
 
 // do not publish 'grpc-rls'
 [publishMavenPublicationToMavenRepository]*.onlyIf { false }
+
+jacocoTestReport {
+    reports {
```
dapengzhang0, line 40: Why add this different setting than other subprojects like grpc-core?

creamsoup: removed.

PR #7008, opened 2020-05-05
rls: add guava dependency

creamsoup:


rls/build.gradle
```diff
@@ -1,5 +1,5 @@
 plugins {
-    id "java"
+    id "java-library"
```
dapengzhang0, line 2: This will not do anything more than `java` plugin unless you use `api` dependency. grpc-lrs is like grpc-grpc-lb and grpc-xds, they don't provide public APIs for users. So `java-library` is not necessary.

creamsoup: ah okay that makes sense.

PR #7012, opened 2020-05-06
all: disable compileClasspath transitivity for gradle build

dapengzhang0:
Set `compileClasspath.transitive = false` and then add all dependencies that have to be added to each subproject.

- Benefit of setting `compileClasspath.transitive = false`:
  + Boost incremental compilation time because the subproject will no longer load the whole transitive dependency tree to compile, instead it only loads dependencies that are explicitly added for the subproject.
  + The compile time dependency structure of a subproject will be flat and more clear, no longer be a tree structure. If our code depends on X and Y with X transitively depends on Y, we have to explicitly add both X and Y in order to compile. More like Bazel. Previously the build can still pass even if Y is not explicitly added.
- Downside:
  + A number of individual io.netty artifacts have to be added for netty.
  + Whenever we use `SettableFuture.create()`, we have to add `libraries.guava_failureaccess` and `libraries.j2objc_annotations` in order to compile.

### DIFF
--- a/android-interop-testing/build.gradle
+++ b/android-interop-testing/build.gradle
@@ -58,6 +58,7 @@ dependencies {
     implementation 'com.android.support:support-fragment:26.1.0'
     implementation 'com.google.android.gms:play-services-base:15.0.1'
     implementation 'com.google.android.gms:play-services-basement:15.0.1'
+    implementation 'org.hamcrest:hamcrest-core:1.3'
 
     implementation project(':grpc-api'),
             project(':grpc-auth'),
@@ -69,6 +70,8 @@ dependencies {
             project(':grpc-stub'),
             project(':grpc-testing'),
             libraries.google_auth_credentials,
+            libraries.guava_failureaccess,
+            libraries.j2objc_annotations,
             libraries.jsr305,
             libraries.junit,
             libraries.protobuf_lite,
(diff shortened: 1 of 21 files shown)


android-interop-testing/build.gradle
```diff
@@ -48,26 +48,43 @@ android {
...
 
     implementation (libraries.google_auth_oauth2_http) {
         exclude group: 'org.apache.httpcomponents'
     }
+    censusApiDependency 'implementation'
     censusGrpcMetricDependency 'implementation'
+    guavaDependency 'implementation'
 
-    compileOnly libraries.javax_annotation
+    compileOnly 'org.hamcrest:hamcrest-core:1.3',
```
ejona86, line 84: I'm very concerned to see all these `compileOnly` deps. `compileOnly` deps are very hard to manage. Why are these `compileOnly` instead of `implementation`?

dapengzhang0: With `implementation`, those dependency will be explicitly included (with runtime scope) in pom file. I'm feeling it looks weird. Otherwise I'm okay with `implementation`.

dapengzhang0: Now moved to `implementation`, PTAL.

ejona86: So you had made them compileOnly to make the POM make more sense?

dapengzhang0: Yes. But maybe it did not make more sense.

PR #7014, opened 2020-05-06
rls: allow test to provide default caching client builder

creamsoup:


rls/src/main/java/io/grpc/rls/RlsLoadBalancer.java
```diff
@@ -106,4 +111,21 @@ public void shutdown() {
       routeLookupClient = null;
     }
   }
+
+  /**
+   * Provides {@link CachingRlsLbClient.Builder} with default settings. This is useful for
+   * testing.
+   */
+  interface CachingRlsLbClientBuilderProvider {
```
dapengzhang0, line 119: Seems all you want is to inject throttle, ]why not just `ThrottlerProvider`?

creamsoup: yeah that was my initial thought. Sooner than later we may need more and more of those Providers. Providing default builder is more flexible without adding test specific codes/classes. e.g. I am thinking we may need to add timeProvider and evictionListener in near future.

PR #7019, opened 2020-05-07
rls: add bazel build rule

creamsoup:


rls/BUILD.bazel
```diff
@@ -0,0 +1,27 @@
...
+    deps = [
+        ":rls_java_grpc",
+        "//api",
+        "//core",
+        "//core:internal",
+        "//core:util",
+        "//stub",
+        "@com_google_code_findbugs_jsr305//jar",
+        "@com_google_guava_guava//jar",
+        "@io_grpc_grpc_proto//:rls_java_proto",
```
dapengzhang0, line 18: If other languages make a breaking change to grpc-proto and grpc-java will fail, we they notice in the first time?

creamsoup: at least, they will notice when they are trying to import to google3. although there shouldn't be any breaking changes.

Issue #7024, opened 2020-05-08
How can I share some value without sharing Context.Key object instance?


hojongs:
### Summary
How can I store and get a value from Context without sharing Context.Key instance?

### Detail

I want to share some value with grpc Context. but I can't get the stored value with Context.Key instance

I tried this snippet
```kotlin
Context.key("A") == Context.key("A")
```

The result was
 ```
false
```

But I expected 
```
true
```

It means if I want to get value from a Context, I should use same Context.Key instance.
It is strange. Because, I'm trying to share some values with Context and when I get the stored value I should use same Context.Key.
To share some value, should share the key object instance. (not the key value)

For example (with kotlin)

```kotlin
// Function A
val myKey = Context.key("myKey")

Context.current()
    .withValue(key, "myVal")
    .attach()
```

```kotlin
// Function B, After call A
Context.key<String>("myKey").get() // == null // I think, It should be "myVal"

myKey.get() // == "myVal" 
```

If I can share `myKey` object instance, I don't need Context to share value.
What about multi-threaded case? It is more difficult to share the Context.Key object instance.

I can't understand use of Context and Context.key to share value.

### In short, My question is.
How can I store and get a value from Context without sharing Context.Key instance?

ericgribkoff:
You're correct that multiple `Context.Key` can share the same debug name but still won't return true when tested for equality. From the doc at https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.html#key-java.lang.String-:

> Create a Context.Key with the given debug name. Multiple different keys may have the same name; the name is intended for debugging purposes and does not impact behavior.

You need to use the same `Context.Key` reference when inserting and retrieving values from the context. A simple way to achieve this is to create your key as essentially a global constant and use it when setting and retrieving values. I didn't look through it carefully but https://github.com/saturnism/grpc-by-example-java/tree/master/metadata-context-example shows example code that uses `Context.Key` to propagate values.

hojongs:
Thank you for your answer. I was missing the doc of Context.key API
But I think it is confuse even now.
Generally, key's name will be recognized it is identifier by people (In my opinion)
the name parameter of Context.key(name) is debugging name, so Let me give a suggestion.
- Make Context.key(name) function have default parameter value (the null)
  - or overload key function without any parameter.
  - It represents the name parameter is optional.
```java
    Key() {
      this(null, null);
    }
```
```java
    Key(String name = null) {
      this(name, null);
    }
```

[...]

PR #7027, opened 2020-05-09
grpc-context: add a option to create Context.Key with default name

hojongs:
## Motivation
- Motivated Issue : #7024 
### Background
  - Context.key(name) is the function to create Context.Key object instance.
  - the name parameter of Context.key(name) is used as debug name and does not impact behavior.
### The Main Point
  - But, Users cannot create Context.Key without specifying debug name
  - It means user must specify the debug name of Context.Key although the user don't want to use it.
  - It is better that a parameter for debug (like the name parameter) is being optional
  - It will make users recognize `multiple different keys may have a same name` from the code before they see the document of the function

### DIFF
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -153,19 +153,21 @@ public class Context {
   }
 
   /**
-   * Create a {@link Key} with the given debug name. Multiple different keys may have the same name;
-   * the name is intended for debugging purposes and does not impact behavior.
+   * Create a {@link Key} with the given debug name.
+   * @param debugString a name is intended for debugging purposes and does not impact behavior;
+   *                    Multiple different keys may have the same debugString.
    */
-  public static <T> Key<T> key(String name) {
-    return new Key<>(name);
+  public static <T> Key<T> key(String debugString) {
+    return new Key<>(debugString);
   }
 
   /**
-   * Create a {@link Key} with the given debug name and default value. Multiple different keys may
-   * have the same name; the name is intended for debugging purposes and does not impact behavior.
+   * Create a {@link Key} with the given debug name and default value.
+   * @param debugString a name is intended for debugging purposes and does not impact behavior;
+   *                    Multiple different keys may have the same debugString.
    */
-  public static <T> Key<T> keyWithDefault(String name, T defaultValue) {
-    return new Key<>(name, defaultValue);
+  public static <T> Key<T> keyWithDefault(String debugString, T defaultValue) {
+    return new Key<>(debugString, defaultValue);
   }
 
   /**
--- a/context/src/main/java/io/grpc/Context.java
+++ b/context/src/main/java/io/grpc/Context.java
@@ -18,8 +18,6 @@ package io.grpc;
 
 import io.grpc.Context.CheckReturnValue;
 import io.grpc.PersistentHashArrayMappedTrie.Node;
-
-import javax.annotation.Nonnull;
 import java.io.Closeable;
 import java.util.ArrayList;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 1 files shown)


context/src/main/java/io/grpc/Context.java
```diff
@@ -152,6 +152,14 @@ private static Storage createStorage(
     }
   }
 
+  /**
+   * Create a {@link Key} with {@value io.grpc.Context.Key#DEFAULT_NAME} value as a debug name. Multiple different keys may have the same name;
+   * the name is intended for debugging purposes and does not impact behavior.
```
ericgribkoff, line 157: A disadvantage of this API is that, by default, debugging issues involving more than `Context.Key` would become very difficult as they would all have the same name. I think we want to instead encourage the use of different identifiers when creating the keys.

context/src/main/java/io/grpc/Context.java
```diff
@@ -946,9 +954,14 @@ public void close() {
    * Key for indexing values stored in a context.
```
ericgribkoff, line 954: Rather than adding the parameter-less `#key()` API above, it might be more helpful to avoid confusion around whether the name of a key "matters" (as in https://github.com/grpc/grpc-java/issues/7024) if the Javadoc for `Context.Key` specifies here as well that `Key` values will only be compared using by reference equality, which otherwise is only specified on the `Context#key` creation methods. 

@ejona86 What do you think?

ejona86: We can't really be reviewing code that hasn't had the CLA signed, but I'm fine with adding documentation to `Context.Key`. I'm not sure it will help _much_, but it doesn't hurt.

What looks broken is that we are still saying 'name' in `Context.key(String name)`. That should be `Context.key(String debugName)`. Do you think that'd help?

ejona86: Actually, `debugString` to match [`Attributes.Key`](https://grpc.github.io/grpc-java/javadoc/io/grpc/Attributes.Key.html).

ejona86: See also https://github.com/grpc/grpc-java/issues/1764#issuecomment-348630514

hojongs: Thank you for fast and helpful your feedback.

I also think it is more safe and useful to change parameter name rather than to add parameter-less `Context#key()`.

I updated my code. I think it will clarify the purpose of the parameter of `#key(debugString)`.

Please review the new change. Thank you.

hojongs:
I want to add `@Nonnull` annotation for the debugString parameter to avoid misusage of API.
But there is a issue that javax.annotation.Nonnull cannot be imported.

How can I add information to mark that the parameter cannot be null?
Maybe.. to call checkNotNull in Context#key() API is better than Nonnull annotation?

Travis CI job log
```
/home/travis/build/grpc/grpc-java/context/src/main/java/io/grpc/Context.java:22: error: cannot find symbol import javax.annotation.Nonnull;
```

ejona86:
grpc-context is our only module that has no dependencies, and we want to keep it that way to ease using it. We can only use documentation and runtime checks. So that means no `@Nonnull`.

context/src/main/java/io/grpc/Context.java
```diff
@@ -153,19 +155,21 @@ private static Storage createStorage(
   }
 
   /**
-   * Create a {@link Key} with the given debug name. Multiple different keys may have the same name;
-   * the name is intended for debugging purposes and does not impact behavior.
+   * Create a {@link Key} with the given debug name.
+   * @param debugString a name is intended for debugging purposes and does not impact behavior;
```
ejona86, line 159: "a name intended for debugging"

context/src/main/java/io/grpc/Context.java
```diff
@@ -153,19 +155,21 @@ private static Storage createStorage(
...
+   *                    Multiple different keys may have the same debugString.
```
ejona86, line 160: If you use a semicolon, then Multiple should be lower-case. It is probably better to put a period after the first part and have this capitalized.

hojongs:
Thank you, Fixed the code as your review above.
and I added a line in javadoc about not null.

ejona86:
@hojongs, thank you!

PR #7029, opened 2020-05-12
Set `SETTINGS_MAX_FRAME_SIZE` appropriately during initial handshake

alexeykudinkin:
Currently `NettyClientTransport.maxMessageSize` isn't set as [`SETTINGS_MAX_FRAME_SIZE`](https://http2.github.io/http2-spec/#SETTINGS_MAX_FRAME_SIZE) within `SETTINGS` handshake therefore capping all the incoming and outgoing data-frames to be of size no more than default cap (16384);

 This change sets it properly, making sure it's advertised appropriately during the handshake.

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
+++ b/netty/src/main/java/io/grpc/netty/NettyClientHandler.java
@@ -136,11 +136,13 @@ class NettyClientHandler extends AbstractNettyHandler {
       boolean autoFlowControl,
       int flowControlWindow,
       int maxHeaderListSize,
+      int maxFrameSize,
       Supplier<Stopwatch> stopwatchFactory,
       Runnable tooManyPingsRunnable,
       TransportTracer transportTracer,
       Attributes eagAttributes,
-      String authority) {
+      String authority
+  ) {
     Preconditions.checkArgument(maxHeaderListSize > 0, "maxHeaderListSize must be positive");
     Http2HeadersDecoder headersDecoder = new GrpcHttp2ClientHeadersDecoder(maxHeaderListSize);
     Http2FrameReader frameReader = new DefaultHttp2FrameReader(headersDecoder);
@@ -161,28 +163,32 @@ class NettyClientHandler extends AbstractNettyHandler {
         autoFlowControl,
         flowControlWindow,
         maxHeaderListSize,
+        maxFrameSize,
         stopwatchFactory,
         tooManyPingsRunnable,
         transportTracer,
         eagAttributes,
-        authority);
+        authority
+    );
   }
 
   @VisibleForTesting
   static NettyClientHandler newHandler(
-      final Http2Connection connection,
-      Http2FrameReader frameReader,
-      Http2FrameWriter frameWriter,
-      ClientTransportLifecycleManager lifecycleManager,
-      KeepAliveManager keepAliveManager,
-      boolean autoFlowControl,
-      int flowControlWindow,
-      int maxHeaderListSize,
-      Supplier<Stopwatch> stopwatchFactory,
-      Runnable tooManyPingsRunnable,
-      TransportTracer transportTracer,
-      Attributes eagAttributes,
-      String authority) {
+          final Http2Connection connection,
+          Http2FrameReader frameReader,
+          Http2FrameWriter frameWriter,
+          ClientTransportLifecycleManager lifecycleManager,
+          KeepAliveManager keepAliveManager,
+          boolean autoFlowControl,
+          int flowControlWindow,
+          int maxHeaderListSize,
+          int maxFrameSeize,
+          Supplier<Stopwatch> stopwatchFactory,
+          Runnable tooManyPingsRunnable,
+          TransportTracer transportTracer,
+          Attributes eagAttributes,
+          String authority
+  ) {
     Preconditions.checkNotNull(connection, "connection");
     Preconditions.checkNotNull(frameReader, "frameReader");
     Preconditions.checkNotNull(lifecycleManager, "lifecycleManager");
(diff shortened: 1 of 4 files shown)


creamsoup:
the use of `maxMessageSize` is for actual payload size restriction (inbound serialized message size - excluding headers etc). So, if we really want to configure the max frame size, we need to introduce a new parameter.

why do you want to configure the max frame size? we want to learn the use case to determine if we want to introduce the new API or not.

alexeykudinkin:
@creamsoup we want to make it configurable so that we could minimize number of hops required to transfer large payloads, since our use-cases is latency sensitive and we have high RTT cost (higher than we would like to have).

Motivational part is well captured in this PR for [grpc-go](https://github.com/grpc/grpc-go/issues/1043#issuecomment-304133038)

alexeykudinkin:
@creamsoup DATA frames [aren't including headers](https://http2.github.io/http2-spec/#DATA) either. Are you referring to gRPC headers?

creamsoup:
i am a bit confused. how does the packet size affect the number of hops? 

it sounds like you want to increase [flowControlWindowSize](https://grpc.github.io/grpc-java/javadoc/io/grpc/netty/NettyChannelBuilder.html#flowControlWindow-int-) to increase overall network utilization. Larger packet size will reduce some overhead, so it will be helpful. but, the gain will be negligible unless the network utilization is already high. Given the high RTT, i imagine the utilization is very low.

What I meant for the `maxInboundMessageSize` is it only considers the protobuf serialized message size.

alexeykudinkin:
@creamsoup we are also increasing `flowControlWindowSize`.

We want to set `MAX_FRAME_SIZE` appropriately to signal to server that we're ready to receive larger frames, to avoid jittering with 16Kb ones when payload is large.

alexeykudinkin:
Sorry for confusion, by "hop" i referred to how many frames from HTTP/2 perspective being casted to the client.

alexeykudinkin:
@creamsoup why do we want to keep those terms separate? And how you want to disambiguate between those terms?

creamsoup:
do we know how much gain from changing packet size?

alexeykudinkin:
@creamsoup you can follow the discussion in issue created for [grpc-go](https://github.com/grpc/grpc-go/issues/1043).

We want to have the knob similar to the one in grpc-go introduced in [PR#1210](https://github.com/grpc/grpc-go/pull/1210)

creamsoup:
that go PR is configuring flow control window size. go is using 64k by default and BDP tunes it up to 16MB IIRC. java just enabled BDP by default starting from 1MB to 8MB (will be available on `1.30.0`).

alexeykudinkin:
@creamsoup sorry, you're right, i was looking at something else and just mixed it up: it's indeed does not touch `SETTINGS_MAX_FRAME_SIZE` adjusting only `SETTINGS_INITIAL_WINDOW_SIZE`

Regardless of it i think it's still beneficial from the performance standpoint leverage elasticity of the DATA frames and sent larger payloads in bigger frames rather than fragmenting it into 16Kb chunks.

creamsoup:
Unless we understand the performance benefit well enough (probably need benchmarking), i don't think it is a good idea to expose this option. I suspect that the gain is not significant enough to expose side effects of large/small packet size.

If you really think this will be beneficial, you should file an issue first. so, we can gather interests and thought about this topic.

alexeykudinkin:
@creamsoup what are the side-effects you're referring to?

creamsoup:
larger frame size can decrease multiplexing efficiency hence the http2 supports multiple streams. it can affect many things including overall concurrency level of application. also affect small message latency etc.

alexeykudinkin:
Which is very similar to `flowControlWindowSize` impact, yet we exposed as it's meaningful lever to optimize the application performance.

I agree that benchmarks would definitely not hurt to have, but i struggle to see why this setting should be treated any differently than FCW size.

creamsoup:
can you explain why it is similar to `flowControlWindowSize`?

alexeykudinkin:
Sorry, not sure that i understand your question completely: this is one of the levers HTTP/2 spec leaves for implementations to decide upon the values most suitable for their use-cases, right?

I just don't see how it's any different from FCW setting: they both allow you to trade in resources consumption for potential performance benefits. Yes, FCW is clearly more powerful lever as it involves synchronization between endpoints. `MAX_FRAME_SIZE` on the flip side could allow us to shave off CPU cycles on framing/deframing bringing it closer to the performance of the low-level transport protocol itself (assuming large enough FCW).

creamsoup:
why don't you create an issue with concrete use case? what you are claiming doesn't have any concrete use case. thus, it is unclear it is actually beneficial or not.

There are countless options from many networking layers. We can't simply add options just because it is configurable. Adding them just decreases use-ability of gRPC with little to no benefit.

alexeykudinkin:
I gave very concrete use-case: fetching medium to large payloads from the server side with high concurrency (tens to hundreds of Gbps) over networks with high RTT cost.

What additional details do you see helping to benefit the conversation. I can create an issue if you think this will be helpful.

>  Adding them just decreases use-ability of gRPC with little to no benefit.

Can you please substantiate this with an example on how this will decrease the usability?

creamsoup:
that's not a use case of the http2 frame size option. you **suspect** that it **may** help which is no where near concrete.  this conversation is not productive in wrong place. like i said, you should file an issue if you believe it is beneficial enough to expose it.

creamsoup:
Closing this PR since it seems like there was a confusion between message size and http2 frame size.

alexeykudinkin:
@creamsoup sent you an email with more context around this, which we can't share here unfortunately.

PR #7030, opened 2020-05-12
interop-testing: adds assertions to empty_stream test case.

plaflamme:
If the server fails (e.g.: with an unimplemented status), the test still succeeds.
This change adds assertions to check the status code as well as the expected number of responses.

### DIFF
--- a/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
+++ b/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
@@ -790,6 +790,8 @@ public abstract class AbstractInteropTest {
         = asyncStub.fullDuplexCall(responseObserver);
     requestObserver.onCompleted();
     responseObserver.awaitCompletion(operationTimeoutMillis(), TimeUnit.MILLISECONDS);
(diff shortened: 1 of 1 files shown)


interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java
```diff
@@ -790,6 +790,8 @@ public void emptyStream() throws Exception {
         = asyncStub.fullDuplexCall(responseObserver);
     requestObserver.onCompleted();
     responseObserver.awaitCompletion(operationTimeoutMillis(), TimeUnit.MILLISECONDS);
+    assertSuccess(responseObserver);
```
ericgribkoff, line 793: Interesting, looks like this test stopped testing for these conditions in https://github.com/grpc/grpc-java/commit/6fbe140959e2117c77ebcf45cff503678881467e#diff-957966121c9393c2a61716d02c6ccf54R393. This brings things back inline with the [spec](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#empty_stream), thanks!

PR #7038, opened 2020-05-13
Fixes #7006 Upgrade Gradle distribution to allow using Java 13.

### LINKED ISSUE #7006: Upgrade to Gradle 6 to support Java 13
### Is your feature request related to a problem?

If you run gradle with Java 13, you will get "Unsupported class file major version 57". This can be fixed by changing to Gradle 6.

### Describe the solution you'd like

Patch:
diff --git a/gradle/wrapper/gradle-wrapper.properties b/gradle/wrapper/gradle-wrapper.properties
index 7c4388a92..562e2c884 100644
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.0-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists

perty:


examples/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.0-all.zip
```
ericgribkoff, line 3: This should stay as bin, `gradle-6.0-bin.zip`.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.0-all.zip
```
ericgribkoff, line 3: This should stay as bin, gradle-6.0-bin.zip.

perty: Looking at it but struggling with compiler errors in the examples. Are they for a future 1.30.0 version?

ejona86:
I think we can just disable gradle module metadata for our entire build (so for the root build.gradle). https://docs.gradle.org/current/userguide/publishing_gradle_module_metadata.html#sub:disabling-gmm-publication

We could maybe disable it for just alts (and any other problematic projects; probably netty-shaded is broken), but we should really understand how it works and any ramifications before enabling it. So it is probably for the best to disable it everywhere.

perty:
Giving up on this. Perhaps the Android version should be separate from Java.

ericgribkoff:
@perty This actually doesn't involve the Android build and is just an issue with Gradle's new (in 6.0) module metadata feature. As @ejona86 suggested, for now this feature can just be disable via adding the following to our `grpc-java/build.gradle` file:

```
    tasks.withType(GenerateModuleMetadata) {
        enabled = false
    }
```

I tested this locally with gradle 6.5.1 and everything seems good. As far as I can tell it looks like I can't directly edit the changes in this PR (did you delete your fork/branch?) but I pushed the necessary changes to my own repo as in https://github.com/grpc/grpc-java/compare/master...ericgribkoff:update_gradle?expand=1. If you'd like to update this PR to do the same, I think we should be able to get this in.

Sorry for the long delay in getting back to you on this!

perty:
No worries.

I deleted my fork, yes, because of an unrelated event.  I will not do
anything, keep going.


Den ons 1 juli 2020 kl 23:42 skrev Eric Gribkoff <notifications@github.com>:

ericgribkoff:
@petry Ok, thanks for starting on this. Closing this PR in favor of https://github.com/grpc/grpc-java/pull/7183

PR #7040, opened 2020-05-13
xds: provide fallback protocol negotiator

sanjaypujare:


xds/src/main/java/io/grpc/xds/internal/sds/XdsServerBuilder.java
```diff
@@ -120,6 +124,12 @@ public XdsServerBuilder intercept(ServerInterceptor interceptor) {
     return this;
   }
 
+  public XdsServerBuilder fallbackProtocolNegotiator(
+      ProtocolNegotiator fallbackProtocolNegotiator) {
+    this.fallbackProtocolNegotiator = fallbackProtocolNegotiator;
```
creamsoup, line 129: null check here would be nice.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
```diff
@@ -87,6 +92,34 @@ public void plaintextClientServer() throws IOException, URISyntaxException {
     assertThat(unaryRpc("buddy", blockingStub)).isEqualTo("Hello buddy");
   }
 
+  @Test
+  public void fallbackProtocolNegotiator_expectException() throws IOException, URISyntaxException {
```
creamsoup, line 96: i am not sure what this is testing. it provides fallbackPN that throws "no fallback negotiation" exception?

xds/src/main/java/io/grpc/xds/internal/sds/XdsServerBuilder.java
```diff
@@ -43,6 +46,7 @@
 
   private final NettyServerBuilder delegate;
   private final int port;
+  private ProtocolNegotiator fallbackProtocolNegotiator;
```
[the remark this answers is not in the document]
sanjaypujare: Yes, I was rethinking null check in `build()` as well. Providing a default for fallback doesn't sound right: they both have similar meaning:  fall back to something or default to something.  So why provide a default for fallback? In that case no fallback is the way to go - which means failing the connection if the Sds PN fails. Is that right?

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: The test makes sure the fallback PN is invoked in an easy way. I can see the contradiction of the fallback PN throwing exception saying "no fallback exception". I can change the text message. In connection with the other thread: if fallback PN is now optional, I can use the same fallback PN (as internal default) that throws an exception? Or else the HandlerPickerHandler can generate an error when there is no fallback PN. Which is better?

xds/src/main/java/io/grpc/xds/internal/sds/XdsServerBuilder.java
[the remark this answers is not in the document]
creamsoup: yes, i think that is natural behavior or at least better than not able to construct XdsServer.
no fallback PN -> no fallback when SdsPN fails.

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
creamsoup: mock PN can be better fit here. `verify(mockFallbackPn).newHandler();`.
The thrown exception should be able to converted to meaningful `Status`, probably `UNAVAILABLE`. the conversion logic is in `WriteBufferingAndExceptionHandler#exceptionCaught`. you can do it in default FallbackPN or `HandlerPickerHandler` whichever makes more sense.

sanjaypujare:
> Looks fair. The server-side builder fallback will change API, but the plumbing inside the ProtocolNegotiator is the important thing for now. We are going to have XdsServerBuilder being the equivalent of xds:// on client-side, and then we'll either pass in XdsCredentials (assuming server credentials exist) or call a method to enable xds credentials (if server credentials don't exist).

We should also remember to rename `XdsServerBuilder` to something else. This is NOT building an xDS server (like TD or Pilot) but a regular gRPC server that is "managed" by xDS. So how about `XdsManagedServerBuilder`? And similar changes to other classes (if needed).

ejona86:
It's the difference between "(xDS Server) Builder" and "xDS (Server Builder)" (or written in plain English, xDS-Server Builder and xDS Server Builder; not to be confused with xDS, Server Builder!) Given it is a ServerBuilder, I don't feel XdsServerBuilder is inappropriate. This will be a cross-language discussion, because we were planning "XdsServerBuilder" for the wrapped languages as well.

sanjaypujare:
> It's the difference between "(xDS Server) Builder" and "xDS (Server Builder)" (or written in plain English, xDS-Server Builder and xDS Server Builder; not to be confused with xDS, Server Builder!) Given it is a ServerBuilder, I don't feel XdsServerBuilder is inappropriate. This will be a cross-language discussion, because we were planning "XdsServerBuilder" for the wrapped languages as well.

Well, naming is subjective so I can't argue with it (and it doesn't affect functionality!) but I suspect many folks would agree that `XdsServerBuilder` is not appropriate in this case.

xds/src/main/java/io/grpc/xds/internal/sds/XdsServerBuilder.java
sanjaypujare: Hmmm, on second thoughts since we are making `fallbackProtocolNegotiator` optional (in the builder) if somebody wants to "nullify" the previously set value, shouldn't we allow null here?

creamsoup: that's little bit subjective. but setting and unsetting on the same builder is weird. calling this has intention of setting Fallback PN, so it is okay to not accept null imo.

sanjaypujare: But I can call it multiple times with different values - the last value wins. It doesn't make sense to disallow null in that case. If I need to clear the previously set value in a different piece of code (over which I have no control) then the only thing I can do is to call this with null assuming it is allowed.

creamsoup: i don't really care which way you use because both has pros and cons. iirc, this topic was mentioned in the api meeting briefly.
make sure this is described in the javadoc.

xds/src/test/java/io/grpc/xds/internal/sds/SdsProtocolNegotiatorsTest.java
```diff
@@ -239,10 +245,11 @@ public SocketAddress localAddress() {
   }
 
   @Test
-  public void serverSdsHandler_nullTlsContext_expectPlaintext() throws IOException {
+  public void serverSdsHandler_nullTlsContext_expectFallbackProtocolNegotiator() {
     SdsProtocolNegotiators.HandlerPickerHandler handlerPickerHandler =
         new SdsProtocolNegotiators.HandlerPickerHandler(
-            grpcHandler, /* xdsClientWrapperForServerSds= */ null);
+            grpcHandler, /* xdsClientWrapperForServerSds= */ null,
+            new FallbackProtocolNegotiator());
```
creamsoup, line 252: no reply for my previous comment. consider mock here. you can `verify` newHandler is called or not. this is cleaner and we can remove L316-344.

sanjaypujare: Your previous comment was for a different file/test `XdsSdsClientServerTest#nullFallbackProtocolNegotiator_expectException` I renamed the test and also changed it. Also that test verifies the server closes so the clients can't connect any more.

For this one: yes, I will look into using a mock - I think it should work.

xds/src/main/java/io/grpc/xds/internal/sds/XdsServerBuilder.java
ejona86: We normally allow people to "undo" changes to the builder, so if null is okay, allowing null is fine. But I'd just remind you not to send too much effort on this part of the API because it will be replaced.

sanjaypujare: `null` is okay so I am going to allow null to override a previous non-null value (and mention so in javadocs)

xds/src/test/java/io/grpc/xds/XdsSdsClientServerTest.java
sanjaypujare: Just used `null` and verified that server is shutdown because of missing fallback PN

xds/src/test/java/io/grpc/xds/internal/sds/SdsProtocolNegotiatorsTest.java
```diff
@@ -239,10 +244,14 @@ public SocketAddress localAddress() {
...
-  public void serverSdsHandler_nullTlsContext_expectPlaintext() throws IOException {
+  public void serverSdsHandler_nullTlsContext_expectFallbackProtocolNegotiator() {
+    ChannelHandler mockChannelHandler = mock(ChannelHandler.class);
+    ProtocolNegotiator mockProtocolNegotiator = mock(ProtocolNegotiator.class);
+    when(mockProtocolNegotiator.newHandler(grpcHandler)).thenReturn(mockChannelHandler);
     SdsProtocolNegotiators.HandlerPickerHandler handlerPickerHandler =
         new SdsProtocolNegotiators.HandlerPickerHandler(
-            grpcHandler, /* xdsClientWrapperForServerSds= */ null);
+            grpcHandler, /* xdsClientWrapperForServerSds= */ null,
+            mockProtocolNegotiator);  // new FallbackProtocolNegotiator()
```
creamsoup, line 254: nit: this comment (ed code?) is confusing

sanjaypujare: oops, good catch. Will fix

Issue #6423, opened 2019-11-13
gRPC Server Exception : Connection reset


halimpuckjava:
hi

why each time client send request to serer, sever respond with internal exeption: 
```
INFO: Transport failed
java.net.SocketException: Connection reset
	at java.base/sun.nio.ch.SocketChannelImpl.throwConnectionReset(SocketChannelImpl.java:345)
	at java.base/sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:376)
	at io.grpc.netty.shaded.io.netty.buffer.PooledByteBuf.setBytes(PooledByteBuf.java:247)
	at io.grpc.netty.shaded.io.netty.buffer.AbstractByteBuf.writeBytes(AbstractByteBuf.java:1147)
	at io.grpc.netty.shaded.io.netty.channel.socket.nio.NioSocketChannel.doReadBytes(NioSocketChannel.java:347)
	at io.grpc.netty.shaded.io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:148)
	at io.grpc.netty.shaded.io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:700)
	at io.grpc.netty.shaded.io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:635)
	at io.grpc.netty.shaded.io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:552)
	at io.grpc.netty.shaded.io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:514)
	at io.grpc.netty.shaded.io.netty.util.concurrent.SingleThreadEventExecutor$6.run(SingleThreadEventExecutor.java:1044)
	at io.grpc.netty.shaded.io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at io.grpc.netty.shaded.io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:830)

```

dapengzhang0:
From the above stack trace there is little information we can find. Apparently the connection is broken, but we can't see details. 
Can you turn on java logging on both the client and server side?

```
.level = ALL
io.grpc.netty.level = FINE
io.netty.handler.codec.http2.Http2FrameLogger.level = FINE
[...]

PR #7041, opened 2020-05-14
netty: log SocketExceptions at FINE, too

AgentK20:
This PR changes the `NettyServerTransport#getLogLevel` method to log `SocketException`s to `LogLevel.FINE`, rather than exclusively pure IOExceptions. This may fix an unintentional regression introduced in c166ec2c4e0fa88ea9dfa6e5766ea04d80a21c8b, although the message in my java version (14.0.1) wouldn't have passed the old logic for quieting either. This also fixes the issue raised in https://github.com/grpc/grpc-java/issues/6423 that was locked for inactivity.

This fixes 
```
[2020/05/14 20:21:52   INFO] [io.grpc.netty.NettyServerTransport.connections] Transport failed
java.net.SocketException: Connection reset
	at java.base/sun.nio.ch.SocketChannelImpl.throwConnectionReset(SocketChannelImpl.java:345)
	at java.base/sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:376)
	at io.netty.buffer.PooledUnsafeDirectByteBuf.setBytes(PooledUnsafeDirectByteBuf.java:288)
	at io.netty.buffer.AbstractByteBuf.writeBytes(AbstractByteBuf.java:1125)
	at io.netty.channel.socket.nio.NioSocketChannel.doReadBytes(NioSocketChannel.java:347)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:148)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:677)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:612)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:529)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:491)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:905)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:832)
```
being logged to level INFO, which occurs whenever a socket is improperly closed by the client, such as with the grpc-health-probe (They've got an [open issue](https://github.com/grpc-ecosystem/grpc-health-probe/issues/34) for this)

### DIFF
--- a/netty/src/main/java/io/grpc/netty/NettyServerTransport.java
+++ b/netty/src/main/java/io/grpc/netty/NettyServerTransport.java
@@ -38,6 +38,7 @@ import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
 import java.io.IOException;
 import java.net.SocketAddress;
+import java.net.SocketException;
 import java.util.List;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.logging.Level;
@@ -186,6 +187,7 @@ class NettyServerTransport implements ServerTransport {
   @VisibleForTesting
   static Level getLogLevel(Throwable t) {
     if (t.getClass().equals(IOException.class)
+        || t.getClass().equals(SocketException.class)
         || QUIET_EXCEPTIONS.contains(t.getClass().getSimpleName())) {
       return Level.FINE;
     }
--- a/netty/src/test/java/io/grpc/netty/NettyServerTransportTest.java
+++ b/netty/src/test/java/io/grpc/netty/NettyServerTransportTest.java
@@ -22,6 +22,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 
 import java.io.IOException;
+import java.net.SocketException;
 import java.util.logging.Level;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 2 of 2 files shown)


creamsoup:
"Connection reset" won't be filtered by previous logic because it does not contains "Connection reset by peer".

@AgentK20 just looking at the title of the original issue, "grpc-health-probe shuts down the gRPC connection non-gracefully", it seems like the log level is proper. does it close connection gracefully?

AgentK20:
They haven't fixed the issue in grpc-health-probe yet, no. This occurs when _any_ gRPC client terminates the connection without properly cleaning it up though, such as when a process is terminated. Since these are relatively "normal" events I think they'd be more appropriately logged at FINE.

A workaround on my end could be to just disable logging of `io.grpc.netty.NettyServerTransport.connections` but I'm not sure if there's any other useful information that may be obscured if I do that.

creamsoup:
okay, we don't want to encourge to disable logging and this seems in the same ballpark of the connection reset. can you add a simple test to prevent future regression? something like this https://github.com/grpc/grpc-java/blob/master/netty/src/test/java/io/grpc/netty/NettyServerTransportTest.java#L45

creamsoup:
thanks @AgentK20, this will be available in `1.30.0` scheduled 6/2/20.

AgentK20:
Happy to help. Thanks guys for the quick response!

Issue #6761, opened 2020-02-26
Profiling for crypter's decryption for ALTS Channel.


veblush:
Under heavy traffic using ALTS, I briefly profiled the client to see where the cpu time went through VisualVM. It seems that it spent significant time running `javax.crypto.ShortBufferException.<init> ()`

**Hot spots (VisualVM)**
```
io.netty.channel.epoll.Native.epollWait0[native] ()	23,995 ms (89.6%)	23,995 ms (9.4%)
org.conscrypt.NativeCrypto.EVP_AEAD_CTX_open[native] ()	994 ms (3.7%)	994 ms (0.4%)
io.netty.channel.unix.FileDescriptor.readAddress[native] ()	891 ms (3.3%)	891 ms (0.4%)
> javax.crypto.ShortBufferException.<init> ()	390 ms (1.5%)	390 ms (0.2%)
org.conscrypt.NativeCrypto.EVP_aead_aes_128_gcm[native] ()	208 ms (0.8%)	208 ms (0.1%)
io.netty.util.internal.PlatformDependent0.copyMemoryWithSafePointPolling ()	201 ms (0.8%)	201 ms (0.1%)
```

**Callstack for javax.crypto.ShortBufferException.<init> (VisualVM)**
```
io.grpc.alts.internal.AltsTsiFrameProtector$Unprotector.handlePayload ()	296 ms (4.5%)	296 ms (4.5%)
 ...
  io.grpc.alts.internal.AesGcmAeadCrypter.decryptAad ()	196 ms (3%)	196 ms (3%)
   javax.crypto.Cipher.doFinal ()	196 ms (3%)	196 ms (3%)
    javax.crypto.CipherSpi.engineDoFinal ()	196 ms (3%)	196 ms (3%)
     javax.crypto.CipherSpi.bufferCrypt ()	196 ms (3%)	196 ms (3%)
      org.conscrypt.OpenSSLCipher$EVP_AEAD.engineDoFinal ()	196 ms (3%)	196 ms (3%)
       org.conscrypt.OpenSSLCipher.engineDoFinal ()	98.8 ms (1.5%)	98.8 ms (1.5%)
        javax.crypto.ShortBufferException.<init> ()	97.2 ms (1.5%)	97.2 ms (1.5%)
```

[...]

PR #7049, opened 2020-05-19
Explicit ALTS buffer management to avoid too many ShortBufferException

veblush:
To avoid having too many `ShortBufferException` thrown in ALTS code path on Java 8, we came up with this workaround creating new managed buffer, filling it, and passing it to underlying Conscrypt not to hit the code path throwing the exception. This might look to introduce another inefficiency but it's more like making it explicit because Conscrypt will do for non-managed buffer which gRPC uses.

Fix: #6761


veblush:
Reviewers: jiangtaoli2016, beschmi

jiangtaoli2016:
@beschmi

Issue #7051, opened 2020-05-20
Support for proto3 field presence


stanley-cheung:
Proto3 introduced the `optional` syntax to support [field presence](https://github.com/protocolbuffers/protobuf/blob/v3.12.0/docs/field_presence.md) tracking in [release 3.12.0](https://github.com/protocolbuffers/protobuf/releases/tag/v3.12.0). We need to add support in our grpc protoc plugin.

See [this doc](https://github.com/protocolbuffers/protobuf/blob/master/docs/implementing_proto3_presence.md#signaling-that-your-code-generator-supports-proto3-optional) for more details. Basically we need to add a function `GetSupportedFeatures()` to our grpc protoc plugin to mark our compliance.

Please plan to do this in the next release. If not, please let me know. Thanks!

stanley-cheung:
FYI, grpc/grpc PR: https://github.com/grpc/grpc/pull/22998

dapengzhang0:
Do we want the protoc flag `--experimental_allow_proto3_optional` on by default?

stanley-cheung:
Yes please turn it on by default. At first I thought the flag only applies to `protoc`. Looks like it applies to the plugin as well (i.e. will fail the plugin if .proto has the new syntax and the flag is not passed in)? Then yes we do want to turn on the flag by default. Thanks!

dapengzhang0:
Fixed by #7052 and #7054 . Backported to v1.30.x with #7058 .

PR #7052, opened 2020-05-20
all: bump protobuf to 3.12.0

dapengzhang0:
Version 3.12.0 is needed for the feature of support for proto3 field presence #7051

PR #7054, opened 2020-05-21
compiler: support proto3 field presence for codegen

dapengzhang0:
Support proto3 field presence #7051.

This is the same change as https://github.com/grpc/grpc/pull/22998/files#diff-6c29ff8771ae54ddf2c14e2ef3eb554fR37-R44


dapengzhang0:
If we want backport this, we also need backport #7052 . That means we would use protobuf 3.12.0 for 1.30.0

ejona86:
Yeah, @stanley-cheung mentioned in #7051 that we were trying to do this for the v1.30 release.

PR #7059, opened 2020-05-22
xds: replace generic with individual client and server SslContextProviders

sanjaypujare:


xds/src/main/java/io/grpc/xds/internal/sds/SdsSslContextProviderHelper.java
```diff
@@ -178,9 +104,9 @@ public void addCallback(Callback callback, Executor executor) {
   }
 
   private void callPerformCallback(
-      Callback callback, Executor executor, final SslContext sslContextCopy) {
-    performCallback(
-        new SslContextGetter() {
+      SslContextProvider.Callback callback, Executor executor, final SslContext sslContextCopy) {
+    sslContextProvider.performCallback(
```
creamsoup, line 108: helper using helpee is weird. is there way to avoid this circular dependency? btw, this class makes sense as an abstract class, but not as a helper class.

sanjaypujare: Yes I realized that but Java doesn't have multiple inheritance (unlike C++) so my only alternative was to inject the helper to achieve code reuse - I really would have hated code duplication.

I was also aware of the Helper class calling back into the caller of the helper but again I thought this is the best way to structure the code without unnecessarily duplicating it. Let me see if I can achieve the same result by restructuring this.

xds/src/main/java/io/grpc/xds/internal/sds/SslContextProvider.java
```diff
@@ -97,6 +77,20 @@ protected void setClientAuthValues(
     }
   }
 
+  /** Returns the DownstreamTlsContext in this SslContextProvider if this is server side. **/
+  public DownstreamTlsContext getDownstreamTlsContext() {
+    checkState(tlsContextHolder instanceof DownstreamTlsContextHolder,
+        "expected DownstreamTlsContextHolder");
+    return ((DownstreamTlsContextHolder)tlsContextHolder).getDownstreamTlsContext();
```
creamsoup, line 84:
```suggestion
    return ((DownstreamTlsContextHolder) tlsContextHolder).getDownstreamTlsContext();
```

xds/src/main/java/io/grpc/xds/internal/sds/SslContextProvider.java
```diff
@@ -97,6 +77,20 @@ protected void setClientAuthValues(
...
+  }
+
+  /** Returns the UpstreamTlsContext in this SslContextProvider if this is client side. **/
+  public UpstreamTlsContext getUpstreamTlsContext() {
+    checkState(tlsContextHolder instanceof UpstreamTlsContextHolder,
+        "expected UpstreamTlsContextHolder");
+    return ((UpstreamTlsContextHolder)tlsContextHolder).getUpstreamTlsContext();
```
creamsoup, line 91:
```suggestion
    return ((UpstreamTlsContextHolder) tlsContextHolder).getUpstreamTlsContext();
```

xds/src/main/java/io/grpc/xds/internal/sds/UpstreamTlsContextHolder.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package io.grpc.xds.internal.sds;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import io.envoyproxy.envoy.api.v2.auth.CommonTlsContext;
+import io.envoyproxy.envoy.api.v2.auth.UpstreamTlsContext;
+
+class UpstreamTlsContextHolder implements TlsContextHolder {
```
creamsoup, line 24: can it be a final class?

xds/src/main/java/io/grpc/xds/internal/sds/UpstreamTlsContextHolder.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public UpstreamTlsContext getUpstreamTlsContext() {
```
creamsoup, line 26: nit: move it before/after L37?

PR #7061, opened 2020-05-26
see if travis likes to upgrade openjdk like this

ejona86:


ejona86:
:-/ I _know_ I changed the repo to be my repo. Don't know why it came up here.

PR #7068, opened 2020-05-27
deps: prepare to bump JUnit to 4.13

### LINKED ISSUE #7053: Bump Guava to 29.0-android
Once we use protobuf-java v3.12.0 #7052, the upper bound of guava version will be `29.0-android`.

elharo:
@dapengzhang0


dapengzhang0:
@elharo Would you please fix failed test?

dapengzhang0:
> I don't want to revert all this work. We can have a new PR to update Guava and then use this to update JUnit.

I'm okay for a new PR. But the junit fix in this PR still has some problems.

dapengzhang0:
> Also, why do we depend on guava-testlib?
Some unit tests are using some utilities in guava-testlib.

dapengzhang0:
@elharo We may not want to upgrade junit to 4.13 as of now, since internally we still use 4.12.

elharo:
google-cloud-java and the rest of the open source GCP Java projects have pretty much already all moved to JUnit 4.13. We're not using anything newly introduced in 4.13 so internal google repos don't block this.

dapengzhang0:
Had a discussion with @ejona86 @voidzcy @sanjaypujare , we would like to keep the existing `ExpectedException` unchanged and only suppress the checker by adding `@SuppressWarnings("ExpectedExceptionChecker")`.  In the future, when grpc migrates to Java 8, we can replace them with `assertThrows()` as recommended by the deprecation javadoc.

ejona86:
Closing in favor of #7468. Much less invasive. It'll be good when we can swap to assertThrows when it is convenient.

Thank you for this PR, even if it didn't get merged. I know it was quite annoying to swap to the try-fail approach. #7079 helped as well; I didn't have any conflicts when I bumped the junit version for testing.

PR #7106, opened 2020-06-09
Fix client call close race condition message leak

### LINKED ISSUE #3557: ClientCalls.ThreadlessExecutor should reject Runnables after end of RPC
As seen in #3537, the executor was being used for multiple RPCs which caused some Runnables to be queued but never run. We are in a position to know they will never execute though, so it'd be nice to have a (possibly unreliable) detection of rejected Runnables to notice issues like this earlier.

This could be accomplished with just a volatile `shutdown` boolean that is checked before `execute()` and a `shutdown()` method. That's racy, but we're just wanting it to detect bugs, not provide correctness.

### LINKED ISSUE #7105: Possible buffer leak in blocking stub DEADLINE_EXCEEDED race condition
One of my unit tests started reporting a netty `ByteBuf` leak when upgrading from 1.26.x to 1.27.0, and it remains in subsequent versions up to 1.29.0. I'm not 100% sure but strongly suspect it's related to #6328, and in particular the interplay of that change with the `ThreadlessExecutor` used by blocking unary calls.

It's a somewhat extreme case, the test itself was there to verify a past unrelated concurrency issue. Basically making many consecutive calls using blocking stub with very short (1ms) deadlines. Though likely rare in the wild I think it does expose a real bug.

Here is example snippet which reproduces the leak error log(s) every time mid-test for me when pointed at a local etcd server:
```java
        ManagedChannel channel = NettyChannelBuilder.forTarget("localhost:2379")
                .usePlaintext().build();
        KVBlockingStub stub = KVGrpc.newBlockingStub(channel);
        for (int i = 0; i < 6000; i++) {
            try {
                stub.withDeadlineAfter(1L, TimeUnit.MILLISECONDS)
                    .put(PutRequest.newBuilder().setValue(bs("some-value"))
                        .setKey(bs("test/" + UUID.randomUUID().toString())).build());
            } catch (Exception e) {
                Status status = Status.fromThrowable(e);
                // Throws a mix of server and client-side DEADLINE_EXCEEDED errors
            }
        }
```

```
[grpc-nio-worker-ELG-1-7] ERROR io.netty.util.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
Created at:
	io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:363)
	io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:187)
        ...
```
I can provide more of the leak access records if needed.

njhill:
As reported in #7105. Not sure if this is how you want it done, but it does fix the problem.

Fixes #7105
Fixes #3557

### DIFF
--- a/core/src/main/java/io/grpc/internal/AbstractClientStream.java
+++ b/core/src/main/java/io/grpc/internal/AbstractClientStream.java
@@ -311,7 +311,10 @@ public abstract class AbstractClientStream extends AbstractStream
      * @param headers the parsed headers
      */
     protected void inboundHeadersReceived(Metadata headers) {
-      checkState(!statusReported, "Received headers on closed stream");
+      // checkState(!statusReported, "Received headers on closed stream");
+      if (statusReported) {
+        return; // possible due to call close race
+      }
       statsTraceCtx.clientInboundHeaders();
 
       boolean compressedStream = false;
--- a/core/src/main/java/io/grpc/internal/ClientCallImpl.java
+++ b/core/src/main/java/io/grpc/internal/ClientCallImpl.java
@@ -53,6 +53,7 @@ import java.io.InputStream;
 import java.nio.charset.Charset;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.Executor;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
@@ -402,7 +403,11 @@ final class ClientCallImpl<ReqT, RespT> extends ClientCall<ReqT, RespT> {
         new LogExceptionRunnable(new DeadlineExceededSendCancelToServerTimer()),
         DEADLINE_EXPIRATION_CANCEL_DELAY_NANOS,
         TimeUnit.NANOSECONDS);
-    executeCloseObserverInContext(observer, status);
+    try {
+      executeCloseObserverInContext(observer, status);
+    } catch (RejectedExecutionException ree) {
+      // call close race
+    }
   }
 
   private void executeCloseObserverInContext(final Listener<RespT> observer, final Status status) {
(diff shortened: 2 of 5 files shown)


njhill:
I guess another option instead of throwing/handling `RejectedExecutionException` might be to just submit to the channel's executor instead (exactly where the REE is thrown in the changes as they are now), though it doesn't look like there's currently a way to access it.

That should be safe since the runnable is always a `SerializingExecutor` here.

njhill:
Thanks @ejona86, I see that the more general part of this is a bit tricky. I wonder if a fallback executor could be avoided. There are two realistic cases I think - temporary thread/queue exhaustion and (permanent) premature executor shutdown.

The latter case constitutes an app bug where all bets are off since the channel is then "broken" at least for async usage. It seems reasonable in this case to just log the error and not run any call close callbacks.

To accommodate the former case, you could schedule a repeating task to re-attempt scheduling of the `SerializingExecutor` onto the user's executor at some interval (like 2sec), after cancelling the call and ensuring to run any non-callback cleanup logic inline (including releasing messages after an `onMessage` is rejected). This has advantages of not requiring a new executor, and of callbacks not being made from an unexpected executor which could have other implications for the app.

Irrespective of that, would a subset of this PR still be worthwhile (specifically the changes to `ThreadlessExecutor`) since they address #3557 and do so in a non-racy way? Of course the original bug would need to be either fixed or reverted first!

ejona86:
@njhill, since #7105 was resolved by reverting the problematic commit, what should we do with this PR? I do think the ThreadlessExecutor changes look good, and given my comment on #3557 it looks like that is safe to do now. Are there other things we want to save?

njhill:
@ejona86 shall I open another PR with just the `ThreadlessExecutor` changes?

ejona86:
@njhill, yeah, that sounds good.

njhill:
@ejona86 PTAL #7798

PR #7108, opened 2020-06-09
alts: add peer attribute access for the ALTS context

d-reidenbach:
@ZhenLian For code review

Added and tested accessing peer attributes map from the handshaker result

### DIFF
--- a/alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
+++ b/alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
@@ -40,13 +40,13 @@ public final class AltsAuthContextTest {
   private static final String TEST_PEER_ATTRIBUTES_KEY = "peer";
   private static final String TEST_PEER_ATTRIBUTES_VALUE = "attributes";
 
-  private Map<String, String> testPeerAttributes = new HashMap<String, String>();
-  
+  private Map<String, String> testPeerAttributes;
   private HandshakerResult handshakerResult;
   private RpcProtocolVersions rpcVersions;
 
   @Before
   public void setUp() {
+    testPeerAttributes = new HashMap<String, String>();
     testPeerAttributes.put(TEST_PEER_ATTRIBUTES_KEY, TEST_PEER_ATTRIBUTES_VALUE);
     rpcVersions =
         RpcProtocolVersions.newBuilder()
--- a/alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
+++ b/alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
@@ -18,10 +18,14 @@ package io.grpc.alts.internal;
 
 import static org.junit.Assert.assertEquals;
 
+
+import java.util.HashMap;
+import java.util.Map;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
+
 
 /** Unit tests for {@link AltsAuthContext}. */
 @RunWith(JUnit4.class)
@@ -35,6 +39,8 @@ public final class AltsAuthContextTest {
   private static final String TEST_LOCAL_SERVICE_ACCOUNT = "local@gserviceaccount.com";
   private static final String TEST_PEER_SERVICE_ACCOUNT = "peer@gserviceaccount.com";
   private static final String TEST_RECORD_PROTOCOL = "ALTSRP_GCM_AES128";
+
+  private static final Map<String, String> TEST_PEER_ATTRIBUTES = new HashMap<String, String>();
 
   private HandshakerResult handshakerResult;
   private RpcProtocolVersions rpcVersions;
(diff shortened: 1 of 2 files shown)


alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
```diff
@@ -36,6 +40,8 @@
   private static final String TEST_PEER_SERVICE_ACCOUNT = "peer@gserviceaccount.com";
   private static final String TEST_RECORD_PROTOCOL = "ALTSRP_GCM_AES128";
 
+  private static final Map<String, String> TEST_PEER_ATTRIBUTES = new HashMap<String, String>();
```
ZhenLian, line 43: Can we use a local variable instead of a static variable?

alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
```diff
@@ -54,11 +60,14 @@ public void setUp() {
                     .setMinor(TEST_MIN_RPC_VERSION_MINOR)
                     .build())
             .build();
+    Identity.Builder peerIdentity = Identity.newBuilder()
+        .setServiceAccount(TEST_PEER_SERVICE_ACCOUNT);
+    peerIdentity.putAllAttributes(TEST_PEER_ATTRIBUTES);
```
ZhenLian, line 65: If we define the map to be a local variable, it doesn't need to be capitalized. We usually do that for some static strings. We can also add some fake values and see if what we got equals what we expect. The key and value of that fake value could be capitalized as a static string.

alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
d-reidenbach: I was not sure so I just made it static since that is what the other sample data was to test the other methods. Do you think I should make it local?

alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
d-reidenbach: Added thanks!

alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
```diff
@@ -18,11 +18,15 @@
 
 import static org.junit.Assert.assertEquals;
 
+
```
ZhenLian, line 21: nit: remove the empty line.

alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
```diff
@@ -18,11 +18,15 @@
...
+import java.util.HashMap;
+import java.util.Map;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
+
```
ZhenLian, line 29: nit: empty line

alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
```diff
@@ -36,11 +40,15 @@
   private static final String TEST_PEER_SERVICE_ACCOUNT = "peer@gserviceaccount.com";
   private static final String TEST_RECORD_PROTOCOL = "ALTSRP_GCM_AES128";
 
+  private Map<String, String> testPeerAttributes = new HashMap<String, String>();
+
```
ZhenLian, line 44: nit: empty line

alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
```diff
@@ -36,11 +40,15 @@
...
 
+  private Map<String, String> testPeerAttributes = new HashMap<String, String>();
+  
+
   private HandshakerResult handshakerResult;
   private RpcProtocolVersions rpcVersions;
 
   @Before
   public void setUp() {
+    testPeerAttributes.put("peer", "attributes");
```
ZhenLian, line 51: How about this: we make the map local, but create two static string like TEST_PEER_ATTRIBUTES_KEY and TEST_PEER_ATTRIBUTES_VALUE.

alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
ZhenLian: Yeah I think so. IMO, the reason why we want to make some test data, like TEST_RECORD_PROTOCOL, static is because these strings might be used multiple times in a place, and it's better to use a `static final` string to refer them, to avoid misspelling,  etc. This is similar to a `const string` in C++.
Map doesn't have this issue, so we'd better keep it local. However, we could make the key and value static. See my inline comments for more.

alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
d-reidenbach: will fix!

alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
d-reidenbach: good catch, fixed in update

alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
ZhenLian: Is this empty line removed? 
Also, does it make sense to not initializing the map here, but moving the initialization to the place where we do SetUp()? That might make the code cleaner IMO.

ZhenLian:
@jiangtaoli2016 Can you please also take a look at this PR please? Thank you so much!

alts/src/test/java/io/grpc/alts/internal/AltsAuthContextTest.java
d-reidenbach: Yeah that makes sense. I believe that was a fragment from when I had the map along with the static variables but good catch

ZhenLian:
@sanjaypujare Hi Sanjay, could you please take a 3rd round of review on the Java coding style, etc? This PR is to add a getter function for class ALTS context. Thank you so much!

ZhenLian:
Thanks all for the review! Merging now...

PR #7131, opened 2020-06-16
xds: replace Envoy's DownstreamTlsContext with internally defined DownstreamTlsContext

sanjaypujare:


xds/src/main/java/io/grpc/xds/internal/sds/ServerSslContextProviderFactory.java
```diff
@@ -35,8 +34,8 @@
   public SslContextProvider createSslContextProvider(
       DownstreamTlsContext downstreamTlsContext) {
     checkNotNull(downstreamTlsContext, "downstreamTlsContext");
-    checkArgument(
-        downstreamTlsContext.hasCommonTlsContext(),
+    checkNotNull(
+        downstreamTlsContext.getCommonTlsContext(),
```
creamsoup, line 38: `DownstreamTlsContext.getCommonTlsContext()` is not marked `@Nullable`. what happens if getCommonTlsContext is a default instance?

sanjaypujare: This is an unlikely scenario or an error in the control plane (for sending faulty Listener data to us). So a Listener has filter-chains and a filter-chain with a "tls" named transport-socket has the DownstreamTlsContext payload. For the scenario you describe to occur, the "tls" named transport-socket should have a default-instance DownstreamTlsContext and/or that DownstreamTlsContext should have a default-instance CommonTlsContext. If the transport-socket is *not* going to be TLS why create the "tls" named transport socket? 

In any case let me look at the code and do a quick check to ensure default-instances are handled properly.

sanjaypujare: Okay, so I ran some tests and default-instances do result into exceptions (in `SdsProtocolNegotiators.ServerSdsHandler`). What I am planning to do is to handle these properly in `SdsProtocolNegotiators.ServerSdsHandler` (as well as `SdsProtocolNegotiators.ClientSdsHandler`) but that will be a separate PR (it can be opened in parallel so you can see it).

sanjaypujare: My other changes depend on changes in this PR, so I will make them once this is done. Works?

creamsoup: sure.

sanjaypujare: So once this PR is merged I am planning to address the exception I mentioned [above](https://github.com/grpc/grpc-java/pull/7131#discussion_r441212806) just to clarify.

PR #7141, opened 2020-06-18
rls: request factory prepends leading '/'

creamsoup:


rls/src/main/java/io/grpc/rls/RlsRequestFactory.java
```diff
@@ -73,11 +73,11 @@
   RouteLookupRequest create(String service, String method, Metadata metadata) {
     checkNotNull(service, "service");
     checkNotNull(method, "method");
-    String path = service + "/" + method;
+    String path = "/" + service + "/" + method;
     Map<String, NameMatcher> keyBuilder = keyBuilderTable.row(path);
     // if no matching keyBuilder found, fall back to wildcard match (ServiceName/*)
     if (keyBuilder.isEmpty()) {
-      keyBuilder = keyBuilderTable.row(service + "/*");
+      keyBuilder = keyBuilderTable.row("/" + service + "/*");
```
dapengzhang0, line 80: Also need update javadoc of `keyBuilderTable`.

creamsoup: good catch. done.

creamsoup:
> Didn't interop test detect this issue?

this is because the rls server of manual integration test is written by me as well (both were wrong in consistent way). discovered this when using CBT rls server.

rls/src/main/java/io/grpc/rls/RlsRequestFactory.java
```diff
@@ -41,7 +41,7 @@
 
   private final String target;
   /**
-   * schema: Path(serviceName/methodName or serviceName/*), rls request headerName, header fields.
+   * schema: Path(/serviceName/methodName or serviceName/*), rls request headerName, header fields.
```
dapengzhang0, line 44: `/serviceName/*` :)

PR #7146, opened 2020-06-19
core: add PerfMark tasks/events for MigratingThreadDeframer

creamsoup:


core/src/main/java/io/grpc/internal/MigratingThreadDeframer.java
```diff
@@ -126,23 +140,33 @@ public void request(final int numMessages) {
...
+                requestFromTransportThread(numMessages);
+              } finally {
+                PerfMark.stopTask("MigratingThreadDeframer.request");
+              }
             }
           });
           return;
         }
         try {
+          PerfMark.startTask("MigratingThreadDeframer.request");
```
ejona86, line 163: Move this before the try?

PR #7158, opened 2020-06-24
github: Remove bug label from newly filed bugs

ejona86:
We've been using the bug label for "confirmed" bugs. Many issues filed as bugs
turn out not to be bugs, but we've not generally remembered to remove the bug
label. This has been causing trouble for tracking bug closure rate, as the data
is now mostly garbage. This change is to put us back into our old flow where
confirmed bugs have the label.

CC @srini100

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,7 +1,6 @@
 ---
 name: Report a bug
 about: Create a bug report to help us improve
-labels: bug
 ---
 
 <!-- Please answer these questions before submitting a bug report. -->


ejona86:
A special label sounds fine. It's not clear how we're using these labels though, and it seems half of the user issues just get a "random" label. Are you finding the auto-applied labels useful?

sanjaypujare:
If the user picks the "file a bug" template then we at least have the user's intent captured in the "potential bug" label. We can then change it to "bug" or remove the label after triaging.

ejona86:
We have the intent captured in the issue description. Are we saying we need a quick-glance form of that intent? When will we benefit from it?

I honestly don't think we can trust users to determine what is an "enhancement" or even "question". "Question" is useful for filtering, but after triage many of them turn into bugs or enhancements.

If we want the label "just so that we know the issue is labeled," that could be fine. Although I don't think we have any queries today to track unlabeled issues.

sanjaypujare:
The current "New Issue" flow requires the user to choose from bug/question/feature request. With this change "no label" effectively implies "potential bug" so I am okay with this.

ejona86:
The template is different for each. The text will be different. But my point is it doesn't really even matter what the user chooses, until we triage it it could be anything.

creamsoup:
when it was added, the goal was less work on our end. we've learned that the auto label doesn't save us any time, so removing it makes sense. other intermediate label sounds cool, but it doesn't really provide any benefit for us.

PR #7160, opened 2020-06-24
xds: import checked.proto from Googleapis' repository

cindyxue:
This PR does the following thing:

Added dependency needed for checked.proto which is used to create a mock Cel interface in Java.

@ZhenLian


ZhenLian:
@voidzcy Could you please also take a look at this PR please? Thanks!

PR #7167, opened 2020-06-29
okhttp: not using MigratingThreadDeframer

creamsoup:


creamsoup:
stack trace when the mentioned race happens (OkHttp on android env)

```
java.lang.IllegalStateException: Headers should be received prior to messages.
at com.google.common.base.Preconditions.checkState(Preconditions.java:511)
at io.grpc.internal.RetriableStream$Sublistener.messagesAvailable(RetriableStream.java:952)
at io.grpc.internal.DelayedStream$DelayedStreamListener$1.run(DelayedStream.java:438)
at io.grpc.internal.DelayedStream$DelayedStreamListener.drainPendingCallbacks(DelayedStream.java:510)
at io.grpc.internal.DelayedStream.drainPendingCalls(DelayedStream.java:168)
at io.grpc.internal.DelayedStream.setStream(DelayedStream.java:132)
at io.grpc.internal.DelayedClientTransport$PendingStream.createRealStream(DelayedClientTransport.java:358)
at io.grpc.internal.DelayedClientTransport$PendingStream.access$300(DelayedClientTransport.java:341)
at io.grpc.internal.DelayedClientTransport$5.run(DelayedClientTransport.java:300)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
at java.lang.Thread.run(Thread.java:818)
```

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -101,6 +101,11 @@
             transport,
             initialWindowSize,
             method.getFullMethodName());
+
+    // TODO(ejona86) Fix a potential data race using MigratingThreadDeframer in OkHttp.
```
ejona86, line 105: You probably need the colon after the TODO. It doesn't matter, but I assume there is some linter somewhere that will complain about that, because somehow that makes us more productive.

okhttp/src/main/java/io/grpc/okhttp/OkHttpClientStream.java
```diff
@@ -101,6 +101,11 @@
...
+    // OkHttp is lock based and it doesn't get benefit from using MigratingThreadDeframer,
```
ejona86, line 106: It did actually see a benefit. So let's just say we're disabling it for the race.

creamsoup:
ugh.. i pressed wrong key and it merged before adding the note.

ejona86:
It is in the referenced issue in the TODO, so that'll be fine.

PR #7172, opened 2020-06-30
stub: Redefine exception handling for StreamObserver

ejona86:
This aligns the documentation with our recommendations expressed elsewhere
("don't throw"). Exceptions are possible, but having a try-catch around each
onNext() would be an anti-pattern for user code as it would be guaranteed to be
buggy. We would much rather have higher-level exception handling, covering many
application statements, call onError() if any of them throw an exception
because the application should generally cancel independent of the source of
the exception.

### DIFF
--- a/stub/src/main/java/io/grpc/stub/StreamObserver.java
+++ b/stub/src/main/java/io/grpc/stub/StreamObserver.java
@@ -49,7 +49,9 @@ public interface StreamObserver<V>  {
    *
    * <p>Although implementations are expected to not throw exceptions, if an exception occurs the
    * caller is encouraged to call {@link #onError} as part of normal application exception handling.
-   * gRPC will attempt to call {@code onError()} when an application's implementation throws.
+   * That is to say, {@code onError} should be called if any related part of the application throws
+   * an exception, generally independent of the source, to avoid leaks. gRPC will call {@code
+   * onError()} when an application's implementation throws.
    *
    * <p>As an exception to the rule and for historical reasons, on server-side, {@code onNext()} may
    * throw with a {@code StatusRuntimeException} if the client cancels the call after {@code
--- a/stub/src/main/java/io/grpc/stub/StreamObserver.java
+++ b/stub/src/main/java/io/grpc/stub/StreamObserver.java
@@ -31,6 +31,10 @@ package io.grpc.stub;
  * not need to be synchronized together; incoming and outgoing directions are independent.
  * Since individual {@code StreamObserver}s are not thread-safe, if multiple threads will be
  * writing to a {@code StreamObserver} concurrently, the application must synchronize calls.
+ *
+ * <p>Implementations are expected to not throw exceptions, except when called incorrectly. For
+ * example, a call to {@code onNext()} after {@code onCompleted()} may throw. {@link Error}s may
+ * also still occur.
  */
 public interface StreamObserver<V>  {
   /**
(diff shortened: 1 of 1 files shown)


stub/src/main/java/io/grpc/stub/StreamObserver.java
```diff
@@ -43,9 +47,18 @@
...
-   * <p>If an exception is thrown by an implementation the caller is expected to terminate the
-   * stream by calling {@link #onError(Throwable)} with the caught exception prior to
-   * propagating it.
+   * <p>Although implementations are expected to not throw exceptions, if an exception occurs the
+   * caller is encouraged to call {@link #onError} as part of normal application exception handling.
+   * That is to say, {@code onError} should be called if any related part of the application throws
+   * an exception, generally independent of the source, to avoid leaks. gRPC will call {@code
+   * onError()} when an application's implementation throws.
+   *
+   * <p>As an exception to the rule and for historical reasons, on server-side, {@code onNext()} may
```
dapengzhang0, line 56: Here `{@code onNext()}` is about the response stream observer, and `{@code onCompleted()}` at the end of this sentence is about the request stream observer?

ejona86: Actually, I realize that is wrong. It is unconditional. Fixed, which side-steps the confusing language.

stub/src/main/java/io/grpc/stub/StreamObserver.java
```diff
@@ -43,9 +47,17 @@
    * server streaming calls, but may receive many onNext callbacks.  Servers may invoke onNext at
    * most once for client streaming calls, but may receive many onNext callbacks.
    *
-   * <p>If an exception is thrown by an implementation the caller is expected to terminate the
-   * stream by calling {@link #onError(Throwable)} with the caught exception prior to
-   * propagating it.
+   * <p>Although implementations are expected to not throw exceptions, if an exception occurs the
```
jdcormie, line 50: Can you avoid the passive voice and just say "implementations should not throw" (here and in 3 other places)?

stub/src/main/java/io/grpc/stub/StreamObserver.java
```diff
@@ -43,9 +47,17 @@
...
+   * caller is encouraged to call {@link #onError} as part of normal application exception handling.
+   * That is to say, {@code onError} should be called if any related part of the application throws
+   * an exception, generally independent of the source, to avoid leaks. gRPC will call {@code
+   * onError()} when an application's implementation throws.
```
jdcormie, line 54: This is helpful but perhaps it should go in the javadoc for the whole interface since it isn't specific to onNext? You could say:

Owners of a StreamObserver are responsible for eventually calling either onError() or onCompleted() to ensure that  resources associated with the stream (memory, file descriptors, etc) are released.

stub/src/main/java/io/grpc/stub/StreamObserver.java
```diff
@@ -43,9 +47,17 @@
...
+   *
+   * <p>As an exception to the rule and for historical reasons, on server-side, {@code onNext()} may
+   * throw with a {@code StatusRuntimeException} if the call is cancelled. This was necessary
+   * because there was not a callback available to notify the service. Services are encouraged to
```
jdcormie, line 58: Consider omitting "This was necessary because ... " since teaching callers the actual historical reason doesn't seem worth the extra reading

stub/src/main/java/io/grpc/stub/StreamObserver.java
```diff
@@ -54,9 +66,9 @@
   /**
    * Receives a terminating error from the stream.
    *
-   * <p>May only be called once and if called it must be the last method called. In particular if an
-   * exception is thrown by an implementation of {@code onError} no further calls to any method are
-   * allowed.
+   * <p>May only be called once and if called it must be the last method called. Although
+   * implementations are expected to not throw exceptions, if an exception is thrown by {@code
+   * onError()} further calls to the observer should be avoided.
```
njhill, line 71: Saying "further calls should be avoided" sounds a bit odd/redundant to me coming immediately after "must be the last method called" (same for `onCompleted()`)

stub/src/main/java/io/grpc/stub/StreamObserver.java
```diff
@@ -43,9 +47,17 @@
    * server streaming calls, but may receive many onNext callbacks.  Servers may invoke onNext at
    * most once for client streaming calls, but may receive many onNext callbacks.
    *
-   * <p>If an exception is thrown by an implementation the caller is expected to terminate the
-   * stream by calling {@link #onError(Throwable)} with the caught exception prior to
-   * propagating it.
+   * <p>Although implementations are expected to not throw exceptions, if an exception occurs the
+   * caller is encouraged to call {@link #onError} as part of normal application exception handling.
```
njhill, line 51: This sounds like a subtle change to the contract, specifically that callers are no longer expected to necessarily pass the caught exception itself to `onError`?

PR #7173, opened 2020-06-30
stub: Only throw on cancellation for streaming responses

ejona86:
Unary are far more common than streaming, and we're throwing for unary even
though it doesn't help the service. Let's stop doing that. We also stop
throwing in onComplete() for all cases, because it doesn't help any service;
it doesn't stop the service's processing and isn't even all that informative
since the cancellation can happen even after onComplete() is called.

-------

This is related to #7172, and the docs there would need to be updated. Whichever is merged last will need to be updated.

Note that I saw it is weird that we call both the cancellation and onError callbacks. We should probably just call onError, but that has a greater chance of "being noticed" so I didn't want it to add risk to this PR.

### DIFF
--- a/stub/src/main/java/io/grpc/stub/ServerCalls.java
+++ b/stub/src/main/java/io/grpc/stub/ServerCalls.java
@@ -48,7 +48,7 @@ public final class ServerCalls {
    */
   public static <ReqT, RespT> ServerCallHandler<ReqT, RespT> asyncUnaryCall(
       UnaryMethod<ReqT, RespT> method) {
-    return new UnaryServerCallHandler<>(method);
+    return new UnaryServerCallHandler<>(method, false);
   }
 
   /**
@@ -58,7 +58,7 @@ public final class ServerCalls {
    */
   public static <ReqT, RespT> ServerCallHandler<ReqT, RespT> asyncServerStreamingCall(
       ServerStreamingMethod<ReqT, RespT> method) {
-    return new UnaryServerCallHandler<>(method);
+    return new UnaryServerCallHandler<>(method, true);
   }
 
   /**
@@ -68,7 +68,7 @@ public final class ServerCalls {
    */
   public static <ReqT, RespT> ServerCallHandler<ReqT, RespT> asyncClientStreamingCall(
       ClientStreamingMethod<ReqT, RespT> method) {
-    return new StreamingServerCallHandler<>(method);
+    return new StreamingServerCallHandler<>(method, false);
   }
 
   /**
@@ -78,7 +78,7 @@ public final class ServerCalls {
    */
   public static <ReqT, RespT> ServerCallHandler<ReqT, RespT> asyncBidiStreamingCall(
       BidiStreamingMethod<ReqT, RespT> method) {
-    return new StreamingServerCallHandler<>(method);
+    return new StreamingServerCallHandler<>(method, true);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


stub/src/main/java/io/grpc/stub/ServerCalls.java
```diff
@@ -211,24 +215,26 @@ public void onReady() {
    * @param method an adaptor to the actual method on the service implementation.
    */
   private static <ReqT, RespT> ServerCallHandler<ReqT, RespT> asyncUnaryRequestCall(
```
dapengzhang0, line 217: I feel this method is just adding an annoying redirection. Why not just call `new UnaryServerCallHandler<>(method, serverStreaming)` directly whenever `asyncUnaryRequestCall` is called?

ejona86: That sounds fine. Do you mind if I do that as a follow-up?

ejona86: Actually, let me do it as a pre-refactor, since I was going to let this PR "sit" for us to discuss it in the API meeting.

ejona86:
API review:
 * We feel comfortable with this change. We don't think many/any existing users will be impacted by this change, as the exception is not guaranteed in the face of cancellation (cancellation is racy). We do think it will help users to simplify their code.
 * It would probably be appropriate to point users to setOnCancelHandler [within the cancellation exception](https://github.com/grpc/grpc-java/blob/825c34f6064bf7df49ee6e0f41148cf516c07fe2/stub/src/main/java/io/grpc/stub/ServerCalls.java#L335). Just to tell them it exists and disables that exception, as an alternative to having a try-catch. We think most users would benefit from using the callback instead of the exception
 * We would _like_ to be even more aggressive and remove the cancellation exception entirely, but have varying opinions on whether that is practical (e.g., whether it would break existing users). There are ways to remove the exception in more cases, even if not entirely.

stub/src/main/java/io/grpc/stub/ServerCalls.java
```diff
@@ -331,10 +344,17 @@ public void setCompression(String compression) {
     @Override
     public void onNext(RespT response) {
       if (cancelled) {
-        if (onCancelHandler == null) {
-          throw Status.CANCELLED.withDescription("call already cancelled").asRuntimeException();
+        if (serverStreamingOrBidi) {
+          throw Status.CANCELLED
+              .withDescription("call already cancelled. "
+                  + "ServerCallStreamObserver.setOnCancelHandler() disables this exception")
```
dapengzhang0, line 350: This message is confusing. It seems a hint to the developer to change their source code: "To disable this exception: use ServerCallStreamObserver.setOnCancelHandler()", not an error message to users.

ejona86: Do you have a suggestion for alternative wording? This is what we discussed in the API review. When would users see this exception?

dapengzhang0: > Do you have a suggestion for alternative wording?

No specific alternatives. I think we should just tell the user that this message is a hint to the developers.

ejona86: Decided on:
"Use ServerCallStreamObserver.setOnCancelHandler() to disable this exception"

PR #7174, opened 2020-07-01
stub: Clarify ServerCallStreamObserver cancellation

ejona86:
This avoids the "changes" description, which could be understood to mean that
the transition is important instead of the steady-state and thus a callback
might not be called (which is not the case).

Much of the language was copied from ServerCall.

### DIFF
--- a/stub/src/main/java/io/grpc/stub/ServerCallStreamObserver.java
+++ b/stub/src/main/java/io/grpc/stub/ServerCallStreamObserver.java
@@ -32,19 +32,28 @@ import io.grpc.ExperimentalApi;
 public abstract class ServerCallStreamObserver<V> extends CallStreamObserver<V> {
 
   /**
-   * If {@code true} indicates that the call has been cancelled by the remote peer.
+   * Returns {@code true} when the call is cancelled and the server is encouraged to abort
+   * processing to save resources, since the client will not be processing any further methods.
+   * Cancellations can be caused by timeouts, explicit cancel by client, network errors, and
+   * similar.
    *
    * <p>This method may safely be called concurrently from multiple threads.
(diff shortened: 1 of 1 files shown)


stub/src/main/java/io/grpc/stub/ServerCallStreamObserver.java
```diff
@@ -32,19 +32,28 @@
 public abstract class ServerCallStreamObserver<V> extends CallStreamObserver<V> {
 
   /**
-   * If {@code true} indicates that the call has been cancelled by the remote peer.
+   * Returns {@code true} when the call is cancelled and the server is encouraged to abort
+   * processing to save resources, since the client will not be processing any further methods.
+   * Cancellations can be caused by timeouts, explicit cancel by client, network errors, and
```
dapengzhang0, line 37: s/explicit cancel by client/explicit cancellation by client

stub/src/main/java/io/grpc/stub/ServerCallStreamObserver.java
```diff
@@ -32,19 +32,28 @@
...
    *
    * <p>This method may safely be called concurrently from multiple threads.
    */
   public abstract boolean isCancelled();
 
   /**
-   * Set a {@link Runnable} that will be called if the calls {@link #isCancelled()} state
-   * changes from {@code false} to {@code true}. It is guaranteed that execution of the
-   * {@link Runnable} are serialized with calls to the 'inbound' {@link StreamObserver}.
+   * Set a {@link Runnable} to be called if the call is cancelled and the server is encouraged to
```
dapengzhang0, line 45: s/Set/Sets

PR #7181, opened 2020-07-04
xds: convert and rename ReferenceCountingSslContextProviderMap to generic ReferenceCountingMap

sanjaypujare:


xds/src/main/java/io/grpc/xds/internal/sds/SslContextProvider.java
```diff
@@ -93,7 +93,8 @@ public UpstreamTlsContext getUpstreamTlsContext() {
   }
 
   /** Closes this provider and releases any resources. */
-  void close() {}
+  @Override
+  public void close() {}
```
dapengzhang0, line 97: Can this be abstract method?

xds/src/main/java/io/grpc/xds/internal/sds/ReferenceCountingMap.java
```diff
@@ -101,19 +101,19 @@ private synchronized SslContextProvider releaseInternal(K key, SslContextProvide
...
-  public interface SslContextProviderFactory<K> {
-    SslContextProvider createSslContextProvider(K key);
+  /** A factory to create a value from the given key. */
+  public interface ValueFactory<K, V extends Closeable> {
+    V create(K key);
   }
 
-  private static class Instance {
-    final SslContextProvider sslContextProvider;
+  private static class Instance<V extends Closeable> {
```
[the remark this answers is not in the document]
sanjaypujare: Done

xds/src/main/java/io/grpc/xds/internal/sds/SslContextProvider.java
sanjaypujare: I left it as a default implementation so the subclasses don't need to define it for a "no-op" `close()`. But making abstract and forcing subclasses to define it (since closing to release resources is so important) is okay as well so I made the change

PR #7183, opened 2020-07-06
Upgrade to Gradle 6.5.1

ericgribkoff:
Fixes https://github.com/grpc/grpc-java/issues/7006

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,6 +41,12 @@ subprojects {
         }
     }
 
+    tasks.withType(GenerateModuleMetadata) {
+        // Module metadata, introduced in Gradle 6.0, conflicts with our publishing task for
+        // grpc-alts and grpc-compiler.
+        enabled = false
+    }
+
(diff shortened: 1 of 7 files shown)


ericgribkoff:
@ejona86 Does it make sense to add a JDK13 travis build?

ejona86:
I don't think we need a JDK 13 travis build. Non-LTS releases have been a regular thorn-in-the-side for release branches. We did that for the Java 9-10 series because it added the new ALPN APIs and restrictions around reflection/unsafe. But really, Java 13 is already dead. It'd need to be Java 14. If we saw that Java 14 is a strange and unpredictable platform behaving unlike Java 11, then we could consider taking on the pain of testing it.

But also, this was just a compilation issue, not runtime. Technically speaking we have no requirement to support Java 13/14 for compilation. If we needed to require Java 6 for compilation, I feel that we'd be free to.

ejona86:
The Android failure looks legit.

